PR #145, opened 2019-04-12
fix the future instance is replaced when the flush frequency is high

jixuan1989:
If we call flush() twice (and in the 2nt calling, `value count` = 0), then the first future instance will be replaced. If there is exception in the first flush process, we will have no idea to catch it.

In this PR, we will wait to finish the last flush process before submitting a new one.

### DIFF
--- a/iotdb/src/main/java/org/apache/iotdb/db/engine/bufferwrite/BufferWriteProcessor.java
+++ b/iotdb/src/main/java/org/apache/iotdb/db/engine/bufferwrite/BufferWriteProcessor.java
@@ -332,16 +332,16 @@ public class BufferWriteProcessor extends Processor {
     }
     lastFlushTime = System.nanoTime();
     // check value count
+    // waiting for the end of last flush operation.
+    try {
+      flushFuture.get();
+    } catch (InterruptedException | ExecutionException e) {
+      LOGGER.error("Encounter an interrupt error when waitting for the flushing, "
+              + "the bufferwrite processor is {}.",
+          getProcessorName(), e);
+      Thread.currentThread().interrupt();
+    }
     if (valueCount > 0) {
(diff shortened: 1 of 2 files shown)


iotdb/src/main/java/org/apache/iotdb/db/engine/bufferwrite/BufferWriteProcessor.java
```diff
@@ -332,16 +332,16 @@ private boolean flushTask(String displayMessage, long version) {
     }
     lastFlushTime = System.nanoTime();
     // check value count
+    // waiting for the end of last flush operation.
+    try {
+      flushFuture.get();
+    } catch (InterruptedException | ExecutionException e) {
```
liukun4515: What should we do? when the exception is `ExecutionException`

iotdb/src/main/java/org/apache/iotdb/db/engine/overflow/io/OverflowProcessor.java
```diff
@@ -488,16 +488,15 @@ private boolean flushTask(String displayMessage) {
           (thisFLushTime - lastFlushTime) / 1000);
     }
     lastFlushTime = System.currentTimeMillis();
-    // value count
+    try {
+      flushFuture.get();
+    } catch (InterruptedException | ExecutionException e) {
```
liukun4515: Same question in bufferwrite processor.

iotdb/src/main/java/org/apache/iotdb/db/engine/bufferwrite/BufferWriteProcessor.java
jixuan1989: I think we need to throw it.

PR #149, opened 2019-04-13
Add disabled mem control & improve memory efficiency

jt2594838:
Add DisabledMemController so the overhead of MemControl is testable.

### DIFF
--- a/iotdb/src/main/java/org/apache/iotdb/db/engine/bufferwrite/BufferWriteProcessor.java
+++ b/iotdb/src/main/java/org/apache/iotdb/db/engine/bufferwrite/BufferWriteProcessor.java
@@ -84,6 +84,9 @@ public class BufferWriteProcessor extends Processor {
   private WriteLogNode logNode;
   private VersionController versionController;
 
+  private boolean isClosed = false;
+  private boolean isFlush = false;
+
   /**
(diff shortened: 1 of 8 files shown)


liukun4515:
maybe the error is caused by memory leak?

jt2594838:
reuse BufferwriteProcessor and OverflowProcessor to improve memory efficiency and avoid memory leakage.

iotdb/src/main/java/org/apache/iotdb/db/engine/bufferwrite/BufferWriteProcessor.java
```diff
@@ -454,38 +485,6 @@ public long getFileSize() {
...
-   */
-  public void rollToNewFile() {
-    // TODO : [MemControl] implement this
-  }
-
-  /**
-   * Check if this TsFile has too big metadata or file. If true, close current file and open a new
-   * one.
-   */
-  private boolean checkSize() {
```
jixuan1989: why you delete this function... 
I think this strategy is useful, and we need to consider where to call this method...

jt2594838: This is done in another place.

PR #157, opened 2019-04-19
Refactor mem controller

jt2594838:
refactor mem controller to monitor metadata produced by insertion.


iotdb/src/main/java/org/apache/iotdb/db/engine/memcontrol/RecordMemController.java
```diff
@@ -154,6 +139,7 @@ public void releaseUsage(Object user, long freeSize) {
           .error("{} requests to free {} bytes while it only registered {} bytes", user,
               freeSize, usage);
       totalMemUsed.addAndGet(-usageLong);
+      usage.set(0);
```
jixuan1989: don't you remove it from memMap?

jt2594838: Because it does not indicate that the user will not use more memory when its usage drop to 0.

jt2594838:
Reaching a usage of 0 does not mean it can escape from being monitored.

jixuan1989:
> Reaching a usage of 0 does not mean it can escape from being monitored.

why? if it reaches 0, just remove it. When it acquires memory again, add it again...

jixuan1989:
pls solve the conflict.

PR #163, opened 2019-04-20
[IOTDB-58] Replace list by array and refactor TsFileResource

jixuan1989:
In this PR, changes are:
BufferWriteProcessor:
- remove `fileName` (because `insertFilePath` is enough)

TsFileResource:
- remove `relativePath`
- using `autoRead` for automatically initialize `startTimeMap` and `endTimeMap`

IMemTable, AbstractMemTable:
- add returned value of `delete()`

TsRecord:
- change `dataPoints` from `List` to Array

And in many other classes, if  `Array` is enough, I replace `List<>` by `Array`.

For insertion, add a new method write(long time, Object value) to replace write(long time, String value), because the latter has to call a switch-case and something like Integer.valueOf(value).

To code reviewers:
TsFileResource is a little hard to read.

PR #168, opened 2019-04-25
Refactor query resource count

fanhualta:
In the current implementation, the counting management control of reading resources in FileNodeProcessor is implemented by lock, which will have a negative impact on the distributed reading process, so I change to use AtomicInteger for counting management of reading resources instead of using lock.

### DIFF
--- a/iotdb/src/main/java/org/apache/iotdb/db/engine/filenode/FileNodeManager.java
+++ b/iotdb/src/main/java/org/apache/iotdb/db/engine/filenode/FileNodeManager.java
@@ -643,7 +643,7 @@ public class FileNodeManager implements IStatistic, IService {
     try {
       LOGGER.debug("Get the FileNodeProcessor: filenode is {}, begin query.",
           fileNodeProcessor.getProcessorName());
-      return fileNodeProcessor.addMultiPassLock();
+      return fileNodeProcessor.addMultiPassCount();
     } finally {
       fileNodeProcessor.writeUnlock();
     }
@@ -694,7 +694,11 @@ public class FileNodeManager implements IStatistic, IService {
     try {
       LOGGER.debug("Get the FileNodeProcessor: {} end query.",
           fileNodeProcessor.getProcessorName());
-      fileNodeProcessor.removeMultiPassLock(token);
+      fileNodeProcessor.decreaseMultiPassCount(token);
+    } catch (FileNodeProcessorException e) {
+      throw new FileNodeManagerException(String
+          .format("FileNodeProcessor of [%s] meets error when ending query",
+              fileNodeProcessor.getProcessorName()), e);
     } finally {
       fileNodeProcessor.writeUnlock();
     }
--- a/iotdb/src/main/java/org/apache/iotdb/db/engine/filenode/FileNodeManager.java
+++ b/iotdb/src/main/java/org/apache/iotdb/db/engine/filenode/FileNodeManager.java
@@ -694,9 +694,8 @@ public class FileNodeManager implements IStatistic, IService {
           fileNodeProcessor.getProcessorName());
       fileNodeProcessor.decreaseMultiPassCount(token);
     } catch (FileNodeProcessorException e) {
-      throw new FileNodeManagerException(String
-          .format("FileNodeProcessor of [%s] meets error when ending query",
-              fileNodeProcessor.getProcessorName()), e);
+      LOGGER.error("Failed to end query: the deviceId {}, token {}.", deviceId, token, e);
+      throw new FileNodeManagerException(e);
     } finally {
       fileNodeProcessor.writeUnlock();
     }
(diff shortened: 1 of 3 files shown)


fanhualta:
> seems ok. But could you review PR 163 first and merge this PR until PR #163 is merged :)

That's ok.

iotdb/src/main/java/org/apache/iotdb/db/engine/filenode/FileNodeManager.java
```diff
@@ -694,7 +692,11 @@ public void endQuery(String deviceId, int token) throws FileNodeManagerException
     try {
       LOGGER.debug("Get the FileNodeProcessor: {} end query.",
           fileNodeProcessor.getProcessorName());
-      fileNodeProcessor.removeMultiPassLock(token);
+      fileNodeProcessor.decreaseMultiPassCount(token);
+    } catch (FileNodeProcessorException e) {
+      throw new FileNodeManagerException(String
+          .format("FileNodeProcessor of [%s] meets error when ending query",
```
liukun4515: May be the the format of log should be consistent with other.
@jixuan1989 We should  give a log guides for our developer.

iotdb/src/main/java/org/apache/iotdb/db/engine/filenode/FileNodeProcessor.java
```diff
@@ -1323,11 +1348,11 @@ private void switchWaitingToWorking()
...
 
-    if (oldMultiPassLock != null) {
-      LOGGER.info("The old Multiple Pass Token set is {}, the old Multiple Pass Lock is {}",
+    if (oldMultiPassCount != null) {
+      LOGGER.info("The old Multiple Pass Token set is {}, the old Multiple Pass Count is {}",
           oldMultiPassTokenSet,
-          oldMultiPassLock);
-      oldMultiPassLock.writeLock().lock();
+          oldMultiPassCount);
+      oldMultiPassLock.lock();
```
liukun4515: original code `oldMultiPassLock.writeLock().lock();` is used to check that all old queries are over.
If all queries are over, we can delete invalid tsfile and overflowfile.
New your changes can't make sure that corner case.

jixuan1989: Does `oldMultiPassLock.lock();` not wait for getting the lock?

fanhualta: Well, unfortunately, it's a bug. I change to use CountDownLatch instead of AtmicLong for oldMultiPassCount because oldMultiPassCount will never increase query token num and it just countdown and use CountDownLatch.await() to wait until all old queries are over

iotdb/src/main/java/org/apache/iotdb/db/engine/filenode/FileNodeManager.java
fanhualta: fixed

fanhualta:
> Kun Liu's comment reminds me something:
> 
> 1. addMultiPassLock() may block if the writeLock works, but addMultiPassCount does not.
> 2. In switchWaitingToWorking(),  why you still use a lock? Don't you want to make it lock-free?

The first question I have answered before, please see that in the previous conversation. 
For the second question, indeed, I find that `lock()` of `oldMultiPassLock` is unnecessary because the whole process is locked in the next step (`writeLock()`) and `unlock()` of `oldMultiPassLock` is just after unlock of the whole processor. So I remove it.

fanhualta:
> Old files can't be deleted until no queries need these files.

That's indeed a bug, it has been fixed already.

jixuan1989:
> > Kun Liu's comment reminds me something:
> > 
> > 1. addMultiPassLock() may block if the writeLock works, but addMultiPassCount does not.
> > 2. In switchWaitingToWorking(),  why you still use a lock? Don't you want to make it lock-free?
> 
> The first question I have answered before, please see that in the previous conversation.
> For the second question, indeed, I find that `lock()` of `oldMultiPassLock` is unnecessary because the whole process is locked in the next step (`writeLock()`) and `unlock()` of `oldMultiPassLock` is just after unlock of the whole processor. So I remove it.

if newMultiPassLock.writeLock.tryLock() success, and then someone calls addMultiPassLock(), the behavior is "block", but addMultiPassCount does not...

fanhualta:
> > > Kun Liu's comment reminds me something:
> > > 
> > > 1. addMultiPassLock() may block if the writeLock works, but addMultiPassCount does not.
> > > 2. In switchWaitingToWorking(),  why you still use a lock? Don't you want to make it lock-free?
> > 
> > 
> > The first question I have answered before, please see that in the previous conversation.
> > For the second question, indeed, I find that `lock()` of `oldMultiPassLock` is unnecessary because the whole process is locked in the next step (`writeLock()`) and `unlock()` of `oldMultiPassLock` is just after unlock of the whole processor. So I remove it.
> 
> if newMultiPassLock.writeLock.tryLock() success, and then someone calls addMultiPassLock(), the behavior is "block", but addMultiPassCount does not...

These methods are executed on the premise of acquiring the current write lock of the processor, so there will be no such situation as you said.

liukun4515:
Please fix the ci.
@fanhualta

PR #174, opened 2019-05-12
fix a bug by removing non-migratable absolute parameterized partition tests

LeiRui:
Absolute parameterized partition tests are non-migratable between different platforms.


jixuan1989:
Hm... you just removed all tests...

PR #180, opened 2019-05-24
[IOTDB-95]keep stack trace when logging or throwing an Exception

jt2594838:
see https://issues.apache.org/jira/browse/IOTDB-95 for details.

### DIFF
--- a/iotdb/src/main/java/org/apache/iotdb/db/qp/QueryProcessor.java
+++ b/iotdb/src/main/java/org/apache/iotdb/db/qp/QueryProcessor.java
@@ -108,7 +108,7 @@ public class QueryProcessor {
     } catch (ParseException e) {
       // e.printStackTrace();
       throw new IllegalASTFormatException(
-          "parsing error,statement: " + sqlStr + " .message:" + e.getMessage());
+          "parsing error,statement: " + sqlStr, e);
     }
     return ParseUtils.findRootNonNullToken(astTree);
   }
--- a/iotdb/src/main/java/org/apache/iotdb/db/service/CloseMergeService.java
+++ b/iotdb/src/main/java/org/apache/iotdb/db/service/CloseMergeService.java
@@ -119,9 +119,7 @@ public class CloseMergeService implements IService {
       startService();
     } catch (Exception e) {
       String errorMessage = String
-          .format("Failed to start %s because of %s", this.getID().getName(),
-              e.getMessage());
-      LOGGER.error(errorMessage);
+          .format("Failed to start %s", this.getID().getName());
       throw new StartupException(errorMessage, e);
     }
   }
--- a/iotdb/src/main/java/org/apache/iotdb/db/service/CloseMergeService.java
+++ b/iotdb/src/main/java/org/apache/iotdb/db/service/CloseMergeService.java
@@ -122,7 +122,7 @@ public class CloseMergeService implements IService {
           .format("Failed to start %s because of %s", this.getID().getName(),
               e.getMessage());
       LOGGER.error(errorMessage);
-      throw new StartupException(errorMessage);
+      throw new StartupException(errorMessage, e);
     }
   }
 
--- a/iotdb/src/main/java/org/apache/iotdb/db/writelog/recover/ExclusiveLogRecoverPerformer.java
+++ b/iotdb/src/main/java/org/apache/iotdb/db/writelog/recover/ExclusiveLogRecoverPerformer.java
@@ -314,8 +314,8 @@ public class ExclusiveLogRecoverPerformer implements RecoverPerformer {
     try {
       FileNodeManager.getInstance().closeOneFileNode(writeLogNode.getFileNodeName());
     } catch (FileNodeManagerException e) {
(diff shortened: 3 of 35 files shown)


iotdb/src/main/java/org/apache/iotdb/db/writelog/recover/ExclusiveLogRecoverPerformer.java
```diff
@@ -316,9 +314,8 @@ private void replayLog() throws RecoverException {
     try {
       FileNodeManager.getInstance().closeOneFileNode(writeLogNode.getFileNodeName());
     } catch (FileNodeManagerException e) {
-      logger.error("Log node {} cannot perform flush after replaying logs! Because {}",
-          writeLogNode.getIdentifier(), e.getMessage());
-      throw new RecoverException(e);
+      throw new RecoverException(String.format("Log node %s cannot perform flush after replaying logs! Because %s",
+          writeLogNode.getIdentifier(), e.getMessage()), e);
```
LeiRui: I am not sure if the e.getMessage() is redundant here.

iotdb/src/main/java/org/apache/iotdb/db/writelog/replay/ConcreteLogReplayer.java
```diff
@@ -57,7 +57,7 @@ public void replay(PhysicalPlan plan, boolean isOverflow) throws ProcessorExcept
       }
     } catch (Exception e) {
       throw new ProcessorException(
-          String.format("Cannot replay log %s, because %s", plan.toString(), e.getMessage()));
+          String.format("Cannot replay log %s, because %s", plan.toString(), e.getMessage()), e);
```
LeiRui: I am not sure if the e.getMessage() is redundant here.

iotdb/src/main/java/org/apache/iotdb/db/writelog/recover/ExclusiveLogRecoverPerformer.java
jixuan1989: > I am not sure if the e.getMessage() is redundant here.
This is the javadoc of Exception(String message, Throwable cause):

> Constructs a new exception with the specified detail message and cause.
> Note that the detail message associated with cause is not automatically incorporated in this exception's detail message.

So, I think Tian is correct.

iotdb/src/main/java/org/apache/iotdb/db/qp/QueryProcessor.java
```diff
@@ -108,7 +108,7 @@ private AstNode parseSQLToAST(String sqlStr) throws IllegalASTFormatException {
     } catch (ParseException e) {
       // e.printStackTrace();
       throw new IllegalASTFormatException(
-          "parsing error,statement: " + sqlStr + " .message:" + e.getMessage());
+          "parsing error,statement: " + sqlStr, e);
```
jixuan1989: why not concat e.getMessage() here...

iotdb/src/main/java/org/apache/iotdb/db/service/CloseMergeService.java
```diff
@@ -122,7 +122,7 @@ public void start() throws StartupException {
           .format("Failed to start %s because of %s", this.getID().getName(),
               e.getMessage());
       LOGGER.error(errorMessage);
-      throw new StartupException(errorMessage);
+      throw new StartupException(errorMessage, e);
```
jixuan1989: Why here throw an exception while an error log is printed.

iotdb/src/main/java/org/apache/iotdb/db/writelog/recover/ExclusiveLogRecoverPerformer.java
jixuan1989: >I am not sure if the e.getMessage() is redundant here.

This is the javadoc of Exception(String message, Throwable cause):

>Constructs a new exception with the specified detail message and cause.
>Note that the detail message associated with cause is not automatically incorporated in this exception's detail message.

So, leaving the concat of the detailed message with cause.getMessage() to the exception handler is more neat, I think.

jt2594838: If the programmer only wants to know the cause of the exception, this may help him find out without looking down to the lower exceptions.
While this is not necessary, I will remove them for consistency.

iotdb/src/main/java/org/apache/iotdb/db/qp/QueryProcessor.java
jt2594838: Other similar usages are removed.

iotdb/src/main/java/org/apache/iotdb/db/service/CloseMergeService.java
jt2594838: Removed.

iotdb/src/main/java/org/apache/iotdb/db/writelog/replay/ConcreteLogReplayer.java
jt2594838: Removed.

PR #183, opened 2019-05-24
[IOTDB-94]IoTDB failed to start client since the required jars are no…

jack870131:
Corresponding JIRA issue link: https://issues.apache.org/jira/browse/IOTDB-94


MyXOF:
Tests failed, you should define IOTDB_CLI_HOME in start-client.bat otherwise how can you use this variable?

jack870131:
Sorry, this problem caused because of the instruction being misunderstood. It cannot count as an issue.

PR #184, opened 2019-05-27
fix bug where IOTDB_HOME is misused

LeiRui:
JIRA IOTDB-102


LeiRui:
After a while I'll make some modifications to README and docs.

PR #185, opened 2019-05-31
[IOTDB-105] fix bug of write lock operation in merge() when IoTDB recover

mdf369:
Currenttly, when IoTDB revover, it is possible that exception of write lock operation will be thrown as shown below.
图片
After checking, during the MERGE of recovery, locking and unlocking operations are not in pairs. This is mainly because FileNodeProcessor.merge() contains 2 locking operations and 3 unlocking operations. In normal process, an extra locking operation is processed before merge(). Therefore, to fix this bug, the locking operation is moved into merge().


qiaojialin:
this problem is fixed in feature_asyc_close_tsfile

PR #188, opened 2019-05-31
[IOTDB-109]Fix native restore tsfile writer if file is complete and append=true

jixuan1989:
https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-109

when using nativeRestoreIOWriter to open a sealed TsFile with append=true mode, the parameter does not work now..

The correct behavior is:

if append == true, then the sealed TsFile's file metadata should be truncated and the existing metadata should be loaded into the memory.

### DIFF
--- a/tsfile/example/src/main/java/org/apache/iotdb/tsfile/TsFileSequenceRead.java
+++ b/tsfile/example/src/main/java/org/apache/iotdb/tsfile/TsFileSequenceRead.java
@@ -22,6 +22,8 @@ import java.io.File;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.List;
+import java.util.NavigableMap;
+import java.util.TreeMap;
 import java.util.stream.Collectors;
 import org.apache.iotdb.tsfile.common.conf.TSFileDescriptor;
 import org.apache.iotdb.tsfile.encoding.decoder.Decoder;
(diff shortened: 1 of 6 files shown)


jixuan1989:
Hm... @jt2594838  test failed..

jixuan1989:
The behavior of the class is:

1. if append == false, the writer will be closed after it is created. 
   a. if the tsfile is broken, the behavior of the class is fixing the broken tsfile and make it sealed.
   b. if the tsfile is completed, the class does nothing.
2. if append == true, the writer will truncate some data:
   a. if the tsfile is broken, then the broken data will be truncated; (And, this PR changes the strategy: the class will try its best to save correct data)
   b. if the tsfile is completed, then the tail of the file (FileMetadata) will be truncated.

Therefore, if append == true, you must call close() in the end of your program; If append == false, you can call close() or not.

jixuan1989:
Well, as the class is renamed as RestorableTsFileIOWriter and does not support append at the end of a complete TsFile anymore, I think we can provide a new function to support appending data at the end of a tsfile:

```java
  RestorableTsFileIOWriter createRestorableTsFileIOWriter(File file) throws IOException {
    long position = file.length();

    try (TsFileSequenceReader reader = new TsFileSequenceReader(file.getAbsolutePath(), false)) {
      // this tsfile is complete
      if (reader.isComplete()) {
        TsFileMetaData metaData = reader.readFileMetadata();
        for (TsDeviceMetadataIndex deviceMetadata : metaData.getDeviceMap().values()) {
          if (position > deviceMetadata.getOffset()){
            position = deviceMetadata.getOffset();
          }
        }
      }
    }

    if (position != file.length()) {
      // if the file is complete, we will remove all file metadatas
      try (FileChannel channel = FileChannel.open(Paths.get(file.getAbsolutePath()), StandardOpenOption.WRITE)){
        channel.truncate(position - 1);//remove the last marker.
      }
    }
    return new RestorableTsFileIOWriter(file);
  }
```

PR #189, opened 2019-06-03
Organize properties

fanhualta:
In this pr, I made the following two changes：
1. Organize properties.
2. Improve the implementation of client sync timing

### DIFF
--- a/iotdb/iotdb/conf/iotdb-engine.properties
+++ b/iotdb/iotdb/conf/iotdb-engine.properties
@@ -17,15 +17,22 @@
 # under the License.
 #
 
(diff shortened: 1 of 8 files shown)


richard-yuan16:
Some settings have units, "this unit is in second", some are not. We should spell them out in comments. Besides, I personally prefer this kind of naming, sync_period_in_seconds.

fanhualta:
> Some settings have units, "this unit is in second", some are not. We should spell them out in comments. Besides, I personally prefer this kind of naming, sync_period_in_seconds.

Accept this advice.

PR #190, opened 2019-06-04
Update mult dir avoid full disk

mdf369:
Currently, Directories doesn't avoid using disks that are full, which is a bug. Therefore, this PR intend to ignore full disks when choosing folder for next tsfile; meanwhile, if disks of all directories are all full, exception will be thrown to warn the users.


iotdb/src/main/java/org/apache/iotdb/db/conf/directories/strategy/MinFolderOccupiedSpaceFirstStrategy.java
```diff
@@ -22,40 +22,41 @@
...
-        min = current;
-      } else if (min == current) {
-        candidates.add(i);
+  private int getMinOccupiedSpaceFolder() throws DiskSpaceInsufficientException {
+    int minIndex = -1;
+    long minSpace = Integer.MAX_VALUE;
+
+    for (int i = 0; i < folders.size(); i++) {
+      String folder = folders.get(i);
+      if (hasSpace(folder)) {
```
jixuan1989: if (!hasSpace(folder)) ?

iotdb/src/main/java/org/apache/iotdb/db/conf/directories/strategy/SequenceStrategy.java
```diff
@@ -19,30 +19,46 @@
...
 
     return index;
   }
 
-  private void updateIndex() {
-    currentIndex++;
-    if (currentIndex >= folders.size()) {
-      currentIndex = 0;
+  private int tryGetNextIndex(int start) throws DiskSpaceInsufficientException {
+    int index = start;
```
jixuan1989: int index = ( start +1 ) % folders.size(); ?

iotdb/src/main/java/org/apache/iotdb/db/conf/directories/strategy/MinFolderOccupiedSpaceFirstStrategy.java
mdf369: fixed

iotdb/src/main/java/org/apache/iotdb/db/conf/directories/strategy/SequenceStrategy.java
mdf369: fixed

PR #195, opened 2019-06-14
[ IOTDB-116]Performance statistics module

little-emotion:
In order to better grasp the performance of database in practical application, we add a module to count the time-consuming of each operation. This module can statistic the avg time-consuming of each operation and the proportion of each operation fall into a time range.A output example is in below. This module uses asynchronous operation to reduce the impact on the normal speed of the database.The queue size is controlled by a param in iotdb-engine.properties, which is 'performance_stat_memory_in_kb'. The  output is in  log_measure.log file. 
Pasted Graphic

**configuration parameter**
location：conf/iotdb-engine.properties
item：
// Is stat performance of sub-module enable
enable_performance_stat=false
// The interval of display statistic result in ms.
performance_stat_display_interval=60000
// The memory used for performance_stat in kb.
performance_stat_memory_in_kb=20


little-emotion:
After change ConcurrentLinkedQueue to ConcurrentCircularArray, the performance of branch time_statistic has improved.
result
When  enable_performance_stat=false, almost no impact on performance.
When  enable_performance_stat=true, about 1.2258% performance degradation.

jixuan1989:
1.2% degradation is acceptable. (So, sometimes classes in JDK are not the best.)

Can you add switch for individual measurement...

And be carefor of the Sonar report...

jixuan1989:
Hi, can you solve the conflict ? @little-emotion

PR #197, opened 2019-06-20
Add UT for DirectoriesStrategy

mdf369:
Add unit test for all strategies of choosing directories for new file.


iotdb/src/main/java/org/apache/iotdb/db/conf/directories/strategy/SequenceStrategy.java
```diff
@@ -53,7 +49,7 @@ public int nextFolderIndex() throws DiskSpaceInsufficientException {
 
   private int tryGetNextIndex(int start) throws DiskSpaceInsufficientException {
     int index = (start + 1) % folders.size();
-    while (!hasSpace(folders.get(index))) {
+    while (!FileUtils.hasSpace(folders.get(index))) {
```
jixuan1989: suppose there is no space any more, then start == -1, and then,.... the process will never finish the while-loop...

iotdb/src/main/java/org/apache/iotdb/db/conf/directories/strategy/MaxDiskUsableSpaceFirstStrategy.java
```diff
@@ -35,7 +36,12 @@ public int getMaxSpaceFolder() throws DiskSpaceInsufficientException {
...
 
     for (int i = 0; i < folders.size(); i++) {
-      long space = getUsableSpace(folders.get(i));
+      String folder = folders.get(i);
+      if (!FileUtils.hasSpace(folder)) {
+        continue;
+      }
+
+      long space = FileUtils.getUsableSpace(folder);
       if (space > maxSpace) {
```
jixuan1989: should be "if( space >= maxSpace)... "
Suppose all folder are empty, if so, you will return -1...

iotdb/src/main/java/org/apache/iotdb/db/conf/directories/strategy/SequenceStrategy.java
jixuan1989: Ok... seems that you guarantee that there must be available space by super.init(). If so, please add comments in the init().

mdf369: added.

iotdb/src/main/java/org/apache/iotdb/db/conf/directories/strategy/MaxDiskUsableSpaceFirstStrategy.java
mdf369: if all folders are empty, maxIndex will not be init as line 41 skip to next loop, so exception will be thrown.

PR #216, opened 2019-06-30
reject future writes if there is no disk spaces any more

jixuan1989:
replace flush task runnable  with a callable function; and reject future writes if there is no disk spaces any more...

(I have not deal with how to avoid WAL recovery when restart the system after the disk space is full)

### DIFF
--- a/iotdb/src/main/java/org/apache/iotdb/db/engine/filenodeV2/FileNodeManagerV2.java
+++ b/iotdb/src/main/java/org/apache/iotdb/db/engine/filenodeV2/FileNodeManagerV2.java
@@ -56,6 +56,11 @@ public class FileNodeManagerV2 implements IService {
       .getLogger(org.apache.iotdb.db.engine.filenodeV2.FileNodeManagerV2.class);
   private static final IoTDBConfig config = IoTDBDescriptor.getInstance().getConfig();
 
+  /*
+   * whether reject all writes (insert, update, delete)
+   */
+  private boolean rejectWrite = false;
+
   /**
    * a folder (system/info/ by default) that persist FileNodeProcessorStore classes. Ends with
    * File.separator Each FileNodeManager will have a subfolder.
@@ -78,6 +83,14 @@ public class FileNodeManagerV2 implements IService {
    * This set is used to store overflowed filenode name.<br> The overflowed filenode will be merge.
    */
   private volatile FileNodeManagerStatus fileNodeManagerStatus = FileNodeManagerStatus.NONE;
+
+  public boolean isRejectWrite() {
+    return rejectWrite;
+  }
+
+  public void setRejectWrite(boolean rejectWrite) {
+    this.rejectWrite = rejectWrite;
+  }
 
   private enum FileNodeManagerStatus {
     NONE, MERGE, CLOSE
@@ -168,7 +181,9 @@ public class FileNodeManagerV2 implements IService {
    * @return an int value represents the insert type, 0: failed; 1: overflow; 2: bufferwrite
    */
   public boolean insert(InsertPlan insertPlan) throws FileNodeManagerException {
-
+    if (rejectWrite) {
+      return false;
+    }
     FileNodeProcessorV2 fileNodeProcessor;
     try {
       fileNodeProcessor = getProcessor(insertPlan.getDeviceId());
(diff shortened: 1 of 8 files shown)


iotdb/src/main/java/org/apache/iotdb/db/engine/filenodeV2/FileNodeManagerV2.java
```diff
@@ -200,22 +215,30 @@ private void writeLog(TSRecord tsRecord, boolean isMonitor, WriteLogNode logNode
...
   }
 
   /**
    * delete data.
    */
-  public void delete(String deviceId, String measurementId, long timestamp)
+  public boolean delete(String deviceId, String measurementId, long timestamp)
       throws FileNodeManagerException {
+    if (rejectWrite) {
+      return false;
```
xingtanzjr: should we block `delete` if disk space is not enough ?

iotdb/src/main/java/org/apache/iotdb/db/engine/filenodeV2/UnsealedTsFileProcessorV2.java
```diff
@@ -314,48 +313,55 @@ private void releaseFlushedMemTableCallback(IMemTable memTable) {
...
 //      if (elapse > 1000) {
 //        LOGGER.info("release a memtable cost: {}", elapse);
 //      }
-      if (IoTDBDescriptor.getInstance().getConfig().isEnableWal()) {
-        getLogNode().notifyEndFlush();
+        if (IoTDBDescriptor.getInstance().getConfig().isEnableWal()) {
+          getLogNode().notifyEndFlush();
+        }
       }
       LOGGER.info("flush a memtable has finished");
```
xingtanzjr: should move this line `LOGGER.info...` inside above if block.

iotdb/src/main/java/org/apache/iotdb/db/engine/filenodeV2/UnsealedTsFileProcessorV2.java
```diff
@@ -314,48 +313,55 @@ private void releaseFlushedMemTableCallback(IMemTable memTable) {
...
-      endFile();
-
-      // for sync close
+      if (flushSuccessed) {
+        //if !flushSuccessed, then the file may be broken, we do not seal the file.
+        endFile();
+      }
+      // for notifying syncClose()
       synchronized (flushingMemTables) {
         flushingMemTables.notify();
```
xingtanzjr: This line means that `Async Close` is done. But if `flushSuccess` is `false`, we should not tell the thread which is waiting that close is done. Maybe we need to reconsider the operation here.

iotdb/src/main/java/org/apache/iotdb/db/engine/memtable/MemTableFlushTaskV2.java
```diff
@@ -94,14 +96,18 @@ public void flushMemTable() {
         "Storage group {} memtable {}, flushing into disk: data sort time cost {} ms.",
         storageGroup, memTable.getVersion(), sortTime);
 
+    Boolean success = false;
     try {
-      ioFlushTaskFuture.get();
+      success = ioFlushTaskFuture.get();
```
xingtanzjr: Rename this parameter to `flushSuceeded`

iotdb/src/main/java/org/apache/iotdb/db/engine/filenodeV2/FlushManager.java
```diff
@@ -34,14 +35,15 @@
...
 
     @Override
-    public void run() {
+    public Boolean call() {
       UnsealedTsFileProcessorV2 unsealedTsFileProcessor = unsealedTsFileProcessorQueue.poll();
       long startTime = System.currentTimeMillis();
+      boolean flushSuccessed = false;
       try {
-        unsealedTsFileProcessor.flushOneMemTable();
+        flushSuccessed = unsealedTsFileProcessor.flushOneMemTable();
```
xingtanzjr: Typo. It should be `flushSecceeded`

iotdb/src/main/java/org/apache/iotdb/db/engine/filenodeV2/FileNodeManagerV2.java
jixuan1989: > should we block `delete` if disk space is not enough ?

because delete is in the lazy mode... It does not delete data on disk immediately.

iotdb/src/main/java/org/apache/iotdb/db/engine/filenodeV2/UnsealedTsFileProcessorV2.java
jixuan1989: If we do not notify the waiting object, maybe they will wait forever...

iotdb/src/main/java/org/apache/iotdb/db/engine/filenodeV2/UnsealedTsFileProcessorV2.java
jixuan1989: fixed

iotdb/src/main/java/org/apache/iotdb/db/engine/memtable/MemTableFlushTaskV2.java
jixuan1989: renamed as `flushSuccessfully`

iotdb/src/main/java/org/apache/iotdb/db/engine/filenodeV2/FlushManager.java
jixuan1989: done

PR #221, opened 2019-07-08
rename a class and modify some comments

LeiRui:


jixuan1989:
FileSeries* means reading data from a data File?

PR #224, opened 2019-07-08
Tsfile write example revised

mikip65:
Enabled TsFile creation based on measurements in JSON file and provided corresponding example and tutorial


tsfile/example/readme.md
```diff
@@ -31,27 +31,37 @@ The example is to show how to write and read a TsFile File.
...
-```
-  The class is to show how to write Tsfile directly,it use the second interface: 
-            public void addMeasurement(MeasurementSchema measurementSchema) throws WriteProcessException
-```
+  This class is to show how to write a Tsfile. It provided two ways:
+  
+   The first one is using a JSON file for measurement(s). The JSON file is an array of JSON objects(schema). The object
+   must include the *measurement_id*, *datatype*, *encoding*, and *compressor*. 
+   
+   An example JSON file is provided in the comments in
```
jixuan1989: json string rather than file

PR #229, opened 2019-07-10
[IOTDB-128] Fix links of nav content in documents

samperson1997:
Fix links of nav content in documents by turning nav content into "Overview" part to fix the bug proposed in [issue IOTDB-128](https://issues.apache.org/jira/browse/IOTDB-128).

### DIFF
--- a/docs/Community-Powered By.md
+++ b/docs/Community-Powered By.md
@@ -21,9 +21,11 @@
 
 <!-- TOC -->
 
-- [Powered By](#powered-by)
-    - [Project and Product names using "IoTDB"](#project-and-product-names-using-iotdb)
-    - [Companies and Organizations](#companies-and-organizations)
+## Outline
+
+- Powered By
+    - Project and Product names using "IoTDB"
+    - Companies and Organizations
 
 <!-- /TOC -->
 ## Powered By
--- a/docs/Development.md
+++ b/docs/Development.md
@@ -21,23 +21,25 @@
 
 <!-- TOC -->
 
-- [Have Questions](#have-questions)
-    - [Mailing Lists](#mailing-lists)
-    - [JIRA issues](#jira-issues)
-- [How to contribute](#how-to-contribute)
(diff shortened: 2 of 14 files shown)


Genius-pig:
Sorry for troubling you. To be honest, I don't think you have fixed this bug. Because the original purpose was when you click some tables of contents(TOC),  jump to the appropriate contents.  Rather than just adding an "outline" in the head of contents.

samperson1997:
> Sorry for troubling you. To be honest, I don't think you have fixed this bug. Because the original purpose was when you click some tables of contents(TOC), jump to the appropriate contents. Rather than just adding an "outline" in the head of contents.

Hi,

We have considered the original purpose, but we thought the nav-content in the head of the documents is not as convenient as on side part (as User Guide documents), so I changed it into "outline" now. Nav content on side part will probably be added in the future. Maybe I should change my description of this PR?

Genius-pig:
> > Sorry for troubling you. To be honest, I don't think you have fixed this bug. Because the original purpose was when you click some tables of contents(TOC), jump to the appropriate contents. Rather than just adding an "outline" in the head of contents.
> 
> Hi,
> 
> We have considered the original purpose, but we thought the nav-content in the head of the documents is not as convenient as on side part (as User Guide documents), so I changed it into "outline" now. Nav content on side part will probably be added in the future. Maybe I should change my description of this PR?

No. As you said, the side part is more convenient.

jixuan1989:
Hi, If no more comment (@Genius-pig ), I will merge this PR first, and wait for the new website in the future.
@samperson1997 , if you have a plan to move the outline from the top to the left side, pls open an issue so that it can remind us.

samperson1997:
> Hi, If no more comment (@Genius-pig ), I will merge this PR first, and wait for the new website in the future.
> @samperson1997 , if you have a plan to move the outline from the top to the left side, pls open an issue so that it can remind us.

Sure, here is the issue: [IOTDB-132](https://issues.apache.org/jira/browse/IOTDB-132).

PR #238, opened 2019-07-12
[IOTDB-133] Fix other mistaken links in documents

Genius-pig:
There are still some other mistaken links.


jixuan1989:
Thanks ~!

PR #244, opened 2019-07-15
[IOTDB-134]Add default parameters for client

lvpengze:
default parameters："-h 127.0.0.1 -p 6667 -u root -pw root"


jixuan1989:
It  is a bigggggggg improvement!!!

PR #245, opened 2019-07-15
refactor readonly mode check

qiaojialin:


jixuan1989:
is there a real case for testing that?

jixuan1989:
There are conflicting files..

qiaojialin:
> is there a real case for testing that?

Qingxin will test that in a real scenario

PR #250, opened 2019-07-16
update RELEASE_NOTES for 0.8.0

qiaojialin:
update release_notes

### DIFF
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -4,29 +4,100 @@ Apache IoTDB (incubating) 0.8.0
 
 This is the first official release of Apache IoTDB after joining the Incubator.
 
+
 New Features
 ------------
 
+IOTDB-1     Add Aggregation query
+IOTDB-4     Asynchronously force sync WAL periodically
 IOTDB-5     Support data deletion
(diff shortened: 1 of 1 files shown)


RELEASE_NOTES
```diff
@@ -4,36 +4,148 @@ Apache IoTDB (incubating) 0.8.0
...
+Enable users define the location of their thrift compiler
+Add example module
+Add a log appender: put info, warn, error log into one file and disable log_info by default
+Recover when resource file does not exist while tsfile is complete
+
 
 Incompatible changes
 --------------------
 
+Data folder structures
```
jixuan1989: You'd better to claim clearly be incompatible with WHO?

JulianFeinauer: Wat is WHO?

JulianFeinauer:
I cannot comment on all things in detail but the update looks excellent and extensive... but I think @jixuan1989 should merge it, when his comment is adressed.

RELEASE_NOTES
qiaojialin: He means with which version... I have updated this description

jixuan1989: +1.

PR #254, opened 2019-07-17
[IOTDB-140]Fixed the bug in calculating chunk datasize and write the size for string variable.

mikip65:
Now it can support string with non-ascii characters.


jixuan1989:
I will merge it.

How about also merge into rel/0.8?

PR #264, opened 2019-07-18
Update jdbc doc 

1160300922:
Update README.md in jdbc and doc in docs/documentation/UserGuideV0.7.0/6-JDBC API


jixuan1989:
um... this PR seems contains not only the JDBC doc supplement but also time unit midification...

PR #266, opened 2019-07-18
Timestamp precision

1160300922:
IoTDB only supports millisecond timestamp-precision. Therefore, I add microsecond and nanosecond to help users have more choice in timestamp precision while inserting or querying data.


qiaojialin:
> I would still suggest a check on the change of time-precision on the system-up to enhance the robustness of IoTDB.

Yes, working on it.

PR #267, opened 2019-07-18
[IOTDB-146] Windows should use different mvn command, like using quoting '-Dmaven test.skip=true'

Genius-pig:
[issue](https://issues.apache.org/jira/browse/IOTDB-146) and detailed [solution](https://stackoverflow.com/questions/6347985/cannot-run-maven-using-mvn-d-argument-within-microsoft-powershell-but-works).


jt2594838:
👍

PR #290, opened 2019-07-23
Fix 'latest' in documents link url to '0.8.0'

samperson1997:
Fix `latest` in documents link url to `0.8.0`, according to [PR #280](https://github.com/apache/incubator-iotdb/pull/280) and [this PR in website](https://github.com/apache/incubator-iotdb-website/pull/26).


jixuan1989:
Hi, I think the PR is ok.... So, if we release v0.9, do we have to replace all `latest` to `v0.9`?

samperson1997:
> Hi, I think the PR is ok.... So, if we release v0.9, do we have to replace all `latest` to `v0.9`?

Yes.. It seems a little weird, but actually I have referred the [User Guide of VUE](https://cn.vuejs.org/v2/guide/), and they also change all the `version` in links when updating new versions.

PR #294, opened 2019-07-23
[IOTDB-152]return data type in TSExecuteStatementResp

little-emotion:
When executing the query statement, after sending executeStatement request to server and return the execResp, the client  send several TSFetchMetadata  requests to server for fetching data types of the query columns. So we can return dataTypes in  executeStatement request to avoid the following fetchmetaData requests.
Thank Rui Lei for discovering this problem.


service-rpc/src/main/thrift/rpc.thrift
```diff
@@ -74,6 +74,9 @@ struct TSExecuteStatementResp {
 	3: optional list<string> columns
 	4: optional string operationType
 	5: optional bool ignoreTimeStamp
+
+  // Data type list of columns in select statement of SQL
+  6: optional list<string> dataTypeList
```
qiaojialin: format

little-emotion: In compiler, these attributes are aligned.

PR #302, opened 2019-07-24
[IOTDB-153]further limit fetchSize to speed up LIMIT&OFFSET query

LeiRui:


LeiRui:
> The only question is that what is `ada` short for?
> (`adaFetchSize`)

adapt.

PR #303, opened 2019-07-25
transform old form to new form and use a parameter to control

SilverNarcissus:
/**
    * transfer old form to new form
    *
    * @param spark your SparkSession
    * @param df    dataFrame need to be tansfer
    *              +---------+-------------+-------------+-------------+-------------+-------------+-------------+
    *              |timestamp|root.ln.d1.m1|root.ln.d1.m2|root.ln.d1.m3|root.ln.d2.m1|root.ln.d2.m2|root.ln.d2.m3|
    *              +---------+-------------+-------------+-------------+-------------+-------------+-------------+
    *              |        1|           11|           12|         null|           21|           22|           23|
    *              +---------+-------------+-------------+-------------+-------------+-------------+-------------+
    * @return tansferred data frame
    *         +---------+-----------+---+---+----+
    *         |timestamp|device_name| m1| m2|  m3|
    *         +---------+-----------+---+---+----+
    *         |        1| root.ln.d2| 21| 22|  23|
    *         |        1| root.ln.d1| 11| 12|null|
    *         +---------+-----------+---+---+----+
    */

and use a parameter to control table form
usage : spark.read.tsfile("path", true, spark)


samperson1997:
I tried the new change:
屏幕快照 2019-07-25 14 36 49

To acquire the new form of df, use `spark.read.tsfile(“test.tsfile”, true, spark)`; To acquire the old form, use `spark.read.tsfile(“test.tsfile”, false, spark)` or just `spark.read.tsfile(“test.tsfile”)` as before.

@qiaojialin Do you think this is acceptable for users to change between old and new forms?

spark-tsfile/src/main/scala/org/apache/iotdb/tsfile/Transformer.scala
```diff
@@ -0,0 +1,174 @@
...
+    *         +---------+-------------+-------------+-------------+-------------+-------------+-------------+
+    *         |timestamp|root.ln.d1.m1|root.ln.d1.m2|root.ln.d1.m3|root.ln.d2.m1|root.ln.d2.m2|root.ln.d2.m3|
+    *         +---------+-------------+-------------+-------------+-------------+-------------+-------------+
+    *         |        1|           11|           12|         null|           21|           22|           23|
+    *         +---------+-------------+-------------+-------------+-------------+-------------+-------------+
+    *
+    */
+  def toOldForm(spark: SparkSession,
+                df: DataFrame): DataFrame = {
+    df.registerTempTable("iotdb_new_form")
```
qiaojialin:
```suggestion
    df.registerTempTable("tsfile_new_form")
```

spark-tsfile/src/main/scala/org/apache/iotdb/tsfile/Transformer.scala
```diff
@@ -0,0 +1,174 @@
...
+    // get all device_name
+    val device_names = spark.sql("select distinct device_name from iotdb_new_form").collect()
```
qiaojialin:
```suggestion
    val device_names = spark.sql("select distinct device_name from tsfile_new_form").collect()
```

spark-tsfile/src/main/scala/org/apache/iotdb/tsfile/Transformer.scala
```diff
@@ -0,0 +1,174 @@
...
+    val table_df = spark.sql("select * from iotdb_new_form")
```
qiaojialin:
```suggestion
    val table_df = spark.sql("select * from tsfile_new_form")
```

spark-tsfile/src/main/scala/org/apache/iotdb/tsfile/Transformer.scala
```diff
@@ -0,0 +1,174 @@
...
+    // +---------+-------------+-------------+-------------+-------------+-------------+-------------+
+
+    device_names.foreach(device_name => {
+      var query = "select " + QueryConstant.RESERVED_TIME
+
+      measurement_names.foreach(measurement_name => {
+        query = query + ", " + measurement_name + " as `" + device_name(0) + "." + measurement_name + "`"
+      })
+
+      query = query + " from iotdb_new_form where device_name = \"" + device_name(0) + "\""
```
qiaojialin:
```suggestion
      query = query + " from tsfile_new_form where device_name = \"" + device_name(0) + "\""
```

spark-tsfile/src/main/scala/org/apache/iotdb/tsfile/package.scala
```diff
@@ -19,15 +19,23 @@
...
 package object tsfile {
-
   /**
     * add a method 'tsfile' to DataFrameReader to read tsfile
     */
   implicit class TsFileDataFrameReader(reader: DataFrameReader) {
-    def tsfile: String => DataFrame = reader.format("org.apache.iotdb.tsfile").load
+    def tsfile(path: String,
+               isNewForm: Boolean = false,
+               spark: SparkSession = null): DataFrame = {
```
qiaojialin: could the parameter spark be omitted and gotten from inside?

spark-tsfile/src/main/scala/org/apache/iotdb/tsfile/package.scala
```diff
@@ -19,15 +19,23 @@
...
-import org.apache.spark.sql.{DataFrame, DataFrameReader, DataFrameWriter}
+import org.apache.spark.sql.{DataFrame, DataFrameReader, DataFrameWriter, SparkSession}
 
 package object tsfile {
-
   /**
     * add a method 'tsfile' to DataFrameReader to read tsfile
     */
   implicit class TsFileDataFrameReader(reader: DataFrameReader) {
-    def tsfile: String => DataFrame = reader.format("org.apache.iotdb.tsfile").load
```
qiaojialin: Is this interface removed: spark.read.tsfile("test.tsfile") ?

You can add an interface, but do not remove the old interface

SilverNarcissus: I use default parameters for upward compatibility, that is, you can still use spark.read.tsfile("test.tsfile")  to get an old form table

spark-tsfile/src/main/scala/org/apache/iotdb/tsfile/package.scala
SilverNarcissus: spark can't be omitted because if we omit spark, we will get an error when we compile our jar, which is "value spark not found". As for got from inside, I can't get from reader and I'm not sure if I get from context all configuration is same as the client wish

SilverNarcissus:
As for the performance consideration. Given that tsfile is read by path, if we want get new table form, we need to split one row of data and build the iterator. I'm not sure whether this step can be parallelized. In current implementation, I just use union function, which means all operations can be in one stage without shuffle.

qiaojialin:
> As for the performance consideration. Given that tsfile is read by path, if we want get new table form, we need to split one row of data and build the iterator. I'm not sure whether this step can be parallelized. In current implementation, I just use union function, which means all operations can be in one stage without shuffle.

I mean when users want to use the new table form, there will be no wide row which contains all time series at all. Therefore, we do not need to split a wide row into several narrow rows.

The following is an example: suppose the file schema has "s1, s2, s3"

In DefaultSource.inferSchema(), we return the "timestamp, deviceId, s1, s2, s3" to spark to generate the new table form.

In DefaultSource.buildReader(), the required schema is "timestamp, deviceId, s1, s2, s3", suppose the PartitionedFile has one chunk group (device1, s1, s2, s3). Then, we could build a tsfile reader to query three time series : device1.s1, device1.s2, device1.s3 and directly construct the result to an InternalRow and return to Spark.

This step can be parallelized. Each executor in spark will read a partition of one tsfile. These executors could work independently.

spark-tsfile/src/test/scala/org/apache/iotdb/tsfile/TSFileSuit.scala
```diff
@@ -214,4 +214,28 @@ class TSFileSuit extends FunSuite with BeforeAndAfterAll {
...
+  }
+
+  test("testTransform3") {
+    val df = spark.read.tsfile(tsfile1, true, spark)
+    df.createOrReplaceTempView("tsfile_table")
+    val newDf = spark.sql("select * from tsfile_table " +
+      "where (`sensor_1`>0 or `sensor_2` < 22) and time < 4")
+    Assert.assertEquals(5, newDf.count())
+  }
+
```
samperson1997: Hi,

The tests need to be changed at the same time : )

LeiRui:
I suggest you spend some time updating the `docs\Documentation\UserGuide\9-Tools-spark.md` as well.

spark-tsfile/src/test/scala/org/apache/iotdb/tsfile/TSFileSuit.scala
```diff
@@ -214,4 +214,28 @@ class TSFileSuit extends FunSuite with BeforeAndAfterAll {
...
 
+  test("testTransform1") {
+    val df = spark.read.tsfile(tsfile1)
+    df.createOrReplaceTempView("tsfile_table")
+    val newDf = spark.sql("select * from tsfile_table " +
+      "where (`device_1.sensor_1`>0 or `device_1.sensor_2` < 22) and time < 4")
+    Assert.assertEquals(3, newDf.count())
+  }
+
+  test("testTransform2") {
```
LeiRui: test("testTransform2")  and test("testTransform3") can't pass.
Please check this problem.

spark-tsfile/src/main/scala/org/apache/iotdb/tsfile/DefaultSource.scala
```diff
@@ -99,50 +109,126 @@ private[tsfile] class DefaultSource extends FileFormat with DataSourceRegister {
...
+            var index = 0
+            requiredSchema.foreach((field: StructField) => {
+              if (field.name == QueryConstant.RESERVED_TIME) {
+                rowBuffer(index) = curRecord.getTimestamp
+              }
+              else if(field.name == NewConverter.DEVICE_NAME){
+                rowBuffer(index) = device_name
+              }
+              else {
+                val pos = paths.indexOf(new org.apache.iotdb.tsfile.read.common.Path(device_name + "." + field.name))
```
qiaojialin:
```suggestion
                val pos = paths.indexOf(new org.apache.iotdb.tsfile.read.common.Path(device_name, field.name))
```

spark-tsfile/src/main/scala/org/apache/iotdb/tsfile/NewConverter.scala
```diff
@@ -0,0 +1,654 @@
...
+
+import scala.collection.JavaConversions._
+import scala.collection.mutable
+import scala.collection.mutable.ListBuffer
+
+/**
+  * This object contains methods that are used to convert schema and data between SparkSQL and TSFile.
+  *
+  */
+object NewConverter {
```
qiaojialin: This converter has many duplicated methods with the old converter, such as toSqlValue, toSqlField. You can combine NewConverter with Converter.

spark-tsfile/src/main/scala/org/apache/iotdb/tsfile/Transformer.scala
```diff
@@ -0,0 +1,174 @@
...
+  * under the License.
+  */
+package org.apache.iotdb.tsfile
+
+import org.apache.iotdb.tsfile.common.constant.QueryConstant
+import org.apache.spark.sql.{SparkSession, _}
+import org.apache.spark.sql.types._
+
+
+object Transformer {
```
qiaojialin: Since you can read dataframe in two forms, this transformer is needless.

spark-tsfile/src/main/scala/org/apache/iotdb/tsfile/package.scala
```diff
@@ -19,15 +19,24 @@
...
   /**
     * add a method 'tsfile' to DataFrameReader to read tsfile
+    * @param reader dataframeReader
     */
   implicit class TsFileDataFrameReader(reader: DataFrameReader) {
-    def tsfile: String => DataFrame = reader.format("org.apache.iotdb.tsfile").load
+    def tsfile(path: String,
+               isNewForm: Boolean = false): DataFrame = {
+      if (isNewForm) {
+        reader.option(DefaultSource.path, path).option(DefaultSource.isNewForm, "new_form").format("org.apache.iotdb.tsfile").load
```
qiaojialin: the parameter could be a boolean

tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/TsFileMetaData.java
```diff
@@ -19,303 +19,310 @@
...
+import java.util.List;
 import java.util.Map;
-import org.apache.iotdb.tsfile.file.metadata.enums.TSDataType;
-import org.apache.iotdb.tsfile.utils.ReadWriteIOUtils;
-import org.apache.iotdb.tsfile.write.schema.MeasurementSchema;
 
 /**
  * TSFileMetaData collects all metadata info and saves in its data structure.
  */
 public class TsFileMetaData {
```
qiaojialin: what did you change?

tsfile/src/main/java/org/apache/iotdb/tsfile/read/TsFileSequenceReader.java
```diff
@@ -55,634 +38,706 @@
...
 
+import java.io.File;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.file.Paths;
+import java.util.*;
+
+import static org.apache.iotdb.tsfile.write.writer.TsFileIOWriter.magicStringBytes;
+
 public class TsFileSequenceReader implements AutoCloseable {
```
qiaojialin: Did you format this class?

tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/TsFileMetaData.java
SilverNarcissus: I added a method to this file which called getDeviceNameInRange and then use format in IDE. Maybe it remove useless import

spark-tsfile/src/main/scala/org/apache/iotdb/tsfile/package.scala
SilverNarcissus: in scala, Boolean is not a class but a primitive type

tsfile/src/main/java/org/apache/iotdb/tsfile/read/TsFileSequenceReader.java
SilverNarcissus: yes, I use IDE default format

spark-tsfile/src/main/scala/org/apache/iotdb/tsfile/Transformer.scala
SilverNarcissus: I think Transformer has 2 function:
1. sometimes we get wide table and then need to transform to narrow table, transformer can be use in this condition.
2. transformer support narrow table to wide table, which can be write into tsfile

LeiRui:
I have to say that the codes you borrowed from are not well written and is difficult to read.

spark-tsfile/src/main/scala/org/apache/iotdb/tsfile/Transformer.scala
qiaojialin: That reminds me... We need to support writing new form to tsfile, which is easier than old form.

spark-tsfile/src/test/scala/org/apache/iotdb/tsfile/TSFileSuit.scala
```diff
@@ -214,4 +223,28 @@ class TSFileSuit extends FunSuite with BeforeAndAfterAll {
...
+    val df = spark.read.tsfile(tsfile1, true)
+    df.createOrReplaceTempView("tsfile_table")
+    val newDf = spark.sql("select * from tsfile_table " +
+      "where `device_name` = 'device_1' and (`sensor_1`>0 or `sensor_2` < 22) and time < 4")
+    Assert.assertEquals(3, newDf.count())
+  }
+
+  test("testTransform3") {
```
LeiRui: Check the `partitionedFile` in the `DefaultSource.buildReader` and you will find that the partitionedFile's start and length just correspond to the whole tsfile. That actually means the partitioned case is not covered in these tests. 

So, to valid the full function of your code, it would be nice to add more tests to **cover the cases when the tsfile is really partitioned into more than one partitionedFile**.

I currently have no feasible solutions for that. I tried to add the following configurations but the partitionedFile turned out to still be the whole tsfile:
```
    val minSplit: Int = 100000
    val maxSplit: Int = 100000
    spark.sparkContext.hadoopConfiguration.setInt("mapred.min.split.size", minSplit)
    spark.sparkContext.hadoopConfiguration.setInt("mapred.max.split.size", maxSplit)
```

qiaojialin: You can try dfs.block.size, and blockSize should be a multiple of io.bytes.per.checksum (512).

SilverNarcissus: After scanning the source code of spark, I found that the parameter to control this is 'spark.sql.files.maxPartitionBytes'. I will add some test case.

LeiRui: excellent

SilverNarcissus: Thanks @qiaojialin and @LeiRui for reviewing my code :)

PR #304, opened 2019-07-25
[IOTDB-154] Remove unuseful package.sh and fix a url in pom.xml.

Genius-pig:
[issue](https://issues.apache.org/jira/browse/IOTDB-154)


jixuan1989:
Thanks for your carefully modification. 
Indeed the package.sh is a legacy when we are not familiar with using Maven to do these things.. It can be removed I think. How about you @qiaojialin

Genius-pig:
@jixuan1989 The problem is that it can't be used, like `./target/lib/*.jar` doesn't exist.

qiaojialin:
> Thanks for your carefully modification.
> Indeed the package.sh is a legacy when we are not familiar with using Maven to do these things.. It can be removed I think. How about you @qiaojialin

+1 for removing that

PR #308, opened 2019-07-26
Update dockerfile to only compile server and client

samperson1997:
Update dockerfile to only compile server and client during building.


samperson1997:
@samperson1997

PR #310, opened 2019-07-26
Final steps towards rel.

JulianFeinauer:
Removed "build" module and moved this to parent pom.
Renamed parent pom without incubating.
Added incubating artefact names in assembly plugin.
Added SHA512 for all artefacts.

Big props to @chrisdutz for his support!


pom.xml
```diff
@@ -584,5 +583,78 @@
...
+                                <goals>
+                                    <goal>single</goal>
+                                </goals>
+                                <!-- heads up: combine.self in the following is highlighted
+                                    as an error in Eclipse's xml editor view.
+                                    Just ignore that.
+                                    See  https://issues.apache.org/jira/browse/MNG-5454  sigh.
+                                 -->
+                                <configuration combine.self="append">
+                                    <finalName>apache-iotdb-${project.version}-incubating</finalName>
```
jixuan1989: Oh, now I get what Justin means... just add "incubating" here is fine..

JulianFeinauer: yeah.. therefore I changed the name of the module back to "iotdb-parent" which is more standard : )

jixuan1989:
Hi, I tried this branch but after running `mvn package`, I do not get the .zip file...

JulianFeinauer:
@jixuan1989 you have to use `mvn package -Papache-release` to enable these steps.

PR #312, opened 2019-07-26
use latest binary file to build the docker image

jixuan1989:
Because the dockerfile depends on the compiled binary file of IoTDB, I modify the dockerfile to use the new binary file.


jixuan1989:
> Hey @jixuan1989 I would suggest to move this file to "src/main/docker" as this fits more to the maven convention. Otherwise it lloks good.

Hi, I move the file to docker/src/main/.

> Why is the Entrypoing commented out?

It is because I think this docker image can be as a base image. We can use it to build a server image or a client image. Or, we can use it to run either a server instance or a client instance by:

* `docker run -p 6667:6667 -v mydata:/iotdb/data -v mylogs:/iotdb/logs -d iotdb:base /iotdb/bin/start-server.sh`
* `docker run -it iotdb:base /cli/bin/start-client.sh -h <C_IP> -p 6667 -u root -pw root` 

However, if we have chance to publish the image to DockerHub, we need open the Entrypoing, I think.

jixuan1989:
> Looks reasonable. Is it rebased? We should also pick it to master.

yes, after the PR is rebased and merge to rel/0.8, we can cherry-pick to master

jixuan1989:
has cherry-picked to the master branch

PR #319, opened 2019-07-27
add jmx interface for getting the active and pending tasks in FlushManager

jixuan1989:
Using jmx interface, we can know how many tasks are pending to be flushed. It is helpful for tuning parameters..

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/flush/pool/AbstractPoolManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/flush/pool/AbstractPoolManager.java
@@ -85,8 +85,4 @@ public abstract class AbstractPoolManager {
 
   public abstract String getName();
 
-  public int getNumberOfWorkingTasks() {
-    return ((ThreadPoolExecutor)pool).getActiveCount();
-  }
-
 }
--- a/server/src/main/java/org/apache/iotdb/db/engine/flush/pool/AbstractPoolManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/flush/pool/AbstractPoolManager.java
@@ -84,4 +84,12 @@ public abstract class AbstractPoolManager {
   public abstract void stop();
 
   public abstract String getName();
+
+  public int getNumberOfWorkingTasks() {
+    return ((ThreadPoolExecutor)pool).getActiveCount();
+  }
+
+  public int getNumberOfPendingTasks() {
+    return ((ThreadPoolExecutor)pool).getQueue().size();
+  }
 }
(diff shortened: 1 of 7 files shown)


jixuan1989:
it is the same with run `show flush task info` on cli...

server/src/main/java/org/apache/iotdb/db/service/ServiceType.java
```diff
@@ -33,7 +33,9 @@
   SYNC_SERVICE("SYNC ServerService", ""),
   PERFORMANCE_STATISTIC_SERVICE("PERFORMANCE_STATISTIC_SERVICE","PERFORMANCE_STATISTIC_SERVICE"),
   TVLIST_ALLOCATOR_SERVICE("TVList Allocator", ""),
-  FLUSH_SERVICE("Flush ServerService", "");
+
+  FLUSH_SERVICE("Flush ServerService",
+      generateJmxName("org.apache.iotdb.db.engine.pool", "Flush Manager"));
```
jt2594838: How about renaming the JmxNames of other services that are registered in JMX too?

jt2594838:
👍 We may add more statistics like the number of submitted tasks and average/maximum task runtime here, but this is good for now.

server/src/main/java/org/apache/iotdb/db/engine/flush/pool/AbstractPoolManager.java
```diff
@@ -84,4 +84,12 @@ public int getCorePoolSize() {
...
 
   public abstract String getName();
+
+  public int getNumberOfWorkingTasks() {
+    return ((ThreadPoolExecutor)pool).getActiveCount();
+  }
+
+  public int getNumberOfPendingTasks() {
+    return ((ThreadPoolExecutor)pool).getQueue().size();
+  }
```
fanhualta: Actually , the method has the same function with `getWaitingTasksNumber` in line 68 of this class.

server/src/main/java/org/apache/iotdb/db/engine/flush/pool/AbstractPoolManager.java
```diff
@@ -84,4 +84,12 @@ public int getCorePoolSize() {
   public abstract void stop();
 
   public abstract String getName();
+
+  public int getNumberOfWorkingTasks() {
+    return ((ThreadPoolExecutor)pool).getActiveCount();
+  }
```
fanhualta: Actually, this method has the same function with `getActiveCnt` in line 64 of this class.

server/src/main/java/org/apache/iotdb/db/engine/flush/pool/AbstractPoolManager.java
jixuan1989: fixed

server/src/main/java/org/apache/iotdb/db/engine/flush/pool/AbstractPoolManager.java
jixuan1989: fixed

PR #322, opened 2019-07-29
modify print function in AbstractClient

LeiRui:
Original:

After changing:


LeiRui:
> Excellent. Though I wonder why there is such a huge difference.

In AbstractClient `print`, `printf` and `println` are redefined:
```
  protected static void printf(String format, Object ... args) {
    SCREEN_PRINTER.printf(format, args);
  }

 protected static void print(String msg) {
    SCREEN_PRINTER.println(msg);
  }

  protected static void println() {
    SCREEN_PRINTER.println();
  }

  protected static void println(String msg) {
    SCREEN_PRINTER.println(msg);
  }

  protected static void println(Object obj) {
    SCREEN_PRINTER.println(obj);
  }
```
And `print` will change the line.

PR #323, opened 2019-07-29
implement rpc compression

Ring-k:
The rpc compression function is implemented.


jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBConnection.java
```diff
@@ -89,7 +90,10 @@ public IoTDBConnection(String url, Properties info) throws SQLException, TTransp
     supportedProtocols.add(TSProtocolVersion.TSFILE_SERVICE_PROTOCOL_V1);
 
     openTransport();
-    client = new TSIService.Client(new TBinaryProtocol(transport));
+    if(Config.rpcThriftCompressionEnable)
+      client = new TSIService.Client(new TCompactProtocol(transport));
+    else
+      client = new TSIService.Client(new TBinaryProtocol(transport));
```
jt2594838: Please use brackets even for a single statement in if or for.

Ring-k: Alright, thanks for your suggestion.

PR #326, opened 2019-07-29
Change map key in TsDigest from String to enum data type

LeiRui:
Changing from
```
private Map<String, ByteBuffer> statistics
```
to 
```
private Map<StatisticType, ByteBuffer> statistics
```
can decrease IO number. Because when performing `TsDigest.deserializeFrom`, `ReadWriteIOUtils.readString` does two basic reads while `ReadWriteIOUtils.readShort` only does once:
```
  public static String readString(InputStream inputStream) throws IOException {
    int strLength = readInt(inputStream);
    byte[] bytes = new byte[strLength];
    int readLen = inputStream.read(bytes, 0, strLength);
    if (readLen != strLength) {
      throw new IOException(String.format("Intend to read %d bytes but %d are actually returned",
          strLength, readLen));
    }
    return new String(bytes, 0, strLength);
  }

  public static short readShort(InputStream inputStream) throws IOException {
    byte[] bytes = new byte[SHORT_LEN];
    int readLen = inputStream.read(bytes);
    if (readLen != SHORT_LEN) {
      throw new IOException(String.format("Intend to read %d bytes but %d are actually returned",
          SHORT_LEN, readLen));
    }
    return BytesUtils.bytesToShort(bytes);
  }
```


LeiRui:
Can someone who has the write access rerun this travis?

jixuan1989:
Also suggest to do not use Map in the class (while keeping the compatibility of TsFile serialization).

jt2594838:
Good, but using map is still too memory-costly. I would suggest using array and the ordinal of the enum as index.

tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/TsDigest.java
```diff
@@ -233,13 +234,58 @@ public boolean equals(Object o) {
...
   }
+
+  public enum StatisticType {
+    max_value, min_value, first, sum, last;
+
+    public static StatisticType deserialize(short i) {
+      if (i >= 5 || i < 0) {
+        throw new IllegalArgumentException("Invalid input: " + i);
+      }
+      switch (i) {
```
jt2594838: Optional: simplify it using`return StatisticType.values()[i];`.

tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/TsDigest.java
```diff
@@ -233,13 +234,58 @@ public boolean equals(Object o) {
...
+          return last;
+      }
+    }
+
+    public static int getSerializedSize() {
+      return Short.BYTES;
+    }
+
+    public short serialize() {
+      switch (this) {
```
jt2594838: Optional: simplify it with "ordinal()"

LeiRui:
I'll try your suggestions to use arrary instead of map today.

LeiRui:
This can be reviewed.

LeiRui:
This can be reviewed.

LeiRui:
Can someone rerun this travis?

PR #327, opened 2019-07-29
Fix markdown link problem

Genius-pig:
Should use %20 to represent space, otherwise it's invalid.


jixuan1989:
has cherry-picked to master.

PR #329, opened 2019-08-02
- Fix the main issues with the assembly of the binary distribution.

chrisdutz:


JulianFeinauer:
Thank you @chrisdutz !

PR #334, opened 2019-08-09
fix start-walchecker scripts for leting user define the wal folder

jixuan1989:
As many users do not set IoTDB_HOME... the script will fail. Therefore, letting user assign the folder is better.

The better way is if $IOTDB_HOME is empty, then let user assign that..


Genius-pig:
Checks fall because of network problem which accounts for that `maven-wrapper.jar` couldn't be downloaded. In fact, to create or update all necessary Maven Wrapper files just execute the following command:
```
mvn -N io.takari:maven:wrapper
```
So I think some files about Maven Wrapper are unnecessary. But we can also imitate spark to use a
 a self-contained Maven.

jt2594838:
+1

jixuan1989:
has cherry-picked to master

PR #335, opened 2019-08-10
fix readme

qiaojialin:
fix some error path in readme, and add a note about generated-soures of thrift and altlr


jixuan1989:
has cherry-picked to master

PR #338, opened 2019-08-13
[IOTDB-158] add metrics service

xijunmu:
add a monitoring web page to show SQL execution status after service startup.Monitor the memory and CPU usage of the current host


Genius-pig:
I think you should open an issue or talk in the mailing list.

PR #339, opened 2019-08-14
[IOTDB-158] add metrics service

xijunmu:
add a monitoring web page to show SQL execution status after service startup.Monitor the memory and CPU usage of the current host


server/src/main/java/org/apache/iotdb/db/metrics/server/MetricsSystem.java
```diff
@@ -0,0 +1,83 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iotdb.db.metrics.server;
+
+import java.util.ArrayList;
+
+import org.apache.iotdb.db.metrics.sink.ConsoleSink;
```
xingtanzjr: Remove unused `import`

server/src/main/java/org/apache/iotdb/db/metrics/server/MetricsSystem.java
```diff
@@ -0,0 +1,83 @@
...
+import org.apache.iotdb.db.metrics.sink.MetricsServletSink;
+import org.apache.iotdb.db.metrics.sink.Sink;
+import org.apache.iotdb.db.metrics.source.JvmSource;
```
xingtanzjr: Remove unused `import`

server/src/main/java/org/apache/iotdb/db/metrics/server/MetricsSystem.java
```diff
@@ -0,0 +1,83 @@
...
+import org.apache.iotdb.db.metrics.source.MetricsSource;
+import org.apache.iotdb.db.metrics.source.Source;
+import org.eclipse.jetty.servlet.ServletContextHandler;
+
+import com.codahale.metrics.MetricRegistry;
+
+public class MetricsSystem {
+	
+	private ArrayList<Sink> sinks = new ArrayList<Sink>();
```
xingtanzjr:
```
private ArrayList<Sink> sinks = new ArrayList<>();
```
Explicit type argument could be removed

server/src/main/java/org/apache/iotdb/db/metrics/server/MetricsSystem.java
```diff
@@ -0,0 +1,83 @@
...
+	private ArrayList<Source> sources = new ArrayList<Source>();
```
xingtanzjr: Same as above

xingtanzjr:
**Please check your code style first**

For example:
* remove unused import
* Typos
* unused functions
* commented codes
...

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -481,6 +484,9 @@ private boolean executeStatementInBatch(String statement, StringBuilder batchErr
 
   @Override
   public TSExecuteStatementResp executeStatement(TSExecuteStatementReq req) {
+	long starttime = System.currentTimeMillis();
```
xingtanzjr: should be `startTime`. Please check other parameters' name, either.

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -523,7 +529,17 @@ public TSExecuteStatementResp executeStatement(TSExecuteStatementReq req) {
       PhysicalPlan physicalPlan;
       physicalPlan = processor.parseSQLToPhysicalPlan(statement, zoneIds.get());
       if (physicalPlan.isQuery()) {
-        return executeQueryStatement(statement, physicalPlan);
+      	resp = executeQueryStatement(statement, physicalPlan);
+      	long endtime = System.currentTimeMillis();
+      	sqlArgument = new SqlArgument(resp,physicalPlan,statement,starttime,endtime);
```
xingtanzjr: Duplicated.
Already pass `statement` as a parameter in constructor but the next line still invoke `setStatement(statement)`

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -523,7 +529,17 @@ public TSExecuteStatementResp executeStatement(TSExecuteStatementReq req) {
...
+      	sqlArgument.setStatement(statement);
+      	sqlArgumentsList.add(sqlArgument);
+      	if(sqlArgumentsList.size()>200) {
+  			for (int i = 0; i < 50; i++) {
+  				sqlArgumentsList.remove(sqlArgumentsList.size()-1);
```
xingtanzjr: It seems that these lines are used to limit total size of saved sqls. But this line seems removes the latest 50 sqls rather than the oldest ones.

server/src/main/java/org/apache/iotdb/db/metrics/ui/MetricsPage.java
```diff
@@ -0,0 +1,172 @@
...
+import java.util.List;
+
+import org.apache.iotdb.db.conf.IoTDBConstant;
+import org.apache.iotdb.db.metrics.server.SqlArgument;
+import org.apache.iotdb.service.rpc.thrift.TSExecuteStatementResp;
+
+import com.codahale.metrics.MetricRegistry;
+
+public class MetricsPage {
+
```
xingtanzjr: Do we have some better way to make `html` as standalone templete rather than hard code them in java file ?

server/src/main/java/org/apache/iotdb/db/metrics/ui/MetricsPage.java
```diff
@@ -0,0 +1,172 @@
...
+			status = "FAILED";
+		}
+
+		app.append(
+			 "<tr>"
+		  +    "<td>"+resp.getOperationType()+"</td>"
+		  +    "<td>"+sdf.format(new Date(queryResult.getStarttime()))+"</td>"
+		  +    "<td>"+sdf.format(new Date(queryResult.getEndtime()))+"</td>"
+		  +    "<td>"+(int)(queryResult.getEndtime()-queryResult.getStarttime())+" ms</td>"
+		  +    "<td style=\"font-size:13px\">"+queryResult.getStatement()+"</td>"
```
xingtanzjr: `hard code style` is not recommended

server/src/main/java/org/apache/iotdb/db/metrics/ui/MetricsPage.java
Genius-pig: IoTDB doesn't have special web UI, it uses Grafana. When it has web UI, the monitoring web page should be added.

xijunmu: Ok, I'm going to try to turn the HTML code into an HTML file

xijunmu: Well, there's some  information on the page about memory ,cpus and JVM

server/src/main/java/org/apache/iotdb/db/metrics/server/ServerArgument.java
```diff
@@ -0,0 +1,338 @@
...
+	 */
+	private double getCpuRateForLinux() {
+		try {
+			long[] c0 = readLinuxCpu();
+			Thread.sleep(CPUTIME);
+			long[] c1 = readLinuxCpu();
+			if (c0 != null && c1 != null) {
+				long idleCpuTime = c1[0] - c0[0];
+				long totalCpuTime = c1[1] - c0[1];
+				return Double.valueOf(100 * (1 - idleCpuTime / totalCpuTime)).doubleValue();
```
samperson1997: Hi,

I have checked the failed tests in [CompressionRatioTest.java, IoTDBConfigDynamicAdapterTest.java](https://travis-ci.org/apache/incubator-iotdb/jobs/577722148), and it is caused by " / by zero" when initializing. Maybe a judgement needs to be added here?

xijunmu: i fix some bug.now test is pass

server/src/main/java/org/apache/iotdb/db/metrics/server/ServerArgument.java
```diff
@@ -0,0 +1,347 @@
...
+				buffer.close();
+			} catch (Exception e2) {
+				e2.printStackTrace();
+			}
+		}
+		return retn;
+	}
+
+	private String substring(String src, int start_idx, int end_idx) {
+		byte[] b = src.getBytes();
```
samperson1997: This method is unnecessary, you can use Java API `src.substring(start_idx, end_idx)` to get the substring. What is different from your method is: `i >= start_idx && i < end_idx`.

server/src/main/java/org/apache/iotdb/db/metrics/server/ServerArgument.java
```diff
@@ -0,0 +1,347 @@
...
+			int kmtidx = line.indexOf("KernelModeTime");
+			int wocidx = line.indexOf("WriteOperationCount");
+			long idletime = 0;
+			long kneltime = 0;
+			long usertime = 0;
+			while ((line = input.readLine()) != null) {
+				if (line.length() < wocidx) {
+					continue;
+				}
+				String caption = substring(line, capidx, cmdidx - 1).trim();
```
samperson1997: I think you can simply use Java API to get substring, and notice the range of `i`. **Other places in this PR** are the same.
```suggestion
				String caption = line.substring(capidx, cmdidx).trim();
```

server/src/main/java/org/apache/iotdb/db/metrics/server/ServerArgument.java
```diff
@@ -0,0 +1,347 @@
...
+			long[] c0 = readLinuxCpu();
+			Thread.sleep(CPUTIME);
+			long[] c1 = readLinuxCpu();
+			if (c0 != null && c1 != null) {
+				long idleCpuTime = c1[0] - c0[0];
+				long totalCpuTime = c1[1] - c0[1];
+				if (totalCpuTime == 0) {
+					return 100.0;
+				}
+				return Double.valueOf(100 * (1 - idleCpuTime / totalCpuTime)).doubleValue();
```
samperson1997: Devision of these two long value `idleCpuTime` and `totalCpuTime` is either 0 or 1. If you want to get double result, you can try to first parse long values to double values.

server/src/main/java/org/apache/iotdb/db/metrics/server/ServerArgument.java
```diff
@@ -0,0 +1,347 @@
...
+			long[] c0 = readWinCpu(Runtime.getRuntime().exec(procCmd));
+			Thread.sleep(CPUTIME);
+			long[] c1 = readWinCpu(Runtime.getRuntime().exec(procCmd));
+			if (c0 != null && c1 != null) {
+				long idletime = c1[0] - c0[0];
+				long busytime = c1[1] - c0[1];
+				if ((busytime + idletime) == 0) {
+					return 100.0;
+				}
+				return Double.valueOf(100 * (busytime) / (busytime + idletime)).doubleValue();
```
samperson1997: Same as above.

server/src/main/resources/iotdb/ui/static/index.html
```diff
@@ -0,0 +1,80 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+<html>
+<head>
+<meta http-equiv="Content-type" content="text/html; charset=utf-8">
+<link rel="stylesheet" href="/static/bootstrap.min.css" type="text/css">
```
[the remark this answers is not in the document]
xijunmu: I think it is useful to refer a local CSS file. Sometimes the service may be deployed in an environment without an external network.And the CSS file is only 124kb

xijunmu:
i fix the web page not fonud exception. now you can access x.x.x.x:8181 get the page

xijunmu:
and unified the style of all the code

samperson1997:
Hi, thanks for your update! I noticed that a test case is failed in Windows platform:
```
[ERROR] Errors: 
[ERROR]   WriteLogNodeManagerTest.testAutoSync:80 � NoClassDefFound Could not initialize...
```
With my macOS environment, I could not reproduce this error... Are you convenient to solve the failure?

samperson1997:
I have tried the Metrics Server on my Ubuntu virtual machine, and it is awesome!
屏幕快照 2019-09-26 20 41 26
However, I only got results of query statements in the table, is it designed on purpose?
Here are my test statements:
```
SET STORAGE GROUP TO root.turbine;
CREATE TIMESERIES root.turbine.d1.s0 WITH DATATYPE=FLOAT, ENCODING=RLE;
insert into root.turbine.d1(timestamp,s0) values(1,24.5);
select s0 from root.turbine.d1;
delete from root.turbine.d1.s0 where time < 2;
insert into root.turbine.d1(timestamp,s0) values(2,25.3);
select s0 from root.turbine.d1;
```
Besides, is it necessary to show the warning message `Can't get the cpu ratio,because this OS is not support` instead of `0% CPU Ratio` for macOS? (as below)
屏幕快照 2019-09-26 20 46 37
Thanks again for your fabulous function and effort! : )

xijunmu:
Initially, I wanted to keep all the SQL execution records.But then I think maybe a lot of insert SQL execution and show this insert sql is not mean.So later I removed  it include create table SQL and authorization SQL, only keep the query SQL

little-emotion:
This is a great work and does not seem to have a significant impact on query performance.

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -504,7 +510,16 @@ public TSExecuteStatementResp executeStatement(TSExecuteStatementReq req) {
       PhysicalPlan physicalPlan;
       physicalPlan = processor.parseSQLToPhysicalPlan(statement, zoneIds.get());
       if (physicalPlan.isQuery()) {
-        return executeQueryStatement(statement, physicalPlan);
+      	resp = executeQueryStatement(statement, physicalPlan);
+      	long endTime = System.currentTimeMillis();
+      	sqlArgument = new SqlArgument(resp,physicalPlan,statement,startTime,endTime);
+      	sqlArgumentsList.add(sqlArgument);
+      	if(sqlArgumentsList.size()>200) {
+  			for (int i = 0; i < 50; i++) {
```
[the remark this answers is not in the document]
xijunmu: already accepted your suggestion

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -80,6 +82,7 @@
 
   private static final Logger logger = LoggerFactory.getLogger(TSServiceImpl.class);
   private static final String INFO_NOT_LOGIN = "{}: Not login.";
+  public static List<SqlArgument> sqlArgumentsList = new ArrayList<>();
```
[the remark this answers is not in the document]
xijunmu: modify the list to vector

jixuan1989:
> Usually where we need something like this we download the resource as part of the build and take the parts we need and then reference that file. By actually adding it to the repo, I guess you have to mention it in the LICENSE or NOTICE files.

Declare it in the LICENSE and NOTICE is a choice. You can find a better way in the following discussion:

https://lists.apache.org/thread.html/29be94b4fa6d201445e736946d1e04f29050349ab5ee0a72717e8d84@%3Cdev.iotdb.apache.org%3E

jixuan1989:
One more thing, if there is a markdown document for introducing the tool, it could be better.  You can put the markdown file under: `docs/Documentation/UserGuide/9-System Tools` (For Chinese version, put it under `docs/Documentation-CHN/UserGuide/9-System Tools`

xijunmu:
I removed the bootstrap.min.css file .Now it is not necessary.Add some CSS styles to myself CSS file

xijunmu:
@samperson1997

xijunmu:
@samperson1997                                                                                                                                                                   
 Can you help me write a user guide document .Thank you very much.

PR #341, opened 2019-08-16
increase configuration parameters of cache

little-emotion:
Adding configuration parameters to metadata caching module：
1、Is enable cache  parameter.
2、Memory occupancy ratio parameters of tsFileMetaData cache and chunkMetaData cache.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/conf/IoTDBDescriptor.java
+++ b/server/src/main/java/org/apache/iotdb/db/conf/IoTDBDescriptor.java
@@ -281,10 +281,17 @@ public class IoTDBDescriptor {
         proportionSum += Integer.parseInt(proportion.trim());
       }
       long maxMemoryAvailable = conf.getAllocateMemoryForRead();
-      conf.setAllocateMemoryForFileMetaDataCache(
-          maxMemoryAvailable * Integer.parseInt(proportions[0].trim()) / proportionSum);
-      conf.setAllocateMemoryForChumkMetaDataCache(
-          maxMemoryAvailable * Integer.parseInt(proportions[1].trim()) / proportionSum);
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/apache/iotdb/db/conf/IoTDBDescriptor.java
```diff
@@ -263,6 +267,26 @@ private void initMemoryAllocate(Properties properties) {
...
+      return;
+    }
+
+    String queryMemoryAllocateProportion = properties
+        .getProperty("filemeta_chunkmeta_free_memory_proportion");
+    if (queryMemoryAllocateProportion != null) {
+      String[] proportions = queryMemoryAllocateProportion.split(":");
+      int proportionSum = 0;
+      for (String proportion : proportions) {
+        proportionSum += Integer.parseInt(proportion.trim());
```
qiaojialin: How about checking minus numbers?

server/src/main/java/org/apache/iotdb/db/engine/cache/TsFileMetadataUtils.java
```diff
@@ -98,4 +98,20 @@ public static TsDeviceMetadata getTsDeviceMetaData(String filePath, Path seriesP
     return pathToChunkMetaDataList;
   }
 
+  public static List<ChunkMetaData> getChunkMetaDataList(String sensor,
```
qiaojialin: Better to use "measurement".

little-emotion: Measurement represents a time series, and what I want to express here is the last layer of the time series path.

PR #347, opened 2019-08-21
[IOTDB-172]fix a bug of TsFileResource updateTime 

LeiRui:
Bug Description: 
When recovering a TsFile's `device -> start time` and `device -> end time` map and the TsFile's resource file happens not to exist, the TsFile itself will be read to update the start and end time map instead. It is implemented by `TsFileRecoverPerformer.recover` as follows:
```
for (TsDeviceMetadataIndex index : deviceMetadataIndexList) {
  TsDeviceMetadata deviceMetadata = reader.readTsDeviceMetaData(index);
  List<ChunkGroupMetaData> chunkGroupMetaDataList = deviceMetadata.getChunkGroupMetaDataList();
  for (ChunkGroupMetaData chunkGroupMetaData : chunkGroupMetaDataList) {
    for (ChunkMetaData chunkMetaData : chunkGroupMetaData.getChunkMetaDataList()) {
      tsFileResource.updateTime(chunkGroupMetaData.getDeviceID(), chunkMetaData.getStartTime());
      tsFileResource.updateTime(chunkGroupMetaData.getDeviceID(), chunkMetaData.getEndTime());
    }
  }
}
```
The `updateTime` function uses the first-coming value as the startTime:
```
  public void updateTime(String deviceId, long time) {
    startTimeMap.putIfAbsent(deviceId, time);
    Long endTime = endTimeMap.get(deviceId);
    if (endTime == null || endTime < time) {
      endTimeMap.put(deviceId, time);
    }
  }
```
So when root.vehicle.d0 has a chunkgroup of chunkMetadata listed as:
s3: startTime=3000, endTime=10000;
s1: startTime=1000, endTime=10000;
s2: startTime=2000, endTime=10000;
By the current code, the updated start time will be the first coming 3000, which is not correct.

---

By the way, I also change `updateTime` in `LogReplayer.replayLogs` to `updateStartTime` and `updateEndTime`. If there is something I haven't noticed, please tell me.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
@@ -221,6 +221,7 @@ public class TsFileResource {
     return processor;
   }
 
+  @Deprecated
   public void updateTime(String deviceId, long time) {
     startTimeMap.putIfAbsent(deviceId, time);
     Long endTime = endTimeMap.get(deviceId);
--- a/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
@@ -220,13 +220,4 @@ public class TsFileResource {
   public TsFileProcessor getUnsealedFileProcessor() {
     return processor;
   }
-
-  @Deprecated
-  public void updateTime(String deviceId, long time) {
-    startTimeMap.putIfAbsent(deviceId, time);
-    Long endTime = endTimeMap.get(deviceId);
-    if (endTime == null || endTime < time) {
-      endTimeMap.put(deviceId, time);
-    }
-  }
 }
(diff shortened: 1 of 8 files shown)


MyXOF:
please add ut to cover this case

LeiRui:
> please add ut to cover this case

added :relaxed:

jixuan1989:
Hi, is there an jira issue attached this PR?
We need to manage the solved version on that.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
```diff
@@ -221,6 +221,7 @@ public TsFileProcessor getUnsealedFileProcessor() {
     return processor;
   }
 
+  @Deprecated
   public void updateTime(String deviceId, long time) {
```
jt2594838: You may just remove this method.

LeiRui:
> Hi, is there an jira issue attached this PR?
> We need to manage the solved version on that.

Added. See IOTDB-172.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
LeiRui: removed

PR #350, opened 2019-08-22
[IOTDB-161] Add ErrorCode of different response errors

samperson1997:
There is already an optional attribute `errorCode` in struct `TS_Status` (but not used before), and it could be used for representing different response errors.

For example, `TIME_SERIES_NOT_EXIST_ERROR` equals 301 by defined. When referring to this error, instead of using `e.getErrorMessage().contains('exists')`, we can just use `e.getErrorCode() == 301`.


server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -114,6 +114,7 @@
   private static final Logger logger = LoggerFactory.getLogger(TSServiceImpl.class);
   private static final String INFO_NOT_LOGIN = "{}: Not login.";
   private static final String ERROR_NOT_LOGIN = "Not login";
+  private static final int SUCCESS_CODE = 200;
```
jt2594838: I would suggest you move this into ErrorCode and rename it as StatusCode if you prefer the HTTP style things.

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -160,8 +161,7 @@ public TSOpenSessionResp openSession(TSOpenSessionReq req) throws TException {
       zoneIds.set(config.getZoneID());
       initForOneSession();
     } else {
-      tsStatus = new TS_Status(TS_StatusCode.ERROR_STATUS);
-      tsStatus.setErrorMessage("login failed. Username or password is wrong.");
+      tsStatus = getErrorStatus(ErrorCode.WRONG_LOGIN_PASSWORD_ERROR, "login failed. Username or password is wrong.");
```
jt2594838: A more thorough solution: make ErrorCode (or StatusCode more precisely) an enum and bind the message with each code.

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -433,16 +429,16 @@ public TSExecuteBatchStatementResp executeBatchStatement(TSExecuteBatchStatement
       }
 
       if (isAllSuccessful) {
-        return getTSBathExecuteStatementResp(TS_StatusCode.SUCCESS_STATUS,
+        return getTSBatchExecuteStatementResp(TS_StatusCode.SUCCESS_STATUS, SUCCESS_CODE,
             "Execute batch statements successfully", result);
       } else {
-        return getTSBathExecuteStatementResp(TS_StatusCode.ERROR_STATUS,
+        return getTSBatchExecuteStatementResp(TS_StatusCode.ERROR_STATUS, ErrorCode.EXECUTE_STATEMENT_ERROR,
```
jt2594838: It feels a bit strange to have StatusCode and ErrorCode at the same time, maybe you can synthesize them.

samperson1997: Hi,

The existing code defined `statusCode` and `errorCode` separately, and `errorCode` is only set when `statusCode` is `ERROR_STATUS`:
```
struct TS_Status {
  1: required TS_StatusCode statusCode
  ...
  // If status is ERROR, then the following fields may be set
  4: optional i32 errorCode // internal error code
  5: optional string errorMessage
}
```
Should I change this struct by combine statusCode and errorCode (and maybe also include errorMessage) ?

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
samperson1997: I didn't bind the message because I thought some error code may have more equivalent messages. For example, `FETCH_METADATA_ERROR` may have messages `Unsupported fetch metadata operation ...` or `Failed to fetch metadata because...`. If the error codes and messages are bound, then maybe the number of error codes will be larger and more detailed.

jt2594838: Yes, I think the more clear the codes, the better, which will help the user faster locate the problem.

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
jt2594838: Yes, those fields seem a bit redundant now.

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBPreparedInsertionStatement.java
```diff
@@ -49,7 +49,7 @@ public boolean execute() throws SQLException {
       req.unsetMeasurements();
       req.unsetTimestamp();
       req.unsetValues();
-      return resp.getStatus().getStatusCode() == TS_StatusCode.SUCCESS_STATUS;
+      return resp.getStatus().getStatusType().getCode() == SUCCESS_CODE;
```
qiaojialin:
```suggestion
      return resp.getStatus().getStatusType() == TSStatusType.SUCCESS_CODE;
```

qiaojialin: Then, you can remove "private static final int SUCCESS_CODE = 200;" 
It's better to use TSStatusType instead of defining an int in each class.

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBStatement.java
```diff
@@ -294,7 +286,7 @@ private boolean executeSQL(String sql) throws TException, SQLException {
     TSExecuteBatchStatementReq execReq = new TSExecuteBatchStatementReq(sessionHandle,
         batchSQLList);
     TSExecuteBatchStatementResp execResp = client.executeBatchStatement(execReq);
-    if (execResp.getStatus().statusCode == TS_StatusCode.SUCCESS_STATUS) {
+    if (execResp.getStatus().getStatusType().getCode() == SUCCESS_CODE) {
```
qiaojialin: Same as above.

jdbc/src/test/java/org/apache/iotdb/jdbc/IoTDBDatabaseMetadataTest.java
```diff
@@ -66,7 +63,8 @@
   @Mock
   private TSFetchMetadataResp fetchMetadataResp;
 
-  private TS_Status Status_SUCCESS = new TS_Status(TS_StatusCode.SUCCESS_STATUS);
+  private TS_StatusType successStatus = new TS_StatusType(200, "");
```
qiaojialin:
```suggestion
  private TS_StatusType successStatus = new TS_StatusType(TSStatusType. SUCCESS_CODE, "");
```

qiaojialin: Try to avoid hard coding

jdbc/src/test/java/org/apache/iotdb/jdbc/IoTDBQueryResultSetTest.java
```diff
@@ -117,7 +105,8 @@
   @Mock
   private TSFetchResultsResp fetchResultsResp;
 
-  private TS_Status Status_SUCCESS = new TS_Status(TS_StatusCode.SUCCESS_STATUS);
+  private TS_StatusType successStatus = new TS_StatusType(200, "");
```
qiaojialin: same as above

PR #352, opened 2019-08-23
[IoTDB-174]Add interfaces for querying device or timeseries number

jack870131:
The corresponding JIRA issue: https://issues.apache.org/jira/browse/IOTDB-174

1. Create an interface for querying the timeseries number under the specified path.

2. Create an interface for querying the node number at the given level in current Metadata Tree (this could be used to query the number of devices).

3. Create an interface for querying the timeseries number of each node at the given level (this could be used to query the number of sensors under each device).


EJTTianYu:
I have reviewed the code, all functions work fine in iotdb 0.9.0-SNAPSHOT.
Approved.

PR #365, opened 2019-08-31
[IOTDB-189] Support compatibility of HDFS File and Java File

samperson1997:
This is the first step to resolve requirement [IOTDB-187](https://issues.apache.org/jira/browse/IOTDB-187): "Enabling to choose storage in local file system or HDFS".

This step contains:
 * Add property `storage_fs` (default `LOCAL`, could be `HDFS` etc) in `iotdb-engine.properties` for users to config.
 * Add `FileFactory` to get files by configuration in `IoTDBConfig`.
 * Add `HdfsFile` which extends `java.io.file` and rewrite methods in HDFS ways (may have some bugs, will be fixed in next steps)

After these steps, we could use `FileFactory.INSTANCE.getFile(...)` to get "file" according to configured file system, which supports compatibility of HDFS file and java file.


tsfile/src/main/java/org/apache/iotdb/tsfile/fileSystem/IoTDBFile.java
```diff
@@ -0,0 +1,183 @@
...
+      files[i] = new IoTDBFile(rawFiles[i].getPath());
+    }
+    return files;
+  }
+
+  public IoTDBFile[] listFiles(FileFilter filter) {
+    if (TSFileConfig.isHdfsStorage) {
+      ArrayList<IoTDBFile> files = new ArrayList<>();
+      try {
+        PathFilter pathFilter = new GlobFilter(filter.toString()); // TODO
```
jt2594838: TODO what?

tsfile/src/main/java/org/apache/iotdb/tsfile/fileSystem/IoTDBFile.java
```diff
@@ -0,0 +1,183 @@
...
+  public IoTDBFile(File parent, String child) {
+    super(parent, child);
+  }
+
+  public IoTDBFile(URI uri) {
+    super(uri);
+  }
+
+  @Override
+  public String getAbsolutePath() {
```
qiaojialin: Consider another way: Use a IoTDBFileFactory to generate different files: java.io.File and HDFSFile(extends java.io.File). Then, we only need to judge the storage type when create and no need to judge in each method.

tsfile/src/main/java/org/apache/iotdb/tsfile/fileSystem/IoTDBFile.java
samperson1997: `PathFilter` / `GlobFilter` in hadoop file system needs _RegExpr_ as param to filter files, however every `FileFilter` in java file system needs to implement interface `FileFilter`. I intend to fix this method to support hadoop file in [next step](https://github.com/apache/incubator-iotdb/pull/362) since this PR just encapsulated "IoTDBFile" to ensure java file system works well...

tsfile/src/main/java/org/apache/iotdb/tsfile/fileSystem/IoTDBFile.java
```diff
@@ -0,0 +1,183 @@
...
+import org.apache.hadoop.fs.*;
+import org.apache.iotdb.tsfile.common.conf.TSFileConfig;
+
+import java.io.File;
+import java.io.FileFilter;
+import java.io.IOException;
+import java.net.URI;
+import java.util.ArrayList;
+
+public class IoTDBFile extends File {
```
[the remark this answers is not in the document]
samperson1997: Thanks for such a good suggestion! I'd like to refactor codes as suggested.

server/src/assembly/resources/conf/tsfile-format.properties
```diff
@@ -47,4 +47,7 @@ value_encoder=PLAIN
 
 # Compression configuration
 # Data compression method, TsFile supports UNCOMPRESSED or SNAPPY. Default value is UNCOMPRESSED which means no compression
-compressor=UNCOMPRESSED
\ No newline at end of file
+compressor=UNCOMPRESSED
+
+# Storage file system. Currently, Tsfile supports to be storied in LOCAL file system or HDFS.
+storage_fs=LOCAL
```
qiaojialin: this parameter should be in the iotdb-engine.properties

tsfile/src/main/java/org/apache/iotdb/tsfile/fileSystem/IoTDBFileFactory.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.io.File;
+import java.net.URI;
+
+public enum IoTDBFileFactory {
+
+  INSTANCE;
+
+  public File getIoTDBFile(String pathname) {
+    if (TSFileConfig.storageFs.equals("HDFS")) {
```
qiaojialin: use an enum

tsfile/src/main/java/org/apache/iotdb/tsfile/fileSystem/IoTDBFileFactory.java
```diff
@@ -0,0 +1,63 @@
...
+      return new HdfsFile(pathname);
+    } else {
+      return new File(pathname);
+    }
+  }
+
+  public File getIoTDBFile(String parent, String child) {
```
qiaojialin: rename to getFile or newFile

PR #371, opened 2019-09-04
fix some small problems discovered when reading code

LeiRui:


LeiRui:
> If these are flaws in these codes you have reviewed, why not add some unit tests to cover these cases?

Thanks. I added a unit test.
But now I realize that these changes in this pr are so minor that they cannot be called "fix", because the original code works correctly with no bug.

PR #375, opened 2019-09-05
[IOTDB-191]Add the delete interface to Session API

jack870131:
Corresponding JIRA issue: https://issues.apache.org/jira/browse/IOTDB-191

Add the delete interface for Session API.

### DIFF
--- a/docs/Documentation-CHN/UserGuide/7-Session API/1-Session API.md
+++ b/docs/Documentation-CHN/UserGuide/7-Session API/1-Session API.md
@@ -75,6 +75,7 @@ public class SessionExample {
 
     insert();
 //    insertRowBatch();
+    delete();
 
     session.close();
   }
@@ -92,6 +93,12 @@ public class SessionExample {
       values.add("3");
       session.insert(deviceId, time, measurements, values);
     }
+  }
+
+  private static void delete() throws IoTDBSessionException {
+    String path = "root.sg1.d1.s1";
+    long deleteTime = 29999;
+    session.delete(path, deleteTime);
   }
 
   private static void insertRowBatch() throws IoTDBSessionException {
--- a/example/session/src/main/java/org/apache/iotdb/SessionExample.java
+++ b/example/session/src/main/java/org/apache/iotdb/SessionExample.java
@@ -44,6 +44,7 @@ public class SessionExample {
 
     insert();
 //    insertRowBatch();
+    delete();
 
     session.close();
   }
@@ -61,6 +62,12 @@ public class SessionExample {
       values.add("3");
       session.insert(deviceId, time, measurements, values);
     }
+  }
+
+  private static void delete() throws IoTDBSessionException {
+    String path = "root.sg1.d1.s1";
+    long deleteTime = 29999;
+    session.delete(path, deleteTime);
   }
 
   private static void insertRowBatch() throws IoTDBSessionException {
--- a/service-rpc/src/main/thrift/rpc.thrift
+++ b/service-rpc/src/main/thrift/rpc.thrift
@@ -246,6 +246,11 @@ struct TSInsertReq {
     4: required i64 timestamp
 }
 
+struct TSDeleteReq {
+    1: required string path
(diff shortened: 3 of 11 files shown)


docs/Documentation-CHN/UserGuide/7-Session API/1-Session API.md
```diff
@@ -94,6 +94,14 @@ public class SessionExample {
     }
   }
```
samperson1997: In this md file you forget to add `delete()` in main method : )

session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -173,6 +173,19 @@ public synchronized TSRPCResp insert(String deviceId, long time, List<String> me
     }
   }
 
+  public synchronized TSRPCResp delete(String deviceId, String measurementId, long time) throws IoTDBSessionException {
```
qiaojialin: The parameters could include a path and a time

example/session/src/main/java/org/apache/iotdb/SessionExample.java
```diff
@@ -63,6 +64,14 @@ private static void insert() throws IoTDBSessionException {
     }
   }
 
+  private static void delete() throws IoTDBSessionException {
+    String deviceId = "root.sg1.d1";
+    String measurementId = "s1";
+    for (long time = 0; time < 30000; time++) {
+      session.delete(deviceId, measurementId, time);
```
qiaojialin: The deleteTime in DeletePlan means deleting all data points in this timeseries whose time is <= deleteTime. Therefore, onle one statement could do the for(): session.delete(deviceId, measurementId, 29999); You can add a comment for the deleteTime field in the DeletePlan.

An easier interface: session.delete("root.sg1.d1.s1", 29999);

qiaojialin:
After you finish the coding. Please add an IT for Session Interfaces. You can refer to the IoTDBCompleteIT. Use session api to set storage group, create timeseries, insert data and delete data... Then use JDBC API to query.

session/src/test/java/org/apache/iotdb/session/utils/EnvironmentUtils.java
```diff
@@ -0,0 +1,195 @@
...
+import org.junit.Assert;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * <p>
+ * This class is used for cleaning test environment in unit test and integration test
+ * </p>
+ *
+ * @author liukun
```
qiaojialin: remove this

service-rpc/src/main/thrift/rpc.thrift
```diff
@@ -304,5 +309,7 @@ service TSIService {
 
 	TSRPCResp insertRow(1:TSInsertReq req);
 
+	TSRPCResp deleteTimeseries(1:TSDeleteReq req);
```
qiaojialin: deleteTimeseries means deleting all data and schema of a timeseries, this should be called delete

jack870131: It would cause error message in thrift because 'delete' is the reserved word, which would show the message below:
```
[ERROR] thrift failed error: [ERROR:C:/Users/asus/Documents/Java/incubator-iotdb/service-rpc/src/main/thrift/rpc.thrift:312] (last token was 'delete')
Cannot use reserved language keyword: "delete"
```

jack870131: I have changed 'deleteTimeseries' into 'deleteRow'. What do you think?

session/src/test/java/org/apache/iotdb/session/utils/EnvironmentUtils.java
jack870131: Done.

service-rpc/src/main/thrift/rpc.thrift
qiaojialin: deleteData is better

PR #380, opened 2019-09-06
Move a util class to test dir and fix some minor problems

Genius-pig:
Move a util class to test dir and fix some minor problem


tsfile/src/main/java/org/apache/iotdb/tsfile/write/writer/TsFileIOWriter.java
```diff
@@ -447,4 +447,12 @@ public void filterChunks(Map<Path, List<Long>> chunkStartTimes) {
       }
     }
   }
+
+  /**
+   * this function is only for Test.
+   * @return TsFileOutput
+   */
+  public TsFileOutput getIOWriterOut() {
+    return this.out;
```
qiaojialin: why do you add this method...

PR #383, opened 2019-09-08
Add more partition file tests in spark-tsfile-connector

LeiRui:
After reviewing pr#303 (transform old form to new form and use a parameter to control), I want to improve the spark-tsfile-connector module with this pr. I did three things: 
- add more tests for partition file cases in the TsFileSuit. Thank @SilverNarcissus for making this possible. 
- polish codes
  - in DefaultSource.builderReader: move queriedSchema behind because narrow table does not use it
  - in DefaultSource.builderReader: remove the queryNext() in next() function because hasNext() has done the job
  - in DefaultSource and two converters: change in.close() to readTsFile.close() and reader.close() respectively
- code format because some codes from pr#303 are not formatted using java-google-style.

### DIFF
--- a/spark-tsfile/src/main/java/org/apache/iotdb/tsfile/qp/Executor.java
+++ b/spark-tsfile/src/main/java/org/apache/iotdb/tsfile/qp/Executor.java
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- * <p>
- * http://www.apache.org/licenses/LICENSE-2.0
- * <p>
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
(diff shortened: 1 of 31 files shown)


SilverNarcissus:
Thanks for your improvement. It's awesome!
I have a question about the test. The order of the result will be determined in any case? As far as I'm concerned, the order of a dataframe is not determined unless you use ORDER BY in sql.

LeiRui:
> Thanks for your improvement. It's awesome!
> I have a question about the test. The order of the result will be determined in any case? As far as I'm concerned, the order of a dataframe is not determined unless you use ORDER BY in sql.

Which test? 
As far as I know, suppose the file is partitioned into 13 partitions, with partition 0 being assigned to task 0, partition 1 being assigned to task 1, ..., then the data in the result table is combined by data from task 0, task 1, .... sequentially.

SilverNarcissus:
> Which test?
> As far as I know, suppose the file is partitioned into 13 partitions, with partition 0 being assigned to task 0, partition 1 being assigned to task 1, ..., then the data in the result table is combined by queried results from task 0, task 1, .... sequentially.

Wow. if spark works like above, my question has been answered.  👍

LeiRui:
If there is no problem, can someone approve this?

PR #384, opened 2019-09-09
[IOTDB-57]rename

little-emotion:
rename mean to avg

### DIFF
--- a/server/src/test/java/org/apache/iotdb/db/integration/IOTDBGroupByIT.java
+++ b/server/src/test/java/org/apache/iotdb/db/integration/IOTDBGroupByIT.java
@@ -127,7 +127,7 @@ public class IOTDBGroupByIT {
   }
 
   @Test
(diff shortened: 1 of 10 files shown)


server/src/test/java/org/apache/iotdb/db/integration/IOTDBGroupByIT.java
```diff
@@ -127,7 +127,7 @@ public void tearDown() throws Exception {
   }
 
   @Test
-  public void countSumMeanTest() throws SQLException {
+  public void countSumavgTest() throws SQLException {
```
LeiRui: countSumAvgTest

little-emotion: fixed

qiaojialin:
Does the doc need to be updated?

PR #386, opened 2019-09-10
[IOTDB-209]Improvement on the Hadoop module

JackieTien97:


LeiRui:
Hi, the travis test reports an error:

[ERROR] testBytesArrayReadAndWrite(org.apache.iotdb.tsfile.hadoop.HDFSInputAndOutputTest) Time elapsed: 0.077 s <<< ERROR! java.io.IOException: (null) entry in command string: null chmod 0644 C:\Users\travis\build\apache\incubator-iotdb\hadoop\testHDFSInputAndOutput.tsfile at org.apache.iotdb.tsfile.hadoop.HDFSInputAndOutputTest.testBytesArrayReadAndWrite(HDFSInputAndOutputTest.java:77)

JackieTien97:
It seems that our windows environment's hadoop configuration doesn't work well, have we set the %HADOOP_HOME% to a folder with winutils.exe in the windows test environment? The following link points out how to configure that.
https://stackoverflow.com/questions/40764807/null-entry-in-command-string-exception-in-saveastextfile-on-pyspark

JackieTien97:
There is a submodule named example in hadoop module, I have write the examples there. Need I move it to example module?

qiaojialin:
> There is a submodule named example in hadoop module, I have write the examples there. Need I move it to example module?

yes

JackieTien97:
> > There is a submodule named example in hadoop module, I have write the examples there. Need I move it to example module?
> 
> yes

hadoop/src/main/java/org/apache/iotdb/tsfile/hadoop/TSFInputSplit.java
```diff
@@ -141,47 +85,203 @@ public long getLength() throws IOException, InterruptedException {
...
+    private String deviceId;
+
+    /**
+     * Name Of all the sensors in the device.
+     */
+    private String[] measurementIds;
+
+    /**
+     * Byte offset of the corresponding data in the file
+     * Notice: include the chunk group header and marker.
```
jt2594838: I don't think we have chunk group header.

hadoop/src/main/java/org/apache/iotdb/tsfile/hadoop/TSFRecordReader.java
```diff
@@ -155,58 +162,65 @@ public NullWritable getCurrentKey() throws IOException, InterruptedException {
...
-      index = 1;
+
+    if (isReadTime && isReadDeviceId) { // Both time and deviceId need to be written into value
+      mapWritable.put(new Text("timestamp"), time);
+      mapWritable.put(new Text("device_id"), deviceIdText);
+    } else if (isReadTime) { // Only Time needs to be written into value
+      mapWritable.put(new Text("timestamp"), time);
+    } else if (isReadDeviceId) { // Only deviceId need to be written into value
+      mapWritable.put(new Text("device_id"), deviceIdText);
     }
```
jt2594838: This might be simplified with
```
if (isReadTime) {
   mapWritable.put(new Text("timestamp"), time);
 }
 if (isReadDeviceId) { 
   mapWritable.put(new Text("device_id"), deviceIdText);
 }
```

hadoop/src/main/java/org/apache/iotdb/tsfile/hadoop/TSFRecordReader.java
```diff
@@ -216,12 +230,12 @@ public float getProgress() throws IOException, InterruptedException {
 
   @Override
   public void close() throws IOException {
-    dataSet = null;
-    hdfsInputStream.close();
+    dataSetList.forEach(queryDataSet -> queryDataSet = null);
+    deviceIdList.forEach(deviceId -> deviceId = null);
```
jt2594838: What is the point of setting the content to null instead of calling clear() or just set the fields to null?

hadoop/src/main/java/org/apache/iotdb/tsfile/hadoop/record/HDFSTSRecord.java
```diff
@@ -0,0 +1,204 @@
...
+
+  @Override
+  public void write(DataOutput out) throws IOException {
+    out.writeLong(time);
+    out.writeUTF(deviceId);
+    out.writeInt(dataPointList.size());
+    for (DataPoint dataPoint : dataPointList) {
+      out.writeShort(dataPoint.getType().serialize());
+      out.writeInt(dataPoint.getMeasurementId().getBytes(StandardCharsets.UTF_8).length);
+      out.write(dataPoint.getMeasurementId().getBytes(StandardCharsets.UTF_8));
```
jt2594838: Why not just use `writeUTF()` as you have done previously?

hadoop/src/main/java/org/apache/iotdb/tsfile/hadoop/TSFInputSplit.java
JackieTien97: Aha, I just copy that from the ChunkGroupMetaData, I will fix it in both files.

hadoop/src/main/java/org/apache/iotdb/tsfile/hadoop/record/HDFSTSRecord.java
JackieTien97: If there are spaces in measurementId, and I use writeUTF(), when I read, I can only use the corresponding readUTF(), I will miss the characters after the first space. This will also cause the rest data to be unusable.

JackieTien97: And I should change the previous deviceId's reading and writing way.

PR #387, opened 2019-09-10
[IOTDB-180] Get rid of JSON format in "show timeseries"

nlosilva1:
JIRA link: [](https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-180)


nlosilva1:
> The README.md and quickstart in docs may need to update

In the README.md and the quickstart in docs/documentation , the example of this query is using "show timeseries root". But the file docs/documentation-CHN/quickstart.md is different from the one in English version, but the same as the quickstart on iotdb.apache.org [1], with "show timeseries" and some other differences between them. Shall we unify all of them as a same version?

 [1]https://iotdb.apache.org/#/Documents/Quick%20Start

PR #399, opened 2019-09-19
[IOTDB-224] fix buffer size and remove cache

SilverNarcissus:
This pr is to fix bug referred in IOTDB-224


LeiRui:
The title of this pr?

PR #415, opened 2019-09-24
[IOTDB-221] Add C++ client

humanfy:


client-cpp/src/IOTDBSession.cpp
```diff
@@ -0,0 +1,236 @@
...
+    TSCloseSessionReq* req = new TSCloseSessionReq();
+    req->__set_sessionHandle(*sessionHandle);
+    try 
+    {
+        TSStatus *resp = new TSStatus();
+        client->closeSession(*resp,*req);
+    } 
+    catch (TException e) 
+    {
+        throw IoTDBSessionException("Error occurs when closing session at server. Maybe server is down.");
```
jt2594838: Better to preserve the error message in e.

client-cpp/compile.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
```
chrisdutz: This would result in a build that excludes Windows users.

client-cpp/src/SessionExample.cpp
```diff
@@ -0,0 +1,67 @@
...
+    
+    vector<string> paths;
+    paths.push_back("root.sg1.d1.s1");
+    paths.push_back("root.sg1.d1.s2");
+    paths.push_back("root.sg1.d1.s3");
+    session->deleteTimeseries(paths);
+    
+    session->deleteStorageGroup("root.sg1");
+    
+    session->close();
```
jt2594838: Please show how to perform a query, which is important.

jixuan1989:
Hi, I will close this PR because a new c++ module is added to the master branch in #1616

PR #417, opened 2019-09-24
[IOTDB-234] Refactor TsFile storage on HDFS

samperson1997:
Refactor TsFile storage on HDFS codes:
* Extract HDFS related files (`HDFSFile`, `HDFSInput`, `HDFSOutput`) into Hadoop module from TSFile module, so that TSFile module will not depend on Hadoop libs.
* Use Java Reflection to get HDFS related files for factories used in TSFile module.
* Use`FSFactoryProducer` to produce `LocalFSFactory` / `HDFSFactory`, `LocalFSInputFactory`/ `HDFSInputFactory` and `LocalFSOutputFactory`/ `HDFSOutputFactory` according to user config, which will be less heavy and slow as well as be more scalable for future design.

**Usage**
1. Build server and Hadoop module by: `mvn clean package -pl server,hadoop -am -Dmaven.test.skip=true`;
2. Copy the target jar of Hadoop module `hadoop-tsfile-0.9.0-SNAPSHOT-jar-with-dependencies.jar` into server target lib folder `.../server/target/iotdb-server-0.9.0-SNAPSHOT/lib`.
Otherwise an error will occur: `ERROR org.apache.iotdb.tsfile.fileSystem.fsFactory.HDFSFactory:62 - Failed to get Hadoop file system. Please check your dependency of Hadoop module. `
3. Edit user config in `iotdb-engine.properties`. Start server and Tsfile will be stored on HDFS.


samperson1997:
> I think using reflection every time is too heavy and slow when all you want to do is to get a new Object. To put it simply, reflections should be used as less frequently as possible.
> 
> Here is my suggestion: use two factories (the LocalFactory and the HDFSFactory) instead of one, and initialize one of them using reflection according to the configuration at the system start-up. Thus the occurrence of reflections can be reduced to only 1.

Hi, many thanks for your useful suggestion!

In the new commit, I tried to reduce the occurrence of reflections to only 1 by initializing `private static Class<?> clazz` when constructing `TSFileFactory`. I think this also reached your main purpose, so I didn't use two factories... Do you think it is acceptable? > <

jt2594838:
If you insist on this design, at least you should save the constructor.
And this test will show you that this is still slower compared with a direct call to new().
[ReflectionTest.txt](https://github.com/apache/incubator-iotdb/files/3722994/ReflectionTest.txt)

samperson1997:
> If you insist on this design, at least you should save the constructor.
> And this test will show you that this is still slower compared with a direct call to new().
> [ReflectionTest.txt](https://github.com/apache/incubator-iotdb/files/3722994/ReflectionTest.txt)

Thanks for the suggested design  @jt2594838 ! I have accepted it by using two factories (`LocalFSFactory` and `HDFSFactory`) with `FSFactoryProducer` to produce either one of them according to user config. I think this design is so much better for being less heavy and slow and being more scalable for future design.

tsfile/src/main/java/org/apache/iotdb/tsfile/fileSystem/HDFSFactory.java
```diff
@@ -0,0 +1,178 @@
...
+          "Failed to get file: {}. Please check your dependency of Hadoop module.",
+          uri.toString(), e);
+      return null;
+    }
+  }
+
+  public BufferedReader getBufferedReader(String filePath) {
+    try {
+      return (BufferedReader) clazz.getMethod("getBufferedReader", String.class)
+          .invoke(clazz.newInstance(), filePath);
```
jt2594838: I think you can save the getter Methods as you have saved the Constructors.

tsfile/src/main/java/org/apache/iotdb/tsfile/fileSystem/FileInputFactory.java
```diff
@@ -19,33 +19,39 @@
...
 import org.apache.iotdb.tsfile.read.reader.DefaultTsFileInput;
 import org.apache.iotdb.tsfile.read.reader.TsFileInput;
-import org.apache.iotdb.tsfile.write.writer.TsFileIOWriter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.IOException;
-import java.nio.file.Paths;
-
 public enum FileInputFactory {
```
jt2594838: Better to perform your refactoring on these two factories, too.

samperson1997: Should I separate these different factories into three packages? like `fsFactory`, `fileInputFactory` and `fileOutputFactory`

tsfile/src/main/java/org/apache/iotdb/tsfile/fileSystem/fileInputFactory/HDFSInputFactory.java
```diff
@@ -17,36 +17,22 @@
...
 
-public enum FileInputFactory {
+public class HDFSInputFactory implements FileInputFactory {
 
-  INSTANCE;
-
-  private static FSType fsType = TSFileDescriptor.getInstance().getConfig().getTSFileStorageFs();
-  private static final Logger logger = LoggerFactory.getLogger(FileInputFactory.class);
+  private static final Logger logger = LoggerFactory.getLogger(HDFSInputFactory.class);
   private static Class<?> clazz;
```
jt2594838: Save the constructor is better.

samperson1997: Sorry for the inconsideration! I have fixed it.

tsfile/src/main/java/org/apache/iotdb/tsfile/fileSystem/fileInputFactory/HDFSInputFactory.java
```diff
@@ -27,13 +28,21 @@
...
   private static final Logger logger = LoggerFactory.getLogger(HDFSInputFactory.class);
-  private static Class<?> clazz;
+  private static Constructor constructor;
+
+  static {
+    try {
+      Class<?> clazz = Class.forName("org.apache.iotdb.tsfile.fileSystem.HDFSInput");
+      constructor = clazz.getConstructor(String.class);
+    } catch (ClassNotFoundException | NoSuchMethodException e) {
+      e.printStackTrace();
```
jt2594838: Do not use stack print, use logger instead.
But on this occasion, I would suggest you throw a runtime exception (the best to be customized) because this is a fatal error.

tsfile/src/main/java/org/apache/iotdb/tsfile/fileSystem/fsFactory/HDFSFactory.java
```diff
@@ -0,0 +1,190 @@
...
+      constructorWithUri = clazz.getConstructor(URI.class);
+      getBufferedReader = clazz.getMethod("getBufferedReader", String.class);
+      getBufferedWriter = clazz.getMethod("getBufferedWriter", String.class, boolean.class);
+      getBufferedInputStream = clazz.getMethod("getBufferedInputStream", String.class);
+      getBufferedOutputStream = clazz.getMethod("getBufferedOutputStream", String.class);
+      listFilesBySuffix = clazz.getMethod("listFilesBySuffix", String.class, String.class);
+      listFilesByPrefix = clazz.getMethod("listFilesByPrefix", String.class, String.class);
+    } catch (ClassNotFoundException | NoSuchMethodException e) {
+      logger.error(
+          "Failed to get Hadoop file system. Please check your dependency of Hadoop module.", e);
```
qiaojialin: So, even if we store data files on the local file system, the server will also print an error log? This should be avoided.

tsfile/src/main/java/org/apache/iotdb/tsfile/fileSystem/FSFactoryProducer.java
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.iotdb.tsfile.fileSystem.fileOutputFactory.HDFSOutputFactory;
+import org.apache.iotdb.tsfile.fileSystem.fileOutputFactory.LocalFSOutputFactory;
+import org.apache.iotdb.tsfile.fileSystem.fsFactory.FSFactory;
+import org.apache.iotdb.tsfile.fileSystem.fsFactory.HDFSFactory;
+import org.apache.iotdb.tsfile.fileSystem.fsFactory.LocalFSFactory;
+
+public class FSFactoryProducer {
+
+  private static FSType fSType = TSFileDescriptor.getInstance().getConfig().getTSFileStorageFs();
+  private static HDFSFactory hdfsFactory = new HDFSFactory();
```
qiaojialin: avoid constructing all fileds in static field

tsfile/src/main/java/org/apache/iotdb/tsfile/fileSystem/fileInputFactory/FileInputFactory.java
```diff
@@ -0,0 +1,28 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iotdb.tsfile.fileSystem.fileInputFactory;
+
+import org.apache.iotdb.tsfile.read.reader.TsFileInput;
+
+public interface FileInputFactory {
```
qiaojialin: why not combine Input and output with FSFactory?

tsfile/src/main/java/org/apache/iotdb/tsfile/fileSystem/fsFactory/HDFSFactory.java
samperson1997: Hi, thanks for  your checking.  I had solved this by avoiding constructing all filed in static field in `FSFactoryProducer`.

tsfile/src/main/java/org/apache/iotdb/tsfile/fileSystem/fileInputFactory/FileInputFactory.java
samperson1997: I think there would be too many methods in one factory in this way... By separating them, we can  easily get the meaning of each factory.

tsfile/src/main/java/org/apache/iotdb/tsfile/fileSystem/fsFactory/HDFSFactory.java
jt2594838: I do not think so. The static codes are run upon the class is loaded. So maybe unless you actually use it, you won't find this log.

tsfile/src/main/java/org/apache/iotdb/tsfile/fileSystem/FSFactoryProducer.java
```diff
@@ -33,36 +33,29 @@
 public class FSFactoryProducer {
 
   private static FSType fSType = TSFileDescriptor.getInstance().getConfig().getTSFileStorageFs();
-  private static HDFSFactory hdfsFactory = new HDFSFactory();
-  private static LocalFSFactory localFSFactory = new LocalFSFactory();
-  private static HDFSInputFactory hdfsInputFactory = new HDFSInputFactory();
-  private static LocalFSInputFactory localFSInputFactory = new LocalFSInputFactory();
-  private static HDFSOutputFactory hdfsOutputFactory = new HDFSOutputFactory();
-  private static LocalFSOutputFactory localFSOutputFactory = new LocalFSOutputFactory();
-
```
jt2594838:
```
  private static FSFactory fsFactory;
  private static FSInputFactory fsInputFactory;
  private static FSOutputFactory fsOutputFactory;
  
  static {
        if (fSType.equals(FSType.HDFS)) {
              // init hdfs factories
        } else {
             // init local factories
        }
  }
```
This may avoid calling too many new(), and hopefully, when not using HDFS you won't get any HDFS log.

PR #421, opened 2019-09-25
[IOTDB-239] Add interface for showing devices

samperson1997:
Create an interface for showing all the devices.

USAGE: `SHOW DEVICES`.

### DIFF
--- a/jdbc/src/test/java/org/apache/iotdb/jdbc/IoTDBDatabaseMetadataTest.java
+++ b/jdbc/src/test/java/org/apache/iotdb/jdbc/IoTDBDatabaseMetadataTest.java
@@ -163,7 +163,7 @@ public class IoTDBDatabaseMetadataTest {
     String standard = "count,\n" + "4,\n";
     try {
       IoTDBDatabaseMetadata metadata = (IoTDBDatabaseMetadata) databaseMetaData;
-      String level = "3";
+      int level = 3;
       ResultSet resultSet = metadata.getNodes(Constant.COUNT_NODES, "root", null, null, level);
       ResultSetMetaData resultSetMetaData = resultSet.getMetaData();
       int colCount = resultSetMetaData.getColumnCount();
@@ -205,7 +205,7 @@ public class IoTDBDatabaseMetadataTest {
             + "root.vehicle.d4,2,\n";
     try {
       IoTDBDatabaseMetadata metadata = (IoTDBDatabaseMetadata) databaseMetaData;
-      String level = "3";
+      int level = 3;
       ResultSet resultSet = metadata.getNodes(Constant.COUNT_NODE_TIMESERIES, "root", null, null, level);
       ResultSetMetaData resultSetMetaData = resultSet.getMetaData();
       int colCount = resultSetMetaData.getColumnCount();
@@ -269,7 +269,7 @@ public class IoTDBDatabaseMetadataTest {
     Set<String> devicesSet = new HashSet<>();
     devicesSet.add("root.vehicle.d0");
 
-    when(fetchMetadataResp.getShowDevices()).thenReturn(devicesSet);
+    when(fetchMetadataResp.getDevices()).thenReturn(devicesSet);
 
     String standard = "Device,\n" + "root.vehicle.d0,\n";
     try {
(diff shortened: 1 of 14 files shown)


server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
```diff
@@ -760,6 +764,15 @@ private void findStorageGroup(MNode node, String path, HashSet<String> res) {
     }
   }
 
+  /**
+   * Get all devices in current Metadata Tree.
+   *
+   * @return a list contains all distinct storage groups
```
qiaojialin:
```suggestion
   * @return a list contains all distinct devices
```

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
```diff
@@ -760,6 +764,15 @@ private void findStorageGroup(MNode node, String path, HashSet<String> res) {
...
+   */
+  HashSet<String> getAllDevices() {
+    return new HashSet<>(getNodesList("3"));
```
qiaojialin: why not using an int?

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -311,15 +311,9 @@ public TSFetchMetadataResp fetchMetadata(TSFetchMetadataReq req) {
...
-          String column = req.getColumnPath();
-          Map<String, List<String>> deviceMap = metadata.getDeviceMap();
-          if (deviceMap == null || !deviceMap.containsKey(column)) {
-            resp.setColumnsList(new ArrayList<>());
-          } else {
-            resp.setColumnsList(deviceMap.get(column));
-          }
+        case "SHOW_DEVICES":
+          Set<String> devices = getAllDevices();
+          resp.setShowDevices(devices);
```
qiaojialin:
```suggestion
          resp.setDevices(devices);
```

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
samperson1997: `getNodesList` method is implemented before and I just call it here, should I change the param type to int? (which I think is better as well)

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
samperson1997: Here I also keep consistent with `SHOW TIMESERIES` and `SHOW STORAGE GROUP` interface, in which they uses `setShowTimeseriesList` and `setShowStorageGroups`... Should I change all of these interfaces? (and I also think your suggestion is better...)

jdbc/src/test/java/org/apache/iotdb/jdbc/IoTDBDatabaseMetadataTest.java
```diff
@@ -240,7 +240,41 @@ public void device() throws Exception {
...
+      resultStr.append("\n");
+      while (resultSet.next()) {
+        for (int i = 1; i <= colCount; i++) {
+          resultStr.append(resultSet.getString(i)).append(",");
+        }
+        resultStr.append("\n");
+      }
+      Assert.assertEquals(resultStr.toString(), standard);
+    } catch (SQLException e) {
+      System.out.println(e);
```
kr11: Simply  print out the exception may bury the error and the TestCase will be regarded as `PASSED`. It's better to replace it with `e.printStackTrace(); fail(e.getMessage());`.

server/src/main/java/org/apache/iotdb/db/metadata/MGraph.java
```diff
@@ -248,6 +248,10 @@ public Metadata getMetadata() throws PathErrorException {
     return mtree.getAllStorageGroup();
   }
 
+  HashSet<String> getAllDevices() {
```
kr11: I note that all its callers Interprete the returned value as `Set` but not `HashSet`.

samperson1997:
Hi @qiaojialin @kr11 ,

Thanks a lot for your detailed review! I have updated codes as you suggested.

PR #424, opened 2019-09-26
[IOTDB-241] Add query and non query interface in session

SilverNarcissus:
Add query and non query interface in session

### DIFF
--- a/session/src/main/java/org/apache/iotdb/session/Session.java
+++ b/session/src/main/java/org/apache/iotdb/session/Session.java
@@ -368,9 +368,4 @@ public class Session {
 
     RpcUtils.verifySuccess(execResp.getStatus());
   }
-
-  // the next record rule without constraints
-
-
-
 }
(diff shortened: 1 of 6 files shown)


session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -294,4 +320,49 @@ public synchronized void setTimeZone(String zoneId) throws TException, IoTDBRPCE
     this.zoneId = ZoneId.of(zoneId);
   }
 
+  /**
+   * check whether this sql is for query
+   * @param sql sql
+   * @return whether this sql is for query
+   */
+  private boolean checkIsQuery(String sql) {
+    return sql.trim().toLowerCase().startsWith("select");
```
jt2594838: I don't think this is enough. Statements like "show" or "list" are also treated as queries.

session/src/main/java/org/apache/iotdb/session/SessionDataSet.java
```diff
@@ -0,0 +1,97 @@
...
+
+    getFlag = false;
+    return record;
+  }
+
+
+  private boolean nextWithoutConstraints(String sql, long queryId)
+      throws SQLException, IoTDBRPCException {
+    if ((recordItr == null || !recordItr.hasNext())) {
+      TSFetchResultsReq req = new TSFetchResultsReq(sql, 512, queryId);
```
jt2594838: I wonder if it is necessary to make the fetchSize hard code.

PR #429, opened 2019-09-26
[IOTDB-205]Support storage-group-level data ttl

jt2594838:
https://issues.apache.org/jira/browse/IOTDB-205


jt2594838:
A design document in CHINESE.
[数据定期删除设计文档.docx](https://github.com/apache/incubator-iotdb/files/3660675/default.docx)

jt2594838:
Experiment settings and results. Pay extra attention to the changes in file sizes.
[config.properties.txt](https://github.com/apache/incubator-iotdb/files/3660682/config.properties.txt)
[iotdb-engine.properties.txt](https://github.com/apache/incubator-iotdb/files/3660683/iotdb-engine.properties.txt)
[insertTestWithDefaultPath_IoTDB_TTL_Write11569292133990.xlsx](https://github.com/apache/
[SERVER_MODE_fit_10_2019_09_24.xlsx](https://github.com/apache/incubator-iotdb/files/3660685/SERVER_MODE_fit_10_2019_09_24.xlsx)
incubator-iotdb/files/3660684/insertTestWithDefaultPath_IoTDB_TTL_Write11569292133990.xlsx)

LeiRui:
Hi, 
This is a review on functionality.
I find the operation `unset TTL to <storage group>` REALLY requires a seasoned user who knows exactly what he/she is doing. And I totally agree with what you have noted in the guidance:
> Notice: When you set TTL to some storage groups, data out of the TTL will be made invisible
immediately, but because the data files may contain both out-dated and living data or the data files may be being used by queries, the physical removal of data is stale. If you increase or unset TTL
just after setting it previously, some previously invisible data may be seen again, but the
physically removed one is lost forever. **In other word, different from delete statement, the
atomicity of data deletion is not guaranteed for efficiency concerns.** **So we recommend that you do not change the TTL once it is set or at least do not reset it frequently, unless you are determined to suffer the unpredictability.** 

I want to draw attention to another point that can be easily misunderstood when using the operation `unset TTL to <storage group>`. That is, the latest insertion time of a storage group, which is the split point used to determine whether the coming data point is sequential or unsequential, will not be “recycled" even if the data before `now()-TTL` are really deleted.

jt2594838:
That is true. We currently do not provide any way to "recycle" the last flushed time of a device other than DELETE STORAGE GROUP statements, even general DELETE DATA statements can not do that. And the "recycling" does not seem to bring many benefits since, on most occasions, data is rolling forward.

Actually, the data TTL has little relation to the last flushed time. And whether the data goes into sequence files or unsequence files should be invisible to users. I do not think users should be concerned about that.

What I have hidden in my implementation is that "UNSET TTL TO root.group1" equals "SET TTL TO root.group1 Long.maxValue". In all, data TTL just avoids the DBA deleting old data manually.

LeiRui:
> And whether the data goes into sequence files or unsequence files should be invisible to users.

I agree.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -580,6 +610,65 @@ public void syncDeleteDataFiles() {
...
+    // copy to avoid concurrent modification of deletion
+    List<TsFileResource> seqFiles = new ArrayList<>(sequenceFileList);
+    List<TsFileResource> unseqFiles = new ArrayList<>(unSequenceFileList);
+
+    for (TsFileResource tsFileResource : seqFiles) {
+      checkFileTTL(tsFileResource, timeBound, true);
+    }
+    for (TsFileResource tsFileResource : unseqFiles) {
+      checkFileTTL(tsFileResource, timeBound, false);
+    }
```
JackieTien97: Why not use Iterator.remove() to avoid concurrent modification of deletion? It can also save the extra list copy time and space.

server/src/main/java/org/apache/iotdb/db/engine/merge/manage/MergeResource.java
```diff
@@ -68,9 +68,21 @@
...
+  }
+
+  public MergeResource(List<TsFileResource> seqFiles, List<TsFileResource> unseqFiles,
+      long timeBound) {
+    this.seqFiles =
+        seqFiles.stream().filter(res -> res.isClosed() && !res.isDeleted()
+            && res.stillLives(timeBound)).collect(Collectors.toList());
+    this.unseqFiles =
+        unseqFiles.stream().filter(res -> res.isClosed() && !res.isDeleted()
+            && res.stillLives(timeBound)).collect(Collectors.toList());
```
JackieTien97: Every time the function `MergeResource` is invoked, two new class will be created because the `timeBound` in lambda express may be different among different invocations.
It's better to write a class to implement the interface Predicate<T> and maintain a field named `timeBound`. And set the field to different value in the function `MergeResource`.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileProcessor.java
```diff
@@ -576,6 +579,8 @@ public String getStorageGroupName() {
       QueryUtils.modifyChunkMetaData(chunkMetaDataList,
           modifications);
 
+      chunkMetaDataList.removeIf(chunkMetaData -> chunkMetaData.getEndTime() < timeLowerBound);
```
JackieTien97: Still the same question. The variable `timeLowerBound` will cause class construction every time the function is invoked.

server/src/main/java/org/apache/iotdb/db/metadata/MetadataOperationType.java
```diff
@@ -34,5 +34,6 @@ private MetadataOperationType(){
   public static final String UNLINK_MNODE_FROM_PTREE = "7";
   public static final String ADD_INDEX_TO_PATH = "8";
   public static final String DELETE_INDEX_FROM_PATH = "9";
-  public static final String DELETE_STORAGE_GROUP_FROM_MTREE = "10";
+  public static final String SET_TTL = "10";
+  public static final String DELETE_STORAGE_GROUP_FROM_MTREE = "11";
```
JackieTien97: Why not directly add a field SET_TTL = "11"?

server/src/main/java/org/apache/iotdb/db/qp/logical/Operator.java
```diff
@@ -73,6 +73,7 @@ public String toString() {
     GRANT_USER_PRIVILEGE, REVOKE_USER_PRIVILEGE, GRANT_USER_ROLE, REVOKE_USER_ROLE, CREATE_ROLE,
     DELETE_ROLE, GRANT_ROLE_PRIVILEGE, REVOKE_ROLE_PRIVILEGE, LIST_USER, LIST_ROLE,
     LIST_USER_PRIVILEGE, LIST_ROLE_PRIVILEGE, LIST_USER_ROLES, LIST_ROLE_USERS,
-    GRANT_WATERMARK_EMBEDDING, REVOKE_WATERMARK_EMBEDDING, DELETE_STORAGE_GROUP,
+    GRANT_WATERMARK_EMBEDDING, REVOKE_WATERMARK_EMBEDDING,
+    TTL, DELETE_STORAGE_GROUP
```
JackieTien97: Why put  TTL before DELETE_STORAGE_GROUP instead of adding it to the end.

server/src/main/java/org/apache/iotdb/db/utils/TestOnly.java
```diff
@@ -0,0 +1,31 @@
...
+package org.apache.iotdb.db.utils;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ElementType.METHOD, ElementType.CONSTRUCTOR})
+@Retention(RetentionPolicy.SOURCE)
+public @interface TestOnly {
```
JackieTien97: Does this mean that any function with this annotation should only be used in UT and IT?

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
jt2594838: image
Unfortunately, the Iterator of ArrayList does not provide concurrent safety, either.

server/src/main/java/org/apache/iotdb/db/engine/merge/manage/MergeResource.java
jt2594838: Yes, but considering the frequency of merge (at least every tens of minutes), such improvement is negligible.
Still, I replace the lambda with a method reference, please have a look.

server/src/main/java/org/apache/iotdb/db/metadata/MetadataOperationType.java
jt2594838: Because when I created the field SET_TTL, there is no such field DELETE_STORAGE_GROUP_FROM_MTREE. And later I merged master which brought DELETE_STORAGE_GROUP_FROM_MTREE.

server/src/main/java/org/apache/iotdb/db/qp/logical/Operator.java
jt2594838: See the previous reply.

server/src/main/java/org/apache/iotdb/db/utils/TestOnly.java
jt2594838: Yes. Although some external packages already include this annotation, I do not want to import them just for adding one annotation.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileProcessor.java
jt2594838: This will require maintaining a ThreadLocal predicate for each query. As far as I am concerned, this works no different from using a lambda since new object construction is unavoidable.

Somehow, I moved `timeLowerBound` into `QueryContext` and replaced this with a method reference.

jt2594838:
Creating a new inner class object or creating a lambda does not seem quite different to me. But I hope using method reference could help a bit.

server/src/main/java/org/apache/iotdb/db/engine/merge/manage/MergeResource.java
```diff
@@ -65,24 +65,28 @@
...
+    this.seqFiles =
+        seqFiles.stream().filter(this::filterResource).collect(Collectors.toList());
     this.unseqFiles =
-        unseqFiles.stream().filter(res -> res.isClosed() && !res.isDeleted())
-            .collect(Collectors.toList());
+        unseqFiles.stream().filter(this::filterResource).collect(Collectors.toList());
+  }
+
+  private boolean filterResource(TsFileResource res) {
+    return res.isClosed() && !res.isDeleted() && res.stillLives(fileTimeBound);
```
JackieTien97: In the before lambda, you didn't judge res.stillLives(fileTimeBound), but in the `filterResource` you judge it. Will there be some bugs after changing it?

fanhualta:
Good job, please fix the conflicts.

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -309,7 +310,7 @@ public TSFetchMetadataResp fetchMetadata(TSFetchMetadataReq req) {
           status = new TSStatus(getStatus(TSStatusCode.SUCCESS_STATUS));
           break;
         case "SHOW_STORAGE_GROUP":
-          Set<String> storageGroups = getAllStorageGroups();
+          Set<String> storageGroups = new HashSet<>(getAllStorageGroups());
```
fanhualta: I think it's unnecessary to wrap `getAllStorageGroups()` with a `new HashSet<>`

server/src/main/java/org/apache/iotdb/db/engine/merge/manage/MergeResource.java
jt2594838: No, because the default value of fileTimeBound is Long.MIN_VALUE, when I call `stillLives()`, it will return true immediately, so it equals to the previous version.

jt2594838: By the way, if you are satisfied with my answers, please resolve the conversations.

docs/Documentation/UserGuide/5-IoTDB SQL Documentation/1-IoTDB Query Statement.md
```diff
@@ -578,3 +578,48 @@ SELECT SUM(Path) (COMMA SUM(Path))* FROM <FromClause> [WHERE <WhereClause>]?
...
+This example will show TTLs of all storage groups.
+Eg.2 SHOW TTL ON root.group1,root.group2,root.group3
+This example will show TTLs of the specified 3 groups.
+```
+
+Notice: When you set TTL to some storage groups, data out of the TTL will be made invisible
+immediately, but because the data files may contain both out-dated and living data or the data files may
+be being used by queries, the physical removal of data is stale. If you increase or unset TTL
+just after setting it previously, some previously invisible data may be seen again, but the
+physically removed one is lost forever. In other word, different from delete statement, the
```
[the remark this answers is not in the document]
jt2594838: fixed

server/src/main/java/org/apache/iotdb/db/engine/StorageEngine.java
```diff
@@ -103,11 +111,32 @@ private StorageEngine() {
...
   }
 
   @Override
   public void stop() {
     syncCloseAllProcessor();
+    ttlCheckThread.shutdownNow();
+    try {
+      ttlCheckThread.awaitTermination(30, TimeUnit.SECONDS);
+    } catch (InterruptedException e) {
+      logger.warn("TTL check thread still doesn't exist after 30s");
```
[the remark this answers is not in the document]
jt2594838: fixed

server/src/main/java/org/apache/iotdb/db/qp/executor/AbstractQueryProcessExecutor.java
```diff
@@ -57,11 +67,44 @@ public QueryDataSet processQuery(PhysicalPlan queryPlan, QueryContext context)
...
+
+    List<String> selectedSgs = showTTLPlan.getStorageGroups();
+
+    List<MNode> storageGroups = MManager.getInstance().getAllStorageGroups();
+    int i = 0;
+    for (MNode mNode : storageGroups) {
+      String sgName = mNode.getFullPath();
+      if (!selectedSgs.isEmpty() && !selectedSgs.contains(sgName)) {
+        continue;
+      }
```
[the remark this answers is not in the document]
jt2594838: I do not know if I got your point. If `selectedSgs.isEmpty()` holds, it means show all storage groups and I should only test the `sgName` when it is non-empty. If I move it outside, show all TTL will result in an empty result set.

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
jt2594838: The data structure required by the upper interface is a `Set`, while the lower interface returns a `List`, so I have to do the conversion.
But actually, it does not need to be a set, as we won't have duplicated SGs. I will change the upper interface to receive a `List` later.

server/src/main/java/org/apache/iotdb/db/qp/executor/AbstractQueryProcessExecutor.java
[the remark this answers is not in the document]
fanhualta: Ok, I get your point.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -862,7 +962,9 @@ public void merge(boolean fullMerge) {
       }
 
       long budget = IoTDBDescriptor.getInstance().getConfig().getMergeMemoryBudget();
-      MergeResource mergeResource = new MergeResource(sequenceFileList, unSequenceFileList);
+      long timeBound = System.currentTimeMillis() - dataTTL;
```
qiaojialin:
```suggestion
      long timeLowerBound = System.currentTimeMillis() - dataTTL;
```

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -878,9 +980,16 @@ public void merge(boolean fullMerge) {
         // cached during selection
         mergeResource.setCacheDeviceMeta(true);
 
+        for (TsFileResource tsFileResource : mergeResource.getSeqFiles()) {
+          tsFileResource.setMerging(true);
+        }
+        for (TsFileResource tsFileResource : mergeResource.getUnseqFiles()) {
+          tsFileResource.setMerging(true);
+        }
```
qiaojialin: Is this code used for fixing merge bug?

server/src/main/java/org/apache/iotdb/db/metadata/MGraph.java
```diff
@@ -205,6 +205,10 @@ boolean checkStorageGroup(String path) {
     throw new PathErrorException(TIME_SERIES_INCORRECT + rootName);
   }
 
+  List<MNode> getAllStorageGroups() {
```
qiaojialin:
```suggestion
  List<MNode> getAllStorageGroupNodes() {
```

server/src/main/java/org/apache/iotdb/db/metadata/MGraph.java
```diff
@@ -259,8 +263,8 @@ public Metadata getMetadata() throws PathErrorException {
     return new Metadata(deviceIdMap);
   }
 
-  Set<String> getAllStorageGroup() {
-    return mtree.getAllStorageGroup();
+  List<String> getAllStorageGroupList() {
```
qiaojialin:
```suggestion
  List<String> getAllStorageGroupNames() {
```

server/src/main/java/org/apache/iotdb/db/metadata/MGraph.java
```diff
@@ -395,18 +390,18 @@ public String toString() {
   /**
    * combine multiple metadata in string format
    */
-  static String combineMetadataInStrings(String[] metadatas) {
-    return MTree.combineMetadataInStrings(metadatas);
+  static String combineMetadataInStrings(String[] metadataArray) {
+    return MTree.combineMetadataInStrings(metadataArray);
```
qiaojialin: is this method a devil?

server/src/main/java/org/apache/iotdb/db/query/context/QueryContext.java
```diff
@@ -87,4 +90,15 @@ public long getJobId() {
...
 
+  public long getQueryTimeLowerBound() {
+    return queryTimeLowerBound;
+  }
+
+  public void setQueryTimeLowerBound(long queryTimeLowerBound) {
+    this.queryTimeLowerBound = queryTimeLowerBound;
+  }
+
+  public boolean chunkNotSatisfy(ChunkMetaData chunkMetaData) {
```
qiaojialin: do we need a pageNotSatisfy()?

docs/Documentation-CHN/UserGuide/4-Operation Manual/7-IoTDB Query Language.md
```diff
@@ -475,6 +475,43 @@ SELECT SUM(Path) (COMMA SUM(Path))* FROM <FromClause> [WHERE <WhereClause>]?
...
+### TTL
+IoTDB支持对存储组级别设置数据存活时间（TTL），这使得IoTDB可以定期、自动地删除一定时间之前的数据。合理使用TTL
+可以帮助您控制IoTDB占用的总磁盘空间以避免出现磁盘写满等异常。并且，随着文件数量的增多，查询性能往往随之下降,
+内存占用也会有所提高。及时地删除一些较老的文件有助于使查询性能维持在一个较高的水平和减少内存资源的占用。
+IoTDB中的TTL操作通可以由以下的语句进行实现：
+
+* 设置 TTL
+```
+SET TTL TO StorageGroupName TTLTime
+Eg. SET TTL TO root.group1 3600000
```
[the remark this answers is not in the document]
jt2594838: Currently not, maybe leave as a future feature.

docs/Documentation-CHN/UserGuide/4-Operation Manual/7-IoTDB Query Language.md
```diff
@@ -475,6 +475,43 @@ SELECT SUM(Path) (COMMA SUM(Path))* FROM <FromClause> [WHERE <WhereClause>]?
...
+* 取消 TTL
+```
+UNSET TTL TO StorageGroupName
+Eg. UNSET TTL TO root.group1
+这个例子展示了如何取消存储组root.group1的TTL，这将使得该存储组接受任意时刻的数据。
+```
+
+* 显示 TTL
+```
+SHOW ALL TTL
```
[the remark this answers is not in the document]
jt2594838: Changed.

server/src/assembly/resources/conf/iotdb-engine.properties
```diff
@@ -171,6 +171,11 @@ concurrent_flush_thread=0
 # (i.e., whether use ChunkBufferPool), value true, false
 chunk_buffer_pool_enable=false
 
+# default TTL for storage groups that are not set TTL by statements, in ms
+# Notice: if this property is changed, previous created storage group which are not set TTL will
+# also be affected
+# default_ttl=36000000
```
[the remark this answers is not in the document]
jt2594838: This is not the default value and I have added more comments to explain this. 0 and even negative values are supported as TTL values. As we also support negative timestamps, I guess there is no need to put such a constraint on this.

server/src/main/java/org/apache/iotdb/db/engine/StorageEngine.java
```diff
@@ -103,11 +112,32 @@ private StorageEngine() {
   @Override
   public void start() {
     // nothing to be done
```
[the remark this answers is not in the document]
jt2594838: Removed.

server/src/main/java/org/apache/iotdb/db/engine/merge/manage/MergeResource.java
```diff
@@ -64,12 +64,28 @@
   private Map<String, MeasurementSchema> measurementSchemaMap = new HashMap<>();
   private Map<MeasurementSchema, IChunkWriter> chunkWriterCache = new ConcurrentHashMap<>();
 
+  private long fileTimeBound = Long.MAX_VALUE;
```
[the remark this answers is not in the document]
jt2594838: Changed.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -178,6 +182,12 @@
   private LinkedList<String> lruForSensorUsedInQuery = new LinkedList<>();
   private static final int MAX_CACHE_SENSORS = 5000;
 
+  /**
+   * when the data in a storage group is older than dataTTL, it is considered invalid and will
```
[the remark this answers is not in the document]
jt2594838: That is the difference between "You are older than 5 years" and "You were born before 2014", I guess there is nothing wrong.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -178,6 +182,12 @@
...
+   * be eventually removed.
+   */
+  private long dataTTL = Long.MAX_VALUE;
```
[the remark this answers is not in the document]
jt2594838: I can't your point. There is only `dataTTL` in SGProcessor and `timeLowerBound` in TsFileResource is derived from it. They are playing different roles and I cannot only use one of them.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -387,6 +408,15 @@ public boolean insert(InsertPlan insertPlan) throws QueryProcessorException {
     }
   }
 
+  /**
+   *
+   * @param time
+   * @return whether the given time falls in ttl
+   */
+  private boolean checkTTL(long time) {
+    return dataTTL == Long.MAX_VALUE || (System.currentTimeMillis() - time) <= dataTTL;
```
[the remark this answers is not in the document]
jt2594838: There is no merit in doing so, persuade me, please.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -387,6 +408,15 @@ public boolean insert(InsertPlan insertPlan) throws QueryProcessorException {
     }
   }
 
+  /**
+   *
+   * @param time
+   * @return whether the given time falls in ttl
+   */
+  private boolean checkTTL(long time) {
```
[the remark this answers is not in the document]
jt2594838: Even though it only serves for insertion now, this method is not logically related to insert, so I would prefer its current name.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -582,6 +612,65 @@ public void syncDeleteDataFiles() {
...
 
+  /**
+   * Iterate each TsFile and try to lock and remove those out of TTL.
+   */
+  public synchronized void checkFilesTTL() {
+    if (dataTTL == Long.MAX_VALUE) {
+      logger.debug("{}: TTL not set, ignore the check", storageGroupName);
+      return;
+    }
+    long timeBound = System.currentTimeMillis() - dataTTL;
```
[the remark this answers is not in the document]
jt2594838: Changed.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -582,6 +612,65 @@ public void syncDeleteDataFiles() {
...
+    try {
+      // prevent new merges and queries from choosing this file
+      resource.setDeleted(true);
+      // the file may be chosen for merge after the last check and before writeLock()
+      // double check to ensure the file is not used by a merge
+      if (resource.isMerging()) {
+        return;
+      }
+      // ensure that the file is not used by any queries
+      if (resource.getMergeQueryLock().writeLock().tryLock()) {
```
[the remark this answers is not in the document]
jt2594838: `Lock()` will block this thread and make other deletable files unable to be deleted.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -582,6 +612,65 @@ public void syncDeleteDataFiles() {
...
+        try {
+          // physical removal
+          resource.remove();
```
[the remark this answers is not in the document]
jt2594838: Excuse me?

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -582,6 +612,65 @@ public void syncDeleteDataFiles() {
...
+          logger.info("Removed a file {} by ttl ({}ms)", resource.getFile().getPath(), dataTTL);
```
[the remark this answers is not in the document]
jt2594838: Added.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
jt2594838: Changed.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
jt2594838: No, it is to notice the TTLChecker that the file should not be deleted.

server/src/main/java/org/apache/iotdb/db/metadata/MGraph.java
jt2594838: Changed.

server/src/main/java/org/apache/iotdb/db/metadata/MGraph.java
jt2594838: What do you mean?

server/src/main/java/org/apache/iotdb/db/metadata/MGraph.java
jt2594838: Changed.

server/src/main/java/org/apache/iotdb/db/query/context/QueryContext.java
jt2594838: That is done by the time filter and have nothing to do with here.

server/src/assembly/resources/conf/iotdb-engine.properties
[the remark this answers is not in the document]
qiaojialin: I think negative TTL and negative timestamps are not the same thing...

server/src/main/java/org/apache/iotdb/db/metadata/MGraph.java
qiaojialin: 这个方法是魔鬼吗，这方法干嘛的，连个例子都没有

jt2594838: 你问我我也不知道，我只是当了一下语法警察。

PR #430, opened 2019-09-26
[IOTDB-193]Create schema automatically

SolomonAnn:
I have completed the issue described in the [[IOTDB-193]](https://issues.apache.org/jira/browse/IOTDB-193). For details, when users allow iotdb to create schema automatically:
- if users execute operations of creating timeseries while they haven't set the storage group, the system will set the storage group automatically.
- if users execute insertion operations while they have set the storage group but haven't created related timeseries, the system will create timeseries automatically.
- if users execute insertion operations directly, the system will set the storage group and create timeseries automatically.


server/src/main/java/org/apache/iotdb/db/engine/StorageEngine.java
```diff
@@ -31,11 +31,7 @@
 import org.apache.iotdb.db.engine.querycontext.QueryDataSource;
 import org.apache.iotdb.db.engine.storagegroup.StorageGroupProcessor;
 import org.apache.iotdb.db.engine.storagegroup.TsFileResource;
-import org.apache.iotdb.db.exception.MetadataErrorException;
-import org.apache.iotdb.db.exception.PathErrorException;
-import org.apache.iotdb.db.exception.ProcessorException;
-import org.apache.iotdb.db.exception.StorageEngineException;
-import org.apache.iotdb.db.exception.StorageEngineFailureException;
+import org.apache.iotdb.db.exception.*;
```
liuruiyiyang: exception.* seems not a good way of import

server/src/main/java/org/apache/iotdb/db/qp/executor/QueryProcessExecutor.java
```diff
@@ -661,4 +717,111 @@ private AuthDataSet executeListUserPrivileges(IAuthorizer authorizer, String use
...
+        predictedDataType, defaultEncoding,
+        CompressionType.valueOf(TSFileDescriptor.getInstance().getConfig().getCompressor()));
+    operateMetadata(metadataPlan);
+  }
+
+  /**
+   * Get predicted DataType of the given value
+   */
+  private TSDataType getPredictedDataType(Object value) {
+    if (value instanceof Boolean) {
```
liuruiyiyang: Under what circumstances will the value object be Boolean or Number? Is there any test code to cover this line?

server/src/main/java/org/apache/iotdb/db/engine/StorageEngine.java
SolomonAnn: I've fixed it.

server/src/assembly/resources/conf/iotdb-engine.properties
```diff
@@ -292,3 +292,31 @@ watermark_module_opened=false
...
+
+####################
+### Configurations for creating schema automatically
+####################
+
+# Is automatically creating schema enable
+enable_auto_create_schema=true
+
+# Storage group level when creating schema automatically is enable
+# The default is 2
```
qiaojialin: add an example, such as root.sg1.d1.s1. when this level is 1, we set root.sg1 as a storage group

server/src/assembly/resources/conf/iotdb-engine.properties
```diff
@@ -292,3 +292,31 @@ watermark_module_opened=false
...
+auto_storage_group_level=2
```
qiaojialin: should root be the 0 level?

server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
```diff
@@ -275,6 +275,36 @@
...
+
+  /**
+   * Storage group level when creating schema automatically is enable
+   */
+  private int autoStorageGroupLevel = 2;
+
+  /**
+   * Boolean encoding when creating schema automatically is enable
+   */
+  private String autoBooleanEncoding = "RLE";
```
qiaojialin: it's better to use TSEncoding

server/src/assembly/resources/conf/iotdb-engine.properties
```diff
@@ -292,3 +292,31 @@ watermark_module_opened=false
...
+# The default is RLE
+auto_long_encoding=RLE
+
+# Double encoding when creating schema automatically is enable
+# The default is GORILLA
+auto_double_encoding=GORILLA
+
+# String encoding when creating schema automatically is enable
+# The default is PLAIN
+auto_string_encoding=PLAIN
```
jt2594838: I wish you could rename these to `default_xxx_encoding` and one day we will not need to specify an encoding in  "create timeseries" statements.

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -1233,6 +1254,22 @@ public String getMetadataInString() {
...
+   */
+  public String getStorageGroupNameByAutoLevel(String fullPath, int level)
+      throws PathErrorException {
+    String[] nodeNames = fullPath.trim().split(DOUB_SEPARATOR);
+    String storageGroupName = nodeNames[0];
+    if (nodeNames.length < level || !storageGroupName.equals(ROOT_NAME)) {
+      throw new PathErrorException(String.format("Timeseries %s is not right.", fullPath));
+    }
+    for (int i = 1; i < level; i++) {
+      storageGroupName += '.' + nodeNames[i];
```
jt2594838: Better to replace '.' with Constant.PATH_SEPARATOR

server/src/main/java/org/apache/iotdb/db/qp/executor/QueryProcessExecutor.java
```diff
@@ -211,15 +211,42 @@ public boolean insert(InsertPlan insertPlan) throws ProcessorException {
...
+            addPathListToMTree(deviceId, measurementList, values);
+          } catch (ProcessorException | PathErrorException e2) {
+            throw new ProcessorException(e2);
+          }
+        }
+      } catch (PathErrorException e2) {
+        if (!conf.isAutoCreateSchemaEnable()) {
+          throw new ProcessorException(e2);
+        } else {
+          addPathListToMTree(deviceId, measurementList, values);
```
jt2594838: I think you can add a flag here and if you have called `addPathListToMTree()`, you can skip the next check since you have already created the time series, there is no chance they do not exist.

server/src/main/java/org/apache/iotdb/db/qp/executor/QueryProcessExecutor.java
```diff
@@ -244,18 +271,47 @@ public boolean insert(InsertPlan insertPlan) throws ProcessorException {
     try {
       String[] measurementList = batchInsertPlan.getMeasurements();
       String deviceId = batchInsertPlan.getDeviceId();
-      MNode node = mManager.getNodeByDeviceIdFromCache(deviceId);
+      Object[] values = collectFirstElements(batchInsertPlan.getColumns());
```
jt2594838: I think you should delay the collection because not all insertions need schema creation.

server/src/main/java/org/apache/iotdb/db/qp/executor/QueryProcessExecutor.java
```diff
@@ -244,18 +271,47 @@ public boolean insert(InsertPlan insertPlan) throws ProcessorException {
...
+      IoTDBConfig conf = IoTDBDescriptor.getInstance().getConfig();
+
+      try {
+        mManager.checkDeviceId(deviceId);
```
jt2594838: This method is called very frequently and it is not a light operation. I am afraid that if you do not add a cache mechanism for it, the performance will downgrade.

server/src/main/java/org/apache/iotdb/db/qp/executor/QueryProcessExecutor.java
```diff
@@ -661,4 +717,111 @@ private AuthDataSet executeListUserPrivileges(IAuthorizer authorizer, String use
...
+
+  /**
+   * Set storage group when creating schema automatically is enable
+   */
+  private void setStorageGroupToMTreeByAutoLevel(String deviceId) throws ProcessorException, PathErrorException {
+    IoTDBConfig conf = IoTDBDescriptor.getInstance().getConfig();
+    MetadataPlan metadataPlan = new MetadataPlan(
+        MetadataOperator.NamespaceType.SET_STORAGE_GROUP,
+        new Path(mManager.getStorageGroupNameByAutoLevel(deviceId, conf.getAutoStorageGroupLevel())));
+    operateMetadata(metadataPlan);
```
jt2594838: I think you can just use the interface in MManager and avoid creating a plan.

server/src/main/java/org/apache/iotdb/db/qp/executor/QueryProcessExecutor.java
```diff
@@ -661,4 +717,111 @@ private AuthDataSet executeListUserPrivileges(IAuthorizer authorizer, String use
...
+        return TSDataType.INT64;
+      } else {
+        return TSDataType.DOUBLE;
+      }
+    } else {
+      return TSDataType.TEXT;
+    }
+  }
+
+  private boolean isNumber(String s) {
```
jt2594838: This function seems a bit complicated. An easier way would be:
```
try {
    Double.parseDouble(s);
} catch(NumberFormatException e) {
    return false;
}
return true;
```
Or maybe you can find a regexp of numbers on the web and use if.
But these might be slower and I would appreciate it if you can compare the three methods.

server/src/main/java/org/apache/iotdb/db/qp/executor/QueryProcessExecutor.java
jt2594838: I think this is possible when using a batched insert.

server/src/test/java/org/apache/iotdb/db/metadata/MManagerBasicTest.java
```diff
@@ -360,4 +363,26 @@ public void testMaximalSeriesNumberAmongStorageGroup() throws MetadataErrorExcep
...
+    try {
+      Assert.assertEquals("root.laptop", manager.getStorageGroupNameByAutoLevel("root.laptop.d1.s1", level));
+    } catch(PathErrorException e) {
+      e.printStackTrace();
+      fail(e.getMessage());
+    }
+    try {
+      manager.getStorageGroupNameByAutoLevel("root1.laptop.d1.s1", level);
+    } catch(PathErrorException e) {
+      Assert.assertEquals("Timeseries root1.laptop.d1.s1 is not right.", e.getMessage());
```
jt2594838: This is not a good way to test exceptions because the tests still pass if there is just no exception, which is also incorrect.
I would usually write code like:
```
boolean caught = false;
try {
  doSomething();
} catch (WantedException e) {
  // verify message
  caught = true;
}
assertTrue(caught);
```
I know this is not the best way, neither, but it can assure the exception is risen and caught. If you find a better way, please inform me.

server/src/main/java/org/apache/iotdb/db/qp/executor/QueryProcessExecutor.java
```diff
@@ -661,4 +717,111 @@ private AuthDataSet executeListUserPrivileges(IAuthorizer authorizer, String use
...
+  private void addPathToMTree(String deviceId, String measurementId, Object value)
+          throws ProcessorException {
+    TSDataType predictedDataType = getPredictedDataType(value);
+    TSEncoding defaultEncoding = getDefaultEncoding(predictedDataType);
+    MetadataPlan metadataPlan = new MetadataPlan(
+        MetadataOperator.NamespaceType.ADD_PATH,
+        new Path(deviceId + "." + measurementId),
+        predictedDataType, defaultEncoding,
+        CompressionType.valueOf(TSFileDescriptor.getInstance().getConfig().getCompressor()));
+    operateMetadata(metadataPlan);
```
fanhualta: You can use interfaces in `MManager` rather than creating a `MetadataPlan`.

server/src/main/java/org/apache/iotdb/db/utils/DeviceIdCache.java
```diff
@@ -0,0 +1,53 @@
...
+      removeFirstObject();
+    }
+  }
+
+  private void removeFirstObject() throws CacheException {
+    if (cache.size() == 0) {
+      return;
+    }
+    String deviceId = cache.iterator().next();
+    cache.remove(deviceId);
```
fanhualta: I think it's a little bit complicated. How about
```suggestion
    if (cache.size() != 0) {
      cache.remove(0);
    }
```
What's more, I think it's better not to split `RemoveObjectIfCacheIsFull` into two methods.

server/src/main/java/org/apache/iotdb/db/utils/DeviceIdCache.java
```diff
@@ -0,0 +1,53 @@
+package org.apache.iotdb.db.utils;
+
+import org.apache.iotdb.db.exception.PathErrorException;
+import org.apache.iotdb.db.exception.StorageGroupException;
+import org.apache.iotdb.tsfile.exception.cache.CacheException;
+
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+public abstract class DeviceIdCache {
```
jt2594838: The number of devices in a system is rather limited and holding all in memory is not a burden, so I do not suggest making it an LRU cache for meaningless overhead. 
Nevertheless, I seriously doubt the performance of using a List, as the `contains()` method can be extremely slow even when there are only thousands of devices (not to mention you are using a COWList). I strongly recommend using a HashSet.

server/src/test/java/org/apache/iotdb/db/integration/IoTDBAutoCreateSchemaIT.java
```diff
@@ -0,0 +1,155 @@
...
+    daemon = IoTDB.getInstance();
+    daemon.active();
+    EnvironmentUtils.envSetUp();
+    conf.setAutoCreateSchemaEnabled(true);
+  }
+
+  @After
+  public void tearDown() throws Exception {
+    daemon.stop();
+    EnvironmentUtils.cleanEnv();
```
jt2594838: Recover in `tearDown()` what configs you have modified in `setUp()`.

server/src/test/java/org/apache/iotdb/db/integration/IoTDBAutoCreateSchemaIT.java
```diff
@@ -0,0 +1,155 @@
...
+   * insert data when storage group hasn't been set and timeseries hasn't been created
+   */
+  @Test
+  public void insertTest2() throws ClassNotFoundException {
+    String[] sqls = {"INSERT INTO root.sg0.d1(timestamp,s3) values(1,\"abc\")",
+        "INSERT INTO root.sg0.d2(timestamp,s2) values(1,123.123)",
+        "INSERT INTO root.sg0.d3(timestamp,s4) values(1,123456)",
+    };
+    executeSQL(sqls);
+  }
```
jt2594838: Add a test where you insert a non-exist time series with different data types.

server/src/test/java/org/apache/iotdb/db/utils/TypeInferenceUtilsTest.java
```diff
@@ -0,0 +1,50 @@
...
+import static org.junit.Assert.assertEquals;
+
+public class TypeInferenceUtilsTest {
+
+  @Test
+  public void isNumberTest() {
+    String[] values = {"123", "123.123", "-123.123", "+123.123", ".123", String.valueOf(Integer.MAX_VALUE),
+        String.valueOf(Integer.MIN_VALUE), "abc", "123.", "123a", ".", "123.1.23"};
+    boolean[] results = {true, true, true, true, true, true,
+        true, false, false, false, false, false};
```
jt2594838: Add coverage of scientific notation.

server/src/main/java/org/apache/iotdb/db/utils/DeviceIdCache.java
SolomonAnn: I'm confused that whether I should use `mNodeCache `directly because its key is just `deviceId`. However, if I choose to use `mNodeCache`, I will have trouble with handling exceptions since I need `PathErrorException `and `StorageGroupException` to distinguish two different cases while `mNodeCache` throws `CacheException` in any case. Do you have any suggestions?

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -556,6 +556,8 @@ private String deletePathFromMTree(String path)
   public void setStorageGroupToMTree(String path) throws MetadataErrorException {
     lock.writeLock().lock();
     try {
+      checkAndGetDataTypeCache.clear();
+      mNodeCache.clear();
```
jt2594838: Is it necessary to clear the cache when you are doing an insertion?

PR #431, opened 2019-09-26
display cache hit rate through jconsole

little-emotion:
1. View cache hit rate through debug level log and MXBean.
2. Log configuration file for iotdb are loaded dynamically every 60s


LeiRui:
excellent!

server/src/assembly/resources/conf/logback.xml
```diff
@@ -19,7 +19,7 @@
     under the License.
 
 -->
-<configuration>
+<configuration scan="true" scanPeriod="100 seconds">
```
jixuan1989: 60s is fine. It is a good feature, you'd better point out this change in the commitlog.

server/src/main/java/org/apache/iotdb/db/service/ServiceType.java
```diff
@@ -35,6 +35,7 @@
   PERFORMANCE_STATISTIC_SERVICE("PERFORMANCE_STATISTIC_SERVICE","PERFORMANCE_STATISTIC_SERVICE"),
   MANAGE_DYNAMIC_PARAMETERS_SERVICE("Manage Dynamic Parameters", "Manage Dynamic Parameters"),
   TVLIST_ALLOCATOR_SERVICE("TVList Allocator", ""),
+  CACHE_HIT_RATE_DISPLAY_SERVICE("CACHE_HIT_RATE_DISPLAY_SERVICE","Cache Hit Rate"),
```
jixuan1989: I'd like use the following method:
```
FLUSH_SERVICE("Flush ServerService",
      generateJmxName("org.apache.iotdb.db.engine.pool", "Flush Manager"));
```
Then you do not need to maintain the mbeanName in CacheHitRateMonitor.

server/src/assembly/resources/conf/logback.xml
little-emotion: ok

server/src/main/java/org/apache/iotdb/db/service/ServiceType.java
little-emotion: changed

server/src/main/java/org/apache/iotdb/db/engine/cache/DeviceMetaDataCache.java
```diff
@@ -153,6 +147,29 @@ public static DeviceMetaDataCache getInstance() {
...
+    if (isHit) {
+      logger.debug(
+          "[ChunkMetaData cache hit] The number of requests for cache is {}, hit rate is {}.",
+          cacheRequestNum.get(), cacheHitNum.get() * 1.0 / cacheRequestNum.get());
+    } else {
+      logger.debug(
+          "[ChunkMetaData cache didn't hit] The number of requests for cache is {}, hit rate is {}.",
+          cacheRequestNum.get(), cacheHitNum.get() * 1.0 / cacheRequestNum.get());
+    }
+  }
```
fanhualta: These codes could be more concise.
```suggestion
      logger.debug(
          "[ChunkMetaData cache {}hit] The number of requests for cache is {}, hit rate is {}.",
          isHit ? "":"didn't ", cacheRequestNum.get(), cacheHitNum.get() * 1.0 / cacheRequestNum.get());
```

server/src/assembly/resources/conf/logback.xml
jt2594838: Do not forget to update the related descriptions in docs.

PR #434, opened 2019-09-27
[IOTDB-244]fix bug when querying with duplicated columns

LeiRui:
JIRA https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-244?filter=allopenissues

In the original code, the client side IoTDBQueryResultSet assumes that the server returns deduplicated query results. However, in the original code, the sever returns query results without the step of deduplication. That's the cause of the bug IOTDB-244.
So, I fix this bug by adding the deduplication step in the server side and uses the same deduplication or map strategy as the client.


server/src/main/java/org/apache/iotdb/db/qp/executor/AbstractQueryProcessExecutor.java
```diff
@@ -63,29 +65,84 @@ protected abstract QueryDataSet processAuthorQuery(AuthorPlan plan, QueryContext
...
-            .setExpression(queryPlan.getExpression());
+
+    deduplicate(queryPlan.getPaths(), deduplicatedPaths);
+    QueryExpression queryExpression = QueryExpression.create().setSelectSeries(deduplicatedPaths)
+        .setExpression(queryPlan.getExpression());
     return queryRouter.query(queryExpression, context);
   }
 
+  /**
+   * Note that the deduplication strategy must be consistent with that of IoTDBQueryResultSet.
```
jt2594838: It is interesting that I did find the modifications of IoTDBQueryResultSet in this PR.

LeiRui: Yes, `IoTDBQueryResultSet` is not changed because I fix this bug by modifying the server side.

LeiRui: In the original code, the client side `IoTDBQueryResultSet` assumes that the server returns deduplicated query results. However, in the original code, the sever returns query results without the step of deduplication. That's the cause of the bug IOTDB-244.
So, I fix this bug by adding the deduplication step in the server side and uses the same deduplication or map strategy.

jt2594838: Got it.
I would appreciate it if you can describe the cause of the bug and your solution briefly at the beginning of the PR. Or you may provide a Jira link and explain it there.

LeiRui:
In the original code, the client side IoTDBQueryResultSet assumes that the server returns deduplicated query results. However, in the original code, the sever returns query results without the step of deduplication. That's the cause of the bug IOTDB-244.
So, I fix this bug by adding the deduplication step in the server side and uses the same deduplication or map strategy as the client.

PR #435, opened 2019-09-27
[IOTDB-174] Fix querying timeseries interface cannot make a query by the specified path prefix

jack870131:
Corresponding to JIRA issue: https://issues.apache.org/jira/browse/IOTDB-174

1. Fix the bug which the querying timeseries number interface cannot make a query by the specified path prefix.
2. Fix the documentation corresponding to the query operation (add the device in path).

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
@@ -31,6 +31,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.sql.SQLException;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import org.apache.iotdb.db.conf.IoTDBConfig;
 import org.apache.iotdb.db.conf.IoTDBConstant;
@@ -849,7 +850,7 @@ public class MManager {
    *
    * @return A HashSet instance which stores all devices info
    */
-  public Set<String> getAllDevices() throws PathErrorException {
+  public Set<String> getAllDevices() throws PathErrorException, SQLException {
 
     lock.readLock().lock();
     try {
@@ -864,11 +865,11 @@ public class MManager {
    *
    * @return A List instance which stores all node at given level
    */
-  public List<String> getNodesList(int nodeLevel) {
+  public List<String> getNodesList(String prefixPath, int nodeLevel) throws SQLException {
 
     lock.readLock().lock();
     try {
-      return mgraph.getNodesList(nodeLevel);
+      return mgraph.getNodesList(prefixPath, nodeLevel);
     } finally {
       lock.readLock().unlock();
     }
--- a/server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
@@ -22,15 +22,12 @@ import com.alibaba.fastjson.JSON;
 import com.alibaba.fastjson.JSONObject;
 import com.alibaba.fastjson.serializer.SerializerFeature;
 import java.io.Serializable;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.sql.SQLException;
+import java.util.*;
+
 import org.apache.iotdb.db.exception.PathErrorException;
 import org.apache.iotdb.db.exception.StorageGroupException;
+import org.apache.iotdb.db.qp.constant.SQLConstant;
 import org.apache.iotdb.tsfile.common.conf.TSFileDescriptor;
 import org.apache.iotdb.tsfile.file.metadata.enums.CompressionType;
 import org.apache.iotdb.tsfile.file.metadata.enums.TSDataType;
@@ -797,7 +794,29 @@ public class MTree implements Serializable {
    * @return a list contains all distinct devices
    */
   Set<String> getAllDevices() {
-    return new HashSet<>(getNodesList(3));
+    HashSet<String> devices = new HashSet<>();
+    MNode node;
+    if ((node = getRoot()) != null) {
+      findDevices(node, SQLConstant.ROOT, devices);
+    }
+    return new LinkedHashSet<>(devices);
+  }
+
+  private void findDevices(MNode node, String path, HashSet<String> res) {
+    if (node == null) {
+      return;
+    }
+    if (node.isLeaf()) {
+      res.add(path);
+      return;
+    }
+    for (MNode child : node.getChildren().values()) {
+      if (child.isLeaf()) {
+        res.add(path);
+      } else {
+        findDevices(child, path + "." + child.toString(), res);
+      }
+    }
   }
 
   /**
@@ -805,11 +824,23 @@ public class MTree implements Serializable {
    *
    * @return a list contains all nodes at the given level
    */
-  List<String> getNodesList(int nodeLevel) {
+  List<String> getNodesList(String schemaPattern, int nodeLevel) throws SQLException {
     List<String> res = new ArrayList<>();
-    MNode rootNode;
-    if ((rootNode = getRoot()) != null) {
-      findNodes(rootNode, "root", res, nodeLevel);
+    String[] nodes = schemaPattern.split("\\.");
+    MNode node;
+    if ((node = getRoot()) != null) {
+      if (nodes[0].equals("root")) {
+        for (int i = 1; i < nodes.length; i++) {
+          if (node.getChild(nodes[i]) != null) {
+            node = node.getChild(nodes[i]);
+          } else {
+            throw new SQLException(nodes[i - 1] + " does not have the child node " + nodes[i]);
+          }
+        }
+        findNodes(node, schemaPattern, res, nodeLevel - (nodes.length - 1));
+      } else {
+        throw new SQLException("Incorrect root node " + nodes[0] + " selected");
+      }
     }
     return res;
   }
@@ -818,7 +849,7 @@ public class MTree implements Serializable {
     if (node == null) {
       return;
     }
-    if (targetLevel == 1) {
+    if (targetLevel == 0) {
       res.add(path);
       return;
     }
--- a/docs/Documentation-CHN/UserGuide/4-Operation Manual/7-IoTDB Query Language.md
+++ b/docs/Documentation-CHN/UserGuide/4-Operation Manual/7-IoTDB Query Language.md
@@ -88,6 +88,38 @@ Eg: IoTDB > SHOW STORAGE GROUP
 Note: This statement can be used in IoTDB Client and JDBC.
 ```
 
+* 显示指定路径下时间序列数语句
+
+```
+COUNT TIMESERIES <Path>
+Eg: IoTDB > COUNT TIMESERIES root
+Eg: IoTDB > COUNT TIMESERIES root.ln
+Eg: IoTDB > COUNT TIMESERIES root.ln.*.*.status
+Eg: IoTDB > COUNT TIMESERIES root.ln.wf01.wt01.status
+Note: The path can be prefix path, star path or timeseries path.
+Note: This statement can be used in IoTDB Client and JDBC.
+```
+
+```
+COUNT TIMESERIES <Path> GROUP BY LEVEL=<INTEGER>
+Eg: IoTDB > COUNT TIMESERIES root GROUP BY LEVEL=1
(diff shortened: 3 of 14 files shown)


server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
```diff
@@ -783,19 +783,23 @@ private void findStorageGroup(MNode node, String path, HashSet<String> res) {
...
+  List<String> getNodesList(String schemaPattern, int nodeLevel) {
     List<String> res = new ArrayList<>();
-    MNode rootNode;
-    if ((rootNode = getRoot()) != null) {
-      findNodes(rootNode, "root", res, nodeLevel);
+    String[] nodes = schemaPattern.split("\\.");
+    MNode node;
+    if ((node = getRoot()) != null) {
+      for (int i = 1; i < nodes.length; i++) {
+        node = node.getChild(nodes[i]);
```
jt2594838: What if the node does not have such a child.

jack870131:
I have added the SQLException to remind users if they enter the wrong child node.

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
```diff
@@ -782,20 +783,28 @@ private void findStorageGroup(MNode node, String path, HashSet<String> res) {
    *
    * @return a list contains all distinct devices
    */
-  Set<String> getAllDevices() {
-    return new HashSet<>(getNodesList(3));
+  Set<String> getAllDevices() throws SQLException {
+    return new HashSet<>(getNodesList("root",2));
```
LeiRui: `device` is not necessarily at level 2.
`device` is at the secod to last level. 
For example, `CREATE TIMESERIES root.demo.m2.m2.s0 WITH DATATYPE=INT32,ENCODING=RLE` and the device should be `root.demo.m2.m2`.

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
```diff
@@ -782,20 +783,28 @@ private void findStorageGroup(MNode node, String path, HashSet<String> res) {
...
-  List<String> getNodesList(int nodeLevel) {
+  List<String> getNodesList(String schemaPattern, int nodeLevel) throws SQLException {
     List<String> res = new ArrayList<>();
-    MNode rootNode;
-    if ((rootNode = getRoot()) != null) {
-      findNodes(rootNode, "root", res, nodeLevel);
+    String[] nodes = schemaPattern.split("\\.");
+    MNode node;
+    if ((node = getRoot()) != null) {
+      for (int i = 1; i < nodes.length; i++) {
```
LeiRui: The start of the schemaPattern is not checked to be `root`. 
I find that `count nodes xx.demo level=2` still returns the count result as normal, which is not right.

LeiRui:
One more suggestion: the header of this pr is a little bit general. Make it more specific would be better.

jack870131:
Please wait a moment for this PR as my computer was crashed these days and was sent for the repair these days.

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
jack870131: @LeiRui Hi,I have no idea about how to implement this as the function of showing all devices was not developed by me. I just did some minor changes to ensure that it won't report the error.

Could you please provide me some advices or discuss with me on how to get this function done? Maybe this should be on another PR. What do you guys think.

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
jack870131: Done.

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
jack870131: @LeiRui Hi, actually I have no idea now about how to implement this interface as it was not developed by me before. What I done is just making sure that the program does not report the error.

However, I would like to discuss with you guys about how to make changes of this function,  could you please give me some advices? I think this should create another PR for this.

jack870131:
Hi, I have no idea now about how to implement the interface which could show all device here as it was not developed by me before. What I done is just making sure that the program does not report the error.

However, I would like to discuss with you guys about how to make changes of this function, could you please give me some advices? I think this should create another PR for this or reopen PR #421 to correct this issue.

jack870131:
> Please modify the SQL Reference document to add `COUNT` clause.

I have added the new schema statement in it. Please have a look.

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
samperson1997: Hi, terribly sorry for the late reply, and I’m responsible for this issue. Do you think it is necessary to add a new param `String schemaPattern` for method `Set<String> getAllDevices()`  as well?

samperson1997: Or maybe I will fix this bug after your PR is merged so that it won't result in too many conflicts since your PR is currently not reporting any error.

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -830,11 +831,11 @@ public Metadata getMetadata() throws PathErrorException {
    *
    * @return A List instance which stores all node at given level
    */
-  public List<String> getNodesList(int nodeLevel) {
+  public List<String> getNodesList(String schemaPattern, int nodeLevel) throws SQLException {
```
qiaojialin:
```suggestion
  public List<String> getNodesList(String prefixPath, int nodeLevel) throws SQLException {
```

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
```diff
@@ -781,20 +777,56 @@ private void findStorageGroup(MNode node, String path, HashSet<String> res) {
    *
    * @return a list contains all distinct devices
    */
-  Set<String> getAllDevices() {
-    return new HashSet<>(getNodesList(3));
+  Set<String> getAllDevices() throws SQLException {
+    List<String> res = new ArrayList<>();
```
qiaojialin:
```suggestion
    List<String> devices = new ArrayList<>();
```

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
```diff
@@ -781,20 +777,56 @@ private void findStorageGroup(MNode node, String path, HashSet<String> res) {
    *
    * @return a list contains all distinct devices
    */
-  Set<String> getAllDevices() {
-    return new HashSet<>(getNodesList(3));
+  Set<String> getAllDevices() throws SQLException {
```
qiaojialin:
```suggestion
  Set<String> getAllDevices() {
```

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
```diff
@@ -781,20 +777,56 @@ private void findStorageGroup(MNode node, String path, HashSet<String> res) {
...
+    List<String> res = new ArrayList<>();
+    MNode node;
+    if ((node = getRoot()) != null) {
+      findDevices(node, "root", res);
```
qiaojialin:
```suggestion
      findDevices(node, SQLConstant.ROOT, res);
```

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
```diff
@@ -781,20 +777,56 @@ private void findStorageGroup(MNode node, String path, HashSet<String> res) {
...
+
+  private void findDevices(MNode node, String path, List<String> res) {
+    if (node == null) {
+      return;
+    }
+    if (node.isLeaf()) {
+      res.add(path);
+      return;
+    }
+    if (node.hasChildren()) {
```
qiaojialin: You already check the node is null or is a leaf. Here, the node should have a child. This if statement is not needed.

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
```diff
@@ -781,20 +777,56 @@ private void findStorageGroup(MNode node, String path, HashSet<String> res) {
...
+      for (MNode child : node.getChildren().values()) {
+        if (child.isLeaf()) {
+          findDevices(child, path, res);
```
qiaojialin:
```suggestion
          res.add(path);
```

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
```diff
@@ -781,20 +777,56 @@ private void findStorageGroup(MNode node, String path, HashSet<String> res) {
...
+  Set<String> getAllDevices() throws SQLException {
+    List<String> res = new ArrayList<>();
+    MNode node;
+    if ((node = getRoot()) != null) {
+      findDevices(node, "root", res);
+    }
+    return new LinkedHashSet<>(res);
+  }
+
+  private void findDevices(MNode node, String path, List<String> res) {
```
qiaojialin: This function is very confused, what's the relationship between node and path?

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
```diff
@@ -781,20 +777,56 @@ private void findStorageGroup(MNode node, String path, HashSet<String> res) {
...
+    }
   }
 
   /**
    * Get all nodes at the given level in current Metadata Tree.
    *
    * @return a list contains all nodes at the given level
    */
-  List<String> getNodesList(int nodeLevel) {
+  List<String> getNodesList(String schemaPattern, int nodeLevel) throws SQLException {
```
qiaojialin: Is this function only for the count? If so, why not return an int?

jt2594838:
There are failing tests.

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
jack870131: @qiaojialin It might not have to declare two separate params here, which it could use node.getName() to get node's name. However, the function above (findStorageGroup) has also declared like this, so I'm not sure if I have to change this declaration.

jack870131: Or are you confusing with the naming of params? Is there any suggestion about how to change it to resolve the confusion?

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
jack870131: No, it is not only for the count. The function "COUNT NODE ... GROUP BY ..."  has also been shared with this method, as it needs paths (path string) in this list so that it could show the node number within the specified path under the specified level.

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
jack870131: Done.

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
jack870131: Done.

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
jack870131: Done.

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
jack870131: Done.

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
jack870131: Done.

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
jack870131: Done.

jack870131: Done.

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
qiaojialin: Here path != node.getName()  , path is the full path (e.g., root.d1.s1) of the node while node.getName only return the node name (e.g., s1). 

But now, after you change the res.add(path), these two parameters are consistent. No need to change.

BTW, if you find some existing codes which are confusing, do not hesitate to optimize them.

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
qiaojialin: ok

PR #439, opened 2019-09-29
Remove jdk constrain of jdk8 and 11

jixuan1989:
* But retain the constrain that jdk<1.8u40 will be forbidden.
* add Apache rat for .checkstyle file.


server/src/assembly/resources/conf/iotdb-env.sh
```diff
@@ -117,10 +117,6 @@ else
   IOTDB_JMX_OPTS="$IOTDB_JMX_OPTS -Xloggc:${IOTDB_HOME}/gc.log"
 fi
 
-if [ "$MAJOR_VERSION" -ne "8" ] && [ "$MAJOR_VERSION" -ne "11" ] ; then
```
MyXOF: why do you remove this check? It seems iotdb only supports jdk8,11 and 13? What if user run iotdb with jdk7 or 9?
Besides, iotdb-env.sh only works for mac and linux, how about windows platform?

jixuan1989: JDK 7 will be forbidden by
```
if [ "$JVM_VERSION" \< "1.8" ] && [ "$JVM_PATCH_VERSION" -lt 40 ] ; then
    echo "IoTDB requires Java 8u40 or later."
    exit 1;
fi
```
I do not try whether JDK9 is supported.

If we maintain the version check, we have to modify the code like :
```
if [ "$MAJOR_VERSION" -ne "8" ] && [ "$MAJOR_VERSION" -ne "11" ] && [ "$MAJOR_VERSION" -ne "13" ]  ; then
```
And now JDK 14 is coming, we need to modify once again:
```
if [ "$MAJOR_VERSION" -ne "8" ] && [ "$MAJOR_VERSION" -ne "11" ] && [ "$MAJOR_VERSION" -ne "13" ] && [ "$MAJOR_VERSION" -ne "14" ]   ; then
```

I will try JDK 9 and 10. If they are supported, I think we can remove this check.

jixuan1989: I have tried JDK9 and JDK10 using OpenJDK Docker image, and IoTDB supports both of them. Therefore, I think removing the code snippet is fine.

PR #440, opened 2019-09-29
reconstruct antlrv3 grammar to improve performance

Ring-k:


jixuan1989:
Hi, the improvement of the performance?

LeiRui:
Hi, could you summarize the core idea of this reconstruction in one or a few sentences?

server/src/main/antlr3/org/apache/iotdb/db/sql/parse/TqlParser.g
```diff
@@ -0,0 +1,726 @@
...
+    | updateStatement
+    | deleteStatement
+    | loadStatement
+    ;
+
+ddlStatement
+    : createTimeseries
+    | deleteTimeseries
+    | setStorageGroup
+    | deleteStorageGroup // todo to implement
```
jt2594838: It is implemented now.

server/src/main/antlr3/org/apache/iotdb/db/sql/parse/TqlParser.g
```diff
@@ -0,0 +1,726 @@
...
+    | K_LIST K_USER K_PRIVILEGES username = ID -> ^(TOK_LIST TOK_PRIVILEGES TOK_ALL ^(TOK_USER $username))
+    | K_LIST K_ROLE K_PRIVILEGES roleName = ID -> ^(TOK_LIST TOK_PRIVILEGES TOK_ALL ^(TOK_ROLE $roleName))
+    | K_LIST K_ALL K_ROLE K_OF K_USER username = ID -> ^(TOK_LIST TOK_ROLE TOK_ALL ^(TOK_USER $username))
+    | K_LIST K_ALL K_USER K_OF K_ROLE roleName = ID -> ^(TOK_LIST TOK_USER TOK_ALL ^(TOK_ROLE $roleName))
+    ;
+
+quitStatement
+    : K_QUIT
+    -> ^(TOK_QUIT)
+    ;
```
jt2594838: I wonder whether the "quit" statement has a chance to reach the server.

Ring-k: Actually, even if the statement is a quit statement, no logical plan will be generated. Do you suggest removing the grammar definition here?

server/src/main/antlr3/org/apache/iotdb/db/sql/parse/TqlParser.g
Ring-k: So far, the function is implemented. Thanks for your reminding.

Ring-k:
The core idea of this reconstruction is
1. The statement is now grouped as ddlStatement, dmlStatement, and administrationStatement, which is more consistent with the conventional sql design.
2. I reorganize the format of constant. In previous version, we have negetive integer, positive integer, usigned integer, etc. To aviod problems of longest match, we need to use "=>" operator. This operator is widely used in previous version. I simplified the grammer definition by just defining the integer and real numbers, regardless of the sign. No "=>" operator anymore. Althought, some acceptable constraints are introduced, like "+" is not allowed to identify a positive value. I think because of less checkings during parsing, the performance is improved, especially in parsing full-digit paths.

Ring-k:
The core idea of this reconstruction is
1. The statement is now grouped as ddlStatement, dmlStatement, and administrationStatement, which is more consistent with the conventional sql design.
2. I reorganized the format of constant. In previous version, we had negetive integer, positive integer, usigned integer, etc. To aviod problems of longest match, we need use "=>" operator. This operator was widely used in previous version. I simplified the grammer definitions by just defining the integer and real numbers, regardless of the sign. No "=>" operators anymore. Althought, some acceptable constraints are introduced, like "+" is not allowed to identify a positive value. I think because of less checkings during parsing, the performance is improved, especially in parsing full-digit paths.

LeiRui:
> "+" is not allowed to identify a positive value

What about updating docs where necessary?

server/src/main/java/org/apache/iotdb/db/qp/constant/TqlParserConstant.java
```diff
@@ -1,4 +1,4 @@
-/*
+/**
```
jixuan1989: use /*  rather than /**

server/src/main/java/org/apache/iotdb/db/qp/strategy/LogicalGenerator.java
```diff
@@ -1,14 +1,14 @@
-/*
+/**
```
jixuan1989: use /*
and seems no "<p>" marker

server/src/main/java/org/apache/iotdb/db/qp/strategy/LogicalGenerator.java
```diff
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p>
```
jixuan1989: same problem

server/src/main/java/org/apache/iotdb/db/qp/strategy/LogicalGenerator.java
```diff
@@ -829,42 +819,56 @@ private long parseTimeUnit(AstNode node) throws LogicalOperatorException {
...
-      seriesValue = parseTokenTime(rightKey);
-    } else if (rightKey.getType() == TSParser.TOK_FLOAT_COMB) {
-      seriesValue = parseTokens(rightKey);
-    } else {
-      seriesValue = rightKey.getText();
+      seriesValue = parseTokenTime(rightKey.getChild(0));
+    }else{
+      seriesValue = cascadeChildrenText(rightKey);
     }
+//    if (rightKey.getType() == TqlParser.TOK_PATH) {
```
jixuan1989: hm... useless codes?

server/src/test/java/org/apache/iotdb/db/qp/plan/LogicalPlanSmallTest.java
```diff
@@ -197,4 +197,21 @@ public void testLimit2()
...
+    try {
+      astTree = ParseGenerator.generateAST(sqlStr); // parse string to ASTTree
+    } catch (ParseException e) {
+      // e.printStackTrace();
+      throw new IllegalASTFormatException(
+              "parsing error,statement: " + sqlStr + " .message:" + e.getMessage());
+    }
+    AstNode astNode = ParseUtils.findRootNonNullToken(astTree);
+    RootOperator operator = generator.getLogicalPlan(astNode);
+    // expected to throw LogicalOperatorException: LIMIT <N>: N must be a positive integer and can not be zero.
```
jixuan1989: Hi, this Test has not expectation?

server/src/main/java/org/apache/iotdb/db/qp/strategy/LogicalGenerator.java
Ring-k: The useless codes are removed. Thanks!

server/src/test/java/org/apache/iotdb/db/qp/plan/LogicalPlanSmallTest.java
Ring-k: I added some expectations. Thanks for your suggestion.

server/src/main/java/org/apache/iotdb/db/qp/strategy/LogicalGenerator.java
Ring-k: I removed the "<p>"s. Thanks for your suggestion.

server/src/main/java/org/apache/iotdb/db/qp/strategy/LogicalGenerator.java
Ring-k: I replaced /** with /*. Thanks for your suggestion.

server/src/main/java/org/apache/iotdb/db/qp/constant/TqlParserConstant.java
Ring-k: I replaced /** with /*. Thanks for your suggestion.

Genius-pig:
>  No "=>" operators anymore.
"=>" operator means always execute predicate. Removing all of them is a real improvement? Syntactic predicates were used to work around a prediction weakness in ANTLR 3.

server/src/main/antlr3/org/apache/iotdb/db/sql/parse/TqlLexer.g
```diff
@@ -0,0 +1,539 @@
...
+fragment
+NUM
+    :   '0'..'9'
+    ;
+
+
+
+
+// ***************************
+fragment A
```
Genius-pig: Why don’t you put them together?
```
fragment
Letter
    : 'a'..'z' | 'A'..'Z'
    ;
```

Ring-k: Separating them makes it easier to define the other key words, easier to make the sql case insensitive. For example, 
K_SELECT
    : S E L E C T
    ;

If I only define Letter, it will be 
K_SELECT
    : 'SELECT' | 'select'
    ;

Ring-k:
> > No "=>" operators anymore.
> 
> "=>" operator means always execute predicate. Removing all of them is a real improvement? Syntactic predicates were used to work around a prediction weakness in ANTLR 3.

For each syntactic predicate, ANTLR defines a special method that returns true or false depending on whether the predicate’s grammar fragment matches the next input symbols. So if syntactic predicate is frequently used, the method will be frequently called. So, in my opinion, a simpler design leads to better performance.

server/src/main/antlr3/org/apache/iotdb/db/sql/parse/TqlLexer.g
jixuan1989: any questions? @Genius-pig

Genius-pig: You can use code to uppercase input String, what @Ring-k has done will lead to Antlr file is too long.
Check out what has spark done. https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/parser/ParseDriver.scala
https://github.com/apache/spark/blob/master/sql/catalyst/src/main/antlr4/org/apache/spark/sql/catalyst/parser/SqlBase.g4

server/src/main/antlr3/org/apache/iotdb/db/sql/parse/TqlParser.g
jt2594838: Yes, redundant things should always be removed.

server/src/main/antlr3/org/apache/iotdb/db/sql/parse/TqlLexer.g
Ring-k: It seems that before parsing, they capitalize all characters in a string, and while it is not case-sensitive to keywords, it is also case-insensitive to paths, values, and so on. To aviod it, I chose the way as it is now. There are some other examples.
https://github.com/antlr/grammars-v3/blob/master/mysql/MySQL.g

I think both ways of implementation are fine. If you think the ANTLR file is too long, separating lexer and parser rules in different files is a solution, as what we have done.

Genius-pig: I agree with you. There are both ways to handle case-insensitive according to https://github.com/antlr/antlr4/blob/master/doc/case-insensitive-lexing.md which is the Antlr official site. Although your way may cause a little slower, it can handle both case-sensitive and case-insensitive keywords.

server/src/main/antlr3/org/apache/iotdb/db/sql/parse/TqlParser.g
Ring-k: These codes are removed. 
Thanks for your suggestion!

PR #445, opened 2019-10-11
generate cpp, go, and python thrift files under service-rpc

jixuan1989:
after run `mvn generate-sources`, you can find cpp, python, go apis under service-rpc/target/generated-source-{language}.

So, you just need to supply how to use these APIs in other PRs.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -886,7 +886,7 @@
                         <version>0.1.11</version>
                         <executions>
                             <execution>
-                                <id>generate-thrift-sources</id>
+                                <id>generate-thrift-sources-java</id>
                                 <phase>generate-sources</phase>
                                 <goals>
                                     <goal>compile</goal>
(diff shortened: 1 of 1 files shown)


jt2594838:
This is fine.
But programmers of other languages do not usually have maven installed, so I doubt whether this is that useful.
And I would appreciate it if you add a parameter like `mvn generate-sources py` to let the user select what they want to build instead of building all at once.

jixuan1989:
> This is fine.
> But programmers of other languages do not usually have maven installed, so I doubt whether this is that useful.
> And I would appreciate it if you add a parameter like `mvn generate-sources py` to let the user select what they want to build instead of building all at once.

Actually, for users who do not install maven, they can use `./mvnw.sh generate-sources` ... 
allowing claiming which language codes users want is good, but I think it is ok now because generating codes is very fast..

What interests me is that  @chrisdutz  said "this is just generating codes, rather compile them as a module"...

PR #455, opened 2019-10-17
[IOTDB-251]improve TSQueryDataSet structure in RPC

LeiRui:
The original definition of TSQueryDataSet leads to the row-at-a-time writing process listed as follows:
```
- TSIService.fetchResults_result.write
  - TSFetchResultsResp.write
    - TSQueryDataSet.write
      - TSRowRecord.write
        - TSDataValue.write
          - TBinaryProtocol.writeDouble
            ---------------
            - TIOStreamTransport.write (这里实际是TSocket.write) 阻塞式socker
              ---------------
              - BufferedOutputStream.write
```
Therefore, by defining a more compact TSQueryDataSet, the time cost of `TSIService.fetchResults_result.write` may likely be decreased.

The original definition of TSQueryDataSet:
```
struct TSFetchResultsResp{
	1: required TSStatus status
	2: required bool hasResultSet
	3: optional TSQueryDataSet queryDataSet
}

struct TSQueryDataSet{
	1: required list\<TSRowRecord> records
}

struct TSRowRecord{
  1: required i64 timestamp
  // column values
  2: required list\<TSDataValue> values
}

struct TSDataValue{
  1: required bool is_empty
  2: optional bool bool_val
  3: optional i32 int_val
  4: optional i64 long_val
  5: optional double float_val
  6: optional double double_val
  7: optional binary binary_val
  8: optional string type
}
```
The new TSQueryDataSet in this pr:
```
struct TSQueryDataSet{
   1: required binary values
   2: required i32 rowCount
}
```


LeiRui:
Test is on the way.

LeiRui:
Credit to @samperson1997 , in his experiment, this pr can decrease the time cost of `fetchResults_result.write` to 27% of its original value.

server/src/main/java/org/apache/iotdb/db/utils/QueryDataSetUtils.java
```diff
@@ -55,66 +55,86 @@ public static TSQueryDataSet convertQueryDataSetByFetchSize(QueryDataSet queryDa
...
+        // use columnOutput to write byte array
+        dataOutputStreams[0].writeLong(rowRecord.getTimestamp());
+        List<Field> fields = rowRecord.getFields();
+        for (int k = 0; k < fields.size(); k++) {
+          Field field = fields.get(k);
+          DataOutputStream dataOutputStream = dataOutputStreams[k + 1]; // DO NOT FORGET +1
+          if (field.getDataType() == null) {
+            dataOutputStream.writeBoolean(true); // is_empty true
+          } else {
+            dataOutputStream.writeBoolean(false); // is_empty false
```
jt2594838: Optional: replace the booleans with a bit map which induce less overhead.

LeiRui: Hi, yes, this is scheduled as a possible future improvement as for the inner structure of TSQueryDataSet's value buffer. And more experiments are needed.
But for now, the core aim of this pr, which is to remove record-wise structure in TSQueryDataSet, is hit.

LeiRui:
> How about this:
> Deprecate the List of RowRecord and instead use a RowBatch, which avoids too many objects such as Fields and RowRecord itself.

Hi, thank you, I have recorded your thought on JIRA https://github.com/apache/incubator-iotdb/pull/455 for future work.

SilverNarcissus:
Thanks for your work. The compact view of dataset will definitely improve the efficiency of network transport. I'm ready to have a try!

PR #458, opened 2019-10-18
[IOTDB-165][TsFile] Fix a example

Genius-pig:
Fix a problem caused by the deleted position() method.

### DIFF
--- a/example/tsfile/src/main/java/org/apache/iotdb/tsfile/TsFileSequenceRead.java
+++ b/example/tsfile/src/main/java/org/apache/iotdb/tsfile/TsFileSequenceRead.java
@@ -23,6 +23,7 @@ import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.List;
 import java.util.stream.Collectors;
+import org.apache.iotdb.tsfile.common.conf.TSFileConfig;
 import org.apache.iotdb.tsfile.common.conf.TSFileDescriptor;
 import org.apache.iotdb.tsfile.encoding.decoder.Decoder;
 import org.apache.iotdb.tsfile.file.MetaMarker;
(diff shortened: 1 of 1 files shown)


jixuan1989:
I  feel a little confusing, why call position() here (but we do not call it in previous version)?

EJTTianYu:
Because in the previous PR , the position() call has been deleted.
image

Genius-pig:
## why I delete `position()` method?
1. This was requested change came from @jt2594838. @qiaojialin and I talked personally, then we decided to delete it.
2. This was literally because every time when you new a `TsFileSequenceReader` object and its constructor would call `loadMetadataSize()` method, `loadMetadataSize()` would call `position()` method. In local mode, it actually called `FileChannel.position()` method. As we know, it was a method would move the position pointer. So questions were **Why to move the position pointer? It is really necessary to move the position pointer when I new a `TsFileSequenceReader` object?**.
## why to move the position pointer?
Let's firstly check out TsFile structure. 
Screen Shot 2019-10-19 at 10 10 39 PM
You can find `MetaMarker.CHUNK_HEADER=0x1` after Magic String. There is a method called `readMarker()` to read `MetaMarker.CHUNK_HEADER=0x1`.
```
  /**
   * read one byte from the input. <br> this method is not thread safe
   */
  public byte readMarker() throws IOException {
    markerBuffer.clear();
    if (ReadWriteIOUtils.readAsPossible(tsFileInput, markerBuffer) == 0) {
      throw new IOException("reach the end of the file.");
    }
    markerBuffer.flip();
    return markerBuffer.get();
  }
```
There is no `position()` method in `readMarker()`. So it used to use `position()` in `loadMetadataSize()` to move the position pointer. But there are many tests and examples use `readMarker()`. After you remove `position()` in `loadMetadataSize()`, you have to use `position()` before `readMarker()` or use `readMarker(TSFileConfig.MAGIC_STRING.getBytes().length+TSFileConfig.MAGIC_STRING.getBytes().length)`. @qiaojialin suggested me to use the first one and I just done it. I also suggest add `@Deprecated` before `readMarker()` in case someone else to use it. 

## It is really necessary to move the position pointer when I new a `TsFileSequenceReader` object?
No. Move the position pointer when you really need it.

## What other jobs does this pr have?
I want to update some tsfile documents in this pr.

Genius-pig:
@jixuan1989 @EJTTianYu

PR #462, opened 2019-10-20
[IOTDB-253]time expression

little-emotion:
1、Supports addition and subtraction duration  based on now()


server/src/main/java/org/apache/iotdb/db/qp/constant/DatetimeUtils.java
```diff
@@ -475,6 +475,66 @@ public static long convertDatetimeStrToLong(String str, ZoneOffset offset, int d
...
+      case "h":
+        res *= 3600_000;
+        break;
+      case "m":
+        res *= 60_000;
+        break;
+      case "s":
+        res *= 1_000;
+        break;
+      default:
```
LeiRui: Where is "ns"?

Plus, would it be better to make `String unit` a constant or enum or something?

LeiRui: Plus Plus, why doesn't the default branch throw an exception?

LeiRui: Plus Plus Plus, if this miss of "ns" is a bug, better add tests to cover this.

docs/Documentation-CHN/UserGuide/2-Concept Key Concepts and Terminology/1-Key Concepts and Terminology.md
```diff
@@ -166,6 +170,42 @@ IoTDB在显示时间戳时可以支持LONG类型以及DATETIME-DISPLAY类型，
 
 </center>
 
+* 相对时间戳
+  
+  相对时间是指与服务器时间```now()```和```DATETIME```类型时间相差一定时间间隔的时间。
+  形式化定义为：
+  ```
+  Duration = (Digit+ ('Y'|'MO'|'W'|'D'|'H'|'M'|'S'|'MS'|'US'|'NS'))+
+  RelativeTime = (now() | DATETIME) ((+|-) Duration)+
```
LeiRui: I looked at the realization and the realization seems to also allow `Duration ((+|-) Duration)+`. What do you think?

little-emotion: 5Y6Y is equal to 11Y. 7D8D is equal to 7D+8D, which is 15D.

server/src/main/java/org/apache/iotdb/db/qp/constant/DatetimeUtils.java
little-emotion: > Plus Plus, why doesn't the default branch throw an exception?
> Plus Plus Plus, if this miss of "ns" is a bug, better add tests to cover this.
This function has the second half.
```
if (timestampPrecision.equals("us")) {
      if (unit.equals("ns")) {
        return value / 1000;
      } else if (unit.equals("us")) {
        return value;
      } else {
        return res * 1000;
      }
    } else if (timestampPrecision.equals("ns")) {
      if (unit.equals("ns")) {
        return value;
      } else if (unit.equals("us")) {
        return value * 1000;
      } else {
        return res * 1000_000;
      }
    } else {
      if (unit.equals("ns")) {
        return value / 1000_0000;
      } else if (unit.equals("us")) {
        return value / 1000;
      } else {
        return res;
      }
    }
```

little-emotion: Yes, I will make String unit a  enum.

docs/Documentation-CHN/UserGuide/2-Concept Key Concepts and Terminology/1-Key Concepts and Terminology.md
little-emotion: oh, it indeed allows Duration ((+|-) Duration)+. I will fix it.

server/src/main/java/org/apache/iotdb/db/qp/strategy/LogicalGenerator.java
```diff
@@ -800,60 +815,87 @@ private TSDataType parseTypeNode(AstNode typeNode) throws LogicalOperatorExcepti
...
+    for (int i = 0; i < durationStr.length(); i++) {
+      char ch = durationStr.charAt(i);
+      if (Character.isDigit(ch)) {
+        tmp *= 10;
+        tmp += (ch - '0');
+      } else {
+        String unit = durationStr.charAt(i) + "";
+        if (i + 1 < durationStr.length() && !Character.isDigit(durationStr.charAt(i + 1))) {
+          i++;
+          unit += durationStr.charAt(i);
```
jixuan1989: e.. is this for recognize "mo"? 

why not put `DurationUnit durationUnit = DurationUnit.valueOf(unit);` here?

server/src/main/java/org/apache/iotdb/db/qp/strategy/LogicalGenerator.java
```diff
@@ -800,60 +815,87 @@ private TSDataType parseTypeNode(AstNode typeNode) throws LogicalOperatorExcepti
...
+
+  private Long parseTokenTime(AstNode astNode) throws LogicalOperatorException {
+    if (astNode.getType() == TqlParser.TOK_DURATION) {
+      return parseTokenDuration(astNode);
+    }
+    return parseTimeFormat(cascadeChildrenText((astNode)));
+  }
+
+  private Long parseTokenDuration(AstNode astNode) {
+    String durationStr = cascadeChildrenText(astNode);
```
jixuan1989: it is better to add some comments to introduce what the AstNode looks like.. to help people understand the codes.

server/src/main/antlr3/org/apache/iotdb/db/sql/parse/TqlLexer.g
```diff
@@ -538,3 +538,8 @@ WS  :   ( ' '
         | '\n'
         ) {$channel=HIDDEN;}
     ;
+
+Duration
```
Ring-k: Hi, it will be better to use "DURATION" to keep it consistent with other rules.

server/src/main/antlr3/org/apache/iotdb/db/sql/parse/TqlLexer.g
```diff
@@ -538,3 +538,8 @@ WS  :   ( ' '
...
+    :
+    (NUM+ ('Y'|'MO'|'W'|'D'|'H'|'M'|'S'|'MS'|'US'|'NS'))+
```
Ring-k: Hi, it seems that the duration unit is case sensitive.  But is it necessary?

little-emotion: Although written in uppercase, it can recognize both uppercase and lowercase.
And I found if it was written in lowercase, it doesn't recognize uppercase or lowercase.

server/src/main/antlr3/org/apache/iotdb/db/sql/parse/TqlLexer.g
little-emotion: OK, I'll fix it.

server/src/main/antlr3/org/apache/iotdb/db/sql/parse/TqlLexer.g
little-emotion: I'll use fragment Letter defined in TqlLexer to keep consistent with other rules.

server/src/main/java/org/apache/iotdb/db/qp/strategy/LogicalGenerator.java
little-emotion: ok, I'll add comments .

server/src/main/java/org/apache/iotdb/db/qp/strategy/LogicalGenerator.java
little-emotion: Yes, this is to identify units with two letters.

PR #464, opened 2019-10-21
Modified Decoder and SequenceReader to support old version of TsFile

HTHou:


tsfile/src/main/java/org/apache/iotdb/tsfile/encoding/decoder/PlainDecoder.java
```diff
@@ -54,26 +55,55 @@ public boolean readBoolean(ByteBuffer buffer) {
 
   @Override
   public short readShort(ByteBuffer buffer) {
+    if (this.endianType == EndianType.LITTLE_ENDIAN) {
+      int ch1 = ReadWriteIOUtils.read(buffer);
+      int ch2 = ReadWriteIOUtils.read(buffer);
+      return (short) (ch1 + (ch2 << 8));
+    }
     return buffer.getShort();
```
jt2594838: A better way would be setting the endian of the buffer using `byteBuffer.order()` when the buffer is created, which can avoid many redundant checks.
Notice: ByteBuffers get from methods like `byteBuffer.slice()` do not inherit the byte order, so you have to set byte order for those ByteBuffers.

tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/TsDigest.java
```diff
@@ -65,18 +67,49 @@ public static int serializeNullTo(ByteBuffer buffer) {
...
+          digest.serializedSize += Short.BYTES + Integer.BYTES + value.remaining();
+    	}
+      }
+      else {
+    	for (int i = 0; i < size; i++) {
+          short n = ReadWriteIOUtils.readShort(inputStream);
+          value = ReadWriteIOUtils.readByteBufferWithSelfDescriptionLength(inputStream);
+          digest.statistics[n] = value;
+          digest.serializedSize += Short.BYTES + Integer.BYTES + value.remaining();
+    	}
```
jt2594838: Please reformat the above codes.

tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/TsFileMetaData.java
```diff
@@ -150,9 +150,11 @@ public static TsFileMetaData deserializeFrom(ByteBuffer buffer) throws IOExcepti
 
     if (ReadWriteIOUtils.readIsNull(buffer)) {
       fileMetaData.createdBy = ReadWriteIOUtils.readString(buffer);
-    }
-    fileMetaData.totalChunkNum = ReadWriteIOUtils.readInt(buffer);
-    fileMetaData.invalidChunkNum = ReadWriteIOUtils.readInt(buffer);
+    }  
+    // if using v0.8.0 TsFile, use 0 to represent missing fields
+    fileMetaData.totalChunkNum = buffer == null ? 0 : ReadWriteIOUtils.readInt(buffer);
+    fileMetaData.invalidChunkNum = buffer == null ? 0 : ReadWriteIOUtils.readInt(buffer);
```
jt2594838: Could this buffer really be null?

tsfile/src/main/java/org/apache/iotdb/tsfile/read/TsFileSequenceReader.java
```diff
@@ -346,6 +341,12 @@ private ChunkHeader readChunkHeader(long position, boolean markerRead) throws IO
    * @return the pages of this chunk
    */
   public ByteBuffer readChunk(ChunkHeader header) throws IOException {
+    if (this.readVersionNumber().startsWith("v")) {
+      config.setEndian("LITTLE_ENDIAN");
+    }
+    else {
+      config.setEndian("BIG_ENDIAN");
+    }
```
jt2594838: I think this will cause a mess if two files of different endian are being read concurrently.

HTHou: > I think this will cause a mess if two files of different endian are being read concurrently.

Yes, it may cause a mess, but I have no other idea about how and where to determine the endian type.

tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/TsFileMetaData.java
HTHou: > Could this buffer really be null?


Did you mean I should change `buffer == null`to `!ReadWriteIOUtils.readIsNull(buffer)` ?

tsfile/src/main/java/org/apache/iotdb/tsfile/encoding/decoder/PlainDecoder.java
HTHou: > A better way would be setting the endian of the buffer using `byteBuffer.order()` when the buffer is created, which can avoid many redundant checks.
> Notice: ByteBuffers get from methods like `byteBuffer.slice()` do not inherit the byte order, so you have to set byte order for those ByteBuffers.

This is a very good point! Thank you.

tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/TsFileMetaData.java
HTHou: > Could this buffer really be null?

Thank you. I've understood this point.

tsfile/src/main/java/org/apache/iotdb/tsfile/read/TsFileSequenceReader.java
jt2594838: You may add a field in TsFileSequenceReader to indicate the byte order and pass it down throughout the query process.

tsfile/src/main/java/org/apache/iotdb/tsfile/encoding/decoder/PlainDecoder.java
```diff
@@ -56,53 +57,39 @@ public boolean readBoolean(ByteBuffer buffer) {
...
+      buffer.order(ByteOrder.LITTLE_ENDIAN);
     }
     return buffer.getFloat();
   }
 
   @Override
   public double readDouble(ByteBuffer buffer) {
     if (this.endianType == EndianType.LITTLE_ENDIAN) {
-      return Double.longBitsToDouble(readLong(buffer));
+      buffer.order(ByteOrder.LITTLE_ENDIAN);
```
jt2594838: This is okay. But the checks can be saved if you set the byte order when the buffer is created instead of when calling these methods.

HTHou: > This is okay. But the checks can be saved if you set the byte order when the buffer is created instead of when calling these methods.

Yes, you are right! Thanks a lot.

tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/TsDigest.java
```diff
@@ -1,250 +1,317 @@
...
+      	        break;
+      	      case "last":
+      	        n = 3;
+      	        break;
+      	      case "sum":
+      	        n = 4;
+      	        break;
+      	      default:
+      	        n = -1;
+            }
```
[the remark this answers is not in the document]
HTHou: > It‘s better to replace these codes with `short n = (short)(StatisticType.valueOf("BATCHINSERT").ordinal()`.

There will be a problem if I use the code you mentioned. The names of StatisticType has changed in new version of TsFile. In old TsFile, for example, their names are `first`, `last`, etc. But in new file, the names change to  `first_value`, `last_value`, etc. 

Therefore, I don't think we should use `short n = (short)(StatisticType.valueOf("BATCHINSERT").ordinal()` at this place.

tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/TsDigest.java
```diff
@@ -1,250 +1,317 @@
...
+      	      n = 2;
+      	      break;
+      	    case "last":
+      	      n = 3;
+      	      break;
+      	    case "sum":
+      	      n = 4;
+      	      break;
+      	    default:
+      	      n = -1;
```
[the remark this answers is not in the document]
HTHou: > The same suggestion as above.

The same reason.

tsfile/src/main/java/org/apache/iotdb/tsfile/read/reader/chunk/ChunkReader.java
```diff
@@ -152,11 +153,13 @@ private PageReader constructPageReaderForNextPage(int compressedPageBodyLength)
               + Arrays.toString(compressedPageBody) + ". Actual:" + chunkDataBuffer
               .remaining());
     }
-
     chunkDataBuffer.get(compressedPageBody, 0, compressedPageBodyLength);
     valueDecoder.reset();
-    PageReader reader = new PageReader(ByteBuffer.wrap(unCompressor.uncompress(compressedPageBody)),
-        chunkHeader.getDataType(),
+    ByteBuffer pageData = ByteBuffer.wrap(unCompressor.uncompress(compressedPageBody));
+    if (endianType.equals("LITTLE_ENDIAN")) {
```
jt2594838: It would be better to make "LITTLE_ENDIAN" a constant or an enum.

HTHou: > It would be better to make "LITTLE_ENDIAN" a constant or an enum.

I have changed "LITTLE_ENDIAN" to an enum.

PR #465, opened 2019-10-22
[IOTDB-261]Check path validity in session

fanhualta:
When using Session to set storage group or create time series, the path needs to be checked in order to reject this creation if it contains special characters.

### DIFF
--- a/session/src/test/java/org/apache/iotdb/session/IoTDBSessionIT.java
+++ b/session/src/test/java/org/apache/iotdb/session/IoTDBSessionIT.java
@@ -280,4 +280,59 @@ public class IoTDBSessionIT {
     session.executeNonQueryStatement(
         "insert into root.sg1.d1(timestamp,s1, s2, s3) values(100, 1,2,3)");
   }
+
+  @Test
+  public void checkPathTest()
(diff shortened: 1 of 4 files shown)


session/src/test/java/org/apache/iotdb/session/IoTDBSessionIT.java
```diff
@@ -280,4 +280,59 @@ private void insert_via_sql() throws TException, IoTDBRPCException {
...
+      throws ClassNotFoundException, SQLException, IoTDBSessionException, TException, IoTDBRPCException {
+    session = new Session("127.0.0.1", 6667, "root", "root");
+    session.open();
+
+    //test set sg
+    checkSetSG(session, "root.vehicle", true);
+    checkSetSG(session, "root.123456", true);
+    checkSetSG(session, "root._1234", true);
+    checkSetSG(session, "root._vehicle", true);
+    checkSetSG(session, "root.\tvehicle", false);
```
jixuan1989: how about root.\\tvehicle

fanhualta: Don't support it.  The check rules here are the same as those of `TSParser.g`.

PR #466, opened 2019-10-22
[IOTDB-208] Bloom filter

SilverNarcissus:
Add bloom filter to skip tsfile which doesn't contains some timeseries


tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/TsFileMetaData.java
```diff
@@ -247,9 +271,59 @@ public int serializeTo(OutputStream outputStream) throws IOException {
...
+   * @param outputStream -output stream to determine byte length
+   * @return -byte length
+   */
+  public int serializeBloomFilter(OutputStream outputStream, List<ChunkGroupMetaData> chunkGroupMetaDataList)
+      throws IOException {
+    int byteLen = 0;
+    BloomFilter filter = buildBloomFilter(chunkGroupMetaDataList);
+
+    byte[] bytes = filter.serialize();
+    byteLen += ReadWriteIOUtils.write(ByteBuffer.wrap(bytes), outputStream);
```
jt2594838:
```
outputStream.write(bytes);
byteLen += bytes.length;
```
 should be enough, there is no need to wrap with a ByteBuffer.

tsfile/src/main/java/org/apache/iotdb/tsfile/read/query/executor/TsFileExecutor.java
```diff
@@ -54,6 +58,25 @@ public TsFileExecutor(IMetadataQuerier metadataQuerier, IChunkLoader chunkLoader
...
+    if(bloomFilter != null) {
+      for (Path path : queryExpression.getSelectedSeries()) {
+        if (bloomFilter.contains(path.getFullPath())) {
+          filteredSeriesPath.add(path);
+        }
+      }
+      if (filteredSeriesPath.isEmpty()) {
+        // return an empty QueryDataSet
+        LOG.warn(
+            "This tsfile not contains all paths which specified in selected series, return empty dataset");
```
jt2594838: `warn` may be too high. Querying old files with newly created time series may often result in this branch, I think `debug` is enough.

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/BloomFilter.java
```diff
@@ -0,0 +1,136 @@
...
+ */
+package org.apache.iotdb.tsfile.utils;
+
+import java.util.BitSet;
+
+public class BloomFilter {
+
+  private static final int MINIMAL_SIZE = 256;
+  private static final int MAXIMAL_HASH_FUNCTION_SIZE = 8;
+  private static final int[] seeds = new int[]{5, 7, 11, 19, 31, 37, 43, 59};
```
jt2594838: seeds -> SEEDS

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/BloomFilter.java
```diff
@@ -0,0 +1,136 @@
...
+  /**
+   * get empty bloom filter
+   *
+   * @param errorPercent the tolerant percent of error of the bloom filter
+   * @param numOfString the number of string want to store in the bloom filter
+   * @return empty bloom
+   */
+  public static BloomFilter getEmptyBloomFilter(double errorPercent, int numOfString) {
+    errorPercent = Math.max(errorPercent, 0.01);
+    errorPercent = Math.min(errorPercent, 0.1);
```
jt2594838: Why adding these constraints?

tsfile/src/main/java/org/apache/iotdb/tsfile/write/writer/TsFileIOWriter.java
```diff
@@ -277,6 +275,10 @@ public void endFile(Schema schema) throws IOException {
     int size = tsFileMetaData.serializeTo(out.wrapAsStream());
     logger.debug("finish flushing the footer {}, file pos:{}", tsFileMetaData, out.getPosition());
 
+    // write bloom filter
+    size += tsFileMetaData.serializeBloomFilter(out.wrapAsStream(), chunkGroupMetaDataList);
+    logger.debug("finish flushing the bloom filter file pos:{}", out.getPosition());
```
jt2594838: Wrap with `if (logger.isDebugEnabled())`, `out.getPosition()` could be costly, please also fix logs above.

jt2594838:
There are failing tests, please have a look.

tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/TsFileMetaData.java
```diff
@@ -110,6 +120,13 @@ public static TsFileMetaData deserializeFrom(InputStream inputStream) throws IOE
     }
     fileMetaData.totalChunkNum = ReadWriteIOUtils.readInt(inputStream);
     fileMetaData.invalidChunkNum = ReadWriteIOUtils.readInt(inputStream);
+    // read bloom filter
+    if(!ReadWriteIOUtils.checkIfMagicString(inputStream)){
```
jt2594838: Maybe you can use `available()` to perform a quick check.

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/BloomFilter.java
SilverNarcissus: There are two reasons:
1.If user set errorPercent smaller than 0 or bigger than 1, iotdb will crash.
2. If user set errorPercent very small, such as 0.00000001, the size of bloom filter will very big (GB level), so I add these constraints

SilverNarcissus:
As for the performance, for 10000 timeseries. Read tsfile metadata will cost 124ms while using bloom filter just cost 4ms. More complex test will be done lately.

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/BloomFilter.java
jt2594838: Yeah, but I still wonder why the specific range [0.01, 0.1] is used. Does any literature argue for that, or you just decide randomly?

jixuan1989:
except how to use it, you need to update the two docs:
 
https://cwiki.apache.org/confluence/display/IOTDB/TsFile+Format

and

docs/Documentation/UserGuide/7-System Design/1-Hierarchy.md

tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/TsFileMetaData.java
jixuan1989: at least you can not use the name of `checkIfMagicString` because you can only know whether the left is not a magic string, you can not know whether it is a correct magic string...  so, `notMagicString()` may be better than now.

tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/TsFileMetaData.java
```diff
@@ -247,9 +271,61 @@ public int serializeTo(OutputStream outputStream) throws IOException {
...
+   * get all path in this tsfile
+   *
+   * @return all path in set
+   */
+  private List<String> getAllPath(List<ChunkGroupMetaData> chunkGroupMetaDataList) {
+    List<String> res = new ArrayList<>();
+    for (ChunkGroupMetaData chunkGroupMetaData : chunkGroupMetaDataList) {
+      String deviceId = chunkGroupMetaData.getDeviceID();
+      for (ChunkMetaData chunkMetaData : chunkGroupMetaData.getChunkMetaDataList()) {
+        res.add(deviceId + PATH_SEPARATOR + chunkMetaData.getMeasurementUid());
```
jixuan1989: um, I just wonder don't we have somewhere to store these sentence in memory?

tsfile/src/main/java/org/apache/iotdb/tsfile/read/query/executor/TsFileExecutor.java
```diff
@@ -54,6 +58,25 @@ public TsFileExecutor(IMetadataQuerier metadataQuerier, IChunkLoader chunkLoader
...
+      }
+      if (filteredSeriesPath.isEmpty()) {
+        // return an empty QueryDataSet
+        LOG.debug(
+            "This tsfile not contains all paths which specified in selected series, return empty dataset");
+        return new DataSetWithoutTimeGenerator(Collections.emptyList(), Collections.emptyList(),
+            Collections.emptyList());
+      }
+    }
+    queryExpression.setSelectSeries(filteredSeriesPath);
```
jixuan1989: actually, I wonder why the performance is improved.
After you do `queryExpression.setSelectSeries(filteredSeriesPath);`,  you have to do `metadataQuerier.loadChunkMetaDatas(queryExpression.getSelectedSeries());` right?

jixuan1989: and, when you run experiments, you'd better discuss in several cases:
if all paths exist; 
if most of paths  exist;
if a few paths exist;
if no paths exist.

jixuan1989:
I have read the codes and I do not know why the performance is improved....

tsfile/src/main/java/org/apache/iotdb/tsfile/read/query/executor/TsFileExecutor.java
jt2594838: 1. If the file contains no required timeseries, it will be very fast using the bloom filter.
2. Even if only some of the timeseries are filtered, it reduces the lower calls to search and compare String. As you know, String comparison can be quite costly.

jt2594838:
Like what jixuan1989 has said, please find a place for your bloom filter in the User Guide.

server/src/main/java/org/apache/iotdb/db/engine/cache/DeviceMetaDataCache.java
```diff
@@ -44,11 +45,9 @@
 
   private static final Logger logger = LoggerFactory.getLogger(DeviceMetaDataCache.class);
   private static final IoTDBConfig config = IoTDBDescriptor.getInstance().getConfig();
-
+  private static final long MEMORY_THRESHOLD_IN_B = config.getAllocateMemoryForChumkMetaDataCache();
```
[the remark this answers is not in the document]
SilverNarcissus: This line was not my code... I don't know why it is here...  Let me ask other people

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/Murmur128Hash.java
```diff
@@ -0,0 +1,153 @@
+package org.apache.iotdb.tsfile.utils;
```
jt2594838: Add the license, please.

tsfile/src/main/java/org/apache/iotdb/tsfile/read/query/executor/TsFileExecutor.java
SilverNarcissus: For this part which is only be used by readOnlyTsfile, if one of the paths not in tsfile, we should just return an empty dataset or throw an exception. Because the original behavior is this.

tsfile/src/main/resources/tsfile-format.properties.template
```diff
@@ -27,6 +27,9 @@ page_size_in_byte=1048576
 # The maximum number of data points in a page, defalut 1024*1024
 max_number_of_points_in_page=1048576
 
+# The acceptable error rate of bloom filter, should be in [0.01, 0.1]
+bloom_filter_error_rate=0.05
```
LeiRui: I am afraid that just adding `bloom_filter_error_rate` to this `tsfile\src\main\resources\tsfile-format.properties.template` does not work.

Because if you check the start-server.bat/sh carefully, you will find that TSFILE_HOME equals IOTDB_HOME and IOTDB_HOME is default to be the direct parent directory of the start up script (so usually it will be like `...\server\target\iotdb-server-0.9.0-SNAPSHOT\`). That means in the TsFileDescriptor.loadProps function, the inputStream of properties will default to use the tsfile-format.properties under the `...\server\target\iotdb-server-0.9.0-SNAPSHOT\conf`, not this `tsfile\src\main\resources\tsfile-format.properties.template`. Note that `...\server\target\iotdb-server-0.9.0-SNAPSHOT\conf\tsfile-format.properties` is generated from `...\server\src\assembly\resources\conf\tsfile-format.properties`. 

So, of course this template should be updated, but `...\server\src\assembly\resources\conf\tsfile-format.properties` should also be updated.

SilverNarcissus: OK I will update that configure file

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/Murmur128Hash.java
SilverNarcissus: done

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/BloomFilter.java
SilverNarcissus: For 1 million time series, 0.01 error rate will use 10MB space. I think this may be too large if we specify a smaller error rate.
And then, bloom filter in hadoop's default error rate is 0.05.
So I choose 0.01 - 0.1

tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/TsFileMetaData.java
```diff
@@ -1,306 +1,385 @@
...
+    return res;
+  }
+
+  /**
+   * build bloom filter
+   * @return bloom filter
+   */
+  private BloomFilter buildBloomFilter(List<ChunkGroupMetaData> chunkGroupMetaDataList) {
+    BloomFilter bloomFilter = BloomFilter
+        .getEmptyBloomFilter(TSFileDescriptor.getInstance().getConfig().getBloomFilterErrorRate(), chunkGroupMetaDataList.size());
```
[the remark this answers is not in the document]
SilverNarcissus: Ok, I will fix it

docs/Documentation/UserGuide/4-Operation Manual/8-TsFile Usage.md
```diff
@@ -620,4 +620,16 @@ and then call:
 TSFileConfig config = TSFileDescriptor.getInstance().getConfig();
 ```
 
+## Bloom filter
+
+Bloom filter check whether a given time series might in the tsfile before load metadata. This can improve the performance of loading metadata and skip tsfile that doesn't contains specified time series.
+If you want to learn more about its mechanism, you can refer to: [wiki page of bloom filter](https://en.wikipedia.org/wiki/Bloom_filter).
+
+#### configuration 
+you can control the false positive rate of bloom filter by following parameter in the config file.
```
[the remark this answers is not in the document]
SilverNarcissus: OK, I will fix it

tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/TsFileMetaData.java
LeiRui: How about add the concatenated path into the bloom filter just within the loop?
Since you don't have to get all paths before you add them into the bloom filter.

tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/TsFileMetaData.java
```diff
@@ -1,306 +1,387 @@
...
+
+  /**
+   * build bloom filter
+   *
+   * @return bloom filter
+   */
+  private BloomFilter buildBloomFilter(List<ChunkGroupMetaData> chunkGroupMetaDataList) {
+    BloomFilter bloomFilter = BloomFilter
+        .getEmptyBloomFilter(TSFileDescriptor.getInstance().getConfig().getBloomFilterErrorRate(),
+            chunkGroupMetaDataList.size());
```
LeiRui: I have a question: chunkGroupMetaDataList.size() doesn't seem to equal the total number of string added to the bloom filter. The total number of strings added is the size of the returned list from `getAllPath` function, right?

tsfile/src/main/java/org/apache/iotdb/tsfile/read/query/executor/TsFileExecutor.java
LeiRui: According to the code, I think the bloom filter takes effect in the iotdb query process, not in the reading process of TsFile. As you know, `metadataQuerier.loadChunkMetaDatas(queryExpression.getSelectedSeries())` is used only in TsFile.

LeiRui: The bloom filter will play a vital role in the following example scenario:

Assume:
- The measurement scheme registered under the storage group `root.sg` is: s0,s1,s2,...,s3200
- The TsFile file1 only contains data from `root.sg.d0.s0`

Then query data from series paths: `root.sg.d0.s1`,`root.sg.d0.s2`,...,`root.sg.d0.s3200`.

As shown in `DeviceMetaDataCache.get`, without the bloom filter, the TsDeviceMetaData of `root.sg.d0` will be read from disk 3200 times. The existent metadata cache mechanism won't help here because List<ChunkMetaData> of `root.sg.d0.s0` doesn't exist in file1.

Now with the bloom filter in the TsFileMetaData, some of the 3200 queried series can be judged to be nonexistent in the file1, therefore eliminating the need to read TsDeviceMetaData from disk.

tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/TsFileMetaData.java
SilverNarcissus: Yes, I think this is a bug. So I have to build list of path and the add it to bloom filter. Only by this I can get the size of the bloom filter

tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/TsFileMetaData.java
```diff
@@ -1,306 +1,388 @@
...
-
-  public void setInvalidChunkNum(int invalidChunkNum) {
-    this.invalidChunkNum = invalidChunkNum;
-  }
-
-  public List<MeasurementSchema> getMeasurementSchemaList() {
-    return new ArrayList<MeasurementSchema>(measurementSchema.values());
-  }
-}
+/**
```
jixuan1989: I remember that we have changed `/**` to `/*`

tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/TsFileMetaData.java
jixuan1989: > Are you going to store all time series for each file in memory? That will surely explode.

I have no idea, i just consider concating the path again and again is time consuming.. (That is why I want to convert path to id)

> How about add the concatenated path into the bloom filter just within the loop?
Yes, but he must know the number of total paths, I am not sure whether he can get it.
```
BloomFilter bloomFilter = BloomFilter
        .getEmptyBloomFilter(TSFileDescriptor.getInstance().getConfig().getBloomFilterErrorRate(),
            paths.size());
```

tsfile/src/main/java/org/apache/iotdb/tsfile/read/query/executor/TsFileExecutor.java
```diff
@@ -54,6 +58,25 @@ public TsFileExecutor(IMetadataQuerier metadataQuerier, IChunkLoader chunkLoader
 
   @Override
   public QueryDataSet execute(QueryExpression queryExpression) throws IOException {
+    // bloom filter
+    BloomFilter bloomFilter = metadataQuerier.getWholeFileMetadata().getBloomFilter();
+    List<Path> filteredSeriesPath = new ArrayList<>();
+    if (bloomFilter != null) {
```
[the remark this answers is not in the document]
jixuan1989: +1 because we have to support v0.8.1 now..

tsfile/src/main/java/org/apache/iotdb/tsfile/read/query/executor/TsFileExecutor.java
jixuan1989: > The bloom filter will play a vital role in the following example scenario:
> 
> Assume:
> 
> * The measurement scheme registered under the storage group `root.sg` is: s0,s1,s2,...,s3200
> * The TsFile file1 only contains data from `root.sg.d0.s0`
> 
> Then query data from series paths: `root.sg.d0.s1`,`root.sg.d0.s2`,...,`root.sg.d0.s3200`.
> 
> As shown in `DeviceMetaDataCache.get`, without the bloom filter, the TsDeviceMetaData of `root.sg.d0` in file1 will be read from disk 3200 times. The existent metadata cache mechanism won't help here because it will only caches List <ChunkMetaData> of those existent series while `root.sg.d0.s0` doesn't exist in file1.
> 
> Now with the bloom filter in the TsFileMetaData, some of the 3200 queried series can be judged to be nonexistent in the file1, therefore eliminating the next step of reading TsDeviceMetaData from disk.

well, but how about if all 3200 paths exist in the file? is there any improvement?

> the TsDeviceMetaData of root.sg.d0 in file1 will be read from disk 3200 times.

By the way, I think it should be improvement...

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/Murmur128Hash.java
jixuan1989: hi, I just want to double check that is this class implemented by you rathar than other projects..

docs/Documentation/UserGuide/4-Operation Manual/8-TsFile Usage.md
```diff
@@ -620,4 +620,16 @@ and then call:
 TSFileConfig config = TSFileDescriptor.getInstance().getConfig();
 ```
 
+## Bloom filter
+
+Bloom filter check whether a given time series might in the tsfile before load metadata. This can improve the performance of loading metadata and skip tsfile that doesn't contains specified time series.
```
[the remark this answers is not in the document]
jt2594838: "check" -> "checks" or "filter" -> "filters"
"might" -> "is"
"load" -> "loading"
"tsfile" -> "the tsfile" or "tsfiles"

docs/Documentation/UserGuide/4-Operation Manual/8-TsFile Usage.md
```diff
@@ -620,4 +620,16 @@ and then call:
...
+If you want to learn more about its mechanism, you can refer to: [wiki page of bloom filter](https://en.wikipedia.org/wiki/Bloom_filter).
+
+#### configuration 
+you can control the false positive rate of bloom filter by the following parameter in the config file config file `tsfile-format.properties` which located at `/server/src/assembly/resources/conf` directory
```
jt2594838: "config file" is repeated
"located" -> "is located"

tsfile/src/test/java/org/apache/iotdb/tsfile/read/ReadOnlyTsFileTest.java
```diff
@@ -151,11 +149,13 @@ void queryNonExistPathTest() {
     QueryExpression queryExpression = QueryExpression.create(paths, expression);
     try {
       QueryDataSet queryDataSet = tsFile.query(queryExpression);
-      fail();
+      Assert.assertEquals(false, queryDataSet.hasNext());
```
jt2594838: Use `assertFalse` instead.

tsfile/src/main/java/org/apache/iotdb/tsfile/read/query/executor/TsFileExecutor.java
LeiRui: > well, but how about if all 3200 paths exist in the file? is there any improvement?

If all 3200 paths have their data existent in this tsfile, that mean this tsfile will also contain their ChunkMetaData, therefore when this tsfile is first read for a certain path, the certain path's and other "hot" sensors' ChunkMetaData will be cached (you can compare it to "disk prefetch").

tsfile/src/main/java/org/apache/iotdb/tsfile/read/query/executor/TsFileExecutor.java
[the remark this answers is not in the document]
SilverNarcissus: When I load metadata, I will check whether a tsfile contains the bloom filter. So we can both support old tsfile format and new tsfile format.

docs/Documentation/UserGuide/4-Operation Manual/8-TsFile Usage.md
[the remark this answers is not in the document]
SilverNarcissus: I will fix it~

docs/Documentation/UserGuide/4-Operation Manual/8-TsFile Usage.md
```diff
@@ -620,4 +620,16 @@ and then call:
...
+```
+# The acceptable error rate of bloom filter, should be in [0.01, 0.1]
+
+bloom_filter_error_rate
```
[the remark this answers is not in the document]
SilverNarcissus: fixed

server/src/assembly/resources/conf/tsfile-format.properties
```diff
@@ -47,4 +47,7 @@ value_encoder=PLAIN
 
 # Compression configuration
 # Data compression method, TsFile supports UNCOMPRESSED or SNAPPY. Default value is UNCOMPRESSED which means no compression
-compressor=UNCOMPRESSED
\ No newline at end of file
+compressor=UNCOMPRESSED
+
+# The acceptable error rate of bloom filter, should be in [0.01, 0.1]
```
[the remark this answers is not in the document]
SilverNarcissus: fixed

server/src/test/java/org/apache/iotdb/db/integration/IoTDBMultiSeriesIT.java
```diff
@@ -138,6 +153,7 @@ private static void insertData()
       statement.execute("flush");
       statement.execute("merge");
 
+
```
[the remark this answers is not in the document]
SilverNarcissus: removed~

tsfile/src/test/java/org/apache/iotdb/tsfile/read/ReadOnlyTsFileTest.java
SilverNarcissus: fixed

docs/Documentation/UserGuide/4-Operation Manual/8-TsFile Usage.md
SilverNarcissus: I will fix it

tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/TsFileMetaData.java
SilverNarcissus: I will fix it

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/BloomFilter.java
```diff
@@ -0,0 +1,142 @@
...
```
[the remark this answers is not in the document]
SilverNarcissus: Ok, I will add it to tsfile config

tsfile/src/main/java/org/apache/iotdb/tsfile/read/query/executor/TsFileExecutor.java
```diff
@@ -54,6 +58,25 @@ public TsFileExecutor(IMetadataQuerier metadataQuerier, IChunkLoader chunkLoader
...
+          LOG.debug(
+              "This tsfile not contains paths " + path
+                  + " which specified in selected series, return empty dataset");
+          return new DataSetWithoutTimeGenerator(Collections.emptyList(), Collections.emptyList(),
+              Collections.emptyList());
+        }
+      }
+    }
+    queryExpression.setSelectSeries(filteredSeriesPath);
+    //
```
qiaojialin: remove this or add comment

tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/TsFileMetaData.java
jixuan1989: Ok, before no more idea can be provided, I will do not block the PR by this issue.

tsfile/src/main/java/org/apache/iotdb/tsfile/read/query/executor/TsFileExecutor.java
jixuan1989: Ok, got it. 
But I think we also need to claim it clearly the improvement happens at which case. @SilverNarcissus

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/Murmur128Hash.java
jixuan1989: My question? @SilverNarcissus

jixuan1989:
Still have two questions left.

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/Murmur128Hash.java
SilverNarcissus: I will update license file.

SilverNarcissus:
Let's explain the improvement of performance case by case. 
## Basic idea
First of all, keep in mind that bloom filter will only improve the performance when **NONE** of the selected time series in the tsfile. It will save the time about loading chunk group metadata. Because bloom filter locate at the end of file metadata and it's very small, it will cause only minor impact overhead when any of the selected time series in the tsfile.
## Target case
Iotdb has lots of tsfile and each tsfile contains some time series. A query want to select some of the time series and these time series aren't present in most of the tsfiles.
## Analyze
1. For one tsfile
     If this tsfile contains 10000 timeseries, reading tsfile metadata will cost 124ms while using bloom filter just cost 4ms. 
2. For iotdb
    This analyze maybe very complex for iotdb has lru cache to cache chunk metadata. So when a query comes, for a specific tsfile there will be three possibilities. Let's say the time of loading metadata is t1, the time of using bloom filter is t2, the time of querying cache is t3, and cache and bloom filter will always hit.
## Three possibilities 
####   This tsfile contains all of the time series.
Time with bloom filter = t3 + t2 + t1 + (n - 1) * t3
Time without bloom filter = t3 + t1 + (n - 1) * t3
overhead : t2
####   This tsfile contains k of the time series.
Time with bloom filter  = t1 + (n - k) * (t3 + t2) + k * t3
Time without bloom filter = (n - k) * (t3 + t1) + k * t3
saving time (n - k) * (t1 - t2)  - t1
####    This tsfile contains none of the time series.
Time with bloom filter  = n * (t3 + t2)
Time without bloom filter = n * (t3 + t1)
saving time n * (t1 - t2)

PR #471, opened 2019-10-23
[IOTDB-265]Re-adjust the threshold size of memtable

fanhualta:
In the dynamic parameter adjustment module, it calculated the average size of each metatable. However, considering that the number of timeseries between storage groups may vary greatly, it's appropriate to judge whether to flush the memtable according to the average memtable size. We need to adjust it according to the number of timeseries in a specific storage group.

jira link: https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-265?filter=allopenissues

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/conf/IoTDBConstant.java
+++ b/server/src/main/java/org/apache/iotdb/db/conf/IoTDBConstant.java
@@ -65,4 +65,6 @@ public class IoTDBConstant {
   public static final String USER = "User";
   public static final String PRIVILEGE = "Privilege";
 
+  public static final int MEMTABLE_NUM_INCREMETN = 4;
+
 }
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/apache/iotdb/db/conf/IoTDBConstant.java
```diff
@@ -65,4 +65,6 @@ private IoTDBConstant() {
   public static final String USER = "User";
   public static final String PRIVILEGE = "Privilege";
 
+  public static final int MEMTABLE_NUM_INCREMETN = 4;
```
qiaojialin:
```suggestion
  public static final int MEMTABLE_NUM_INCREMENT = 4;
```

or  MEMTABLE_IN_EACH_STORAGE_GROUP

PR #472, opened 2019-10-24
[IOTDB-267]reduce IO operations in deserializing chunk header

little-emotion:
Currently, deserializing the chunk header requires 2 disk read operations. One is reading chunkheader length, the other is reading chunkheader content. So this pr is for reducing IO operations in  chunk header deserialization.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/qp/physical/PhysicalPlan.java
+++ b/server/src/main/java/org/apache/iotdb/db/qp/physical/PhysicalPlan.java
@@ -94,7 +94,7 @@ public abstract class PhysicalPlan {
     if (valueLen == NULL_VALUE_LEN) {
       return null;
     }
-    return ReadWriteIOUtils.readStringWithoutLength(buffer, valueLen);
+    return ReadWriteIOUtils.readStringWithLength(buffer, valueLen);
   }
 
   public static class Factory {
--- a/tsfile/src/main/java/org/apache/iotdb/tsfile/file/footer/ChunkGroupFooter.java
+++ b/tsfile/src/main/java/org/apache/iotdb/tsfile/file/footer/ChunkGroupFooter.java
@@ -106,7 +106,7 @@ public class ChunkGroupFooter {
     buffer = ByteBuffer.allocate(getSerializedSize(size));
     ReadWriteIOUtils.readAsPossible(input, offsetVar, buffer);
     buffer.flip();
-    String deviceID = ReadWriteIOUtils.readStringWithoutLength(buffer, size);
+    String deviceID = ReadWriteIOUtils.readStringWithLength(buffer, size);
     long dataSize = ReadWriteIOUtils.readLong(buffer);
     int numOfChunks = ReadWriteIOUtils.readInt(buffer);
     return new ChunkGroupFooter(deviceID, dataSize, numOfChunks);
--- a/tsfile/src/main/java/org/apache/iotdb/tsfile/file/header/ChunkHeader.java
+++ b/tsfile/src/main/java/org/apache/iotdb/tsfile/file/header/ChunkHeader.java
@@ -76,7 +76,8 @@ public class ChunkHeader {
   }
 
   public static int getSerializedSize(String measurementID) {
-      return Byte.BYTES + Integer.BYTES + getSerializedSize(measurementID.getBytes(TSFileConfig.STRING_CHARSET).length);
+    return Byte.BYTES + Integer.BYTES + getSerializedSize(
+        measurementID.getBytes(TSFileConfig.STRING_CHARSET).length);
   }
 
   private static int getSerializedSize(int measurementIdLength) {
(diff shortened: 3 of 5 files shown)


LeiRui:
Sounds good +1

little-emotion:
> One more push: If we record the total chunk size in the ChunkMetadata, we do not have to split the read of ChunkHeader and chunk data into 2 reads.

I've thought about it, but there are two other things to consider.
1. This will change the tsfile structure. I think it's best to change the file structure when upgrading a large version.
2. Although the granularity of data read by iotdb is chunk, we have been discussing whether we should change the granularity of data read by iotdb to page, especially in the case of insufficient memory.

jt2594838:
> > One more push: If we record the total chunk size in the ChunkMetadata, we do not have to split the read of ChunkHeader and chunk data into 2 reads.
> 
> I've thought about it, but there are two other things to consider.
> 
> 1. This will change the tsfile structure. I think it's best to change the file structure when upgrading a large version.
> 2. Although the granularity of data read by iotdb is chunk, we have been discussing whether we should change the granularity of data read by iotdb to page, especially in the case of insufficient memory.

1. That is okay, this change could be listed in the change list of 0.9.0. Fear not the changes.

2. Reading page by page will surely induce more IO, considering current read performance, I do think it is much beneficial. Moreover, you can adaptively choose to cache a page or a chunk depending on your available memory, which has no conflicts with the current code. And, in my opinion,  trade one IO for only another 8 bytes is so charming.

LeiRui:
> > > One more push: If we record the total chunk size in the ChunkMetadata, we do not have to split the read of ChunkHeader and chunk data into 2 reads.
> > 
> > 
> > I've thought about it, but there are two other things to consider.
> > 
> > 1. This will change the tsfile structure. I think it's best to change the file structure when upgrading a large version.
> > 2. Although the granularity of data read by iotdb is chunk, we have been discussing whether we should change the granularity of data read by iotdb to page, especially in the case of insufficient memory.
> 
> 1. That is okay, this change could be listed in the change list of 0.9.0. Fear not the changes.
> 2. Reading page by page will surely induce more IO, considering current read performance, I do think it is much beneficial. Moreover, you can adaptively choose to cache a page or a chunk depending on your available memory, which has no conflicts with the current code. And, in my opinion,  trade one IO for only another 8 bytes is so charming.

I can't get the second point. By `induce` you mean reduce or increase?

jt2594838:
> > > > One more push: If we record the total chunk size in the ChunkMetadata, we do not have to split the read of ChunkHeader and chunk data into 2 reads.
> > > 
> > > 
> > > I've thought about it, but there are two other things to consider.
> > > 
> > > 1. This will change the tsfile structure. I think it's best to change the file structure when upgrading a large version.
> > > 2. Although the granularity of data read by iotdb is chunk, we have been discussing whether we should change the granularity of data read by iotdb to page, especially in the case of insufficient memory.
> > 
> > 
> > 
> > 1. That is okay, this change could be listed in the change list of 0.9.0. Fear not the changes.
> > 2. Reading page by page will surely induce more IO, considering current read performance, I do think it is much beneficial. Moreover, you can adaptively choose to cache a page or a chunk depending on your available memory, which has no conflicts with the current code. And, in my opinion,  trade one IO for only another 8 bytes is so charming.
> 
> I can't get the second point. By `induce` you mean reduce or increase?

Increase, of course. Assuming we have 5 pages in a chunk, you will probably do 6 IOs (plus reading the header) to read the chunk if you read pages one by one.

qiaojialin:
> > > > > One more push: If we record the total chunk size in the ChunkMetadata, we do not have to split the read of ChunkHeader and chunk data into 2 reads.
> > > > 
> > > > 
> > > > I've thought about it, but there are two other things to consider.
> > > > 
> > > > 1. This will change the tsfile structure. I think it's best to change the file structure when upgrading a large version.
> > > > 2. Although the granularity of data read by iotdb is chunk, we have been discussing whether we should change the granularity of data read by iotdb to page, especially in the case of insufficient memory.
> > > 
> > > 
> > > 
> > > 1. That is okay, this change could be listed in the change list of 0.9.0. Fear not the changes.
> > > 2. Reading page by page will surely induce more IO, considering current read performance, I do think it is much beneficial. Moreover, you can adaptively choose to cache a page or a chunk depending on your available memory, which has no conflicts with the current code. And, in my opinion,  trade one IO for only another 8 bytes is so charming.
> > 
> > 
> > I can't get the second point. By `induce` you mean reduce or increase?
> 
> Increase, of course. Assuming we have 5 pages in a chunk, you will probably do 6 IOs (plus reading the header) to read the chunk if you read pages one by one.

As far as I know, we do read a Chunk into memory in one IO and then deserialize page one by one in memory.

LeiRui:
According to my experiment on this pr, the time cost ratio of `readChunkHeader` to `readChunk` is still 2:1, which overturns our former assumption that it is because `readChunkHeader` calls the read operation twice that the ratio is 2:1. 

image

I discussed this result with @little-emotion yesterday. And we plan to do more experiments on this to find out what really causes 2:1.

PR #473, opened 2019-10-24
[IOTDB-266]NullPoint exception when reading not existed devices using ReadOnlyTsFile

little-emotion:
update error info of querying non-existent path

### DIFF
--- a/tsfile/src/test/java/org/apache/iotdb/tsfile/read/ReadOnlyTsFileTest.java
+++ b/tsfile/src/test/java/org/apache/iotdb/tsfile/read/ReadOnlyTsFileTest.java
@@ -110,12 +110,13 @@ public class ReadOnlyTsFileTest {
 
   @Test
   public void test2() throws InterruptedException, WriteProcessException, IOException {
-    int minRowCount = 1000, maxRowCount=100000;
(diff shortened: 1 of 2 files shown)


tsfile/src/test/java/org/apache/iotdb/tsfile/read/ReadOnlyTsFileTest.java
```diff
@@ -139,5 +140,19 @@ void queryTest2() throws IOException {
...
+  void queryNonExistPathTest() {
+    ArrayList<Path> paths = new ArrayList<>();
+    paths.add(new Path("dr.s1"));
+    paths.add(new Path("d2.s1"));
+    IExpression expression = new GlobalTimeExpression(TimeFilter.gt(1480562664760L));
+    QueryExpression queryExpression = QueryExpression.create(paths, expression);
+    try {
+      QueryDataSet queryDataSet = tsFile.query(queryExpression);
+    } catch (IOException e) {
+      Assert.assertNotNull(e.getMessage(), "No measurement path : dr");
```
jt2594838: Add a `fail()`  after `QueryDataSet queryDataSet = tsFile.query(queryExpression);` in case that the Exception does not rise.
Why use `assertNotNull`?

PR #474, opened 2019-10-24
[IOTDB-233] Keep metadata plan clear

HTHou:
The metadata operator and metadataPlan have both a path and a deletePathList, which is confusing. It seems that they are used in different scenarios: 

private Path path;  (for set storage group, create timeseries)
private List<Path> deletePathList;  (for delete timeseries)

It's better to make it clear. Maybe by using more types of physical plans.

jira link: https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-233?filter=allopenissues


server/src/main/java/org/apache/iotdb/db/qp/executor/QueryProcessExecutor.java
```diff
@@ -110,13 +112,15 @@ public boolean processNonQuery(PhysicalPlan plan) throws ProcessorException {
         return operateWatermarkEmbedding(((DataAuthPlan) plan).getUsers(), true);
       case REVOKE_WATERMARK_EMBEDDING:
         return operateWatermarkEmbedding(((DataAuthPlan) plan).getUsers(), false);
+      case METADATA:
```
jt2594838: I think this should be removed and should not occur anywhere.

server/src/main/java/org/apache/iotdb/db/qp/physical/sys/AddPathPlan.java
```diff
@@ -0,0 +1,111 @@
...
+    return stringBuilder.toString();
+  }
+  
+  @Override
+  public List<Path> getPaths() {
+    List<Path> ret = new ArrayList<>();
+    if (path != null) {
+      ret.add(path);
+    }
+    return ret;
```
jt2594838: Use `Collections.singletonList`

server/src/main/java/org/apache/iotdb/db/qp/physical/sys/AddPathPlan.java
```diff
@@ -0,0 +1,111 @@
...
+    return props;
+  }
+
+  public void setProps(Map<String, String> props) {
+    this.props = props;
+  }
+  
+  @Override
+  public String toString() {
+    String ret = String.format("seriesPath: %s%nresultDataType: %s%nencoding: %s%nnamespace type:"
```
qiaojialin: refactor this...

PR #480, opened 2019-10-27
Update rpc and tsfile changelists

LeiRui:
Hi, I saw that the changelists of rpc and tsfile are out of date, so I spent some time organizing them. They become more clear to me now. Hope these docs can be maintained in the future as well.


samperson1997:
This is such a great effort! 🎉The change lists are quite organized much better than before! It must be really helpful for developers to see these clear and tidy change list for [TsFile](https://iotdb.apache.org/#/Development/Tsfile) and [RPC](https://iotdb.apache.org/#/Development/RPC) in our website page.

PR #501, opened 2019-10-31
[IOTDB-283] Modify rules of datatype inference

SolomonAnn:


session/src/test/java/org/apache/iotdb/session/IoTDBSessionIT.java
```diff
@@ -97,6 +100,28 @@ public void test()
...
+    session.createTimeseries("root.sg4.d1.s2", TSDataType.INT64, TSEncoding.RLE,
+        CompressionType.SNAPPY);
+    session.createTimeseries("root.sg4.d1.s3", TSDataType.INT64, TSEncoding.RLE,
+        CompressionType.SNAPPY);
+    insertRowBatchTest1("root.sg4.d1");
+
+    // do not set storage group and create timeseries
+    insertRowBatchTest1("root.sg5.d1");
+
+    conf.setAutoCreateSchemaEnabled(false);
```
jt2594838: I would suggest you move these tests elsewhere and give it a more proper name.

PR #506, opened 2019-11-01
[IOTDB-259] level query of path

HTHou:
when we don't know a specific timeseries path, we'd like to complete the optional path level by level(like the dir auto-complete in Linux), so we need the level query capcity of path.

For example, given there're two timeseries: root.a.s1.d1, root.a.s2.d1, we know the storage group root.a, but we are not sure the optional path of the next level. so we want the level query: give the prefix root.a, iotdb can return the already exist optional path of the next level : {s1,s2} , then we can get the new path root.a.s1

jira link: [https://issues.apache.org/jira/browse/IOTDB-259](https://issues.apache.org/jira/browse/IOTDB-259)


LeiRui:
Good feature, but I wonder how to write such a query. There is no user guide in the `docs/Documentation/UserGuide/4-Operation Manual/7-IoTDB Query Statement.md`.

LeiRui:
If you change the rpc.thrift, please remember to update the rpc-changelist.md under the service-rpc module.

HTHou:
> Good feature, but I wonder how to write such a query. There is no user guide in the `docs/Documentation/UserGuide/4-Operation Manual/7-IoTDB Query Statement.md`.

Since my another [pr](https://github.com/apache/incubator-iotdb/pull/500) about documents has not been merged yet, and that pr changes the name and the path of the document you mentioned. Therefore if I update doc here, it may cause a mess. I will add the statement into doc after that pr be merged. 

The usage I can show here. 
Screen Shot 2019-11-01 at 4 53 04 PM

HTHou:
> If you change the rpc.thrift, please remember to update the rpc-changelist.md under the service-rpc module.

Thank you for reminding me. I'll update rpc-changelist.md soon.

service-rpc/src/main/thrift/rpc.thrift
```diff
@@ -184,6 +184,7 @@ struct TSFetchMetadataResp{
 		8: optional set<string> devices
 		9: optional list<string> nodesList
 		10: optional map<string, string> nodeTimeseriesNum
+		11: optional set<string> childPaths
```
jt2594838: It is not much related to this PR, but seeing this, I think maybe we should break down this Resp into several smaller ones so that it is more clear what it responds to.

HTHou: > It is not much related to this PR, but seeing this, I think maybe we should break down this Resp into several smaller ones so that it is more clear what it responds to.

Did you mean we can separate `TSFetchMetadataResp` into some ones like `TSFetchDevicesResp`, `TSFetchNodeListResp`, etc? 
That's a good point.

jt2594838: Yes, besides you, other people are adding things into this structure, making it bigger and bigger. I don't think it is a good thing.

LeiRui:
> It looks good to me. But should we support paths with wildcards "*" in this query?

From my opinion, I think there is no need to support this.
By `show child paths <null>|<prefixPath>`, we know exactly where we are standing at and just want to know what paths exist in the next level.

PR #507, opened 2019-11-01
[IOTDB-282] add "show version"

Genius-pig:
Screen Shot 2019-11-01 at 3 49 01 PM


LeiRui:
Please update docs.

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBStatement.java
```diff
@@ -287,6 +288,10 @@ private boolean executeSQL(String sql) throws TException, SQLException {
         resultSet = databaseMetaData.getNodes(Constant.COUNT_NODES, path, null, null, level);
         return true;
       }
+    } else if(sqlToLowerCase.startsWith(SHOW_VERSION_COMMAND_LOWERCASE)) {
```
LeiRui: startWith? So if I query "show version xxxxxx", it can also run successfully. Is that right?

Genius-pig: done

LeiRui:
You answer the client's `show version` sql by refering to a constant string in jdbc:
```
static final String GLOBAL_DB_VERSION = "0.9.0-SNAPSHOT";
```
That way, your current realization is actually showing the version of jdbc, not the server.

Maybe you should develop `show client version` and `show server version` to handle the two different versions. Or just `show version` but return both the client's and the server's versions.

What do other guys think?

liutaohua:
i think the client version is show in shell ，like ./iotdb-client.sh -v ！
when you connect to server ，“show version” command is an sql，must return from server side

Genius-pig:
@LeiRui this is a server version. I have already changed it.

LeiRui:
> @LeiRui @liutaohua this is a server version. Returned JDBC version is before I force-push, I have already changed it. It actually uses
> 
> ```java
> IoTDBConstant.VERSION
> ```

Got it. Thanks for your hard work.
I still have some suggestions:
1. If you change rpc.thrift, remember to update the rpc-changelist.md under the service-rpc module.
2. What about just disabling the `getDatabaseProductVersion()` method by throw an exception? Since the now returned result `VERSION` is meaningless anyway and this method is not used anywhere.
3. You better add some tests in the `IoTDBMetadataFetchIT`.

Besides, I find that the server version will appear in the client shell:
image
So I wonder if the user specifically wants to get the version information using jdbc? We better clarify what the user really wants.

LeiRui:
By the way, let's can call it a day :)  You can reply me tomorrow.

Genius-pig:
1. done
2. done
3. done

Please check out.
I think users want to get the server version. @LeiRui

LeiRui:
> I think users want to get the server version. @LeiRui

Yes I think so. I am just a little confused since the server version already appears in the client shell. But this pr is ok anyway.

> 3. done

Good job. To be perfect, I expect you to add another test in the `IoTDBMetadataFetchIT` similar to `showTimeseriesTest1` and `showStorageGroupTest`, because if you have noticed, these two tests use `statement.execute(sql)` instead of `databaseMetaData.getColumns`.
By the way, please help rename `showTimeseriesTest1` to `showTimeseriesTest` for clarity.

liutaohua:
if somebody to write gui client ，how to show server version @LeiRui

LeiRui:
> if somebody to write gui client ，how to show server version @LeiRui

Yes, that is the user case of this pr.

Genius-pig:
> > I think users want to get the server version. @LeiRui
> 
> Yes I think so. I am just a little confused since the server version already appears in the client shell. But this pr is ok anyway.
> 
> > 1. done
> 
> Good job. To be perfect, I expect you to add another test in the `IoTDBMetadataFetchIT` similar to `showTimeseriesTest1` and `showStorageGroupTest`, because if you have noticed, these two tests use `statement.execute(sql)` instead of `databaseMetaData.getColumns`.
> By the way, please help rename `showTimeseriesTest1` to `showTimeseriesTest` for clarity.

Done

liukun4515:
why not add the commit id within version?

Genius-pig:
> why not add the commit id within version?

What is commit id and why add it?

docs/Documentation/UserGuide/5-Operation Manual/4-SQL Reference.md
```diff
@@ -30,6 +30,18 @@ In this part, we will introduce you IoTDB's Query Language. IoTDB offers you a S
...
+
+```sql
+show version
+```
+
+```
++---------------------------------------------------------------------------+
+|                                                             0.9.0-SNAPSHOT|
++---------------------------------------------------------------------------+
+It costs 00.176s
```
jt2594838: I wonder why there are two '0's.

server/src/test/java/org/apache/iotdb/db/integration/IoTDBMetadataFetchIT.java
```diff
@@ -201,6 +202,29 @@ public void databaseMetaDataTest() throws ClassNotFoundException, SQLException {
...
+        if(hasResultSet) {
+          try(ResultSet resultSet = statement.getResultSet()) {
+            ResultSetMetaData resultSetMetaData = resultSet.getMetaData();
+            Assert.assertEquals(resultSetMetaData.getColumnLabel(1), IoTDBConstant.VERSION);
+          }
+        }
+      } catch (Exception e) {
+        fail(e.getMessage());
+      } finally {
+        connection.close();
```
jt2594838: This call is unnecessary since you have put `connection` to try resource control.

docs/Documentation/UserGuide/5-Operation Manual/4-SQL Reference.md
Genius-pig: done

server/src/test/java/org/apache/iotdb/db/integration/IoTDBMetadataFetchIT.java
Genius-pig: You mean to delete `connect.close()`.

qiaojialin:
> why not add the commit id within version?

show version is for users, not developers.

Use case:
Suppose client-0.9.0 could connect both server-0.8.0 and server-0.9.0. And we add a sql in 0.9.0, users need to decide which sql they use based on the version.

server/src/test/java/org/apache/iotdb/db/integration/IoTDBMetadataFetchIT.java
jt2594838: Yes

PR #509, opened 2019-11-01
update python example to adapt to new QueryDataSet

jt2594838:


client-py/src/client_example.py
```diff
@@ -59,6 +59,54 @@
...
+    for i in range(row_count):
+        records.append([times[i]])
+
+    for i in range(len(dataTypeList)):
+        type = dataTypeList[i]
+        for j in range(row_count):
+            is_null = value_bytes[0]
+            value_bytes = value_bytes[1:]
+            if is_null == 1:
+                records[i].append('null')
```
LeiRui: Hi, this should be records[j], I think.

jt2594838: Fixed

PR #514, opened 2019-11-02
[IOTDB-287] Restrict users to only use domain names and IP addresses.

Genius-pig:
https://issues.apache.org/jira/browse/IOTDB-287
I suggest don't use `matcher.group()`.
Refer to https://stackoverflow.com/questions/36247902/java-regex-string-check-if-ipport-string-contains-a-valid-ipv4-or-dns-address


liukun4515:
please add unit test.

Genius-pig:
> please add unit test.

@liukun4515 add more unit test.

Genius-pig:
> Use a more precise title like "Allow domain name in IoTDB URL".
> And please check the documents to see if there is anything that should be changed.

`Pattern.compile("([^;]*):([^;]*)/");` The regular expression can be used to represent anything.  
Rather than allowing the use of domain names, What I change is more likely to restrict users to only use domain names and IP addresses. @jt2594838

jt2594838:
At least you should give what regex you are improving, and better together with what for.

PR #516, opened 2019-11-03
[IOTDB-288] Fix unclear error message notification for "creating timeseries" clause

samperson1997:
Fix unclear error message notification for "creating timeseries" clause.
As for the example in JIRA issue, current error for statement `CREATE TIMESERIES root.mi.d1.RANDOM_foo:BYTE WITH DATATYPE=INT32, ENCODING=RLE.` would be:
`Statement format is not right: parsing error,statement: CREATE TIMESERIES root.sg1.d1.RANDOM_foo:BYTE WITH DATATYPE=INT64, ENCODING=RLE .message:line 1:40 character ':' not supported here`


server/src/main/java/org/apache/iotdb/db/qp/QueryProcessor.java
```diff
@@ -111,7 +108,7 @@ private AstNode parseSQLToAST(String sqlStr) throws IllegalASTFormatException {
     } catch (ParseException e) {
       // e.printStackTrace();
```
jt2594838: Remove this.

PR #525, opened 2019-11-06
[IOTDB-260] Make paths can concatenate

Genius-pig:
https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-260
For example, you can use
```
  SELECT status FROM root.ln.wf01.{wt01,wt02}
```
Screen Shot 2019-11-06 at 1 35 45 PM

### DIFF
--- a/server/src/main/antlr3/org/apache/iotdb/db/sql/parse/TqlLexer.g
+++ b/server/src/main/antlr3/org/apache/iotdb/db/sql/parse/TqlLexer.g
@@ -257,6 +257,10 @@ K_PATH
     : P A T H
     ;
 
+K_SUBPATH
+    : S U B P A T H
+    ;
+
 K_LOAD
     : L O A D
     ;
--- a/server/src/main/antlr3/org/apache/iotdb/db/sql/parse/TqlLexer.g
+++ b/server/src/main/antlr3/org/apache/iotdb/db/sql/parse/TqlLexer.g
@@ -396,6 +396,14 @@ RS_BRACKET
     : ']'
     ;
 
+L_BRACKET
+    : '{'
+    ;
+
+R_BRACKET
+    : '}'
+    ;
+
 STRING_LITERAL
    : DQUOTA_STRING
    | SQUOTA_STRING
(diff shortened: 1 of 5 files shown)


qiaojialin:
what if the path contains *, could you handle it correctly?

Genius-pig:
> what if the path contains *, could you handle it correctly?
```
IoTDB> SELECT * FROM root.ln.wf01.{wt01,wt02}
SELECT * FROM root.ln.wf01.{wt01,wt02}
+-----------------------------------+-----------------------------+-----------------------------+-----------------------------+-----------------------------+
|                               Time|     root.ln.wf01.wt01.status|root.ln.wf01.wt01.temperature|     root.ln.wf01.wt02.status|root.ln.wf01.wt02.temperature|
+-----------------------------------+-----------------------------+-----------------------------+-----------------------------+-----------------------------+
|      1970-01-01T08:00:00.100+08:00|                         true|                         null|                         true|                         null|
|      1970-01-01T08:00:00.200+08:00|                        false|                        20.71|                        false|                        20.71|
+-----------------------------------+-----------------------------+-----------------------------+-----------------------------+-----------------------------+
Total line number = 2
It costs 0.096s
```

LeiRui:
Glad to see this pr :)

Two places you need to take care of:
1. Add docs!!!
2. Change the title of this pr to a more appropriate one.

Genius-pig:
> Glad to see this pr :)
> 
> Two places you need to take care of:
> 
> 1. Add docs!!!
> 2. Change the title of this pr to a more appropriate one.

What kind of title do you suggest? I don't think "Make paths can concatenate" is a bad title.

LeiRui:
> What kind of title do you suggest? I don't think "Make paths can concatenate" is a bad title.

 I still insist my opinion. There are many ways to make paths concatenate. For example, when executing `select s1,s2 from root.sg.d0`, the query executor will concatenate the prefix path `root.sg.d0` with the two suffix paths `s1`&`s2`. That's why I think `Make paths can concatenate" is not a good title.

If you need suggestions, I would say like "**support multiple timeseries selection using curly braces in the WHERE clause**".

PR #530, opened 2019-11-07
[IOTDB-295] Refactor db.exception

samperson1997:
Refactor `db.exception`:

* Extract uniform message into exception classes
* Create base exception `ProcessException`, which is extended by all exceptions
* Add different error codes to all exceptions. When an exception is caught and a higher-level exception is thrown, the error code will keep and pass (so that users will know the detailed error reason)
* Refactor packages under `db.exception`
exception


server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -425,15 +481,15 @@ public static TSDataType getSeriesType(String path)
...
+          throw new QueryProcessException(
               "aggregate does not support " + aggrType + " function.");
       }
     }
     return MManager.getInstance().getSeriesType(path);
   }
 
   protected List<String> getPaths(String path)
-      throws MetadataErrorException {
+      throws MetadataException {
```
HTHou: It is OK, but I just feel it's better to merge these two lines to one.

LeiRui:
Hi, I think you can think about sharing this refactored information somewhere in the docs. People would like it.

samperson1997:
> Hi, I think you can think about sharing this refactored information somewhere in the docs. People would like it.

Hi @LeiRui , thanks a lot for your reminding! I have updated the [related User Guide document chapter](https://github.com/apache/incubator-iotdb/blob/ec26f7f7bb3f157e25b5bf060c9b2da6a6e0b5fc/docs/Documentation/UserGuide/4-Client/2-Programming%20-%20JDBC.md) and added the refactor information in it (at the very end of this document).

samperson1997:
> It looks good to me except some very small and easy-fix issues. Please note there are also some conflicts to be resolved.

Hi @HTHou , thank you so much for your effort and detailed review! I have fixed the issues you have mentioned in the latest commit, and I'll keep an eye on format issues when solving conflicts in the future 😄...

LeiRui:
I wonder two things and I am not so sure:
1. In `IoTDBStatement.executeSQL`, there are many branches. I see that the last branch is equipped with StatusCode. Do the previous branches need that？
2. Does the client need to print the errorCode? As in the `AbstractClient.executeQuery`, when an exeception is caught, the current information printed is:
```
    } catch (Exception e) {
      println("Msg: " + e.getMessage());
      handleException(e);
    }
```

LeiRui:
> I wonder two things and I am not so sure:
> 
> 1. In `IoTDBStatement.executeSQL`, there are many branches. I see that the last branch is equipped with StatusCode. Do the previous branches need that？
> 2. Does the client need to print the errorCode? As in the `AbstractClient.executeQuery`, when an exeception is caught, the current information printed is:
> 
> ```
>     } catch (Exception e) {
>       println("Msg: " + e.getMessage());
>       handleException(e);
>     }
> ```

For the first question, I see that Tian Jiang has fixed and added StatusCode in the previous branches 😄

PR #543, opened 2019-11-11
[IOTDB-300]reconnection to requestStmtId when broke pip

liutaohua:
1. Start the IOTDB-Client and connect to the server successfully

2. Restart the server side

3. And then if you execute any command on the client, you will receive an error message: 'MSG: cannot get ID for statement'

I will refer to the *executeQuery()* method of reconnection for repair.


LeiRui:
I just encountered the same problem today and you solve it 👍

liutaohua:
It's so nice of you to comfort me 😘

PR #552, opened 2019-11-13
Delete users/*.profile

HTHou:
remove data/system/user/*.profile


server/src/main/java/org/apache/iotdb/db/service/IoTDB.java
```diff
@@ -88,7 +90,16 @@ private void setUp() throws StartupException {
     if (IoTDBDescriptor.getInstance().getConfig().isEnableStatMonitor()) {
       StatMonitor.getInstance().recovery();
     }
-
+    // Delete data/system/users/*.profile
+    File userDirectory = new File("server/target/iotdb-server-0.9.0/data/system/users");
```
qiaojialin:
```suggestion
    File usersDir = SystemFileFactory.INSTANCE.getFile(
        IoTDBDescriptor.getInstance().getConfig().getSystemDir() + File.separator + "users");
```

server/src/main/java/org/apache/iotdb/db/service/IoTDB.java
```diff
@@ -88,7 +91,17 @@ private void setUp() throws StartupException {
     if (IoTDBDescriptor.getInstance().getConfig().isEnableStatMonitor()) {
       StatMonitor.getInstance().recovery();
     }
-
+    // Delete data/system/users/*.profile
+    File usersDir = SystemFileFactory.INSTANCE.getFile(
+        IoTDBDescriptor.getInstance().getConfig().getSystemDir() + File.separator + "users");
```
samperson1997: It might be more convenient to use interface `File getFile(String parent, String child)`:
```suggestion
        IoTDBDescriptor.getInstance().getConfig().getSystemDir(), "users");
```

HTHou: Thanks for your suggestion. :)

qiaojialin:
fixed in  https://github.com/apache/incubator-iotdb/pull/555

PR #555, opened 2019-11-14
[IOTDB-303]fix user profile when online upgrading

EJTTianYu:


EJTTianYu:
> add a change_list of system file in server model

fixed

PR #559, opened 2019-11-15
[IOTDB-297]Fix show info bug

fanhualta:
When I execute "show dynamic parameters" and "show flush task info" commands, no useful information is output:

After fixing the bug, the result is as follow:
image


server/src/main/java/org/apache/iotdb/db/conf/IoTDBConstant.java
```diff
@@ -54,12 +54,10 @@ private IoTDBConstant() {
...
 
-  // for cluster, set read consistency level
-  public static final String SET_READ_CONSISTENCY_LEVEL_PATTERN = "set\\s+read.*level.*";
-
-  public static final String SHOW_FLUSH_TASK_INFO = "show\\s+flush\\s+task\\s+info";
-
-  public static final String SHOW_DYNAMIC_PARAMETERS = "show\\s+dynamic\\s+parameters";
+  // show info
+  public static final String ITEM = "                             item";
+  public static final String PARAMETER = "                                          parameter";
```
mdf369: Why put many spaces?

fanhualta: Just for the output format. If remove these spaces, its output is as follow:
image

mdf369:
approve

PR #560, opened 2019-11-15
[IOTDB-247] Optimize Dynamic Config with ActiveTimeSeriesCounter

liuruiyiyang:
Change List:
1. Add ```ActiveTimeSeriesCounter``` to count active time series number when flushing a meltable of each storage group, . The meltable flush threshold is depends on the dynamic active time series number instead of the created time series number, since some time series may not be written after created.
2. Fix the launch failure when enable StatMonitor.


server/src/main/java/org/apache/iotdb/db/conf/adapter/ActiveTimeSeriesCounter.java
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  @Override
+  public double getActiveRatio(String storageGroup) {
+    lock.writeLock().lock();
+    double ratio;
+    try {
+      ratio = activeRatioMap.get(storageGroup);
+    } catch (Exception e) {
+      LOGGER.error("Get {} active ratio failed", storageGroup, e);
```
fanhualta: Why do you try catch here? Is there any chance to occur an exception?

server/src/main/java/org/apache/iotdb/db/conf/adapter/ActiveTimeSeriesCounter.java
```diff
@@ -0,0 +1,140 @@
...
+      for (double number : activeTimeSeriesNumMap.values()) {
+        totalActiveTsNum += number;
+      }
+      for (Map.Entry<String, Long> entry : activeTimeSeriesNumMap.entrySet()) {
+        double activeRatio = 0;
+        if (totalActiveTsNum > 0) {
+          activeRatio = entry.getValue() / totalActiveTsNum;
+        }
+        activeRatioMap.put(entry.getKey(), activeRatio);
+        LOGGER.debug("{}: storage group {} has active ratio {}", Thread.currentThread().getName(),
```
fanhualta: `has active ratio `-> `has an active ratio:`

server/src/main/java/org/apache/iotdb/db/service/IoTDB.java
```diff
@@ -107,6 +100,13 @@ private void setUp() throws StartupException {
     registerManager.register(MetricsService.getInstance());
     JMXService.registerMBean(getInstance(), mbeanName);
 
+    // When registering statMonitor, we should start recovering some statistics
+    // with latest values stored
+    // Warn: registMonitor() method should be called after systemDataRecovery()
+    if (IoTDBDescriptor.getInstance().getConfig().isEnableStatMonitor()) {
+      StatMonitor.getInstance().recovery();
+    }
+
```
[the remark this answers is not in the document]
liuruiyiyang: To fix the launch failure when enable StatMonitor.

server/src/test/java/org/apache/iotdb/db/conf/adapter/HyperLogLogTest.java
```diff
@@ -0,0 +1,71 @@
...
+          card.offer(a);
+        }
+        double p = (card.cardinality() - set.size()) / (double) set.size();
+        errorSum += Math.abs(p);
+      }
+      // allow 1% error
+      assertEquals(0, errorSum / testNum, 0.01);
+      avgErrors[floorIndex] = String.format("%.5f", errorSum / testNum);
+    }
+    System.out.println(Arrays.toString(avgErrors));
```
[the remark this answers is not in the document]
liuruiyiyang: Fixed

server/src/test/java/org/apache/iotdb/db/conf/adapter/HyperLogLogTest.java
```diff
@@ -0,0 +1,71 @@
...
+    // output: [0.00299, 0.00914, 0.00331, 0.00294, 0.00348, 0.00450, 0.00605, 0.00581, 0.00580, 0.00585, 0.00668, 0.00672, 0.00748, 0.00753, 0.00742]
+    // time cost: 11s
```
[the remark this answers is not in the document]
liuruiyiyang: Fixed

server/src/main/java/org/apache/iotdb/db/conf/adapter/ActiveTimeSeriesCounter.java
```diff
@@ -0,0 +1,140 @@
...
+  public void init(String storageGroup) {
+    storageGroupHllMap.put(storageGroup, new HyperLogLog(LOG2M));
+    activeRatioMap.put(storageGroup, 0D);
+    activeTimeSeriesNumMap.put(storageGroup, 0L);
+  }
+
+  @Override
+  public void offer(String storageGroup, String device, String measurement) {
+    try {
+      storageGroupHllMap.get(storageGroup).offer(device + measurement);
```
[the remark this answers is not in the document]
liuruiyiyang: Fixed

server/src/main/java/org/apache/iotdb/db/conf/adapter/ActiveTimeSeriesCounter.java
liuruiyiyang: Fixed

server/src/main/java/org/apache/iotdb/db/conf/adapter/ActiveTimeSeriesCounter.java
```diff
@@ -0,0 +1,140 @@
...
+          measurement, e);
+    }
+  }
+
+  @Override
+  public void updateActiveRatio(String storageGroup) {
+    lock.writeLock().lock();
+    try {
+      // update the active time series number in the newest memtable to be flushed
+      activeTimeSeriesNumMap.put(storageGroup, storageGroupHllMap.get(storageGroup).cardinality());
```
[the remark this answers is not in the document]
liuruiyiyang: Modified.

server/src/main/java/org/apache/iotdb/db/conf/adapter/ActiveTimeSeriesCounter.java
liuruiyiyang: Just in case some unexpected test class exceptions, though I have fixed them for now.

server/src/main/java/org/apache/iotdb/db/conf/adapter/ActiveTimeSeriesCounter.java
```diff
@@ -59,11 +59,12 @@ public void init(String storageGroup) {
 
   @Override
   public void offer(String storageGroup, String device, String measurement) {
+    String path = device + "." + measurement;
```
fanhualta: It's better to replace "." with `IoTDBConstant.PATH_SEPARATOR`.

server/src/main/java/org/apache/iotdb/db/conf/adapter/ActiveTimeSeriesCounter.java
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  @Override
+  public double getActiveRatio(String storageGroup) {
+    lock.writeLock().lock();
+    double ratio;
+    try {
+      ratio = activeRatioMap.get(storageGroup);
+    } catch (Exception e) {
+      LOGGER.error("Get {} active ratio failed", storageGroup, e);
```
fanhualta: I don't think it's elegant if we just do `try-catch` in lots of places just in case `unexpected test class`.

liuruiyiyang: Fixed

server/src/main/java/org/apache/iotdb/db/conf/adapter/ActiveTimeSeriesCounter.java
liuruiyiyang: Fixed

LeiRui:
Your commit log is impressive 👍

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -646,6 +648,7 @@ public boolean deleteStorageGroupsFromMTree(List<Path> deletePathList)
           IoTDBConfigDynamicAdapter.getInstance().addOrDeleteStorageGroup(-1);
           mgraph.deleteStorageGroup(delStorageGroup);
           seriesNumberInStorageGroups.remove(delStorageGroup);
+          ActiveTimeSeriesCounter.getInstance().delete(delStorageGroup);
```
LeiRui: Do you need to updateActiveRatio here, or updateActiveRatio before the next use such as in `TsFileProcessor.getMemtableSizeThresholdBasedOnSeriesNum`?

server/src/test/java/org/apache/iotdb/db/conf/adapter/ActiveTimeSeriesCounterTest.java
```diff
@@ -0,0 +1,122 @@
...
+      this.sensorNum = sensorNum;
+      this.storageGroup = storageGroup;
+      this.finished = finished;
+    }
+
+    @Override
+    public void run() {
+      try {
+        for (int j = 0; j < sensorNum; j++) {
+          ActiveTimeSeriesCounter.getInstance().offer(storageGroup, "device_0", "sensor_" + j);
```
LeiRui: How about offering some duplicated entries for soundness?

server/src/main/java/org/apache/iotdb/db/conf/adapter/ActiveTimeSeriesCounter.java
```diff
@@ -0,0 +1,142 @@
...
+
+  @Override
+  public void offer(String storageGroup, String device, String measurement) {
+    String path = device + IoTDBConstant.PATH_SEPARATOR + measurement;
+    try {
+      storageGroupHllMap.get(storageGroup).offer(path);
+    } catch (Exception e) {
+      storageGroupHllMap.putIfAbsent(storageGroup, new HyperLogLog(LOG2M));
+      storageGroupHllMap.get(storageGroup).offer(path);
+      LOGGER.error("Register active time series root.{}.{} failed", storageGroup, path, e);
```
LeiRui: Note that a ususal case is  storageGoup = 'root.vehicle', device= 'root.vehicle.d0', measurement='s3'. So `root.{}.{}` may not apply to these situations.

server/src/main/java/org/apache/iotdb/db/conf/adapter/ActiveTimeSeriesCounter.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+  }
+
+  @Override
+  public void updateActiveRatio(String storageGroup) {
+    lock.writeLock().lock();
+    try {
+      long activeTimeSeriesNum = storageGroupHllMap.get(storageGroup).cardinality();
+      if (activeTimeSeriesNum != activeTimeSeriesNumMap.get(storageGroup)) {
```
LeiRui: It seems that if the latest `storageGroupHllMap.get(storageGroup).cardinality()` just happens to equal to `activeTimeSeriesNumMap.get(storageGroup)`, the `activeRatio` wouldn't be updated. Is that a bug?

Besides, the logic of the judgment condition - `activeTimeSeriesNum != activeTimeSeriesNumMap.get(storageGroup)` - looks a little bit wierd to me. If you want to judge whether the activeRatio of a storageGroup needs updating, you need to take other storageGroups into consideration. But the judgment condition here looks more like a decoration since the HLL counter is initialized every time in this function and thus the new counted result has a high probability of not equaling to `activeTimeSeriesNumMap.get(storageGroup)` (but still may equal hence the bug mentioned above, right?).

server/src/test/java/org/apache/iotdb/db/conf/adapter/ActiveTimeSeriesCounterTest.java
liuruiyiyang: The duplicated entries cases is included in the HyperLogLogTest class I think.

server/src/main/java/org/apache/iotdb/db/conf/adapter/ActiveTimeSeriesCounter.java
liuruiyiyang: I think there is a case that when  the memtable is going to be flushed after the storage group is deleted, it may cause the exception. And for any exception situations we may not aware currently or in the future, it is necessary to try catch here.

server/src/test/java/org/apache/iotdb/db/conf/adapter/ActiveTimeSeriesCounterTest.java
LeiRui: OK.

server/src/main/java/org/apache/iotdb/db/conf/adapter/ActiveTimeSeriesCounter.java
liuruiyiyang: If the activeTimeSeriesNum is not changed then the activeTimeSeriesNumMap is no need to be modified.

server/src/main/java/org/apache/iotdb/db/conf/adapter/ActiveTimeSeriesCounter.java
liuruiyiyang: fixed

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
liuruiyiyang: It would be unsafe for the memory control since the delete storage group's memtable may not be flushed yet.

PR #566, opened 2019-11-18
[IOTDB-263] Refactor IoTDBQueryResultSet of JDBC

HTHou:
Link: https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-263?filter=allopenissues

### DIFF
--- a/jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
+++ b/jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
@@ -19,11 +19,13 @@
 
 package org.apache.iotdb.jdbc;
 
+import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
 import java.math.BigDecimal;
 import java.math.MathContext;
 import java.net.URL;
+import java.nio.ByteBuffer;
 import java.sql.Array;
 import java.sql.Blob;
 import java.sql.Clob;
@@ -43,7 +45,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Calendar;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import org.apache.iotdb.rpc.IoTDBRPCException;
@@ -77,11 +78,13 @@ public class IoTDBQueryResultSet implements ResultSet {
   private Map<String, Integer> columnInfoMap; // used because the server returns deduplicated columns
   private List<String> columnTypeDeduplicatedList; // deduplicated from columnTypeList
   private RowRecord record;
-  private Iterator<RowRecord> recordItr;
   private int rowsFetched = 0;
   private int maxRows; // defined in TsfileStatement
   private int fetchSize;
   private boolean emptyResultSet = false;
+  
+  private TSQueryDataSet tsQueryDataSet = null;
+  private ByteBuffer byteBuffer = null;
 
   // 0 means it is not constrained in sql
   private int rowsLimit = 0;
@@ -406,7 +409,16 @@ public class IoTDBQueryResultSet implements ResultSet {
 
   @Override
   public double getDouble(String columnName) throws SQLException {
-    return Double.parseDouble(getValueByName(columnName));
+    checkRecord();
+    int index = columnInfoMap.get(columnName) - 2;
+    Field field = record.getFields().get(index);
+    if (field.getDataType() != null) {
+      return field.getDoubleV();
+    }
+    else {
+      throw new SQLException(
+          String.format("The value got by %s (column name) is NULL.", columnName));
+    }
   }
 
   @Override
@@ -436,7 +448,16 @@ public class IoTDBQueryResultSet implements ResultSet {
 
   @Override
   public float getFloat(String columnName) throws SQLException {
-    return Float.parseFloat(getValueByName(columnName));
+    checkRecord();
+    int index = columnInfoMap.get(columnName) - 2;
+    Field field = record.getFields().get(index);
+    if (field.getDataType() != null) {
+      return field.getFloatV();
+    }
+    else {
+      throw new SQLException(
+          String.format("The value got by %s (column name) is NULL.", columnName));
+    }
   }
 
   @Override
@@ -451,7 +472,16 @@ public class IoTDBQueryResultSet implements ResultSet {
 
   @Override
   public int getInt(String columnName) throws SQLException {
-    return Integer.parseInt(getValueByName(columnName));
+    checkRecord();
+    int index = columnInfoMap.get(columnName) - 2;
+    Field field = record.getFields().get(index);
+    if (field.getDataType() != null) {
+      return field.getIntV();
+    }
+    else {
+      throw new SQLException(
+          String.format("The value got by %s (column name) is NULL.", columnName));
+    }
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
```diff
@@ -1251,15 +1295,14 @@ private String getValueByName(String columnName) throws SQLException {
...
       return String.valueOf(record.getTimestamp());
     }
-    int tmp = columnInfoMap.get(columnName);
-    int i = 0;
-    for (Field field : record.getFields()) {
-      i++;
-      if (i == tmp - 1) {
-        return field.getDataType() == null ? null : field.getStringValue();
-      }
+    int index = columnInfoMap.get(columnName) - 2;
```
SilverNarcissus: This 2 has appeared several times, may be you can let it be a constant

HTHou:
> Thanks for your works! There are only 2 issues, please have a look. :D

The issues you mentioned have been fixed. Thank you so much for your feedback. :)

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
```diff
@@ -66,6 +67,7 @@
   private final String TIMESTAMP_STR = "Time";
```
SilverNarcissus: This should be static?

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
```diff
@@ -66,6 +67,7 @@
...
   private static final String limitStr = "LIMIT";
   private static final String offsetStr = "OFFSET";
+  private static final int firstColumnTypeIndex = 2;
```
SilverNarcissus: Constant naming should be like this :   LIMIT_STR

HTHou: Did you mean also change other constant names like `limitStr` and `offsetStr` to `LIMIT_STR` and `OFFSET_STR`?

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
HTHou: Fixed.

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
HTHou: > Constant naming should be like this : LIMIT_STR

Thank you. Fixed

jdbc/src/test/java/org/apache/iotdb/jdbc/IoTDBQueryResultSetTest.java
```diff
@@ -229,7 +229,7 @@ public void testQuery() throws Exception {
               + "103,null,199,null,null,\n"
               + "105,11.11,199,33333,11.11,\n"
               + "1000,1000.11,55555,22222,1000.11,\n"; // Note the LIMIT&OFFSET clause takes effect
-      Assert.assertEquals(resultStr.toString(), standard);
+      Assert.assertEquals(standard, resultStr.toString());
```
LeiRui: careful 👍

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
```diff
@@ -315,18 +319,16 @@ public boolean getBoolean(int columnIndex) throws SQLException {
 
   @Override
   public boolean getBoolean(String columnName) throws SQLException {
-    String b = getValueByName(columnName);
-    if (b == null) {
+    checkRecord();
+    int index = columnInfoMap.get(columnName) - START_INDEX;
+    Field field = record.getFields().get(index);
+    if (field.getDataType() != null) {
```
LeiRui: Is it necessary to check whether the field.gerDataType equals boolean?
If the datatype is not boolean, I guess the field.getBoolV is null too.

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
```diff
@@ -1251,15 +1294,14 @@ private String getValueByName(String columnName) throws SQLException {
...
     }
-    int tmp = columnInfoMap.get(columnName);
-    int i = 0;
-    for (Field field : record.getFields()) {
-      i++;
-      if (i == tmp - 1) {
-        return field.getDataType() == null ? null : field.getStringValue();
-      }
+    int index = columnInfoMap.get(columnName) - START_INDEX;
+    Field field = record.getFields().get(index);
```
LeiRui: What if the index is out of bound? 
In the old code, no `i == tmp - 1` so null is returned.
Besides, will field ever be null?

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
```diff
@@ -1251,15 +1294,14 @@ private String getValueByName(String columnName) throws SQLException {
...
+    if (field != null) {
+      return field.getStringValue();
```
LeiRui: In the old code, if `field.getDataType` equals null, then `null` is returned instead of a String valued "null". I think there is difference between a null String and a String whose value is "null", although the printed result is the same "null".

jdbc/src/main/java/org/apache/iotdb/jdbc/Utils.java
```diff
@@ -83,21 +86,95 @@ static IoTDBConnectionParams parseUrl(String url, Properties info)
...
+        boolean is_empty = BytesUtils.byteToBool(byteBuffer.get());
+        if (is_empty) {
+          dataOutputStreams[i].writeBoolean(true);
+        } else {
+          dataOutputStreams[i].writeBoolean(false);
+          switch (dataType) {
+            case BOOLEAN:
+              boolean booleanValue = BytesUtils.byteToBool(byteBuffer.get());
+              dataOutputStreams[i].writeBoolean(booleanValue);
+              valueOccupation += 1;
```
LeiRui: So you convert the `column-wise` buffer to a `row-wise` buffer to replace the `List<RowRecord>` structure, and then get a rowRecord from the `row-wise` buffer each time.

I see that `List<RowRecord>` is replaced by a more compact structure.
But it still looks a little wierd to me. A value derived from the buffer is then written into buffer again... Do you have any opinions @jt2594838?

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
HTHou: > Is it necessary to check whether the field.getDataType equals boolean?
> If the datatype is not boolean, I guess the field.getBoolV is null too.

Yes, you are right. It would be better to check whether the field.getDataType is the correct datatype.

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
HTHou: > What if the index is out of bound? In the old code, if no `i == tmp - 1`, `null` is returned.
> Besides, will field ever possible be assigned null in this statement?
The point you mentioned is exactly a problem that need be fixed. Thanks.

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
HTHou: Fixed. Thank you.

jdbc/src/main/java/org/apache/iotdb/jdbc/Utils.java
HTHou: Yes, you are talking about what I am thinking. The buffer fetched from server is `column-wise`. If we can construct the `row-wise` buffer in server, we won't need to convert the buffer again.

LeiRui: Good idea! 
I look back on [my development document](https://tower.im/teams/138444/todos/7987/) of `column-wise` buffer and can't find any strong reason against a `row-wise` alternative.

PR #577, opened 2019-11-22
[IOTDB-23] Add Internal Monitor Metrics and Code Refactor

liuruiyiyang:
Change List:
1. Add some internal monitor metrics
```
+---------------------------------------------------------------------------+---------------------------------------------+--------+--------+
|                                                                 Timeseries|                                Storage Group|DataType|Encoding|
+---------------------------------------------------------------------------+---------------------------------------------+--------+--------+
|                                      root.MONITOR.write.global.FAIL_POINTS|                                 root.MONITOR|   INT64|     RLE|
|                                        root.MONITOR.write.global.OK_POINTS|                                 root.MONITOR|   INT64|     RLE|
|                                             root.MONITOR.request.TOTAL_REQ|                                 root.MONITOR|   INT64|     RLE|
|                                                  root.MONITOR.fileSize.WAL|                                 root.MONITOR|   INT64|     RLE|
|                                                 root.MONITOR.fileSize.DATA|                                 root.MONITOR|   INT64|     RLE|
|                                               root.MONITOR.system.FREE_MEM|                                 root.MONITOR|   INT64|     RLE|
|                                      root.MONITOR.system.FREE_PHYSICAL_MEM|                                 root.MONITOR|   INT64|     RLE|
|                                      root.MONITOR.system.USED_PHYSICAL_MEM|                                 root.MONITOR|   INT64|     RLE|
|                                              root.MONITOR.system.CPU_USAGE|                                 root.MONITOR|   INT64|     RLE|
+---------------------------------------------------------------------------+---------------------------------------------+--------+--------+
```
which includes the insert point number, request received number, WAL and sequence TsFile size, JVM and System physical memory size, etc.

2. Remove some useless code and do some refactor.


jixuan1989:
1. why you monitor the `sequence` folder but no `unsequence` folder?
2.   What is the difference between memory and physical memory? and I think the following are not required:

root.MONITOR.system.TOTAL_MEM
root.MONITOR.system.TOTAL_PHYSICAL_MEM
root.MONITOR.system.MAX_MEM

liuruiyiyang:
1. I will monitor the data dir size which includes both ```sequence``` and ```unsequence``` files, is that OK?
2. The previous is for JVM and the latter is for the server node. I will consider remove some not-frequently-changed metrics.

qiaojialin:
hi, how to use this?

liuruiyiyang:
@qiaojialin 
hi, to enable the Internal Monitor, set ```enable_stat_monitor=true``` in the ```iotdb-engine.properties``` file before starting the IoTDB server service. Then the above metrics/time series will be created automatically and you can control the monitor data frequency and the data duration by setting some related parameters of the ```Statistics Monitor configuration``` section in the ```iotdb-engine.properties``` file.

liutaohua:
Is dynamic configuration supported?

liuruiyiyang:
@liutaohua 
You mean setting ```enable_stat_monitor``` to enable or disable this function after starting the server service? I don't think we can do that currently.

server/src/main/java/org/apache/iotdb/db/engine/StorageEngine.java
```diff
@@ -437,4 +468,55 @@ public void deleteTsfile(File deletedTsfile) throws StorageEngineException {
...
+        .convertToTSRecord(getStatParamsHashMap(), MONITOR_METRIC_PREFIX);
+    HashMap<String, TSRecord> ret = new HashMap<>();
+    ret.put(MONITOR_METRIC_PREFIX, tsRecord);
+    return ret;
+  }
+
+  @Override
+  public void registerStatMetadata() {
+    Map<String, String> hashMap = new HashMap<>();
+    for (StorageEngineMetrics kind : StorageEngineMetrics.values()) {
```
qiaojialin:
```suggestion
    for (StorageEngineMetrics type : StorageEngineMetrics.values()) {
```

server/src/main/java/org/apache/iotdb/db/engine/StorageEngine.java
```diff
@@ -437,4 +468,55 @@ public void deleteTsfile(File deletedTsfile) throws StorageEngineException {
...
+      String seriesPath = MONITOR_METRIC_PREFIX
+          + IoTDBConstant.PATH_SEPARATOR
+          + kind.name();
+      hashMap.put(seriesPath, MonitorConstants.DATA_TYPE_INT64);
+      Path path = new Path(seriesPath);
+      try {
+        addTimeSeries(path, TSDataType.valueOf(MonitorConstants.DATA_TYPE_INT64),
```
qiaojialin: why not use TSDatatype.INT64 directly?

server/src/main/java/org/apache/iotdb/db/engine/StorageEngine.java
```diff
@@ -437,4 +468,55 @@ public void deleteTsfile(File deletedTsfile) throws StorageEngineException {
...
+            TSEncoding.valueOf("RLE"), CompressionType.valueOf(
```
qiaojialin:
```suggestion
            TSEncoding.RLE, CompressionType.valueOf(
```

server/src/main/java/org/apache/iotdb/db/monitor/MonitorConstants.java
```diff
@@ -20,90 +20,70 @@
...
-import java.util.concurrent.atomic.AtomicLong;
 import org.apache.iotdb.db.conf.IoTDBConfig;
+import org.apache.iotdb.db.conf.IoTDBConstant;
 import org.apache.iotdb.db.conf.IoTDBDescriptor;
-import org.apache.iotdb.db.monitor.collector.FileSize;
 
 public class MonitorConstants {
 
   private static IoTDBConfig config = IoTDBDescriptor.getInstance().getConfig();
   public static final String DATA_TYPE_INT64 = "INT64";
```
qiaojialin: remove this, use TSDataType.INT64 instead

jixuan1989:
discuss:

To save memory and get better compression ( and the write speed), how about put all monitor-series into one device?
 
If so,
1. In the memory, only one ChunkGroup is needed, which makes the memory more efficient;
2. Less ChunkGroupMetadata is generated;
3. You can use one insert(InsertPlan) rather than a for-loop.

@qiaojialin , any suggestions?

PR #605, opened 2019-11-30
Fix Travis CI for WinOS + OpenJDK11

jixuan1989:
As `choco install` is quite unstable, I give up using choco finally.
Now, On WinOS, I download JDK from java.net.
Hope it can work well  for a long long time!

This PR needs `squash and merge`


liutaohua:
if work a long long time , we need a simple file server

PR #609, opened 2019-12-01
[IOTDB-306] new chunkReader for avg count

liutaohua:
In the old interface, data filtering could not be use the chunk level, which I think may not have been considered in the previous design. Now the IAggregateReader interface has been added `hasNextChunk` and `nextChunkMeta`.

The advantage of this is that `Statistics` can be used in the `AggrFunction`, there is no need to read `page`


jixuan1989:
aha?
1. why comment plugins in pom?
2. working PR? for what ?

server/src/main/java/org/apache/iotdb/db/query/aggregation/AggregateFunction.java
```diff
@@ -119,4 +120,7 @@ public abstract void calcAggregationUsingTimestamps(long[] timestamps, int lengt
   public TSDataType getResultDataType() {
     return resultDataType;
   }
+
+  public abstract void calculateValueFromChunkData(
```
JackieTien97: It's better to change the function name to 'calculateValueFromChunkMetaData'

server/src/main/java/org/apache/iotdb/db/query/aggregation/impl/FirstValueAggrFunc.java
```diff
@@ -151,4 +152,18 @@ public void calcAggregationUsingTimestamps(long[] timestamps, int length,
...
+      throws QueryProcessException {
+    if (resultData.isSetTime()) {
+      return;
+    }
+
+    Object firstVal = chunkMetaData.getStatistics().getFirstValue();
+    if (firstVal == null) {
+      throw new QueryProcessException("chunkMetaData contains no FIRST value");
+    }
+    resultData.putTimeAndValue(0, firstVal);
```
JackieTien97: Why here just set the timestamp to 0? You can get the time through chunkMetaData.getStatistics().getStartTime().

tsfile/src/main/java/org/apache/iotdb/tsfile/read/reader/series/FileSeriesReader.java
```diff
@@ -54,23 +55,13 @@ public FileSeriesReader(IChunkLoader chunkLoader, List<ChunkMetaData> chunkMetaD
...
-      if (chunkSatisfied(chunkMetaData)) {
-        // chunk metadata satisfy the condition
-        initChunkReader(chunkMetaData);
-
-        if (chunkReader.hasNextBatch()) {
-          return true;
-        }
+    if (hasNextChunk()) {
+      if (chunkReader.hasNextBatch()) {
+        return true;
```
JackieTien97: I think here you should use while(hasNextChunk()) instead of if(hasNextChunk).
Because, even if the current chunkReader doesn't have the needed batch, it doesn't mean that next Chunk won't have.

server/src/main/java/org/apache/iotdb/db/query/reader/universal/IterateReader.java
```diff
@@ -42,19 +43,13 @@ public IterateReader(int readerSize) {
...
-    while (nextSeriesReaderIndex < readerSize) {
-      boolean isConstructed = constructNextReader(nextSeriesReaderIndex++);
-      if (isConstructed && currentSeriesReader.hasNext()) {
-        curReaderInitialized = true;
-        return true;
-      }
+    curReaderInitialized = false;
+    if (hasNextChunk()) {
+      curReaderInitialized = true;
+      return currentSeriesReader.hasNext();
```
JackieTien97: Same above. The if statement should be while statement.
And as the former logic, you can set the curReaderInitialized to true only if currentSeriesReader.hasNext() return true.

PR #610, opened 2019-12-01
add a assembly.xml to define which files will be packed into source-release.zip

jixuan1989:
Considering we packed some incorrect files in the last release (RC1 to RC4), I think writing an assembly.xml file for generating the source-release.zip is a good way to avoid that.

What is more, it is very helpful when we switch different branches for releasing. for example, in v0.9.0, there is a module called session, which does not exist in 0.8.0. So, if I switch the branch from 0.9 to 0.8 (or rel/0.8) for a minor version releasing (e.g., 0.8.2), I have to remove the session folder manually. If I forgot that, a dirty source-release.zip is generated. But if we have an assembly.xml,  it will avoid that.

I have written the source-assembly.xml now. The assembly file looks ugly because we have to declare the files one by one. (I tried the <moduleSet> tag, which is more convenient. But because our module.artifactId != module.folder name, there is some issues that I can not solve. Finally I gave up using <moduleSet> tag).

If the PR is accepted, then for all contributors please notice that:

(1) If you add new source files into a src/ docs/ license folder, it is ok.

(2) If you add a new module,  create a new folder, or write a plain text file under the project's (or a module's) root directory, you have to maintain the source-assembly.xml file.. Otherwise the new added files will not be packed into source-release.zip.

Hope in this way, we can no longer get  dirty source-release.zip files.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -920,6 +920,9 @@
                                     See  https://issues.apache.org/jira/browse/MNG-5454  sigh.
                                  -->
                                 <configuration combine.self="append">
+                                    <descriptors>
+                                        <descriptor>source-assembly.xml</descriptor>
+                                    </descriptors>
                                     <finalName>apache-iotdb-${project.version}-incubating</finalName>
                                 </configuration>
                             </execution>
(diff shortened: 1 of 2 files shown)


jixuan1989:
I will close this because using `mvn release:perform` can get a clean packed file........
So sad that I spend about 6 hours to learn the plugin and finish the file :( But also happy that do not need to maintain such a file in the project.

PR #614, opened 2019-12-03
[IOTDB-306] new reader with ChunkMetaData for aggregateFunction

liutaohua:
In the old interface, data filtering could not be use the chunk level, which I think may not have been considered in the previous design. Now the IAggregateReader interface has been added `hasNextChunk`  `nextChunkMeta` `hasNextPageInCurrentChunk`.

The advantage of this is that Statistics can be used in the AggrFunction, there is no need to read page

This some test, **i restart the server after each kind of test**:

1. 10,000 data without chunkMeta and unseq
IoTDB> `select count(*) from root`
It costs 0.290s
IoTDB> `select first_value(*) from root`
It costs 0.657s
IoTDB> `select last_value(*) from root`
It costs 0.156s
IoTDB> `select sum(*) from root`
It costs 0.224s
IoTDB> `select avg(*) from root`
It costs 0.272s
IoTDB> `select max_time(*) from root`
It costs 0.117s

1. 10,000 data with chunkMeta but without unseq
IoTDB>  `select count(*) from root`
It costs 0.006s
IoTDB> `select first_value(*) from root`
It costs 0.006s
IoTDB> `select sum(*) from root`
It costs 0.005s
IoTDB> `select avg(*) from root`
It costs 0.006s
IoTDB> `select max_time(*) from root`
It costs 0.004s


server/src/main/java/org/apache/iotdb/db/query/executor/AggregateEngineExecutor.java
```diff
@@ -212,40 +214,49 @@ private AggreResultData handleLastMaxTimeWithOutTimeGenerator(AggregateFunction
...
-        // cal by pageData
-        BatchData batchData = sequenceReader.nextBatch();
-        if (batchData.length() > 0) {
-          if (lastBatchTimeStamp > batchData.currentTime()) {
+    while (sequenceReader.hasNextChunk()) {
+      ChunkMetaData chunkMetaData = sequenceReader.nextChunkMeta();
+      if (chunkMetaData != null && canUseHeader(function, chunkMetaData.getStartTime(),
+          chunkMetaData.getEndTime(), unSequenceReader, timeFilter)) {
+        function.calculateValueFromChunkMetaData(chunkMetaData);
+        continue;
```
LeiRui: The `if (isChunkEnd) {break;}` logic is missing here. You could have broken out of the loop early.

liutaohua: I have finished the modification, but many `else` have been deleted, please review again

server/src/main/java/org/apache/iotdb/db/query/executor/AggregateEngineExecutor.java
```diff
@@ -148,23 +149,32 @@ private AggreResultData aggregateWithoutValueFilter(AggregateFunction function,
...
-        sequenceReader.skipPageData();
-      } else {
-        // cal by pageData
-        function.calculateValueFromPageData(sequenceReader.nextBatch(), unSequenceReader);
+    while (sequenceReader.hasNextChunk()) {
+      ChunkMetaData chunkMetaData = sequenceReader.nextChunkMeta();
+      if (chunkMetaData != null && canUseHeader(function, chunkMetaData.getStartTime(),
+          chunkMetaData.getEndTime(), unSequenceReader, filter)) {
+        function.calculateValueFromChunkMetaData(chunkMetaData);
+        continue;
```
LeiRui: Same problem here. 
`continue` makes all the early break out of "first_value" or "min_time" aggregate query impossible.

LeiRui:
In short, the performance of `aggregate query without value filters` is optimized by enabling sequence data skipping at the chunk level whenever possible. To sum up the updated facilities:
- Aggregate Readers
  - `IAggregateReader`: add three interface methods
    - `hasNextChunk`
    - `nextChunkMeta`
    - `hasNextPageInCurrentChunk`
  - Readers which implement `IAggregateReader` are thus updated:
    - `MemChunkReader`
    - `FileSeriesReaderAdapter`
    - `IterateReader`
  - `FileSeriesReader` in the TsFile module is updated.
- Aggregate Functions
  - `AggregateFunction`: add an abstract method `calculateValueFromChunkMetaData`
  - Functions which extend `AggregateFunction` are thus updated:
    - `AvgAggrFunc`
    - `CountAggrFunc`
    - `MaxValueAggrFunc`
    - `MinValueAggrFunc`
    - `FirstValueAggrFunc`
    - `LastValueAggrFunc`
    - `MaxTimeAggrFunc`
    - `MinTimeAggrFunc`
- Aggregate Engine Executor
  - `AggregateEngineExecutor.aggregateWithoutValueFilter` method is modified
  - `AggregateEngineExecutor.handleLastMaxTimeWithOutTimeGenerator` method is modified

It looks good to me. Thank @liutaohua for your hard work.

PR #616, opened 2019-12-03
remove some unnecessary codes

qiaojialin:


server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -803,32 +803,23 @@ private TSExecuteStatementResp executeAuthQuery(PhysicalPlan plan) {
...
-    }
-
-    // check file level set
-    try {
-      checkFileLevelSet(paths);
-    } catch (StorageGroupException e) {
-      logger.error("meet error while checking file level.", e);
-      return getTSExecuteStatementResp(
-          getStatus(TSStatusCode.CHECK_FILE_LEVEL_ERROR, e.getMessage()));
-    }
```
samperson1997: Should `CHECK_FILE_LEVEL_ERROR` be removed in `TSStatusCode` and [related User Guide document](https://iotdb.apache.org/#/Documents/progress/chap4/sec2) too? (The error code number is 304)

qiaojialin: yes, I removed it

PR #619, opened 2019-12-03
[IOTDB-330] Improve the reading method of unsequence data

samperson1997:
Improve the reading method of unsequence data:
* Make `UnseqResourceMergeReader` implement `IBatchReader`, and use `PriorityMergeReader` as a variable instead of extending it.
* Make `PriorityMergeReader` not implement `IPointReader`, which returns Element directly instead of returning TimeValuePair
* Split TimeValuePair in `PriorityMergeReader` heap element into time and value


server/src/main/java/org/apache/iotdb/db/query/reader/resourceRelated/UnseqResourceMergeReader.java
```diff
@@ -42,136 +42,133 @@
...
+
+      // create and add DiskChunkReader
+      while (priorityMergeReader.current().getTime() >= nextMetaDataStartTime) {
+        diskChunkReaderWrap = new ChunkReaderWrap(metaData, metaData.getChunkLoader(), timeFilter);
+        priorityMergeReader
+            .addReaderWithPriority(diskChunkReaderWrap.getIPointReader(), metaData.getPriority());
+        index++;
+        if (index < metaDataList.size()) {
+          metaData = metaDataList.get(index);
+          nextMetaDataStartTime = metaData.getStartTime();
```
LeiRui: What if here `index = metaDataList.size()` and the last updated `nextMetaDataStartTime` <= `priorityMergeReader.current()`, will the `while (priorityMergeReader.current().getTime() >= nextMetaDataStartTime)` loop forever?

samperson1997: Hi, thanks for your reminding! I have modified the code logic in a better way.

server/src/main/java/org/apache/iotdb/db/query/reader/universal/PriorityMergeReader.java
```diff
@@ -23,143 +23,137 @@
...
     while (!heap.isEmpty()) {
       Element e = heap.poll();
       e.close();
     }
   }
 
-  class Element {
+  public class Element {
 
     IPointReader reader;
```
LeiRui: From my current understanding, there is a future TODO:
 `ChunkReader` is going to implement only the `IBatchReader` operator. In other words, there will be no more `IPointReader` or related methods such as ` TimeValuePair timeValuePair = reader.next();`.

LeiRui: The future TODO is out of the scope of this pr. This pr is good.

server/src/main/java/org/apache/iotdb/db/query/reader/resourceRelated/UnseqResourceMergeReader.java
```diff
@@ -42,136 +42,130 @@
...
+      metaData = metaDataList.get(index);
+      // create and add DiskChunkReader
+      while (index < metaDataList.size()) {
+        diskChunkReaderWrap = new ChunkReaderWrap(metaData, metaData.getChunkLoader(), timeFilter);
+        priorityMergeReader
+            .addReaderWithPriority(diskChunkReaderWrap.getIPointReader(), metaData.getPriority());
+        metaData = metaDataList.get(index++);
+        if (priorityMergeReader.current().getTime() < metaData.getStartTime()) {
+          break;
+        }
```
LeiRui: Thanks for your quick reaction. 
But ... the modified logic senses more strange to me, especially the two places:
1) The `index++` logic makes the last `metaData` of `metaDataList` got at line 133 can never be added when its `getStartTime() <= priorityMergeReader.current().getTime()`, because the program will exit in the next loop at line 129.
2) the `addReaderWithPriority` is directly used before applying the `if (priorityMergeReader.current().getTime() < metaData.getStartTime())` check.

You can modify or discuss about this later. Take your time :)

server/src/main/java/org/apache/iotdb/db/query/reader/universal/PriorityMergeReader.java
samperson1997: Yes, I intend to make `DiskChunkReader` and `MemChunkReader` only implement `IBatchReader` in next step.

server/src/main/java/org/apache/iotdb/db/query/reader/resourceRelated/UnseqResourceMergeReader.java
samperson1997: Sorry for my fault... I have changed the order, how do you think about the codes now? 😂

PR #620, opened 2019-12-03
[IOTDB-337]Add timestamp precision properties for grafana

fanhualta:
In the grafana module, timestamps are in milliseconds. In this case, when users choose "us" or "ns" in engine, it'll not work correctly. This pr is to fix this bug.

The jira link: https://issues.apache.org/jira/browse/IOTDB-337

### DIFF
--- a/grafana/src/main/java/org/apache/iotdb/web/grafana/dao/impl/BasicDaoImpl.java
+++ b/grafana/src/main/java/org/apache/iotdb/web/grafana/dao/impl/BasicDaoImpl.java
@@ -18,6 +18,10 @@
  */
 package org.apache.iotdb.web.grafana.dao.impl;
 
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
 import java.sql.Connection;
 import java.sql.DatabaseMetaData;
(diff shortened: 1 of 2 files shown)


grafana/src/main/java/org/apache/iotdb/web/grafana/dao/impl/BasicDaoImpl.java
```diff
@@ -45,19 +50,43 @@
...
+  private static final String CONFIG_PROPERTY_FILE = "application.properties";
+
   private final JdbcTemplate jdbcTemplate;
 
+  private static long TIMESTAMP_RADIX = 1L;
+
   @Autowired
   public BasicDaoImpl(JdbcTemplate jdbcTemplate) {
     this.jdbcTemplate = jdbcTemplate;
+    try (InputStream inputStream = new FileInputStream(new File(CONFIG_PROPERTY_FILE))) {
```
Genius-pig: Why use properties like that? https://www.baeldung.com/properties-with-spring

fanhualta: I think this way is ok for this issue.

PR #624, opened 2019-12-04
[IOTDB-287] Allow anything in a JDBC URL.

Genius-pig:
In this [pr](https://github.com/apache/incubator-iotdb/pull/514l), it restricts users to only use domain names and IP addresses. Obviously, this makes users can’t use a custom domain name like `master`. Now, you can use a JDBC URL like `jdbc:iotdb://anything:port/ or jdbc:iotdb://anything:port`.

### DIFF
--- a/jdbc/src/main/java/org/apache/iotdb/jdbc/Utils.java
+++ b/jdbc/src/main/java/org/apache/iotdb/jdbc/Utils.java
@@ -51,31 +51,17 @@ public class Utils {
       return params;
     }
     boolean isUrlLegal = false;
-    Pattern pattern = Pattern.compile("^"
-        + "(((?!-)[A-Za-z0-9-]{1,63}(?<!-)\\.)+[A-Za-z]{2,6}" // Domain name
-        + "|"
-        + "localhost" // localhost
-        + "|"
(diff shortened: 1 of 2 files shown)


jdbc/src/main/java/org/apache/iotdb/jdbc/Utils.java
```diff
@@ -51,31 +51,17 @@ static IoTDBConnectionParams parseUrl(String url, Properties info)
...
-    } else {
-      DomainAndPort = subURL.split(":");
-    }
-    params.setHost(DomainAndPort[0]);
-    params.setPort(Integer.parseInt(DomainAndPort[1]));
     if (!isUrlLegal) {
-      throw new IoTDBURLException("Error url format, url should be jdbc:iotdb://domain|ip:port/ or jdbc:iotdb://domain|ip:port");
+      throw new IoTDBURLException("Error url format, url should be jdbc:iotdb://anything:port/ or jdbc:iotdb://anything:port");
     }
+    params.setHost(matcher.group(1));
```
jixuan1989: have you tried if the url is not matched?

Genius-pig: Add one test of `jdbc:iotdb//test6667`.

PR #626, opened 2019-12-04
Change the return content in TSQueryDataSet and remove the row record in IoTDBQueryResultSet

JackieTien97:
There is no need to construct a row record every time the next() function is called, we can extract the value from the corresponding value byte buffer.
Before, we use a byte size bool to indicate whether it is a null value, but the null values are rare, so the space is wasted.
Now, I use a bitmap to indicate whether it is a null value, in this way, each value only consumes extra one bit size instead of one byte size.


jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
```diff
@@ -716,10 +689,75 @@ public boolean next() throws SQLException {
...
+
+      record[i+1] = null;
+      // the column value is not null, we should move the corresponding value buffer to next position
+      if (!isNull(i, rowsIndex)) {
+        ByteBuffer valueBuffer = tsQueryDataSet.valueList.get(i);
+        TSDataType dataType = TSDataType.valueOf(columnTypeList.get(i));
+        switch (dataType) {
+          case BOOLEAN:
+            record[i+1] = new byte[1];
+            valueBuffer.get(record[i+1]);
```
qiaojialin: format the code using google-style

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
```diff
@@ -76,12 +52,13 @@
...
   private Map<String, Integer> columnInfoMap; // used because the server returns deduplicated columns
   private List<String> columnTypeDeduplicatedList; // deduplicated from columnTypeList
-  private RowRecord record;
+  private int rowsIndex = 0;
   private int fetchSize;
   private boolean emptyResultSet = false;
 
   private TSQueryDataSet tsQueryDataSet = null;
-  private ByteBuffer byteBuffer = null;
+  private byte[][] record;
```
qiaojialin: add some javadoc for this field..

You could separate the time bytes from this record. This record could rename to values

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
```diff
@@ -716,10 +689,75 @@ public boolean next() throws SQLException {
     if (emptyResultSet) {
       return false;
     }
-    record = Utils.getRowRecord(byteBuffer, columnTypeDeduplicatedList);
+    constructOneRow();
     return true;
   }
 
+  private void constructOneRow() {
+    record[0] = new byte[Long.BYTES];
```
qiaojialin: for the timestamp column, the record[0] could be created by a new byte[Long.BYTES] and reused. No need to new a byte[Long.BYTES] each time.

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
```diff
@@ -716,10 +689,75 @@ public boolean next() throws SQLException {
...
+          case INT64:
+            record[i+1] = new byte[Long.BYTES];
+            valueBuffer.get(record[i+1]);
+            break;
+          case FLOAT:
+            record[i+1] = new byte[Float.BYTES];
+            valueBuffer.get(record[i+1]);
+            break;
+          case DOUBLE:
+            record[i+1] = new byte[Double.BYTES];
```
qiaojialin: also, all non-text fields could be treated as the same with the timestamp column

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
```diff
@@ -716,10 +689,75 @@ public boolean next() throws SQLException {
...
+            break;
+          case TEXT:
+            int length = valueBuffer.getInt();
+            record[i+1] = ReadWriteIOUtils.readBytes(valueBuffer, length);
+            break;
+          default:
+            throw new UnSupportedDataTypeException(
+                    String.format("Data type %s is not supported.", columnTypeList.get(i)));
+        }
+      } else {
```
qiaojialin: remove this

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
```diff
@@ -1172,7 +1210,7 @@ public boolean wasNull() throws SQLException {
   }
 
   private void checkRecord() throws SQLException {
-    if (record == null) {
+    if (Objects.isNull(record)) {
```
qiaojialin: the record could not be null, You already init this field in the constructor of this class. Remove this check

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
```diff
@@ -76,12 +52,13 @@
   private List<String> columnTypeList; // no deduplication
   private Map<String, Integer> columnInfoMap; // used because the server returns deduplicated columns
```
qiaojialin: add some javadoc

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
```diff
@@ -355,12 +333,12 @@ public double getDouble(int columnIndex) throws SQLException {
   public double getDouble(String columnName) throws SQLException {
     checkRecord();
     int index = columnInfoMap.get(columnName) - START_INDEX;
-    Field field = record.getFields().get(index);
-    if (field.getDataType() != null && field.getDataType() == TSDataType.DOUBLE) {
-      return field.getDoubleV();
-    } else {
+    if (record[index+1] != null) {
+      return Utils.bytesToDouble(record[index+1]);
```
qiaojialin: We have a ReadWriteIOUtils, you could use it.

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
```diff
@@ -1191,14 +1229,29 @@ private String findColumnNameByIndex(int columnIndex) throws SQLException {
...
-    if (index < 0 || index >= record.getFields().size()) {
+    if (index < 0 || index >= record.length-1 || record[index+1] == null) {
       return null;
     }
-    Field field = record.getFields().get(index);
-    return field.getDataType() == null ? null : field.getStringValue();
+    TSDataType dataType = TSDataType.valueOf(columnTypeList.get(index));
+    switch (dataType) {
+      case BOOLEAN:
+        return String.valueOf(BytesUtils.bytesToBool(record[index+1]));
```
qiaojialin: use ReadWriteIOUtils or BytesUtils in all places

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -931,7 +883,7 @@ public TSFetchResultsResp fetchResults(TSFetchResultsReq req) {
       } else {
         result = QueryDataSetUtils.convertQueryDataSetByFetchSize(queryDataSet, req.fetchSize);
       }
-      boolean hasResultSet = (result.getRowCount() != 0);
+      boolean hasResultSet = (result.bufferForTime().limit() != 0);
```
qiaojialin: limit or remaining?

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
JackieTien97: what if the record value has been set to null before

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
JackieTien97: Both ok, when there is no data, limit=position=0

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
```diff
@@ -716,10 +693,75 @@ public boolean next() throws SQLException {
...
     return true;
   }
 
+  private void constructOneRow() {
+    tsQueryDataSet.time.get(time);
+    for (int i = 0; i < tsQueryDataSet.bitmapList.size(); i++) {
+      ByteBuffer bitmapBuffer = tsQueryDataSet.bitmapList.get(i);
+      // another new 8 row, should move the bitmap buffer position to next byte
+      if (rowsIndex % 8 == 0) {
+        bitmapBuffer.get();
```
qiaojialin: this could be buffered in a List<Byte>

JackieTien97: ok

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
```diff
@@ -76,12 +52,15 @@
...
+  private int rowsIndex = 0; // used to record the row index in current TSQueryDataSet
   private int fetchSize;
   private boolean emptyResultSet = false;
 
   private TSQueryDataSet tsQueryDataSet = null;
-  private ByteBuffer byteBuffer = null;
+  private byte[] time; // used to cache the current time value
+  private byte[][] values; // used to cache the current row record value
+  private byte[] currentBitmap; // used to cache the current bitmap for every column
+  private static final int flag = 0x80; // used to do `or` operation with bitmap to judge whether the value is null
```
qiaojialin:
```suggestion
  private static final int flag = 0x80; // used to do `and` operation with bitmap to judge whether the value is null
```

PR #629, opened 2019-12-05
[IOTDB-343] Add test method in session

SilverNarcissus:
Add test method in session


SilverNarcissus:
> the code is fine...I just wonder is it necessary

This feature is designed for helping our user to analyze the time consuming about their code. User can change almost nothing to analyze the performance about preprocessing.

PR #631, opened 2019-12-09
Add data to TSExecuteStatementResp

liutaohua:
In the original design, a query needed to return metadata before communicating with the server, which was too bloated in the `data.size () < fetchSize` scenario, so I thought it was necessary to reduce communication to the server once


server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -859,60 +875,71 @@ public TSFetchResultsResp fetchResults(TSFetchResultsReq req) {
 
       QueryDataSet queryDataSet;
       if (!queryId2DataSet.get().containsKey(req.queryId)) {
-        queryDataSet = createNewDataSet(req);
+        queryDataSet = createNewDataSet(req.queryId);
```
qiaojialin: Users may call  "executeQueryStatement(TSExecuteStatementReq req)"  in line 651 to execute a query. The dataset needs to be also added in this function. 

It's better to move the fill data into executeQueryStatement(long statementId, PhysicalPlan plan), then the if statement in fetchResults could be removed, becuase the dataset must be saved in queryId2DataSet and queryId2Plan.

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -568,6 +568,18 @@ public TSExecuteStatementResp executeStatement(TSExecuteStatementReq req) {
         if (sqlArgumentsList.size() > MAX_SIZE) {
           sqlArgumentsList.subList(0, DELETE_SIZE).clear();
         }
+        //fill data
+        if (resp.operationHandle.hasResultSet) {
+          long queryId = resp.getOperationHandle().operationId.queryId;
+          QueryDataSet newDataSet = createNewDataSet(queryId);
+          TSQueryDataSet tsQueryDataSet = getTsQueryDataSet(req.fetchSize, newDataSet);
+          TSFetchResultsResp tsFetchResultsResp = getTsFetchResultsResp(queryId, tsQueryDataSet);
```
qiaojialin: maybe adding a checkDataSetIsNull will make the codes clear

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBQueryResultSet.java
```diff
@@ -671,34 +690,51 @@ public void moveToInsertRow() throws SQLException {
...
+        RpcUtils.verifySuccess(resp.getStatus());
+      } catch (IoTDBRPCException e) {
+        throw new IoTDBSQLException(e.getMessage(), resp.getStatus());
+      }
+      if (!resp.hasResultSet) {
+        emptyResultSet = true;
+      } else {
+        tsQueryDataSet = resp.getQueryDataSet();
+        rowsIndex = 0;
+        constructOneRow();
```
qiaojialin: it's better to move this line outside

qiaojialin:
The queryId2Plan is not necessary

PR #645, opened 2019-12-12
enable Sonarcloud

jixuan1989:
This PR finishes add the following functions for CI:

1. add code quality check on Travis. That is, if too many code reliability problems or Vulnerabilities are introduced, the travis test will fail.

2. add a code-coverage report on travis. When submit a PR, the code-coverage will be calculated. If the coverage ratio is less than the thresholds, the test will fail.
Currently, the method, branch, and class coverage requirements are all 0%. (I have to set the threshold  as 0, otherwise the test will fail...) 

3. enable sonar code quality analysis on Jenkins just for the master branch.

4. remove cobertura plugin as the plugin does not support jdk11 and is not maintained anymore.

5. add JaCoCo for code coverage check, use `mvn test -Pcode-coverage` to enable it.

6. Fix that `-DskipTests` does not work for hadoop module.

TODO: Integrate coveralls.io.


jixuan1989:
need squash

PR #650, opened 2019-12-13
[IOTDB-361]Refactor session management

jt2594838:
1. Remove uses of ThreadLocals in TSServiceImpl
2. TS_SessionHandles are replaced with sessionIds
3. TSOperationHandles are replaced with queryIds
4. Make queryId in TSServiceImpl generated by QueryResourceManager

To remove the constraint of serving each client with a separate thread, ThreadLocals in TSServiceImpl are removed. Instead, each session (connection) will be given a sessionId as the unique identifier. In every request that needs authentication, the sessionId must be passed along with to indicate who the client is.

This also fundamentally supports query management in the distributed version.


service-rpc/src/main/thrift/rpc.thrift
```diff
@@ -273,6 +237,22 @@ struct ServerProperties {
...
+struct TSQueryDataSet{
+   // ByteBuffer for time column
+   1: required binary time
+   // ByteBuffer for each column values
+   2: required list<binary> valueList
+   // Bitmap for each column to indicate whether it is a null value
+   3: required list<binary> bitmapList
+}
+
+struct TSColumnSchema{
```
qiaojialin: is this used?

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBPreparedStatement.java
```diff
@@ -18,40 +18,56 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.apache.iotdb.service.rpc.thrift.TSIService.Iface;
 
 public class IoTDBPreparedStatement extends IoTDBStatement implements PreparedStatement {
 
   private String sql;
   private static final String METHOD_NOT_SUPPORTED_STRING = "Method not supported";
+  private long queryId;
```
JackieTien97: It seems that this field hasn't been used.

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBStatement.java
```diff
@@ -73,29 +69,23 @@
...
+  private long stmtId = -1;
+  private long queryId = -1;
 
   /**
    * Constructor of IoTDBStatement.
    */
   IoTDBStatement(IoTDBConnection connection, TSIService.Iface client,
-      TS_SessionHandle sessionHandle,
-      int fetchSize, ZoneId zoneId) throws SQLException {
+      Long sessionId, int fetchSize, ZoneId zoneId) throws SQLException {
```
JackieTien97: Long seesionId or long

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBStatement.java
```diff
@@ -358,7 +349,7 @@ private boolean executeSQL(String sql) throws TException, SQLException {
 
   private int[] executeBatchSQL() throws TException, SQLException {
```
JackieTien97: Why the queryId is not set in this sql function

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBPreparedStatement.java
qiaojialin: No need to review this class, the preparedStatement is removed in f_batch_reader branch.

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBConnection.java
```diff
@@ -434,6 +443,10 @@ private void openSession() throws SQLException {
     } catch (TException e) {
       throw new SQLException(String.format("Can not establish connection with %s.",
           params.getJdbcUriString()), e);
+    } catch (IoTDBRPCException e) {
+      // failed to connect, disconnect from the server
+      transport.close();
```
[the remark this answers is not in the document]
jt2594838: closed

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBStatement.java
jt2594838: changed to long

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBStatement.java
jt2594838: queryId is mainly for the resource control of data queries, which are not allowed in the batch statement, so it is not necessary to set it in batch execution.

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBPreparedStatement.java
jt2594838: removed

server/src/main/java/org/apache/iotdb/db/query/executor/EngineQueryRouter.java
```diff
@@ -139,7 +139,7 @@ public QueryDataSet fill(List<Path> fillPaths, long queryTime, Map<TSDataType, I
       QueryContext context)
       throws StorageEngineException, QueryProcessException, IOException {
 
-    long nextJobId = context.getJobId();
+    long nextJobId = context.getQueryId();
```
fanhualta: Maybe it's better to unify `queryId` and `jobId`. It looks like confused if you mix using them.

LeiRui:
This pr make the code many times clearer than before :tada:

LeiRui:
> 1. Remove uses of ThreadLocals in TSServiceImpl
> 2. TS_SessionHandles are replaced with sessionIds
> 3. TSOperationHandles are replaced with queryIds
> 4. Make queryId in TSServiceImpl generated by QueryResourceManager
> 
> To remove the constraint of serving each client with a separate thread, ThreadLocals in TSServiceImpl are removed. Instead, each session (connection) will be given a sessionId as the unique identifier. In every request that needs authentication, the sessionId must be passed along with to indicate who the client is.
> 
> This also fundamentally supports query management in the distributed version.

I have two questions to confirm:
1. Note that TSServiceImpl.currSessionId is still ThreadLocal<>(). So to be accurate, the use of ThreadLocal is not totally removed in TSServiceImpl.
2. `generateQueryId` is used in `TSServiceImpl.executeUpdateStatement`.  In `generateQueryId` method, `QueryFileManager.addQueryId` will new HashSet<>(). Is that a waste for update statements?

jt2594838:
> > 1. Remove uses of ThreadLocals in TSServiceImpl
> > 2. TS_SessionHandles are replaced with sessionIds
> > 3. TSOperationHandles are replaced with queryIds
> > 4. Make queryId in TSServiceImpl generated by QueryResourceManager
> > 
> > To remove the constraint of serving each client with a separate thread, ThreadLocals in TSServiceImpl are removed. Instead, each session (connection) will be given a sessionId as the unique identifier. In every request that needs authentication, the sessionId must be passed along with to indicate who the client is.
> > This also fundamentally supports query management in the distributed version.
> 
> I have two questions to confirm:
> 
> 1. Note that TSServiceImpl.currSessionId is still ThreadLocal<>(). So to be accurate, the use of ThreadLocal is not totally removed in TSServiceImpl.
> 2. `generateQueryId` is used in `TSServiceImpl.executeUpdateStatement`.  In `generateQueryId` method, `QueryFileManager.addQueryId` will new HashSet<>(). Is that a waste for update statements?

1. Yes. I realized later that this ThreadLocal is still needed to retrieve the resources when the client is down so I added it, but I forget to update my description.
2. Good point. I think the data queries and non-data queries should be treated differently.

PR #657, opened 2019-12-16
Serialize filters

jt2594838:
In order to pass the query plan to different nodes in the distributed version, the filters should be able to serialize and deserialize, which is implemented in this PR.


tsfile/src/main/java/org/apache/iotdb/tsfile/read/filter/basic/UnaryFilter.java
```diff
@@ -51,4 +60,36 @@ public FilterType getFilterType() {
...
   public abstract Filter clone();
+
+  @Override
+  public void serialize(DataOutputStream outputStream) {
+    try {
+      outputStream.write(getSerializeId().ordinal());
+      outputStream.write(filterType.ordinal());
+      ReadWriteIOUtils.writeObject(value, outputStream);
+    } catch (IOException ignored) {
+      // ignored
```
qiaojialin: log?

jt2594838: This DataOutputStream is backed-up with a ByteArrayOutputStream, technically, there is no chance that an exception will be thrown.

fanhualta: But I don't think there was no chance to throw an Exception even it sounds safe technically.  In case it has an exception,  it'll be hopeless to locate the bug even if it's a bug from JDK.

PR #658, opened 2019-12-16
mem table use skip list map

liutaohua:


jixuan1989:
hi, any performance result?

liutaohua:
> hi, any performance result?

Write is 4x slower than before

PR #659, opened 2019-12-16
[IOTDB-370] fix a multi-thread problem.

Genius-pig:
fix a multi-thread problem.
https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-370


Genius-pig:
> Should use some flag to distinguish using parser1 or parser2 logic, rather than use Exception to switch between the two logics.

I implement the two-stage method to improve speed. I get this idea from the antlr4 site. I think using a flag also needs to catch the exception. https://github.com/antlr/antlr4/blob/master/doc/faq/general.md. When you input the wrong SQL statement, the speed will be almost double slower. But who would enter so many wrong SQLs?

PR #677, opened 2019-12-26
[IOTDB-340] Remove unnecessary getting data types from MManager 

liutaohua:
In a QueryPlan,  the selected paths with data types, predicates (expression) are included.

private List<Path> paths;
private List<TSDataType> dataTypes;
private IExpression expression;
However, when executing a QueryPlan in the AbstractQueryProcessExecutor, only paths and expression are passed to the EngineQueryRouter through QueryExpression. 

Therefore, the data type information is lost and needs to get from manager again, which is a waste.

It's better to pass the full QueryPlan to the EngineQueryRouter for all query types: raw data query, fill query, group by query, aggregation query. 

The QueryExpression is a parameter in the query interface of TsFile, we do not need to use it in the server, because we could get more information on the server.


tsfile/src/main/java/org/apache/iotdb/tsfile/read/expression/QueryExpression.java
```diff
@@ -20,11 +20,13 @@
 
 import java.util.ArrayList;
 import java.util.List;
+import org.apache.iotdb.tsfile.file.metadata.enums.TSDataType;
 import org.apache.iotdb.tsfile.read.common.Path;
 
 public class QueryExpression {
```
jt2594838: This class is no longer used in the main code. You should either reuse it or remove it.

server/src/main/java/org/apache/iotdb/db/qp/strategy/PhysicalGenerator.java
```diff
@@ -350,6 +351,33 @@ private PhysicalPlan transformQuery(QueryOperator queryOperator)
...
+  private void deduplicate(QueryOperator queryOperator, QueryPlan queryPlan)
+      throws QueryProcessException {
+    if (queryPlan instanceof AggregationPlan || queryPlan instanceof GroupByPlan) {
+      if (!queryPlan.isGroupByDevice()) {
+        deduplicateAggregation(queryPlan);
+      }
+    } else {
+      deduplicate(queryPlan);
+    }
+    if (queryPlan.getDeduplicatedPaths() == null) {
```
qiaojialin: The following 3 if could be avoided

server/src/main/java/org/apache/iotdb/db/qp/strategy/PhysicalGenerator.java
```diff
@@ -333,6 +333,7 @@ private PhysicalPlan transformQuery(QueryOperator queryOperator)
       List<Path> paths = queryOperator.getSelectedPaths();
       queryPlan.setPaths(paths);
     }
+    deduplicate(queryOperator, queryPlan);
 
     queryPlan.checkPaths(executor);
```
qiaojialin: this may be unnecessary

server/src/main/java/org/apache/iotdb/db/qp/strategy/PhysicalGenerator.java
```diff
@@ -368,5 +395,58 @@ private PhysicalPlan transformQuery(QueryOperator queryOperator)
...
+  private void deduplicateAggregation(QueryPlan queryPlan)
+      throws QueryProcessException {
+    List<Path> paths = queryPlan.getPaths();
+    List<String> aggregations = queryPlan.getAggregations();
+    if (paths == null) {
+      throw new QueryProcessException("Parameters should not be null.");
+    }
+    if (paths.size() != aggregations.size()) {
+      return;
+    }
```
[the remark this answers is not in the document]
qiaojialin: if the user write a SQL: select count(s0),s0 from root.*    , the Antlr will throw an exception. Therefore, this check could be removed.

server/src/main/java/org/apache/iotdb/db/qp/strategy/PhysicalGenerator.java
```diff
@@ -368,5 +396,66 @@ private PhysicalPlan transformQuery(QueryOperator queryOperator)
...
   }
 
+
+  private void deduplicateAggregation(QueryPlan queryPlan)
+      throws QueryProcessException {
+    List<Path> paths = queryPlan.getPaths();
+    List<String> aggregations = queryPlan.getAggregations();
+
+    Map<Path, TSDataType> dataTypeMapping = queryPlan.getDataTypeMapping();
+    if (paths == null) {
```
qiaojialin: This could also be removed.
If the user selects some non-existing paths, the ConcatPathOptimizer will throw an exception.

server/src/main/java/org/apache/iotdb/db/qp/strategy/PhysicalGenerator.java
```diff
@@ -368,5 +396,66 @@ private PhysicalPlan transformQuery(QueryOperator queryOperator)
...
+    }
+    queryPlan.setDeduplicatedPaths(deduplicatedPaths);
+    queryPlan.setDeduplicatedDataTypes(deduplicatedDataTypes);
+    queryPlan.setDeduplicatedAggregations(deduplicatedAggregations);
+  }
+
+  private void deduplicate(QueryPlan queryPlan)
+      throws QueryProcessException {
+    List<Path> paths = queryPlan.getPaths();
+    if (paths == null) {
```
qiaojialin: remove this also

server/src/main/java/org/apache/iotdb/db/qp/physical/crud/QueryPlan.java
```diff
@@ -33,6 +35,12 @@
 
   private List<Path> paths = null;
   private List<TSDataType> dataTypes = null;
+
+  private List<Path> deduplicatedPaths = null;
+  private List<TSDataType> deduplicatedDataTypes = null;
+  private List<String> deduplicatedAggregations = new ArrayList<>();
+  private Map<Path, TSDataType> dataTypeMapping = new HashMap<>();
```
qiaojialin: This field may be useless, you can use deduplicatedPaths and deduplicatedDataTypes.

server/src/main/java/org/apache/iotdb/db/query/dataset/groupby/GroupByWithValueFilterDataSet.java
```diff
@@ -67,9 +69,9 @@ public GroupByWithValueFilterDataSet(long queryId, List<Path> paths, long unit,
   /**
    * init reader and aggregate function.
    */
-  public void initGroupBy(QueryContext context, List<String> aggres, IExpression expression)
-      throws StorageEngineException, QueryProcessException, IOException {
-    initAggreFuction(aggres);
+  public void initGroupBy(QueryContext context, List<String> aggres, List<TSDataType> dataTypes,
```
qiaojialin: change this to a private method that called in GroupByWithValueFilterDataSet constructor

server/src/main/java/org/apache/iotdb/db/query/dataset/groupby/GroupByWithoutValueFilterDataSet.java
```diff
@@ -70,9 +71,10 @@ public GroupByWithoutValueFilterDataSet(long queryId, List<Path> paths, long uni
   /**
    * init reader and aggregate function.
    */
-  public void initGroupBy(QueryContext context, List<String> aggres, IExpression expression)
+  public void initGroupBy(QueryContext context, List<String> aggres, List<TSDataType> dataTypes,
```
qiaojialin: Change this to a private method and called in constructor. More, the constructor could receive a GroupByPlan as a parameter, which is clearer

qiaojialin:
Could the deduplicated codes be removed in IoTDBQueryResultSet? Just use the deduplicated paths passed by server

PR #684, opened 2019-12-27
[IOTDB-389] Parallel Raw data query

JackieTien97:
I apply the producer-consumer module and only change the NewEngineDataSetWithoutValueFilter.

Add a blocking queue that buffers a predefined(5 in this version) number of BatchDatas for each time series. Once a BatchData is produced and the blocking queue is not full, the producer submit itself to generate another task to produce for this time series before it ends itself.

The main thread also the consumer is responsible for taking from the blocking of each time series to align them. If one queue is empty, it means the data of that time series is not ready, main thread just blocked and waiting for it ready.

The performance evaluation:

In my mac, one device, ten sensors, long datatype, RLE & SNAPPY. Each time series contains 1 billion data points. Total 562M data on disk.

I query the raw data of ten time series with a time filter that results in 10, 20 and 30 millionpoints in total.

Here are the results:
(master version: 994dc325c) (multiThread version: 47a1e0c)
___________________________________________________________________________________
point number per sensor   |    master(single-thread)      |     TYMultiThread        |
-------------------------------------------------------------------------------
10,000,000                         |                16160ms                |             10930ms          |
-------------------------------------------------------------------------------
20,000,000                         |               25970ms                |             20227ms          |
-------------------------------------------------------------------------------
30,000,000                         |               42424ms                |             30281ms          |
-------------------------------------------------------------------------------

It improves about 30%.

And we analyze the results, each time the main thread(consumer) to take from the queue, the queue is always full. It means that almost all time is spent on the data align.
And the analysis is supported by the JProfile results.


qiaojialin:
Multi-thread:  most time is caused by fillRPCDataSet, the fetchResult is asynchronous. 
image

Single-thread in master
image

server/src/main/java/org/apache/iotdb/db/query/dataset/NewEngineDataSetWithoutValueFilter.java
```diff
@@ -19,63 +19,135 @@
...
+        // the reader has next batch
+        if (reader.hasNextBatch()) {
+          BatchData batchData = reader.nextBatch();
+          blockingQueue.put(batchData);
+          // if the queue also has free space
+          // just submit another itself
+          if (blockingQueue.remainingCapacity() > 0) {
+            pool.submit(this);
+          }
+          // the queue has no more space
```
qiaojialin: put into the else

server/src/main/java/org/apache/iotdb/db/query/dataset/NewEngineDataSetWithoutValueFilter.java
```diff
@@ -19,62 +19,136 @@
...
+  // noMoreDataInQueue can still be true
+  // its usage is to tell the consumer thread not to call the take() method.
+  private boolean[] noMoreDataInQueueArray;
+
   private BatchData[] cachedBatchDataArray;
 
   private static final int FLAG = 0x01;
 
+  // capacity for blocking queue
+  private static final int BLOCKING_QUEUE_CAPACITY = 5;
```
qiaojialin: add to iotdb-engine.properties?
rename to  CACHED_BATCH_DATA_NUMBER or so

server/src/main/java/org/apache/iotdb/db/query/dataset/NewEngineDataSetWithoutValueFilter.java
```diff
@@ -244,8 +321,34 @@ public TSQueryDataSet fillBuffer(int fetchSize, WatermarkEncoder encoder) throws
...
+      noMoreDataInQueueArray[seriesIndex] = true;
+    }
+    // there are more batch data in this time series queue
+    else {
+      cachedBatchDataArray[seriesIndex] = batchData;
+
+      synchronized (seriesReaderWithoutValueFilterList.get(seriesIndex)) {
+        // we only need to judge whether to submit another task when the queue is not full
+        if (blockingQueueList.get(seriesIndex).remainingCapacity() > 0) {
+          SeriesReaderWithoutValueFilter reader = seriesReaderWithoutValueFilterList.get(seriesIndex);
```
qiaojialin: move this to the outside of synchronized

server/src/main/java/org/apache/iotdb/db/query/dataset/NewEngineDataSetWithoutValueFilter.java
```diff
@@ -274,22 +377,27 @@ protected RowRecord nextWithoutConstraint() throws IOException {
...
         if (!cachedBatchDataArray[seriesIndex].hasCurrent()) {
-          if (seriesReaderWithoutValueFilterList.get(seriesIndex).hasNextBatch()) {
-            cachedBatchDataArray[seriesIndex] = seriesReaderWithoutValueFilterList
-                .get(seriesIndex).nextBatch();
+          // still have remaining batch data in queue
+          if (!noMoreDataInQueueArray[seriesIndex]) {
+            try {
+              fillCache(seriesIndex);
+            } catch (InterruptedException e) {
+              e.printStackTrace();
```
jt2594838: Use logger.

server/src/main/java/org/apache/iotdb/db/query/dataset/NewEngineDataSetWithoutValueFilter.java
```diff
@@ -86,7 +88,7 @@ public void run() {
           reader.setManagedByQueryManager(false);
         }
       } catch (IOException | InterruptedException e) {
-        e.printStackTrace();
+        LOGGER.error(e.getMessage());
```
jt2594838: Log the exception instead of merely its message unless the exception is very tiny otherwise, it may be a disaster in future debugs.

SilverNarcissus:
Nice work!

PR #685, opened 2019-12-27
abstract TsFileFlushPolicy and allow specifying storage groups in flush command

jt2594838:
To make the different replicas of a data partition consistent, the distributed version needs to force all replicas close file by the leader instead of on their own. So the TsFileFlushPolicy is abstracted to make this behavior controllable.

Moreover, since the leader should be able to make a follower close one of its storage group, the grammar of the flush command is enhanced to support finer granularity.


server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -485,6 +491,25 @@ private boolean execAdminCommand(String statement, long sessionId) throws Storag
     }
   }
 
+  private void execFlush(String statement) throws StorageGroupNotSetException {
+    String[] args = statement.split("\\s+");
+    if (args.length == 1) {
+      StorageEngine.getInstance().syncCloseAllProcessor();
+    } else if (args.length == 2){
+      String[] storageGroups = args[1].split(",\\s*");
```
fanhualta: What do you want to match here? If the command is "FLUSH root.a.b,  root.b.c", the length if `args` is 3 which is invalid. So in your gramma, you don't allow space between `storageGroups`. So it's unnecessary to split `,\\s*`, just `,` is ok.

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -485,6 +491,25 @@ private boolean execAdminCommand(String statement, long sessionId) throws Storag
...
+      for (String storageGroup : storageGroups) {
+        StorageEngine.getInstance().asyncCloseProcessor(storageGroup, true);
+        StorageEngine.getInstance().asyncCloseProcessor(storageGroup, false);
+      }
+    } else {
+      String[] storageGroups = args[1].split(",\\s*");
```
fanhualta: The same as the above.

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
jt2594838: fixed

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
jt2594838: fixed

PR #698, opened 2019-12-30
Define disable align clause

Ring-k:
The disable align clause is defined in the ANTLR file. And the logical plan can be generated now.


HTHou:
I'm working on the issue IOTDB-396, but I am not familiar with antlr4. Your work help me a lot. Thank you! :)

PR #700, opened 2019-12-31
Continue writing the last unclosed file

jt2594838:
As the TsFile close is strictly controlled in the distributed version, if a node arbitrarily closes the TsFile that is not ordered to close, the file consistency of different nodes will be broken. So during start-up, we should only close files that are confirmed to close.  

To support this, a flag is created when a file is to be closed and the flag is removed after its close is done.


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -287,24 +288,42 @@ private void continueFailedRenames(File fileFolder, String suffix) {
...
+    for (int i = 0; i < tsFiles.size(); i++) {
+      TsFileResource tsFileResource = tsFiles.get(i);
       sequenceFileList.add(tsFileResource);
       TsFileRecoverPerformer recoverPerformer = new TsFileRecoverPerformer(storageGroupName + "-"
-          , schema, versionController, tsFileResource, false);
-      recoverPerformer.recover();
-      tsFileResource.setClosed(true);
+          , schema, versionController, tsFileResource, false, i == tsFiles.size() - 1);
+      RestorableTsFileIOWriter writer = recoverPerformer.recover();
+      if (i != tsFiles.size() - 1) {
```
qiaojialin: We could have more than one unclosed TsFiles in a SG

PR #704, opened 2020-01-02
Add log for deadlock

JackieTien97:


server/src/main/java/org/apache/iotdb/db/query/dataset/NewEngineDataSetWithoutValueFilter.java
```diff
@@ -66,7 +68,9 @@ public void run() {
             if (batchData.isEmpty()) {
               continue;
             }
+            logger.info("Reader-" + index + ": start putting batch data into queue");
```
jixuan1989: Hi, if you want to merge the codes into master branch, `logger.info` is too high level... 
I suggest you use logger.debug and modify logback.xml to claim the log level of NewEngineDataSetWithoutValueFilter.java  as debug. （But do not forget to recover the logback.xml file after the test)

```
<logger level="debug" name="org.apache.iotdb.db.conf.adapter.IoTDBConfigDynamicAdapter">
        <appender-ref ref="stdout"/>
    </logger>
```

Or, if you just want to use travis to reproduce the issue, you can manually trigger a request on travis by using the following parameter:

`-Dlogback.configurationFile=/temporary/logback.xml`

jixuan1989: And `if (logger.isDebugEnable())` is needed if there are too many string concat operations.

PR #707, opened 2020-01-06
[IOTDB-404]fix the bug when recover the last crashed file

HTHou:
It also fixed the issue https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-404?filter=allissues


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -308,14 +308,16 @@ private void recoverSeqFiles(List<TsFileResource> tsFiles) throws StorageGroupPr
       TsFileRecoverPerformer recoverPerformer = new TsFileRecoverPerformer(storageGroupName + "-"
           , schema, versionController, tsFileResource, false, i == tsFiles.size() - 1);
       RestorableTsFileIOWriter writer = recoverPerformer.recover();
-      if (i != tsFiles.size() - 1) {
+      if (i != tsFiles.size() - 1 || !writer.canWrite()) {
         // not the last file, just close it
```
qiaojialin: change the comment, what  !write.canWrite means?

HTHou: Fixed.

PR #709, opened 2020-01-06
[IOTDB-401]Correct the calculation of a chunk if there is no data in the chunk

jixuan1989:
when flushing a chunk, tsfile will calculate the size of the chunk and compare the size and the actual data size in the output stream.
When calculating the size, the chunk header is calculated (which has the measurement id and therefore the size always > 0 ). However, if we find the chunk has no data points, we do not flush anything.
In this case, the size ( >0 ) != the actual size (=0).

This PR fixes the above case.


tsfile/src/main/java/org/apache/iotdb/tsfile/write/chunk/ChunkWriterImpl.java
```diff
@@ -231,6 +231,9 @@ public long estimateMaxSeriesMemSize() {
 
   @Override
   public long getCurrentChunkSize() {
+    if (this.getCurrentDataSize() == 0) {
```
qiaojialin: Is this check necessary?

jt2594838: You may save the calculation result and use it in line238 to avoid duplicated calculation.

And if all chunks in a chunk group are empty, this chunk group should not be flushed into the file, which will generate an empty ChunkGroupMetadata and waste some space.

jixuan1989: `this.getCurrentDataSize()` actually is `pageBuffer.size()` while `pageBuffer` call `outputstream.count`.
Therefore, I remove getCurrentDataSize() and use pageBuffer.size() instead.

@qiaojialin , yes I think so. 
No matter whether you check the serialized size in flushAll... method, this method is needed and should check the data size here.

PR #714, opened 2020-01-08
Set whether to ignore time in JDBC.

Genius-pig:
Set whether to ignore time in JDBC.


jixuan1989:
Hi,
as you have the write permission now, it is recommended to use apache repo directly and thereby it can enable SonarCloud check.

Genius-pig:
> Hi,
> as you have the write permission now, it is recommended to use apache repo directly and thereby it can enable SonarCloud check.

How?

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -585,6 +585,7 @@ private TSExecuteStatementResp executeQueryStatement(long statementId, PhysicalP
         resp = getAuthQueryColumnHeaders(plan);
       } else if (plan instanceof ShowPlan) {
         resp = getShowQueryColumnHeaders((ShowPlan) plan);
+        resp.setIgnoreTimeStamp(true);
```
qiaojialin: this could be omitted

PR #715, opened 2020-01-08
Change sync log and sync system dir

fanhualta:
This pr fixes some minor issues:
1.  Extract sync log from the server log to be independent dir in logs dir.
2. Change the directory of sync dir from `data` directory to `system` dir which contains global info for one single instance, including `UUID file`, `lock file`, `device owner map file`.


server/src/assembly/resources/conf/logback-sync.xml
```diff
@@ -0,0 +1,139 @@
...
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+
+-->
+<configuration scan="true" scanPeriod="60 seconds">
```
jixuan1989: why add a new file? how to use it?
it is better to add a appender in previous logback.xml for sync.

jixuan1989:
try to merge master first.

server/src/assembly/resources/conf/logback-sync.xml
fanhualta: fixed.

PR #726, opened 2020-01-09
Made all tests Locale Invariant by always explicitly defaulting to Lo…

JulianFeinauer:
…cale.ENGLISH.

Fixes comment in my VOTE Email https://lists.apache.org/thread.html/rfef7f0035f28c3ee81ce32135e46e53774ce19b4af0bc06db426d4e3%40%3Cdev.iotdb.apache.org%3E

### DIFF
--- a/server/src/test/java/org/apache/iotdb/db/integration/IoTDBDeletionIT.java
+++ b/server/src/test/java/org/apache/iotdb/db/integration/IoTDBDeletionIT.java
@@ -236,12 +236,12 @@ public class IoTDBDeletionIT {
 
       // prepare BufferWrite data
       for (int i = 10001; i <= 20000; i++) {
-        statement.execute(String.format(insertTemplate, i, i, i, (double) i, "\'" + i + "\'",
+        statement.execute(String.format(Locale.ENGLISH, insertTemplate, i, i, i, (double) i, "\'" + i + "\'",
             i % 2 == 0));
       }
       // prepare Overflow data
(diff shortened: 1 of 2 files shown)


sunjincheng121:
I would like to have a look.

jixuan1989:
> Thanks for the PR @JulianFeinauer
> The PR looks good to me for fix this issue, I only left one comment for simple the code change and one suggestion as follows:
> 
> I checked the entire project and found that `String.format` is used in many places(not only in test code). We are fortunate to find this `Local` problem in ITCase, but we cannot rule out that there are unpredictable problems hidden in other places, so I think would be better make a util similar with `String.format` in IoTDB , and set the configuration we need in the util, such as `Local` information, so that even if other problems are found in the future, we can fix it as a whole by modifying one place.
> 
> What do you think? @JulianFeinauer @jixuan1989

I  think many String.format() is used for log. Because all messages are in English, I wonder why a Locale is needed...

JulianFeinauer:
Excellent idea... The issue is that with e. G German locals decimal points are printed as comma...

JulianFeinauer:
@jixuan1989 if String.format is used for log only we have no issue but in the places above it is used to concatenate (or expand) Query Strings. And here, for some Locales (like e.g. German) the decimal points are expanded to decimal commas which leads to syntactically wrong statements, of course. So this, for sure, has to be Locale Invariant!

jixuan1989:
I have checked. All String.format are used for either printing log or test. So, it should be ok.

PR #727, opened 2020-01-10
[IOTDB-412]Paths are not correctly deduplicated

liutaohua:
In the previous design, we needed to move all the work removing the duplicate `Path` to the `PhysicalGenerator` class to do this, but without the `group by device` because it is a special way of working, we kept the deduplication work in the `DeviceIterateDataSet` class.

So, when a `group by device` query after `PhysicalGenerator`, `deduplicatedPaths` and `deduplicatedDataTypes` should be an empty state, and the `paths` and `datatypes` should be meet the requirements of the implementation of `DeviceIterateDataSet` , for more complex description. 

I created a `ut` `PhysicalPlanTest.testDeduplicatedPath` to describe these things.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/qp/strategy/PhysicalGenerator.java
+++ b/server/src/main/java/org/apache/iotdb/db/qp/strategy/PhysicalGenerator.java
@@ -418,6 +418,7 @@ public class PhysicalGenerator {
   }
 
   private void deduplicate(QueryPlan queryPlan) {
+    //The deduplication of a GroupByDevice query is done in the dataset
     if (queryPlan.isGroupByDevice()) {
       return;
     }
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/iotdb/db/qp/strategy/PhysicalGenerator.java
```diff
@@ -416,11 +418,12 @@ private void generateDataTypes(QueryPlan queryPlan) throws PathException {
   }
 
   private void deduplicate(QueryPlan queryPlan) {
+    if (queryPlan.isGroupByDevice()) {
+      return;
```
jt2594838: Add a comment to explain this like "The deduplication of a GroupByDevice query is done in the dataset".

liutaohua: done, thanks for your advice

PR #728, opened 2020-01-11
[IOTDB-414] fix the client display and add COMPRESSION keyword.

Genius-pig:
fix the client display.

### DIFF
--- a/client/src/main/java/org/apache/iotdb/client/AbstractClient.java
+++ b/client/src/main/java/org/apache/iotdb/client/AbstractClient.java
@@ -273,7 +273,7 @@ public abstract class AbstractClient {
         }
       } else {
         for (int i = 1; i <= colCount; i++) {
-          printColumnData(resultSetMetaData, res, i, zoneId);
+          printf(formatValue, res.getString(i));
         }
       }
       println();
@@ -475,10 +475,6 @@ public abstract class AbstractClient {
     StringBuilder blockLine = new StringBuilder();
     if (printTimestamp) {
       blockLine.append("+").append(StringUtils.repeat('-', maxTimeLength)).append("+");
-    } else {
-      blockLine.append("+");
-    }
-    if(printTimestamp) {
       if (resultSetMetaData.getColumnName(2).equals(GROUPBY_DEVICE_COLUMN_NAME)) {
         maxValueLength = measurementColumnLength;
       } else {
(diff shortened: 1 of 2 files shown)


client/src/main/java/org/apache/iotdb/client/AbstractClient.java
```diff
@@ -478,15 +478,17 @@ private static void printBlockLine(boolean printTimestamp, int colCount,
...
+        int tmp = Integer.MIN_VALUE;
+        for (int i = 1; i <= colCount; i++) {
+          int len = resultSetMetaData.getColumnLabel(i).length();
+          tmp = Math.max(tmp, len);
+        }
+        maxValueLength = tmp;
       }
-      maxValueLength = tmp;
     }
     if(printTimestamp) {
```
JackieTien97: Why not merge this if(printTimestamp) into next?

JackieTien97: And I try your version, it seems that you didn't fix the "show storage group" problem. It still show the error message.

SolomonAnn: > And I try your version, it seems that you didn't fix the "show storage group" problem. It still show the error message.

The problem has been fixed. Perhaps you need to recompile the project.

JackieTien97: Sorry for that, I try it again and it really works.
And you can split the `printBlockLine` function into two functions.
Because, the logic in that function is completely split by the `printTimestamp`.
image

SolomonAnn: The "create timeseries" statement with the "compression" argument can't still print successfully.

JackieTien97: it can create successfully, but it does the wrong thing, I think we should choose one bewteen syntax `compression` and `compressor`.
image

Genius-pig: Compression is a wrong syntax, the wrong syntax doesn't mean it gives an error message. Instead, It is identified as a normal property and doesn't influence anything. So if you use `create timeseries root.sg1.d2.s1 with datatype=INT64, encoding=RLE, compression=SNAPPY;`, root.sg1.d2.s1 will use UNCOMPRESSED.

PR #739, opened 2020-01-14
[hotfix] Simplify the setting of Locale defaults for String.format().

sunjincheng121:
I would like to make the change simple which I mentioned here: https://github.com/apache/incubator-iotdb/pull/726#discussion_r365042005

What do you think?

### DIFF
--- a/server/src/test/java/org/apache/iotdb/db/integration/IoTDBDeletionIT.java
+++ b/server/src/test/java/org/apache/iotdb/db/integration/IoTDBDeletionIT.java
@@ -53,6 +53,7 @@ public class IoTDBDeletionIT {
 
   @Before
   public void setUp() throws Exception {
+    Locale.setDefault(Locale.ENGLISH);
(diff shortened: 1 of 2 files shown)


sunjincheng121:
@jixuan1989 @JulianFeinauer It's this PR make sense to you?

sunjincheng121:
Thanks for your review @JulianFeinauer !

At the beginning I really wanted to restore the locale in the tearDown () method,  but in that way, we should keep the original `Locale` info which is make the change a little bit complex, furthermore, currently, we always need `Locale.ENGLISH` in all of testCase, so, I think the change is fine. 

But I am fine, if you really want to do it, which make the code looks pretty complete.

What do you think?

JulianFeinauer:
@sunjincheng121 I agree with your arguments and do not insist on em and as your change is, as you say "somewhat" idempotent we can leave it like that. 👍

PR #741, opened 2020-01-14
[IOTDB-421] Fix merge problem

SilverNarcissus:
After dividing storage groups by time partition, merge part contains a potential bug about merging different partition's data into one partition.
So I change the logic of merge and only merge the data from the same time partition


jt2594838:
It is okay, but it seems you are trying to bypassing this problem instead of solving it directly.
Please also dig out the root fo this bug, or someone may have to one day.

SilverNarcissus:
This problem has been fixed at https://github.com/apache/incubator-iotdb/pull/743 more gracefully.

PR #743, opened 2020-01-15
[IOTDB-421]fix that selected seq files for merge are not sorted

jt2594838:
Because the current MergeFileSelector select merge files by unseq files if the unseq files overlap seq files in a reverted order, the selected seq files will be out of order.
Example:
1. write a seq file A, which ranges [10, 20]
2. write a seq file B, which ranges [20, 30]
3. write an unseq file C, which ranges [20, 30]
4. write an unseq file D, which ranges [10, 20]
C overlaps B and D overlaps and because C is generated before D, as a result, the selected seq files will be [B, A] instead of [A, B], which will make the following merge process write data into a wrong file.

To resolve this, the selected seq files are reordered before starting the actual merge task.


SilverNarcissus:
This fixing is amazing!   I have tested it with large data which contains unordered problem. It gives the right result and the code is brilliant! Thanks!

PR #753, opened 2020-01-17
Add iterators to TVList to prevent data copy

liutaohua:
In the previous design, when you used in-memory data:

1. the `TVList` structure was `clone` into an `IWritableMemChunk` named `chunkCopy`

1. And then copied again into a `List<TimeValuePair>` structure

1. If multiple data happens to exist in `flushingMemTable`, `MemSeriesLazyMerger` will use `addAll(List<TimeValuePair>)` to copy again

so, I think we should directly use `TVList` to implement `IPointReader` interface and use it externally

- Add the `Iterator` class and `getIterator` method for `TVlist` 

- Use `PriorityMergeReader` to sort the data in `ReadOnlyMemChunk`

Now, Data was copied only when it is used


server/src/main/java/org/apache/iotdb/db/engine/memtable/AbstractMemTable.java
```diff
@@ -195,19 +199,19 @@ public boolean isEmpty() {
...
-      IWritableMemChunk chunkCopy = new WritableMemChunk(dataType, memChunk.getTVList().clone());
-      chunkCopy.setTimeOffset(undeletedTime);
-      sorter = chunkCopy;
     }
-    return new ReadOnlyMemChunk(measurement, dataType, sorter, props);
+    long undeletedTime = findUndeletedTime(deviceId, measurement, timeLowerBound);
+    IWritableMemChunk memChunk = memTableMap.get(deviceId).get(measurement);
+    TVList chunkCopy = memChunk.getTVList().clone();
+
+    chunkCopy.setVersion(getVersion());
```
qiaojialin: put this to clone method of TVList

server/src/main/java/org/apache/iotdb/db/engine/memtable/WritableMemChunk.java
```diff
@@ -223,51 +223,6 @@ public void setTimeOffset(long offset) {
     list.setTimeOffset(offset);
   }
 
-  @Override
-  public synchronized List<TimeValuePair> getSortedTimeValuePairList() {
```
qiaojialin: remove the sortedList field in this class and unused import

server/src/main/java/org/apache/iotdb/db/engine/querycontext/ReadOnlyMemChunk.java
```diff
@@ -18,106 +18,69 @@
...
   private TSDataType dataType;
-  private TimeValuePairSorter memSeries;
-  private List<TimeValuePair> sortedTimeValuePairList;
 
   Map<String, String> props;
   private int floatPrecision = TSFileDescriptor.getInstance().getConfig().getFloatPrecision();
+  private ChunkMetaData cachedMetaData;
+
+  private PriorityMergeReader mergeReader;
+  private PriorityMergeReader chunkedReader;
```
qiaojialin: I suggest using one ReadOnlyMemChunk to store one chunk in a memtable. The TsFileProcessor return a Pair<List<ReadOnlyMemtable>, List<ChunkMetadata>>
Init the chunk metadata in the constructor in this class and cache it. 
Therefore, we only need one chunkreader, and it should not be a mergereader, we could directly get the  IPointReader in TVList.
In the AbstractMemTable, the version should be init by Long.MAX_VALUE for a working memtable, the memtable version is set only when it comes to a  flushing memtable.

server/src/main/java/org/apache/iotdb/db/engine/querycontext/ReadOnlyMemChunk.java
```diff
@@ -18,106 +18,69 @@
...
-          sortedTimeValuePairList.replaceAll(x -> new TimeValuePair(x.getTimestamp(),
-              new TsDouble(
-                  MathUtils.roundWithGivenPrecision(x.getValue().getDouble(), floatPrecision))));
-          break;
-        default:
-          break;
-      }
+    mergeReader = new PriorityMergeReader(floatPrecision);
+    chunkedReader = new PriorityMergeReader(floatPrecision);
+    for (TVList pair : memSeries) {
```
qiaojialin:
```suggestion
    for (TVList tvlist : memSeries) {
```

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileProcessor.java
```diff
@@ -583,28 +587,28 @@ public String getStorageGroupName() {
       String measurementId, TSDataType dataType, Map<String, String> props, QueryContext context) {
     flushQueryLock.readLock().lock();
     try {
-      MemSeriesLazyMerger memSeriesLazyMerger = new MemSeriesLazyMerger();
+      List<TVList> tvLists = new ArrayList<>();
```
qiaojialin: replace this with List<ReadOnlyMemchunk>

server/src/main/java/org/apache/iotdb/db/query/reader/universal/PriorityMergeReader.java
```diff
@@ -42,6 +47,11 @@
   public PriorityMergeReader() {
   }
 
+  public PriorityMergeReader(int floatPrecision) {
```
qiaojialin: If we use a List<ReadOnlyMemchunk> , this class should not be changed.
is this a bug before?

server/src/main/java/org/apache/iotdb/db/utils/datastructure/TVList.java
```diff
@@ -452,11 +470,62 @@ void updateMinTimeAndSorted(long[] time, int start, int end) {
...
+        hasCachedPair = true;
+        cur++;
+        return true;
+      }
+      return hasCachedPair;
+    }
+
+    @Override
+    public TimeValuePair nextTimeValuePair() {
+      hasCachedPair = false;
```
qiaojialin: if (hasCachedPair || hasNextTimeValuePair) {
    hasCachedPair = false
   return cachedTimeValuePair
} else {
   throw new IOException("no next time value pair");
}

server/src/main/java/org/apache/iotdb/db/engine/memtable/WritableMemChunk.java
liutaohua: done

server/src/main/java/org/apache/iotdb/db/engine/memtable/AbstractMemTable.java
liutaohua: done

server/src/main/java/org/apache/iotdb/db/engine/querycontext/ReadOnlyMemChunk.java
liutaohua: done

server/src/main/java/org/apache/iotdb/db/utils/datastructure/TVList.java
liutaohua: done

server/src/main/java/org/apache/iotdb/db/engine/querycontext/ReadOnlyMemChunk.java
liutaohua: this is a pretty nice design

server/src/main/java/org/apache/iotdb/db/query/reader/universal/PriorityMergeReader.java
liutaohua: It's for formatting data in memory

PR #760, opened 2020-02-02
[IOTDB-449] Manage TsFileSequenceReader in query by FileReaderManager

liutaohua:
Manage all TsFileSequenceReader in FileReaderManager when query


server/src/main/java/org/apache/iotdb/db/query/reader/seriesRelated/SeriesReader.java
```diff
@@ -482,13 +476,7 @@ public VersionPair(long version, T data) {
...
 
-  public void closeReader() throws IOException {
-    if (firstChunkMetaData != null) {
-      firstChunkMetaData.getChunkLoader().close();
-    }
-    for (IChunkLoader openedChunkLoader : openedChunkLoaders) {
-      openedChunkLoader.close();
-    }
+  public void closeReader() {
   }
```
JackieTien97: It seems that you can safely delete this method.

liutaohua: done

PR #762, opened 2020-02-03
[IOTDB-445] Unify the keyword of "timestamp" and "time"

Genius-pig:
use both timestamp and time.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/qp/strategy/LogicalGenerator.java
+++ b/server/src/main/java/org/apache/iotdb/db/qp/strategy/LogicalGenerator.java
@@ -1167,12 +1167,15 @@ public class LogicalGenerator extends SqlBaseBaseListener {
       return parseOrExpression(ctx.orExpression());
     } else {
       Path path = null;
-      BasicFunctionOperator basic = null;
+      BasicFunctionOperator basic;
       if (ctx.prefixPath() != null) {
         path = parsePrefixPath(ctx.prefixPath());
       }
       if (ctx.suffixPath() != null) {
         path = parseSuffixPath(ctx.suffixPath());
+      }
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/apache/iotdb/db/qp/strategy/LogicalGenerator.java
```diff
@@ -1167,13 +1167,16 @@ private FilterOperator parsePredicate(PredicateContext ctx) {
...
-      BasicFunctionOperator basic = null;
+      BasicFunctionOperator basic;
       if (ctx.prefixPath() != null) {
         path = parsePrefixPath(ctx.prefixPath());
       }
       if (ctx.suffixPath() != null) {
         path = parseSuffixPath(ctx.suffixPath());
       }
+      if(ctx.TIME() != null || ctx.TIMESTAMP() != null) {
+        path = new Path(SQLConstant.RESERVED_TIME);
```
jixuan1989: I am not familiar with this part but there is one question, why there is no code about `path = new Path(SQLConstant.RESERVED_TIME);` in current master branch? (Or, why you add such a line of code?).
If no such codes, how current master branch support `select ..... where time > ...`?

qiaojialin: I guess the current master branch treats "time" as a time series path, it is not distinguished here

jixuan1989: OK.. As I am not familiar with this part. I do not know whether it is right... You can ignore my question when you decide whether to merge this pr.

jixuan1989:
The failed  travis-ci/pr shows some potential bugs, though these bugs are not caused by this PR. 
I have downloaded the log.

PR #765, opened 2020-02-04
[IOTDB-335] Merge different aggregations on same timeseries

samperson1997:
Separate query executions of the same timeseries with different aggregate functions can be optimized

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/query/executor/AggregationExecutor.java
+++ b/server/src/main/java/org/apache/iotdb/db/query/executor/AggregationExecutor.java
@@ -19,6 +19,12 @@
 
 package org.apache.iotdb.db.query.executor;
 
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import org.apache.iotdb.db.conf.IoTDBDescriptor;
 import org.apache.iotdb.db.exception.StorageEngineException;
 import org.apache.iotdb.db.exception.path.PathException;
@@ -31,8 +37,8 @@ import org.apache.iotdb.db.query.dataset.SingleDataSet;
 import org.apache.iotdb.db.query.factory.AggreResultFactory;
 import org.apache.iotdb.db.query.reader.IReaderByTimestamp;
 import org.apache.iotdb.db.query.reader.seriesRelated.ISeriesReader;
-import org.apache.iotdb.db.query.reader.seriesRelated.SeriesReaderByTimestamp;
 import org.apache.iotdb.db.query.reader.seriesRelated.SeriesReader;
+import org.apache.iotdb.db.query.reader.seriesRelated.SeriesReaderByTimestamp;
 import org.apache.iotdb.db.query.timegenerator.EngineTimeGenerator;
 import org.apache.iotdb.tsfile.file.metadata.enums.TSDataType;
 import org.apache.iotdb.tsfile.file.metadata.statistics.Statistics;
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/iotdb/db/query/executor/AggregationExecutor.java
```diff
@@ -121,23 +127,21 @@ private AggregateResult aggregateOneSeries(int i, Filter timeFilter, QueryContex
         //cal by pageheader
         if (seriesReader.canUseCurrentPageStatistics()) {
           Statistics pageStatistic = seriesReader.currentPageStatistics();
-          aggregateResult.updateResultFromStatistics(pageStatistic);
-          if (aggregateResult.isCalculatedAggregationResult()) {
```
qiaojialin: If one series relates to a   min_time and first_value, we'd better return result if all aggregateResult  has calculated aggregation results.

server/src/main/java/org/apache/iotdb/db/query/executor/AggregationExecutor.java
```diff
@@ -80,39 +82,43 @@ public QueryDataSet executeWithoutValueFilter(QueryContext context)
...
-  }
 
+    return constructDataSet(Arrays.asList(aggregateResultList));
+  }
 
-  /**
-   * get aggregation result for one series
-   */
-  private AggregateResult aggregateOneSeries(int i, Filter timeFilter, QueryContext context)
+  private List<AggregateResult> aggregateOneSeries(Map.Entry<Path, List<Integer>> series,
```
qiaojialin: add javadoc and an example

given : 
paths: s1, s2, s3, s1
aggregations: count, sum, count, sum

seriesMap:  
s1 -> 0, 3
s2 -> 2
s3 -> 3

samperson1997: Hi, thanks! I have intended put the example to `mergeSameSeries` method

qiaojialin: this function could be renamed to  groupAggregationsBySeries

server/src/main/java/org/apache/iotdb/db/query/executor/AggregationExecutor.java
samperson1997: Thanks for your proposal. I use an `aggregateResultList` to store whether the `aggregateResult` is calculated... which is consistent with the `aggregateResultList`. Do you think it is okay?

server/src/main/java/org/apache/iotdb/db/query/executor/AggregationExecutor.java
samperson1997: Done.

server/src/main/java/org/apache/iotdb/db/query/executor/AggregationExecutor.java
```diff
@@ -129,10 +129,15 @@ public QueryDataSet executeWithoutValueFilter(QueryContext context)
...
-          aggregateResult.updateResultFromStatistics(chunkStatistics);
-          if (aggregateResult.isCalculatedAggregationResult()) {
-            isCalculatedList.set(i, true);
+          if (Boolean.FALSE.equals(isCalculatedList.get(i))) {
+            AggregateResult aggregateResult = aggregateResultList.get(i);
+            aggregateResult.updateResultFromStatistics(chunkStatistics);
+            if (aggregateResult.isCalculatedAggregationResult()) {
+              isCalculatedList.set(i, true);
+            }
+            if (isCalculatedList.stream().allMatch(element -> element == Boolean.TRUE)) {
```
qiaojialin: replace this with an counter

PR #767, opened 2020-02-04
Fix sync schema pos bug

fanhualta:


qiaojialin:
In which case it is null?

fanhualta:
> In which case it is null?

After the file is created for the first time, the system fails before the pos data is written, and then the file will be empty when the IoTDB system is restarted, which will lead to null.

PR #773, opened 2020-02-06
Add SeriesReaderByTimestampTest for SeriesReaderByTimestamp

jack870131:
Add the unit test for SeriesReaderByTimestamp.


jack870131:
> I think it's better to extract the prepare methods to another test util class.

@JackieTien97 @qiaojialin I have added a util class for preparing series and files. Please have a look. Thx.

server/src/test/java/org/apache/iotdb/db/query/reader/series/SeriesReaderByTimestampTest.java
```diff
@@ -0,0 +1,65 @@
+package org.apache.iotdb.db.query.reader.series;
```
qiaojialin: add apache header

jack870131: @qiaojialin Done.

PR #784, opened 2020-02-10
[IOTDB-470]fix IllegalArgumentException when there exists 0 byte TsFile

EJTTianYu:


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -265,9 +265,17 @@ private void recover() throws StorageGroupProcessorException {
       recoverUnseqFiles(unseqTsFiles);
 
       for (TsFileResource resource : seqTsFiles) {
+        //After recover, case the TsFile's length is equal to 0, delete both the TsFileResource and the file itself
+        if (resource.getFile().length() == 0) {
+          deleteTsfile(resource.getFile());
```
jixuan1989: resource.remove() seems better.

EJTTianYu: it seems that resource.remove() just remove the related file, while deleteTsFile method could also remove the TsFileResource from sequenceFileList/unSequenceFileList.
Does we need to reserve the TsFileResource when the TsFile is 0 byte length

qiaojialin: If the tsfile length is 0. the resource should be removed from the resourceList, otherwise exception may occur in the query process

PR #789, opened 2020-02-11
Self-serve modification of git topics

liutaohua:


jixuan1989:
Wow, so it works in this way?
how about adding a TSDB label?

liutaohua:
> Wow, so it works in this way?

 right 💯

PR #790, opened 2020-02-11
[IoTDB-468] Restructure QueryPlan

Alima777:
The AlignByDevice query ( which is called groupByDevice before ) and general query aligning by time are both storaged in QueryPlan now. However, they invoke different query process bascially, so it's necessary to restructure the QueryPlan and seperate the AlignByDeviceQuery from it. 

The QueryPlan is designed as base class, which has AlignByTimePlan and AlignByDevicePlan as subclasses. This restructure may bring some usage changes, because the general query which uses QueryPlan directly before has to instantiate AlignByTimePlan now.

If you think the design is not friendly for users, welcome to comment.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/qp/physical/crud/AlignByDevicePlan.java
+++ b/server/src/main/java/org/apache/iotdb/db/qp/physical/crud/AlignByDevicePlan.java
@@ -1,0 +1,162 @@
+package org.apache.iotdb.db.qp.physical.crud;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import org.apache.iotdb.db.qp.logical.Operator;
+import org.apache.iotdb.db.qp.logical.Operator.OperatorType;
+import org.apache.iotdb.tsfile.file.metadata.enums.TSDataType;
+import org.apache.iotdb.tsfile.read.expression.IExpression;
+
+public class AlignByDevicePlan extends QueryPlan {
+
+  private List<String> measurements; // for group by device sql, e.g. temperature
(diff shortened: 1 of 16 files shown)


server/src/main/java/org/apache/iotdb/db/qp/physical/crud/AlignByDevicePlan.java
```diff
@@ -0,0 +1,162 @@
+package org.apache.iotdb.db.qp.physical.crud;
```
samperson1997: Hi, don't forget the Apache license.

server/src/main/java/org/apache/iotdb/db/qp/physical/crud/AlignByTimePlan.java
```diff
@@ -0,0 +1,58 @@
+package org.apache.iotdb.db.qp.physical.crud;
```
samperson1997: Same above.

qiaojialin:
resubmit to new_series_reader

server/src/main/java/org/apache/iotdb/db/qp/physical/crud/AlignByDevicePlan.java
Alima777: Thank you! I will fix it.

server/src/main/java/org/apache/iotdb/db/qp/physical/crud/AlignByTimePlan.java
Alima777: Thank you! I will fix it.

PR #798, opened 2020-02-13
del ChunkLoaderImpl LruCache 

liutaohua:


jixuan1989:
Hi, the reason for discarding the cache?

liutaohua:
> Hi, the reason for discarding the cache?

Chunkloader has a problem with the use it , it is `new` in most cases, so cache will not be used! And a huge cache will cause OOM. 
This issue has been fixed and submitted by Jackie tien

PR #802, opened 2020-02-14
Cached chunk

JackieTien97:


server/src/main/java/org/apache/iotdb/db/engine/cache/ChunkCache.java
```diff
@@ -0,0 +1,153 @@
...
+
+    cacheRequestNum.incrementAndGet();
+
+    try {
+      readLock.lock();
+      if (lruCache.containsKey(chunkMetaData)) {
+        cacheHitNum.incrementAndGet();
+        printCacheLog(true);
+        Chunk chunk = lruCache.get(chunkMetaData);
+        return new Chunk(chunk.getHeader(), chunk.getData().duplicate(), chunk.getDeletedAt(), reader.getEndianType());
```
qiaojialin: why not return the chunk got from cache?

server/src/main/java/org/apache/iotdb/db/engine/cache/ChunkCache.java
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  public static ChunkCache getInstance() {
+    return ChunkCacheHolder.INSTANCE;
+  }
+
+  public Chunk get(ChunkMetaData chunkMetaData, TsFileSequenceReader reader) throws IOException {
+    if (!cacheEnable) {
+      Chunk chunk = reader.readMemChunk(chunkMetaData);
+      return new Chunk(chunk.getHeader(), chunk.getData().duplicate(), chunk.getDeletedAt(), reader.getEndianType());
```
qiaojialin: why new a Chunk?

server/src/main/java/org/apache/iotdb/db/engine/cache/ChunkCache.java
```diff
@@ -0,0 +1,153 @@
...
+      printCacheLog(true);
+      Chunk chunk = lruCache.get(chunkMetaData);
+      readLock.unlock();
+      return new Chunk(chunk.getHeader(), chunk.getData().duplicate(), chunk.getDeletedAt(), reader.getEndianType());
+    }
+    printCacheLog(false);
+    Chunk chunk = reader.readMemChunk(chunkMetaData);
+    lruCache.put(chunkMetaData, chunk);
+    writeLock.unlock();
+    return new Chunk(chunk.getHeader(), chunk.getData().duplicate(), chunk.getDeletedAt(), reader.getEndianType());
```
qiaojialin:
```suggestion
    return chunk;
```

server/src/main/java/org/apache/iotdb/db/engine/cache/ChunkCache.java
JackieTien97: The chunk has a field byte buffer, it should be reused.

server/src/main/java/org/apache/iotdb/db/engine/cache/ChunkCache.java
```diff
@@ -0,0 +1,153 @@
...
+  private final ReadWriteLock lock = new ReentrantReadWriteLock();
+  private final Lock readLock = lock.readLock();
+  private final Lock writeLock = lock.writeLock();
+
+
+  private ChunkCache() {
+    lruCache = new LRULinkedHashMap<ChunkMetaData, Chunk>(MEMORY_THRESHOLD_IN_CHUNK_CACHE, true) {
+      @Override
+      protected long calEntrySize(ChunkMetaData key, Chunk value) {
+        return RamUsageEstimator.sizeOf(value) + RamUsageEstimator.sizeOf(value);
```
liutaohua: why this return valueSize * 2 ?

server/src/main/java/org/apache/iotdb/db/engine/cache/ChunkCache.java
```diff
@@ -0,0 +1,153 @@
...
+        Chunk chunk = lruCache.get(chunkMetaData);
+        return new Chunk(chunk.getHeader(), chunk.getData().duplicate(), chunk.getDeletedAt(), reader.getEndianType());
+      }
+    } finally {
+      readLock.unlock();
+    }
+
+    writeLock.lock();
+    if (lruCache.containsKey(chunkMetaData)) {
+      writeLock.unlock();
```
liutaohua: Why not down the lock to readLock

server/src/main/java/org/apache/iotdb/db/query/reader/chunk/DiskChunkLoader.java
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  @Override
+  public void close() throws IOException {
+    reader.close();
+  }
+
+  @Override
+  public void clear() {
+    // no cache need clear
```
liutaohua: i think , this method can delete for  interface, neither memChunk nor diskChunk don't need to clear ,and the cache is  LRU

server/src/main/java/org/apache/iotdb/db/engine/cache/ChunkCache.java
JackieTien97: it's a typo, should be RamUsageEstimator.sizeOf(key) + RamUsageEstimator.sizeOf(value);
Fixed

PR #804, opened 2020-02-14
[IOTDB-467] The CLI displays query results in a batch manner

liutaohua:
If the query result set has too many lines. Each time CLI prints 1000 lines and waits for the command of the user. If the user enters an "ENTER", the CLI prints another 1000 line. If the user enters a "q", the CLI stop to print the Result and wait for the next command.

### DIFF
--- a/client/src/main/java/org/apache/iotdb/client/AbstractClient.java
+++ b/client/src/main/java/org/apache/iotdb/client/AbstractClient.java
@@ -106,6 +106,7 @@ public abstract class AbstractClient {
   static ServerProperties properties = null;
 
   private static final PrintStream SCREEN_PRINTER = new PrintStream(System.out);
+  private static final Scanner SCREEN_INPUT = new Scanner(System.in);
 
   static void init() {
     keywordSet.add("-" + HOST_ARGS);
@@ -124,7 +125,7 @@ public abstract class AbstractClient {
   }
 
   private static void printCount(int cnt) {
-      println("Total line number = " + cnt);
+    println("Total line number = " + cnt);
   }
 
   static Options createOptions() {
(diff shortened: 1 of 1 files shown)


client/src/main/java/org/apache/iotdb/client/AbstractClient.java
```diff
@@ -532,45 +539,61 @@ private static void executeQuery(IoTDBConnection connection, String cmd) {
...
+   * @return List<List < String>> result
    * @throws SQLException throw exception
    */
   private static List<List<String>> cacheResult(ResultSet resultSet, List<Integer> maxSizeList,
       int columnCount, ResultSetMetaData resultSetMetaData, ZoneId zoneId)
       throws SQLException {
     boolean printTimestamp = !((IoTDBQueryResultSet) resultSet).isIgnoreTimeStamp();
     List<List<String>> lists = new ArrayList<>(columnCount);
-    for(int i = 1; i <= columnCount; i++) {
-      List<String> list = new ArrayList<>(1001);
```
Genius-pig: change it to 1001 or maxPrintRowCount + 1. It has one more line that is name header row.

client/src/main/java/org/apache/iotdb/client/AbstractClient.java
```diff
@@ -532,45 +539,61 @@ private static void executeQuery(IoTDBConnection connection, String cmd) {
...
+  private static void printNextBatchResult(ZoneId zoneId, ResultSet resultSet,
+      ResultSetMetaData resultSetMetaData, int columnLength, List<Integer> maxSizeList,
+      int preLinesCount)
+      throws SQLException {
+    maxSizeList.clear();
+    List<List<String>> lists = cacheResult(resultSet, maxSizeList, columnLength,
+        resultSetMetaData, zoneId);
+    output(lists, maxSizeList);
+    if (isReachMaxLine) {
+      println(String.format("Reach maxPrintRowCount = %s lines, enter q to exit, other to continue",
```
qiaojialin:
```suggestion
      println(String.format("Printed %s lines, enter q to exit, other to continue",
```

PR #806, opened 2020-02-15
[IOTDB-467] The CLI displays query results in a batch manner

dynamicDr:


Genius-pig:
Thank you for your contribution. But it has been done by https://github.com/apache/incubator-iotdb/pull/804

liutaohua:
> Thank you for your contribution. But it has been done by #804

welcome for new contributor ，#804 was closed

client/src/main/java/org/apache/iotdb/client/AbstractClient.java
```diff
@@ -90,8 +93,9 @@
   static int maxTimeLength = ISO_DATETIME_LEN;
   static int maxValueLength = 15;
   static String TIMESTAMP_PRECISION = "ms";
+  static int lineCount = 0;
 
-  private static boolean isReachMaxLine = false;
+  private static boolean isReachEnd = false;
```
qiaojialin:
```suggestion
  private static boolean isReachEnd = true;
```

dynamicDr:
> Hi, thanks for your contribution! The code is clean and the CLI has a good user experience.
> 
> Three suggestions:
> 
> (1) if you use the IDEA and mac, please import the java-google-style.xml and option+command+L to format the code.
> (2) Fortunately, the other PR for the same issue is closed by the author :) In order not to do duplicated work, it's better to leave a comment on the Jira to claim that you will solve the issue.
> (3) We have a weixin group to communicate, you could join through my weixin: qiaojialin1993

Thanks for your advice!
I've made some changes on it.

PR #819, opened 2020-02-17
[IOTDB-496]An extra line in CLI when the result is empty

dynamicDr:


client/src/main/java/org/apache/iotdb/client/AbstractClient.java
```diff
@@ -607,13 +612,16 @@ private static void executeQuery(IoTDBConnection connection, String cmd) {
   }
 
   private static void output(List<List<String>> lists, List<Integer> maxSizeList) {
-    printBlockLine(maxSizeList);
-    printRow(lists, 0, maxSizeList);
-    printBlockLine(maxSizeList);
-    for (int i = 1; i < lists.get(0).size(); i++) {
-      printRow(lists, i, maxSizeList);
+    int size = lists.get(0).size();
```
qiaojialin: Hi, please print the table header information if it's an empty set.

PR #827, opened 2020-02-20
[IOTDB-501] let Travis CI more stable

jixuan1989:
This PR solves some issues in UTs and ITs.

1. [not solved now] Port 8181, 6667, and 31999 can not be binded because they are used.
2. [Solved] ActiveTimeSeriesCounter offer and update methods throw null.
3. [Solved] Hive and Hadoop print debug/info log
4. [Solved] Some folders can not be deleted because the folder is not empty.


jixuan1989:
This PR leaves a known issue case:
If a `delete timeseries` is called and the deletion command will remove all time series in a SG, then the data files in the SG will be deleted,
At this time, if there are queries on these time series at the same time, then the data files readers are not closed now.  And it will lead to deleting files failed.

PR #833, opened 2020-02-24
[IOTDB-514] Extend aggregation

yx-zhang:


server/src/main/java/org/apache/iotdb/db/query/aggregation/AggregateResult.java
```diff
@@ -209,4 +209,6 @@ protected void setBinaryValue(Binary binaryValue) {
   protected boolean hasResult() {
     return hasResult;
   }
+
+  abstract protected void merge(AggregateResult another);
```
qiaojialin: make it public

PR #837, opened 2020-02-25
[IOTDB-444] Enhanced wildcard query on device path

samperson1997:
Now we have `show child paths PATH_PREFIX` to show the children nodes of the PREFIX PATH. However, sometimes we need more complex queries, like
1. Show all path prefixes (level6) which start with "root.a.b.*.d" (level1=root, level2=a, and level3=b, level4 can be everything, and level5=d).
2. Show all path prefixes which start with "root.a.b*" (level1= root, level2=a and level3 starts with b).
3. Combine 1 and 2 together. e.g., "root.a.b*.\*.d.\*", etc..
4. For select queries, it is also good to support wildcard like "select * from root.a.\*.b\*" (which means that the level4 starts with 'b')

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
@@ -33,6 +33,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.TreeSet;
 import org.apache.iotdb.db.conf.IoTDBConstant;
 import org.apache.iotdb.db.conf.IoTDBDescriptor;
 import org.apache.iotdb.db.exception.metadata.IllegalPathException;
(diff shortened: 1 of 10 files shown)


server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
```diff
@@ -503,13 +504,20 @@ private void findPath(MNode node, String[] nodes, int idx, String parent,
...
-    if (!(PATH_WILDCARD).equals(nodeReg)) {
+    if (!nodeReg.contains(PATH_WILDCARD)) {
       if (node.hasChild(nodeReg)) {
         findPath(node.getChild(nodeReg), nodes, idx + 1, parent + node.getName() + PATH_SEPARATOR,
             timeseriesSchemaList);
       }
     } else {
       for (MNode child : node.getChildren().values()) {
+        String prefix = "";
+        if (!nodeReg.equals(PATH_WILDCARD)) {
```
qiaojialin: use regex match

samperson1997: Thanks! Updated in the latest commit.

PR #838, opened 2020-02-26
[IOTDB-429] return empty dataset instead of throw exception

qiaojialin:


jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBResultMetadata.java
```diff
@@ -59,10 +59,7 @@ public String getColumnClassName(int arg0) throws SQLException {
   }
 
   @Override
-  public int getColumnCount() throws SQLException {
-    if (columnInfoList == null || columnInfoList.isEmpty()) {
-      throw new SQLException("No column exists");
-    }
+  public int getColumnCount() {
     return columnInfoList.size();
```
JackieTien97: What if columnInfoList is null? I think you can return 0 if it is null.

jdbc/src/test/java/org/apache/iotdb/jdbc/IoTDBResultMetadataTest.java
```diff
@@ -120,7 +124,7 @@ public void testGetColumnType() throws SQLException {
     } catch (Exception e) {
       flag = true;
     }
-    assertEquals(flag, true);
+    assertTrue(flag);
```
JackieTien97: If you also return 0 when it's null, like I suggested above, don't forget to change assert here.

server/src/test/java/org/apache/iotdb/db/integration/IoTDBSimpleQueryTest.java
```diff
@@ -41,6 +44,62 @@ public void tearDown() throws Exception {
...
+      Assert.assertEquals(1, resultSet.getMetaData().getColumnCount());
+      while(resultSet.next()) {
+        fail();
+      }
+
+      resultSet = statement.executeQuery(
+          "select count(*) from root where time >= 1 and time <= 100 "
+              + "group by ([0, 100), 20ms, 20ms) align by device");
+      // has an empty time column
+      Assert.assertEquals(2, resultSet.getMetaData().getColumnCount());
```
JackieTien97: The comment here should be has an empty time column and device column

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
```diff
@@ -228,9 +229,12 @@ private QueryDataSet processDataQuery(QueryPlan queryPlan, QueryContext context)
...
       queryDataSet = new AlignByDeviceDataSet((AlignByDevicePlan) queryPlan, context, queryRouter);
     } else {
-      if (queryPlan instanceof GroupByPlan) {
+      if (queryPlan.getPaths() == null || queryPlan.getPaths().isEmpty()) {
+        // no time series are selected, return EmptyDataSet
+        return new EmptyDataSet();
+      } else if (queryPlan instanceof GroupByPlan) {
         GroupByPlan groupByPlan = (GroupByPlan) queryPlan;
-        return queryRouter.groupBy(groupByPlan, context);
+        queryDataSet =  queryRouter.groupBy(groupByPlan, context);
```
JackieTien97: No need to do this change, the limit and offset are ignored in group by.

JackieTien97: Or you can add a JIRA to report this, actually in syntax, we support limit and offset in group by, but we just ignore these two in the code.

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBResultMetadata.java
samperson1997: I think it is okay to have null `columnInfoList`, which means some error occurs. If the size `columnInfoList` is 0, it means it is an empty list. What do you think?

qiaojialin: Actually, in the normal process, this couldn't be null, but in case others set a null in the future, check null is better

jdbc/src/test/java/org/apache/iotdb/jdbc/IoTDBResultMetadataTest.java
qiaojialin: fixed

server/src/test/java/org/apache/iotdb/db/integration/IoTDBSimpleQueryTest.java
qiaojialin: fixed

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
qiaojialin: I updated the docs

PR #841, opened 2020-02-26
[IOTDB-476] Improve the ExpressionOptimizer

Ring-k:
Some logic in ExpressionOptimizer is modified to make the optimizer generate more efficient expression. For example, before the modification, `s1 > 10 or time < 1` will be tranformed to 
``
ORExpression (
    SingleSeriesExpression(s1 , value >10)
    ORExpression(
        SingleSeriesExpression(s1 , time <10),
        SingleSeriesExpression(s2 , time <10)
    )
)
``

With the new logic, the query will be transformed to 
``
ORExpression (
    SingleSeriesExpression(s1 , value >10 or time <10)
    SingleSeriesExpression(s2 , time <10)
)
``

With less SingleSeriesExpressions, the query can be executed more efficiently. Experiments shows that, if we have 3 devices, each containing 3 sensors, and the query is `t<4838 or d2.s2>0.5`, the latency decreases around 10%.


tsfile/src/main/java/org/apache/iotdb/tsfile/read/expression/util/ExpressionOptimizer.java
```diff
@@ -108,13 +108,66 @@ private IExpression handleOneGlobalTimeFilter(GlobalTimeExpression globalTimeExp
...
+   * @param rightExpression The IExpression to be merged into the first IExpression
+   * @return a merged IExpression, which is initially based on the input leftExpression
+   */
+  private IExpression mergeSecondTreeToFirstTree(IExpression leftExpression,
+      IExpression rightExpression) {
+    if (rightExpression.getType() == ExpressionType.SERIES) {
+      SingleSeriesExpression leaf = (SingleSeriesExpression) rightExpression;
+      updateFilterWithOr(leftExpression, leaf.getFilter(), leaf.getSeriesPath());
+      return leftExpression;
+    } else if (rightExpression.getType() == ExpressionType.SERIES) {
```
SilverNarcissus: Two of if are same, please check the logic

tsfile/src/main/java/org/apache/iotdb/tsfile/read/expression/util/ExpressionOptimizer.java
```diff
@@ -108,13 +108,66 @@ private IExpression handleOneGlobalTimeFilter(GlobalTimeExpression globalTimeExp
...
+    } else {
+      return BinaryExpression.or(leftExpression, rightExpression);
+    }
+  }
+
+  /**
+   * This method search  the node in the input expression, whose path is identical to the input
+   * path, then merges its filter and the input filter with relation OR.
+   *
+   * @return an expression, some of whose node contains the input filter.
```
SilverNarcissus: The return value is boolean but not an expression

tsfile/src/main/java/org/apache/iotdb/tsfile/read/expression/util/ExpressionOptimizer.java
Ring-k: It's a mistake. Thanks for your reminding.

tsfile/src/main/java/org/apache/iotdb/tsfile/read/expression/util/ExpressionOptimizer.java
Ring-k: The return value should be of boolean type. I will modify the it. Thanks for your suggestion.

PR #846, opened 2020-02-26
[IOTDB-512] file reader close bug

JackieTien97:


jixuan1989:
The reported bug by SonarCloud is not an real issue. 
But it is better if you can add some comments to notify readers, (and use ignore annotation to ignore it).

As for the failed test in Travis CI + Win, how do you think?
Is that a bug in the Merge module?

server/src/main/java/org/apache/iotdb/db/query/control/FileReaderManager.java
```diff
@@ -199,6 +202,26 @@ void decreaseFileReaderReference(TsFileResource tsFile, boolean isClosed) {
    * integration tests will not conflict with each other.
    */
   public synchronized void closeAndRemoveAllOpenedReaders() throws IOException {
+    Iterator<Map.Entry<TsFileResource, TsFileSequenceReader>> iterator = closedFileReaderMap.entrySet().iterator();
+    while (iterator.hasNext()) {
+      Map.Entry<TsFileResource, TsFileSequenceReader> entry = iterator.next();
+      entry.getValue().close();
+      if (resourceLogger.isInfoEnabled()) {
```
qiaojialin: change to debug

server/src/main/java/org/apache/iotdb/db/query/control/FileReaderManager.java
```diff
@@ -199,6 +202,26 @@ void decreaseFileReaderReference(TsFileResource tsFile, boolean isClosed) {
...
+        resourceLogger.info("{} closedTsFileReader is closed.", entry.getValue().getFileName());
+      }
+      closedReferenceMap.remove(entry.getKey());
+      iterator.remove();
```
qiaojialin: duplicated?

server/src/main/java/org/apache/iotdb/db/query/control/FileReaderManager.java
```diff
@@ -199,6 +202,26 @@ void decreaseFileReaderReference(TsFileResource tsFile, boolean isClosed) {
...
+    }
+    iterator = unclosedFileReaderMap.entrySet().iterator();
+    while (iterator.hasNext()) {
+      Map.Entry<TsFileResource, TsFileSequenceReader> entry = iterator.next();
+      entry.getValue().close();
+      if (resourceLogger.isInfoEnabled()) {
+        resourceLogger.info("{} unclosedTsFileReader is closed.", entry.getValue().getFileName());
+      }
+      unclosedReferenceMap.remove(entry.getKey());
+      iterator.remove();
```
qiaojialin: duplicated

server/src/main/java/org/apache/iotdb/db/query/control/FileReaderManager.java
```diff
@@ -199,6 +202,26 @@ void decreaseFileReaderReference(TsFileResource tsFile, boolean isClosed) {
...
+    }
     for (Map.Entry<TsFileResource, TsFileSequenceReader> entry : closedFileReaderMap.entrySet()) {
```
qiaojialin: this should be removed..

server/src/main/java/org/apache/iotdb/db/query/control/QueryFileManager.java
```diff
@@ -77,8 +74,9 @@ private void addUsedFilesForQuery(long queryId, List<TsFileResource> resources)
       // this file may be deleted just before we lock it
       if (tsFileResource.isDeleted()) {
         Map<Long, Set<TsFileResource>> pathMap = !isClosed ? unsealedFilePathsMap : sealedFilePathsMap;
-        pathMap.get(queryId).remove(tsFileResource);
-        FileReaderManager.getInstance().decreaseFileReaderReference(tsFileResource, isClosed);
+        if (pathMap.get(queryId).remove(tsFileResource)) {
```
qiaojialin: what If does not contain the tsfileResource, is it a bug?

server/src/main/java/org/apache/iotdb/db/query/control/FileReaderManager.java
JackieTien97: different map. iterator is the closedFileReaderMap's iterator.

server/src/main/java/org/apache/iotdb/db/query/control/QueryFileManager.java
JackieTien97: Nope, It may have been decreased because of it's deleted, but another thread may also enter into the same block, it also remove it(although it's an empty operation, the resource has been removed by another thread, so the remove operation return false), in this case, the second thread should not decrease it again.

PR #865, opened 2020-03-01
Several fixes.

JulianFeinauer:
- Fixed an issue in Utils and added test
- Fixed BasicDaoImpl (query was invalid) in IoTDB-Grafana

### DIFF
--- a/grafana/src/main/java/org/apache/iotdb/web/grafana/dao/impl/BasicDaoImpl.java
+++ b/grafana/src/main/java/org/apache/iotdb/web/grafana/dao/impl/BasicDaoImpl.java
@@ -18,20 +18,6 @@
  */
 package org.apache.iotdb.web.grafana.dao.impl;
 
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.sql.Connection;
-import java.sql.DatabaseMetaData;
-import java.sql.ResultSet;
-import java.sql.SQLException;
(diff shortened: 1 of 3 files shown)


grafana/src/main/java/org/apache/iotdb/web/grafana/dao/impl/BasicDaoImpl.java
```diff
@@ -104,9 +103,10 @@ public Object doInConnection(Connection connection) throws SQLException {
   public List<TimeValues> querySeries(String s, Pair<ZonedDateTime, ZonedDateTime> timeRange) {
     Long from = zonedCovertToLong(timeRange.left);
     Long to = zonedCovertToLong(timeRange.right);
-    String sql = "SELECT " + s.substring(s.lastIndexOf('.') + 1) + " FROM root."
-        + s.substring(0, s.lastIndexOf('.')) + " WHERE time > " + from * TIMESTAMP_RADIX
-        + " and time < " + to * TIMESTAMP_RADIX;
+    // How many rows will the result have?
```
jixuan1989: currently, we can not get how many rows before we get the whole ResultSet.
===

 I have an idea, we may implement it by using the summaries (count of each page/chunk) info in IoTDB. 

@qiaojialin , how do you think?

qiaojialin: do you want to get the precise result or a fuzy one?

PR #866, opened 2020-03-02
Cherry pick the changes of the distributed version

jt2594838:
To support the distributed version, some changes have been made to two modules, server, and TsFile.  Unfortunately, such changes often conflict with those who are working on the master branch. So I would like to merge such changes into master ahead to reduce further conflicts.

The changes are mainly extractions of interfaces and code fixes and should not affect the normal functionality.


server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
```diff
@@ -611,6 +614,14 @@ private String getHdfsDir() {
     return dataDirs;
   }
 
+  public boolean isEnableMetricService() {
```
jixuan1989: there is a field called `private boolean enableMetricsWebService = true;` You need to delete `enableMetricsWebService` and use your field.

qiaojialin: Hi, I will fix this

PR #871, opened 2020-03-02
[IOTDB-529] Relative times and NOW() operator cannot be used in Group By

HTHou:
We can use relative times and NOW() operator in Group By clause. 

Screen Shot 2020-03-02 at 11 02 00 PM


jixuan1989:
Hi, can you add a UT for that?
@JulianFeinauer , you can merge it into local master branch if you really need it in a hurry.
I will merge this PR after a UT is added.

HTHou:
> Hi, can you add a UT for that?
Added. Thank you so much for reminding.

server/src/test/java/org/apache/iotdb/db/integration/IOTDBGroupByIT.java
```diff
@@ -506,6 +506,45 @@ public void countSumAvgInnerIntervalTest() {
     }
   }
 
+  @Test
+  public void countSumAvgRelativeTimesTest() {
```
jixuan1989: Ok.. This UT is for testing that there is no data.
How about add one more like: 
```java
@Test
  public void usingNowFunction() {

    try (Connection connection = DriverManager.
        getConnection("jdbc:iotdb://127.0.0.1:6667/", "root", "root");
      boolean hasResultSet = statement.execute(
          "select count(temperature), sum(temperature), avg(temperature) from "
              + "root.ln.wf01.wt01 where time > 3"
              + "GROUP BY ([1, 30), 0ms)");
      fail();
    } catch (Exception e) {
      assertTrue(e instanceof IoTDBSQLException);
    }

    try (Connection connection = DriverManager.
        getConnection("jdbc:iotdb://127.0.0.1:6667/", "root", "root");
      statement.execute("INSERT INTO root.ln.wf01.wt01(timestamp,temperature,status, hardware) "
          + "values(now(), 35.5, false, 650)");
      ResultSet resultSet = statement.executeQuery(
          "select count(temperature), sum(temperature), avg(temperature) from "
              + "root.ln.wf01.wt01 "
              + "GROUP BY ([now() - 1h, now()), 1h)");
      assertTrue(resultSet.next());
      //resultSet.getLong(1) is the timestamp
      assertEquals(1, Integer.valueOf(resultSet.getString(2)).intValue());
      assertEquals(35.5, Float.valueOf(resultSet.getString(3)).floatValue(), 0.01);
      assertEquals(35.5, Double.valueOf(resultSet.getString(4)).doubleValue(), 0.01);

    } catch (Exception e) {
      e.printStackTrace();
      fail(e.getMessage());
    }
  }
```

HTHou: Thank you. I added your test code into this PR.

PR #876, opened 2020-03-03
[IOTDB-430] fix a bug when cache non-align resultSet

HTHou:
Screen Shot 2020-03-03 at 8 29 18 PM

When the times and the values are null, I'm not sure if it is good to show the string `null` as above. 

Please share your opinion about that. Thanks!


qiaojialin:
Hi, the table header of disable align will be changed to :  time, path, value
Then, this may not be a problem

PR #879, opened 2020-03-03
Hot fix for LAST query bug when using with disable align

wshao08:
This pr is a hot fix for last query.
1. Last query failed when using with disable align and align by device.
2. Added more integration tests for cases when querying last value among unseqenced files.


server/src/main/java/org/apache/iotdb/db/qp/strategy/PhysicalGenerator.java
```diff
@@ -242,7 +242,12 @@ private PhysicalPlan transformQuery(QueryOperator queryOperator)
     } else {
       queryPlan = new RawDataQueryPlan();
     }
-    if (queryOperator.isAlignByDevice()) {
+    if (queryPlan instanceof LastQueryPlan) {
+      // Last query result set will not be affected by alignment
+      queryPlan.setAlignByTime(true);
```
qiaojialin: false?

PR #880, opened 2020-03-03
[IOTDB-538]add a simple connection pool for session api

jixuan1989:
Well, I think the pool is more easy to use than the description in IOTDB-538.

We provided a connection pool (`SessionPool) for Native API.
Using the interface, you need to define the pool size.
If you can not get a session connection in 60 secondes, there is a warning log but the program will hang.
If a session has finished an operation, it will be put back to the pool automatically.
If a session connection is broken, the session will be removed automatically and the pool will try 
to create a new session and redo the operation.
For query operations:
1. When using SessionPool to query data, the result set is `SessionDataSetWrapper`;
2. Given a `SessionDataSetWrapper`, if you have not scanned all the data in it and stop to use it,
you have to call `SessionPool.closeResultSet(wrapper)` manually;
3. When you call `hasNext()` and `next()` of a `SessionDataSetWrapper` and there is an exception, then
you have to call `SessionPool.closeResultSet(wrapper)` manually;
Examples: ```session/src/test/java/org/apache/iotdb/session/pool/SessionPoolTest.java```


jixuan1989:
I need to add a new api to enable user closing a SessionPool.

session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
```diff
@@ -0,0 +1,670 @@
...
+            }
+            session = queue.poll();
+          }
+        }
+        if (session != null) {
+          return session;
+        }
+      }
+      if (logger.isDebugEnabled()) {
+        logger.error("Create a new Session {}, {}, {}, {}", ip, port, user, password);
```
qiaojialin: should the password be printed?

session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
```diff
@@ -0,0 +1,670 @@
...
+      }
+    }
+  }
+
+  /**
+   * use batch interface to insert sorted data times in row batch must be sorted before!
+   *
+   * @param rowBatch data batch
+   */
+  public TSExecuteBatchStatementResp insertSortedBatch(RowBatch rowBatch)
```
qiaojialin: Is it needed to export the methods of Session in this Pool? the codes are double...

session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
```diff
@@ -0,0 +1,670 @@
...
+  public TSExecuteBatchStatementResp insertBatch(RowBatch rowBatch) throws IoTDBSessionException {
+    while (true) {
+      Session session = getSession();
+      try {
+        TSExecuteBatchStatementResp resp = session.insertBatch(rowBatch);
+        putBack(session);
+        return resp;
+      } catch (IoTDBSessionException e) {
+        if (e.getCause() instanceof TException) {
+          // TException means the connection is broken, remove it and get a new one.
```
qiaojialin: get a new one?

jixuan1989: will fix.

session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
jixuan1989: >  the codes are double...
Ah?

qiaojialin: the interfaces are double...

jixuan1989:
The attachment is a patch for v0.9.1. 
just use `git apply --check THE FILE` to check it is ok and then use `git apply THE FILE` to use it.
[session_pool_0.9.1.patch.txt](https://github.com/apache/incubator-iotdb/files/4327592/session_pool_0.9.1.patch.txt)

qiaojialin:
The Session pool is already merged into rel/0.9 branch. If you want to use it in 0.9.x, rel/0.9 branch is the best choice

PR #886, opened 2020-03-05
[IOTDB-545] Update checkPathValidity in Session

samperson1997:
In session,  support path like: `root.sg`, `root.sg.1.2.3`, `root.sg.d1.1_1`, `root.sg."1.2.3"`, `root.sg.'1.2.3'`

### DIFF
--- a/session/src/main/java/org/apache/iotdb/session/Config.java
+++ b/session/src/main/java/org/apache/iotdb/session/Config.java
@@ -28,11 +28,12 @@ public class Config {
   public static final int DEFAULT_FETCH_SIZE = 10000;
   public static final int DEFAULT_TIMEOUT_MS = 0;
 
-  public static final String NODE_MATCHER = "([a-zA-Z0-9_-]+)";
+  public static final String NODE_MATCHER = "[" + PATH_SEPARATOR + "]" + "([a-zA-Z0-9_]+)";
 
   // for path like: root.sg1.d1."1.2.3" or root.sg1.d1.'1.2.3', only occurs in the end of the path and only occurs once
-  public static final String NODE_WITH_QUOTATION_MARK_MATCHER = "\"|\'a-zA-Z0-9_-[.a-zA-Z0-9_-]+\"|\'";
-  public static final String PATH_MATCHER = PATH_ROOT + "([" + PATH_SEPARATOR + "]" + NODE_MATCHER
-      + ")+[" + NODE_WITH_QUOTATION_MARK_MATCHER + "]?";
+  public static final String NODE_WITH_QUOTATION_MARK_MATCHER =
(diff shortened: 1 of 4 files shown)


session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -654,8 +648,7 @@ public void executeNonQueryStatement(String sql) throws TException, IoTDBRPCExce
   private void checkPathValidity(String path) throws IoTDBSessionException {
```
qiaojialin: add UT for this

samperson1997: Added. Thanks for reminding.

session/src/main/java/org/apache/iotdb/session/Config.java
```diff
@@ -28,7 +28,12 @@
...
   public static final int DEFAULT_TIMEOUT_MS = 0;
 
+  public static final String NODE_MATCHER = "[" + PATH_SEPARATOR + "]" + "([a-zA-Z0-9_]+)";
+
+  // for path like: root.sg1.d1."1.2.3" or root.sg1.d1.'1.2.3', only occurs in the end of the path and only occurs once
+  public static final String NODE_WITH_QUOTATION_MARK_MATCHER =
+      "[" + PATH_SEPARATOR + "][\"|\']([a-zA-Z0-9_]+)(" + NODE_MATCHER + ")+[\"|\']";
   public static final String PATH_MATCHER =
-      PATH_ROOT + "([" + PATH_SEPARATOR + "](([a-zA-Z_][a-zA-Z0-9_-]*)|([+-]?[0-9]+)))+";
+      PATH_ROOT + "(" + NODE_MATCHER + ")+(" + NODE_WITH_QUOTATION_MARK_MATCHER + ")?";
```
vesense: Why not use `Pattern` here?

session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -654,8 +648,7 @@ public void executeNonQueryStatement(String sql) throws TException, IoTDBRPCExce
...
     if (!Pattern.matches(PATH_MATCHER, path)) {
```
vesense: If a pattern is to be used multiple times, compiling it once and reusing it will be more efficient than invoking this method each time.

session/src/main/java/org/apache/iotdb/session/Config.java
samperson1997: Fixed.

session/src/main/java/org/apache/iotdb/session/Session.java
samperson1997: Hi, thanks a lot for this suggestion! That's quite useful, and I have updated it in the latest commit. : )

vesense:
Thanks @samperson1997  LGTM +1

PR #888, opened 2020-03-06
[IOTDB-442]Add session api

SilverNarcissus:
Add some new session api proposed in IOTDB-422


session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -357,6 +392,23 @@ private Object sortList(Object valueList, TSDataType dataType, Integer[] index)
...
+   * or insertBatch method
+   *
+   * @see Session#insertInBatch(List, List, List, List)
+   * @see Session#insertBatch(RowBatch)
+   */
+  public TSStatus insert(String deviceId, long time, List<String> measurements,
+      Object... values) throws IoTDBSessionException {
+    List<String> stringValues = new ArrayList<>();
+    for (Object o : values) {
+      stringValues.add(o.toString());
```
jixuan1989: I think this implementation is just an syntactic sugar...
Is there a possible way to send exact class instances (integer, float, etc..) to the server directly, to avoid the server calling functions like `Integer.parse()`

SilverNarcissus: Yes, but after using byte buffer, the performance goes down. The test result is at: https://tower.im/teams/138444/todos/12089/
If we want to let performance raise, we may refactor InsertPlan.

PR #916, opened 2020-03-17
fix execute flush command while inserting bug

qiaojialin:
When calling flush command. 

StorageEngine call syncCloseAllTsFileProcessors for each StorageGroupProcessor. 
Then, there may be an insert, which causes a new TsFileResource to be added to the list.
Then, StorageEngine call closeAllResource for each StorageGroupProcessor, the newly created resource is marked as closed but it is actually an empty file.

When executing query, the file will be seen as a closed file and read TsFileMetadata, the error occurs.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/StorageEngine.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/StorageEngine.java
@@ -303,7 +303,7 @@ public class StorageEngine implements IService {
   public void syncCloseAllProcessor() {
     logger.info("Start closing all storage group processor");
     for (StorageGroupProcessor processor : processorMap.values()) {
-      processor.syncCloseAllTsFileProcessors();
+      processor.syncCloseAllWorkingTsFileProcessors();
     }
   }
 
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/apache/iotdb/db/engine/StorageEngine.java
```diff
@@ -303,9 +303,7 @@ public void insert(InsertPlan insertPlan)
   public void syncCloseAllProcessor() {
     logger.info("Start closing all storage group processor");
     for (StorageGroupProcessor processor : processorMap.values()) {
-      processor.waitForAllCurrentTsFileProcessorsClosed();
-      //TODO do we need to wait for all merging tasks to be finished here?
-      processor.closeAllResources();
```
JackieTien97: Don't we have to close the file resources? Should we move it to syncCloseAllTsFileProcessors() instead of directly deleting it?

qiaojialin: It's already be closed in closeUnsealedTsFileProcessor() when the close call back

PR #926, opened 2020-03-20
avoid flushing empty memtable

qiaojialin:
When no data is successfully written to memtable because of wrong data type or failed to write WAL. An empty memtable will be flushed.

This pr avoid flushing the empty memtable.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
@@ -282,13 +282,7 @@ public class StorageGroupProcessor {
             .putAll(resource.getEndTimeMap());
         partitionLatestFlushedTimeForEachDevice.computeIfAbsent(timePartitionId, id -> new HashMap<>())
             .putAll(resource.getEndTimeMap());
-
-        for (Map.Entry<String, Long> mapEntry : resource.getEndTimeMap().entrySet()) {
-          if (globalLatestFlushedTimeForEachDevice.getOrDefault(mapEntry.getKey(), Long.MIN_VALUE)
-                  < mapEntry.getValue()) {
-            globalLatestFlushedTimeForEachDevice.put(mapEntry.getKey(), mapEntry.getValue());
-          }
-        }
+        globalLatestFlushedTimeForEachDevice.putAll(resource.getEndTimeMap());
       }
     }
   }
--- a/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
@@ -466,7 +466,7 @@ public class StorageGroupProcessor {
     }
   }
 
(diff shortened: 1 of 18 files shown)


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -282,14 +282,7 @@ private void recover() throws StorageGroupProcessorException {
...
             .putAll(resource.getEndTimeMap());
-
-        for (Map.Entry<String, Long> mapEntry : resource.getEndTimeMap().entrySet()) {
-          if (!globalLatestFlushedTimeForEachDevice.containsKey(mapEntry.getKey())
-              || globalLatestFlushedTimeForEachDevice.get(mapEntry.getKey())
-                  < mapEntry.getValue()) {
-            globalLatestFlushedTimeForEachDevice.put(mapEntry.getKey(), mapEntry.getValue());
-          }
-        }
+        globalLatestFlushedTimeForEachDevice.putAll(resource.getEndTimeMap());
```
SilverNarcissus: How could you ensure resource end time is larger than global end time?

SilverNarcissus: I know that~

jt2594838: I think the files are sorted by the partition number and version, so the later one cannot have timestamps earlier than the previous one.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -598,23 +591,24 @@ private void insertBatchToTsFileProcessor(BatchInsertPlan batchInsertPlan,
       return;
     }
 
-    tsFileProcessor.insertBatch(batchInsertPlan, start, end, results);
+    try {
+      tsFileProcessor.insertBatch(batchInsertPlan, start, end, results);
+    } catch (WriteProcessException e) {
+      logger.error("insert to TsFileProcessor error ", e);
+      return;
```
jt2594838: How could the client know anything about this?

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileProcessor.java
```diff
@@ -167,10 +165,9 @@ public boolean insert(InsertPlan insertPlan) throws QueryProcessException {
     if (IoTDBDescriptor.getInstance().getConfig().isEnableWal()) {
       try {
         getLogNode().write(insertPlan);
-      } catch (IOException e) {
-        logger.error("{}: {} write WAL failed", storageGroupName,
-            tsFileResource.getFile().getName(), e);
-        return false;
+      } catch (Exception e) {
+        throw new WriteProcessException(String.format("%s: %s write WAL failed, because %s",
+            storageGroupName, tsFileResource.getFile().getAbsolutePath(), e.getMessage()));
```
jt2594838: Preserve the stack trace.

PR #932, opened 2020-03-23
fix sqlArgumentsList concurrent modification error

qiaojialin:
In the Meteorological scenario, the sqlArgumentsList throw ConcurrentModificationException, use a synchronized to avoid.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
+++ b/server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
@@ -1312,7 +1312,11 @@ public class TSServiceImpl implements TSIService.Iface, ServerContext {
   }
 
   public static List<SqlArgument> getSqlArgumentList() {
-    return sqlArgumentList;
+    List<SqlArgument> readCopy;
+    synchronized (sqlArgumentList) {
+      readCopy = new ArrayList<>(sqlArgumentList);
+    }
+    return readCopy;
   }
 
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -133,7 +133,7 @@
   private static final int DELETE_SIZE = 50;
   private static final String ERROR_PARSING_SQL =
       "meet error while parsing SQL to physical plan: {}";
-  public static Vector<SqlArgument> sqlArgumentsList = new Vector<>();
+  public static final Vector<SqlArgument> sqlArgumentsList = new Vector<>();
```
vesense: why not use `CopyOnWriteArrayList`?

jixuan1989: After reading codes, I find `sqlArgumentsList` is for MetricService, which is for logging the query history (and its execution plan..). So, the field is write-intense and read-rare. 
Is CopyOnWriteArrayList has better performance than `synchronized ` in this way?

Besides, @qiaojialin, don't you need to add lock for reading the list? (line 46 in QueryServlet.java)

qiaojialin: Hi, the sqlArgumentList is a write-intensive with nearly no read. I plan to use the Collections.synchronizedList to avoid synchronization everywhere.

PR #934, opened 2020-03-23
Delete dataTypeMapping etc fields in QueryPlan

Alima777:
DataTypeMapping, deviceTomeasurementMap etc fields take lots of memory space while they can be replaced by other fields actually. Therefore, I delete these two fields here. 
Tip: If some fields are no need to add, please use existing fields to replace them.


server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -548,13 +548,13 @@ public TSDataType getSeriesType(String path) throws MetadataException {
...
-   * Get all devices under given prefixPath.
+   * Get all device nodes under given prefixPath.
    *
    * @param prefixPath a prefix of a full path. if the wildcard is not at the tail, then each
    * wildcard can only match one level, otherwise it can match to the tail.
-   * @return A HashSet instance which stores devices names with given prefixPath.
+   * @return A list which stores device nodes with given prefixPath.
    */
-  public Set<String> getDevices(String prefixPath) throws MetadataException {
+  public List<MNode> getDevices(String prefixPath) throws MetadataException {
```
samperson1997: Actually I think it's better to use `String` here so that you won't use `getFullPath()` when call this method ... What do you think?

server/src/main/java/org/apache/iotdb/db/query/dataset/AlignByDeviceDataSet.java
```diff
@@ -21,12 +21,16 @@
...
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import org.apache.iotdb.db.exception.StorageEngineException;
+import org.apache.iotdb.db.exception.metadata.MetadataException;
 import org.apache.iotdb.db.exception.query.QueryProcessException;
+import org.apache.iotdb.db.metadata.MManager;
```
samperson1997: Remove some useless imports

server/src/main/java/org/apache/iotdb/db/query/dataset/AlignByDeviceDataSet.java
```diff
@@ -116,25 +119,29 @@ protected boolean hasNextWithoutConstraint() throws IOException {
...
+          }
         } else {
-          Path path = new Path(currentDevice, column);
-          tsDataTypes.add(tsDataTypeMap.get(path));
-          executePaths.add(path);
+          if (measurementOfGivenDevice.contains(column)) {
+            executeColumns.add(column);
+            executePaths.add(new Path(currentDevice.getFullPath(), column));
+            tsDataTypes.add(measurementDataTpeMap.get(column));
+          }
```
samperson1997: try to extract the same process to avoid duplicated codes:
```
String measurement = column;
        if (dataSetType == DataSetType.GROUPBY || dataSetType == DataSetType.AGGREGATE) {
          measurement = column.substring(column.indexOf('(') + 1, column.indexOf(')'));
          if (measurementOfGivenDevice.contains(measurement)) {
            executeAggregations.add(column.substring(0, column.indexOf('(')));
          }
        }
        if (measurementOfGivenDevice.contains(measurement)) {
          executeColumns.add(measurement);
          executePaths.add(new Path(currentDevice.getFullPath(), measurement));
          tsDataTypes.add(measurementDataTpeMap.get(measurement));
        }
```
(please first ensure codes above work : ) )

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
Alima777: If using `String` here, we have to find the node again by path string in AlignByDeviceDataSet. So storaging MNode directly here can reduce that process.

server/src/main/java/org/apache/iotdb/db/query/dataset/AlignByDeviceDataSet.java
Alima777: Fixed.

server/src/main/java/org/apache/iotdb/db/query/dataset/AlignByDeviceDataSet.java
Alima777: Thank you! It's very helpful for me.

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
samperson1997: Sure. I didn't figure out a better way either... Maybe we could keep this then : )

Alima777: Actually It doesn't cost that much time finding device node. Like what you said, to keep the consistence with other methods, I deciced to change it to return List<String>.

PR #935, opened 2020-03-23
Integrate data file version recording with time partitioning

jt2594838:
Previously the data file version recording is not integrated with time partitioning, so versions of different partitions may be mixed. 

This PR will make file versions of different partitions be managed separately using a map.

Some other minor fixes and extractions are also included to improve the code quality.


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -1638,12 +1646,98 @@ public void loadNewTsFile(TsFileResource newTsFileResource)
...
+        // the two files overlap in the device
+        return 0;
+      }
+    }
+    if (hasPre && hasSubsequence) {
+      // some devices are newer but some devices are older, the two files overlap in general
+      return 0;
+    }
+    if (!hasPre && hasSubsequence) {
+      // all devices in B are newer than those in A
```
fanhualta: All devices in A are newer than those in B.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -1638,12 +1646,98 @@ public void loadNewTsFile(TsFileResource newTsFileResource)
...
+      return -1;
+    }
+    // all devices in B are older than those in A
```
fanhualta: All devices in A are older than those in B.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -1638,12 +1646,98 @@ public void loadNewTsFile(TsFileResource newTsFileResource)
...
+        // skip files that are not in the partition as the new file and the last empty file
+        continue;
+      }
+
+      int fileComparison = compareTsFileDevices(newTsFileResource, localFile);
+      switch (fileComparison) {
+        case 0:
+          // some devices are newer but some devices are older, the two files overlap in general
+          return POS_OVERLAP;
+        case -1:
```
fanhualta: case 1:

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -1638,12 +1646,98 @@ public void loadNewTsFile(TsFileResource newTsFileResource)
...
+
+  /**
+   * Compare each device in the two files to find the time relation of them.
+   * @param fileA
+   * @param fileB
+   * @return -1 if fileA is totally older than fileB (A < B)
+   *          0 if fileA is partially older than fileB and partially newer than fileB (A X B)
+   *          1 if fileA is totally newer than fileB (B < A)
+   */
+  private int compareTsFileDevices(TsFileResource fileA, TsFileResource fileB) {
```
fanhualta: I think this method needs to be  checked carefully.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -1638,12 +1646,98 @@ public void loadNewTsFile(TsFileResource newTsFileResource)
...
+
+      int fileComparison = compareTsFileDevices(newTsFileResource, localFile);
+      switch (fileComparison) {
+        case 0:
+          // some devices are newer but some devices are older, the two files overlap in general
+          return POS_OVERLAP;
+        case -1:
+          // all devices in the local file are newer than the new file, the new file can be
+          // inserted before the new file
+          return i - 1;
```
fanhualta: `return i;`
It needs to add in the cur pos and move all newer files backward.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -1638,12 +1646,98 @@ public void loadNewTsFile(TsFileResource newTsFileResource)
...
+        default:
+          // all devices in the local file are older than the new file, proceed to the next file
+          insertPos = i;
```
fanhualta: `return i + 1;`
It needs to be inserted after the older files.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -1559,61 +1601,24 @@ public void loadNewTsFileForSync(TsFileResource newTsFileResource)
...
-      if (isOverlap) {
+      if (insertPos == POS_OVERLAP) {
         loadTsFileByType(LoadTsFileType.LOAD_UNSEQUENCE, tsfileToBeInserted, newTsFileResource);
       } else {
 
         // check whether the file name needs to be renamed.
-        if (subsequentIndex != sequenceFileTreeSet.size() || preIndex != -1) {
-          String newFileName = getFileNameForLoadingFile(tsfileToBeInserted.getName(), preIndex,
-              subsequentIndex, getTimePartitionFromTsFileResource(newTsFileResource));
+        if (insertPos != sequenceFileTreeSet.size() - 1 || insertPos != -1) {
```
fanhualta: You can use `if(! sequenceFileTreeSet.isEmpty())`.

The origin `if (subsequentIndex != sequenceFileTreeSet.size() || preIndex != -1)` is equals to above.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -1702,24 +1796,26 @@ private void removeFullyOverlapFiles(TsFileResource resource, Iterator<TsFileRes
...
-    if (preIndex == -1) {
+    if (insertIndex == -1) {
       preTime = 0L;
     } else {
-      String preName = sequenceList.get(preIndex).getFile().getName();
+      String preName = sequenceList.get(insertIndex).getFile().getName();
       preTime = Long.parseLong(preName.split(IoTDBConstant.TSFILE_NAME_SEPARATOR)[0]);
     }
-    if (subsequentIndex == sequenceFileTreeSet.size()) {
+    if (insertIndex == sequenceFileTreeSet.size() - 1) {
```
fanhualta: `insertIndex == sequenceFileTreeSet.size()`

Logic:
If it needs to be inserted to be the last in the list.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -1559,61 +1601,24 @@ public void loadNewTsFileForSync(TsFileResource newTsFileResource)
...
         loadTsFileByType(LoadTsFileType.LOAD_UNSEQUENCE, tsfileToBeInserted, newTsFileResource);
       } else {
 
         // check whether the file name needs to be renamed.
-        if (subsequentIndex != sequenceFileTreeSet.size() || preIndex != -1) {
-          String newFileName = getFileNameForLoadingFile(tsfileToBeInserted.getName(), preIndex,
-              subsequentIndex, getTimePartitionFromTsFileResource(newTsFileResource));
+        if (insertPos != sequenceFileTreeSet.size() - 1 || insertPos != -1) {
+          String newFileName = getFileNameForLoadingFile(tsfileToBeInserted.getName(), insertPos,
+              getTimePartitionFromTsFileResource(newTsFileResource));
```
fanhualta: Just a note: currently, `insertPos ` is equals to `subsequentIndex `.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
jt2594838: I have changed the assignments of hasPre and hasSubsequence, now hasPre means "B has previous data to A", so this should remain.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
jt2594838: See above.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
jt2594838: I believe it should be -1, which means data in the new file has smaller timestamps than the local file, so we can insert the new file before the local file.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
jt2594838: Yes, I mistakenly reversed the condition and now it is fixed.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
jt2594838: As I have explained in the javadoc that the file will be inserted between [i, i +1], I think it does not really matter.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
jt2594838: See above.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
jt2594838: Changed. But I doubt the quality of them, as the file being inserted at the head or the tail does not necessarily indicate the set is empty.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
jt2594838: You have misunderstood the meaning of insertIndex, please read the Javadoc. For example, if the insertIndex is -1, it means the new file will be inserted between [-1, 0], not that the new file will be inserted at -1.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -1655,7 +1655,7 @@ public void loadNewTsFile(TsFileResource newTsFileResource)
    */
   private int findInsertionPosition(TsFileResource newTsFileResource) {
     File tsfileToBeInserted = newTsFileResource.getFile();
-    long newFilePartitionId = Long.parseLong(tsfileToBeInserted.getParent());
+    long newFilePartitionId = Long.parseLong(tsfileToBeInserted.getParentFile().getName());
```
fanhualta: The parent file of the file may not be its time range. It's better to get time range from the time.
Such as:
`load /data/1584445970309-101-1.tsfile`

jt2594838: Fixed. But there is still one thing remaining, what if the file belongs to more than one partition? Currently, I will report an error and a delicate solution will be left as a future job.

fanhualta: Yes, I agree with you.

PR #943, opened 2020-03-25
Update AlignByDeviceQuery Doc

Alima777:
After deleting the dateTypeMapping structure in QueryPlan, the doc need to be updated too.


Alima777:
> Hi, since the English design documents are merged, you'd better modify English version too. : )

Hi, thank you for reminding. :）

PR #966, opened 2020-03-31
Fix visible metadata, version, work processor bugs when recovering

qiaojialin:
Fix many bugs when recovering:

(1) when recovering the last unclosed SEQ file, the TsFileProcessor is added to workUnsequenceTsFileProcessors.
(2) when recovering the last unclosed UNSEQ file, the TsFileProcessor isn's added to any workTsFileProcessors.
(3) when recovering, the metadata get from TsFile is set to chunkMetadataListMap, however, visibale metadata is recovered from chunkGroupInfoList. Therefore, the persisted Chunk is not visible.
(4) when recovering, the TsFileRecoverPerformer needs to receive a partitionId. However, before recovering, the startTime & endTime maps are empty. So the partitionId is -1, which leads to the wrong version of Deletion and memtable version in recovery. I get the partitionId from resource path. No more -1 partitions in the future.


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
```diff
@@ -471,4 +473,12 @@ public void setProcessor(TsFileProcessor processor) {
   public TimeseriesMetadata getTimeSeriesMetadata() {
     return timeSeriesMetadata;
   }
+
+  public long getTimePartition() {
+    if (startTimeMap != null && !startTimeMap.isEmpty()) {
+      return StorageEngine.getTimePartition(startTimeMap.values().iterator().next());
+    }
+    String[] splits = FilePathUtils.splitTsFilePath(this);
+    return Long.parseLong(splits[splits.length - 2]);
```
SilverNarcissus: If we sync a tsfile from other iotdb, will this tsfile has time partition directory? Maybe we should use startTimeMap to get time partition

tsfile/src/main/java/org/apache/iotdb/tsfile/write/writer/TsFileIOWriter.java
```diff
@@ -65,14 +66,19 @@
...
-  protected Map<Path, List<ChunkMetadata>> chunkMetadataListMap = new TreeMap<>();
+
+
+  // current flushed Chunk
   private ChunkMetadata currentChunkMetadata;
+  // current flushed ChunkGroup
+  protected List<ChunkMetadata> chunkMetadataList = new ArrayList<>();
+  // all flushed ChunkGroup,  device -> List<ChunkMetadata>
+  protected List<Pair<String, List<ChunkMetadata>>> chunkGroupMetadataList = new ArrayList<>();
+
```
HTHou: Since Pair<String, List<ChunkMetadata> is similar to previous ChunkGroupMetadata, I'm thinking if it's necessary to add ChunkGroupMetadata back instead of using the pair.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
qiaojialin: Thanks, I use the startTimeMap first, then the path

tsfile/src/main/java/org/apache/iotdb/tsfile/write/writer/TsFileIOWriter.java
qiaojialin: Good idea, I have added it

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
jt2594838: It is not guaranteed that the other IoTDB will use the same partition interval, so it is better to check the whole startTimeMap and endTimeMap to see if any device has crossed a partition and report an exception if so.

qiaojialin: restart and load could use different getPartition method, one check partition and the other not. I will add a method with partition check

PR #986, opened 2020-04-04
[IOTDB-578] Cluster module should use the same thrift version with io…

vesense:
…tdb-parent

### DIFF
--- a/cluster/pom.xml
+++ b/cluster/pom.xml
@@ -35,11 +35,6 @@
     </properties>
     <dependencies>
         <dependency>
-            <groupId>org.apache.thrift</groupId>
-            <artifactId>libthrift</artifactId>
-            <version>0.12.0</version>
-        </dependency>
-        <dependency>
             <groupId>org.apache.iotdb</groupId>
             <artifactId>service-rpc</artifactId>
             <version>0.10.0-SNAPSHOT</version>


cluster/pom.xml
```diff
@@ -34,11 +34,6 @@
         <cluster.ut.skip>${cluster.test.skip}</cluster.ut.skip>
     </properties>
     <dependencies>
-        <dependency>
-            <groupId>org.apache.thrift</groupId>
-            <artifactId>libthrift</artifactId>
```
jixuan1989, line 39: I think we can only remove the version here.
In the parent pom, the libthrift is claimed in dependencyManagement, rather than dependencies.

vesense: @jixuan1989 the `cluster` module is dependent on `service-rpc`, and `service-rpc` module has the `libthrift` dependency which version extends from parent. So we don't need to add the `libthrift` dependency to cluster module separately.

```
service-rpc/pom.xml

 <dependency>
            <groupId>org.apache.thrift</groupId>
            <artifactId>libthrift</artifactId>
</dependency>
```

jt2594838:
thrift 0.13.0 used by the stand-alone IoTDB will cause bugs in Mac OS and that is why we use a separate version in the distributed version.
PLEASE, do not merge any outside code without informing the committers who are working on this branch.

PR #1008, opened 2020-04-08
enable compression in session pool

qiaojialin:


jixuan1989:
Hi, please add an ignore annotation.

PR #1009, opened 2020-04-08
Revert "[IOTDB-578] Cluster module should use the same thrift version with io…"

jt2594838:
Reverts apache/incubator-iotdb#986

### DIFF
--- a/cluster/pom.xml
+++ b/cluster/pom.xml
@@ -35,6 +35,11 @@
     </properties>
     <dependencies>
         <dependency>
+            <groupId>org.apache.thrift</groupId>
+            <artifactId>libthrift</artifactId>
+            <version>0.12.0</version>
+        </dependency>
+        <dependency>
             <groupId>org.apache.iotdb</groupId>
             <artifactId>service-rpc</artifactId>
             <version>0.10.0-SNAPSHOT</version>


jixuan1989:
> thrift 0.13.0 used by the stand-alone IoTDB will cause bugs in Mac OS and that is why we use a separate version in the distributed version.
PLEASE, do not merge any outside code without informing the committers who are working on this branch.

If so, 
1. should we downgrade the thrift version in the server module?
2. add a comment on the top of the dependency in the cluster pom.xml to claim why you have to do that.

vesense:
> > thrift 0.13.0 used by the stand-alone IoTDB will cause bugs in Mac OS and that is why we use a separate version in the distributed version.
> > PLEASE, do not merge any outside code without informing the committers who are working on this branch.
> 
> If so,
> 
> 1. should we downgrade the thrift version in the server module?
> 2. add a comment on the top of the dependency in the cluster pom.xml to claim why you have to do that.

Agree with @jixuan1989  We'd better keep the same version in project as far as possible.

liutaohua:
I think you should follow the principle of majority voting and not merge code without anyone else knowing.
And we need to add more comments to specific places so that contributors/committers can note why .

PR #1011, opened 2020-04-08
[IOTDB-583] fix concurrent modification exception

qiaojialin:
The upgrade thread traverses the list while new resources is added.


qiaojialin:
> It can reduce the concurrecny exception, but I am not sure whether new ArrayList(Collection) is atomic..

Maybe not atomic, but the most important is we need to release 0.9.2...

PR #1016, opened 2020-04-09
Remove disable align syntax within LAST query

wshao08:
As discussed, the result set of **Last** query will always be displayed in a fixed three column table format and no longer support "**disable align**" syntax. 
Now the user will get an exception if adding a "**disable align**" at the end of a **Last** SQL.


qiaojialin:
image

fix the test

PR #1027, opened 2020-04-10
[IOTDB-348] add SetTTLPlan, DataAuthPlan and DeletePlan

neuyilan:


cluster/src/test/java/org/apache/iotdb/cluster/common/TestRemoteFileSnapshot.java
```diff
@@ -56,7 +57,7 @@ public void getRemoteSnapshot() {
     int startTime = 12345;
     for (RemoteTsFileResource tsFileResource : remoteDataFiles) {
       // fake a file for each resource
-      String[] splits = tsFileResource.getFile().getPath().split("\\\\");
+      String[] splits = tsFileResource.getFile().getPath().split(File.separator);
```
jt2594838, line 60: It is good to take different platforms into account. But `split()` receives a regex so I wonder whether the non-escaped separator string works.
Have you tested it on both Linux and Windows?

jt2594838:
Good job. But there is more to discuss TTL and DataAuth.
TTLs and users' info must also be put into the meta snapshot so that when the logs are removed, the stale nodes can still catch up. This can be done within the PR or you can start an issue and work on it in another PR, both fine to me, how do you think?

neuyilan:
> Good job. But there is more to discuss TTL and DataAuth.
> TTLs and users' info must also be put into the meta snapshot so that when the logs are removed, the stale nodes can still catch up. This can be done within the PR or you can start an issue and work on it in another PR, both fine to me, how do you think?

I see that the relevant work of snapshot is not perfect at present, so I would like to reopen a new issue to complete this part of work.

cluster/src/test/java/org/apache/iotdb/cluster/common/TestRemoteFileSnapshot.java
neuyilan: yes, I have tested it on both Linux and Windows, and it passed.

PR #1028, opened 2020-04-11
[IoTDB-575]add default jmx user and password; fix issues that jmx can not be accessed remotely

jixuan1989:
Though PR#979 has requires users to use JMX with authorization, it breaks our rule "make IoTDB available out of the box".
Users have to add files into their JDK HOME and then start IoTDB.

This PR fixes that and gives two default user/passwords for JMX.

Besides, I notice there are some unused codes in our JMXService module, so I simplify them.


jixuan1989:
close this for more work

PR #1030, opened 2020-04-11
upgrade site and add docs/zh

SailVR:


jixuan1989:
Hi, 
I saw Win Travis failed because out-of-memory.
Please apply the following patch 
```
git apply travis.patch.txt
```
[travis.patch.txt](https://github.com/apache/incubator-iotdb/files/4466167/travis.patch.txt)

It will split the website builder into a single container.

jixuan1989:
Hi,

please add the following content into pom.xml

After Line 569:
```
                            <!-- only for Travis CI with WinOS-->
                            <exclude>hadoopbin</exclude>
                             <exclude>windowssystem32</exclude>

```

PR #1031, opened 2020-04-12
[IOTDB-589] Upgrade moquette version to 0.13

vesense:
* upgraded moquette version to 0.13
* moved the jcenter repo to server module


vesense:
@jixuan1989 Can you take a look?

image

distribution/pom.xml
```diff
@@ -83,6 +83,19 @@
             </plugin>
         </plugins>
     </build>
+    <repositories>
```
jixuan1989, line 86: does this module also require bintray?

vesense: yes. distribution has the iotdb-server dependency. it will throw error without the repo.
https://travis-ci.org/github/apache/incubator-iotdb/builds/673966792
image

PR #1032, opened 2020-04-12
[IOTDB-573]claim the error output can be igonred in UT/ITs

jixuan1989:
PR #1015 introduced WrappedRunnable but print too many error messages in UT.

I make the error messages easy to be distinguished and reduce the number of test threads.


vesense:
LGTM

PR #1033, opened 2020-04-12
disable mqtt server in default config

liutaohua:
I don't think MQTT server should be started without the user's knowledge.

please have a look , @jixuan1989  @vesense


vesense:
_Why the previous default value is true?_
As an IOT database, mqtt should be supported more natively (as a built-in service). And the service started by default is for the convenience of users.

_Do we need to let users know about this change?_
Yes, of course. I think we should specify the changes in the 0.10 release note.

_Can we change it to the default false?_
The reason why the default value is true is explained above. I'm OK for both. And I'm willing to listen to others.

jixuan1989:
+1 for  @vesense 
I also think it should be enable by default.
Can you give some reasons to explain why you want to disable it by default?

liutaohua:
I think MQTT server just an extension connector. 
Opening up an extra port for infrequently used features can be very dangerous.

jixuan1989:
> Opening up an extra port for infrequently used features can be very dangerous.

you are right. But I think it is not the reason that we disable it. Instead, we need to:

1. guarantee there is no security problem in our implementation;
2. Only expose the port to authorized IPs by firewall.

> I think MQTT server just an extension connector.

How to say... maybe many companies use their customized TCP protocols, but MQTT is a symbolic meaning for IoT area.

liutaohua:
What you said is very contradictory. Does MQTT communicate with clients or with background services?
If client, how to set authorized IPs?
If BG-server ,why not use thrift?

jixuan1989:
Hi,

Yes, when people use MQTT,  the scenario is let devices connect to the MQTT server directly.

Normally, the "device" is a car, a smart electric meter, etc..

If the "device" has a const IP, then you can use firewall to allow which IPs can access the port (Almost all Cloud platforms, aliyun, aws, etc., support that. Normally they are called “安全组”).

If the "device" has no a const IP, then maybe two solutions can be chosen:

1. The device reports its IP to someone and then be added into the firewall;
2. Use a complex USERNAME/PASSWORD in MQTT protocol.

jixuan1989:
I think how to guarantee the MQTT port safety is not the focus of this discussion.

I think  whether we enable the port by default just depends on what role we want IoTDB plays:

- Do we want to wave the flag that "IoTDB is for IoT area. Therefore, it put the international Standards in IoT area in priority."

liutaohua:
Well, hopefully it will prompt the user at the time of publication that an mqtt-server is turned on and that the user remembers to change the default user password

liutaohua:
According to the results of the discussion[1], I still suggest the default close the port, so I reopen the pr

[1]https://lists.apache.org/thread.html/rc86378cf67aca3635536ce9baa437b82128f5e07e0f0e345bec8d0a7%40%3Cdev.iotdb.apache.org%3E

jixuan1989:
Hi @liutaohua , can you grant me the write permission to your repo? 
I'd like to add some logs to see why there are three Travic CI failed.

liutaohua:
> Hi @liutaohua , can you grant me the write permission to your repo?
> I'd like to add some logs to see why there are three Travic CI failed.

 sure, thank you very much for your work

PR #1048, opened 2020-04-14
Switch to use new TsFile interfaces in Last query executor

wshao08:
This is a PR to utilize the new TsFile interfaces in Last query executor implementation, and remove the deprecated interfaces.
 
Refer to https://issues.apache.org/jira/browse/IOTDB-587 for the new TsFile improvements.


server/src/main/java/org/apache/iotdb/db/query/fill/PreviousFill.java
```diff
@@ -141,15 +141,13 @@ private TimeValuePair retrieveValidLastPointFromSeqFiles() throws IOException {
           FileLoaderUtils.loadTimeSeriesMetadata(
               resource, seriesPath, context, timeFilter, allSensors);
       if (timeseriesMetadata != null) {
-        if (timeseriesMetadata.getStatistics().canUseStatistics()
-            && endtimeContainedByTimeFilter(timeseriesMetadata.getStatistics())) {
+        if (endtimeContainedByTimeFilter(timeseriesMetadata.getStatistics())) {
```
JackieTien97, line 144: Why you delete the judgement `timeseriesMetadata.getStatistics().canUseStatistics()`.
What if the file has been modified and timeseriesMetadata can not been used, for example, it has been deleted from the endTime of it.

wshao08: Fixed

PR #1051, opened 2020-04-15
[IOTDB-526] Support some metadata queries

Ring-k:
The method getNodesList is implemented in the cluster version. The bat files are executed to test the new function.

SET STORAGE GROUP TO root.ed0;
SET STORAGE GROUP TO root.rf1;
SET STORAGE GROUP TO root.fh2;
SET STORAGE GROUP TO root.zh3;
...
create timeseries root.ed0.wf01.wt01.status with datatype=BOOLEAN,encoding=PLAIN;
create timeseries root.rf1.wf01.wt01.status with datatype=BOOLEAN,encoding=PLAIN;
create timeseries root.fh2.wf01.wt01.status with datatype=BOOLEAN,encoding=PLAIN;
create timeseries root.zh3.wf01.wt01.status with datatype=BOOLEAN,encoding=PLAIN;
...

The results are: 
+-----+
|count|
+-----+
|10000|
+-----+
Total line number = 1
It costs 0.043s


cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
```diff
@@ -81,10 +97,48 @@ public QueryDataSet processQuery(PhysicalPlan queryPlan, QueryContext context)
...
+      pool.submit(() -> {
+        DataClient client = null;
+        try {
+          client = metaGroupMember.getDataClient(node);
+        } catch (IOException e) {
+          logger.error("Failed to connect to node: {}", node, e);
+        }
+        AtomicReference<List<String>> response = new AtomicReference<>(null);
+        handler.setResponse(response);
+        handler.setContact(node);
```
jt2594838, line 122: The same handler is used in different threads and its response and contact are set concurrently, so I am afraid that only the result from one node is returned because there is only one response in the handler.

And the answer is still correct because the default replica number and node number are both 3, so timeseries are fully replicated in each node and even if you only query one node, it will have all timeseries. Thus I suggest you test in a different configuration, like 4-node-3-replica or 3-node-2-replica.

cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
```diff
@@ -81,10 +97,48 @@ public QueryDataSet processQuery(PhysicalPlan queryPlan, QueryContext context)
...
+            }
+          } catch (TException e) {
+            logger.error("Error occurs when getting node lists in node {}.", node, e);
+          } catch (InterruptedException e) {
+            logger.error("Interrupted when getting node lists in node {}.", node, e);
+          }
+        }
+        List<String> paths = response.get();
+        if (paths != null) {
+          nodeListSet.addAll(response.get());
```
jt2594838, line 137: As `nodeListSet` is modified in different threads, maybe you should use a concurrent class.

By the way, the naming `nodeListSet` is confusing, probably better to just use `nodeSet`.

jt2594838:
The current implementation is good, but not the best. The problem is that when replications are used, nodes will have overlapped timeseries. For example, node1,2,3 all have timeseries1,2,3, so 3 times as much timeseries will be transferred to respond to the query, let alone further deduplication.

My suggestion is: you should query each group instead of each node, and the queried node will know from the parameter `header` which group you are querying, so it will only send the timeseries that are actually managed by the node (Remember we use the storage group to partition timeseries schema).

It is up to you to make an improvement within this pr or start another pr. But if you choose the latter one, please supplement your idea in the related Jira issue.

cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
```diff
@@ -107,9 +108,6 @@ public QueryDataSet processQuery(PhysicalPlan queryPlan, QueryContext context)
     ExecutorService pool = new ScheduledThreadPoolExecutor(THREAD_POOL_SIZE);
 
     for (Node node : metaGroupMember.getAllNodes()) {
-      if (node.equals(metaGroupMember.getThisNode())) {
-        continue;
-      }
```
jt2594838: Why not skip the local node?

cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
```diff
@@ -138,13 +136,53 @@ public QueryDataSet processQuery(PhysicalPlan queryPlan, QueryContext context)
...
+      pool.submit(() -> {
+        DataClient client = null;
+        try {
+          client = metaGroupMember.getDataClient(node);
+        } catch (IOException e) {
+          logger.error("Failed to connect to node: {}", node, e);
+        }
+        AtomicReference<List<String>> response = new AtomicReference<>(null);
+        handler.setResponse(response);
+        handler.setContact(node);
```
jt2594838, line 164: Please notice the concurrent problem as the above one.

cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
Ring-k: I thought maybe I can use `client == null` to check if it is the local node. Since it makes no differentce, I use the `if (node.equals(metaGroupMember.getThisNode()))` as before.

cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
Ring-k: Thanks for your reminding. `ConcurrentSkipListSet` is used instead of `Set` now.

cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
Ring-k: Yes, there's a problem. Thanks for your helpful advice. The initialization of  `response` now is inside each thread.

Ring-k: Yes, there's a problem. Thanks for your reminding. The initialization of handler should be inside each thread.

cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
```diff
@@ -121,38 +124,41 @@ public QueryDataSet processQuery(PhysicalPlan queryPlan, QueryContext context)
         synchronized (response) {
           try {
             if (client != null) {
-              client.getNodeList(null, schemaPattern, level, handler);
+              client.getNodeList(node, schemaPattern, level, handler);
               response.wait(connectionTimeoutInMS);
             }
           } catch (TException e) {
             logger.error("Error occurs when getting node lists in node {}.", node, e);
```
jt2594838: It will be more robust if, for each group, you will try querying each node in the group until one result is returned. So that even if one node is down, you can still get the result from its replicas.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/DataGroupMember.java
```diff
@@ -1369,6 +1370,24 @@ public void getChildNodePathInNextLevel(Node header, String path,
...
+    if (!syncLeader()) {
+      resultHandler.onError(new LeaderUnknownException(getAllNodes()));
+      return;
+    }
+    try {
+      List<List<String>> res = new ArrayList<>();
+      for(String[] element : MManager.getInstance().getAllMeasurementSchema(path)){
+        res.add(Arrays.asList(element));
+      }
+      resultHandler.onComplete(res);
```
jt2594838, line 1385: The current implementation cannot avoid the problem of sending duplicated metadata.
The solution may be modifying the interfaces in MManager to add a filter that will skip the unsatisfying timeseries that do not belong to the slots managed by the member.
It is a little tricky and it is okay if you do it in another PR.

cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
Ring-k: Thanks for your suggestion.

cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
```diff
@@ -81,22 +99,172 @@ public QueryDataSet processQuery(PhysicalPlan queryPlan, QueryContext context)
...
-    throw new UnsupportedOperationException("Not implemented");
-    //return metaGroupMember.getNodeList(schemaPattern, level);
+  protected List<String> getNodesList(String schemaPattern, int level)
+      throws IOException, TException, InterruptedException, MetadataException {
+    ConcurrentSkipListSet<String> nodeSet = new ConcurrentSkipListSet<>(
+        MManager.getInstance().getNodesList(schemaPattern, level));
+
+    ExecutorService pool = new ScheduledThreadPoolExecutor(THREAD_POOL_SIZE);
+
+    for (PartitionGroup group : metaGroupMember.getPartitionTable().getLocalGroups()) {
```
jt2594838, line 109: Local groups are the groups that the local node is in and they do not necessarily cover all groups. For example, if replica number is 2, and there are 4 nodes 1,2,3,4 the local groups of 1 will only be [4,1] and [1,2].

cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
```diff
@@ -81,22 +99,172 @@ public QueryDataSet processQuery(PhysicalPlan queryPlan, QueryContext context)
...
+            }
+          } catch (IOException e) {
+            logger.error("Failed to connect to node: {}", node, e);
+          } catch (TException e) {
+            logger.error("Error occurs when getting node lists in node {}.", node, e);
+          } catch (InterruptedException e) {
+            logger.error("Interrupted when getting node lists in node {}.", node, e);
+            Thread.currentThread().interrupt();
+          }
+        }
```
jt2594838, line 142: It would be better to leave a log here if you fail to get result from all nodes in the group.

cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
Ring-k: That's right. Thanks for your reminding.

cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
Ring-k: Yes, it would be more clear.

PR #1057, opened 2020-04-15
[IOTDB-528]grafana querySeries group by 

SailVR:


jixuan1989:
Hi @SailVR ,  add descriptions to tell reviewers what you are doing is important..

SailVR:
When using itodb-grafana integration through SimpleJson it seems like it does not do any aggregation at all.

In situations where one selects a big time range it should do some subsampling with group by to limit the number of rows being sent.

PR #1071, opened 2020-04-20
[IOTDB-603]add Feedback page for users

jixuan1989:
Hi,

As we are using dev@ mailing list to serve for developers and users together, it is better to add a page to point out it.

Otherwise, users may think that the dev@ is just for developers.


liutaohua:
how about  add a `公众号`

jixuan1989:
@liutaohua the 公众号 is already added.

PR #1074, opened 2020-04-20
Fix OOM

fanhualta:
This pr fixed the OOM issue by two approaches:
1.  Use a new tool for estimating memtable size: **RamUsageEstimator**.
2. Limit the total size of the WAL to one-tenth of the write memory.


server/src/main/java/org/apache/iotdb/db/utils/MemUtils.java
```diff
@@ -88,7 +86,7 @@ public static long getRecordSize(BatchInsertPlan batchInsertPlan, int start, int
         case TEXT:
           memSize += (end - start) * 8L;
           for (int j = start; j < end; j++) {
-            memSize += ((Binary[]) batchInsertPlan.getColumns()[i])[j].getLength();
+            memSize += RamUsageEstimator.sizeOf(((Binary[]) batchInsertPlan.getColumns()[i])[j]);
```
qiaojialin, line 89: Hi, this may slow down the write performance of TEXT data heavily. It's better to do sampling or calculate the constant header of Binary and byte[].

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -44,6 +44,7 @@
 import org.apache.iotdb.db.conf.IoTDBDescriptor;
 import org.apache.iotdb.db.conf.directories.DirectoryManager;
 import org.apache.iotdb.db.engine.StorageEngine;
+import org.apache.iotdb.db.engine.cache.RamUsageEstimator;
```
jt2594838, line 47: This import is added but I cannot find usages of it in this PR, so maybe it is unused?

server/src/main/java/org/apache/iotdb/db/utils/MemUtils.java
jt2594838: Agreed.

fanhualta: I have modified the way to estimate the size of Binary. In new way, the efficiency is nearly same with the origin way which approved that it'll not affect the performance of the engine.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
fanhualta: fixed

PR #1081, opened 2020-04-21
Upgrade mlog

HTHou:
To upgrade old mlog, we add empty informations to all old `CREATE_TIMESERIES` records. 

We also add one more property `iotdbVersion` in system.properties for confirming whether we need to do upgrade.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfigCheck.java
+++ b/server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfigCheck.java
@@ -126,15 +126,17 @@ public class IoTDBConfigCheck {
       }
       if (properties.getProperty("iotdb_version") == null) {
         logger.warn("Lower iotdb version detected, upgrading old mlog file... ");
-        MLogWriter.upgradeMLog(IoTDBDescriptor.getInstance().getConfig().getSchemaDir(), 
+        boolean upgradeSuccessful = MLogWriter.upgradeMLog(IoTDBDescriptor.getInstance().getConfig().getSchemaDir(), 
             MetadataConstant.METADATA_LOG);
-        logger.info("Old mlog file is upgraded.");
-        try (FileOutputStream outputStream = new FileOutputStream(file.toString())) {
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfigCheck.java
```diff
@@ -120,6 +124,19 @@ private void checkFile(String filepath) {
...
+        logger.warn("Lower iotdb version detected, upgrading old mlog file... ");
+        MLogWriter.upgradeMLog(IoTDBDescriptor.getInstance().getConfig().getSchemaDir(), 
+            MetadataConstant.METADATA_LOG);
+        logger.info("Old mlog file is upgraded.");
+        try (FileOutputStream outputStream = new FileOutputStream(file.toString())) {
+          properties.setProperty("timestamp_precision", timestampPrecision);
+          properties.setProperty("storage_group_time_range", String.valueOf(partitionInterval));
+          properties.setProperty("tsfile_storage_fs", tsfileFileSystem);
+          properties.setProperty("iotdb_version", iotdbVersion);
+          properties.store(outputStream, "System properties:");
```
qiaojialin, line 137: The new file should be stored after upgrading all system files.

server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfigCheck.java
```diff
@@ -120,6 +124,21 @@ private void checkFile(String filepath) {
             .getProperty("tsfile_storage_fs") + " !");
         System.exit(-1);
       }
+      if (properties.getProperty("iotdb_version") == null) {
+        logger.warn("Lower iotdb version detected, upgrading old mlog file... ");
```
qiaojialin, line 128:
```suggestion
        logger.info("Lower iotdb version {} detected, upgrading old mlog file... ", iotdbVersion);
```

HTHou: Fixed

PR #1086, opened 2020-04-22
update merge selector to filter close

zhanglingzhe0820:


server/src/main/java/org/apache/iotdb/db/engine/merge/selector/MaxFileMergeFileSelector.java
```diff
@@ -201,6 +201,17 @@ private void updateSelectedFiles(long newCost, TsFileResource unseqFile) {
     }
   }
 
+  private boolean checkClosed(TsFileResource unseqFile) {
+    boolean isClosed = unseqFile.isClosed();
```
jt2594838, line 205: You may return here as soon as the unseqFile itself is found unclosed.

PR #1090, opened 2020-04-23
[IOTDB-604] support more robust remote reader

mychaow:


cluster/src/test/java/org/apache/iotdb/cluster/query/reader/RemoteSeriesReaderByTimestampTest.java
```diff
@@ -71,18 +74,93 @@ public void fetchSingleSeriesByTimestamp(Node header, long readerId, long time,
...
+    }
+    assertEquals(TestUtils.getNode(2), sourceInfo.getCurrentNode());
+
+    // all node failed
+    reader.setClientForTest(badClient);
+    try {
+      reader.getValueInTimestamp(90);
+      fail();
+    } catch (IOException e) {
+      //ignore
```
jt2594838, line 163: It would be better to check the information or cause of the exception and see if it is what you have expected.

cluster/src/main/java/org/apache/iotdb/cluster/query/reader/DataSourceInfo.java
```diff
@@ -0,0 +1,163 @@
...
+  private PartitionGroup partitionGroup;
+  private TSDataType dataType;
+  private SingleSeriesQueryRequest request;
+  private RemoteQueryContext context;
+  private MetaGroupMember metaGroupMember;
+  private List<Node> nodes;
+  private DataClient curClient;
+  private boolean isNoData;
+
+  private Map<Node, Boolean> visitedNodes;
```
jt2594838, line 64: Recording the failed nodes is one way, but there is a chance that previously down nodes will go back online when the query proceeds, or maybe there is just a temporary network partition that caused the node time out.

For example, the query reads data of [1, 100] from node1 and node2, when it is reading 10, node1 fails, but later when it is reading 90, node1 recovers but node2 fails and the current implementation will miss a chance to let the query go on. Such a situation is more common when the query is complicated and time-consuming.

So I advise that do not record the visited node, but every time `nextDataClient` is called, traverse `nodes` once from the current position and only return null when the traverse ends but no nodes respond.

mychaow: ok, I will change the code.

PR #1102, opened 2020-04-26
[IOTDB-348] Implement LoadConfigurationPlan

Ring-k:
This pull request implements how the LoadConfigurationPlan is transmitted among nodes in the cluster and then executed.

There are two kinds of LoadConfigurationPlan, one being executed locally, the other globally.

The following statements are tested:

- LOAD CONFIGURATION           // locally
- LOAD CONFIGURATION -g      // globally


cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterDescriptor.java
```diff
@@ -204,4 +206,53 @@ private void loadProps() {
...
+  }
+
+  public void loadHotModifiedProps(Properties properties, boolean toCheckProperties)
+      throws QueryProcessException {
+    if (toCheckProperties && !checkProperties(properties)) {
+      throw new QueryProcessException(
+          "Failed to load configuration from properties because some are missing locally.");
+    }
+    // TODO implement some hot modified properties to load
+    logger.info("Set cluster configuration {}", properties);
```
jt2594838, line 234: Please find somewhere to add the documents about what configurations are supported even if it is only an empty list now.

cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterDescriptor.java
```diff
@@ -204,4 +206,53 @@ private void loadProps() {
...
+    }
+    try (InputStream inputStream = new FileInputStream(new File(url))) {
+      Properties localProperties = System.getProperties();
+      localProperties.load(inputStream);
+      Set<String> localPropertyNames = localProperties.stringPropertyNames();
+      for (String propertyName : properties.stringPropertyNames()) {
+        if (!localPropertyNames.contains(propertyName)) {
+          return false;
+        }
+      }
```
jt2594838, line 250: I am not sure if I got it right, but the parameter `properties` is loaded from `getPropUrl()` and you also load `getPropUrl()` into `localProperties`, so it seems that `properties` must be a subset of `localProperties`.

server/src/main/java/org/apache/iotdb/db/qp/physical/sys/LoadConfigurationPlan.java
```diff
@@ -17,19 +17,107 @@
...
+          Map<String, String> propertiesMap = new HashMap<>();
+          for (Entry<Object, Object> entry : properties.entrySet()) {
+            propertiesMap.put(String.valueOf(entry.getKey()), String.valueOf(entry.getValue()));
+          }
+          stream.writeInt(propertiesMap.size());
+          for(Entry entry : propertiesMap.entrySet()){
+            putString(stream, String.valueOf(entry.getKey()));
+            putString(stream, String.valueOf(entry.getValue()));
+          }
+//          ReadWriteIOUtils.write(propertiesMap, stream);
```
jt2594838, line 97: Please remove this.

server/src/main/java/org/apache/iotdb/db/qp/physical/sys/LoadConfigurationPlan.java
```diff
@@ -17,19 +17,107 @@
...
+      stream.writeInt(propertiesArray.length);
+      for (Properties properties : propertiesArray) {
+        if (properties == null) {
+          stream.writeInt(0);
+        } else {
+          stream.writeInt(1);
+          Map<String, String> propertiesMap = new HashMap<>();
+          for (Entry<Object, Object> entry : properties.entrySet()) {
+            propertiesMap.put(String.valueOf(entry.getKey()), String.valueOf(entry.getValue()));
+          }
```
jt2594838, line 91: I wonder if it is necessary to put the properties into a new map, as you have the entrySet, you can just write them into the stream within this loop.

server/src/main/java/org/apache/iotdb/db/qp/physical/sys/LoadConfigurationPlan.java
```diff
@@ -17,19 +17,107 @@
...
+    if (loadConfigurationPlanType == LoadConfigurationPlanType.GLOBAL) {
+      int propertiesNum = buffer.getInt();
+      propertiesArray = new Properties[propertiesNum];
+      for (int i = 0; i < propertiesArray.length; i++) {
+        if (buffer.getInt() == 1) {
+          propertiesArray[i] = new Properties();
+          int size = buffer.getInt();
+          Map<String, String> values = new HashMap<>(size);
+          for(int j = 0; j < size; j++){
+            values.put(readString(buffer), readString(buffer));
```
jt2594838, line 115: Maybe you could just use `Properties.setProperty` and discard the map.

cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterDescriptor.java
Ring-k: It seems that the check there is unnecessary. I will remove it.

server/src/main/java/org/apache/iotdb/db/qp/physical/sys/LoadConfigurationPlan.java
Ring-k: Thanks for your reminding.

server/src/main/java/org/apache/iotdb/db/qp/physical/sys/LoadConfigurationPlan.java
Ring-k: Thanks. It is not necessary.

PR #1103, opened 2020-04-27
[IOTDB-617] Support alter one time series's tag/attribute property

JackieTien97:


server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -959,10 +966,77 @@ public void changeOffset(String path, long offset) throws MetadataException {
...
+        if (beforeValue != null && !beforeValue.equals(value)) {
+          tagIndex.get(key).get(beforeValue).remove(leafMNode);
+          if (tagIndex.get(key).get(beforeValue).isEmpty()) {
+            tagIndex.get(key).remove(beforeValue);
+          }
+        }
+
+        // if the key doesn't exist or the value is not equal to the new one
+        // we should add a new key-value to inverted index map
+        if (beforeValue == null || !beforeValue.equals(value)) {
```
qiaojialin, line 1019:
```suggestion
        } else {
        // if the key doesn't exist or the value is not equal to the new one
        // we should add a new key-value to inverted index map
```

PR #1112, opened 2020-04-28
[IOTDB-354] Implement batch append entries

SilverNarcissus:
Implement batch append entries


cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterFillExecutor.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.Set;
+import org.apache.iotdb.cluster.server.member.MetaGroupMember;
+import org.apache.iotdb.db.exception.StorageEngineException;
+import org.apache.iotdb.db.query.context.QueryContext;
+import org.apache.iotdb.db.query.executor.FillQueryExecutor;
+import org.apache.iotdb.db.query.fill.IFill;
+import org.apache.iotdb.tsfile.file.metadata.enums.TSDataType;
+import org.apache.iotdb.tsfile.read.common.Path;
+
+public class ClusterFillExecutor extends FillQueryExecutor {
```
jt2594838, line 33: The class is moved to another package and modified, you should remove it.

SilverNarcissus: Sure. I will fix it~

cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
```diff
@@ -384,7 +385,89 @@ public void appendEntry(AppendEntryRequest request, AsyncMethodCallback resultHa
...
+
+  /**
+   * Find the local previous log of "log". If such log is found, discard all local logs behind it
+   * and append "log" to it. Otherwise report a log mismatch.
+   *
+   * @param logs
+   * @return Response.RESPONSE_AGREE when the log is successfully appended or Response
+   * .RESPONSE_LOG_MISMATCH if the previous log of "log" is not found.
+   */
+  private long appendEntries(List<Log> logs) {
```
LebronAl, line 420: I've found that this way of appendEntries has a high concurrency bug, this is the fix [document](https://shimo.im/docs/JWhvXw8vKyqvt6X8). Please refer to function `appendEntry`'s implementation.

jt2594838: I guess LebronAl is talking about the chance that the logs are disjoint with the local last log, but as `appendEntries` is only used in catch-up, I think the possibility of that case is relatively low.

Still, it would be better to check the index of the first coming log and the last local log, which I previously hoped to be done inside the LogManager. As you can see, other users may not bother to do the checks (the may not know the checks actually), so any necessary checks are better to be done within underlined classes.

LebronAl: Actually,I used to think that the interface raftLogManager exposed to followers to call append logs was `maybeAppend`, and in that function we can handle this check, but now appendEntry's RPC doesn't have lastLogTerm, lastLogIndex, leaderCommit. So i have to use `append `.
So now we can either add some fields to the appendEntry rpc to support the use of maybeAppend, or we can do this check inside `append`. What's your opinion?

LebronAl: In my view, `maybeAppend ` will make catchUp more fault-tolerant.When catching up in batch, even if the index of the first log sent is smaller than follower's lastLogIndex, some previous logs can be removed. Otherwise, if the index of the first Log is smaller than  follower's commitindex, such a batch of logs cannot be appended.

cluster/src/main/java/org/apache/iotdb/cluster/log/catchup/LogCatchUpTask.java
```diff
@@ -39,17 +43,25 @@
...
+  private List<Log> logs;
+  private boolean useBatch = false;
 
   public LogCatchUpTask(List<Log> logs, Node node, RaftMember raftMember) {
     this.logs = logs;
     this.node = node;
     this.raftMember = raftMember;
   }
 
+  public LogCatchUpTask(List<Log> logs, Node node, RaftMember raftMember, boolean useBatch) {
```
LebronAl, line 58: I think maybe it's better to make `useBatch` a configuration item instead of calling it manually every time

SilverNarcissus: Sure, I will fix it

cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
```diff
@@ -398,7 +399,92 @@ public void appendEntry(AppendEntryRequest request, AsyncMethodCallback resultHa
...
+      return Response.RESPONSE_AGREE;
+    }
+
+    long resp;
+    synchronized (logManager) {
+      if (logs.get(0).getCurrLogIndex() > logManager.getLastLogIndex() + 1) {
+        // the incoming log points to an illegal position, reject it
+        resp = Response.RESPONSE_LOG_MISMATCH;
+      } else {
+        logManager.append(logs);
```
jt2594838, line 445: LebronAl has now provided a safer method called `maybeAppendBatch` in `RaftLogManager`, please switch to that and check the return value.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
```diff
@@ -398,7 +399,92 @@ public void appendEntry(AppendEntryRequest request, AsyncMethodCallback resultHa
...
+      AsyncMethodCallback resultHandler) {
+    long leaderTerm = request.getTerm();
+    long localTerm;
+
+    synchronized (term) {
+      // if the request comes before the heartbeat arrives, the local term may be smaller than the
+      // leader term
+      localTerm = term.get();
+      if (leaderTerm < localTerm) {
+        logger.debug("{} rejected the AppendEntryRequest for term: {}/{}", name, leaderTerm,
```
jt2594838, line 474: The implementation of `checkRequestTerm` for `AppendEntryRequest` has been changed slightly, so please see to it and make necessary modifications.
By the way, the log message should use `AppendEntriesRequest` here instead of `AppendEntriesRequest`.

SilverNarcissus: Fixed~

cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
SilverNarcissus: Fine~ Fixed~

PR #1126, opened 2020-04-29
[IOTDB-620] Add the ability to persist and maintain voteFor for the raft node and amend the judgment conditions for elections

LebronAl:


cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
```diff
@@ -516,11 +546,20 @@ public void setLastHeartbeatReceivedTime(long lastHeartbeatReceivedTime) {
 
   public void setLeader(Node leader) {
     if (!Objects.equals(leader, this.leader)) {
-      logger.info("{} has become a follower of {}", getName(), leader);
+      if (!Objects.equals(leader, this.thisNode)) {
+        logger.info("{} has become a follower of {}", getName(), leader);
+      }
       this.leader = leader;
```
jt2594838, line 552: It seemed strange here.

PR #1128, opened 2020-04-29
Using version controler to build new file in log 

SilverNarcissus:
Using version controler to build new file in log


cluster/src/main/java/org/apache/iotdb/cluster/log/manage/serializable/SyncLogDequeSerializer.java
```diff
@@ -71,6 +73,8 @@
   private long maxAvailableTime = Long.MAX_VALUE;
```
jt2594838, line 73: I think you should also change the names and comments of `minAvailableTime` and `maxAvailableTime`.

PR #1129, opened 2020-04-29
[IOTDB-630] Add a jdbc-like way to fetch data in session

JackieTien97:


example/session/src/main/java/org/apache/iotdb/SessionExample.java
```diff
@@ -193,15 +195,11 @@ private static void insertRecords() throws IoTDBConnectionException, BatchExecut
...
-   *
-   *      device1
-   * time s1, s2, s3
-   * 1,   1,  1,  1
-   * 2,   2,  2,  2
-   * 3,   3,  3,  3
-   *
+   * <p>
+   * device1 time s1, s2, s3 1,   1,  1,  1 2,   2,  2,  2 3,   3,  3,  3
+   * <p>
```
qiaojialin, line 202: format back

PR #1134, opened 2020-04-30
[IOTDB-631] Using new TsFile MetadataIndex to optimize query and cache

samperson1997:
As the new TsFile MetadataIndex is designed, we could do more optimization for querying process and TimeseriesMetadata Cache.
* provide a new interface `List<TimeseriesMetadata> readTimeseriesMetadata` to get more sensors' TimeseriesMetadata and reduce IO.
* use a new method `int binarySearchInTimeseriesMetadataList` for binary searching in TimeseriesMetadataList

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/cache/TimeSeriesMetadataCache.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/cache/TimeSeriesMetadataCache.java
@@ -19,6 +19,13 @@
 
 package org.apache.iotdb.db.engine.cache;
 
+import java.io.IOException;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 import org.apache.iotdb.db.conf.IoTDBConfig;
 import org.apache.iotdb.db.conf.IoTDBConstant;
 import org.apache.iotdb.db.conf.IoTDBDescriptor;
@@ -30,22 +37,17 @@ import org.apache.iotdb.tsfile.read.common.Path;
 import org.apache.iotdb.tsfile.utils.BloomFilter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import java.io.IOException;
-import java.util.Objects;
-import java.util.Set;
-import java.util.concurrent.atomic.AtomicLong;
-import java.util.concurrent.locks.ReadWriteLock;
-import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 /**
- * This class is used to cache <code>TimeSeriesMetadata</code> in IoTDB. The caching
- * strategy is LRU.
+ * This class is used to cache <code>TimeSeriesMetadata</code> in IoTDB. The caching strategy is
+ * LRU.
  */
 public class TimeSeriesMetadataCache {
 
   private static final Logger logger = LoggerFactory.getLogger(TimeSeriesMetadataCache.class);
   private static final IoTDBConfig config = IoTDBDescriptor.getInstance().getConfig();
-  private static final long MEMORY_THRESHOLD_IN_TIME_SERIES_METADATA_CACHE = config.getAllocateMemoryForTimeSeriesMetaDataCache();
+  private static final long MEMORY_THRESHOLD_IN_TIME_SERIES_METADATA_CACHE = config
+      .getAllocateMemoryForTimeSeriesMetaDataCache();
   private static boolean cacheEnable = config.isMetaDataCacheEnable();
 
   private final LRULinkedHashMap<TimeSeriesMetadataCacheKey, TimeseriesMetadata> lruCache;
@@ -58,9 +60,11 @@ public class TimeSeriesMetadataCache {
 
   private TimeSeriesMetadataCache() {
     logger.info("TimeseriesMetadataCache size = " + MEMORY_THRESHOLD_IN_TIME_SERIES_METADATA_CACHE);
-    lruCache = new LRULinkedHashMap<TimeSeriesMetadataCacheKey, TimeseriesMetadata>(MEMORY_THRESHOLD_IN_TIME_SERIES_METADATA_CACHE, true) {
+    lruCache = new LRULinkedHashMap<TimeSeriesMetadataCacheKey, TimeseriesMetadata>(
+        MEMORY_THRESHOLD_IN_TIME_SERIES_METADATA_CACHE, true) {
       int count = 0;
       long averageSize = 0;
+
       @Override
       protected long calEntrySize(TimeSeriesMetadataCacheKey key, TimeseriesMetadata value) {
         if (count < 10) {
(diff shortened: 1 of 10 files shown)


server/src/main/java/org/apache/iotdb/db/engine/cache/TimeSeriesMetadataCache.java
```diff
@@ -125,9 +130,20 @@ public TimeseriesMetadata get(TimeSeriesMetadataCacheKey key, Set<String> allSen
...
+
+      List<TimeseriesMetadata> timeSeriesMetadataList = reader
+          .readTimeseriesMetadata(key.device, allSensors);
+      if (!allSensors.isEmpty()) {
+        // put TimeSeriesMetadata of all sensors used in this query into cache
+        timeSeriesMetadataList.forEach(timeseriesMetadata -> {
+          lruCache.put(new TimeSeriesMetadataCacheKey(key.filePath, key.device,
+              timeseriesMetadata.getMeasurementId()), timeseriesMetadata);
+        });
+      }
```
qiaojialin, line 145: Hi, the key.measurement should be also in the allSensors set. Therefore, we only need to query the allSensor, put them into cache. Then return the queried sensor's TimeseriesMetadata.

tsfile/src/main/java/org/apache/iotdb/tsfile/read/TsFileSequenceReader.java
```diff
@@ -314,28 +311,76 @@ public TsFileMetadata readFileMetadata() throws IOException {
...
+        MetadataIndexNode metadataIndexNode = MetadataIndexNode.deserializeFrom(buffer);
+        measurementMetadataIndexPair = getMetadataAndEndOffset(metadataIndexNode,
+            measurement, MetadataIndexNodeType.INTERNAL_MEASUREMENT);
+      }
+      buffer = readData(measurementMetadataIndexPair.left.getOffset(),
+          measurementMetadataIndexPair.right);
+      while (buffer.hasRemaining()) {
+        timeseriesMetadataList.add(TimeseriesMetadata.deserializeFrom(buffer));
+      }
+      int searchResult = binarySearchInTimeseriesMetadataList(timeseriesMetadataList,
```
qiaojialin, line 356: check if the size of measurements reaches a threshold, just traversing from start to end is quicker, no need to binary search for each

tsfile/src/main/java/org/apache/iotdb/tsfile/read/TsFileSequenceReader.java
```diff
@@ -314,28 +311,76 @@ public TsFileMetadata readFileMetadata() throws IOException {
...
+        path.getMeasurement());
+    return searchResult >= 0 ? timeseriesMetadataList.get(searchResult) : null;
+  }
+
+  public List<TimeseriesMetadata> readTimeseriesMetadata(String device, Set<String> measurements)
+      throws IOException {
+    readFileMetadata();
+    MetadataIndexNode deviceMetadataIndexNode = tsFileMetaData.getMetadataIndex();
+    Pair<MetadataIndexEntry, Long> metadataIndexPair = getMetadataAndEndOffset(
+        deviceMetadataIndexNode, device, MetadataIndexNodeType.INTERNAL_DEVICE);
```
qiaojialin, line 339: I'm not clear about the MetadataIndexNodeType.INTERNAL_DEVICE, how can I see this parameter, the queried node type? Shouldn't this be Leaf_Device?

server/src/main/java/org/apache/iotdb/db/engine/cache/TimeSeriesMetadataCache.java
samperson1997: Great! Fixed.

tsfile/src/main/java/org/apache/iotdb/tsfile/read/TsFileSequenceReader.java
samperson1997: How to define the threshold? So should we use 2 different methods for 2 situation (measurements number > threshold and < threshold)?

tsfile/src/main/java/org/apache/iotdb/tsfile/read/TsFileSequenceReader.java
samperson1997: I could explain why I use INTERAL_DEVICE / INTERNAL_MEASUREMENT in method `getMetadataAndEndOffset`:

The statement which ends the recursion is: `if (!childIndexEntry.left.getChildNodeType().equals(type))`. For example, when searching for a device node, we look for LEAF_DEVICE, so we return when it is not INTERNAL_DEVICE. Likewise, when searching for a measurement node, we look for LEAF_MEASUREMENT, so we return when it is not INTERNAL_MEASUREMENT.

Why not judging by `== LEAF_DEVICE` or `== LEAF_MEASUREMENT`? Because this works for the situation when we are searching for device, but the index tree does **not** have the **device level** and only has the **measurement level**. Or we will never meet the LEAF_DEVICE.

server/src/main/java/org/apache/iotdb/db/engine/cache/TimeSeriesMetadataCache.java
samperson1997: Hi, when I tried to only query the allSensor, I found some tests failed... which means that the allSensors set doesn't contain key.measurement... Is it a bug in query module? Or it was designed as this?

samperson1997:
Hi @qiaojialin , thanks a lot for your review! I have taken your suggestions with a little bit change as is mentioned below:
* I've noticed that `allSensors` may not contain `key.measurement` during executing tests, so I put the key in the `allSensors` before sequenceReader's read process.
* I've taken the `max_degree_of_index_node` for the measurement number's threshold, do you think it is acceptable? : )

server/src/main/java/org/apache/iotdb/db/engine/cache/TimeSeriesMetadataCache.java
qiaojialin: it's a bug..

tsfile/src/main/java/org/apache/iotdb/tsfile/read/TsFileSequenceReader.java
qiaojialin: yes， just calculate the search number is ok

server/src/main/java/org/apache/iotdb/db/query/timegenerator/ServerTimeGenerator.java
```diff
@@ -33,8 +36,6 @@
 import org.apache.iotdb.tsfile.read.query.timegenerator.TimeGenerator;
 import org.apache.iotdb.tsfile.read.reader.IBatchReader;
 
-import java.io.IOException;
-
 /**
  * A timestamp generator for query with filter. e.g. For query clause "select s1, s2 form root where
  * s3 < 0 and time > 100", this class can iterate back to every timestamp of the query.
```
HTHou, line 41: For query clause "select s1, s2 **form** root -> For query clause "select s1, s2 **from** root

samperson1997: Thanks!  Good catch : )

PR #1150, opened 2020-05-02
[IOTDB-636]Grafana connector does not use correct time unit

jixuan1989:
On rel/0.9, the grafana connector uses the time unit like this:

```
@Autowired
public BasicDaoImpl(JdbcTemplate jdbcTemplate) {
  this.jdbcTemplate = jdbcTemplate;
  Properties properties = new Properties();
  String tsPrecision = properties.getProperty("timestamp_precision", "ms");
  switch (tsPrecision) {
    case "us":
      TIMESTAMP_RADIX = 1000;
      break;
    case "ns":
      TIMESTAMP_RADIX = 1000_000;
      break;
    default:
      TIMESTAMP_RADIX = 1;
  }
  logger.info("Use timestamp precision {}", tsPrecision);
}

```
That is to say, it always is "ms"! No configuration file is read!

### DIFF
--- a/docs/UserGuide/7-Ecosystem Integration/1-Grafana.md
+++ b/docs/UserGuide/7-Ecosystem Integration/1-Grafana.md
@@ -153,7 +153,7 @@ isDownSampling=true
 # defaut sampling intervals
 interval=1m
 # aggregation function to use to downsampling the data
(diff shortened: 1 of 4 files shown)


samperson1997:
Hi, extraordinary catch. Is it also a bug in master branch ... ?

jixuan1989:
> Hi, extraordinary catch. Is it also a bug in master branch ... ?

Hive fixed it in PR #1149

PR #1157, opened 2020-05-06
[IOTDB-567] Apache IoTDB's Peer Tracker for The Raft Protocol

LebronAl:


cluster/src/main/java/org/apache/iotdb/cluster/log/catchup/CatchUpTask.java
```diff
@@ -0,0 +1,169 @@
...
+        // if follower return RESPONSE.AGREE with this empty log, then start sending real logs from this emptyContentLog's index.
+        // which means the log which index is matchIndex will be send twice, but at the first time it sent an empty log.
+        logs.subList(0, index).clear();
+        if (logger.isDebugEnabled()) {
+          logger.debug("{} makes {} catch up with {} cached logs", raftMember.getName(), node,
+              logs.size());
+        }
+        return false;
+      }
+      index--;
```
jt2594838, line 133: It would be better to perform a binary search, but it is fine for now.

cluster/src/main/java/org/apache/iotdb/cluster/log/catchup/CatchUpTask.java
```diff
@@ -0,0 +1,169 @@
...
+
+  public void run() {
+    try {
+      if (checkMatchIndex()) {
+        SnapshotCatchUpTask task = new SnapshotCatchUpTask(logs, snapshot, node, raftMember);
+        task.run();
+      } else {
+        LogCatchUpTask task = new LogCatchUpTask(logs, node, raftMember);
+        task.run();
+      }
```
jt2594838, line 156: Batched log catch-up has been introduced, better to switch to better.

cluster/src/main/java/org/apache/iotdb/cluster/log/catchup/CatchUpTask.java
LebronAl: Inconsistencies are generally rare, and there aren't many missing logs, so this may actually be enough. If a performance bottleneck occurs, I will modify here ~

cluster/src/main/java/org/apache/iotdb/cluster/log/catchup/CatchUpTask.java
LebronAl: Batched log catch-up or not is only related to the configuration item, and the code here supports batched log catch-up if related configuration item has been set~

PR #1158, opened 2020-05-06
Add Grafana IoTDB Bridge Artefact to distrbution in tools/grafana fol…

JulianFeinauer:
…der.

NOTE: Should also be picked to develop.

### DIFF
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -96,5 +96,11 @@
             <version>${project.version}</version>
             <type>zip</type>
         </dependency>
+        <dependency>
+            <groupId>org.apache.iotdb</groupId>
+            <artifactId>iotdb-grafana</artifactId>
+            <version>${project.version}</version>
+            <type>war</type>
+        </dependency>
     </dependencies>
 </project>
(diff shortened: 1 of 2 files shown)


JulianFeinauer:
@chrisdutz thanks for the hint I changed it to use the dependencySet. You can have a look if you like :)

chrisdutz:
LGTM now :-)

jixuan1989:
LGTM.
Chris is an expert of Maven... If he says "LGTM", it really looks good for how to use Maven :D

JulianFeinauer:
Oh, sorry too late @jixuan1989. We could do separate downloads if we use multiple assembly files. Then user can decide what to download. Thats the way to go, I guess?

jixuan1989:
It is ok. I have approved it...

Maybe we can separate it in v0.10 and release more connectors at that time.

JulianFeinauer:
+1. So I do not pick it for master, yet?
WDYT? (currently about to do that :) ).

jixuan1989:
If we can not implement separating these binaries into several files quickly, I think we can pick this commit to master... to keep consistent between 0.9 and the master branch...
And then consider separating them as a new feature. 
How do you think?

jixuan1989:
Besides, when we are in the incubator, there is an issue that we can retain at most 10 files. Otherwise, there will be a warning on this page to say incubator project can have at most 10 files, please delete stale version files: http://incubator.apache.org/clutch/iotdb.html (the Release Chapter).

PR #1161, opened 2020-05-06
[IOTDB-348] Fix DeleteTimeseriesPlan result inconsistent problem

Ring-k:
This pull request fixes DeleteTimeseriesPlan result inconsistent problem. 

Some meta data checks are skipped in the cluster version. Some messages sent to MManager are now replaced by broadcast in the cluster.

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
@@ -445,50 +445,6 @@ public class ClusterPlanExecutor extends PlanExecutor {
   }
 
   @Override
-  protected boolean deleteTimeSeries(DeleteTimeSeriesPlan deleteTimeSeriesPlan) throws QueryProcessException {
-    List<Path> deletePathList = deleteTimeSeriesPlan.getPaths();
-    try {
-      deleteDataOfTimeSeries(deletePathList);
-      Set<String> emptyStorageGroups = new HashSet<>();
-      List<String> failedNames = new LinkedList<>();
-      for (Path path : deletePathList) {
-        Pair<Set<String>, String> pair = deleteTimeseries(path.toString());
-        emptyStorageGroups.addAll(pair.left);
-        if (!pair.right.isEmpty()) {
-          failedNames.add(pair.right);
-        }
-      }
-      for (String deleteStorageGroup : emptyStorageGroups) {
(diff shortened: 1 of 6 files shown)


cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
```diff
@@ -373,8 +373,85 @@ protected void loadConfiguration(LoadConfigurationPlan plan) throws QueryProcess
...
+      Node header = group.getHeader();
+      if (header.equals(metaGroupMember.getThisNode())) {
+        continue;
+      }
+      pool.submit(() -> {
+        DeleteTimeseriesHandler handler = new DeleteTimeseriesHandler();
+        AtomicReference<DeleteTimeseriesRespPair> response = new AtomicReference<>(null);
+        handler.setResponse(response);
+
+        for (Node node : group) {
```
jt2594838, line 399: As a node exists in multiple groups, a node will perform the same deletion several times and I am not sure if it is meaningful. Why not just send the deletion to each node in allNodes?

cluster/src/main/java/org/apache/iotdb/cluster/server/handlers/caller/DeleteTimeseriesHandler.java
```diff
@@ -0,0 +1,56 @@
...
+    logger.info("Received delete timeseries result from {}", contact);
+    synchronized (result) {
+      result.set(resp);
+      result.notifyAll();
+    }
+  }
+
+  @Override
+  public void onError(Exception exception) {
+    logger.warn("Cannot delete timeseries from {}, because ", contact, exception);
```
jt2594838, line 46: I think you should notify `result` when an error occurs.

cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
Ring-k: Thanks for your advice. I've changed the logic.

cluster/src/main/java/org/apache/iotdb/cluster/server/handlers/caller/DeleteTimeseriesHandler.java
Ring-k: Thanks for your reminding.

PR #1166, opened 2020-05-07
[IOTDB-642] Data can't be inserted correctly by alias

Alima777:
If data is inserted using alias, it will be stored by alias path which leads to the empty result in query.

set storage group to root.sg;
create timeseries root.sg.d1.s1(speed)with datatype=FLOAT, encoding=RLE;
insert into root.sg.d1(timestamp,speed) values(1,1);
select s1 from root.sg.d1;
select speed from root.sg.d1;

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
+++ b/server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
@@ -871,7 +871,7 @@ public class PlanExecutor implements IPlanExecutor {
         }
         LeafMNode measurementNode = (LeafMNode) node.getChild(measurement);
         schemas[i] = measurementNode.getSchema();
-        if (measurement != measurementNode.getName()){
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
```diff
@@ -871,8 +871,12 @@ public void insert(InsertPlan insertPlan) throws QueryProcessException {
         }
         LeafMNode measurementNode = (LeafMNode) node.getChild(measurement);
         schemas[i] = measurementNode.getSchema();
+        if (measurement != measurementNode.getName()){
```
HTHou, line 874:
```suggestion
        schemas[i] = measurementNode.getSchema();
        if (measurement != measurementNode.getName()) {
```

Alima777: Fixed~

PR #1181, opened 2020-05-09
[IOTDB-659] Move the same construct dataset logic to service-rpc

JackieTien97:


service-rpc/src/main/java/org/apache/iotdb/rpc/AbstractIoTDBDataSet.java
```diff
@@ -0,0 +1,406 @@
...
+import org.apache.iotdb.service.rpc.thrift.TSQueryDataSet;
+import org.apache.iotdb.service.rpc.thrift.TSStatus;
+import org.apache.iotdb.tsfile.exception.write.UnSupportedDataTypeException;
+import org.apache.iotdb.tsfile.file.metadata.enums.TSDataType;
+import org.apache.iotdb.tsfile.utils.BytesUtils;
+import org.apache.iotdb.tsfile.utils.ReadWriteIOUtils;
+import org.apache.thrift.TException;
+
+
+public class AbstractIoTDBDataSet {
```
qiaojialin, line 42:
```suggestion
public class IoTDBRPCDataSet {
```

PR #1194, opened 2020-05-12
[IOTDB-667] add retry in session and jdbc.execute

qiaojialin:
Retry to shield the connection reset in thrift.

### DIFF
--- a/session/src/main/java/org/apache/iotdb/session/Session.java
+++ b/session/src/main/java/org/apache/iotdb/session/Session.java
@@ -200,10 +200,13 @@ public class Session {
         }
         isClosed = true;
         open(enableRPCCompression, connectionTimeoutInMs);
+        flag = true;
+        break;
       } catch (Exception e) {
         try {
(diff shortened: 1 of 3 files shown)


session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -185,6 +191,29 @@ public synchronized void close() throws IoTDBConnectionException {
...
+        isClosed = true;
+        open(enableRPCCompression, connectionTimeoutInMs);
+        flag = true;
+        break;
+      } catch (Exception e) {
+        try {
+          Thread.sleep(RETRY_INTERVAL);
+        } catch (InterruptedException e1) {
+          Thread.currentThread().interrupt();
+          logger.error("reconnect is interrupted.", e1);
```
jt2594838, line 210: reconnect -> reconnection

qiaojialin: fixed

PR #1198, opened 2020-05-12
[IOTDB-355] Start up checks for seed nodes

Ring-k:
This pull request implements the start up checks for seed nodes. It mainly modifies the `addSeedNodes` method in `MetaGroupMember`.
The seed nodes should have identical values of partition interval, hash salt and relication number.


cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -319,31 +333,119 @@ private void initSubServers() throws TTransportException, StartupException {
...
+  }
+
+  private boolean analyseStartUpCheckResult(int consistentNum, int inconsistentNum,
+      int totalSeedNum, long timeElapsed) {
+    if (consistentNum >= (totalSeedNum + 1) / 2) {
+      // break the loop and establsih the cluster
+      return true;
+    } else if (inconsistentNum >= (totalSeedNum + 1) / 2) {
+      // this node is not consistent with the cluster, shut down
+      logger.debug("The configuration of this node is inconsistent with the cluster.");
```
jt2594838, line 431: The log levels should be "error" here.

service-rpc/src/main/thrift/cluster.thrift
```diff
@@ -370,10 +370,12 @@ service TSMetaService extends RaftService {
   **/
   AddNodeResponse addNode(1: Node node, 2: StartUpStatus startUpStatus)
 
+  CheckStatusResponse  checkStatus(1: StartUpStatus startUpStatus)
+
   /**
   * Remove a node from the cluster. If the node is not in the cluster or the cluster size will
   * less than replication number, the request will be rejected.
-  * return -1(RESPONSE_AGREE) or -3(RESPONSE_REJECT) or -9(RESPONSE_CLUSTER_TOO_SMALL)
+  * return -1(RESPONSE_AGREE) or -3(RESPONSE_REJECT) or -9(RESPONSE_CLUSTEMetaGroupMemberR_TOO_SMALL)
```
jt2594838, line 378: Please revert this.

jt2594838:
I think there is one more thing to check, the `seed_nodes`.
If the checker is already in a cluster, then it should confirm that the `seed_nodes` it receives is a subset of its `allNodes`.
If the checker is not in a cluster, which means the initial cluster is to be built, then it should confirm that the `seed_nodes` is exactly the same as its.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -905,6 +936,63 @@ private boolean processAddNodeLocally(Node node, StartUpStatus startUpStatus,
...
+      List<Node> remoteSeedNodes) {
+    return isClusterEstablished ? seedNodesContains(localSeedNodes, remoteSeedNodes)
+        : seedNodesEquals(localSeedNodes, remoteSeedNodes);
+  }
+
+  private boolean seedNodesEquals(List<Node> thisNodeList, List<Node> thatNodeList) {
+    Node[] thisNodeArray = thisNodeList.toArray(new Node[0]);
+    Node[] thatNodeArray = thatNodeList.toArray(new Node[0]);
+    Arrays.sort(thisNodeArray, this::compareSeedNode);
+    Arrays.sort(thatNodeArray, this::compareSeedNode);
```
jt2594838, line 949: Is it necessary to copy them to an array rather than just use `List.sort()`?

PR #1204, opened 2020-05-13
[IOTDB-622] add count records for all databases

mychaow:


server/src/main/java/org/apache/iotdb/db/qp/physical/crud/GroupByPlan.java
```diff
@@ -33,6 +33,8 @@
   // if it is left close and right open interval
   private boolean leftCRightO = true;
 
+  private boolean byTime = false;
```
qiaojialin, line 36: The GroupByPlan is actually GroupByTimePlan, so no need to add a byTime field, we can rename the class.

server/src/main/java/org/apache/iotdb/db/query/dataset/groupby/GroupByLevelDataSet.java
```diff
@@ -0,0 +1,217 @@
...
+      }
+    }
+
+    // prepare final records
+    // group by level
+    // path -> count
+    Map<String, Long> finalPaths = new TreeMap<String, Long>();
+
+    for (int j = 0; j < this.paths.size(); j++) {
+      String[] tmpPath = this.paths.get(j).getFullPath().split("\\.");
```
qiaojialin, line 95: Actually, we support path like this:  root.sg.d."s.1"
s.1 is measurement name

server/src/main/java/org/apache/iotdb/db/query/dataset/groupby/GroupByLevelDataSet.java
```diff
@@ -0,0 +1,217 @@
...
+        for (int k = 0; k <= level; k++) {
+          if (k == 0) {
+            path.append(tmpPath[k]);
+          } else {
+            path.append("." + tmpPath[k]);
+          }
+        }
+        key = path.toString();
+      }
+      logger.debug("records " + record.getFields().toString());
```
qiaojialin, line 111: add if (logger.isDebugEnabled())

server/src/main/java/org/apache/iotdb/db/query/dataset/groupby/GroupByLevelDataSet.java
```diff
@@ -0,0 +1,217 @@
...
+      if (res == null) {
+        record.addField(null);
+        continue;
+      }
+      record.addField(res.getResult(), res.getResultDataType());
+    }
+    return record;
+  }
+
+  private RowRecord mergeRecords(RowRecord newRecord, RowRecord oldRecord) {
```
qiaojialin, line 199: please add javadoc and an example

server/src/main/java/org/apache/iotdb/db/query/dataset/groupby/GroupByLevelDataSet.java
```diff
@@ -0,0 +1,217 @@
...
+  protected boolean hasNextWithoutConstraint() throws IOException {
+    return index < records.size();
+  }
+
+  @Override
+  protected RowRecord nextWithoutConstraint() {
+    return records.get(index++);
+  }
+
+  private void initGroupBy()
```
qiaojialin, line 144: initGroupByLevel?

server/src/main/java/org/apache/iotdb/db/query/executor/QueryRouter.java
```diff
@@ -113,34 +114,45 @@ protected AggregationExecutor getAggregationExecutor(AggregationPlan aggregation
...
   @Override
   public QueryDataSet groupBy(GroupByPlan groupByPlan, QueryContext context)
-      throws QueryFilterOptimizationException, StorageEngineException, QueryProcessException {
-    long unit = groupByPlan.getInterval();
-    long slidingStep = groupByPlan.getSlidingStep();
-    long startTime = groupByPlan.getStartTime();
-    long endTime = groupByPlan.getEndTime();
+    throws QueryFilterOptimizationException, StorageEngineException, QueryProcessException, IOException {
+
+    logger.debug("paths:" + groupByPlan.getPaths() + " level:" + groupByPlan.getLevel() + " byTime:" + groupByPlan.isByTime());
```
qiaojialin, line 119: check if debug enabled

server/src/main/java/org/apache/iotdb/db/query/dataset/groupby/GroupByLevelDataSet.java
```diff
@@ -0,0 +1,217 @@
...
+  private static final Logger logger = LoggerFactory
+    .getLogger(GroupByLevelDataSet.class);
+
+  private List<RowRecord> records = new ArrayList<>();
+  private int index = 0;
+
+  private int level = -1;
+  private boolean isByTime = false;
+  protected long queryId;
+  private GroupByEngineDataSet dataSet;
```
samperson1997, line 56: `dataSet` is not used as a global variable. Actually, `level` and `isByTime` could also be converted to local variables

server/src/main/antlr4/org/apache/iotdb/db/qp/strategy/SqlBase.g4
```diff
@@ -241,6 +241,13 @@ groupByClause
       COMMA DURATION
       (COMMA DURATION)?
       RR_BRACKET
+    | GROUP BY LEVEL OPERATOR_EQ INT
+    | GROUP BY LR_BRACKET
+            timeInterval
+            COMMA DURATION
+            (COMMA DURATION)?
+            RR_BRACKET
+            COMMA LEVEL OPERATOR_EQ INT
```
qiaojialin, line 250: No need to modify the group by ([1,10), 1ms) clause.

server/src/main/java/org/apache/iotdb/db/query/dataset/groupby/GroupByLevelDataSet.java
```diff
@@ -0,0 +1,217 @@
...
+      record.addField(res.getResult(), res.getResultDataType());
+    }
+    return record;
+  }
+
+  private RowRecord mergeRecords(RowRecord newRecord, RowRecord oldRecord) {
+    if (oldRecord.getFields().isEmpty()) {
+      oldRecord.setFields(newRecord.getFields());
+    } else {
+      assert oldRecord.getFields().size() == newRecord.getFields().size();
```
samperson1997, line 203: I think you'd better use `logger.error` instead of assertion, what do you think?

server/src/main/java/org/apache/iotdb/db/qp/strategy/LogicalGenerator.java
```diff
@@ -664,6 +668,7 @@ public void enterGroupByFillClause(SqlBaseParser.GroupByFillClauseContext ctx) {
     super.enterGroupByFillClause(ctx);
     queryOp.setGroupBy(true);
     queryOp.setFill(true);
+    queryOp.setGroupByTime(true);
```
qiaojialin, line 671: since there is more than one group by, it's better to use full name.   GroupByLevel and GroupByTime

server/src/main/java/org/apache/iotdb/db/query/dataset/groupby/GroupByLevelDataSet.java
```diff
@@ -0,0 +1,217 @@
...
+    .getLogger(GroupByLevelDataSet.class);
+
+  private List<RowRecord> records = new ArrayList<>();
+  private int index = 0;
+
+  private int level = -1;
+  private boolean isByTime = false;
+  protected long queryId;
+  private GroupByEngineDataSet dataSet;
+  private GroupByPlan groupByPlan;
```
qiaojialin, line 57: Replace this with an AggregationPlan

samperson1997:
Besides, maybe you should maintain the User Guide documents (in `chapter 5.2` and `chapter 5.4`) and System Design documents (in `chapter 5.6`)

samperson1997:
> We could add the group by level=n to AggregationPlan, not the GroupByPlan.
> 
> The group by plan segment the query interval [1,10) to many 2ms ranges and do aggregation in each range. Add a level=1 may have confusion.
> e.g., select count(s1) from root.turbine1.d1 group by ([1,10),2ms), level=1

Actually I think maybe we should define a new type of plan, which extends AggregationPlan? From my point of view, a new plan would be better instead of just using a field to classify, and the code would be more clear

server/src/main/java/org/apache/iotdb/db/query/dataset/groupby/GroupByLevelDataSet.java
mychaow: ok, assert will help me found bug fastly. I can change it to error.

mychaow: And, I will maintain the User Guide docluments after merge.

server/src/main/java/org/apache/iotdb/db/query/executor/QueryRouter.java
mychaow: ok

server/src/main/java/org/apache/iotdb/db/query/dataset/groupby/GroupByLevelDataSet.java
```diff
@@ -0,0 +1,208 @@
...
+  private GroupByTimePlan groupByTimePlan;
+  private QueryContext context;
+
+  private Map<Path, GroupByExecutor> pathExecutors = new HashMap<>();
+  private Map<Path, List<Integer>> resultIndexes = new HashMap<>();
+
+  public GroupByLevelDataSet(QueryContext context, GroupByTimePlan plan, GroupByEngineDataSet dataSet)
+    throws QueryProcessException, StorageEngineException, IOException {
+    this.queryId = context.getQueryId();
+    this.paths = plan.getPaths();
```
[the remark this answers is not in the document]
mychaow: ok

server/src/main/java/org/apache/iotdb/db/utils/FilePathUtils.java
```diff
@@ -46,4 +54,82 @@ public static String regularizePath(String filePath){
...
+   * @param pathIndex
+   * @return
+   */
+  public static Map<String, Long> getPathByLevel(List<Path> rawPaths, int level, Map<Integer, String> pathIndex) {
+    // pathGroupByLevel -> count
+    Map<String, Long> finalPaths = new TreeMap<>();
+
+    int i = 0;
+    for (Path value : rawPaths) {
+      String[] tmpPath = value.getFullPath().split("\\.");
```
Alima777, line 71: Hey, just like @qiaojialin said, you can reuse MetaUtils.getNodesName() method here~

server/src/main/java/org/apache/iotdb/db/qp/physical/crud/QueryPlan.java
```diff
@@ -36,6 +36,8 @@
   private int rowLimit = 0;
   private int rowOffset = 0;
 
+  private int level = -1;
```
Alima777, line 39: I think this `level` parameter shouldn't be inserted here in Queryplan which is the base class. Maybe groupByTimePlan or AggregationPlan is better.

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -549,7 +540,7 @@ private TSExecuteStatementResp internalExecuteQueryStatement(String statement,
         if (plan.getOperatorType() == OperatorType.FILL) {
           throw new QueryProcessException("Fill doesn't support disable align clause.");
         }
-        if (plan.getOperatorType() == OperatorType.GROUPBY) {
+        if (plan.getOperatorType() == GROUPBYTIME) {
```
Alima777, line 543:
```suggestion
        if (plan.getOperatorType() == OperatorType.GROUPBYTIME) {
```

mychaow:
Three places used the group by level operation: count timeseries/select count/select count group by  time interval. I don't think it's ok to create a new plan to unified them.

mychaow:
you means QueryRouter.groupby?

mychaow:
> The last one request: you'd better add some comments in GroupByTimeDataSet about what it's for. It's confusing for me at first until reading the test for it.

Do you mean the QueryRouter.groupBy? I does not change any logic of the GroupByTimeDataSet

server/src/main/java/org/apache/iotdb/db/query/dataset/groupby/GroupByTimeDataSet.java
```diff
@@ -0,0 +1,96 @@
...
+import org.apache.iotdb.tsfile.read.common.RowRecord;
+import org.apache.iotdb.tsfile.read.filter.basic.Filter;
+import org.apache.iotdb.tsfile.read.query.dataset.QueryDataSet;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.*;
+
+public class GroupByTimeDataSet extends QueryDataSet {
```
Alima777, line 42:
```suggestion
// To process statement like, select count group by time interval, level = x
public class GroupByTimeDataSet extends QueryDataSet {
```

PR #1208, opened 2020-05-14
[IOTDB-678] LOCAL_IP and SEED_NODES should be consistent in the conf file

neuyilan:
If the LOCAL_IP and SEED_NODES in the configuration file have both a local ip(127.0.0.1) and a real ip, the iotdb service cannot be connected to other nodes, causing raft communication to fail, and the cluster is always in the election state.

I think the LOCAL_IP and SEED_NODES should be consistent in the conf file; for the pseudo distributed mode, it can be all local ip, for the distributed mode, it must be the real ip, otherwise it will report an error at startup.

please leave your options, thanks.

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/ClusterMain.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/ClusterMain.java
@@ -20,6 +20,7 @@ package org.apache.iotdb.cluster;
 
 import java.io.IOException;
 import java.util.Arrays;
+import java.util.List;
 import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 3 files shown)


neuyilan:
> Great, but how about adding "localhost"? Although we can it "IP", but it accepts domain name too.
> 
> And one other thing, what if the user mix domain names with IPs? For example, by modifying `HOST`, the user maps "host1,host2,host3" to "192.168.1.1, 192.168.1.2, 192.168.1.3", however, he uses "host1" in the local IP but "192.168.168.1.1" in the seed nodes.

Good idea, I have solved this in this latest commit,  where users can specify the configuration using IP and domain names, please review again, thanks.

PR #1212, opened 2020-05-15
[IOTDB-680] Make LRUCache more accurate

JackieTien97:


server/src/main/java/org/apache/iotdb/db/engine/cache/AccountableString.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+package org.apache.iotdb.db.engine.cache;
+
+import java.util.Objects;
+import org.apache.iotdb.tsfile.common.cache.Accountable;
+
+public class AccountableString implements Accountable {
+
+  private final String string;
+  private long RAMSize;
```
qiaojialin, line 27:
```suggestion
  private long ramSize;
```

server/src/main/java/org/apache/iotdb/db/engine/cache/ChunkMetadataCache.java
```diff
@@ -47,45 +47,55 @@
...
+      protected long calEntrySize(AccountableString key, List<ChunkMetadata> value) {
         if (value.isEmpty()) {
-          return key.getBytes().length + averageSize * value.size();
+          return RamUsageEstimator.sizeOf(key) + RamUsageEstimator.shallowSizeOf(value);
         }
+        long entrySize;
         if (count < 10) {
-          long currentSize = RamUsageEstimator.shallowSizeOf(value.get(0)) + RamUsageEstimator
-              .shallowSizeOf(value.get(0).getStatistics());
+          long currentSize = CHUNK_METADATA_FIXED_RAM_SIZE + RamUsageEstimator
```
qiaojialin, line 78: This may be not enough, it also has a statistics, in which different data type may have different occupation, especially the TEXT type.

server/src/main/java/org/apache/iotdb/db/engine/cache/TimeSeriesMetadataCache.java
```diff
@@ -198,11 +206,14 @@ public void remove(TimeSeriesMetadataCacheKey key) {
...
   }
 
-  public static class TimeSeriesMetadataCacheKey {
+  public static class TimeSeriesMetadataCacheKey implements Accountable {
+
+    private final String filePath;
+    private final String device;
+    private final String measurement;
+
+    private long RAMSize;
```
qiaojialin, line 215:
```suggestion
    private long ramSize;
```

server/src/main/java/org/apache/iotdb/db/engine/cache/TimeSeriesMetadataCache.java
```diff
@@ -134,7 +142,7 @@ public TimeseriesMetadata get(TimeSeriesMetadataCacheKey key, Set<String> allSen
       timeSeriesMetadataList.forEach(timeseriesMetadata ->
           lruCache.put(new TimeSeriesMetadataCacheKey(key.filePath, key.device,
               timeseriesMetadata.getMeasurementId()), timeseriesMetadata));
-      return lruCache.get(key);
+      return new TimeseriesMetadata(lruCache.get(key));
```
qiaojialin, line 145: As we already replace the reader by resource, do we need to copy? 
But the resource may be also removed from memory, so copy is fine...

PR #1215, opened 2020-05-16
Added Blogpost from Jialin and Blog Section.

JulianFeinauer:


samperson1997:
Wow, I must say, the translation is really great! : )

JulianFeinauer:
@samperson1997 Indeed, DeepL does an awesome job and everyhting was understandable. I just did a bit of "polishing" (which did not change understanding but just make it a bit more conmfortable for readers).

PR #1216, opened 2020-05-16
[IOTDB-681] Fix bugs caused by batch manner displaying query results in CLI

samperson1997:
After entering and executing some wrong SQL statements, execute the correct statement.
At this time, the returned result is: empty set, which is unexpected.


samperson1997:
Hi @qiaojialin, I think this will also fix the problem of printing one less result : )

PR #1221, opened 2020-05-18
[IOTDB-653]Separate LAST cache from MTree

wshao08:
JIRA Issue: https://issues.apache.org/jira/browse/IOTDB-653


server/src/main/java/org/apache/iotdb/db/engine/cache/LastCacheManager.java
```diff
@@ -0,0 +1,142 @@
...
+    return SingletonClassInstance.instance;
+  }
+
+  public TimeValuePair get(String timeseriesPath) throws MetadataException {
+    String storageGroupName = MManager.getInstance().getStorageGroupName(timeseriesPath);
+    if (storageGroupMap.containsKey(storageGroupName)) {
+      return storageGroupMap.get(storageGroupName).get(timeseriesPath);
+    } else {
+      StorageGroupLastCache newEntry = new StorageGroupLastCache(storageGroupName);
+      storageGroupMap.put(storageGroupName, newEntry);
```
jt2594838, line 49: storageGroupMap is not a concurrent one, so I doubt the concurrent safety when multiple storage groups put a cache into it concurrently.

server/src/main/java/org/apache/iotdb/db/engine/cache/LastCacheManager.java
```diff
@@ -0,0 +1,142 @@
...
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import org.apache.iotdb.db.exception.metadata.MetadataException;
+import org.apache.iotdb.db.metadata.MManager;
+import org.apache.iotdb.tsfile.read.TimeValuePair;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class LastCacheManager {
+
+  private HashMap<String, StorageGroupLastCache> storageGroupMap;
```
jt2594838, line 33: It would be better to type a variable using an interface rather than specific classes to introduce more flexibility.

server/src/main/java/org/apache/iotdb/db/engine/cache/LastCacheManager.java
```diff
@@ -0,0 +1,142 @@
...
+    if (storageGroupMap.containsKey(storageGroupName)) {
+      storageGroupMap.get(storageGroupName).put(timeseriesPath, timeValuePair, highPriorityUpdate);
+    } else {
+      StorageGroupLastCache newEntry = new StorageGroupLastCache(storageGroupName);
+      newEntry.put(timeseriesPath, timeValuePair, highPriorityUpdate);
+      storageGroupMap.put(storageGroupName, newEntry);
+    }
+  }
+
+  public void clearCache(String timeseriesPath) throws MetadataException {
```
jt2594838, line 66: Add a comment that this method does not only remove the cache of the given timeseries but also its belonging storage group's caches.

server/src/main/java/org/apache/iotdb/db/query/executor/LastQueryExecutor.java
```diff
@@ -22,7 +22,10 @@
 
 import static org.apache.iotdb.db.conf.IoTDBConstant.COLUMN_VALUE;
 import static org.apache.iotdb.db.conf.IoTDBConstant.COLUMN_TIMESERIES;
+
+import java.sql.Time;
```
jt2594838, line 26: Is this used?

server/src/main/java/org/apache/iotdb/db/query/executor/LastQueryExecutor.java
```diff
@@ -179,7 +177,8 @@ public static TimeValuePair calculateLastPairForOneSeries(
     }
 
     // Update cached last value with low priority
-    node.updateCachedLast(resultPair, false, Long.MIN_VALUE);
+    LastCacheManager.getInstance().put(seriesPath.getFullPath(), resultPair, false);
+    //node.updateCachedLast(resultPair, false, Long.MIN_VALUE);
```
jt2594838, line 181: Remove the commented code.

server/src/test/java/org/apache/iotdb/db/integration/IoTDBGroupByFillIT.java
```diff
@@ -316,6 +317,7 @@ public void previousUntilLastTest1() {
     try (Connection connection = DriverManager.
             getConnection("jdbc:iotdb://127.0.0.1:6667/", "root", "root");
          Statement statement = connection.createStatement()) {
+      LastCacheManager.getInstance().clear();
```
jt2594838, line 320: I would suggest moving this into `clearEnv` in `EnvironmentUtils` so that other people will not forget to do the cleaning when they add new tests.

wshao08: Fixed

server/src/main/java/org/apache/iotdb/db/query/executor/LastQueryExecutor.java
wshao08: Fixed

server/src/main/java/org/apache/iotdb/db/query/executor/LastQueryExecutor.java
wshao08: Removed

server/src/main/java/org/apache/iotdb/db/engine/cache/LastCacheManager.java
wshao08: Fixed

server/src/main/java/org/apache/iotdb/db/engine/cache/LastCacheManager.java
wshao08: Fixed

wshao08:
As the Last implementation in cluster has been change, this PR is no longer needed.

PR #1224, opened 2020-05-18
[IoTDB-531]fix issue when grafana visualize boolean data.

jixuan1989:
This is for fixing the issue that grafana can not visualize boolean data

### DIFF
--- a/distribution/src/assembly/distribution.xml
+++ b/distribution/src/assembly/distribution.xml
@@ -34,6 +34,14 @@
             <outputDirectory>${file.separator}</outputDirectory>
             <unpack>true</unpack>
         </dependencySet>
+        <dependencySet>
+            <includes>
+                <include>*:iotdb-grafana:*</include>
+            </includes>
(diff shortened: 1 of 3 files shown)


distribution/src/assembly/distribution.xml
```diff
@@ -34,6 +34,14 @@
             <outputDirectory>${file.separator}</outputDirectory>
             <unpack>true</unpack>
         </dependencySet>
+        <dependencySet>
+            <includes>
+                <include>*:iotdb-grafana:*</include>
+            </includes>
+            <outputDirectory>tools/grafana/</outputDirectory>
+            <outputFileNameMapping>${artifact.artifactId}.${artifact.extension}</outputFileNameMapping>
+            <unpack>false</unpack>
```
JulianFeinauer, line 43: should we also include the config here? This is something I forgot in 0.9.3 branch. WDYT?

jixuan1989: needed and have corrected.

PR #1230, opened 2020-05-20
[IOTDB-665] validate physical plan before fowarding

Ring-k:
This pull request implements physical plan validation before fowarding. The client connects and sends requests to a node in the cluster, which should response some messages if the statement is invalid.

The validation is implemented in `MetaGroupMember` class, `validatePlan()` method.

```
Currently implemented:
    - SetStorageGroupPlan: 
        + path should not exist
    - InsertPlan:
        + path should exist
        + the value can be parsed into the data type of the path
```

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
@@ -1415,6 +1415,9 @@ public class MetaGroupMember extends RaftMember implements TSMetaService.AsyncIf
 
   private TSStatus validateSetStorageGroupPlan(SetStorageGroupPlan plan) {
     String path = plan.getPath().getFullPath();
+    if(path.trim().equals("root")){
+      return StatusUtils.PATH_ILLEGAL;
+    }
     if (getAllStorageGroupNames().contains(path)) {
       return StatusUtils.PATH_ALREADY_EXIST_ERROR;
     }
--- a/cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
@@ -23,6 +23,7 @@ import static org.apache.iotdb.cluster.server.RaftServer.connectionTimeoutInMS;
 import static org.apache.iotdb.cluster.server.RaftServer.queryTimeOutInSec;
 import static org.apache.iotdb.db.conf.IoTDBConstant.PATH_WILDCARD;
 import static org.apache.iotdb.db.utils.SchemaUtils.getAggregationType;
+import static org.apache.iotdb.tsfile.file.metadata.enums.TSDataType.*;
 
 import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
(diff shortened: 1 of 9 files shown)


cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -1397,6 +1401,91 @@ public TSStatus executeNonQuery(PhysicalPlan plan) {
...
+    }
+    // TODO add more pre-validations
+    return result;
+  }
+
+  private TSStatus validateSetStorageGroupPlan(SetStorageGroupPlan plan) {
+    String path = plan.getPath().getFullPath();
+    if (getAllStorageGroupNames().contains(path)) {
+      return StatusUtils.PATH_ALREADY_EXIST_ERROR;
+    }
```
jt2594838, line 1420: I think you should also check the legalness of the path. It should start with "root' but cannot be "root".

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -1397,6 +1401,91 @@ public TSStatus executeNonQuery(PhysicalPlan plan) {
...
+  private TSStatus validateInsertPlan(InsertPlan plan) {
+    List<Path> paths = (plan).getPaths();
+    String[] values = (plan).getValues();
+    List<String> pathStrings = new ArrayList<>();
+    for (Path path : paths) {
+      pathStrings.add(path.getFullPath());
+    }
+    List<MeasurementSchema> measurementSchemas;
+    try {
+      measurementSchemas = pullTimeSeriesSchemas(pathStrings);
```
jt2594838, line 1433: Check the schema cache in MManager using MManager.getSeriesType() first, if it cannot be found then add them to the list and pull them.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -1397,6 +1401,91 @@ public TSStatus executeNonQuery(PhysicalPlan plan) {
...
+      case INT32:
+        Integer.parseInt(value);
+        break;
+      case INT64:
+        Long.parseLong(value);
+        break;
+      case BOOLEAN:
+        if (!(value.equals("true") || value.equals("TRUE")
+            || value.equals("false") || value.equals("FALSE")
+            || value.equals("0") || value.equals("1"))) {
```
jt2594838, line 1473: We have defined these strings in SQLConstants, better to use them. And maybe you can use `equalsIgnoreCase` to increase robustness.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
Ring-k: That's right. Thanks for your reminding.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
Ring-k: The schema cache can help to accelerate the validations. Thanks for your suggestions.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
Ring-k: Oh, I didn't notice that. That's convenient. Thanks for your suggestion :)

PR #1245, opened 2020-05-23
Add test insertion interfaces for session and add primitiveArray pool config

wshao08:


server/src/assembly/resources/conf/iotdb-engine.properties
```diff
@@ -417,4 +417,7 @@ value_encoder=PLAIN
 compressor=SNAPPY
 
 # bloom filter error rate in TsFile
-bloom_filter_error_rate=0.05
\ No newline at end of file
+bloom_filter_error_rate=0.05
+
+# primitive array pool size in memory
```
jixuan1989, line 422: array size, not pool size

server/src/assembly/resources/conf/iotdb-engine.properties
```diff
@@ -207,6 +207,9 @@ chunk_buffer_pool_enable=false
 # data.
 # default_ttl=36000000
 
+# primitive array size (Byte) in array pool
```
qiaojialin, line 210:
```suggestion
# primitive array size (length of each array) in array pool
```

PR #1246, opened 2020-05-23
[IOTDB-696] Fix recover close file in each partition

HTHou:
When recovering, if the last file in one partition is not closed, it should not be closed. We need to add it to workingTsFileProcessor in this partition.

However, we only leave the global last file unclosed.


qiaojialin:
Sorry.. this is fixed in cluster_new branch.

PR #1252, opened 2020-05-24
delete docs num label

SailVR:


jixuan1989:
Please send an email to the mailing list to let others know about this and upload their document modifications ASAP. Otherwise once this PR is merged, they may occur a lot of conflicts. 

On the mailing list, you also need to say why you want to remove the serial number.

PR #1255, opened 2020-05-24
[IOTDB-700]Feature/iotdb 700 add openid

JulianFeinauer:
As discussed with @jixuan1989 in Slack he will have a look on this Branch and we will discuss how to integrate multiple Auth Providers in the config.


server/src/main/java/org/apache/iotdb/db/auth/authorizer/OpenIdAuthorizer.java
```diff
@@ -0,0 +1,248 @@
...
+    OpenIdAuthorizer(String providerUrl) throws AuthException, URISyntaxException, ParseException, IOException {
+        this(getJWKfromProvider(providerUrl));
+    }
+
+    private static JSONObject getJWKfromProvider(String providerUrl) throws URISyntaxException, IOException, ParseException, AuthException {
+        if (providerUrl == null) {
+            throw new IllegalArgumentException("OpenID Connect Provider URI must be given!");
+        }
+
+        //
```
xingtanzjr, line 75: Let's remove empty comment here : )

jixuan1989: welcome back @xingtanzjr

server/src/main/java/org/apache/iotdb/db/auth/authorizer/LocalFileAuthorizer.java
```diff
@@ -19,44 +19,32 @@
 package org.apache.iotdb.db.auth.authorizer;
 
 import java.io.File;
+import java.io.IOException;
+import java.net.URISyntaxException;
+
+import com.nimbusds.oauth2.sdk.ParseException;
```
jixuan1989, line 25: the unused "import" should be removed..

server/src/test/java/org/apache/iotdb/db/auth/authorizer/OpenIdAuthorizerTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+        OpenIdAuthorizer authorizer = new OpenIdAuthorizer(JSONObjectUtils.parse(OPEN_ID_PUBLIC_JWK));
+        boolean login = authorizer.login(jwt, null);
```
jixuan1989, line 28: has to consider how to write the Test. 
Either use a Mock or use a stable OpenID server.

JulianFeinauer:
@jixuan1989 I added some english users documentation and fixed the sonar issues (the last I think is fine, better log more than less). So I am open for any further feedback!

jixuan1989:
Hi @JulianFeinauer , pls pay attention to then the Travis error  (using WinOS, some resource is not released when the JWT connection fails. If we do not fix the issue, then all Travis after merging this PR will fail...)

jixuan1989:
As for the sonar issue about the log, I think following Sonar tips is fine.

As you have already thrown the exception to the caller. The handler who receives the exception should handle the exception or print the exception. 

But in this case, the exception log may be printed twice.

image

JulianFeinauer:
@jixuan1989 I fixed everything. I hope you are happy with me now? :)

jixuan1989:
Hi @JulianFeinauer , Look good to me. 
I will add the Chinese user guide later.
How about add this feature into v0.10.

PR #1256, opened 2020-05-25
delete serial num

SailVR:


HTHou:
Can you please tell us why it's necessary to delete the numbers? It will help others to understand the meaning of this PR. Thanks.

samperson1997:
> Can you please tell us why it's necessary to delete the numbers? It will help others to understand the meaning of this PR. Thanks.

Actually I think the numbers can be used to maintain the correct order of the documents in IDE ... And it would be easy for developers to index the documents : (

qiaojialin:
Hi, @HTHou @samperson1997 , the reason to remove the number is when I want to add one document between 1 and 2, I need to modify all the following documents' numbers, which is very disturbing. After removing the number, we just need to maintain the order in config.js.

PR #1262, opened 2020-05-25
[IOTDB-692] Fix merge update bug

zhanglingzhe0820:


zhanglingzhe0820:
The previous version is loop with chunkMetaHeap, but chunkMetaHeap skip sensors isn't in seq file, so the unseq sensor in test(s3) isn't writen.

PR #1264, opened 2020-05-26
[IOTDB-708] add config for inferring data type from string value 

qiaojialin:


server/src/assembly/resources/conf/iotdb-engine.properties
```diff
@@ -390,6 +390,18 @@ enable_auto_create_schema=true
 #      we will set root.sg0 as the storage group if storage group level is 1
 default_storage_group_level=1
 
+# ALL data types: BOOLEAN, INT32, INT64, FLOAT, DOUBLE, TEXT
+
+# when receive boolean string "true" or "false", we register the time series as BOOLEAN type
+boolean_string_infer_type=BOOLEAN
+
+# when receive an integer string "67", we register the time series as INT32 type
+integer_string_infer_type=FLOAT
```
jixuan1989, line 399: INT32 by default, not float

neuyilan: IMO，it's better to set FLOAT as default like InfluxDB, and I encountered a scenario where the first number the user started to transmit is generally 0, which is actually of type float. In this case, if the default configuration is INT32, the insertion will not succeed.

PR #1287, opened 2020-05-29
remove echo in cli

Ring-k:


jixuan1989:
why use Scanner to replace ConsoleReader? Any benefit?

Ring-k:
> why use Scanner to replace ConsoleReader? Any benefit?

ConsoleReader will repeat the input.

qiaojialin:
> why use Scanner to replace ConsoleReader? Any benefit?

The repeat is like this:

image

jixuan1989:
Hi, do we revert this PR? 
Now we can not use "up arrow" in the console, which is worse than print the command one more time..

PR #1289, opened 2020-05-29
Add force stop storage engine interface

qiaojialin:
The shut down is not simulated because syncCloseAllProcessors in StorageEngine.stop(), need to refactor


server/src/test/java/org/apache/iotdb/db/integration/IoTDBRecoverUnclosedIT.java
```diff
@@ -0,0 +1,267 @@
...
+        Assert.assertEquals(3, cnt);
+      }
+
+    } catch (Exception e) {
+      e.printStackTrace();
+      fail(e.getMessage());
+    }
+
+    EnvironmentUtils.shutdownDaemon();
+    EnvironmentUtils.reactiveDaemon();
```
JackieTien97, line 155:
```suggestion
    EnvironmentUtils.restartDaemon();
```

server/src/test/java/org/apache/iotdb/db/integration/IoTDBRecoverUnclosedIT.java
```diff
@@ -0,0 +1,267 @@
...
+    
+    insertMoreData();
+
+    EnvironmentUtils.shutdownDaemon();
+    EnvironmentUtils.reactiveDaemon();
```
JackieTien97, line 160:
```suggestion
    EnvironmentUtils.restartDaemon();
```

server/src/main/java/org/apache/iotdb/db/service/RegisterManager.java
```diff
@@ -64,4 +65,21 @@ public void deregisterAll() {
...
+   * stop all service and clear iService list.
+   */
+  public void shutdownAll() {
+    //we stop JMXServer at last
+    Collections.reverse(iServices);
+    for (IService service : iServices) {
+      try {
+        service.shutdown(10000);
+      } catch (Exception e) {
+        logger.error("Failed to stop {} because:", service.getID().getName(), e);
```
JackieTien97, line 79: I think you should rethrow the exception, and in the tests, you should catch it, and add failed() clause.

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -176,6 +176,7 @@ public synchronized void init() {
     try {
       tagLogFile = new TagLogFile(config.getSchemaDir(), MetadataConstant.TAG_LOG);
 
+      isRecovering = true;
```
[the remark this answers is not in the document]
HTHou: Actually, when close and restart the IoTDB using API, we don't get a new Instance of MManager. If I didn't set `isRecovering = true` here, the value of `isRecovering` would be null.

server/src/test/java/org/apache/iotdb/db/integration/IoTDBRecoverUnclosedIT.java
HTHou: Fixed, thanks~

server/src/test/java/org/apache/iotdb/db/integration/IoTDBRecoverUnclosedIT.java
HTHou: Fixed~

server/src/main/java/org/apache/iotdb/db/service/RegisterManager.java
HTHou: Very good point! Thanks.

PR #1290, opened 2020-05-29
[IOTDB-718]Fix wrong time precision of NOW()

wshao08:
JIRA issue: https://issues.apache.org/jira/browse/IOTDB-718

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/qp/strategy/LogicalGenerator.java
+++ b/server/src/main/java/org/apache/iotdb/db/qp/strategy/LogicalGenerator.java
@@ -1483,7 +1483,7 @@ public class LogicalGenerator extends SqlBaseBaseListener {
         case "ns":
           return System.currentTimeMillis() * 1000_000 + (System.nanoTime() - startupNano) % 1000_000;
         case "us":
-          return System.currentTimeMillis() * 1000 + (System.nanoTime() - startupNano) % 1000;
+          return System.currentTimeMillis() * 1000 + (System.nanoTime() - startupNano) / 1000 % 1000;
         default:
           return System.currentTimeMillis();
       }
--- a/server/src/main/java/org/apache/iotdb/db/qp/strategy/LogicalGenerator.java
+++ b/server/src/main/java/org/apache/iotdb/db/qp/strategy/LogicalGenerator.java
@@ -1475,7 +1475,15 @@ public class LogicalGenerator extends SqlBaseBaseListener {
       throw new SQLParserException("input timestamp cannot be empty");
     }
     if (timestampStr.equalsIgnoreCase(SQLConstant.NOW_FUNC)) {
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/apache/iotdb/db/qp/strategy/LogicalGenerator.java
```diff
@@ -1475,7 +1475,15 @@ long parseTimeFormat(String timestampStr) throws SQLParserException {
       throw new SQLParserException("input timestamp cannot be empty");
     }
     if (timestampStr.equalsIgnoreCase(SQLConstant.NOW_FUNC)) {
-      return System.currentTimeMillis();
+      String timePrecision = IoTDBDescriptor.getInstance().getConfig().getTimestampPrecision();
+      switch (timePrecision) {
+        case "ns":
+          return System.currentTimeMillis() * 1000_000L;
```
HTHou, line 1481: I'm just a little bit curious why don't you use `System.nanoTime()` here.

jixuan1989: When users use `now()` and set the precision as `ns` or `ms`, they want to generate dense data. if you you `System.currentTimeMillis()*1000` or `1000_000`, then many timstamps will repeat...

qiaojialin: System.nanoTime is not the current time in nano second, it is usually used for timing:

long start = System.nanoTime();
long elapse = System.nanoTime() - start;

And, there is no System. microsecond()...

Alima777: I agree with @qiaojialin and @jixuan1989 both.  System.nanoTime can't get the current time correctly, while the precision of *1000 or *1000_100 is unchanged actually. We need to rethink about the implementation way.

wshao08: Use Instant to generate nano seconds instead.

jixuan1989: I think `Instant.now()` is a good solution. 
Another way is just using the last 6 digits of System.nanoTime(), as in a JVM instance (which means if you do not restart the JVM), System.nanoTime() is keeping increasing.

Alima777: `Instant.getNano()` returns the nanoseconds, but actully it's still millisecond resolution. I'm not sure it's useful here.

jixuan1989: I think what we need is to keep the nano time as monotonic increasing.

PR #1295, opened 2020-06-01
[IOTDB-729] shutdown uncessary threadpool

qiaojialin:
Shutdown recover thread pool and upgrade threadpool to make jstack clear.


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -1537,6 +1538,7 @@ private void upgradeTsFileResourceCallBack(TsFileResource tsFileResource) {
           }
         }
       }
+      UpgradeSevice.getINSTANCE().stop();
```
HTHou, line 1541: If we need to stop UpgradeSevice after upgrading complete, I think we should also stop it if there is no file to be upgraded.

qiaojialin: fixed

PR #1298, opened 2020-06-01
[IOTDB-731] Continue write inside InsertPlan

qiaojialin:
If some measurements are failed to write, other also should pass

### DIFF
--- a/server/src/test/java/org/apache/iotdb/db/integration/IoTDBSimpleQueryIT.java
+++ b/server/src/test/java/org/apache/iotdb/db/integration/IoTDBSimpleQueryIT.java
@@ -194,8 +194,8 @@ public class IoTDBSimpleQueryIT {
       ResultSet resultSet = statement.executeQuery("select s0, s1 from root.sg1.d0");
 
       while(resultSet.next()) {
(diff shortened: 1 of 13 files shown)


server/src/test/java/org/apache/iotdb/db/integration/IoTDBSimpleQueryIT.java
```diff
@@ -169,6 +172,34 @@ public void testFirstOverlappedPageFiltered() throws SQLException, ClassNotFound
...
+        fail();
+      } catch (IoTDBSQLException e) {
+        assertTrue(e.getMessage().contains("s1"));
+      }
+
+      ResultSet resultSet = statement.executeQuery("select s0, s1 from root.sg1.d0");
+
+      while(resultSet.next()) {
+        assertEquals(resultSet.getInt("root.sg1.d0.s0"), 1);
+        assertEquals(resultSet.getString("root.sg1.d0.s1"), null);
```
samperson1997, line 198:
```suggestion
        assertEquals(1, resultSet.getInt("root.sg1.d0.s0"));
        assertEquals(null, resultSet.getString("root.sg1.d0.s1"));
```

PR #1302, opened 2020-06-01
[IOTDB-615]fix session.insertTablets api and add test methods

jixuan1989:
This PR is for fix the following issues:

1. session.insertTables() is incorrect.
2. No testInsert* methods in session and sessionPool.

### DIFF
--- a/service-rpc/src/main/thrift/rpc.thrift
+++ b/service-rpc/src/main/thrift/rpc.thrift
@@ -303,6 +303,8 @@ service TSIService {
 
 	TSExecuteBatchStatementResp testInsertTablet(1:TSInsertTabletReq req);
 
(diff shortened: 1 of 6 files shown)


service-rpc/src/main/thrift/rpc.thrift
```diff
@@ -303,6 +303,8 @@ service TSIService {
 
 	TSExecuteBatchStatementResp testInsertTablet(1:TSInsertTabletReq req);
 
+  TSExecuteBatchStatementResp testInsertTablets(1:TSInsertTabletsReq req);
```
SilverNarcissus, line 306: Hi~  You should also update the rpc-changlist.md

qiaojialin: fixed

PR #1303, opened 2020-06-01
Add Metrics Service based on Micrometer.

JulianFeinauer:
* Prometheus Endpoint is exposed based on Micrometer Prometheus.
* Add IoTDBRegistry which stores all Metrics as IoTDB timeseries under root._metrics
* Added Monitoring at several points in the programm
* Added IService Implementation for new Service

This is not final but rather a Work-In-Process where I would like to get Feedback regarding the Architecture.


jixuan1989:
some suggestions:

1. use `latency` rather than `duration`;

2. use InsertPlan to write data rather than use SQL;

3. I think it is better to keep the server module as lightweight as possible;
  - add a switch to enable/disable collect metrics (please read the 4th suggestion. If the performance impaction can be ignored, we do not need this switch);
  - add a switch to enable/disable writing metrics into IoTDB 
  - move the HTTP service and Prometheus into a new module, e.g., Prometheus-registry. Then the dependency of micrometer-registry-prometheus can be moved out of the server module. 
(1) To implement it, you may have to extract some implement in MicrometerServerService.java as an interface, and just retain the IoTDBRegistry as the default implementation; 
(2) In `prometheus-registry` module, you can add the dependency and reimplement the interface. 
(3) suppose we have a new module `prometheus-registry`, do not let the `server` module depends on it. If so, we may need using the Java reflect strategy to load the new implementation in (2).

4. better to attach a report that the performance impaction after adding this feature

micrometer-registry-iotdb/src/main/java/org/apache/iotdb/micrometer/IoTDBRegistry.java
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    @Override
+    protected <T> Gauge newGauge(Meter.Id id, T obj, ToDoubleFunction<T> valueFunction) {
+        return new DefaultGauge<>(id, obj, valueFunction);
+    }
+
+    @Override
+    protected Counter newCounter(Meter.Id id) {
+        return new CumulativeCounter(id);
```
jixuan1989, line 90: the CumulativeCounter uses an array (DoubleAdder) to records each increased value. But the counter has only an interface count() . Is that needed to save all historical increased values (Should we change to some other counter implementation)?

liangyuanpeng:
So what problem have this PR now?

jixuan1989:
Hi Julian, please see https://cwiki.apache.org/confluence/display/IOTDB/Monitor+Module
and 
https://github.com/apache/iotdb/pull/4194

SteveYurongSu:
The new metrics framework has been accepted by the community, and the implementation of this PR has been referenced.

Thanks, Julian! @JulianFeinauer

I will close this PR.

JulianFeinauer:
Thanks for the feedback @SteveYurongSu and congrats to the effort!

PR #1308, opened 2020-06-03
[IOTDB-738] Fix measurements has blank

qiaojialin:
(1) In MTree.getDeviceNodeWithAutoCreating(), the StorageGroup node will be created as an InternalMNode, which will cause the following MTree.getNodeByPathWithStorageGroupCheck() always throw StorageGroupNotExistException.

(2) If the measurement name has blank, when we generate new Path, the blank will be trim out. Then we create a path without blank. But in the following check, we will get the child with blank, then it stuck.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
@@ -888,6 +888,9 @@ public class MManager {
             MNode fullNode = getNodeByPath(parent.getFullPath() + IoTDBConstant.PATH_SEPARATOR + child);
             if (fullNode == null) {
               logger.warn("realChild is null, qilepale");
+            } else {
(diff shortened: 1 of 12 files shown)


server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -288,7 +289,11 @@ public void createTimeseries(CreateTimeSeriesPlan plan, long offset) throws Meta
         }
         storageGroupName =
             MetaUtils.getStorageGroupNameByLevel(path, config.getDefaultStorageGroupLevel());
-        setStorageGroup(storageGroupName);
+//        try {
+          setStorageGroup(storageGroupName);
+//        } catch (StorageGroupAlreadySetException e1) {
+//          ignore
+//        }
```
HTHou, line 296: Remove this

qiaojialin: fixed

PR #1311, opened 2020-06-04
[IOTDB-695] Accelerate the count timeseries query

Alima777:
For count timeseries, we get all timeseries and then calculate the count. Which occupy much memory and not efficient. It's better to count while traversing the MTree.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
@@ -701,15 +701,15 @@ public class MManager {
   }
 
   /**
-   * To calculate the count of timeseries in the given level for given prefix path.
+   * To calculate the count of nodes in the given level for given prefix path.
    *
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -692,6 +692,20 @@ public String getStorageGroupName(String path) throws MetadataException {
...
 
+  /**
+   * To calculate the count of timeseries for given prefix path.
+   */
+  public int getAllTimeseriesCount(String prefixPath) throws MetadataException {
+    lock.readLock().lock();
+    try {
+      return mtree.getAllTimeseriesCount(prefixPath);
+    } catch (MetadataException e) {
+      throw new MetadataException(e);
```
qiaojialin, line 703: no need

Alima777: Fixed.

PR #1319, opened 2020-06-05
[Distrubuted] Peer tracker optimization

LebronAl:
during matchIndex checking,start checking from follower's lastLogIndex instead of leader's lastLogIndex to accelerate checking

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/log/catchup/CatchUpTask.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/log/catchup/CatchUpTask.java
@@ -55,13 +55,13 @@ public class CatchUpTask implements Runnable {
   }
 
   /**
-   *
    * @return true if the matched index exceed the memory log bound so a snapshot is necessary, or
    * false if the catch up can be done using memory logs.
    * @throws TException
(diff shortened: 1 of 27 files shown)


cluster/src/main/java/org/apache/iotdb/cluster/log/catchup/CatchUpTask.java
```diff
@@ -55,28 +55,32 @@ public CatchUpTask(Node node, Peer peer, RaftMember raftMember) {
...
   /**
-   *
    * @return true if the matched index exceed the memory log bound so a snapshot is necessary, or
    * false if the catch up can be done using memory logs.
    * @throws TException
    * @throws InterruptedException
    */
   boolean checkMatchIndex() throws TException, InterruptedException, LeaderUnknownException {
+    boolean isLogDebug = logger.isDebugEnabled();
+    long lo = -1, hi;
```
jt2594838: I think it is fine to put multiple variables in a single line, but this will add Sonar code smells, so it is better to put them into separate lines.

LebronAl: OK~

PR #1323, opened 2020-06-07
IOTDB-750 Allow SubDevices in IoTDB.

JulianFeinauer:


server/src/main/java/org/apache/iotdb/db/metadata/mnode/LeafMNode.java
```diff
@@ -24,14 +24,9 @@
...
 
-import java.util.Collections;
 import java.util.Map;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
-public class LeafMNode extends MNode {
-
-  private static final Logger logger = LoggerFactory.getLogger(LeafMNode.class);
+public class LeafMNode extends InternalMNode {
```
qiaojialin, line 29: Call this MeasurementMNode is better.

Alima777: I agree with that. Otherwise the concept of leaf will be confusing.

JulianFeinauer: 👍

JulianFeinauer:
Thanks all reviewers so far, especially the suggestion to change LeafMNode name and to clarify intent (Add Measurements BELOW a Measurement(s), NOT add device below device), so a Node can at the same time be a device and a timeseries. I added all changes and rebased : )

PR #1328, opened 2020-06-08
[IOTDB-753] Add a judgement before removing entry from tag inverted index

JackieTien97:


samperson1997:
Hi, should this PR be merged into `rel/0.10` ?

jixuan1989:
> Hi, should this PR be merged into `rel/0.10` ?

Ok.  

By the way, pay attention to the code smell (use log.isWarnEnable() before calling log.warn() because there are too many string concat.)

PR #1332, opened 2020-06-09
[IOTDB-727] Use the same String in MManager and TsFileResource

HTHou:


HTHou:
> How about adding some tests to ensure that the device strings are the same object?

Thanks. Great idea! I'll add some tests soon.

PR #1341, opened 2020-06-10
[ISSUE-1281] Show timeseries with heat sort

JackieTien97:
The 'orderByHeatClause' is added to support this feature. The detailed syntax is:

orderByHeatClause
    : ORDER BY HEAT
    ;

It's an optional clause, if user doesn't specify it. The action of 'show timeseries' will just be like before, the result set is sorted by the registered time of the time series.

If it is specified, firstly we will obtain all the satisfying timeseries as before, then we will sort them by the last point's timestamp. That means if the timeseries is inserted recently, it will be shown in the front. 

In the implementation, as we have got the LeafMNode before, so we can use the cachedLastValuePair field in it to get last timestamp, if the cachedLastValuePair is null, we will construct a LastQueryPlan to get it.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
@@ -18,6 +18,8 @@
  */
 package org.apache.iotdb.db.metadata;
 
+import static java.util.stream.Collectors.toList;
+
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileReader;
@@ -36,7 +38,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
-import java.util.TreeSet;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import org.apache.iotdb.db.conf.IoTDBConfig;
 import org.apache.iotdb.db.conf.IoTDBConstant;
@@ -772,7 +773,9 @@ public class MManager {
       if (value2Node.isEmpty()) {
         throw new MetadataException("The key " + plan.getKey() + " is not a tag.");
       }
-      Set<LeafMNode> allMatchedNodes = new TreeSet<>(Comparator.comparing(MNode::getFullPath));
+
+      List<LeafMNode> allMatchedNodes = new ArrayList<>();
+
       if (plan.isContains()) {
         for (Entry<String, Set<LeafMNode>> entry : value2Node.entrySet()) {
           String tagValue = entry.getKey();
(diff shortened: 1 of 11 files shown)


server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -798,6 +800,18 @@ public int getNodesCountInGivenLevel(String prefixPath, int level) throws Metada
           }
         }
       }
+
+      // if ordered by heat, we sort all the timeseries by the descending order of the last insert timestamp
+      if (plan.isOrderByHeat()) {
+        allMatchedNodes = allMatchedNodes.stream().sorted(
+            Comparator.comparingLong(MTree::getLastTimeStamp).reversed()
+                .thenComparing(MNode::getFullPath)).collect(toList());
```
qiaojialin, line 808: do we need to order by path?

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
```diff
@@ -657,7 +713,8 @@ private int getCountInGivenLevel(MNode node, int targetLevel) {
   /**
    * Iterate through MTree to fetch metadata info of all leaf nodes under the given seriesPath
    *
-   * @param timeseriesSchemaList List<timeseriesSchema>
+   * @param timeseriesSchemaList Collection<timeseriesSchema> result: [name, alias, storage group,
+   *                          dataType, encoding, compression, offset, lastTimeStamp]
    */
   private void findPath(MNode node, String[] nodes, int idx, String parent,
```
qiaojialin, line 719: it's better to return List<MeasurementMNode> instead of List<String>, Zesong may be working on this.

JackieTien97: It actually returns List, I forgot to change the comments back.

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
Alima777: @qiaojialin Hi, in my opinion, order by path is better than disorder relatively.

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
```diff
@@ -624,6 +631,34 @@ private int getCountInGivenLevel(MNode node, int targetLevel) {
     return cnt;
   }
 
+  /**
+   * Get all time series schema under the given path order by insert frequency
+   *
+   * <p>result: [name, alias, storage group, dataType, encoding, compression, offset]
```
Alima777, line 637: Hi, you forgot the lastTimestamp column here~

PR #1342, opened 2020-06-10
fix some error in documents

jixuan1989:
some docs still use incorrect URLs.
some docs are confusing （Like the version info)


HTHou:
Doc of watermark Tool
``client`` should be ``CLI``

Screen Shot 2020-06-10 at 4 05 43 PM

jixuan1989:
has merged to rel/v0.10

PR #1344, opened 2020-06-10
[Distributed] change infertype

LebronAl:


server/src/main/java/org/apache/iotdb/db/qp/physical/crud/InsertPlan.java
```diff
@@ -461,9 +461,7 @@ public void deserialize(ByteBuffer buffer) {
     }
 
     // the types are not inferred before the plan is serialized
-    if (types[0] == null) {
-      this.inferType = true;
-    }
+    this.inferType = true;
```
qiaojialin, line 464: Hi, the serializedPlan should already have specific data types. I think this field should be false.

LebronAl: I found I can't insert data to cluster_new just an hour ago.And I found the changes in this [commit](https://github.com/apache/incubator-iotdb/commit/34ea36f50fcee41f8f18a01a24bfee8b040f9a52) resulting this bug.Now @jt2594838  has fix this bug by serializing the inferFlag in this [commit](https://github.com/apache/incubator-iotdb/commit/cad7680e67dccfc8de1de85fe6f9e70539f10641).I will close this PR.
Thanks for reviewing!

PR #1345, opened 2020-06-11
[IOTDB-759] Refactor MNode by removing InternalMNode

samperson1997:
* refactor `MNode` by removing `InternalMNode` and including all its methods.
* [Experimental] Add serialization method for `MNode`


tsfile/src/main/java/org/apache/iotdb/tsfile/utils/ReadWriteIOUtils.java
```diff
@@ -105,15 +105,15 @@ public static byte readByte(ByteBuffer buffer) {
   /**
    * write if the object not equals null. Eg, object eauals null, then write false.
    */
```
HTHou, line 107:
```suggestion
  /**
   * write if the object equals null. Eg, object equals null, then write true.
   */
```

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/ReadWriteIOUtils.java
```diff
@@ -105,15 +105,15 @@ public static byte readByte(ByteBuffer buffer) {
...
-  public static int writeIsNotNull(Object object, OutputStream outputStream) throws IOException {
-    return write(object != null, outputStream);
+  public static int writeIsNull(Object object, OutputStream outputStream) throws IOException {
+    return write(object == null, outputStream);
   }
 
   /**
    * write if the object not equals null. Eg, object eauals null, then write false.
    */
```
HTHou, line 114:
```suggestion
  /**
   * write if the object equals null. Eg, object equals null, then write true.
   */
```

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/ReadWriteIOUtils.java
samperson1997: Thanks for the good catch !!!

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/ReadWriteIOUtils.java
samperson1997: Fixed

server/src/main/java/org/apache/iotdb/db/metadata/mnode/MNode.java
```diff
@@ -44,49 +57,100 @@
...
+  public void deleteChild(String name) throws DeleteFailedException {
+    if (children.containsKey(name)) {
+      Lock writeLock;
+      // if its child node is leaf node, we need to acquire the write lock of the current device node
+      if (children.get(name) instanceof MeasurementMNode) {
+        writeLock = lock.writeLock();
+      } else {
+        // otherwise, we only need to acquire the write lock of its child node.
+        writeLock = (children.get(name)).lock.writeLock();
+      }
```
Alima777, line 101: I'm not sure about this. Whether should we change the lock logic of this method? As the measurementNode maybe internalNode now.

samperson1997: Good point, thanks! I've fixed it.

server/src/main/java/org/apache/iotdb/db/metadata/mnode/MNode.java
```diff
@@ -44,49 +53,96 @@
...
+
+  protected transient ReadWriteLock lock = new ReentrantReadWriteLock();
 
   /**
    * Constructor of MNode.
    */
   public MNode(MNode parent, String name) {
     this.parent = parent;
     this.name = name;
+    this.children = new LinkedHashMap<>();
```
qiaojialin, line 67: init when used

samperson1997: Fixed

server/src/main/java/org/apache/iotdb/db/metadata/mnode/MNode.java
```diff
@@ -56,37 +69,85 @@ public MNode(MNode parent, String name) {
...
+  }
 
   /**
    * get the child with the name
    */
-  public abstract MNode getChild(String name);
+  public MNode getChild(String name) {
+    if (children != null && children.containsKey(name)) {
+      return children.get(name);
+    }
```
qiaojialin, line 124: MNode child = null;
if (children != null) {
    child = children.get(name);
}
if (child != null) {
    return child;
}

samperson1997: Thanks!

server/src/main/java/org/apache/iotdb/db/metadata/mnode/MNode.java
```diff
@@ -127,4 +201,43 @@ public String getName() {
...
+
+  public void setChildren(Map<String, MNode> children) {
+    this.children = children;
+  }
+
+  public void serializeTo(BufferedWriter bw) throws IOException {
+    serializeChildren(bw);
+
+    String s = String.valueOf(MetadataConstant.MNODE_TYPE);
+    s += "," + name + ",";
```
jixuan1989, line 213: ah..... String.format or StringBuilder?

samperson1997: Actually it's the IDE that recommends me to use `String` instead of `StringBuilder`...
image

samperson1997: I'll try `Striing.format`

jixuan1989: And another serialization method has the same issue.
My another question is,... seems these serialize() methods are not used...

samperson1997: Fixed. These methods will be used in my next PR for [issue IOTDB-726](https://issues.apache.org/jira/browse/IOTDB-726)

samperson1997: Emm...  As is mentioned in [this page](https://stackoverflow.com/questions/44117788/performance-between-string-format-and-stringbuilder) which compares the performance between String.format and StringBuilder thanks to @Alima777 , I'll use StringBuilder anyway.

PR #1359, opened 2020-06-12
cherry-pick Use the same String in MManager and TsFileResource

jixuan1989:
cherry-pick #1332 to v0.10.


jixuan1989:
I will close it as according to the discussion on the mailing list, it will be released in v.11

PR #1365, opened 2020-06-15
[IOTDB-761]cherry-picked from v0.10: flink-tsfile-connecto and flink-example dependency conflict: "Some Enforcer rules have failed. Look above for specific messages explaining why the rule failed"

jixuan1989:
There are two dependency conflicts that cause the maven failure:

Flink 1.10 uses commons-compress 1.18 while hadoop-tsfile (Hadoop 2.7.3) module uses 1.4.1;
Flink 1.10 uses commons-math3 3.1.1 while hadoop-tsfile (Hadoop 2.7.3) module uses 3.5
So, when running `mvn release:parepare`, exception message will be thrown like:

Some Enforcer rules have failed. Look above for specific messages explaining why the rule failed.

The solution is to exclude the stale version of those dependencies.

this is

### DIFF
--- a/example/flink/pom.xml
+++ b/example/flink/pom.xml
@@ -35,6 +35,14 @@
             <groupId>org.apache.iotdb</groupId>
             <artifactId>flink-iotdb-connector</artifactId>
             <version>${project.version}</version>
+            <exclusions>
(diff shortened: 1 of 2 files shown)


jixuan1989:
@vesense and @sunjincheng121 , could you have a review?
(By the way, does Flink has plan to upgrade its dependencies?)

PR #1370, opened 2020-06-16
[IOTDB-769]Fix precision lost when using PLAIN for FLOAT/DOUBLE

SailVR:


server/src/test/java/org/apache/iotdb/db/utils/datastructure/TVListTest.java
```diff
@@ -26,8 +26,18 @@
...
+  @Test
+  public void testDoubleTVList1(){
+    TVList tvList = new DoubleTVList();
+    for (long i = 0; i < 1000; i++) {
+      double j = 1.2345678+Math.random();
+      tvList.putDouble(i, j);
+    }
+    tvList.sort();
+    tvList.getTimeValuePair(0);
+  }
```
HTHou, line 40: LGTM. Why not add a `testFloatTVList` as well?

SailVR: OK，add

server/src/test/java/org/apache/iotdb/db/utils/datastructure/TVListTest.java
```diff
@@ -26,8 +26,29 @@
...
 
-public class LongTVListTest {
+public class TVListTest {
 
+  @Test
+  public void testFloatTVList1(){
+    TVList tvList = new FloatTVList();
+    for (long i = 0; i < 1000; i++) {
+      double j = 1.2345678+Math.random();
+      tvList.putDouble(i, j);
```
HTHou, line 36: Still double here?

PR #1373, opened 2020-06-16
[IOTDB-771] Add audit log

qiaojialin:
Add audit log for connection and oprations.

Log file name: log_audit.log


server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -1061,6 +1061,10 @@ public ServerProperties getProperties() {
   @Override
   public TSExecuteBatchStatementResp insertRecords(TSInsertRecordsReq req) {
     TSExecuteBatchStatementResp resp = new TSExecuteBatchStatementResp();
+    auditLogger
```
jixuan1989, line 1064: better to add ifDebugEnabled

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -1337,6 +1345,8 @@ public TSExecuteBatchStatementResp createMultiTimeseries(TSCreateMultiTimeseries
       logger.info(INFO_NOT_LOGIN, IoTDBConstant.GLOBAL_DB_NAME);
       return RpcUtils.getTSBatchExecuteStatementResp(TSStatusCode.NOT_LOGIN_ERROR);
     }
+    auditLogger.info("Session-{} create multi timeseries, first is {}", currSessionId.get(),
```
jixuan1989, line 1348: better to print how many time series

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
qiaojialin: fixed

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
qiaojialin: added

PR #1374, opened 2020-06-16
Fix session doc

SilverNarcissus:
Fix session doc


jixuan1989:
Is the Chinese doc consistent?

PR #1377, opened 2020-06-16
[IOTDB-772] The start-server.sh uses tool long java command in WinOS

jixuan1989:
on WinOS, we can use "lib\*" to refer all jar files in the lib folder...


jixuan1989:
cherry-picked to rel/0.10

PR #1381, opened 2020-06-17
[IOTDB-773] add tag attribute size in config check

qiaojialin:


samperson1997:
Please help add `max_degree_of_index_node` as an immutable property : )

PR #1387, opened 2020-06-18
[IOTDB-709] Create schema automatically

Ring-k:
This pull request implements schema automatic creation in distributed version.

If an InsertPlan is executed but an exception is encountered because the storage group is not set yet, or the timeseries is not created, the MetaGroupMember will first execute a SetStorageGroupPlan, or CreateTimeseriesPlan, and then execute the InsertPlan again. 

These plans will be routed among the cluster, so that the corresponding nodes will be notified.


cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -1520,12 +1534,31 @@ private TSStatus processPartitionedPlan(PhysicalPlan plan) throws UnsupportedPla
...
+          return executeNonQuery(plan);
+        } catch (MetadataException e) {
+          logger.info("Failed to set storage group of device id {}", deviceId);
+        }
+      }
+      logger.error("{}: Cannot found storage groups for {}", name, plan);
       return StatusUtils.NO_STORAGE_GROUP;
     }
-    logger.debug("{}: The data groups of {} are {}", name, plan, planGroupMap);
+    logger.error("{}: The data groups of {} are {}", name, plan, planGroupMap);
```
jt2594838, line 1561: This should still be debug.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -1551,6 +1584,20 @@ TSStatus forwardPlan(Map<PhysicalPlan, PartitionGroup> planGroupMap) {
         subStatus = forwardPlan(entry.getKey(), entry.getValue());
       }
       if (subStatus.getCode() != TSStatusCode.SUCCESS_STATUS.getStatusCode()) {
+        if (entry.getKey() instanceof InsertPlan
+            && subStatus.getCode() == TSStatusCode.STORAGE_ENGINE_ERROR.getStatusCode()
```
jt2594838, line 1588: I think we should add a specific status code like "TSStatusCode.NO_TIMESERIES", as this one is too abstract and we are not sure whether it is because of no metadata.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -1551,6 +1584,20 @@ TSStatus forwardPlan(Map<PhysicalPlan, PartitionGroup> planGroupMap) {
...
+            && IoTDBDescriptor.getInstance().getConfig().isAutoCreateSchemaEnabled()) {
+          // try to create timeseries
+          boolean hasCreate = autoCreateTimeseries((InsertPlan) entry.getKey(), entry.getValue());
```
jt2594838, line 1591: "entry.getValue()" is the data group you are going to send the plan to, and it may not be the group that should hold the metadata, so create timeseries plans should be forward to the group that should hold the metadata.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -1569,6 +1616,70 @@ TSStatus forwardPlan(Map<PhysicalPlan, PartitionGroup> planGroupMap) {
...
+        DataClient client = getDataClient(node);
+        Map<String, Boolean> result = SyncClientAdaptor
+            .getUnregisteredMeasurements(client, partitionGroup.getHeader(), seriesList);
+        for (Map.Entry<String, Boolean> entry : result.entrySet()) {
+          if (!entry.getValue()) {
+            unregistered.add(entry.getKey());
+          }
+        }
+      } catch (TException | IOException e) {
+        logger.error("{}: cannot getting unregistered series list {} from {}", name,
```
jt2594838, line 1672: getting -> get
I would suggest print something like `"{} and other {} paths", seriesList.get(0), seriesList.size() - 1` instead of `"{}", serieseList`, in case that the list is too long.
And if you want to print a list in a log, you can just use the list as a parameter and there is no need to wrap it with "Arrays.toString(seriesList.toArray(new String[0]))".

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -1569,6 +1616,70 @@ TSStatus forwardPlan(Map<PhysicalPlan, PartitionGroup> planGroupMap) {
...
+    List<String> unregisteredSeriesList = getUnregisteredSeriesList(seriesList, partitionGroup);
+    for (String seriesPath : unregisteredSeriesList) {
+      int index = seriesList.indexOf(seriesPath);
+      TSDataType dataType = TypeInferenceUtils
+          .getPredictedDataType(insertPlan.getValues()[index], true);
+      TSEncoding encoding = getDefaultEncoding(dataType);
+      CompressionType compressionType = TSFileDescriptor.getInstance().getConfig().getCompressor();
+      CreateTimeSeriesPlan createTimeSeriesPlan = new CreateTimeSeriesPlan(new Path(seriesPath),
+          dataType, encoding, compressionType, null, null, null, null);
+      TSStatus result = executeNonQuery(createTimeSeriesPlan);
```
jt2594838, line 1642: Please add a TODO here: "TODO-Cluster: add executeNonQueryBatch()", so we can cut the number of communications later.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -1569,6 +1616,70 @@ TSStatus forwardPlan(Map<PhysicalPlan, PartitionGroup> planGroupMap) {
...
+      if (result.getCode() != TSStatusCode.SUCCESS_STATUS.getStatusCode()) {
+        logger.error("{} failed to execute create timeseries {}", thisNode, seriesPath);
+        return false;
```
jt2594838, line 1645: The schema may be auto-created concurrently, so I think the status code should be carefully checked, and if it says that the timeseries are already created, we should also return true.

service-rpc/src/main/thrift/cluster.thrift
```diff
@@ -350,6 +350,8 @@ service TSDataService extends RaftService {
 
   list<binary> getAggrResult(1:GetAggrResultRequest request)
 
+  map<string, bool> isMeasurementsRegistered(1: Node header, 2: list<string> measurements)
```
jt2594838, line 353: A suggestion about the interface design: I think the return type can be a list<bool> with the length of `measurements` because the sender clearly know what series it queries and the receiver need not send them back.

Another thing, I think "measurement" is not proper here, better just call them timeseries.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -1520,12 +1534,31 @@ private TSStatus processPartitionedPlan(PhysicalPlan plan) throws UnsupportedPla
     } catch (MetadataException e) {
       logger.error("Cannot route plan {}", plan, e);
     }
-    // the storage group is not found locally, forward it to the leader
+    // the storage group is not found locally
     if (planGroupMap == null || planGroupMap.isEmpty()) {
-      logger.debug("{}: Cannot found storage groups for {}", name, plan);
+      if (plan instanceof InsertPlan && IoTDBDescriptor.getInstance().getConfig()
```
neuyilan, line 1539: How about CreateTimeSeriesPlan? and other plans which can cause first check storage group exist or not ?

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
Ring-k: Thanks for your reminding. I will add the comment.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
Ring-k: Alright, I will revert.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
Ring-k: Thanks for your suggestion. I will change that.

service-rpc/src/main/thrift/cluster.thrift
Ring-k: I will change the interface, specifically, the return, to be a list f string, `List<String>,` which are  unregistered series.

Ring-k:
> Good job. But I have a question: if you enable AutoCreation, the receiver of the Insertion plan will create schemas locally and the sender will not sense a schema missing. How do you resolve it?

A new property `ENABLE_AUTO_CREATE_SCHEMA` is added in cluster configurations, which will replace the one in standalone version. In other word, the `ENABLE_AUTO_CREATE_SCHEMA` in iotdb-engine.properties is alwayes treated as `FALSE`. Therefore, the auto schema creation on storage engine level will never be triggered.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
Ring-k: The problem you mentioned refers to `CreateTimeseriesPlan`, `InsertTimeseriesPlan`, `DeleteTimeseriesPlan` and `DeleteTimeseriesPlan`. However, if storage group is not set when executing `DeleteTimeseriesPlan` and `DeleteTimeseriesPlan`, logically there's no necessity for the storage group to be created. As in the method `executeNonQuery` in `MetaGroupMember`, these two plans will trigger `processNonPartitionDataPlan` and `processNonPartitionMetaPlan` respectively. `CreateTimeseriesPlan` and `InsertTimeseriesPlan` are "partitioned plans". As a result, if storage group is not set for these 2 types, auto creation will be performed.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -1667,6 +1710,18 @@ TSStatus forwardPlan(List<PartitionGroup> partitionGroups, PhysicalPlan plan) {
         status = forwardPlan(plan, partitionGroup);
       }
       if (status.getCode() != TSStatusCode.SUCCESS_STATUS.getStatusCode()) {
+        if (plan instanceof InsertPlan
```
LebronAl, line 1713: It seems that this function `forwardPlan` will only be used in `processNonPartitionedDataPlan`, which will be called only when the type of plan is `DeletePlan` or `DeleteTimeSeriesPlan`, so maybe these new code in this function is useless?

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -1530,9 +1544,38 @@ private TSStatus processPartitionedPlan(PhysicalPlan plan) throws UnsupportedPla
...
+        // try to set storage group
+        String deviceId = ((InsertPlan) plan).getDeviceId();
+        try {
+          String storageGroupName = MetaUtils
+              .getStorageGroupNameByLevel(deviceId, IoTDBDescriptor.getInstance()
+                  .getConfig().getDefaultStorageGroupLevel());
+          SetStorageGroupPlan setStorageGroupPlan = new SetStorageGroupPlan(
+              new Path(storageGroupName));
+          TSStatus setStorageGroupResult = executeNonQuery(setStorageGroupPlan);
+          if (setStorageGroupResult.getCode() != TSStatusCode.SUCCESS_STATUS.getStatusCode()) {
```
LebronAl, line 1560: maybe you can use `processNonPartitionedMetaPlan` directly to avoid unnecessary judgement

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -1530,9 +1544,38 @@ private TSStatus processPartitionedPlan(PhysicalPlan plan) throws UnsupportedPla
...
+          .isEnableAutoCreateSchema()) {
+        // try to set storage group
+        String deviceId = ((InsertPlan) plan).getDeviceId();
+        try {
+          String storageGroupName = MetaUtils
+              .getStorageGroupNameByLevel(deviceId, IoTDBDescriptor.getInstance()
+                  .getConfig().getDefaultStorageGroupLevel());
+          SetStorageGroupPlan setStorageGroupPlan = new SetStorageGroupPlan(
+              new Path(storageGroupName));
+          TSStatus setStorageGroupResult = executeNonQuery(setStorageGroupPlan);
```
LebronAl, line 1559: maybe you can use `processNonPartitionedMetaPlan` directly to avoid unnecessary judgement

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
LebronAl: maybe you can use `processPartitionedPlan` directly to avoid unnecessary judgement

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -1684,6 +1739,77 @@ TSStatus forwardPlan(List<PartitionGroup> partitionGroups, PhysicalPlan plan) {
...
+    for (String seriesPath : unregisteredSeriesList) {
+      int index = seriesList.indexOf(seriesPath);
+      TSDataType dataType = TypeInferenceUtils
+          .getPredictedDataType(insertPlan.getValues()[index], true);
+      TSEncoding encoding = getDefaultEncoding(dataType);
+      CompressionType compressionType = TSFileDescriptor.getInstance().getConfig().getCompressor();
+      CreateTimeSeriesPlan createTimeSeriesPlan = new CreateTimeSeriesPlan(new Path(seriesPath),
+          dataType, encoding, compressionType, null, null, null, null);
+      // TODO-Cluster: add executeNonQueryBatch()
+      TSStatus result = executeNonQuery(createTimeSeriesPlan);
```
LebronAl, line 1776: maybe you can use `processPartitionedPlan` directly to avoid unnecessary judgement

cluster/src/main/java/org/apache/iotdb/cluster/server/member/DataGroupMember.java
```diff
@@ -1073,7 +1072,12 @@ public void pullTimeSeriesSchema(PullSchemaRequest request,
 
   /**
    * Create an IPointReader of "path" with “timeFilter” and "valueFilter". A synchronization with
+<<<<<<< HEAD
+   * the leader will be performed first to preserve strong consistency. TODO-Cluster: also support
+   * weak consistency
+=======
    * the leader will be performed according to consistency level
+>>>>>>> origin/cluster_new
```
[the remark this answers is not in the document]
Ring-k: Thanks for your reminding.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/DataGroupMember.java
```diff
@@ -1610,6 +1623,34 @@ public void getAggrResult(GetAggrResultRequest request,
...
+   * Check if the given measurements are registered or not
+   * @param header
+   * @param timeseriesList
+   * @param resultHandler
+   * @throws TException
+   */
+  @Override
+  public void getUnregisteredTimeseries(Node header, List<String> timeseriesList,
+      AsyncMethodCallback<List<String>> resultHandler) throws TException {
+    if (!syncLeader()) {
```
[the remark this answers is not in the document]
Ring-k: Thank you for your suggestion. The method `syncLeaderWithConsistencyCheck` will be used.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/DataGroupMember.java
```diff
@@ -1610,6 +1623,34 @@ public void getAggrResult(GetAggrResultRequest request,
...
+      resultHandler.onError(new LeaderUnknownException(getAllNodes()));
+      return;
+    }
+    List<String> result = new ArrayList<>();
+    for (String seriesPath : timeseriesList) {
+      try {
+        List<String> path = MManager.getInstance().getAllTimeseriesName(seriesPath);
+        if (path.size() != 1) {
+          throw new MetadataException("Size of the path is not 1.");
```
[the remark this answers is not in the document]
Ring-k: Thanks for your suggestion. The information of the series will be included.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -1530,9 +1544,29 @@ private TSStatus processPartitionedPlan(PhysicalPlan plan) throws UnsupportedPla
...
+        String deviceId = ((InsertPlan) plan).getDeviceId();
+        try {
+          String storageGroupName = MetaUtils
+              .getStorageGroupNameByLevel(deviceId, IoTDBDescriptor.getInstance()
+                  .getConfig().getDefaultStorageGroupLevel());
+          SetStorageGroupPlan setStorageGroupPlan = new SetStorageGroupPlan(
+              new Path(storageGroupName));
+          TSStatus setStorageGroupResult = executeNonQuery(setStorageGroupPlan);
+          if (setStorageGroupResult.getCode() != TSStatusCode.SUCCESS_STATUS.getStatusCode()) {
+            throw new MetadataException("Failed to set storage group " + storageGroupName);
```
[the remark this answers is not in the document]
Ring-k: Thanks for your suggestion. The status code will be included in the exception.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
Ring-k: You are right. Thanks for your reminding. These codes will be moved to `TSStatus forwardPlan(Map<PhysicalPlan, PartitionGroup> planGroupMap, PhysicalPlan plan)`.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
Ring-k: Good idea. Thanks for your suggestion.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
Ring-k: Thanks for your suggestion. The error code `TIMESERIES_NOT_EXIST` will be introduced.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
Ring-k: Thanks for your suggestion. An extra condition will be added in the `if` clause to make sure of it.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -1530,9 +1544,39 @@ private TSStatus processPartitionedPlan(PhysicalPlan plan) throws UnsupportedPla
...
+          // try to create timeseries
+          boolean isAutoCreateTimeseriesSuccess = autoCreateTimeseries((InsertPlan)plan);
+          if(!isAutoCreateTimeseriesSuccess){
+            throw new MetadataException(
+                String.format("Failed to create timeseries from InsertPlan automatically.")
+            );
+          }
+          return executeNonQuery(plan);
+        } catch (MetadataException e) {
+          logger.error("Failed to set storage group or create timeseries, because {}", e.getMessage());
```
jt2594838, line 1576: It would be better to use `e` (preserve stack traces).

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -1684,6 +1743,77 @@ TSStatus forwardPlan(List<PartitionGroup> partitionGroups, PhysicalPlan plan) {
...
+        List<String> result = SyncClientAdaptor
+            .getUnregisteredMeasurements(client, partitionGroup.getHeader(), seriesList);
+        unregistered.addAll(result);
+      } catch (TException | IOException e) {
+        logger.error("{}: cannot getting unregistered {} and other {} paths from {}", name,
+            seriesList.get(0), seriesList.get(seriesList.size() - 1), node, e);
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        logger.error("{}: getting unregistered series list {} is interrupted from {}", name,
+            Arrays.toString(seriesList.toArray(new String[0])), node, e);
```
jt2594838, line 1811: Please refactor this message like the former one to avoid printing too many seriesPaths.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -1684,6 +1743,77 @@ TSStatus forwardPlan(List<PartitionGroup> partitionGroups, PhysicalPlan plan) {
...
+   * @return
+   */
+  List<String> getUnregisteredSeriesList(List<String> seriesList, PartitionGroup partitionGroup) {
+    Set<String> unregistered = new HashSet<>();
+    for (Node node : partitionGroup) {
+      try {
+        DataClient client = getDataClient(node);
+        List<String> result = SyncClientAdaptor
+            .getUnregisteredMeasurements(client, partitionGroup.getHeader(), seriesList);
+        unregistered.addAll(result);
```
jt2594838, line 1804: I think you should break once a non-null result is returned, and I do not think using a Set is necessary.

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
```diff
@@ -914,8 +914,26 @@ public void insert(InsertPlan insertPlan) throws QueryProcessException {
...
+          while(curException.getCause() != null){
+            curException = (Exception) curException.getCause();
+          }
+          if(!(curException instanceof PathNotExistException)){
+            isPathNotExistException = false;
+            break;
+          }
+        }
+        if(isPathNotExistException){
+          throw new PathNotExistException(failedPaths.toString());
```
jt2594838, line 932: I suggest you just add a field of List<String> in PathNotExistException, since concatenating a long string list is time-consuming.

Besides, please perform a reformat.

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
```diff
@@ -914,8 +914,26 @@ public void insert(InsertPlan insertPlan) throws QueryProcessException {
...
-        throw new StorageEngineException(
-            "failed to insert points " + insertPlan.getFailedMeasurements());
+        // check if all path not exist exceptions
+        List<String> failedPaths = new ArrayList<>(insertPlan.getFailedMeasurements().keySet());
+        List<Exception> exceptions = new ArrayList<>(insertPlan.getFailedMeasurements().values());
+        boolean isPathNotExistException = true;
+        for(Exception e : exceptions){
+          Exception curException = e;
+          while(curException.getCause() != null){
+            curException = (Exception) curException.getCause();
```
jt2594838, line 924: You may just define this as a Throwable so you may avoid the typecast.

Ring-k: Thanks for yoru suggestion.

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
Ring-k: An attribute is added in the PathNotExistExeption. Thanks for your suggestion.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
Ring-k: You are right. Finishing the loop is not necessary.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
Ring-k: Thanks for your reminding.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
Ring-k: Thanks for your suggestion.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
Ring-k: You are right. Thanks.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -1586,7 +1586,7 @@ private TSStatus processPartitionedPlan(PhysicalPlan plan) throws UnsupportedPla
           }
           return executeNonQuery(plan);
         } catch (MetadataException e) {
-          logger.error("Failed to set storage group or create timeseries, because {}", e);
+          logger.error(String.format("Failed to set storage group or create timeseries, because %s", e));
```
jt2594838, line 1589: Why change this?

Ring-k: It was recognized as a bug according to rules of sonar, "Fewer arguments provided (0) than placeholders specified (1) ".

PR #1389, opened 2020-06-18
Fix spark tsfile master

SilverNarcissus:
Only search those measurements in required schema


SilverNarcissus:
> Hi, bug fixing for version 0.10 needs to be merged to `rel/0.10`, and I'm afraid that you have to open another PR for `master`

Sure~   I have 3 PR for 0.9.3, 0.10.0 and master

PR #1392, opened 2020-06-20
[IOTDB-777]Missing dot between deviceId and sensorId in chunkMetadaCache

jack870131:
Issue description: https://issues.apache.org/jira/browse/IOTDB-777

In ChunkMetadataCache, key of this map LRULinkedHashMap<AccountableString, List<ChunkMetadata>> lruCache should be "file path dot deviceId dot sensorId", which the key of chunkMetadataCache was missing the separator between deviceId and sensorId.


Alima777:
Hi, you have to modify the write and read process of this cache together. What you modify is just the read process, in this way, you can not get any ChunkMetaData from this cache.

PR #1399, opened 2020-06-21
[IOTDB-736] Query performance tracing

Alima777:
This PR provides the performance tracing function by **adding a performance log** and gives a switch statement "**Tracing on/off**" to control whether to print this log, which is off by default.

The log records including:
1. the query sql, and the start time of this query
3. the number of series paths
4. the number of distinct tsfiles, including sequence and unsequence
5. the number of chunks, and average size of each chunk.

This pr concentrates on raw data query, it may modify something for aggregation query and group by query later.


JulianFeinauer:
Good work but I would suggest to write the log file directly not via logger. Because a user could disable it by accident although the property is active.

Alima777:
> Good work but I would suggest to write the log file directly not via logger. Because a user could disable it by accident although the property is active.

Thank you for your good suggestion. I've fixed it.

HTHou:
Great work!
Just one suggestion. It would be better to add some doc of this new feature. :)

Alima777:
> Great work!
> Just one suggestion. It would be better to add some doc of this new feature. :)

Thank you~ I will add it later. :D

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -524,10 +546,30 @@ public TSExecuteStatementResp executeQueryStatement(TSExecuteStatementReq req) {
    *             AuthorPlan
    */
   private TSExecuteStatementResp internalExecuteQueryStatement(String statement,
-      long statementId, PhysicalPlan plan, int fetchSize, String username) {
+      long statementId, PhysicalPlan plan, int fetchSize, String username) throws IOException {
     auditLogger.info("Session {} execute Query: {}", currSessionId.get(), statement);
     long startTime = System.currentTimeMillis();
     long queryId = -1;
+    if (plan instanceof QueryPlan && config.isEnablePerformanceTracing()) {
+      File performanceDir = SystemFileFactory.INSTANCE.getFile(config.getPerformanceDir());
```
qiaojialin, line 554: put all these performance related to a class TracingManager, which manage a writer, log file.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -1233,6 +1237,20 @@ public QueryDataSource query(String deviceId, String measurementId, QueryContext
       if (filePathsManager != null) {
         filePathsManager.addUsedFilesForQuery(context.getQueryId(), dataSource);
       }
+
+      // exclude repetitive tsFile to calculate the number of it
+      if (config.isEnablePerformanceTracing()) {
+        if (seqFile == null) {
+          seqFile = new HashSet<>(seqResources);
```
qiaojialin, line 1244: If I query multiple times, when do you clear this?

server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
```diff
@@ -185,6 +184,11 @@
    */
   private String syncDir = "data" + File.separator + "system" + File.separator + "sync";
 
+  /**
+   * Performance tracing directory, stores performance tracing files
+   */
+  private String performanceDir = "data" + File.separator + "system" + File.separator + "performance";
```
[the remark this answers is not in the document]
Alima777: Fixed.

server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfigCheck.java
```diff
@@ -125,24 +138,23 @@ private IoTDBConfigCheck() {
...
-   *
+   * <p>
    * When init: create system.properties directly
-   *
-   * When upgrading the system.properties:
-   * (1) create system.properties.tmp
-   * (2) delete system.properties
-   * (2) rename system.properties.tmp to system.properties
+   * <p>
+   * When upgrading the system.properties: (1) create system.properties.tmp (2) delete
```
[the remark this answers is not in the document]
Alima777: OK, my fault, all changes back..

server/src/main/java/org/apache/iotdb/db/query/reader/series/SeriesReader.java
```diff
@@ -251,6 +257,13 @@ private void unpackAllOverlappedTimeSeriesMetadataToCachedChunkMetadata(
       unpackOneTimeSeriesMetadata(firstTimeSeriesMetadata);
       firstTimeSeriesMetadata = null;
     }
+
+    // try to calculate the total number of chunk and time-value points in chunk
+    if (IoTDBDescriptor.getInstance().getConfig().isEnablePerformanceTracing()) {
+      totalChunkNum += cachedChunkMetadata.size();
```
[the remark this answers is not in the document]
Alima777: Good suggestion, Fixed.

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
Alima777: Fixed.

server/src/main/java/org/apache/iotdb/db/qp/strategy/LogicalGenerator.java
```diff
@@ -223,6 +226,20 @@ public void enterFlush(FlushContext ctx) {
     initializedOperator = flushOperator;
   }
 
+  @Override
+  public void enterTracingOn(TracingOnContext ctx) {
+    super.enterTracingOn(ctx);
+    IoTDBDescriptor.getInstance().getConfig().setEnablePerformanceTracing(true);
```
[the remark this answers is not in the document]
Alima777: Fixed.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
Alima777: Fixed.

server/src/main/java/org/apache/iotdb/db/query/control/QueryResourceManager.java
```diff
@@ -85,20 +109,49 @@ public void registerTempExternalSortFile(long queryId,
...
 
     SingleSeriesExpression singleSeriesExpression = new SingleSeriesExpression(selectedPath,
         filter);
-    return StorageEngine.getInstance().query(singleSeriesExpression, context, filePathsManager);
+    QueryDataSource queryDataSource = StorageEngine.getInstance()
+        .query(singleSeriesExpression, context, filePathsManager);
+    // calculate the distinct number of seq and unseq tsfiles
+    if (config.isEnablePerformanceTracing()) {
+      Set<TsFileResource> seqFileNum = seqFileNumMap.get(context.getQueryId());
+      Set<TsFileResource> unseqFileNum = unseqFileNumMap.get(context.getQueryId());
```
qiaojialin, line 122: you could use computeIfAbsent

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -548,6 +568,11 @@ private TSExecuteStatementResp internalExecuteQueryStatement(String statement,
       resp.setOperationType(plan.getOperatorType().toString());
       // generate the queryId for the operation
       queryId = generateQueryId(true);
+      if (plan instanceof QueryPlan && config.isEnablePerformanceTracing()) {
+        TracingManager.getInstance().writeQueryId(queryId);
+        TracingManager.getInstance().writeStartTime();
+        TracingManager.getInstance().writePathsNum(plan.getPaths().size());
```
qiaojialin, line 574: put it to one method, this method write one line and \n

server/src/main/java/org/apache/iotdb/db/query/control/QueryResourceManager.java
```diff
@@ -85,20 +109,49 @@ public void registerTempExternalSortFile(long queryId,
...
+    return queryDataSource;
   }
 
   /**
    * Whenever the jdbc request is closed normally or abnormally, this method must be invoked. All
    * query tokens created by this jdbc request must be cleared.
    */
   public void endQuery(long queryId) throws StorageEngineException {
+    try {
+      if (config.isEnablePerformanceTracing() && chunkNumMap.get(queryId) != null) {
```
qiaojialin, line 143: Not all query unpacks the TimeseriesMetadata, we only unpack it when needed. So the chunkNumMap may do not contain the queryId, you'd better check the seqFileNum or check one by one

server/src/main/java/org/apache/iotdb/db/query/control/TracingManager.java
```diff
@@ -0,0 +1,117 @@
...
+    return TracingManagerHelper.INSTANCE;
+  }
+
+  public void writeSeperator() throws IOException {
+    writer.write("-----------------------------");
+    writer.newLine();
+    writer.flush();
+  }
+
+  public void writeStatement(String statement) throws IOException {
```
qiaojialin, line 67: each line should have a QueryId in prefix

Alima777: All above Fixed. Thank you for detailed review ：)

server/src/main/java/org/apache/iotdb/db/conf/IoTDBConstant.java
```diff
@@ -97,6 +97,8 @@ private IoTDBConstant() {
   public static final String SCHEMA_FOLDER_NAME = "schema";
   public static final String SYNC_FOLDER_NAME = "sync";
   public static final String QUERY_FOLDER_NAME = "query";
+  public static final String TRACING_FOLDER_NAME = "tracing";
```
jixuan1989, line 100: not used

server/src/main/java/org/apache/iotdb/db/query/control/TracingManager.java
```diff
@@ -0,0 +1,135 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class TracingManager {
+
+  private static final Logger logger = LoggerFactory.getLogger(TracingManager.class);
+  private BufferedWriter writer;
+
+  public TracingManager(String dirName, String logFileName){
+    File performanceDir = SystemFileFactory.INSTANCE.getFile(dirName);
```
jixuan1989, line 38: perfromanceDir -> tracingDir

server/src/main/java/org/apache/iotdb/db/query/control/TracingManager.java
```diff
@@ -0,0 +1,135 @@
...
+  public void writeQueryInfo(long queryId, String statement, int pathsNum) throws IOException {
+    StringBuilder builder = new StringBuilder("-----------------------------\n");
+    builder.append("Query Id: ").append(queryId)
+        .append(" - Query Statement: ").append(statement)
+        .append("\nQuery Id: ").append(queryId)
+        .append(" - Start time: ")
+        .append(new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS").format(System.currentTimeMillis()))
+        .append("\nQuery Id: ").append(queryId)
+        .append(" - Number of series paths: ").append(pathsNum);
+    writer.write(builder.toString());
```
jixuan1989, line 70: `writer.write()` should be protected by `synchnorized`

server/src/main/java/org/apache/iotdb/db/query/control/TracingManager.java
```diff
@@ -0,0 +1,135 @@
...
+    writer.newLine();
+    writer.flush();
```
jixuan1989, line 72: no need to call `flush`

server/src/main/java/org/apache/iotdb/db/query/control/TracingManager.java
```diff
@@ -0,0 +1,135 @@
...
+  public void writeQueryInfo(long queryId, String statement) throws IOException {
+    StringBuilder builder = new StringBuilder("-----------------------------\n");
+    builder.append("Query Id: ").append(queryId)
+        .append(" - Query Statement: ").append(statement)
+        .append("\nQuery Id: ").append(queryId)
+        .append(" - Start time: ")
+        .append(new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS").format(System.currentTimeMillis()));
+    writer.write(builder.toString());
+    writer.newLine();
+    writer.flush();
```
jixuan1989, line 85: same issue.

server/src/main/java/org/apache/iotdb/db/query/control/TracingManager.java
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  public void writePathsNum(long queryId, int pathsNum) throws IOException {
+    writer.write(new StringBuilder("Query Id: ").append(queryId)
+        .append(" - Number of series paths: ").append(pathsNum).toString());
+    writer.newLine();
+    writer.flush();
```
jixuan1989, line 92: same issue

jixuan1989: use StringBuilder to get string first (do not need to wrap with `sync`), and wrap `writer.write` with `sync`

server/src/main/java/org/apache/iotdb/db/query/control/TracingManager.java
```diff
@@ -0,0 +1,135 @@
...
+    // to avoid the disorder info of multi query
+    // add query id as prefix of each info
+    writer.write(new StringBuilder("Query Id: ").append(queryId)
+        .append(" - Number of tsfiles: ").append(seqFileNum + unseqFileNum)
+        .append("\nQuery Id: ").append(queryId)
+        .append(" - Number of sequence files: ").append(seqFileNum)
+        .append("\nQuery Id: ").append(queryId)
+        .append(" - Number of unsequence files: ").append(unseqFileNum).toString());
+    writer.newLine();
+    writer.flush();
```
jixuan1989, line 105: same issue

server/src/main/java/org/apache/iotdb/db/query/control/TracingManager.java
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  public void writeChunksInfo(long queryId, long totalChunkNum, long totalChunkSize) throws IOException {
+    writer.write(new StringBuilder("Query Id: ").append(queryId)
+        .append(" - Number of chunks: ").append(totalChunkNum)
+        .append("\nQuery Id: ").append(queryId)
+        .append(" - Average size of chunks: ").append(totalChunkSize / totalChunkNum).toString());
+    writer.newLine();
+    writer.flush();
```
jixuan1989, line 114: same issue

server/src/main/java/org/apache/iotdb/db/query/control/QueryResourceManager.java
```diff
@@ -85,20 +110,50 @@ public void registerTempExternalSortFile(long queryId,
...
 
     SingleSeriesExpression singleSeriesExpression = new SingleSeriesExpression(selectedPath,
         filter);
-    return StorageEngine.getInstance().query(singleSeriesExpression, context, filePathsManager);
+    QueryDataSource queryDataSource = StorageEngine.getInstance()
+        .query(singleSeriesExpression, context, filePathsManager);
+    // calculate the distinct number of seq and unseq tsfiles
+    if (config.isEnablePerformanceTracing()) {
+      seqFileNumMap.computeIfAbsent(context.getQueryId(), k -> new HashSet<>())
+          .addAll(queryDataSource.getSeqResources());
```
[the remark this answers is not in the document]
Alima777: Hi, the TsFileResources used for this query in this map will be removed after printing tracing log. Then it can be cleaned by GC.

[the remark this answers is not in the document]
Alima777: You can see the Code in QueryResourceManager 136-146.

server/src/main/java/org/apache/iotdb/db/conf/IoTDBConstant.java
Alima777: It's used in IoTDBDescriptor `conf.setTracingDir(FilePathUtils.regularizePath(conf.getBaseDir() + IoTDBConstant.TRACING_FOLDER_NAME));`

server/src/main/java/org/apache/iotdb/db/query/control/TracingManager.java
Alima777: Fixed.

server/src/main/java/org/apache/iotdb/db/query/control/TracingManager.java
```diff
@@ -0,0 +1,135 @@
...
+        .append(new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS").format(System.currentTimeMillis()))
+        .toString());
+    writer.newLine();
+    writer.flush();
+  }
+
+  private static class TracingManagerHelper {
+
+    private static final TracingManager INSTANCE = new TracingManager(
+        IoTDBDescriptor.getInstance().getConfig().getTracingDir(),
```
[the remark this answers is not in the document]
Alima777: All above fixed.

server/src/main/java/org/apache/iotdb/db/query/control/QueryResourceManager.java
Alima777: Fixed.

server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
```diff
@@ -185,6 +184,11 @@
    */
   private String syncDir = "data" + File.separator + "system" + File.separator + "sync";
 
+  /**
+   * Performance tracing directory, stores performance tracing files
+   */
+  private String tracingDir = "data" + File.separator + "tracing";
```
jixuan1989, line 190: here should be "TRACING_FOLDER_NAME"..

server/src/main/java/org/apache/iotdb/db/query/control/QueryResourceManager.java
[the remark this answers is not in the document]
jixuan1989: this method just query one time series.. 
Suppose we have 2 time series to be queried, maybe the resources corresponding to the first time series can be GC released before the second one is queried.

server/src/main/java/org/apache/iotdb/db/query/control/TracingManager.java
```diff
@@ -0,0 +1,145 @@
...
+    writer.newLine();
+  }
+
+  public void writePathsNum(long queryId, int pathsNum) throws IOException {
+    StringBuilder builder = new StringBuilder("Query Id: ").append(queryId)
+        .append(" - Number of series paths: ").append(pathsNum);
+    synchronized (writer) {
+      writer.write(builder.toString());
+    }
+    writer.newLine();
```
jixuan1989, line 96: `writer.newLine()`  also needs to be wrapped...

jixuan1989: considering thread A writes "abcdefg", and thread B writes  "\n", then the result may be "abc\ndefg".

(By the way, why not add "\n" into the end of the string?)

server/src/main/java/org/apache/iotdb/db/query/control/QueryResourceManager.java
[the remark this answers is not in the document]
Alima777: But if the first time series is not kept in this map, we can not calculate the total number of distinct tsfiles in the whole query. Or we give the accessed times of tsfile instead? In this way, we only keep a accessed number for one time series.

server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
Alima777: I think you are right. But the Dir strings in this file are all defined this way....

server/src/main/java/org/apache/iotdb/db/query/control/TracingManager.java
Alima777: Actually, I researched and also did a test by myself, the process of BufferedWriter.write() self is thread safe...
For example, thread A invokes `writer.write("aaa")`, thread B invokes `writer.write("bbb")`, the result will only be "aaabbb" Or "bbbaaa". So the `synchronized` is not needed. And writer.flush() is also needed.....

Sorry, I will changed all back. Remove `synchronized` and add flush() back.

server/src/main/java/org/apache/iotdb/db/query/control/QueryResourceManager.java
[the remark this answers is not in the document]
Alima777: Hi, I replace those objects with WeakReference as you said. I believe it will make sense in the GC. Thank you very much for your good suggestion. Please check it.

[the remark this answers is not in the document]
Alima777: And it will reflect the correct performance in case the tsfileResources are GC and reloaded from disk in the future.

PR #1408, opened 2020-06-23
move getSeriesSchemas to MManager

mychaow:


server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
```diff
@@ -907,6 +907,17 @@ public void delete(Path path, long timestamp) throws QueryProcessException {
...
+    throws MetadataException {
+    return mManager.getSeriesSchemasAndLock(insertPlan.getDeviceId(), insertPlan.getMeasurements(), insertPlan);
+  }
+
+  protected MeasurementSchema[] getSeriesSchemas(InsertTabletPlan insertTabletPlan)
+    throws MetadataException {
+    return mManager.getSeriesSchemasAndLock(insertTabletPlan.getDeviceId(),
+      insertTabletPlan.getMeasurements(), insertTabletPlan);
+  }
+
```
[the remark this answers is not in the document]
mychaow: These functions is just for cluster version, cluster version will override this two function.

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -1800,4 +1807,207 @@ private void checkMTreeModified() {
...
+    deviceNode = getDeviceNodeWithAutoCreateAndReadLock(deviceId);
+    // To reduce the String number in memory, set the deviceId from MManager to insertPlan
+    if (plan instanceof InsertPlan) {
+      ((InsertPlan) plan).setDeviceId(deviceNode.getFullPath());
+    } else if (plan instanceof InsertTabletPlan) {
+      ((InsertTabletPlan) plan).setDeviceId(deviceNode.getFullPath());
+    } else {
+      throw new MetadataException(String.format(
+        "Only support insert and insertTablets, deviceId[%s], plans [%s]", deviceId, plan.getOperatorType()));
+    }
```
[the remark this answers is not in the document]
mychaow: ok

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -1800,4 +1807,207 @@ private void checkMTreeModified() {
...
+                trueType));
+            } else {
+              // mark failed measurement
+              ((InsertTabletPlan) plan).markMeasurementInsertionFailed(i);
+              continue;
+            }
+          }
+        }
+
+        if ((plan instanceof InsertPlan) && isCreated) {
```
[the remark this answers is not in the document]
mychaow: if a measurement exist, but the value is string (like jdbc insert), we need to convert the value to the true type

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
```diff
@@ -915,227 +926,19 @@ public void insert(InsertPlan insertPlan) throws QueryProcessException {
...
-      case FLOAT:
-        return conf.getDefaultFloatEncoding();
-      case DOUBLE:
-        return conf.getDefaultDoubleEncoding();
-      case TEXT:
-        return conf.getDefaultTextEncoding();
-      default:
-        throw new UnSupportedDataTypeException(
-            String.format("Data type %s is not supported.", dataType.toString()));
+      mManager.unlockInsert(insertPlan.getDeviceId());
```
jt2594838, line 934: I would suggest you separate the locking part from `getSeriesSchemasAndLock` and explicitly call it in this method, as it is always better to explicitly put `lock` and `unlock` in the same code block.

mychaow: ok

server/src/main/java/org/apache/iotdb/db/qp/physical/crud/InsertPlan.java
```diff
@@ -57,9 +58,9 @@
   private TSDataType[] types;
   private MeasurementSchema[] schemas;
 
-  // if inferType is false, use the type of values directly
-  // if inferType is true, values is String[], and infer types from them
-  private boolean inferType = false;
+  // if values is String[], isNeedInferType must be true, so we could infer types from them
```
qiaojialin, line 61:
```suggestion
  // if isNeedInferType is true, the values must be String[], so we could infer types from them
```

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -1846,4 +1836,219 @@ public void createMTreeSnapshot() {
...
+              ((InsertTabletPlan) plan).markMeasurementInsertionFailed(i);
+            } else if (plan instanceof InsertPlan) {
+              ((InsertPlan) plan).markMeasurementInsertionFailed(i);
+            }
+            continue;
+          }
+        }
+
+        // maybe need to convert value type to the true type
+        if ((plan instanceof InsertPlan)) {
```
qiaojialin, line 1913:
```suggestion
        if ((plan instanceof InsertPlan && ((InsertPlan) plan).isNeedInferType()) {
```

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -1846,4 +1836,219 @@ public void createMTreeSnapshot() {
...
+          }
+          break;
+        case BOOLEAN:
+          if (!(plan.getValues()[loc] instanceof Boolean)) {
+            plan.getValues()[loc] =
+              Boolean.parseBoolean(String.valueOf(plan.getValues()[loc]));
+          }
+          break;
+        case TEXT:
+          // need to do nothing
```
qiaojialin, line 1974: The String needs to be cast to Binary

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -1846,4 +1836,219 @@ public void createMTreeSnapshot() {
...
+          }
+        } else {
+          throw e;
+        }
+      }
+    }
+    return schemas;
+  }
+
+  private void changeValueToRealType(InsertPlan plan, int loc, TSDataType type) throws MetadataException {
```
qiaojialin, line 1939:
```suggestion
  private void changeStringValueToRealType(InsertPlan plan, int loc, TSDataType type) throws MetadataException {
```

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -1846,4 +1836,219 @@ public void createMTreeSnapshot() {
...
+    plan.getTypes()[loc] = type;
+    try {
+      switch (type) {
+        case INT32:
+          if (!(plan.getValues()[loc] instanceof Integer)) {
```
qiaojialin, line 1944: Do we need this check? I think only String value need to be inferred.

mychaow: yes, could remove this check.

PR #1409, opened 2020-06-23
Support partition removal

jt2594838:
Partitions can be removed with the following statement:
`DELETE PARTITION root.sg1 0,1,2,3,4` which deletes the first five partitions of storage group root.sg1.


server/src/main/java/org/apache/iotdb/db/qp/constant/SQLConstant.java
```diff
@@ -151,6 +151,7 @@ private SQLConstant() {
   public static final int TOK_LOAD_CONFIGURATION_LOCAL = 86;
 
   public static final int TOK_SHOW_MERGE_STATUS = 87;
+  public static final int TOK_DELETE_PARTITION = 87;
```
jixuan1989, line 154: still 87?

jt2594838: fixed

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -2410,4 +2410,58 @@ public boolean isFileAlreadyExist(TsFileResource tsFileResource, long partitionN
...
+   * remove all partitions that satisfy a filter.
+   * @param filter
+   */
+  public void removePartitions(TimePartitionFilter filter) {
+    // this requires blocking all other activities
+    insertLock.writeLock().lock();
+    mergeLock.writeLock().lock();
+    try {
+      // abort ongoing merges
+      MergeManager.getINSTANCE().abortMerge(storageGroupName);
```
jixuan1989, line 2424: I notice you add `storageGroupSubTasks` and `storageGroupMainTasks` in 6.9. When to clear the finished tasks?

jt2594838: There is a TaskCleanerThread in MergeManager.

PR #1416, opened 2020-06-24
[IOTDB-726] fix the lineNumber of mlog/snapshot is incorrect after restarting the server 

jixuan1989:
After this PR, the behavior is:
when restarting a server,
- if the mlog exists, it is fine and maybe a snapshot will be created;
- if the snapshot exists, then the mlog must also exist, and its lineNumber should be larger than snapshot, otherwise exception will be thrown.


server/src/main/java/org/apache/iotdb/db/metadata/MLogWriter.java
```diff
@@ -169,4 +169,12 @@ private void newLine() throws IOException {
   int getLineNumber() {
     return lineNumber;
   }
+
+  /**
+   * only used for initialize a mlog file writer.
+   * @param number
+   */
+  void settLineNumber(int number) {
```
samperson1997, line 177:
```suggestion
  void setLineNumber(int number) {
```

PR #1417, opened 2020-06-24
[IOTDB-774]Handle result of "show timeseries" as iterator

SailVR:
Currently, "show timeseries" will cache all timeseries in memory, which may cause OOM.


SailVR:
> Please fix the potential bugs according to the failed tests in `IoTDBSortedShowTimeseriesIT`, `IoTDBTagAlterIT` and `IoTDBTagIT`

OK，Thanks

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -500,7 +501,14 @@ private TSExecuteStatementResp internalExecuteQueryStatement(String statement,
     long queryId = -1;
     try {
       TSExecuteStatementResp resp = getQueryResp(plan, username); // column headers
-
+	  
+	  if(plan instanceof ShowTimeSeriesPlan){
+        if (plan instanceof ShowTimeSeriesPlan) {
```
samperson1997, line 506: Duplicated?

SailVR: modified

PR #1418, opened 2020-06-24
[IOTDB-778] Do not handle measurement name with quotaion uniformly

Genius-pig:
https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-778


tsfile/src/test/java/org/apache/iotdb/tsfile/read/common/PathTest.java
```diff
@@ -41,7 +41,7 @@ public void construct() throws Exception {
     testPath(path, "", "", "");
     // with quote;
     path = new Path("root.d1.r1.\"x1.x2.x3\"");
-    testPath(path, "root.d1.r1", "x1.x2.x3", "root.d1.r1.\"x1.x2.x3\"");
+    testPath(path, "root.d1.r1", "\"x1.x2.x3\"", "root.d1.r1.\"x1.x2.x3\"");
```
Alima777, line 44: Add more tests here, e.g. with single quote

tsfile/src/test/java/org/apache/iotdb/tsfile/read/common/PathTest.java
```diff
@@ -41,7 +41,9 @@ public void construct() throws Exception {
...
```
SilverNarcissus, line 44: Also add some test about " in 'xxx' or ' in "xxx". It will make it more robust. Thanks~

Genius-pig: a path only supports a double quotes.

tsfile/src/test/java/org/apache/iotdb/tsfile/read/common/PathTest.java
Genius-pig: a path only supports double-quotes.

JulianFeinauer:
Just discussed with xiangdong about this PR. Why is this closed? It looks good, not?

PR #1423, opened 2020-06-26
[IOTDB-782] [Experimental] Clear mlog when creating snapshot

samperson1997:
Clear `mlog.txt` when creating snapshot, and thus avoid reading and skipping the lines which are already in `mtree.snapshot`.


samperson1997:
**Experiment result**: (just in my computer)
When recovering from 3000K timeseries, deserializing MTree from `mlog.txt` (including reading and skipping the useless lines) takes up to 7845ms:
```
02:22:05.232 [main] DEBUG org.apache.iotdb.db.metadata.MManager - spend 7845 ms to deserialize mtree from mlog.txt
```
After deleting this useless part, the time could reduce by up to **19.5%**. Maybe when the number of timeseries is larger (for example, 10M), the "useless lines" are more, so we could save more time.

Could you please help and do an experiment with 10M timeseries and compare the result with the origin method @jixuan1989 ? I hope the result will be better this time.

server/src/main/java/org/apache/iotdb/db/metadata/MLogWriter.java
```diff
@@ -160,6 +162,14 @@ public static void upgradeMLog(String schemaDir, String logFileName) throws IOEx
     FSFactoryProducer.getFSFactory().moveFile(tmpLogFile, logFile);
   }
 
+  public void clear() throws IOException {
+    writer.close();
+    Files.delete(logFile.toPath());
+    FileWriter fileWriter = new FileWriter(logFile, true);
```
SilverNarcissus, line 168: Maybe you should use file.getChannel().truncate(0) to clear mlog

PR #1431, opened 2020-06-29
change the way to get MManager

mychaow:
for cluster version could use different MManager


server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -178,6 +178,10 @@ public synchronized void removeItem(String key) {
             MTREE_SNAPSHOT_THREAD_CHECK_TIME, TimeUnit.SECONDS);
   }
 
+  /**
+   * we should not use this function in other place,
```
SilverNarcissus, line 182: Other place is where?  May be you miss something after ','

mychaow: yes, I will add the comments

SilverNarcissus:
Fantastic work!

PR #1433, opened 2020-06-29
Abstract InsertPlan

LebronAl:
Hi, this PR is working on an abstraction of InsertPlan and InsertTabletPlan, which will rename InsertPlan to InsertRowPlan and make an abstraction from the two plans

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/qp/physical/crud/InsertPlan.java
+++ b/server/src/main/java/org/apache/iotdb/db/qp/physical/crud/InsertPlan.java
@@ -16,219 +16,33 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+
 package org.apache.iotdb.db.qp.physical.crud;
 
-import org.apache.iotdb.db.conf.IoTDBConstant;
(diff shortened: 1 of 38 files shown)


LebronAl:
@jt2594838 @qiaojialin PTAL

server/src/main/java/org/apache/iotdb/db/qp/physical/crud/InsertPlan.java
```diff
@@ -369,97 +81,16 @@ public int getFailedMeasurementNumber() {
...
-    this.types = new TSDataType[measurementSize];
-    this.values = new Object[measurementSize];
-    try {
-      setValues(buffer);
-    } catch (QueryProcessException e) {
-      e.printStackTrace();
+  /**
+   * @param index failed measurement index
+   */
+  public void markMeasurementInsertionFailed(int index) {
```
SilverNarcissus, line 87: maybe you should use "markFailedMeasurementInsertion"

LebronAl: OK~

PR #1435, opened 2020-06-29
Add version info in MTree snapshot file name

samperson1997:
Add version info in MTree snapshot file name, in case the format of MTree is updated later.
snapshot file name: `mtree-1.snapshot`
temporary file name: `mtree-1.snapshot.tmp`


samperson1997:
> Excellent work! How about let '-' be a constant for later use?

Hi, actually I refer to `StorageGroupProcessor` and `TsFileProcessor`, where `-` is used directly in file name ; )

samperson1997:
> Excellent work! How about let '-' be a constant for later use?

Thanks! Fixed

PR #1443, opened 2020-06-30
[IOTDB-774]Fix "show timeseries" OOM problem

SailVR:
Currently, "show timeseries" will cache all timeseries in memory, which may cause OOM. 
We defined a ShowTimeSeriesDataSet，Let the server return data in batches
Later，We will optimize the situation “The defined Limit larger than fetchSize”


SailVR:
> Good job.
> 
> However, what will happen if the user defined limit is larger than the fetch size? It seems that you didn't deal with this case in your code.

When the limit is larger than fetchSize, we press fetchSize several times in the background. Let me add one test when limit>fetchSize

SailVR:
> > Good job.
> > However, what will happen if the user defined limit is larger than the fetch size? It seems that you didn't deal with this case in your code.
> 
> When the limit is larger than fetchSize, we press fetchSize several times in the background. Let me add one test when limit>fetchSize

IoTDBSimpleQueryIT.testShowTimeseriesDataSet4()

PR #1448, opened 2020-07-01
[IOTDB-716] add lZ4 compress method

SilverNarcissus:
Support lZ4 compress method


tsfile/pom.xml
```diff
@@ -53,6 +53,11 @@
             <groupId>commons-io</groupId>
             <artifactId>commons-io</artifactId>
         </dependency>
+        <dependency>
+            <groupId>net.jpountz.lz4</groupId>
+            <artifactId>lz4</artifactId>
+            <version>1.3.0</version>
+        </dependency>
```
HTHou, line 60: I think we should update the LICENSE-binary as well. :)

SilverNarcissus: Sure~ I will fix it~

jixuan1989:
SnappyUnCompressor.java 
```java
    @Override
    public int uncompress(byte[] byteArray, int offset, int length, byte[] output, int outOffset)
        throws IOException {
      Snappy.uncompressedLength(byteArray, offset, length);
      return Snappy.uncompress(byteArray, offset, length, output, outOffset);
    }
```
`Snappy.uncompressedLength(byteArray, offset, length);` is not used... either remove it or check whether the length of the output[] satisfies.

jixuan1989:
besides, when something error in uncompression progress, how about throwing ioexception.

SilverNarcissus:
> SnappyUnCompressor.java
> 
> ```java
>     @Override
>     public int uncompress(byte[] byteArray, int offset, int length, byte[] output, int outOffset)
>         throws IOException {
>       Snappy.uncompressedLength(byteArray, offset, length);
>       return Snappy.uncompress(byteArray, offset, length, output, outOffset);
>     }
> ```
> 
> `Snappy.uncompressedLength(byteArray, offset, length);` is not used... either remove it or check whether the length of the output[] satisfies.

Sure, fixed~

SilverNarcissus:
> besides, when something error in uncompression progress, how about throwing ioexception.

Nice suggestion~ fixed~

SilverNarcissus:
> Don't forget to resolve conflicts : )

Sure~ Fixed~

tsfile/src/main/java/org/apache/iotdb/tsfile/compress/ICompressor.java
```diff
@@ -52,6 +52,8 @@ static ICompressor getCompressor(CompressionType name) {
         return new NoCompressor();
       case SNAPPY:
         return new SnappyCompressor();
+      case LZ4:
+        return new IOTDBLZ4Compressor();
```
qiaojialin, line 56: To be consistent with LZ4UnCompressor and simplified, rename this to LZ4Compressor?

tsfile/src/main/java/org/apache/iotdb/tsfile/compress/IUnCompressor.java
```diff
@@ -184,4 +186,82 @@ public CompressionType getCodecName() {
...
+    @Override
+    public byte[] uncompress(byte[] bytes) {
+      if (bytes == null) {
+        return new byte[0];
+      }
+
+      try {
+        return decompressor.decompress(bytes, MAX_COMPRESS_RATIO * bytes.length);
+      } catch (RuntimeException e) {
+        logger.error(
```
qiaojialin, line 227: I think it's better to throw this exception and let the upper level handle it.

tsfile/src/main/java/org/apache/iotdb/tsfile/compress/IUnCompressor.java
```diff
@@ -184,4 +186,82 @@ public CompressionType getCodecName() {
...
+    public int uncompress(ByteBuffer compressed, ByteBuffer uncompressed) {
+      if (compressed == null || !compressed.hasRemaining()) {
+        return 0;
+      }
+
+      try {
+        decompressor.decompress(compressed, uncompressed);
+        return compressed.limit();
+      } catch (RuntimeException e) {
+        logger.error(
```
qiaojialin, line 256: throw IOException

tsfile/src/main/java/org/apache/iotdb/tsfile/compress/ICompressor.java
SilverNarcissus: The package I import is called LZ4Compressor, so I can't use this name

tsfile/src/main/java/org/apache/iotdb/tsfile/compress/IUnCompressor.java
SilverNarcissus: Sure~

tsfile/src/main/java/org/apache/iotdb/tsfile/compress/IUnCompressor.java
SilverNarcissus: Sure~

PR #1452, opened 2020-07-02
Fix logic of creating MTree snapshot

samperson1997:
Fix logic of creating MTree snapshot.
The snapshot should only be created when `System.currentTimeMillis() - logFile.lastModified() >= mtreeSnapshotThresholdTime` AND `logWriter.getLineNumber() >= mtreeSnapshotInterval`

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
@@ -1785,27 +1785,25 @@ public class MManager {
       // the logWriter is not initialized now, we skip the check once.
       return;
     }
-    if (System.currentTimeMillis() - logFile.lastModified() >= mtreeSnapshotThresholdTime
-        && logWriter.getLineNumber() > 0) {
-      logger.info("Start creating MTree snapshot, because {} ms elapse.",
+    if (System.currentTimeMillis() - logFile.lastModified() < mtreeSnapshotThresholdTime) {
+      logger.info("MTree snapshot need not be created. Time from last modification: {} ms.",
           System.currentTimeMillis() - logFile.lastModified());
-      createMTreeSnapshot();
-    } else if (logWriter.getLineNumber() >= mtreeSnapshotInterval) {
-      logger.info("Start creating MTree snapshot, because of {} new lines are added.",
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -1785,27 +1785,25 @@ private void checkMTreeModified() {
       // the logWriter is not initialized now, we skip the check once.
       return;
     }
-    if (System.currentTimeMillis() - logFile.lastModified() >= mtreeSnapshotThresholdTime
-        && logWriter.getLineNumber() > 0) {
-      logger.info("Start creating MTree snapshot, because {} ms elapse.",
+    if (System.currentTimeMillis() - logFile.lastModified() < mtreeSnapshotThresholdTime) {
+      logger.info("MTree snapshot need not be created. Time from last modification: {} ms.",
```
jixuan1989, line 1789: how about debug level

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -1785,27 +1785,25 @@ private void checkMTreeModified() {
...
           System.currentTimeMillis() - logFile.lastModified());
-      createMTreeSnapshot();
-    } else if (logWriter.getLineNumber() >= mtreeSnapshotInterval) {
-      logger.info("Start creating MTree snapshot, because of {} new lines are added.",
+    } else if (logWriter.getLineNumber() < mtreeSnapshotInterval) {
+      logger.info("MTree snapshot need not be created. New mlog line number: {}.",
```
jixuan1989, line 1792: how about debug level

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
samperson1997: Sure

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
samperson1997: Fixed

PR #1457, opened 2020-07-03
delete timeseries when path is not found

Ring-k:
Problem >>> If the DeleteTimeseriesPlan is routed globally and the metadata is removed, and the other node needs to retrieve the full paths, then the metadata is nowhere to be found. 

Solution >>> In this implementation, the full paths will be retrieved before plan routing.


cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -1510,6 +1511,32 @@ private TSStatus processNonPartitionedMetaPlan(PhysicalPlan plan) {
    * @return
    */
   private TSStatus processNonPartitionedDataPlan(PhysicalPlan plan) {
+    if(plan instanceof DeleteTimeSeriesPlan){
+      List<Path> originalPaths = ((DeleteTimeSeriesPlan)plan).getPaths();
```
LebronAl, line 1515: format

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -1510,6 +1511,32 @@ private TSStatus processNonPartitionedMetaPlan(PhysicalPlan plan) {
...
+        });
+      }
+      getAllPathsService.shutdown();
+      try {
+        getAllPathsService.awaitTermination(RaftServer.getQueryTimeoutInSec(), TimeUnit.SECONDS);
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        logger.error("Unexpected interruption when waiting for get all paths services to stop", e);
+      }
+      ((DeleteTimeSeriesPlan)plan).setPaths(new ArrayList<>(fullPaths));
```
LebronAl, line 1538: format

cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
```diff
@@ -707,5 +709,24 @@ public void delete(Path path, long timestamp) throws QueryProcessException {
...
+  @Override
+  public void delete(DeletePlan deletePlan) throws QueryProcessException {
+    try {
+      Set<String> existingPaths = new HashSet<>();
+      for (Path p : deletePlan.getPaths()) {
+        existingPaths.addAll(getPathsName(p.getFullPath()));
+      }
+      if (existingPaths.isEmpty()) {
+        logger.info("TimeSeries does not exist and its data cannot be deleted");
+        return;
```
LebronAl, line 721: This change surely can abandon unnecessary exception, but what if the user deletes a time series which really doesn't exist? I think distributed IoTDB should also throw an exception to user as same as stand-alone IoTDB. Maybe you should add a judgement in `processNonPartitionedDataPlan` after getting all paths.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
Ring-k: Thank you for your reminding :)

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
Ring-k: Thank you for your reminding :)

cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
Ring-k: Thank you for your suggestions :)

LebronAl:
LGTM

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -1522,6 +1532,44 @@ private TSStatus processNonPartitionedDataPlan(PhysicalPlan plan) {
...
+          List<String> fullPathStrs = getMatchedPaths(pathStr);
+          if (fullPathStrs.size() == 0) {
+            nonExistPaths.add(pathStr);
+            logger.error("Path {} is not found.", pathStr);
+          }
+          for (String fullPathStr : fullPathStrs) {
+            fullPaths.add(new Path(fullPathStr));
+          }
+        } catch (MetadataException e) {
+          logger.error("Failed to get full paths of the prefix path: {} because {}.", pathStr, e);
```
LebronAl, line 1555: remove last `{}` or use e.getMessage()

Ring-k: Thanks for your reminding ;)

cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
jt2594838: Agreed, information returned to the user should be consistent with the stand-alone version as much as possible.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -1522,6 +1532,44 @@ private TSStatus processNonPartitionedDataPlan(PhysicalPlan plan) {
     }
   }
 
+  DeleteTimeSeriesPlan getDeleteTimeseriesPlanWithFullPaths(DeleteTimeSeriesPlan plan)
+      throws PathNotExistException {
+    List<Path> originalPaths = plan.getPaths();
+    ConcurrentSkipListSet<Path> fullPaths = new ConcurrentSkipListSet<>();
+    ConcurrentSkipListSet<String> nonExistPaths = new ConcurrentSkipListSet<>();
+    ExecutorService getAllPathsService = Executors
```
jt2594838, line 1540: Suggestion: change the method signature to `List<String> getMatchedPaths(List<Path> originalPaths)` so that this method can be applied in other situations.

cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
```diff
@@ -707,5 +709,24 @@ public void delete(Path path, long timestamp) throws QueryProcessException {
     }
   }
 
+  @Override
+  public void delete(DeletePlan deletePlan) throws QueryProcessException {
+    try {
+      Set<String> existingPaths = new HashSet<>();
+      for (Path p : deletePlan.getPaths()) {
+        existingPaths.addAll(getPathsName(p.getFullPath()));
```
jt2594838, line 717: `getFullPath()` here will still perform a metadata query on other nodes, so I doubt whether your changes work.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
Ring-k: Good idea! I will extract the method.

cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
Ring-k: These codes seems redundent now, because the DeletePlan to be executed contains full paths. I will remove these codes.

PR #1459, opened 2020-07-05
fix maven warning about losing some plugin versions

jixuan1989:
Jenkins build fails but do not know why...

we can fix some warning issues :

image

For more info, read https://builds.apache.org/job/IoTDB-Pipeline/job/master/412/execution/node/67/log/


qiaojialin:
image

Deleting the target folder will resolve this problem

PR #1464, opened 2020-07-06
add cluster version mmanager

mychaow:


cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
```diff
@@ -0,0 +1,233 @@
...
+  private MetaPuller metaPuller;
+
+  protected CMManager() {
+    super();
+    metaPuller = MetaPuller.getInstance();
+    int remoteCacheSize = config.getmRemoteSchemaCacheSize();
+    mRemoteMetaCache = new LRUCache<String, MeasurementMeta>(remoteCacheSize) {
+      @Override
+      protected MeasurementMeta loadObjectByKey(String key) throws IOException {
+        throw new IOException(key + " not found!");
```
jt2594838, line 58: Throwing an exception would fill the stack trace, so how about changing it to return null?

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  @Override
+  public TSDataType getSeriesType(String path) throws MetadataException {
+    try {
+      cacheLock.readLock().lock();
+      MeasurementMeta measurementMeta = mRemoteMetaCache.get(path);
+      return measurementMeta.getMeasurementSchema().getType();
+    } catch (IOException e) {
+      //do nothing
```
jt2594838, line 110: When is the schema pulled in this case?

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
```diff
@@ -0,0 +1,233 @@
...
+          MeasurementMeta measurementMeta = mRemoteMetaCache.get(deviceId + measurements[i]);
+          measurementSchemas[i] = measurementMeta.getMeasurementSchema();
+        } catch (IOException ex) {
+          allExist = false;
+          break;
+        }
+      }
+      cacheLock.readLock().unlock();
+      if (!allExist) {
+        throw new MetadataException(deviceId + " has some mesurements not found");
```
jt2594838, line 158: It would be better to show which series is not found and just throw PathNotExistException.

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
```diff
@@ -0,0 +1,233 @@
...
+      return measurementMeta.getTimeValuePair();
+    } catch (IOException e) {
+      // do nothing
+    }
+    return super.getLastCache(seriesPath);
+  }
+
+  @Override
+  public MeasurementSchema[] getSeriesSchemasAndReadLockDevice(String deviceId, String[] measurementList, InsertPlan plan) throws MetadataException {
+    //TODO cluster also need to lock device node
```
jt2594838, line 212: I think this is urgent, or you can extract the lock method separately. Otherwise, an IllegalMonitorStateException will be thrown because the next step will try to unlock but it is not locked.

cluster/src/main/java/org/apache/iotdb/cluster/metadata/MetaPuller.java
```diff
@@ -0,0 +1,182 @@
...
+      logger.debug("{}: pulling schemas of {} and other {} paths from {} groups", metaGroupMember.getName(),
+        prefixPaths.get(0), prefixPaths.size() - 1,
+        partitionGroupPathMap.size());
+    }
+    for (Map.Entry<PartitionGroup, List<String>> partitionGroupListEntry : partitionGroupPathMap
+      .entrySet()) {
+      PartitionGroup partitionGroup = partitionGroupListEntry.getKey();
+      List<String> paths = partitionGroupListEntry.getValue();
+      pullTimeSeriesSchemas(partitionGroup, paths, schemas);
+    }
```
jt2594838, line 107: How about using parallel pulling?

cluster/src/main/java/org/apache/iotdb/cluster/server/handlers/caller/AppendNodeEntryHandler.java
```diff
@@ -66,7 +67,7 @@ public void onComplete(Long response) {
         logger.debug("{}: Received an agreement from {} for {}, remaining votes to succeed: {}",
             member.getName(), receiver, log, remaining);
         if (remaining == 0) {
-          logger.debug("{}: Log {} is accepted by the quorum", member.getName(), log);
+          logger.debug("{}: Log {} is accepted by the quorum {}", member.getName(), log, log.getCurrLogIndex());
```
jt2594838, line 70: Adding the parameter after "quorum" is a bit confusing, maybe we can use:
`"{}: Log[{}] {} is accepted by the quorum", member.getName(), log.getCurrLogIndex(), log`

cluster/src/test/java/org/apache/iotdb/cluster/log/applier/DataLogApplierTest.java
```diff
@@ -184,15 +182,15 @@ public void testApplyInsert()
...
-    assertFalse(dataSet.hasNext());
+//    insertPlan.setDeviceId(TestUtils.getTestSg(4));
+//    applier.apply(log);
+//    dataSet = query(Collections.singletonList(TestUtils.getTestSeries(4, 0)), null);
+//    assertTrue(dataSet.hasNext());
+//    record = dataSet.next();
+//    assertEquals(1, record.getTimestamp());
+//    assertEquals(1, record.getFields().size());
+//    assertEquals(1.0, record.getFields().get(0).getDoubleV(), 0.00001);
+//    assertFalse(dataSet.hasNext());
```
jt2594838, line 193: Why are these commented?

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
mychaow: ok

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
mychaow: the caller will pull by themselves.

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
mychaow: ok

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
mychaow: yes, I will fix it

cluster/src/main/java/org/apache/iotdb/cluster/metadata/MetaPuller.java
mychaow: It will be complicated, I will do it later

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
mychaow: yes, I will fix it

jt2594838: Yes, I added that just as a temporary solution, but I still hope you consider to do some refactoring, maybe not in this PR, though, since generating exceptions is somehow burdensome.

jt2594838:
I think it will be ready after you fix these conflicts.

One more thing, please do not use force-push, or I will not be able to view changes from my last review and I must review the changed files from the beginning.

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
mychaow: ok，by the way, I use force-push just because some modifications are small, so I think it could be merged with the previous  commit. Anyway, I will avoid do it again.

PR #1481, opened 2020-07-10
[IOTDB-792] do not get MManager.readLock when inserting to avoid deadlock

qiaojialin:


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -835,29 +823,17 @@ private void insertToTsFileProcessor(InsertRowPlan insertRowPlan, boolean sequen
...
+    String[] measurementList = plan.getMeasurements();
+    for (int i = 0; i < measurementList.length; i++) {
+      if (plan.getValues()[i] == null) {
+        continue;
       }
-    } catch (MetadataException e) {
-      // skip last cache update if the local MTree does not contain the schema
-    } finally {
+      // Update cached last value with high priority
       if (node != null) {
```
mychaow, line 834: you can move this line out of cycle. `if (node != null) return`

qiaojialin: actually, this is for test... Some tests insert data into StorageGroupProcessor directely

mychaow: the code for test? Why do we need merge it into master? I think the test code should be put into the test branch.

qiaojialin: fixed

PR #1494, opened 2020-07-14
[IOTDB-768] Extend more functions in aggregation by level

HTHou:
Now support: 
sum, min_time, max_time, min_value, max_value
To do: 
avg, last_value, first_value


server/src/main/java/org/apache/iotdb/db/query/executor/AggregationExecutor.java
```diff
@@ -318,28 +318,80 @@ private void aggregateWithValueFilter(List<AggregateResult> aggregateResults,
...
-
-      RowRecord curRecord = FilePathUtils.mergeRecordByPath(record, finalPaths, pathIndex);
-
       List<Path> paths = new ArrayList<>();
       List<TSDataType> dataTypes = new ArrayList<>();
-      for (int i = 0; i < finalPaths.size(); i++) {
-        dataTypes.add(TSDataType.INT64);
+      RowRecord curRecord = null;
+      Set<String> finalPaths = FilePathUtils.getPathByLevel(plan, pathIndex);
+      switch (aggregation) {
```
mychaow, line 344: you should check the datatype of the data points, sum could be caculated by int and float, but text and boolean should not be supportted

HTHou: Yes, the check is necessary. Fixing~

HTHou:
Since there is another PR fixing this issue, I closed this PR.

PR #1508, opened 2020-07-16
[IOTDB-798] [To rel/0.10] fix a set rowLimit and rowOffset bug

HTHou:
https://github.com/apache/incubator-iotdb/issues/1505


qiaojialin:
It seems that limit is not supported in group by fill

HTHou:
> It seems that limit is not supported in group by fill

Sure, fixed~ Thanks!

qiaojialin:
This is a bug fix of group by. We'd better to support limit in group by fill also.

Besides, I prefer to set the limit and offset in one place, not in each query type

HTHou:
> We'd better to support limit in group by fill also.

Yes, we can support it in 0.11. 

> Besides, I prefer to set the limit and offset in one place, not in each query type

Fixed, please take a look.

HTHou:
> This is a bug fix of group by. We'd better to support limit in group by fill also.
> 
> Besides, I prefer to set the limit and offset in one place, not in each query type

I updated the antlr part to support limit in group by fill.

PR #1512, opened 2020-07-17
[IOTDB-801]Separate data port and heartbeat port

neuyilan:


cluster/src/main/java/org/apache/iotdb/cluster/client/async/AsyncDataHeartbeatClient.java
```diff
@@ -0,0 +1,122 @@
...
+import org.apache.thrift.protocol.TProtocolFactory;
+import org.apache.thrift.transport.TNonblockingSocket;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Notice: Because a client will be returned to a pool immediately after a successful request, you
+ * should not cache it anywhere else or there may be conflicts.
+ */
+public class AsyncDataHeartbeatClient extends AsyncClient {
```
jt2594838, line 40: I think maybe you can just extend AsyncDataClient and override the factory and constructor so there will be fewer duplicated codes.

Or maybe you can extract heartbeat-related methods in RaftService into a new service, so it can be better decoupled.

cluster/src/main/java/org/apache/iotdb/cluster/utils/ClusterUtils.java
```diff
@@ -48,11 +50,24 @@
 
   public static final long START_UP_CHECK_TIME_INTERVAL = 5; // second
 
+  /**
+   * the data group member's heartbeat offset relative to the {@link ClusterConfig#getLocalDataPort()},
+   * which means the dataHeartbeatPort = getLocalDataPort() + DATA_HEARTBEAT_OFFSET.
+   */
+  public static final int DATA_HEARTBEAT_OFFSET = 1;
```
jt2594838, line 57: How about using "DATA_HEARTBEAT_PORT_OFFSET" so it would be more clear what the port is for?

cluster/src/main/java/org/apache/iotdb/cluster/client/async/AsyncDataHeartbeatClient.java
neuyilan: Good idea, I will resolve it.

cluster/src/main/java/org/apache/iotdb/cluster/utils/ClusterUtils.java
neuyilan: sure, i will change it

PR #1526, opened 2020-07-20
[IOTDB-809] Separate read and write configurations

Ring-k:


LebronAl:
Great job!
BTW, I'm actually not sure what these two read-write timeout configurations mean.For `write_operation_timeout_ms`, if it means that a write request can't execute more than `write_operation_timeout_ms ` in user's view, then maybe we should track the entire request to implement this.It will be more complicated then...

Ring-k:
These two configurations are thresholds for read and write requests from user's view. When it reaches the threshold, the execution should terminate and the user should be informed.

LebronAl:
> These two configurations are thresholds for read and write requests from user's view. When it reaches the threshold, the execution should terminate and the user should be informed. What this pull request do is to separate the read and write constraints from `connection_timeout_ms` to improve flexibility of the configurations.

In async mode, it seems that there are more places which should be controlled by these configurations.Such as a raft broadcast in `sendLogToFollowers`,please make some checks to cover all~
image
In sync mode,how are you going to apply these configurations?

Ring-k:
> > These two configurations are thresholds for read and write requests from user's view. When it reaches the threshold, the execution should terminate and the user should be informed. What this pull request do is to separate the read and write constraints from `connection_timeout_ms` to improve flexibility of the configurations.
> 
> In async mode, it seems that there are more places which should be controlled by these configurations.Such as a raft broadcast in `sendLogToFollowers`,please make some checks to cover all~
> image
> In sync mode,how are you going to apply these configurations?

I see. But the `connection_time_ms` is not totally replaced by `write_operation_timeout_ms` and `read_operation_timeout_ms`.  It is still used in heartbeat among nodes, or async mode. Is it really necessary to separate read and write there?

cluster/src/main/java/org/apache/iotdb/cluster/client/rpcutils/TimeoutChangeableTFastFramedTransport.java
```diff
@@ -0,0 +1,38 @@
...
+  TTransport underlying;
+
+  public TimeoutChangeableTFastFramedTransport(TTransport underlying) {
+    super(underlying);
+    this.underlying = underlying;
+  }
+
+  public void setTimeout(int timeout){
+    ((TSocket)underlying).setTimeout(timeout);
+  }
```
jt2594838, line 37: It would be better if you enforce the parameter in the constructor to be TSocket, so there will be less chance to trigger a ClassCastException when the code is modified by another person.

cluster/src/main/java/org/apache/iotdb/cluster/client/sync/SyncDataClient.java
```diff
@@ -39,13 +39,20 @@
...
     this.node = node;
     this.pool = pool;
     getInputProtocol().getTransport().open();
   }
 
+  public void setTimeout(int timeout) {
+    ((TimeoutChangeableTFastFramedTransport) (getInputProtocol().getTransport()))
+        .setTimeout(timeout);
+    ((TimeoutChangeableTFastFramedTransport) (getOutputProtocol().getTransport()))
+        .setTimeout(timeout);
```
jt2594838, line 53: The same transport is used in both input and output, so it is unnecessary to set twice.

cluster/src/assembly/resources/conf/iotdb-cluster.properties
```diff
@@ -58,7 +58,13 @@ max_concurrent_client_num=10000
...
 
 # connection time out (ms) among raft nodes
-connection_time_out_ms=20000
+connection_timeout_ms=20000
+
+# write operation timeout threshold (ms)
+write_operation_timeout_ms=30000
+
+# read operation timeout threshold (ms)
+read_operation_timeout_ms=30000
```
jt2594838, line 67: Add comments explaining that the configs are only for internal communications, not for the whole operation.

cluster/src/main/java/org/apache/iotdb/cluster/client/rpcutils/TimeoutChangeableTFastFramedTransport.java
Ring-k: Sure. That's more clear.

cluster/src/main/java/org/apache/iotdb/cluster/client/sync/SyncDataClient.java
Ring-k: Indeed. I will remove one.

cluster/src/assembly/resources/conf/iotdb-cluster.properties
Ring-k: Thanks for  your advice. I will add these comments.

PR #1529, opened 2020-07-21
Fix recover bug with vm

JackieTien97:
After adding vm, the restart is not performed as planned.


server/src/main/java/org/apache/iotdb/db/engine/flush/MemTableFlushTask.java
```diff
@@ -62,9 +63,11 @@
   private volatile boolean noMoreEncodingTask = false;
   private volatile boolean noMoreIOTask = false;
 
-  public MemTableFlushTask(IMemTable memTable, RestorableTsFileIOWriter writer, String storageGroup) {
+  public MemTableFlushTask(IMemTable memTable, RestorableTsFileIOWriter writer, String storageGroup,
```
qiaojialin, line 66: add javadoc about these parameters

PR #1533, opened 2020-07-22
[IOTDB-807]Make heartbeat thread commit log tasks asynchronous

neuyilan:


cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
```diff
@@ -1564,4 +1582,33 @@ public void setAppendLogThreadPool(ExecutorService appendLogThreadPool) {
...
+    @Override
+    public void onSuccess() {
+      synchronized (syncLock) {
+        syncLock.notifyAll();
+      }
+    }
+
+    @Override
+    public void onError(Exception e) {
+      logger.error("async commit log failed, {}", e.toString());
```
jt2594838, line 1610: `"async commit log failed", e` is more generally used.

PR #1541, opened 2020-07-22
Add file list on disk 

qiaojialin:


qiaojialin:
> LGTM. And shall we add the new feature that is in data/tracing ?

Yes, I forgot it. You can add it after this PR is merged

PR #1542, opened 2020-07-22
modify docs from v0.10.0 to v0.10.1

jixuan1989:


jixuan1989:
this branch is not clean.

PR #1543, opened 2020-07-23
[IOTDB-811] fix upgrading mlog many times when upgrading system.properties crashed

HTHou:


server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfigCheck.java
```diff
@@ -191,6 +191,18 @@ public void checkConfig() throws IOException {
...
+      File mlogFile = SystemFileFactory.INSTANCE.getFile(SCHEMA_DIR + File.separator
+          + MetadataConstant.METADATA_LOG);
+      File tmpMLogFile = SystemFileFactory.INSTANCE.getFile(mlogFile.getAbsolutePath()
+          + ".tmp");
+
+      if (!mlogFile.delete()) {
+        throw new IOException("Deleting " + mlogFile + "failed.");
+      }
+      // rename tmpLogFile to mlog
+      FSFactoryProducer.getFSFactory().moveFile(tmpMLogFile, mlogFile);
```
samperson1997, line 205: It suddenly occurs to me that `mlog` is a system file but `FSFactoryProducer.getFSFactory()` is for data files (like tsfile)... This may result in the problem that when upgrading 0.9 version tsfiles, HDFS cannot find `mlog` because it is actually stored in local.

Sorry I forgot and ignored it in the PR #1540 yesterday... 😣
```suggestion
      FileUtils.moveFile(tmpMLogFile, mlogFile);
```

PR #1544, opened 2020-07-23
[IOTDB-810] Set Configuration of Max Number of Logs in Memory

Ring-k:


LebronAl:
I think when the max_num_of_logs_in_mem threshold is reached we should delete the logs in memory to min_num_of_logs_in_mem rather than throwing an exception to the client.
These two configurations  `max_num_of_logs_in_mem` and `log_deletion_check_interval_second ` are both used to ensure that the Raft logs did not take up whole memory  rather than slow down the write rate for the client.

cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterConfig.java
```diff
@@ -63,7 +63,12 @@
   /**
    * max number of committed logs to be saved
    */
-  private int maxNumberOfLogs = 100;
+  private int minNumOfLogsInMem = 100;
+
+  /**
+   * max number of committed logs in memory
+   */
+  private int maxNumOfLogsInMem = 100;
```
jt2594838, line 71: Maybe this should be 1000 or at least a different number from min.

cluster/src/assembly/resources/conf/iotdb-cluster.properties
```diff
@@ -66,8 +66,11 @@ max_unsnapshoted_log_size=134217728
 # whether to use batch append entries in log catch up
 use_batch_in_catch_up=true
 
-# max number of committed logs to be saved
-max_number_of_logs=100
+# minimum number of committed logs in memory
+min_num_of_logs_in_mem=100
+
+# maximum number of committed logs in memory
+max_num_of_logs_in_mem=1000
```
jt2594838, line 73: Add more specific comments like
 "# minimum number of committed logs in memory, after each log deletion, at most such number of logs will remain in memory. Increasing the number will reduce the chance to use snapshot in catch-ups, but will also increase memory footprint"
"# maximum number of committed logs in memory, when reached, a log deletion will be triggered. Increasing the number will reduce the chance to use snapshot in catch-ups, but will also increase memory footprint"

cluster/src/main/java/org/apache/iotdb/cluster/log/manage/RaftLogManager.java
```diff
@@ -415,6 +423,9 @@ public void commitTo(long newCommitIndex, boolean ignoreExecutionExceptions)
               .clear();
         }
         try {
+          if (committedEntryManager.getTotalSize() + entries.size() > maxNumOfLogsInMem) {
+            throw new LogNumberOutOfBoundException();
+          }
```
jt2594838, line 428: You should trigger a log deletion and proceed the committing instead of throwing an exception.

cluster/src/main/java/org/apache/iotdb/cluster/exception/LogNumberOutOfBoundException.java
```diff
@@ -0,0 +1,35 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iotdb.cluster.exception;
+
+public class LogNumberOutOfBoundException extends Exception{
```
jt2594838, line 22: This exception is unneeded, and we should solve it silently by triggering a log deletion.

Ring-k: Thanks for your advice. I will remove it.

cluster/src/assembly/resources/conf/iotdb-cluster.properties
Ring-k: Sure, that will be more clear.

cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterConfig.java
Ring-k: Sorry for the mistake. I will correct it.

cluster/src/main/java/org/apache/iotdb/cluster/log/manage/RaftLogManager.java
Ring-k: I see. Thanks for your suggestion.

Ring-k:
> I think when the max_num_of_logs_in_mem threshold is reached we should delete the logs to min_num_of_logs_in_mem rather than throwing an exception to the client.
> These two configurations `max_num_of_logs_in_mem` and `log_deletion_check_interval_second ` are both used to ensure that the Raft logs did not take up whole memory rather than slow down the write rate for the client.

I got you. I will change these logics.

PR #1561, opened 2020-07-25
limit log size of batch catch up task

mychaow:


cluster/src/main/java/org/apache/iotdb/cluster/log/catchup/LogCatchUpTask.java
```diff
@@ -50,8 +52,9 @@
 
   // sending logs may take longer than normal communications
   private static final long SEND_LOGS_WAIT_MS = 5 * 60 * 1000L;
+  private static final int LOG_NUM_IN_BATCH = 1024;
+  private static final int LEFT_SIZE_IN_REQUEST = 1024;
```
LebronAl, line 56: These variables seems to be easier to maintain and modify in the clusterUtil or clusterConfig

cluster/src/main/java/org/apache/iotdb/cluster/log/catchup/LogCatchUpTask.java
```diff
@@ -178,11 +181,21 @@ void doLogCatchUpInBatch() throws TException, InterruptedException {
...
+      long totalLogSize = 0;
+      int newStart = i;
+      for (int curNum = 0; curNum < LOG_NUM_IN_BATCH && i < logs.size(); i++, curNum++) {
+        ByteBuffer logData = logs.get(i).serialize();
+        totalLogSize += logData.array().length;
+        // we should send logs who's size is smaller than the max frame size of thrift
+        // left 200 byte for other fields of AppendEntriesRequest
+        if (totalLogSize > IoTDBDescriptor.getInstance().getConfig().getThriftMaxFrameSize() - LEFT_SIZE_IN_REQUEST) {
+          break;
+        }
```
jt2594838, line 195: Report an error or warning here if the logList is empty when breaking, which means the configured size is too small to hold even one log and this may end up in a dead loop.

But more importantly, does it matter if a request's size exceeds FrameSize? Will not thrift break a request into several frames? If so, what is the point to enforce request size, and if not, why does thrift call it a FramedTransport?

mychaow: Not break a request into several frames. If a request is too large, it will be failed. A frame hold a full request data.  https://blog.csdn.net/duyuanhai/article/details/53332981, https://andrewpqc.github.io/2019/02/24/thrift/

PR #1572, opened 2020-07-28
fix doc links

LebronAl:
Hi,I find some wrong links in docs~


LebronAl:
@qiaojialin PTAL

qiaojialin:
@LebronAl Hi, the link works correctly with .html, no need to change to md.

PR #1580, opened 2020-07-29
[IOTDB-821] Flush Raft Logs in Batches

Ring-k:


cluster/src/main/java/org/apache/iotdb/cluster/log/manage/serializable/SyncLogDequeSerializer.java
```diff
@@ -621,6 +650,7 @@ private void actuallyDeleteFile() {
         logger.debug("Examining file for removal, file: {}, len: {}, removedLogSize: {}", logFile
             , logFile.length(), removedLogSize);
       }
+      long len = logFile.length();
```
jt2594838, line 653: Where is this used？

Ring-k: It was used for debugging. I will remove it. Thanks for your reminding.

PR #1582, opened 2020-07-29
[IOTDB-822] Flush Raft Logs Periodically

Ring-k:


cluster/src/main/java/org/apache/iotdb/cluster/log/manage/RaftLogManager.java
```diff
@@ -100,6 +100,16 @@ public RaftLogManager(StableEntryManager stableEntryManager, LogApplier applier,
...
             TimeUnit.SECONDS);
+
+    /**
+     * flush log to file periodically
+     */
+    int logFlushTimeIntervalMS = ClusterDescriptor.getInstance().getConfig()
+        .getForceRaftLogPeriodInMS();
+    if (ClusterDescriptor.getInstance().getConfig().isEnableRaftLogPersistence()) {
+      executorService.scheduleAtFixedRate(this::flushLogPeriodically, logFlushTimeIntervalMS,
+          logFlushTimeIntervalMS, TimeUnit.MILLISECONDS);
```
jt2594838, line 111: shutdown this when the manager is closed.

Ring-k: Thanks for your advice. I will use a new executor service for raft log writing.

PR #1588, opened 2020-07-31
Enable iotdb to set useasyncserver and useasyncapplier through configuration files

wyh-ztf:
It is convenient to change the corresponding variables in the code by modifying the configuration file during testing. It is not necessary to change the code and recompile before each test.

### DIFF
--- a/cluster/src/assembly/resources/conf/iotdb-cluster.properties
+++ b/cluster/src/assembly/resources/conf/iotdb-cluster.properties
@@ -107,11 +107,11 @@ enable_auto_create_schema=true
 # Weak consistency do not synchronize with the leader and simply use the local data
 consistency_level=mid
 
-#is use async server
+#Whether to use an asynchronous server
 is_use_async_server=true
 
-#is use sync applier
-is_use_async_applier=true
+#Whether to use asynchronous applications
+is_use_async_applier=false
 
 # is raft log persistence enabled
 is_enable_raft_log_persistence=true
--- a/cluster/src/assembly/resources/conf/iotdb-cluster.properties
+++ b/cluster/src/assembly/resources/conf/iotdb-cluster.properties
@@ -107,10 +107,10 @@ enable_auto_create_schema=true
 # Weak consistency do not synchronize with the leader and simply use the local data
 consistency_level=mid
 
-#Whether to use an asynchronous server
+# Whether to use asynchronous server
 is_use_async_server=true
 
-#Whether to use asynchronous applier
+# Whether to use asynchronous applier
 is_use_async_applier=false
 
 # is raft log persistence enabled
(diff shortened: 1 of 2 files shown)


cluster/src/assembly/resources/conf/iotdb-cluster.properties
```diff
@@ -107,6 +107,12 @@ enable_auto_create_schema=true
 # Weak consistency do not synchronize with the leader and simply use the local data
 consistency_level=mid
 
+#is use async server
+is_use_async_server=true
+
+#is use sync applier
+is_use_async_applier=true
```
neuyilan, line 114: please make the annotation more clearer.

cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterDescriptor.java
```diff
@@ -244,6 +244,16 @@ private void loadProps() {
         .getProperty("enable_auto_create_schema",
             String.valueOf(config.isEnableAutoCreateSchema()))));
 
+    config.setUseAsyncServer(
+            Boolean.parseBoolean(properties.getProperty("is_use_async_server",
+                    String.valueOf(config.isUseAsyncServer()))));
+    logger.info("server+{}",config.isUseAsyncServer());
```
neuyilan, line 250: remove the debug log

cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterDescriptor.java
```diff
@@ -244,6 +244,16 @@ private void loadProps() {
...
+
+    config.setUseAsyncApplier(
+            Boolean.parseBoolean(properties.getProperty("is_use_async_applier",
+                    String.valueOf(config.isUseAsyncApplier()))));
+    logger.info("app+{}",config.isUseAsyncApplier());
```
neuyilan, line 255: the same as above

cluster/src/assembly/resources/conf/iotdb-cluster.properties
neuyilan, line 114: the default value is false. make sure consistent with the code

neuyilan:
LGTM

cluster/src/assembly/resources/conf/iotdb-cluster.properties
```diff
@@ -107,6 +107,12 @@ enable_auto_create_schema=true
 # Weak consistency do not synchronize with the leader and simply use the local data
 consistency_level=mid
 
+#Whether to use an asynchronous server
+is_use_async_server=true
+
+#Whether to use asynchronous applications
+is_use_async_applier=false
```
neuyilan, line 114: applications -> applier

PR #1590, opened 2020-08-03
Use readLock in `getActiveRatio`

dddddai:
```java
@Override
  public double getActiveRatio(String storageGroup) {
    lock.writeLock().lock();
    double ratio;
    try {
      ratio = activeRatioMap.get(storageGroup);
    } catch (Exception e) {
      ratio = 0;
      LOGGER.error("Get active ratio failed", e);
    } finally {
      lock.writeLock().unlock();
    }
    return ratio;
  }
```

It should use readLock instead of writeLock here


jixuan1989:
Thanks for your contribution!
LGTM.

PR #1597, opened 2020-08-05
add tsfilemanage-merge

zhanglingzhe0820:


server/src/main/java/org/apache/iotdb/db/engine/tsfilemanagement/TsFileManagement.java
```diff
@@ -0,0 +1,87 @@
...
+
+  public abstract boolean isEmpty(boolean sequence);
+
+  public abstract int size(boolean sequence);
+
+  public abstract void recover();
+
+  public abstract void forkCurrentFileList();
+
+  protected abstract void merge(ReadWriteLock hotCompactionMergeLock);
```
JackieTien97, line 68: It's better to add some java doc for these abstract methods. It will be useful for someone else to implement another TsFileManagement.

server/src/main/java/org/apache/iotdb/db/engine/tsfilemanagement/level/LevelTsFileManagement.java
```diff
@@ -0,0 +1,508 @@
...
+
+  private void deleteLevelFiles(List<TsFileResource> vmMergeTsFiles) {
+    logger.debug("{} [hot compaction] merge starts to delete file", storageGroupName);
+    for (TsFileResource vmMergeTsFile : vmMergeTsFiles) {
+      deleteLevelFile(vmMergeTsFile);
+    }
+    for (int i = 0; i < maxLevelNum; i++) {
+      sequenceTsFileResources.get(i).removeAll(vmMergeTsFiles);
+      unSequenceTsFileResources.get(i).removeAll(vmMergeTsFiles);
+    }
```
JackieTien97, line 79: Why we need to iterate each level to delete vmMergeTsFiles? Cause, according to my understanding, they should be on the same level.

server/src/main/java/org/apache/iotdb/db/engine/tsfilemanagement/level/LevelTsFileManagement.java
```diff
@@ -0,0 +1,508 @@
...
+  }
+
+  private void flushAllFilesToLastLevel(List<List<TsFileResource>> currMergeFiles,
+      HotCompactionLogger hotCompactionLogger, boolean sequence,
+      ReadWriteLock hotCompactionMergeLock) throws IOException {
+    TsFileResource sourceFile = currMergeFiles.get(0).get(0);
+    File newTargetFile = createNewTsFileName(sourceFile.getTsFile(), maxLevelNum - 1);
+    TsFileResource targetResource = new TsFileResource(newTargetFile);
+    HotCompactionUtils.merge(targetResource, getTsFileList(sequence),
+        storageGroupName, hotCompactionLogger, new HashSet<>(), sequence);
```
JackieTien97, line 105: It seems that we should use currMergeFiles instead of `getTsFileList(sequence)`.

server/src/main/java/org/apache/iotdb/db/engine/tsfilemanagement/level/LevelTsFileManagement.java
```diff
@@ -0,0 +1,508 @@
...
+  }
+
+  @Override
+  public void forkCurrentFileList() {
+    forkTsFileList(forkedSequenceTsFileResources, sequenceTsFileResources);
+    forkTsFileList(forkedUnSequenceTsFileResources, unSequenceTsFileResources);
+  }
+
+  private void forkTsFileList(List<List<TsFileResource>> forkedSequenceTsFileResources,
+      List<List<TsFileResource>> sequenceTsFileResources) {
```
JackieTien97, line 371: change the parameter name. it can be either sequenceTsFileResources or unSequenceTsFileResources

server/src/main/java/org/apache/iotdb/db/engine/tsfilemanagement/level/LevelTsFileManagement.java
```diff
@@ -0,0 +1,508 @@
...
+    forkedSequenceTsFileResources.clear();
+    for (int i = 0; i < maxLevelNum - 1; i++) {
```
JackieTien97, line 373: sequenceTsFileResources' size is `maxLevelNum + 1`, so should the `maxLevelNum - 1` be `maxLevelNum`?

zhanglingzhe0820: no, the maxLevelNum - 1 is the last level of hot compaction, and the maxLevelNum is for the all merge completed level

PR #1598, opened 2020-08-05
server log_info.log does not contain INFO level log

haimeiguo:
when using log.info method, the result is output to log_all file, now updated to log_info file

### DIFF
--- a/server/src/assembly/resources/conf/logback.xml
+++ b/server/src/assembly/resources/conf/logback.xml
@@ -192,7 +192,7 @@
         <appender-ref ref="FILEDEBUG"/>
         <appender-ref ref="FILEWARN"/>
         <appender-ref ref="FILEERROR"/>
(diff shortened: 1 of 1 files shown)


server/src/assembly/resources/conf/logback.xml
```diff
@@ -192,7 +192,7 @@
         <appender-ref ref="FILEDEBUG"/>
         <appender-ref ref="FILEWARN"/>
         <appender-ref ref="FILEERROR"/>
-        <appender-ref ref="FILEALL"/>
```
neuyilan, line 195: The issue is to resolve the problem that the log_info.log do not contains the INFO level log,
log_all.log contains logs which  level is greater than or equal to INFO level,  please do not remove the 'FILEALL' log.

jixuan1989:
Thanks for your contribution!

PR #1600, opened 2020-08-05
[ISSUE#1599]check datatype with encoding.

Genius-pig:
check datatype with encoding.


jixuan1989:
pls fix sonar code smells.

PR #1634, opened 2020-08-19
Enable to remove node in the distributed environment

wyh-ztf:


cluster/src/main/java/org/apache/iotdb/cluster/ClusterMain.java
```diff
@@ -211,6 +211,10 @@ private static void doRemoveNode(String[] args) throws IOException {
     String ip = args[1];
     int metaPort = Integer.parseInt(args[2]);
     ClusterConfig config = ClusterDescriptor.getInstance().getConfig();
+    if(config.getReplicationNum()==1||config.getReplicationNum()==2) {
+      logger.error("When default_ replica_ When num = 1 or 2, the node cannot be removed temporarily, because in the current implementation, datagroup cannot achieve majority agreement");
```
neuyilan, line 215: “default_ replica_ When“ what this mean?

cluster/src/assembly/resources/sbin/remove-node.sh
```diff
@@ -0,0 +1,75 @@
...
+launch_service()
+{
+	class="$1"
+	iotdb_parms="-Dlogback.configurationFile=${IOTDB_CONF}/logback.xml"
+	iotdb_parms="$iotdb_parms -DIOTDB_HOME=${IOTDB_HOME}"
+	iotdb_parms="$iotdb_parms -DTSFILE_HOME=${IOTDB_HOME}"
+	iotdb_parms="$iotdb_parms -DIOTDB_CONF=${IOTDB_CONF}"
+	iotdb_parms="$iotdb_parms -Dname=iotdb\.IoTDB"
+	ip=`sed '/^cluster_rpc_ip=/!d;s/.*=//' $IOTDB_CONF/iotdb-cluster.properties`
+  meta_port=`sed '/^internal_meta_port=/!d;s/.*=//' $IOTDB_CONF/iotdb-cluster.properties`
```
neuyilan, line 67: code format

neuyilan (changes requested):
Good Job!

jt2594838 (changes requested):
Thanks for your contribution, but there are a few points that we should explain to you:
1. The node being removed is not bound to a local node, and it is very likely that we remove a node by another node, especially when the node to be removed is unavailable. So the node should be set by the user manually.
2. The removal of a node does not require the agreement of any data group, and the only thing that matters is that the meta group reaches an agreement, and data groups will be adjusted accordingly, so it does not concern the number of replicas.
3. Node-removal is still under development and requires testing, so it would be better not to pay too much attention to it currently.
4. The whole protect follows Google-style which is described in the project root directory, please see to it.

wyh-ztf:
> explain

PR #1635, opened 2020-08-19
[IOTDB-860] Emend the async log applier

neuyilan:


jt2594838 (changes requested):
The fix itself has a potential problem that when new logs keep coming in, it is very likely to time out.
Now I see you want to make sure when the snapshot is taken, all previous logs are applied. This side effect has been noticed but back to then, AsyncApplier is merely in the experimental stage. Since you have fixed one of the problems, I would like to give some pieces of advice to make AsyncApplier complete:
1. When starting to take a snapshot, record the current commitIndex.
2. Block until all logs whose indices <= the recorded commitIndex are applied. (Use RaftLogManager to do so instead of LogApplier)
3. Prevent the log cleaner thread to clean logs that are not applied.
4. Change `StorageEngine.getInstance().syncCloseAllProcessor();` in `takeSnapshot()` to send a flush plan within the group. (So the file sequence will not be broken by snapshots)
5. When committed logs are recovered during start-up, re-apply all of them. (Notice that operation sequences in IoTDB are idempotent)

neuyilan:
> The fix itself has a potential problem that when new logs keep coming in, it is very likely to time out.
> Now I see you want to make sure when the snapshot is taken, all previous logs are applied. This side effect has been noticed but back to then, AsyncApplier is merely in the experimental stage. Since you have fixed one of the problems, I would like to give some pieces of advice to make AsyncApplier complete:
> 
> 1. When starting to take a snapshot, record the current commitIndex.
> 2. Block until all logs whose indices <= the recorded commitIndex are applied. (Use RaftLogManager to do so instead of LogApplier)
> 3. Prevent the log cleaner thread to clean logs that are not applied.
> 4. Change `StorageEngine.getInstance().syncCloseAllProcessor();` in `takeSnapshot()` to send a flush plan within the group. (So the file sequence will not be broken by snapshots)
> 5. When committed logs are recovered during start-up, re-apply all of them. (Notice that operation sequences in IoTDB are idempotent)

Great suggestion, I think the 1-3 items you mentioned is to make sure that when do snapshot, new logs can not be added and the snapshot task should not take long time.  however now the implementation can block the new logs coming in, as when do snapshot we should get the _logManager_ lock, so new logs can not be committed(commited log also  need to get the  _logManager_ lock ), we just need to wait all the previous committed log applied when do snapshot.

I'm going to rethink the 4-5 suggestions.

cluster/src/main/java/org/apache/iotdb/cluster/log/manage/FilePartitionedSnapshotLogManager.java
```diff
@@ -47,22 +52,47 @@
...
+      logger.info("{}: no need to flush processor", getName());
+      return;
+    }
+    List<Path> storageGroups = new ArrayList<>();
+    for (Map.Entry<String, StorageGroupProcessor> entry : processorMap.entrySet()) {
+      Path path = new Path(entry.getKey());
+      storageGroups.add(path);
+    }
+    FlushPlan plan = new FlushPlan(null, true, storageGroups);
+    dataGroupMember.flushFileWhenDoSnapshot(plan);
```
jt2594838, line 76: I am afraid that we cannot flush the whole storage group when time partitioning is enabled. Because in that case, a storage group will be managed by several data groups, if you flush one storage group without notifying other data groups, the file integrity of other data groups will be broken.
So you should either flush other related groups (which is rather hard to find), or only flush partitions that belong to the data group.

jt2594838 (changes requested):

cluster/src/main/java/org/apache/iotdb/cluster/log/manage/FilePartitionedSnapshotLogManager.java
neuyilan: Sure, thanks  for your advice, I'd like to only flush the partitions that belong to the data group

jt2594838 (approved):

PR #1655, opened 2020-08-28
Fix sync bug

fanhualta:
In some scenario, the thrift put some bytes in the ByteBuffer, so record its position and reset it could avoid this error.


jixuan1989:
You'd better describe what this PR is for...

qiaojialin (approved):

PR #1656, opened 2020-08-28
Add raw data query Interface in Session 

haimeiguo:
Added raw data query interface in session. When user specifies the query as raw data query, the system will process it directly and skip the sql parser. 

jira: https://issues.apache.org/jira/browse/IOTDB-670


server/src/main/java/org/apache/iotdb/db/qp/Planner.java
```diff
@@ -65,6 +72,46 @@ public PhysicalPlan parseSQLToPhysicalPlan(String sqlStr, ZoneId zoneId)
     return physicalGenerator.transformToPhysicalPlan(operator);
   }
 
+  /**
+   * convert raw data query to physical plan directly
+   */
+  public PhysicalPlan rawDataQueryToPhysicalPlan(List<String> paths, long startTime, long endTime)
```
mychaow, line 78: rawDataQueryToPhysicalPlan  ----> rawDataQueryReqToPhysicalPlan
and use the class rawDataQueryReq as parameter, maybe we will support more complicated query req

session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -913,6 +914,27 @@ public SessionDataSet executeQueryStatement(String sql)
         execResp.isIgnoreTimeStamp());
   }
 
+  public SessionDataSet executeRawDataQuery(List<String> paths, long startTime, long endTime)
```
mychaow, line 917: add comments,  like the query is select * from paths where time >= startTime and time < endTime, time interval include startTime and exclude endTime

server/src/main/java/org/apache/iotdb/db/qp/Planner.java
```diff
@@ -65,6 +72,46 @@ public PhysicalPlan parseSQLToPhysicalPlan(String sqlStr, ZoneId zoneId)
...
+      fromOp.addPrefixTablePath(path);
+    }
+    selectOp.addSelectPath(new Path(""));
+
+    queryOp.setSelectOperator(selectOp);
+    queryOp.setFromOperator(fromOp);
+
+    //set time filter operator
+    FilterOperator filterOp = new FilterOperator(SQLConstant.KW_AND);
+    Path timePath = new Path("time");
```
mychaow, line 97: "time" should be defined as a constant string in IoTDBConsant

session/src/main/java/org/apache/iotdb/session/Session.java
mychaow: and change all other "time" in the code to the constant string, like in BasicOperatorType

server/src/main/java/org/apache/iotdb/db/qp/Planner.java
haimeiguo: Got it, thank you Chao for your comments!

thrift/src/main/thrift/rpc.thrift
```diff
@@ -222,6 +222,16 @@ struct TSDeleteDataReq {
...
 }
 
+struct TSRawDataQueryReq {
+    1: required i64 sessionId
+    2: required list<string> paths
+    3: optional i32 fetchSize
+    4: required i64 startTime
+    5: required i64 endTime
+    6: required i64 statementId
+    7: required string statement
```
neuyilan, line 232: What is the statement used for? can you give some examples?

session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -913,6 +914,27 @@ public SessionDataSet executeQueryStatement(String sql)
...
+          throws StatementExecutionException, IoTDBConnectionException {
+
+    String statement = "raw data query";
+    TSRawDataQueryReq execReq = new TSRawDataQueryReq(sessionId, paths, startTime, endTime, statementId, "");
```
neuyilan, line 921: as the statement parameter is empty, why we need the parameter?

session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -913,6 +914,27 @@ public SessionDataSet executeQueryStatement(String sql)
         execResp.isIgnoreTimeStamp());
   }
 
+  public SessionDataSet executeRawDataQuery(List<String> paths, long startTime, long endTime)
+          throws StatementExecutionException, IoTDBConnectionException {
+
+    String statement = "raw data query";
```
neuyilan, line 920: please make the statement is a constant field to avoid too many local variables to created and GC.

session/src/main/java/org/apache/iotdb/session/Session.java
haimeiguo: Hi Houliang, thank you for pointing out. I will pass the statement string to parameter. We will need statement and statementId in closeSession() when executing multiple queries.

session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -83,6 +83,7 @@
   private ZoneId zoneId;
   private long statementId;
   private int fetchSize;
+  private String rawDataQuery = "raw data query";
```
mychaow, line 86: use final to define it as constant, and move the string to other utils class, you may need other query strings later.

server/src/main/java/org/apache/iotdb/db/qp/Planner.java
```diff
@@ -94,7 +100,7 @@ public PhysicalPlan rawDataQueryToPhysicalPlan(List<String> paths, long startTim
 
     //set time filter operator
     FilterOperator filterOp = new FilterOperator(SQLConstant.KW_AND);
-    Path timePath = new Path("time");
+    Path timePath = new Path(TIME);
```
mychaow, line 103: BasicOperatorType also has the raw string "time", it's too ugly, you could replace all this "time" by constant TIME

mychaow (approved):

example/session/src/main/java/org/apache/iotdb/SessionExample.java
```diff
@@ -352,6 +353,24 @@ private static void query() throws IoTDBConnectionException, StatementExecutionE
...
+    paths.add("root.sg1.d1.s1");
+    paths.add("root.sg1.d1.s2");
+    paths.add("root.sg1.d1.s3");
+    long startTime = 10L;
+    long endTime = 200L;
+
+    SessionDataSet dataSet;
+    dataSet = session.executeRawDataQuery(paths, startTime, endTime);
+    System.out.println(dataSet.getColumnNames());
+    dataSet.setFetchSize(1024); // default is 10000
```
neuyilan, line 367: It's better not comment on the end of the line

neuyilan (approved):

example/session/src/main/java/org/apache/iotdb/SessionExample.java
haimeiguo: removed. Thank you!

session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -913,6 +914,38 @@ public SessionDataSet executeQueryStatement(String sql)
...
+   * @param endTime excluded
+   * @return
+   * @throws StatementExecutionException
+   * @throws IoTDBConnectionException
+   */
+
+  public SessionDataSet executeRawDataQuery(List<String> paths, long startTime, long endTime)
+          throws StatementExecutionException, IoTDBConnectionException {
+
+    String statement = SessionUtils.rawDataQuery;
```
qiaojialin, line 931: In JDBC or executeQueryStatement in Session, we send a SQL to the server, the SQL is used in a web metric(SqlArgument), which is disabled by default. The SQL is not important.

We could remove the statement field in the rpc. if we need it, we could reconstruct a sql in the server based on other parameters.

session/src/main/java/org/apache/iotdb/session/SessionUtils.java
```diff
@@ -28,6 +28,8 @@
 
 public class SessionUtils {
 
+  public static final String rawDataQuery = "raw data query";
```
qiaojialin, line 31: this could be removed

thrift/src/main/thrift/rpc.thrift
```diff
@@ -238,6 +238,16 @@ struct TSDeleteDataReq {
     4: required i64 endTime
 }
 
+struct TSRawDataQueryReq {
+    1: required i64 sessionId
+    2: required list<string> paths
+    3: optional i32 fetchSize
+    4: required i64 startTime
+    5: required i64 endTime
+    6: required i64 statementId
```
qiaojialin, line 247: The statementId is generated by the server, not the client. So this field is not needed.

You could refer to generateQueryId(boolean isDataQuery) in TSServiceImpl.

session/src/main/java/org/apache/iotdb/session/Session.java
haimeiguo: Got it. Thank you Jialin for your comments!

thrift/src/main/thrift/rpc.thrift
mychaow: we need statementId to record queryId and release resource when close session.

session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -913,6 +914,37 @@ public SessionDataSet executeQueryStatement(String sql)
...
+
+    TSExecuteStatementResp execResp;
+    try {
+      execResp = client.executeRawDataQuery(execReq);
+    } catch (TException e) {
+      throw new IoTDBConnectionException(e);
+    }
+
+    RpcUtils.verifySuccess(execResp.getStatus());
+    return new SessionDataSet("", execResp.getColumns(), execResp.getDataTypeList(),
```
qiaojialin, line 942: It seems that the statement(sql) in SessionDataSet is not needed. We could remove the statement field in TSFetchResultsReq in rpc.thrift.

mychaow: > It seems that the statement(sql) in SessionDataSet is not needed. We could remove the statement field in TSFetchResultsReq in rpc.thrift.

we should not remove the field, because TSFetchResults has the field in the previous release 0.10. User may use old client to query data from new release, we should keep our system backward compatible.

qiaojialin: It's ok, but it seems that the client-0.10 can not connect to server-0.11 already...

PR #1661, opened 2020-08-30
- Made it possible to load the config from the classpath

chrisdutz:
By using the URL class, we can not only use Files as config source, but also classpath resources.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -256,6 +256,11 @@
                 <version>1.2.17</version>
             </dependency>
             <dependency>
+                <groupId>net.jpountz.lz4</groupId>
+                <artifactId>lz4</artifactId>
+                <version>1.3.0</version>
+            </dependency>
+            <dependency>
                 <groupId>org.eclipse.jetty</groupId>
                 <artifactId>jetty-server</artifactId>
                 <version>${jetty.version}</version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -255,6 +255,11 @@
                 <artifactId>log4j</artifactId>
                 <version>1.2.17</version>
             </dependency>
+            <dependency>
```
jixuan1989, line 258: Why ... add LZ4 dependency here?

jixuan1989 (commented):
The Test on WinOS failed... need to check why.
maybe because the current Integration Test mechanism can not delete the correct folders after this change.

pom.xml
chrisdutz: I couldn't build IoTDB due to an error of the dependency plugin ... whithout this there were two different version versions in the reactor.

chrisdutz:
That's strange ... I simply changed the way the InputStream is generated. If used with the previous config options, it should still have the same result.

chrisdutz:
But while working on this ... I did notice the tests sometimes created directories and resources outside the "target" directory. This is considered not good practice as a failing test will prevent a cleanup ... in my case I couldn't do a new build till I found them all as otherwise the RAT test would have failed the build.

jixuan1989:
> But while working on this ... I did notice the tests sometimes created directories and resources outside the "target" directory. This is considered not good practice as a failing test will prevent a cleanup ... in my case I couldn't do a new build till I found them all as otherwise the RAT test would have failed the build.

Agree with you. I will do the refactor in the following days.

pom.xml
jixuan1989: OK, I checked the dependency tree, spark-core uses LZ4 v1.4  while others (e.g., tsfile) use v1.3

jixuan1989 (approved):

PR #1662, opened 2020-08-31
[IOTDB-837] ArrayIndexOutOfBoundsException if the measurementId size is not consistent with the value size

wyh-ztf:
ArrayIndexOutOfBoundsException if the measurementId size is not consistent with the value size

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/qp/strategy/PhysicalGenerator.java
+++ b/server/src/main/java/org/apache/iotdb/db/qp/strategy/PhysicalGenerator.java
@@ -31,6 +31,7 @@ import org.apache.iotdb.db.exception.metadata.MetadataException;
 import org.apache.iotdb.db.exception.query.LogicalOperatorException;
 import org.apache.iotdb.db.exception.query.LogicalOptimizeException;
 import org.apache.iotdb.db.exception.query.QueryProcessException;
+import org.apache.iotdb.db.exception.runtime.SQLParserException;
 import org.apache.iotdb.db.metadata.PartialPath;
 import org.apache.iotdb.db.qp.constant.SQLConstant;
 import org.apache.iotdb.db.qp.logical.Operator;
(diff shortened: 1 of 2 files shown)


neuyilan (commented):
Please make the title clearer.

server/src/main/java/org/apache/iotdb/db/qp/strategy/PhysicalGenerator.java
```diff
@@ -176,6 +176,13 @@ public PhysicalPlan transformToPhysicalPlan(Operator operator) throws QueryProce
               "For Insert command, cannot specified more than one seriesPath: " + paths);
         }
```
qiaojialin, line 177: This check could also be removed. I do not know how to trigger this. The sql parser should have already checked.

server/src/main/java/org/apache/iotdb/db/qp/strategy/PhysicalGenerator.java
```diff
@@ -176,6 +176,13 @@ public PhysicalPlan transformToPhysicalPlan(Operator operator) throws QueryProce
...
 
+        if (insert.getMeasurementList().length != insert.getValueList().length) {
+          throw new ArrayIndexOutOfBoundsException(
```
qiaojialin, line 180: You could throw an SQLParserException.

qiaojialin (commented):
Good job!

server/src/main/java/org/apache/iotdb/db/qp/strategy/PhysicalGenerator.java
wyh-ztf: Thank you for your opinion. I will revise it

qiaojialin (commented):
I forget one thing. For each bug fix, please add a test to verify it

neuyilan (approved):

wyh-ztf:
> I forget one thing. For each bug fix, please add a test to verify it

ok,i will add a test to verify it

qiaojialin (approved):

qiaojialin:
Hi, we usually use square brackets [] around the issue number in the title

wyh-ztf:
> Hi, we usually use square brackets [] around the issue number in the title

OK, I'll pay attention next time. Thanks for your help

PR #1663, opened 2020-08-31
[IOTDB-855] Put back SessionPool resources if RuntimeException occurs

wyh-ztf:


jixuan1989:
looks fine.
But my only concern is that is there some runtimeError breaking the season down? (I do not know whether it may happen. But if so, put the session back to the pool is not a good idea).

qiaojialin (approved):

wyh-ztf:
> down

There are two types of running errors that I have found. They are illegargumentexception and indexoutofboundsexception. I have tested them. After catching these two kinds of running errors, the session is still robust

wyh-ztf:
> looks fine.
> But my only concern is that is there some runtimeError breaking the season down? (I do not know whether it may happen. But if so, put the session back to the pool is not a good idea).

There are two types of running errors that I have found. They are illegargumentexception and indexoutofboundsexception. I have tested them. After catching these two kinds of running errors, the session is still robust

PR #1665, opened 2020-09-01
IOTDB-851: enhance failure tolerance when recover WAL

haimeiguo:
WAL will skip error series instead of all series in the device

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/writelog/recover/LogReplayer.java
+++ b/server/src/main/java/org/apache/iotdb/db/writelog/recover/LogReplayer.java
@@ -164,16 +164,14 @@ public class LogReplayer {
     }
     if (plan instanceof InsertRowPlan) {
       InsertRowPlan tPlan = (InsertRowPlan) plan;
+      ((InsertRowPlan) plan).setNeedInferType(false);
       tPlan.setSchemasAndTransferType(schemas);
+      checkDataTypeAndMarkFailed(schemas, tPlan);
       recoverMemTable.insert(tPlan);
     } else {
       InsertTabletPlan tPlan = (InsertTabletPlan) plan;
       tPlan.setSchemas(schemas);
-      for (int i = 0; i < schemas.length; i++) {
-        if (schemas[i].getType() != tPlan.getDataTypes()[i]) {
-          tPlan.markFailedMeasurementInsertion(i);
-        }
-      }
+      checkDataTypeAndMarkFailed(schemas, tPlan);
       recoverMemTable.insertTablet(tPlan, 0, tPlan.getRowCount());
     }
   }
@@ -182,5 +180,13 @@ public class LogReplayer {
   private void replayUpdate(UpdatePlan updatePlan) {
     // TODO: support update
     throw new UnsupportedOperationException("Update not supported");
+  }
+
+  private void checkDataTypeAndMarkFailed(final MeasurementSchema[] schemas, InsertPlan tPlan) {
+    for (int i = 0; i < schemas.length; i++) {
+      if (schemas[i] == null || schemas[i].getType() != tPlan.getDataTypes()[i]) {
+        tPlan.markFailedMeasurementInsertion(i);
+      }
+    }
   }
 }
--- a/server/src/main/java/org/apache/iotdb/db/writelog/recover/LogReplayer.java
+++ b/server/src/main/java/org/apache/iotdb/db/writelog/recover/LogReplayer.java
@@ -164,8 +164,11 @@ public class LogReplayer {
     }
     if (plan instanceof InsertRowPlan) {
       InsertRowPlan tPlan = (InsertRowPlan) plan;
+      //only inferTyp when users pass a String value
+      //WAL already serializes the real data type, so no need to infer typ
       ((InsertRowPlan) plan).setNeedInferType(false);
       tPlan.setSchemasAndTransferType(schemas);
+      //mark failed plan manually 
       checkDataTypeAndMarkFailed(schemas, tPlan);
       recoverMemTable.insert(tPlan);
     } else {
--- a/server/src/main/java/org/apache/iotdb/db/writelog/recover/LogReplayer.java
+++ b/server/src/main/java/org/apache/iotdb/db/writelog/recover/LogReplayer.java
@@ -169,6 +169,11 @@ public class LogReplayer {
     } else {
       InsertTabletPlan tPlan = (InsertTabletPlan) plan;
       tPlan.setSchemas(schemas);
+      for (int i = 0; i < schemas.length; i++) {
+        if (schemas[i].getType() != tPlan.getDataTypes()[i]) {
+          tPlan.markFailedMeasurementInsertion(i);
+        }
+      }
       recoverMemTable.insertTablet(tPlan, 0, tPlan.getRowCount());
     }
   }
--- a/server/src/main/java/org/apache/iotdb/db/writelog/recover/LogReplayer.java
+++ b/server/src/main/java/org/apache/iotdb/db/writelog/recover/LogReplayer.java
@@ -164,8 +164,8 @@ public class LogReplayer {
     }
     if (plan instanceof InsertRowPlan) {
       InsertRowPlan tPlan = (InsertRowPlan) plan;
-      //only inferTyp when users pass a String value
-      //WAL already serializes the real data type, so no need to infer typ
+      //only infer type when users pass a String value
+      //WAL already serializes the real data type, so no need to infer type
       ((InsertRowPlan) plan).setNeedInferType(false);
       tPlan.setSchemasAndTransferType(schemas);
       //mark failed plan manually 
--- a/server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
@@ -686,10 +686,15 @@ public class MManager {
       MeasurementSchema[] measurementSchemas = new MeasurementSchema[measurements.length];
       for (int i = 0; i < measurementSchemas.length; i++) {
         if (!deviceNode.hasChild(measurements[i])) {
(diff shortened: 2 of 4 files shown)


mychaow (approved):

qiaojialin (commented):
Good work~ Since there are some historical reason and more scenarios, I would like to offer more information.

First, there are two bug scenarios:

(1) 
insert into root.turbine1.d1(timestamp,s1,s2) values(1,1.1,2.2);
delete timeseries root.turbine1.d1.s1
create timeseries root.turbine1.d1.s1 with datatype=INT32, encoding=RLE, compression=SNAPPY
restart IoTDB
select * from root

The InsertRowPlan that deserialized from WAL should set isNeedInserfType as false. We only inferTyp when users pass a String value, the WAL already serializes the real data type, so no need to infer type. So, we need to markFailedMeasurementInsertion manually for insertRowPlan.

(2) 
insert into root.turbine1.d1(timestamp,s1,s2) values(1,1.1,2.2);
delete timeseries root.turbine1.d1.s1
restart IoTDB
select * from root

This will throw a metadata exception when querying the schema of the InsertPlan, because a measurement is deleted in MManager.

One way to fix this is to add a parameter of IoTDB.metaManager.getSchemas() indicating that it should return a MeasurementSchema[] with some nulls instead of throw exception when recovering wal. 

Finally, it's better to recover the two cases by an IT. There is an IoTDBRestartIT class, you could use EnvironmentUtils.restartDaemon()  to simulate the restart of IoTDB and verify the bugs.

qiaojialin:
Here is one case

@Test
  public void testRecoverWALMismatchDataType()
      throws Exception {
    EnvironmentUtils.envSetUp();
    Class.forName(Config.JDBC_DRIVER_NAME);

    try(Connection connection = DriverManager
        .getConnection(Config.IOTDB_URL_PREFIX + "127.0.0.1:6667/", "root",
            "root");
        Statement statement = connection.createStatement()){
      statement.execute("insert into root.turbine1.d1(timestamp,s1,s2) values(1,1.1,2.2);");
      statement.execute("delete timeseries root.turbine1.d1.s1;");
      statement.execute("create timeseries root.turbine1.d1.s1 with datatype=INT32, encoding=RLE, compression=SNAPPY;");
    }

    EnvironmentUtils.restartDaemon();


    try(Connection connection = DriverManager
        .getConnection(Config.IOTDB_URL_PREFIX + "127.0.0.1:6667/", "root",
            "root");
        Statement statement = connection.createStatement()){

      boolean hasResultSet = statement.execute("select * from root");
      assertTrue(hasResultSet);
      ResultSet resultSet = statement.getResultSet();
      int cnt = 0;
      while (resultSet.next()) {
        cnt++;
      }
      assertEquals(1, cnt);
    }

    EnvironmentUtils.cleanEnv();
  }

haimeiguo:
> Good work~ Since there are some historical reason and more scenarios, I would like to offer more information.
> 
> First, there are two bug scenarios:
> 
> (1)
> insert into root.turbine1.d1(timestamp,s1,s2) values(1,1.1,2.2);
> delete timeseries root.turbine1.d1.s1
> create timeseries root.turbine1.d1.s1 with datatype=INT32, encoding=RLE, compression=SNAPPY
> restart IoTDB
> select * from root
> 
> The InsertRowPlan that deserialized from WAL should set isNeedInserfType as false. We only inferTyp when users pass a String value, the WAL already serializes the real data type, so no need to infer type. So, we need to markFailedMeasurementInsertion manually for insertRowPlan.
> 
> (2)
> insert into root.turbine1.d1(timestamp,s1,s2) values(1,1.1,2.2);
> delete timeseries root.turbine1.d1.s1
> restart IoTDB
> select * from root
> 
> This will throw a metadata exception when querying the schema of the InsertPlan, because a measurement is deleted in MManager.
> 
> One way to fix this is to add a parameter of IoTDB.metaManager.getSchemas() indicating that it should return a MeasurementSchema[] with some nulls instead of throw exception when recovering wal.
> 
> Finally, it's better to recover the two cases by an IT. There is an IoTDBRestartIT class, you could use EnvironmentUtils.restartDaemon() to simulate the restart of IoTDB and verify the bugs.

haimeiguo:
> Here is one case
> 
> @test
> public void testRecoverWALMismatchDataType()
> throws Exception {
> EnvironmentUtils.envSetUp();
> Class.forName(Config.JDBC_DRIVER_NAME);
> 
> ```
> try(Connection connection = DriverManager
>     .getConnection(Config.IOTDB_URL_PREFIX + "127.0.0.1:6667/", "root",
>         "root");
>     Statement statement = connection.createStatement()){
>   statement.execute("insert into root.turbine1.d1(timestamp,s1,s2) values(1,1.1,2.2);");
>   statement.execute("delete timeseries root.turbine1.d1.s1;");
>   statement.execute("create timeseries root.turbine1.d1.s1 with datatype=INT32, encoding=RLE, compression=SNAPPY;");
> }
> 
> EnvironmentUtils.restartDaemon();
> 
> 
> try(Connection connection = DriverManager
>     .getConnection(Config.IOTDB_URL_PREFIX + "127.0.0.1:6667/", "root",
>         "root");
>     Statement statement = connection.createStatement()){
> 
>   boolean hasResultSet = statement.execute("select * from root");
>   assertTrue(hasResultSet);
>   ResultSet resultSet = statement.getResultSet();
>   int cnt = 0;
>   while (resultSet.next()) {
>     cnt++;
>   }
>   assertEquals(1, cnt);
> }
> 
> EnvironmentUtils.cleanEnv();
> ```
> 
> }



> Good work~ Since there are some historical reason and more scenarios, I would like to offer more information.
> 
> First, there are two bug scenarios:
> 
> (1)
> insert into root.turbine1.d1(timestamp,s1,s2) values(1,1.1,2.2);
> delete timeseries root.turbine1.d1.s1
> create timeseries root.turbine1.d1.s1 with datatype=INT32, encoding=RLE, compression=SNAPPY
> restart IoTDB
> select * from root
> 
> The InsertRowPlan that deserialized from WAL should set isNeedInserfType as false. We only inferTyp when users pass a String value, the WAL already serializes the real data type, so no need to infer type. So, we need to markFailedMeasurementInsertion manually for insertRowPlan.
> 
> (2)
> insert into root.turbine1.d1(timestamp,s1,s2) values(1,1.1,2.2);
> delete timeseries root.turbine1.d1.s1
> restart IoTDB
> select * from root
> 
> This will throw a metadata exception when querying the schema of the InsertPlan, because a measurement is deleted in MManager.
> 
> One way to fix this is to add a parameter of IoTDB.metaManager.getSchemas() indicating that it should return a MeasurementSchema[] with some nulls instead of throw exception when recovering wal.
> 
> Finally, it's better to recover the two cases by an IT. There is an IoTDBRestartIT class, you could use EnvironmentUtils.restartDaemon() to simulate the restart of IoTDB and verify the bugs.

Hi Jialin, 

Thank you so much for your offering and test cases. I will look into it:)

server/src/main/java/org/apache/iotdb/db/writelog/recover/LogReplayer.java
```diff
@@ -164,11 +164,14 @@ private void replayInsert(InsertPlan plan) throws WriteProcessException, QueryPr
     }
     if (plan instanceof InsertRowPlan) {
       InsertRowPlan tPlan = (InsertRowPlan) plan;
+      ((InsertRowPlan) plan).setNeedInferType(false);
```
mychaow, line 167: add comment to explain the reason

mychaow (approved):

qiaojialin (approved):
great! just one minor issue

server/src/main/java/org/apache/iotdb/db/writelog/recover/LogReplayer.java
mychaow: I think it will make others confused in the deserialization of InsertRowPlan. Cause that we should gurantee that the result of serialization and deserialization is the same, this set operation should be better in the here.

qiaojialin: +1

mychaow: > Please put the ((InsertRowPlan) plan).setNeedInferType(false); into the deserialization of InsertRowPlan

I think we should make the two function serialize(ByteBuffer) and Serialize(DataOutputStream)  have same action, so it's not a good way to setNeedInferType in deserialization.

server/src/main/java/org/apache/iotdb/db/writelog/recover/LogReplayer.java
```diff
@@ -164,11 +164,17 @@ private void replayInsert(InsertPlan plan) throws WriteProcessException, QueryPr
     }
     if (plan instanceof InsertRowPlan) {
       InsertRowPlan tPlan = (InsertRowPlan) plan;
+      //only infer type when users pass a String value
+      //WAL already serializes the real data type, so no need to infer type
+      ((InsertRowPlan) plan).setNeedInferType(false);
       tPlan.setSchemasAndTransferType(schemas);
+      //mark failed plan manually 
+      checkDataTypeAndMarkFailed(schemas, tPlan);
```
jixuan1989, line 172: seems `setSchemasAndTransferType ` has marked the related data point failed already.

jixuan1989 (changes requested):

server/src/main/java/org/apache/iotdb/db/writelog/recover/LogReplayer.java
haimeiguo: Hi thank you for your review! 
Here i set plan.setNeedInferType(false)
in setSchemasAndTransferType(schemas), it will only execute this.schemas = schemas
it will not mark the data points as failed

jixuan1989: I see.

jixuan1989 (approved):

PR #1694, opened 2020-09-06
[IOTDB-813] Show storage group under given path prefix

yhwang-hbl:


server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
```diff
@@ -502,6 +502,54 @@ private void findStorageGroup(
...
+    }
+    String nodeReg = MetaUtils.getNodeRegByIdx(idx, nodes);
+    if (!(PATH_WILDCARD).equals(nodeReg)) {
+      if (node.hasChild(nodeReg)) {
+        findStorageGroupPaths(
+            node.getChild(nodeReg),
+            nodes,
+            idx + 1,
+            parent + node.getName() + PATH_SEPARATOR,
+            storageGroupPaths);
```
qiaojialin, line 543: Hi, please put these into one or two lines.

qiaojialin (approved):
Good work!

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
yhwang-hbl: Copy that. I'll change it

PR #1714, opened 2020-09-09
[IOTDB-869] Failed to insert data via Windows CLI

wshao08:
JIRA issue: https://issues.apache.org/jira/browse/IOTDB-869

This issue occurs because IoTDB failed to load "iotdb-engine.properties" when starting on Windows.
When the file path is in the format of "D:/path" on WinOS, it will not be recognized as valid protocol when creating a URL object.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/conf/IoTDBDescriptor.java
+++ b/server/src/main/java/org/apache/iotdb/db/conf/IoTDBDescriptor.java
@@ -123,14 +123,10 @@ public class IoTDBDescriptor {
     // If the url doesn't start with "file:" or other protocols, it's provided as a normal path.
     // If valid, we return a URL converted from this file path.
     File confFile = new File(urlString);
-    if (confFile.exists()) {
-      try {
-        return confFile.toURI().toURL();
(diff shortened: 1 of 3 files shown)


jixuan1989 (approved):

server/src/main/java/org/apache/iotdb/db/conf/IoTDBDescriptor.java
```diff
@@ -118,12 +119,14 @@ public URL getPropsUrl() {
...
-      urlString = "file:" + urlString;
-    }
+
+    // If the url doesn't start with "file:" or other protocols, it's provided as a normal path.
+    // If valid, we return a URL converted from this file path.
+    File confFile = new File(urlString);
     try {
+      if (confFile.exists()) {
+        return confFile.toURI().toURL();
+      }
```
chrisdutz, line 129: This change will again break the ability to load a config properites resource from classpath ...

chrisdutz (changes requested):

chrisdutz (approved):
LGTM, even if the other parts don't really have an effect on my issue.

PR #1717, opened 2020-09-09
[IOTDB-881] Fix Double quotes bug in Align by device

Alima777:
This pr fixes:

1. **IOTDB-881 bug:** Double quoted meaurement is recognized as constant in Align by device. Only single quoted measurement should be recognized as constant after double quotes can be used in path.

2. Modify splitPathToDetachedPath() method: support nested double quotes now.

3. Extract some logic in PhysicalGenerator transfrom() method.


Genius-pig:
Can you add support nested double quotes in `Path`?

qiaojialin (commented):
Update the user guide

Alima777:
> Can you add support nested double quotes in `Path`?

Hi, thank you for reminding.

Alima777:
> Update the user guide

Updated.

Genius-pig (approved):

qiaojialin (approved):
Could the split method in MetaUtils and Path be combined?

Issue #1726, opened 2020-09-11
Refactor `SingleFileLogReader` iteraotor logic


yuqi1129:
Currently `SingleFileLogReader#hasNext` is not very efficiency and clear, we can simplify this mothod

jixuan1989:
Hi yuqi, any ideas (and want to have a try)? :D

yuqi1129:
Yes, I'am working on it

PR #1727, opened 2020-09-11
[IOTDB-888] Refactor `SingleFileLogReader` iterator logic

yuqi1129:


jixuan1989:
thanks @yuqi1129 .
issue #1726  @jt2594838 , can you have a review?

server/src/main/java/org/apache/iotdb/db/writelog/io/SingleFileLogReader.java
```diff
@@ -53,54 +54,69 @@
...
-      if (logStream.available() < LEAST_LOG_SIZE) {
-        return false;
-      }
 
-      int logSize = logStream.readInt();
-      if (logSize <= 0) {
-        return false;
-      }
-      buffer = new byte[logSize];
+    if (Objects.nonNull(batchLogReader)) {
```
jixuan1989, line 66: IMO,   keep `if (batchLogReader != null && batchLogReader.hasNext())` is better...

server/src/main/java/org/apache/iotdb/db/writelog/io/SingleFileLogReader.java
```diff
@@ -53,54 +54,69 @@
...
 
-      final long checkSum = logStream.readLong();
-      checkSummer.reset();
-      checkSummer.update(buffer, 0, logSize);
-      if (checkSummer.getValue() != checkSum) {
-        throw new IOException(String.format("The check sum of the No.%d log batch is incorrect! In "
-            + "file: "
-            + "%d Calculated: %d.", idx, checkSum, checkSummer.getValue()));
+    try {
+      while (logStream.available() > LEAST_LOG_SIZE) {
```
jixuan1989, line 75: why add `while` here?

jixuan1989 (changes requested):
let me try to understand what you want to improve..
the main idea is: do not re-check hasNext() when calling next(), right?

server/src/main/java/org/apache/iotdb/db/writelog/io/SingleFileLogReader.java
```diff
@@ -53,54 +54,69 @@
...
-    return true;
+
+    hasNext = false;
+    return hasNext;
   }
 
   @Override
   public PhysicalPlan next() {
-    if (!hasNext()){
+    if (!hasNext){
```
jt2594838, line 119: `next()` in an iterator should not depend on `hasNext()`, that is, the caller should be able to iterate without calling `hasNext()`. While in your implementation, if the caller calls `next()` without calling `hasNext()`, it will get nothing at all.

And I do not understand other code changes in this PR, could you please explain their improvements to me?

jt2594838 (changes requested):

server/src/main/java/org/apache/iotdb/db/writelog/io/SingleFileLogReader.java
yuqi1129: You are right,  we should invoke `hasNext()` when we call next(), i change the logic here.

server/src/main/java/org/apache/iotdb/db/writelog/io/SingleFileLogReader.java
yuqi1129: if `batchLogReader ` is null, then we should find a position in logStream where we create BatchLogReader and make `hasNext()` is true repeatedly, it's in fact a loop

server/src/main/java/org/apache/iotdb/db/writelog/io/SingleFileLogReader.java
yuqi1129: Yes,  if `hasNext()` should be called, there is not need to make any changes here

server/src/main/java/org/apache/iotdb/db/writelog/io/SingleFileLogReader.java
jt2594838: I wonder how could we get a new BatchLogReader but it has nothing to read and yet it is not corrupted. If there is indeed such a situation, please add associated unit test.

jt2594838 (changes requested):

jixuan1989:
ping.  This PR needs maintainers.

PR #1731, opened 2020-09-14
[IOTDB-870] tags and attributes output

haimeiguo:
tags and attributes in result dataset are in json format


server/src/main/java/org/apache/iotdb/db/query/dataset/ShowTimeSeriesResult.java
```diff
@@ -134,13 +151,25 @@ public static ShowTimeSeriesResult deserialize(ByteBuffer buffer) {
     result.encoding = ReadWriteIOUtils.readString(buffer);
     result.compressor = ReadWriteIOUtils.readString(buffer);
```
liutaohua, line 152: we have the enums of these strings, I suggest that we use the enums to serialize, u can find it in :
```
package org.apache.iotdb.tsfile.file.metadata.enums;
```

server/src/main/java/org/apache/iotdb/db/utils/QueryUtils.java
```diff
@@ -127,12 +128,27 @@ public static void constructPathAndDataTypes(List<PartialPath> paths, List<TSDat
     paths.add(new PartialPath(COLUMN_TIMESERIES_COMPRESSION, false));
     dataTypes.add(TSDataType.TEXT);
```
liutaohua, line 129: duplicate code, how about:
```
  private static PartialPath[] resourcePaths = new PartialPath[]{new PartialPath(COLUMN_TIMESERIES, false), .....};
  private static TSDataType[] resourceTypes = new TSDataType[]{TSDataType.TEXT, .....};

  public static void constructPathAndDataTypes(List<PartialPath> paths, List<TSDataType> dataTypes, List<ShowTimeSeriesResult> timeseriesList) {
    Collections.addAll(paths,resourcePaths);
    Collections.addAll(dataTypes,resourceTypes);
    ......
  }
```

server/src/main/java/org/apache/iotdb/db/utils/QueryUtils.java
```diff
@@ -127,12 +128,27 @@ public static void constructPathAndDataTypes(List<PartialPath> paths, List<TSDat
...
 
-    Set<String> tagAndAttributeName = new TreeSet<>();
+    //check if timeseries result has tag or attribute
+    boolean hasTag = false;
+    boolean hasAttribute = false;
     for (ShowTimeSeriesResult result : timeseriesList) {
```
liutaohua, line 134: I don't think there's any need to check,  it's ok to show `null || spaces`, like :
```
+--------------+-----+-------------+--------+--------+-----------+----+----------+
|    timeseries|alias|storage group|dataType|encoding|compression|tags|attributes|
+--------------+-----+-------------+--------+--------+-----------+----+----------+
|root.sg1.d1.s1| null|     root.sg1|   INT64|     RLE|     SNAPPY|null|      null|
|root.sg1.d1.s2| null|     root.sg1|   INT64|     RLE|     SNAPPY|null|      null|
|root.sg1.d1.s3| null|     root.sg1|   INT64|     RLE|     SNAPPY|null|      null|
+--------------+-----+-------------+--------+--------+-----------+----+----------+

```

liutaohua (commented):
nice work , but I have some suggestions that I hope you'll take

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -994,14 +993,13 @@ private boolean match(PartialPath fullPath, String[] prefixNodes) {
...
-                ansString.right[3], ansString.right[4], Collections.emptyMap()));
+                ansString.right[3], ansString.right[4], Collections.emptyMap(), Collections.emptyMap()));
           } else {
             // has tags/attributes
             Pair<Map<String, String>, Map<String, String>> pair =
                 tagLogFile.read(config.getTagAttributeTotalSize(), tagFileOffset);
-            pair.left.putAll(pair.right);
             res.add(new ShowTimeSeriesResult(ansString.left.getFullPath(), ansString.right[0], ansString.right[1], ansString.right[2],
-                ansString.right[3], ansString.right[4], pair.left));
+                ansString.right[3], ansString.right[4], pair.left, pair.right));
```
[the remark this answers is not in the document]
haimeiguo: edited. thank you :)

server/src/main/java/org/apache/iotdb/db/utils/QueryUtils.java
```diff
@@ -152,10 +142,11 @@ public static QueryDataSet getQueryDataSet(List<ShowTimeSeriesResult> timeseries
       updateRecord(record, result.getName());
       updateRecord(record, result.getAlias());
       updateRecord(record, result.getSgName());
-      updateRecord(record, result.getDataType());
-      updateRecord(record, result.getEncoding());
-      updateRecord(record, result.getCompressor());
-      updateRecord(record, result.getTagAndAttribute(), paths);
+      updateRecord(record, result.getDataType().toString());
+      updateRecord(record, result.getEncoding().toString());
+      updateRecord(record, result.getCompressor().toString());
```
liutaohua, line 147: i think that too many type cast , how about used only in `serialize/deserialize`.

server/src/main/java/org/apache/iotdb/db/utils/QueryUtils.java
```diff
@@ -166,26 +157,28 @@ public static QueryDataSet getQueryDataSet(List<ShowTimeSeriesResult> timeseries
...
```
liutaohua, line 168: same

server/src/main/java/org/apache/iotdb/db/utils/QueryUtils.java
```diff
@@ -112,36 +122,16 @@ public static void filterQueryDataSource(QueryDataSource queryDataSource,
...
+    Collections.addAll(paths, resourcePaths);
+    Collections.addAll(dataTypes, resourceTypes);
   }
 
   public static QueryDataSet getQueryDataSet(List<ShowTimeSeriesResult> timeseriesList,
       ShowTimeSeriesPlan showTimeSeriesPlan, QueryContext context) {
     List<PartialPath> paths = new ArrayList<>();
     List<TSDataType> dataTypes = new ArrayList<>();
-    constructPathAndDataTypes(paths, dataTypes, timeseriesList);
+    constructPathAndDataTypes(paths, dataTypes);
```
liutaohua, line 134: I think that `constructPath` and `updateRecord` behavior belongs to `Dataset`, how about move it to `ShowTimeseriesDataSet ` class?

and next line ,  the `showTimeSeriesPlan` already exists in dataset...


@qiaojialin  what do you think?

liutaohua:
Don't forget merge _`main`_ branch

server/src/main/java/org/apache/iotdb/db/utils/QueryUtils.java
qiaojialin: It seems that the constructPathAndDataTypes is only used for ShowTimeseries, so moving it to ShowTimeseriesDataSet is ok.

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
```diff
@@ -566,7 +567,8 @@ private QueryDataSet processShowStorageGroup(ShowStorageGroupPlan showStorageGro
   private QueryDataSet processShowTimeseries(ShowTimeSeriesPlan showTimeSeriesPlan,
       QueryContext context) throws MetadataException {
     List<ShowTimeSeriesResult> timeseriesList = showTimeseries(showTimeSeriesPlan, context);
-    return QueryUtils.getQueryDataSet(timeseriesList, showTimeSeriesPlan, context);
+    ShowTimeseriesDataSet showTimeSeriesDataSet = new ShowTimeseriesDataSet(showTimeSeriesPlan, context);
+    return showTimeSeriesDataSet.getQueryDataSet(timeseriesList);
```
qiaojialin, line 571: ShowTimeseriesData should play as a query processor. 
It accepts a query plan and generates results.

server/src/main/java/org/apache/iotdb/db/query/dataset/ShowTimeseriesDataSet.java
```diff
@@ -19,44 +19,123 @@
...
+      return;
+    }
+    Field field = new Field(TSDataType.TEXT);
+    field.setBinaryV(new Binary(s));
+    record.addField(field);
   }
 
   @Override
   protected boolean hasNextWithoutConstraint() throws IOException {
+    if (!hasSetRecord) {
```
liutaohua, line 135: i think that this code can be optimizing:
1. Take the data directly from the `MManager` and initialize it in the `constructor`
2. if the total amount of data is less than the `fetchSize`, which means there is no more data, reducing the access to the `MManager`.

> when hasLimit = false, limit was means fetchSize

3. if `index == result.size()== fetchSize`, get data from `MManager` again.

server/src/main/java/org/apache/iotdb/db/query/dataset/ShowTimeseriesDataSet.java
```diff
@@ -19,44 +19,123 @@
...
+
+  private void updateRecord(RowRecord record, Map<String, String> map) {
+    String text = map.entrySet().stream()
+        .map(e -> "\"" + e.getKey() + "\"" + ":" + "\"" + e.getValue() + "\"")
+        .collect(Collectors.joining(","));
+
+    updateRecord(record, text.length() == 0 ? null : "{" + text + "}");
+  }
+
+  private void updateRecord(RowRecord record, String s) {
```
liutaohua, line 123: `putRecord` method should also be modified to `private`

liutaohua (commented):
victory is in sight

liutaohua:
LGTM

liutaohua (approved):

server/src/main/java/org/apache/iotdb/db/conf/IoTDBConstant.java
```diff
@@ -70,6 +70,8 @@ private IoTDBConstant() {
   public static final String COLUMN_DEVICES = "devices";
   public static final String COLUMN_COLUMN = "column";
   public static final String COLUMN_COUNT = "count";
+  public static final String COLUMN_TAG = "tags";
+  public static final String COLUMN_ATTRIBUTE = "attributes";
```
HTHou, line 74:
```suggestion
  public static final String COLUMN_TAGS = "tags";
  public static final String COLUMN_ATTRIBUTES = "attributes";
```

qiaojialin (approved):
Good work!

PR #1737, opened 2020-09-15
[IOTDB-897] Remove duplicate maven plugin in module tsfile and website

yuqi1129:


site/pom.xml
```diff
@@ -39,30 +39,6 @@
     </distributionManagement>
     <build>
         <plugins>
-            <!-- copy source codes into vue required folders-->
```
jixuan1989, line 42: Hi, why you thought this is duplicated...
have you tried `mvn package -pl site -P site` ?

hadoop/pom.xml
```diff
@@ -85,26 +85,6 @@
                     </archive>
                 </configuration>
             </plugin>
-            <plugin>
```
jixuan1989, line 88: this plugin is for generating `hadoop-tsfile-0.11.0-SNAPSHOT-jar-with-dependencies.jar` file, which includes all dependencies for using hdfs and tsfile.
If the user uses maven, indeed it is useless, but for users who are not use maven, I think it is useful (or someone who uses maven but there are version conflicts for hadoop dependencies).

So, I have no idea about whether to remove this plugin? either is ok.

What do you think?

hadoop/pom.xml
```diff
@@ -40,7 +40,7 @@
         <dependency>
             <groupId>org.apache.iotdb</groupId>
             <artifactId>tsfile</artifactId>
-            <version>0.11.0-SNAPSHOT</version>
```
jixuan1989, line 43: +1

jixuan1989 (commented):
have you tried to run on the modified codebase (especially for the site module)?

site/pom.xml
yuqi1129: You are right,  i mistake it

hadoop/pom.xml
yuqi1129: Ok, if `hadoop-tsfile-0.11.0-SNAPSHOT-jar-with-dependencies.jar` is needed somewhere, i do think there is no enough reason to remove it.

Thanks for your kindly review~

yuqi1129:
This pull pull is mark as invalid, close it

PR #1738, opened 2020-09-16
#1730 fix bug that cross-test generates data out of the target folder

jixuan1989:
PR #1730  did not fix the problem that cross-test generates data out of the target folder.

1. use iotdb-server:test-jar to import the EnvironmentUtils.java
2. put all configurations into test/resources

@mychaow


HTHou (approved):

qiaojialin (approved):

mychaow:
> PR #1730 did not fix the problem that cross-test generates data out of the target folder.
> 
> 1. use iotdb-server:test-jar to import the EnvironmentUtils.java
> 2. put all configurations into test/resources
> 
> @mychaow

Got it!

PR #1753, opened 2020-09-23
set props to null by default in createTimeseries to save memory

qiaojialin:


JackieTien97 (changes requested):
Actually, Collections.emptyMap() always return a same Constant object. I don't think it will save memory. Even if you set the reference to null, the reference itself will take 4B same as before. Overall, EMPTY_MAP is a constant value in Collections, it will be in the memory as long as you use the `Collections` class.

qiaojialin:
Thanks Yuan!

PR #1762, opened 2020-09-24
fix ImportCsvToolIT bug

mychaow:


HTHou (approved):

jixuan1989:
it is strange why the IT passed...
The first screenshot is on Linux and the second is on Win on Travis.
image
image

mychaow:
> it is strange why the IT passed...
> The first screenshot is on Linux and the second is on Win on Travis.
> image
> image

We run different script in  linux and windows, so the output is also difference in linux and windows.

PR #1764, opened 2020-09-25
Made the source plugin create source-jars even for SNAPSHOT builds

chrisdutz:
Right now we are not producing source jars for SNAPSHOT builds. So if you're using a SNAPSHOT version and trying to debug something, the IDE will have to use decompiling instead, which is a little unfortunate.

With this change the source jars are created for every build and not only the apache-release builds.


qiaojialin (approved):
Thank you, Chris!

jixuan1989:
Thanks, the master of maven. :D

PR #1768, opened 2020-09-27
[IOTDB-908] Implement ResultSetMetadata interface for JDBC

jack870131:
JIRA Issue: https://issues.apache.org/jira/browse/IOTDB-908

I have now implemented two methods, which are getColumnClassName() & getColumnTypeName(). Please have a review.

Btw, We still need some further discussions about how to deal with other methods such as getTableName() and getSchemaName(). What exactly are they specific to?


HTHou (approved):
LGTM
It will be better if you can add some UT.

PR #1778, opened 2020-09-29
remove some redundant code

haimeiguo:


Alima777 (commented):
I think the comments should be kept, which is not redundant... All else looks good.

haimeiguo:
> I think the comments should be kept, which is not redundant... All else looks good.

added. Thank you!

Alima777 (approved):

qiaojialin (approved):
Really clear!

PR #1780, opened 2020-09-29
[IOTDB-921] Solve the statement twice call bug

JackieTien97:


Alima777 (approved):

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBStatement.java
```diff
@@ -320,12 +320,12 @@ private ResultSet executeQuerySQL(String sql) throws TException, SQLException {
     if (execResp.queryDataSet == null) {
       this.resultSet = new IoTDBNonAlignJDBCResultSet(this, execResp.getColumns(),
           execResp.getDataTypeList(), execResp.columnNameIndexMap, execResp.ignoreTimeStamp, client, sql, queryId,
-          sessionId, execResp.nonAlignQueryDataSet);
+          sessionId, execResp.nonAlignQueryDataSet.deepCopy());
     }
     else {
       this.resultSet = new IoTDBJDBCResultSet(this, execResp.getColumns(),
           execResp.getDataTypeList(), execResp.columnNameIndexMap, execResp.ignoreTimeStamp, client, sql, queryId,
-          sessionId, execResp.queryDataSet);
```
yuqi1129, line 328: Better to write some unit test

yuqi1129: `queryDataSet.deepCopy()` seems do not really deeply copy memory,  deep copy only copy   
`List`

PR #1781, opened 2020-09-29
[IOTDB-921] Solve the statement twice call bug

yuqi1129:


yuqi1129:
Someone has take this issue

PR #1782, opened 2020-09-30
[IOTDB-927] Remove fastjson introduced by rocketmq

yuqi1129:


jixuan1989:
@liuruiyiyang , can you have an evaluation for this change?  I am not sure whether RocketMQ client can run as normal if there is no a JSON dependency.
By the way, is there some useful UT/IT that can be introduced from RocektMQ here?

liuruiyiyang:
The fastJson is a dependency of rocketmq-client indeed, since it is used in the rocketmq-remoting component for serialization which is used by almost all rocketmq component including broker, name server and client.

PR #1783, opened 2020-09-30
Remove unnecessary lock in MManager

SilverNarcissus:
There is an unnecessary lock in MManager. I have removed this in this PR.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
@@ -681,25 +681,24 @@ public class MManager {
     }
   }
 
-  public MeasurementSchema[] getSchemas(PartialPath deviceId, String[] measurements)
+  public MeasurementMNode[] getMNodes(PartialPath deviceId, String[] measurements)
       throws MetadataException {
     lock.readLock().lock();
     try {
       MNode deviceNode = getNodeByPath(deviceId);
-      MeasurementSchema[] measurementSchemas = new MeasurementSchema[measurements.length];
-      for (int i = 0; i < measurementSchemas.length; i++) {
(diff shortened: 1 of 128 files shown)


HTHou:
Should we consider to submit a PR to rel/0.10 branch as well?

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -1121,7 +970,6 @@ public StorageGroupMNode getStorageGroupNodeByPath(PartialPath path) throws Meta
    */
   public MNode getDeviceNodeWithAutoCreateAndReadLock(
       PartialPath path, boolean autoCreateSchema, int sgLevel) throws MetadataException {
```
HTHou, line 972: If there is no read lock anymore, what about change the method name to `getDeviceNodeWithAutoCreate`?

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -1205,25 +1039,16 @@ public String getDeviceId(PartialPath path) {
     return device;
   }
 
+  // concurrent warning: you should get parent read lock before calling this method
```
HTHou, line 1042: Remove it?

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
qiaojialin: fixed

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
qiaojialin: fixed

qiaojialin (approved):

HTHou (approved):

PR #1786, opened 2020-09-30
[IOTDB-929] Update show timeseries information in readme

yuqi1129:


HTHou (commented):
Very good job. 
It may be not related to this PR, but I would like to mention the link at line 300 of `README.md` should be fixed. It will be great if you can change it. Thanks.

yuqi1129:
@HTHou , No problem, I will update this PR later

samperson1997 (approved):
Thanks for this vital fix!

Alima777 (approved):
Nice work~ One suggestion is that force-push every time is not a good way especially in a big patch... as we can not see the changes of each commit.

yuqi1129:
OK, force-push is indeed not suggested in PR. The reason why i try to
overlap the code before is that i do not want to have too many commit
message.
Till now, Git can squash commits into one, so i do not need to use force
push again, Thanks

Xiangwei Wei <notifications@github.com> 于2020年9月30日周三 下午4:23写道：

HTHou (approved):

PR #1789, opened 2020-10-02
spelling correct

ayushgupta1915:


HTHou:
Sorry, I don't think this is a meaningful PR...

PR #1791, opened 2020-10-02
[IOTDB-912] Implement wasNull method in AbstractIoTDBJDBCResultSet

jack870131:
The wasNull method had been implemented and getXXX method should be modified according to the JDBC specification while there is null.

### DIFF
--- a/service-rpc/src/main/java/org/apache/iotdb/rpc/IoTDBRpcDataSet.java
+++ b/service-rpc/src/main/java/org/apache/iotdb/rpc/IoTDBRpcDataSet.java
@@ -54,6 +54,7 @@ public class IoTDBRpcDataSet {
   public int fetchSize;
   public boolean emptyResultSet = false;
   public boolean hasCachedRecord = false;
+  public boolean lastReadWasNull;
 
 
   public byte[][] values; // used to cache the current row record value
@@ -284,9 +285,11 @@ public class IoTDBRpcDataSet {
     checkRecord();
     int index = columnOrdinalMap.get(columnName) - START_INDEX;
     if (!isNull(index, rowsIndex - 1)) {
+      lastReadWasNull = false;
       return BytesUtils.bytesToBool(values[index]);
     } else {
-      throw new StatementExecutionException(String.format(VALUE_IS_NULL, columnName));
+      lastReadWasNull = true;
+      return false;
     }
   }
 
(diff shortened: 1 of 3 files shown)


service-rpc/src/main/java/org/apache/iotdb/rpc/IoTDBRpcDataSet.java
```diff
@@ -312,9 +317,23 @@ public float getFloat(String columnName) throws StatementExecutionException {
...
       return BytesUtils.bytesToFloat(values[index]);
     } else {
-      throw new StatementExecutionException(String.format(VALUE_IS_NULL, columnName));
+      lastReadWasNull = true;
+      return 0;
+    }
+  }
+
+  public short getShort(String columnName) throws StatementExecutionException {
+    checkRecord();
```
HTHou, line 329: Actually, we are not supporting short type of data yet...

jack870131: I think so, too. I felt weird that why it was implemented. How about just throwing exception while it has been used?

HTHou: Yes, look good to me. 👍

HTHou (approved):

service-rpc/src/main/java/org/apache/iotdb/rpc/IoTDBRpcDataSet.java
```diff
@@ -343,9 +352,11 @@ public long getLong(String columnName) throws StatementExecutionException {
     }
     int index = columnOrdinalMap.get(columnName) - START_INDEX;
     if (!isNull(index, rowsIndex - 1)) {
+      lastReadWasNull = false;
       return BytesUtils.bytesToLong(values[index]);
     } else {
-      throw new StatementExecutionException(String.format(VALUE_IS_NULL, columnName));
```
qiaojialin, line 348: Maybe throw an exception is better? which can help users find bugs in client

jack870131: The reason why I remove the exception here is that it was not consistent with JDBC specification as we discussed in [IOTDB-912](https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-912?filter=allopenissues).

qiaojialin: ok thanks!

PR #1793, opened 2020-10-03
IOTDB-932 Removed (wrong and redundant) Verification of Paths in Sess…

JulianFeinauer:
…ion API.

I added the Test `IoTDBSessionSimpleIT.createTimeSeriesWithDoubleTicks()` which failed before the fix and works afterwards.


server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -1649,15 +1632,6 @@ protected TSStatus executeNonQueryPlan(PhysicalPlan plan) {
         : RpcUtils.getStatus(TSStatusCode.EXECUTE_STATEMENT_ERROR);
   }
 
-  private TSStatus checkPathValidity(String path) {
-    if (!PATH_PATTERN.matcher(path).matches()) {
```
qiaojialin, line 1653: Hi, could the PATH_PATTERN also be removed?

qiaojialin (approved):

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
JulianFeinauer: Good catch, thanks. Indeed I just pushed an Update and if tests are okay I will merge!

JulianFeinauer:
@qiaojialin the code is fine now from my perspective but the build fails (due to things that are totally unrelated to my work). Is this a current issue or is there something I can do? Otherwise we could merge it?

qiaojialin:
Hi @JulianFeinauer  Currently, the web always builds failed... If you are familiar with Vue, you could have a look :) This PR is good~

PR #1796, opened 2020-10-05
[IOTDB-886]fix float loss precision

EJTTianYu:


HTHou (approved):

server/src/main/java/org/apache/iotdb/db/utils/TypeInferenceUtils.java
```diff
@@ -52,6 +54,10 @@ private static boolean isBoolean(String s) {
         .equalsIgnoreCase(SQLConstant.BOOLEAN_FALSE);
   }
 
+  private static boolean isConvertFloatPrecisionLack(String s){
+    return Long.parseLong(s) > (2 << 24);
```
qiaojialin, line 58: This may be time-consuming when insert integer string.

qiaojialin (approved):

server/src/main/java/org/apache/iotdb/db/utils/TypeInferenceUtils.java
```diff
@@ -63,6 +69,9 @@ public static TSDataType getPredictedDataType(Object value, boolean inferType) {
         return booleanStringInferType;
       } else if (isNumber(strValue)){
         if (!strValue.contains(TsFileConstant.PATH_SEPARATOR)) {
+          if (isConvertFloatPrecisionLack(strValue)) {
```
jixuan1989, line 72: how about just check the length of the string?

qiaojialin: this will only happen at the first insertion of a time series, don't worry

PR #1815, opened 2020-10-11
[IOTDB-940] Supports value filter for chunk and page statistics to ac…

yuqi1129:
…celerate query


JackieTien97 (changes requested):
Actually, we can't use the value filter when there are unseq files. For example,
seq: (1, 1); (2, 2); (3, 3)
unseq: (3, 4);

The query sql is `select s1 from root.sg1.d1 where s1 <= 3`;
If we use the value filter, we will get the point (3, 3) which should not be in the result set.

PR #1816, opened 2020-10-12
Put all log level setting together

HTHou:


qiaojialin (approved):

server/src/assembly/resources/conf/logback.xml
```diff
@@ -196,6 +194,8 @@
         <appender-ref ref="FILEALL"/>
         <appender-ref ref="stdout"/>
     </root>
+    <logger level="info" name="org.apache.iotdb.db.service"/>
```
qiaojialin, line 197: This could be removed...

PR #1819, opened 2020-10-12
Fix Session raw data query bug.

JackieTien97:
We should not generate a statement id for the query. Instead, we should use the statement id in the session itself.

### DIFF
--- a/session/src/main/java/org/apache/iotdb/session/Session.java
+++ b/session/src/main/java/org/apache/iotdb/session/Session.java
@@ -372,7 +372,7 @@ public class Session {
       buffersList.add(buffer);
     }
     request.setValuesList(buffersList);
-    return  request;
+    return request;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


qiaojialin (approved):

session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -931,19 +936,21 @@ public void executeNonQueryStatement(String sql)
...
-   * query eg. select * from paths where time >= startTime and time < endTime
-   * time interval include startTime and exclude endTime
+   * query eg. select * from paths where time >= startTime and time < endTime time interval include
+   * startTime and exclude endTime
+   *
    * @param paths
    * @param startTime included
-   * @param endTime excluded
+   * @param endTime   excluded
    * @return
```
HTHou, line 945: Update javadoc?

HTHou (approved):

PR #1821, opened 2020-10-12
[IOTDB-926] Support reconnection of Session

JackieTien97:


session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -239,7 +244,15 @@ public void insertTablet(Tablet tablet, boolean sorted)
     try {
       RpcUtils.verifySuccess(client.insertTablet(request));
     } catch (TException e) {
-      throw new IoTDBConnectionException(e);
+      if (reconnect()) {
```
neuyilan, line 247: The reconnect() method will call the open() method again, would this cause an infinite loop？
If the open fails, I think we can just throw an exception, as next time the insert request came can call the reconnect() method.

session/src/main/java/org/apache/iotdb/session/Config.java
```diff
@@ -24,5 +24,6 @@
   public static final String DEFAULT_PASSWORD = "password";
   public static final int DEFAULT_FETCH_SIZE = 10000;
   public static final int DEFAULT_TIMEOUT_MS = 0;
-
+  public static final int RETRY_NUM = 3;
+  public static final long RETRY_INTERVAL = 1000;
```
neuyilan, line 28: RETRY_INTERVAL_MS?

qiaojialin: +1

session/src/main/java/org/apache/iotdb/session/Session.java
JackieTien97: It seems that the open() doesn't call the reconnect(), there won't be any infinite loops.

neuyilan: yeah, I read the code in the wrong place, it indeed does not have the problems.

session/src/main/java/org/apache/iotdb/session/Config.java
JackieTien97: fix

qiaojialin (approved):

PR #1823, opened 2020-10-12
fix some regression bugs

Ring-k:
This pull request fixes some regression bugs. In the following case:
  - show timeseries, where the key is not a tag
  - show child paths,  where the path does not exist
distributed and standalone versions should show same error message on client without useless exception information.

### DIFF
--- a/cluster/src/assembly/resources/conf/iotdb-engine.properties
+++ b/cluster/src/assembly/resources/conf/iotdb-engine.properties
@@ -325,7 +325,7 @@ enable_auto_create_schema=false
 
 # Storage group level when creating schema automatically is enabled
 # e.g. root.sg0.d1.s2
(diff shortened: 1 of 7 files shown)


cluster/src/assembly/resources/conf/iotdb-engine.properties
```diff
@@ -326,7 +326,7 @@ enable_auto_create_schema=false
 # Storage group level when creating schema automatically is enabled
 # e.g. root.sg0.d1.s2
 #      we will set root.sg0 as the storage group if storage group level is 2
```
HTHou, line 328:
```suggestion
#      we will set root.sg0 as the storage group if storage group level is 1
```

jt2594838 (approved):

PR #1827, opened 2020-10-15
[Distributed]fix code smell

neuyilan:


cluster/src/main/java/org/apache/iotdb/cluster/log/LogDispatcher.java
```diff
@@ -244,25 +244,20 @@ private void appendEntriesAsync(List<ByteBuffer> logList, AppendEntriesRequest r
...
-        start = System.nanoTime();
-      }
+      Timer.Statistic.RAFT_SENDER_WAIT_FOR_PREV_LOG.setStartTime();
       if (!member.waitForPrevLog(peer, currBatch.get(0).getLog())) {
         logger.warn("{}: node {} timed out when appending {}", member.getName(), receiver,
             currBatch.get(0).getLog());
         return;
       }
-      Timer.Statistic.RAFT_SENDER_WAIT_FOR_PREV_LOG.addNanoFromStart(start);
+      Timer.Statistic.RAFT_SENDER_WAIT_FOR_PREV_LOG.calCostTime();
```
jt2594838, line 253: I don't think the change is thread-safe, please confirm it.

cluster/src/main/java/org/apache/iotdb/cluster/server/Timer.java
```diff
@@ -183,14 +184,33 @@ public void add(long val) {
...
-    public void addNanoFromStart(long start) {
+    public void setStartTime() {
+      setStartTime(System.nanoTime());
+    }
+
+    public void setStartTime(long startTime) {
+      if (ENABLE_INSTRUMENTING) {
+        this.startTime = startTime;
+      }
+    }
```
jt2594838, line 197: The condition should be put into the first method, as `System.nanoTime()` is a heavy call and should be avoided when not using instrumenting.

jt2594838 (changes requested):

cluster/src/main/java/org/apache/iotdb/cluster/log/LogDispatcher.java
neuyilan: you're right, its indeed cause concurrent problem, I'll fix it. thanks!

cluster/src/main/java/org/apache/iotdb/cluster/server/Timer.java
neuyilan: Got it

jt2594838 (approved):

PR #1836, opened 2020-10-20
[IOTDB-948] Use binary search when checking the follower's match index when catchup	

neuyilan:
Now, when the follower lags behind the leader a lot, one catchup task will be triggered. When catchup, it will judge whether a certain log saved in the leader's memory is consistent with the follower. If so, the leader can use logs in its memory to catch up. If the leader cannot find a log in memory that matches the follower, the leader needs to send snapshots.

When checking whether the leader's log match with the follower's, the current implementation starts from the largest index of the stored log in memory, and traverses from large to small to see whether it is consistent with the follower. However, each traversal will send one RPC. Since the maximum number of logs in memory is 1000, there may be 1000 times of RPC, and the time is O( n ). If it becomes a binary search, it only needs O(log n), and 1000 logs only need 10 RPC communications, it'll save communication cost.

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/log/catchup/CatchUpTask.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/log/catchup/CatchUpTask.java
@@ -36,6 +36,7 @@ import org.apache.iotdb.cluster.server.NodeCharacter;
 import org.apache.iotdb.cluster.server.Peer;
 import org.apache.iotdb.cluster.server.member.RaftMember;
 import org.apache.iotdb.cluster.utils.ClientUtils;
+import org.apache.iotdb.db.utils.TestOnly;
 import org.apache.thrift.TException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -1691,7 +1693,7 @@ protected TSStatus executeNonQueryPlan(PhysicalPlan plan) {
     } catch (QueryProcessException e) {
       logger.error("meet error while processing non-query. ", e);
       Throwable cause = e;
-      while (cause.getCause() != null){
+      while (cause.getCause() != null && cause.getCause().getCause() != null) {
```
jt2594838, line 1696: Could you please explain this?

jt2594838 (approved):

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
neuyilan: this is used to fix the UT tests that the exception only shows the message bug but lacks the exception type. I saw you have fixed in the commit[1], but I think it's better to be consistent with the single server.
So if you agree with me, I would re-add the exception type you've deleted in [1], otherwise, I'll change the code here as normal.
[1] https://github.com/apache/iotdb/commit/83e916a9e7bc618f348b7078bd79e62db6d9d546

neuyilan: have changed the code as normal

LebronAl (approved):
LGTM

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
jt2594838: I think it would be better to think from the user side. Such exceptions does few meaning to users, so we'd better remove them for clarity.

PR #1841, opened 2020-10-21
change tsfile_manage_strategy default LEVEL_STRATEGY

zhanglingzhe0820:


server/src/assembly/resources/conf/iotdb-engine.properties
```diff
@@ -335,7 +335,7 @@ chunk_merge_point_threshold=20480
 
 # How many thread will be set up to perform hot compaction, 30 by default.
```
qiaojialin, line 336:
```suggestion
# How many thread will be set up to perform hot compaction, 10 by default.
```

server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
```diff
@@ -559,7 +559,7 @@
    * How many thread will be set up to perform hot compaction, 30 by default. Set to 1 when less
```
qiaojialin, line 559:
```suggestion
   * How many thread will be set up to perform hot compaction, 10 by default. Set to 1 when less
```

server/src/assembly/resources/conf/iotdb-engine.properties
zhanglingzhe0820: okay, I will solve itv

server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
zhanglingzhe0820: okay, I will solve it

qiaojialin:
duplicated with https://github.com/apache/iotdb/pull/1930

PR #1846, opened 2020-10-22
[IOTDB-872] Use system timezone in CLI (Session)

Alima777:
Remove the default timezone in server side. 
**Using the system default time zone of each client** to decide the time zone of server operation.


session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -916,23 +922,14 @@ public boolean checkTimeseriesExists(String path)
...
-
-    TSGetTimeZoneResp resp;
-    try {
-      resp = client.getTimeZone(sessionId);
-    } catch (TException e) {
-      throw new IoTDBConnectionException(e);
-    }
-    RpcUtils.verifySuccess(resp.getStatus());
-    return resp.getTimeZone();
+    return ZoneId.systemDefault().getId();
```
liutaohua, line 929: There is something wrong with the logic. 
It should be the default value that the server stores when the client is not registered with the server.

session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -916,23 +922,14 @@ public boolean checkTimeseriesExists(String path)
...
   }
 
-  private synchronized void setTimeZone(String zoneId)
+  public synchronized void setTimeZone(String zoneId)
       throws StatementExecutionException, IoTDBConnectionException {
     TSSetTimeZoneReq req = new TSSetTimeZoneReq(sessionId, zoneId);
     TSStatus resp;
```
liutaohua, line 935: There may be a bug here, because the value stored by the server does not necessarily change when an error is encountered, but the local cached zoneId does

session/src/test/java/org/apache/iotdb/session/pool/SessionPoolTest.java
```diff
@@ -163,7 +163,7 @@ private void correctQuery(SessionPool pool) {
 
   @Test
   public void tryIfTheServerIsRestart() {
-    SessionPool pool = new SessionPool("127.0.0.1", 6667, "root", "root", 3, 1, 6000, false);
+    SessionPool pool = new SessionPool("127.0.0.1", 6667, "root", "root", 3, 1, 6000, false, null);
```
HTHou, line 166: So we have to fill a `null` to the constructor of SessionPool if we don't want to set timezone, right? 
I think we'd better don't change these tests and add an extra test of setting timezone.

thrift/src/main/thrift/rpc.thrift
```diff
@@ -79,9 +79,10 @@ struct TSOpenSessionResp {
 // Open a session (connection) on the server against which operations may be executed.
 struct TSOpenSessionReq {
   1: required TSProtocolVersion client_protocol = TSProtocolVersion.IOTDB_SERVICE_PROTOCOL_V3
-  2: optional string username
-  3: optional string password
-  4: optional map<string, string> configuration
+  2: required string zoneId
+  3: optional string username
+  4: optional string password
+  5: optional map<string, string> configuration
```
HTHou, line 85: Please update `thrift/rpc-changelist.md`.

Alima777: Fixed.

session/src/test/java/org/apache/iotdb/session/pool/SessionPoolTest.java
Alima777: Not Exactly. We can use a constructor which doesn't need a timezone parameter instead.

HTHou:
Consider add some setting timezone tests in `SessionTest` or `SessionPoolTest`?

Alima777:
> Consider add some setting timezone tests in `SessionTest` or `SessionPoolTest`?

Added~

HTHou (approved):
LGTM

PR #1851, opened 2020-10-23
[IOTDB-951] Fix auto-creating schema in parallel may write unrecognizable timeseries in mlog.txt

wshao08:
Jira Issue: https://issues.apache.org/jira/browse/IOTDB-951

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/MLogWriter.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/MLogWriter.java
@@ -21,9 +21,12 @@ package org.apache.iotdb.db.metadata;
 import java.io.BufferedReader;
 import java.io.BufferedWriter;
 import java.io.File;
+import java.io.FileOutputStream;
 import java.io.FileReader;
 import java.io.FileWriter;
 import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.channels.FileChannel;
 import java.nio.file.Files;
 import java.util.Map;
 import org.apache.iotdb.db.engine.fileSystem.SystemFileFactory;
@@ -38,6 +41,7 @@ public class MLogWriter {
   private static final String STRING_TYPE = "%s,%s,%s"; 
   private File logFile;
   private BufferedWriter writer;
+  private FileOutputStream fileOutputStream;
   private int lineNumber;
 
   public MLogWriter(String schemaDir, String logFileName) throws IOException {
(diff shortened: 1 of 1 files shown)


HTHou (approved):

JackieTien97 (changes requested):
Firstly, the method need not to be synchronnized.
Secondly, the FileChannel instance could be inialized in the contruct function, and shouldn't contruct it every time in the writeTimeseries function.

server/src/main/java/org/apache/iotdb/db/metadata/MLogWriter.java
```diff
@@ -38,6 +41,8 @@
   private static final String STRING_TYPE = "%s,%s,%s"; 
   private File logFile;
   private BufferedWriter writer;
```
JackieTien97, line 43: You can use fileChannel to totally replace it.

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/metadata/MLogWriter.java
```diff
@@ -53,40 +58,50 @@ public MLogWriter(String schemaDir, String logFileName) throws IOException {
...
-    newLine();
+    final String newLine = System.getProperty("line.separator");
+    buf.append(newLine);
+
+    channel.write(ByteBuffer.wrap(buf.toString().getBytes()));
+    channel.force(true);
+    ++lineNumber;
   }
 
   public void deleteTimeseries(String path) throws IOException {
```
samperson1997, line 107: I think you could use `channel.write()` to implement all the functions in this class, including `deleteTimeseries`, `setStorageGroup`, `deleteStorageGroup`, `setTTL`, `changeOffset` and so on

server/src/main/java/org/apache/iotdb/db/metadata/MLogWriter.java
SilverNarcissus, line 43: We can use file channel to replace it. Check other place where it is used ~

SilverNarcissus (changes requested):
Nice work! Just few issues, please have a look~

server/src/main/java/org/apache/iotdb/db/metadata/MLogWriter.java
wshao08: Fixed

wshao08: Fixed

SilverNarcissus (approved):
Brilliant work!  Look good to me

samperson1997 (approved):

PR #1852, opened 2020-10-23
Remove duplicate declaration of maven.assembly plugin

liushangkun:
Remove duplicate declaration of plugin org.apache.maven.plugins:maven-assembly-plugin in tsfile model


qiaojialin (approved):
Thank you!

PR #1863, opened 2020-10-26
[IOTDB-958] Fix MLog bug and add switch for MTree snapshot

samperson1997:
* Fix MLog bug which is not solved in PR #1851 ;
* Add switch for MTree snapshot


SilverNarcissus (approved):
Nice work!

JackieTien97 (approved):

PR #1866, opened 2020-10-26
fix some regression bugs

Ring-k:
This pull request fixes some regression bugs, including:
 - error message in distributed version containing unnecessary exception information
 - always showing error message when deleting timeseries

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
@@ -1334,7 +1334,7 @@ public abstract class RaftMember {
           .getStatus(Arrays.asList(((BatchInsertionException) cause).getFailingStatus()));
     }
     TSStatus tsStatus = StatusUtils.getStatus(StatusUtils.EXECUTE_STATEMENT_ERROR,
-        cause.getClass().getName() + ":" + cause.getMessage());
+        cause.getMessage());
     if (cause instanceof IoTDBException) {
       tsStatus.setCode(((IoTDBException) cause).getErrorCode());
     }
--- a/cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
@@ -1339,6 +1339,9 @@ public abstract class RaftMember {
     }
     TSStatus tsStatus = StatusUtils.getStatus(StatusUtils.EXECUTE_STATEMENT_ERROR,
         cause.getMessage());
+    if (cause instanceof RuntimeException) {
+      logger.error("RuntimeException during executing {}", log, cause);
+    }
     if (cause instanceof IoTDBException) {
       tsStatus.setCode(((IoTDBException) cause).getErrorCode());
     }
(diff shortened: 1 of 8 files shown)


mychaow (approved):
LGTM

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
```diff
@@ -371,7 +374,12 @@ protected QueryDataSet processDataQuery(QueryPlan queryPlan, QueryContext contex
       } else if (queryPlan instanceof LastQueryPlan) {
         queryDataSet = queryRouter.lastQuery((LastQueryPlan) queryPlan, context);
       } else {
-        queryDataSet = queryRouter.rawDataQuery((RawDataQueryPlan) queryPlan, context);
+        try {
+          queryDataSet = queryRouter.rawDataQuery((RawDataQueryPlan) queryPlan, context);
+        } catch (IndexOutOfBoundsException e) {
```
LebronAl, line 379: Sorry, I'm a little confused about when will this method throw a IndexOutOfBoundsException?

LebronAl (approved):
LGTM

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
Ring-k: When we have some nonsense predicates, such as 
 - select status, temperature from root.ln.wf01.wt01 where time < 2017-11-01T00:05:00.000 and time > 2017-11-01T00:12:00.000

jt2594838: I believe this is a misuse, since it does not involve any index, and I would prefer creating a new Exception like `IllegalTimeRangeException` of `IllegalPredicateException`. You may correct it within this PR or just create an issue about it.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
```diff
@@ -1334,7 +1338,7 @@ private TSStatus handleLogExecutionException(
           .getStatus(Arrays.asList(((BatchInsertionException) cause).getFailingStatus()));
     }
     TSStatus tsStatus = StatusUtils.getStatus(StatusUtils.EXECUTE_STATEMENT_ERROR,
-        cause.getClass().getName() + ":" + cause.getMessage());
+        cause.getMessage());
     if (cause instanceof IoTDBException) {
```
jt2594838, line 1342: This is not relevant to this PR, but it occurs to me that RunTimeExceptions sometimes does not provide useful messages, and it would difficult to locate the problem.
So please add this:
```
if (cause instanceof RunTimeException) {
  logger.error("RuntimeException during executing {}", log , cause);
}
```

jt2594838 (approved):
Good job.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
Ring-k: That's a good idea.

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
Ring-k: To make it more clear, I created EmptyIntervalException.

PR #1868, opened 2020-10-27
Stop iotdb instance when RPC ip address or port is unavailable

HTHou:


server/src/main/java/org/apache/iotdb/db/service/RPCService.java
```diff
@@ -61,12 +63,17 @@ public void initTProcessor() throws ClassNotFoundException,IllegalAccessExceptio
...
+    try {
+      thriftServiceThread = new ThriftServiceThread(processor,
+          getID().getName(), ThreadName.RPC_CLIENT.getName(),
+          config.getRpcAddress(), config.getRpcPort(), config.getRpcMaxConcurrentClientNum(),
+          config.getThriftServerAwaitTimeForStopService(),
+          new RPCServiceThriftHandler(impl),
+          IoTDBDescriptor.getInstance().getConfig().isRpcThriftCompressionEnable());
+    }
+    catch (RPCServiceException e) {
+      throw new IllegalAccessException(e.getMessage());
```
jt2594838, line 75: Format here.

Besides, instead of shutting down immediately, I would suggest a few retries, which may prevent test failures when the socket is not shut down timely.

jt2594838 (approved):

server/src/main/java/org/apache/iotdb/db/service/RPCService.java
HTHou: Good idea.
So do you think we should add the retries in this pr or do it in another pr later? I don't see any test failed so far....

PR #1869, opened 2020-10-27
[IOTDB-1131]Improve wal by using differential logging

jt2594838:
WALs may contain duplicated fields like deviceIds, measurementsId and data types. This PR aims to eliminate them to improve WAL performance when using small batchSize and without damaging the performance when using large batchSize.

Design:
[Differential Logging.docx](https://github.com/apache/iotdb/files/5442493/Differential.Logging.docx)

Evaluation:
[differentiation_logging.xlsx](https://github.com/apache/iotdb/files/5442494/differentiation_logging.xlsx)


server/src/main/java/org/apache/iotdb/db/writelog/io/ILogReaderFactory.java
```diff
@@ -0,0 +1,24 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iotdb.db.writelog.io;
+
+import java.io.File;
```
SilverNarcissus, line 22: This file is empty. Please have a look

server/src/main/java/org/apache/iotdb/db/utils/datastructure/RandomAccessArrayDeque.java
```diff
@@ -0,0 +1,939 @@
...
+import java.util.function.Consumer;
+
+/**
+ * RandomAccessArrayDeque is a modified version of ArrayDeque from java.util, providing the ability
+ * to access an element by index.
+ * Most codes are copied directly because the fields are package-private and cannot be accessed
+ * by a simple extension.
+ * @param <E>
+ */
+public class RandomAccessArrayDeque<E> extends AbstractCollection<E>
```
SilverNarcissus, line 44: Wow~ That's a fantastic structure but without test. May be we should add some test for it.

server/src/main/java/org/apache/iotdb/db/writelog/recover/LogReplayer.java
```diff
@@ -91,17 +91,24 @@ public void replayLogs() {
         logNodePrefix + FSFactoryProducer.getFSFactory().getFile(insertFilePath).getName());
 
     ILogReader logReader = logNode.getLogReader();
+    long readLogTime = 0;
+    long replayLogTime = 0;
     try {
       while (logReader.hasNext()) {
         try {
+          long start = System.nanoTime();
```
SilverNarcissus, line 99: System.nanoTime() will call "clock_gettime" which is a system call. So calculating the replay time of every row may have some performance impact.

SilverNarcissus (changes requested):
Nice work and detailed design with analyze. Only have some small issues, please have a look~

server/src/main/java/org/apache/iotdb/db/writelog/recover/LogReplayer.java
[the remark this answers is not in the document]
jt2594838: Removed.

server/src/main/java/org/apache/iotdb/db/writelog/io/ILogReaderFactory.java
jt2594838: Removed.

server/src/main/java/org/apache/iotdb/db/utils/datastructure/RandomAccessArrayDeque.java
jt2594838: Since the class is mainly copied from java.util, I think most of it is trustable, and I only add some tests of `get()`.

SilverNarcissus (approved):
Look good to me~

jixuan1989:
ping.

HTHou:
The conflicts fixed. The PR looks better now!

PR #1898, opened 2020-10-29
[IOTDB-968] Support time predicate in select last

wshao08:
JIRA issue: https://issues.apache.org/jira/browse/IOTDB-968

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/query/executor/LastQueryExecutor.java
+++ b/server/src/main/java/org/apache/iotdb/db/query/executor/LastQueryExecutor.java
@@ -181,7 +181,7 @@ public class LastQueryExecutor {
   }
 
   private static boolean satisfyFilter(Filter filter, TimeValuePair tvPair) {
-    if (filter != null &&
+    if (filter == null ||
         filter.satisfy(tvPair.getTimestamp(), tvPair.getValue().getValue())) {
       return true;
     }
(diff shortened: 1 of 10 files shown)


server/src/main/java/org/apache/iotdb/db/query/executor/LastQueryExecutor.java
```diff
@@ -127,103 +127,64 @@ protected TimeValuePair calculateLastPairForOneSeries(
...
   }
 
-  private static TimeValuePair constructLastPair(long timestamp, Object value,
-                                                 TSDataType dataType) {
-    return new TimeValuePair(timestamp, TsPrimitiveType.getByType(dataType, value));
+  private static boolean satisfyFilter(Filter filter, TimeValuePair tvPair) {
+    if (filter != null &&
+        filter.satisfy(tvPair.getTimestamp(), tvPair.getValue().getValue())) {
+      return true;
+    }
```
JackieTien97, line 187: if the filter is null, shoudn't we return true?

server/src/main/java/org/apache/iotdb/db/query/executor/fill/LastPointReader.java
```diff
@@ -216,6 +216,9 @@ private boolean shouldUpdate(long time, long version, long newTime, long newVers
   }
 
   private boolean endtimeContainedByTimeFilter(Statistics statistics) {
+    if (timeFilter == null) {
+      return false;
+    }
```
JackieTien97, line 221: Same as before, if the filter is null, shoudn't we return true?

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/query/executor/LastQueryExecutor.java
wshao08: Fixed

server/src/main/java/org/apache/iotdb/db/query/executor/fill/LastPointReader.java
wshao08: Fixed

JackieTien97 (approved):

HTHou (approved):

PR #1902, opened 2020-10-29
[Distributed] Fix create multi time series plan serialization

LebronAl:


server/src/main/java/org/apache/iotdb/db/qp/physical/sys/CreateMultiTimeSeriesPlan.java
```diff
@@ -164,6 +164,7 @@ public void serialize(DataOutputStream stream) throws IOException {
 
     if (alias != null) {
       stream.write(1);
+      stream.writeInt(alias.size());
```
mychaow, line 167: If we support variable length of alias array, how to know which alias match the measurement? I think we should support the same size of alias array and measurement, but the alias can be null. 

How do you think?

mychaow (commented):
Good job!

server/src/main/java/org/apache/iotdb/db/qp/physical/sys/CreateMultiTimeSeriesPlan.java
LebronAl: You are right~I misunderstood the code here,I will fix this.
Thanks!

mychaow (approved):

PR #1917, opened 2020-11-02
Add restrictions for LAST query filter

wshao08:
Considering query performance, we add restrictions to support only '>' or '>=' in LAST query currently.


server/src/main/java/org/apache/iotdb/db/query/executor/LastQueryExecutor.java
```diff
@@ -145,13 +145,17 @@ public static TimeValuePair calculateLastPairForOneSeriesLocally(
...
+          return null;
         }
       }
 
       if (node != null && node.getCachedLast() != null) {
         TimeValuePair timeValuePair =  node.getCachedLast();
         if (timeValuePair != null && satisfyFilter(filter, timeValuePair)) {
           return timeValuePair;
+        } else if (timeValuePair != null) {
+          return null;
```
JackieTien97, line 158: If the cached last time-value pair doesn't satify the filter, shouldn't we keep to call the calculateLastPairByScanningTsFiles() function?

JackieTien97 (changes requested):

JackieTien97 (approved):

qiaojialin (approved):

PR #1955, opened 2020-11-05
[IOTDB-978] lsof does not exist in stop-server.sh

yhwangBonc:


server/src/assembly/resources/sbin/stop-server.sh
```diff
@@ -21,10 +21,14 @@
 
 IOTDB_CONF="`dirname "$0"`/../conf"
 rpc_port=`sed '/^rpc_port=/!d;s/.*=//' ${IOTDB_CONF}/iotdb-engine.properties`
-PID=$(lsof -t -i:${rpc_port})
-
-if [ "$PID" = "" ]; then
+if type lsof > /dev/null; then
+  PID=$(lsof -t -i:${rpc_port})
+else
```
zhanglingzhe0820, line 26: why not just use "ps ax..." as 'lsof' does not exist in most operating system

zhanglingzhe0820 (commented):
Does keep 'lsof...' command necessarily?

zhanglingzhe0820 (approved):

jt2594838 (approved):

yhwangBonc (commented):
> Does keep 'lsof...' command necessarily?

lsof can guarantee that only the services started by itself will be killed, while ps ax will stop all services, so I think lsof should be used first

PR #1968, opened 2020-11-06
Fix the risk of deadlock by WeakReference

asdf2014:


server/src/main/java/org/apache/iotdb/db/engine/cache/TimeSeriesMetadataCache.java
```diff
@@ -60,6 +64,7 @@
 
   private final ReadWriteLock lock = new ReentrantReadWriteLock();
 
+  private final Map<String, WeakReference<String>> devices = Collections.synchronizedMap(new WeakHashMap<>());
```
jixuan1989, line 67: this should be discussed. 
We may have too many devices and the memory is limited..

asdf2014: Hi, @jixuan1989 . Thanks for your reply. Indeed, taking this into consideration, so WeakReference is used here, which is more GC friendly. Moreover, I think the previous method has the risk of deadlock and the memory usage is not small.

JackieTien97 (approved):

HTHou (approved):

server/src/main/java/org/apache/iotdb/db/engine/cache/TimeSeriesMetadataCache.java
asdf2014: Hi, @JackieTien97 . FYI, https://stackoverflow.com/a/134154/4719867

PR #1975, opened 2020-11-07
fix nullpointer exception if no password is set when login

jixuan1989:
How to reproduce:

run iotdb-grafana and access http://127.0.0.1:8888/search

TODO: we may break the connection if a user login failed, to avoid some malicious connections (which will take a thread for each connection).

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/auth/authorizer/BasicAuthorizer.java
+++ b/server/src/main/java/org/apache/iotdb/db/auth/authorizer/BasicAuthorizer.java
@@ -102,7 +102,7 @@ public abstract class BasicAuthorizer implements IAuthorizer, IService {
   @Override
   public boolean login(String username, String password) throws AuthException {
     User user = userManager.getUser(username);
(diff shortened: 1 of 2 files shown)


qiaojialin (approved):

server/src/main/java/org/apache/iotdb/db/auth/authorizer/BasicAuthorizer.java
```diff
@@ -102,7 +102,7 @@ public static IAuthorizer getInstance() throws AuthException {
   @Override
   public boolean login(String username, String password) throws AuthException {
     User user = userManager.getUser(username);
-    return user != null && user.getPassword().equals(AuthUtils.encryptPassword(password));
+    return user != null && password!=null && user.getPassword().equals(AuthUtils.encryptPassword(password));
```
qiaojialin, line 105:
```suggestion
    return user != null && password != null && user.getPassword().equals(AuthUtils.encryptPassword(password));
```

PR #2046, opened 2020-11-13
client-go

yanhongwangg:


client-go/src/main/iotdb_session/get_session.go
```diff
@@ -0,0 +1,97 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package iotdbSession
+
+const DefaultUser = "root"
```
neuyilan, line 22:
```suggestion
const {
DefaultUser = "root"
DefaultPasswd = "root"
}

```

client-go/src/main/iotdb_session/get_session.go
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package iotdbSession
```
neuyilan, line 20:
```suggestion
package session
```

client-go/src/main/iotdb_session/get_session.go
```diff
@@ -0,0 +1,97 @@
...
+	})
+}
+
+func withFetchSize(fetchSize int32) DialOption {
+	return newFuncOption(func(session *Session) {
+		session.FetchSize = fetchSize
+	})
+}
+
+//默认参数
```
neuyilan, line 72: change it to English comment

client-go/src/main/iotdb_session/get_session.go
```diff
@@ -0,0 +1,97 @@
...
+
+type SessionConn struct {
+	session Session
+}
+
+func NewSession(host string, port string, opts ...DialOption) Session {
+	sessionConn := &SessionConn{
+		session: defaultOptions(),
+	}
+	//循环调用opts
```
neuyilan, line 90: same as above

client-go/src/main/iotdb_session/session.go
```diff
@@ -0,0 +1,290 @@
...
+	"log"
+	"net"
+	"os"
+	"strings"
+	"time"
+)
+
+const protocolVersion = rpc.TSProtocolVersion_IOTDB_SERVICE_PROTOCOL_V3
+
+var client *rpc.TSIServiceClient
```
neuyilan, line 36: Group Similar Declarations, This also applies to constants, variables, and type declarations.

https://github.com/uber-go/guide/blob/master/style.md#group-similar-declarations

client-go/src/main/iotdb_session/session.go
```diff
@@ -0,0 +1,290 @@
...
+var sessionId int64
+var isClose bool = true
+var trans thrift.TTransport
+var err error
+var requestStatementId int64
+
+func (s *Session) Open(enableRPCCompression bool, connectionTimeoutInMs int) error {
+	dir, _ := os.Getwd()
+	os.Mkdir(dir+"\\logs", os.ModePerm)
```
neuyilan, line 45: I think the logs should be redesign,
I suggest using the log : https://github.com/sirupsen/logrus

client-go/src/main/iotdb_session/utils/field.go
```diff
@@ -0,0 +1,122 @@
...
+
+import (
+	"encoding/json"
+)
+
+type Field struct {
+	DataType int32
+}
+
+var boolV bool
```
neuyilan, line 30: the same as above

client-go/src/test/Session_test.go
```diff
@@ -0,0 +1,185 @@
+/**
```
neuyilan, line 1: the file name should all use lower case

neuyilan (changes requested):

client-go/src/main/iotdb_session/utils/field.go
yanhongwangg: Thank you for your guidance

client-go/src/main/iotdb_session/session.go
yanhongwangg: good idea!!!

client-go/src/main/iotdb_session/session.go
yanhongwangg: I think what you said is very good

JackieTien97 (changes requested):

client-go/src/main/client/utils/row_record.go
```diff
@@ -0,0 +1,68 @@
...
+			res += strconv.FormatInt(objValue.(int64), 10)
+			break
+		case "FLOAT":
+
+			res += strconv.FormatFloat(float64(objValue.(float32)), 'f', 1, 32)
+			break
+		case "DOUBLE":
+			res += strconv.FormatFloat(objValue.(float64), 'f', -1, 64)
+			break
+		case "TEXT":
```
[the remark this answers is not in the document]
yanhongwangg: copy that

client-go/src/main/client/utils/iotdb_constants.go
```diff
@@ -0,0 +1,52 @@
...
+
+const (
+	UNCOMPRESSED int32 = 0
+	SNAPPY       int32 = 1
+	GZIP         int32 = 2
+	LZO          int32 = 3
+	SDT          int32 = 4
+	PAA          int32 = 5
+	PLA          int32 = 6
+	LZ4          int32 = 7
```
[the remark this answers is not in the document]
yanhongwangg: Well, you're right

client-go/src/main/client/session.go
```diff
@@ -0,0 +1,708 @@
...
+	client             *rpc.TSIServiceClient
+	sessionId          int64
+	trans              thrift.TTransport
+	err                error
+	requestStatementId int64
+	ts                 string
+	sg                 string
+	dv                 string
+	Log                = logrus.New()
+)
```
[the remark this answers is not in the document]
yanhongwangg: Thanks for your guidance

client-go/pom.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
yanhongwangg: I agree with you

client-go/README.md
```diff
@@ -0,0 +1,304 @@
...
+*Enter project directory eg:cd session
+
+*Download dependency : go get github.com/yanhongwangg/incubator-iotdb@client-go
+
+*eg:session_example.go
+
+```
+package main
+
+import (
```
neuyilan, line 39: if those codes same as the session_example.go, I think we can just add the session_example.go link here.

client-go/src/main/client/session.go
```diff
@@ -0,0 +1,698 @@
...
+	dataSet.CloseOperationHandle()
+	return result
+}
+
+func (s *Session) Close() error {
+	tSCloseSessionReq := rpc.NewTSCloseSessionReq()
+	tSCloseSessionReq.SessionId = s.sessionId
+	status, err := s.client.CloseSession(context.Background(), tSCloseSessionReq)
+	if err != nil {
+		Log.WithError(err).Error("Error occurs when closing session at server. Maybe server is down. Error message")
```
neuyilan, line 100: please make the error message more clear.

client-go/src/main/client/session.go
```diff
@@ -0,0 +1,698 @@
...
+/*
+ *delete one storage group
+ *
+ *param
+ *storageGroupId: string, storage group name (starts from root)
+ *
+ *return
+ *error: correctness of operation
+ */
+func (s *Session) DeleteStorageGroup(storageGroupId string) {
```
neuyilan, line 140: the method no return value, the annotation should be consistent with the codes,  the same for other methods

client-go/src/main/client/session.go
```diff
@@ -0,0 +1,698 @@
...
+ */
+func (s *Session) SetStorageGroup(storageGroupId string) error {
+	status, err := s.client.SetStorageGroup(context.Background(), s.sessionId, storageGroupId)
+	if err != nil {
+		Log.WithError(err).Error("setting storage group failed")
+	} else {
+		Log.WithFields(logrus.Fields{
+			"sg":   storageGroupId,
+			"code": status.Code,
+		}).Info("setting storage group success")
```
neuyilan, line 126: I think if the operation succeeds, we do not need to print the info message, as this will print so many logs for users who use the go-client. I think the debug message is ok. and the default log level is info.

client-go/src/main/client/session.go
```diff
@@ -0,0 +1,698 @@
...
+ *dataTypes: [][]int32, each element of outer list indicates sensor data types of a device
+ *values: [][]interface{}, values to be inserted, for each device
+ *timestamps: []int64, timestamps for records
+ *
+ *return
+ *error: correctness of operation
+ */
+func (s *Session) InsertRecords(deviceIds []string, measurements [][]string, dataTypes [][]int16, values [][]interface{},
+	timestamps []int64) {
+	s.dv = strings.Replace(strings.Trim(fmt.Sprint(deviceIds), "[]"), " ", ",", -1)
```
neuyilan, line 423: for every inset operation, this may time-consume, If you want to print the devices,  I think the `deviceIds[]` is just enough.

neuyilan (changes requested):

client-go/README.md
yanhongwangg: It's exactly the same. I'll post the link

client-go/src/main/client/session.go
yanhongwangg: But the direct output deviceids [] is the address

client-go/src/main/client/session.go
yanhongwangg: copy that

client-go/src/main/client/session.go
```diff
@@ -0,0 +1,667 @@
...
+/*
+ *delete multiple storage group
+ *
+ *param
+ *storageGroupIds: []string, paths of the target storage groups
+ *
+ */
+func (s *Session) DeleteStorageGroups(storageGroupIds []string) {
+	status, err := s.client.DeleteStorageGroups(context.Background(), s.sessionId, storageGroupIds)
+	s.sg = strings.Replace(strings.Trim(fmt.Sprint(storageGroupIds), "[]"), " ", ",", -1)
```
neuyilan, line 153: Do not do the code here as you just want to print the devices in debug log, for every operation, the code will be called, it'll cost cpu
`s.sg = strings.Replace(strings.Trim(fmt.Sprint(storageGroupIds), "[]"), " ", ",", -1)`

client-go/src/main/client/session.go
```diff
@@ -0,0 +1,667 @@
...
+ *dataTypes: []int32, data types for time series
+ *encodings: []int32, encodings for time series
+ *compressors: []int32, compressing types for time series
+ *
+ */
+func (s *Session) CreateMultiTimeseries(paths []string, dataTypes []int32, encodings []int32, compressors []int32) {
+	request := rpc.TSCreateMultiTimeseriesReq{SessionId: s.sessionId, Paths: paths, DataTypes: dataTypes,
+		Encodings: encodings, Compressors: compressors}
+	status, err := s.client.CreateMultiTimeseries(context.Background(), &request)
+	s.ts = strings.Replace(strings.Trim(fmt.Sprint(paths), "[]"), " ", ",", -1)
```
neuyilan, line 202: same as above！

client-go/src/main/client/session.go
neuyilan: then you can do it if the log debug level is enabled, if the log level is info,  for every operation the code will be called once, it's unnecessary!

neuyilan (changes requested):

client-go/src/main/client/session.go
yanhongwangg: thanks for your guidance

PR #2049, opened 2020-11-14
[IOTDB-1007]Fix session pool concurrency and leakage issue when pool.close is called

jixuan1989:
Current session Pool has some concurrency problems:

1.

public static void main() {
    queryByRowRecord(); // this method open a new thread to use session.
    Thread.sleep(1000);
    pool.close(); // this method close pool.
}
Then the pool will clear all existing sessions, which may lead to some leakage and deadlock.

2. `notifyAll()` is not called, which will lead to `wait(1000)` waits too long time.

3. More log is printed.

### DIFF
--- a/session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
+++ b/session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
@@ -76,6 +76,8 @@ public class SessionPool {
   private boolean enableCompression = false;
   private ZoneId zoneId;
 
+  private boolean closed;//whether the queue is closed.
+
   public SessionPool(String ip, int port, String user, String password, int maxSize) {
     this(ip, port, user, password, maxSize, Config.DEFAULT_FETCH_SIZE, 60_000, false, null);
   }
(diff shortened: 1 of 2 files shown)


session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
```diff
@@ -76,6 +76,8 @@
   private boolean enableCompression = false;
   private ZoneId zoneId;
 
+  private boolean closed;//whether the queue is closed.
```
JackieTien97, line 79: I think it should be volatile, cause it is used outside a synchronized block somewhere below

session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
```diff
@@ -110,20 +112,64 @@ public SessionPool(String ip, int port, String user, String password, int maxSiz
...
+            } else {
+              return session;
+            }
+          }
+        } catch (IoTDBConnectionException e) {
+          //if exception, we will throw the exception.
+          //Meanwhile, we have to set size--
+          synchronized (this) {
+            size--;
+            this.notifyAll();
```
JackieTien97, line 154: I think notify() is better, cause only one thread can get session.

session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
```diff
@@ -178,6 +207,10 @@ private void putBack(Session session) {
     queue.push(session);
     synchronized (this) {
       this.notifyAll();
```
JackieTien97, line 209: notify() is better.

session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
```diff
@@ -223,10 +260,12 @@ public void closeResultSet(SessionDataSetWrapper wrapper) {
   }
 
   private synchronized void removeSession() {
+    logger.warn("Remove a broken Session {}, {}, {}", ip, port, user);
+    size--;
+    this.notifyAll();
```
JackieTien97, line 265: notify() is better.

JackieTien97 (changes requested):

session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
jixuan1989: but it does not take native effective if a thread sees the value of `closed` with a latency.

session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
jixuan1989: fixed.

session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
jixuan1989: fixed.

JackieTien97 (approved):

PR #2060, opened 2020-11-17
division by zero bug

haimeiguo:
division by zero bug, when endtime - starttime < interval

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/qp/strategy/PhysicalGenerator.java
+++ b/server/src/main/java/org/apache/iotdb/db/qp/strategy/PhysicalGenerator.java
@@ -687,8 +687,10 @@ public class PhysicalGenerator {
     if (queryPlan instanceof GroupByTimePlan) {
       GroupByTimePlan plan = (GroupByTimePlan) queryPlan;
       // the actual row number of group by query should be calculated from startTime, endTime and interval.
-      fetchSize = Math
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/apache/iotdb/db/qp/strategy/PhysicalGenerator.java
```diff
@@ -687,8 +687,10 @@ private void deduplicate(QueryPlan queryPlan, int fetchSize)
     if (queryPlan instanceof GroupByTimePlan) {
       GroupByTimePlan plan = (GroupByTimePlan) queryPlan;
       // the actual row number of group by query should be calculated from startTime, endTime and interval.
-      fetchSize = Math
-          .min((int) ((plan.getEndTime() - plan.getStartTime()) / plan.getInterval()), fetchSize);
+      long interval = (plan.getEndTime() - plan.getStartTime()) / plan.getInterval();
```
neuyilan, line 690: What about the plan.getInterval() is zero?  does the code has been checked before?

haimeiguo: yes. if you set interval = 0, it will throw exception: "Interval must more than 0"

neuyilan: got it

samperson1997 (approved):

PR #2063, opened 2020-11-17
Refactor LastQueryExecutor to separate the execution into multiple stages

wshao08:
A merge lock has already been added locking the whole Last query procedure, however, this procedure may take a long time. 
This PR refactor the Last query execution process, separating it into multiple stages:

1. Get Last result by accessing cache.
2. Hold merge lock and obtain query resource.
3. Scan TsFiles to get Last value.

Now only the stage 2 will be locked.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/query/executor/QueryRouter.java
+++ b/server/src/main/java/org/apache/iotdb/db/query/executor/QueryRouter.java
@@ -225,8 +225,12 @@ public class QueryRouter implements IQueryRouter {
   @Override
   public QueryDataSet lastQuery(LastQueryPlan lastQueryPlan, QueryContext context)
       throws StorageEngineException, QueryProcessException, IOException {
-    LastQueryExecutor lastQueryExecutor = new LastQueryExecutor(lastQueryPlan);
+    LastQueryExecutor lastQueryExecutor = getLastQueryExecutor(lastQueryPlan);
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/apache/iotdb/db/query/executor/QueryRouter.java
```diff
@@ -225,12 +225,8 @@ public QueryDataSet groupByFill(GroupByTimeFillPlan groupByFillPlan, QueryContex
   @Override
   public QueryDataSet lastQuery(LastQueryPlan lastQueryPlan, QueryContext context)
       throws StorageEngineException, QueryProcessException, IOException {
-    LastQueryExecutor lastQueryExecutor = getLastQueryExecutor(lastQueryPlan);
+    LastQueryExecutor lastQueryExecutor = new LastQueryExecutor(lastQueryPlan);
     return lastQueryExecutor.execute(context, lastQueryPlan);
   }
 
-  protected LastQueryExecutor getLastQueryExecutor(LastQueryPlan lastQueryPlan) {
```
qiaojialin, line 232: this may be used in cluster

jt2594838: Yes, indeed. I hope you can make some interfaces out of it, which would better not be modified so frequently.

JackieTien97 (approved):

PR #2070, opened 2020-11-18
Update level compaction delete and change list lock

zhanglingzhe0820:


qiaojialin (approved):

JackieTien97 (approved):

server/src/main/java/org/apache/iotdb/db/engine/compaction/level/LevelCompactionTsFileManagement.java
```diff
@@ -338,12 +342,14 @@ public void recover() {
...
-                deleteLevelFiles(timePartition, currMergeFile);
+                deleteLevelFilesInDisk(currMergeFile);
+                deleteLevelFilesInList(timePartition, currMergeFile);
               }
             } else {
               for (List<TsFileResource> currMergeFile : unSequenceTsFileResources
                   .get(timePartition)) {
-                deleteLevelFiles(timePartition, currMergeFile);
+                deleteLevelFilesInDisk(currMergeFile);
+                deleteLevelFilesInList(timePartition, currMergeFile);
```
EJTTianYu, line 352: Why not combine these two statements into one function, because most of the time they appear at the same time

EJTTianYu (approved):

PR #2078, opened 2020-11-19
[IOTDB-1017] Fix file version bug

SilverNarcissus:
Currently tsfile resource version is generated by version controller. When recover, version will increase 100 times but not just 1. This PR is to fix this issue.


jt2594838 (approved):

samperson1997 (approved):
Excellent code!

SilverNarcissus:
For sonar smell, if we don't pass that parameter, we have to split file name to get file version which cost CPU and memory.

PR #2113, opened 2020-11-24
when auto create schema, we need to wait for sg ready

mychaow:


cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
```diff
@@ -416,6 +424,35 @@ public void createSchema(PhysicalPlan plan) throws MetadataException {
...
+        break;
+      }
+
+      if (System.currentTimeMillis() - startTime > ClusterDescriptor.getInstance().getConfig().getConnectionTimeoutInMS()) {
+        break;
+      } else {
+        try {
+          Thread.sleep(1);
+        } catch (InterruptedException e) {
+          logger.debug("Failed to wait for creating sgs for plan");
```
LebronAl, line 449: It would be better if  the log here contains specified log~

LebronAl (approved):
LGTM

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
```diff
@@ -416,6 +424,35 @@ public void createSchema(PhysicalPlan plan) throws MetadataException {
...
+    while (true) {
+      int successNum = deviceIds.size();
+      for (PartialPath deviceId : deviceIds) {
+        PartialPath storageGroupName = MetaUtils
+          .getStorageGroupPathByLevel(deviceId, IoTDBDescriptor.getInstance()
+            .getConfig().getDefaultStorageGroupLevel());
+        if (IoTDB.metaManager.isStorageGroup(storageGroupName)) {
+          successNum --;
+        }
+      }
```
jt2594838, line 438: All deviceIds are verified in every loop, which is unnecessary.
It would be better to copy deviceIds into another list and remove them once verified, or use a boolean array to indicate whether the deviceId has been verified.

jt2594838 (approved):

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
mychaow: ok

jt2594838 (approved):

PR #2114, opened 2020-11-24
[To rel/0.11] change mem control lock usage

HTHou:


server/src/assembly/resources/conf/iotdb-engine.properties
```diff
@@ -174,6 +174,7 @@ timestamp_precision=ms
 wal_buffer_size=16777216
 
 # When a TsFile's file size (in byte) exceeds this, the TsFile is forced closed.
+# It may cause memTable size smaller, if it sets a large value
```
qiaojialin, line 177:
```suggestion
# It may cause memTable size smaller if it is a large value
```

qiaojialin (approved):

qiaojialin (approved):

Issue #2110, opened 2020-11-22
Outdated documentation at https://hub.docker.com/r/apache/iotdb


fagnercarvalho:
The documentation to start the IoTDB CLI seems to be wrong at https://hub.docker.com/r/apache/iotdb.

1. Run `docker run -d -p 6667:6667 -p 31999:31999 -p 8181:8181 --name some-iotdb apache/iotdb:latest`
2. Run `docker exec -it some-iotdb /bin/bash`
3. Run `start-client.sh`
4. See error `bash: start-client.sh: command not found`

**Expected behavior**
The command should be something like `/iotdb/sbin/start-cli.sh`.

**Desktop:**
 - OS: Windows
 - Version: latest

jixuan1989:
Ah yes, many thanks! 
Before 0.10, it is `start-client.sh`
After 0.10, it is `start-cli.sh`

HTHou:
Hi, the user guide in docker hub has been updated, thanks for reminding!

fagnercarvalho:
Thank you for jumping on this so quickly! 

I checked the repo and I found a few more places where this is outdated.

- https://github.com/apache/iotdb/blob/master/docs/UserGuide/System%20Tools/Watermark%20Tool.md
- https://github.com/apache/iotdb/blob/master/docs/zh/UserGuide/System%20Tools/Watermark%20Tool.md
- https://github.com/apache/iotdb/blob/master/docker/ReadMe.md

I think we can close this after fixing the name in these files.

HTHou:
Thank you so much~ 
I've submitted a PR to fix it.

PR #2120, opened 2020-11-25
[Issue-2110] Update start-client to start-cli

HTHou:
https://github.com/apache/iotdb/issues/2110


samperson1997 (approved):
Thanks for ur quick fix ... and I must admit that the user who proposed this issue is so careful : )

PR #2141, opened 2020-11-29
Fix compaction add list bug when last level exceeds max file num

zhanglingzhe0820:


qiaojialin (approved):

qiaojialin:
already cherry-picked to rel/0.11

PR #2144, opened 2020-11-30
[IOTDB-1011] Memtable sort in query

SilverNarcissus:
Now a query will copy and sort TV list from memtable but never put it back to memtable. That's means when another query come, same data will be sort again. So we should put it back to memtable.

JIRA: https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-1011
Design: https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=165225810
Test report: https://cwiki.apache.org/confluence/display/IOTDB/Sort+Memtable+in+inserting+test+report


samperson1997 (approved):
LGTM

jixuan1989:
how about the performance?

jixuan1989 (approved):
+1 for this PR.
But is there some way to improve the TVList Buffer Pool better?
As currently only the tvlists in a memtable which never be queried can be reused.

PR #2150, opened 2020-12-01
[IOTDB-1035] Fix bug in getDeviceTimeseriesMetadata when querying non-exist device

samperson1997:
When querying a non-exist device, `getDeviceTimeseriesMetadata()` method may still return some `TimeseriesMetadata`. This is caused by `getMetadataAndEndOffset()`.


samperson1997:
This bug also causes bugs in merge strategy, which is mentioned in ISSUE #2124 .

PR #2152, opened 2020-12-01
Update compaction level list delete

zhanglingzhe0820:
Now in LEVEL_COMPACTION, we delete file in list by iterating all levels.
So in this pr, we delete level file list by passing level parameter to deleteLevelFiles function.


qiaojialin (approved):

qiaojialin:
has cherry-picked to rel/0.11

PR #2153, opened 2020-12-01
change mem control lock usage

HTHou:


server/src/main/java/org/apache/iotdb/db/exception/WriteProcessRejectException.java
```diff
@@ -0,0 +1,35 @@
...
+package org.apache.iotdb.db.exception;
+
+import org.apache.iotdb.rpc.TSStatusCode;
+
+public class WriteProcessRejectException extends WriteProcessException {
+
+  private static final long serialVersionUID = -4217324287547595610L;
+
+  public WriteProcessRejectException(String message) {
+    super(message, TSStatusCode.WRITE_PROCESS_ERROR.getStatusCode());
```
jixuan1989, line 28: add a new code?

jixuan1989 (changes requested):
why not add a new error code for write_reject_because_overload?

server/src/main/java/org/apache/iotdb/db/exception/WriteProcessRejectException.java
HTHou: Good Idea!

qiaojialin (approved):

jixuan1989 (approved):

jixuan1989:
Is this related to some known bug? If so, we need to cherry-pick to rel/0.11, otherwise  please delete `mem_control_op` branch

PR #2154, opened 2020-12-01
[Mem control] Move one of the insert check out from the sg lock

HTHou:


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -710,10 +717,17 @@ public void insert(InsertRowPlan insertRowPlan) throws WriteProcessException {
    */
   @SuppressWarnings("squid:S3776") // Suppress high Cognitive Complexity warning
   public void insertTablet(InsertTabletPlan insertTabletPlan) throws BatchInsertionException {
+    TSStatus[] results = new TSStatus[insertTabletPlan.getRowCount()];
```
mychaow, line 720: I think it's better to move the two line after blockInsertionIfReject  for that we need not to create them before we get memory.

HTHou: Sure~

mychaow (approved):

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -710,6 +717,15 @@ public void insert(InsertRowPlan insertRowPlan) throws WriteProcessException {
    */
   @SuppressWarnings("squid:S3776") // Suppress high Cognitive Complexity warning
   public void insertTablet(InsertTabletPlan insertTabletPlan) throws BatchInsertionException {
+    if (enableMemControl) {
+      try {
+        blockInsertionIfReject();
+      } catch (WriteProcessException e) {
+        TSStatus[] results = new TSStatus[insertTabletPlan.getRowCount()];
+        Arrays.fill(results, RpcUtils.SUCCESS_STATUS);
```
jixuan1989, line 725: SUCCESS_STATUS？

HTHou: That is what I am confused...

HTHou: That is a mistake, I'll fix it. Thanks.

qiaojialin (approved):

PR #2155, opened 2020-12-01
[IOTDB-1034] Show timeseries error in Chinese on Windows

Alima777:
Change the charset of Windows console to UTF-8 does not fix the problem, since the column header is wrong again...
We'd better unify the charset to "UTF-8" instead of letting it depend on the platform.


qiaojialin (approved):

qiaojialin:
has cherry-picked to rel/0.11

PR #2164, opened 2020-12-02
Remove static resps

Genius-pig:
remove static response which‘s header info is duplicate to the dataset header info.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
+++ b/server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
@@ -18,8 +18,6 @@
  */
 package org.apache.iotdb.db.service;
 
-import static org.apache.iotdb.db.qp.physical.sys.ShowPlan.ShowContentType.TIMESERIES;
-
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.sql.SQLException;
@@ -650,7 +648,6 @@ public class TSServiceImpl implements TSIService.Iface, ServerContext {
     long startTime = System.currentTimeMillis();
     long queryId = -1;
     try {
-      TSExecuteStatementResp resp = getQueryResp(plan, username); // column headers
 
       // In case users forget to set this field in query, use the default value
       if (fetchSize == 0) {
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -725,19 +719,30 @@ private TSExecuteStatementResp internalExecuteQueryStatement(String statement,
...
+
+      if (plan instanceof QueryPlan) {
+        resp = getQueryColumnHeaders(plan, username);
+      }
       // create and cache dataset
       QueryDataSet newDataSet = createQueryDataSet(queryId, plan);
+
+      if (plan instanceof ShowPlan || plan instanceof AuthorPlan) {
+        resp = getListDataResp(newDataSet);
+      }
```
Alima777, line 733: Rename this method as `getListDataSetHeaders` to keep the consistence.

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -725,19 +719,30 @@ private TSExecuteStatementResp internalExecuteQueryStatement(String statement,
       if (plan instanceof AuthorPlan) {
         plan.setLoginUserName(username);
       }
+
+      TSExecuteStatementResp resp = null;
+
+      if (plan instanceof QueryPlan) {
+        resp = getQueryColumnHeaders(plan, username);
+      }
```
Alima777, line 727:
```suggestion
      // execute it before createDataSet since it may change the content of query plan
      if (plan instanceof QueryPlan) {
        resp = getQueryColumnHeaders(plan, username);
      }
```

Alima777 (commented):
LGTM

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
Genius-pig: fixed

jixuan1989:
Should `paths` in QueryDataSet  be `columns`?

Alima777 (approved):

Alima777:
> Should `paths` in QueryDataSet be `columns`?

`paths` is for data fetching while `columns` is metadata.

PR #2168, opened 2020-12-03
[Distributed]limit the max read frame size

neuyilan:


mychaow (approved):

cluster/src/main/java/org/apache/iotdb/cluster/client/rpcutils/TElasticFramedTransport.java
```diff
@@ -26,6 +26,13 @@
 
 public class TElasticFramedTransport extends TFastFramedTransport {
 
+  /**
+   * It is used to prevent the size of the parsing package from being too large and allotting the
```
mychaow, line 30: allotting -> allocating, which is more universal

mychaow (changes requested):

cluster/src/main/java/org/apache/iotdb/cluster/client/rpcutils/TElasticFramedTransport.java
neuyilan: Thanks

mychaow (approved):

PR #2170, opened 2020-12-03
set max number of open files in start-node.sh

yanhongwangg:


neuyilan (approved):

LebronAl (approved):
LGTM

cluster/src/assembly/resources/sbin/start-node.sh
```diff
@@ -18,6 +18,8 @@
 # under the License.
 #
 
+# Set max number of open files
+ulimit -n 65535
```
liutaohua, line 22: What happens if the current user is not `root`?

liutaohua: some suggestions but not blocking:
1. check whether the current user is `root`
2. check if the current `ulimit` is greater than `65535`
3. Add obvious hints that require manual modification

jt2594838:
This should better be put into env.sh instead of start-node.sh, so it may affect both stand-alone and cluter verison.

jt2594838:
By the way, we did not add this because we thought that it was improper for an application to modify OS configuration without a proper request to the user.

mychaow:
> By the way, we did not add this because we thought that it was improper for an application to modify OS configuration without a proper request to the user.

I think we could do this for user. The user may not be familiar with the system, but if they have used it in production system, it's too late to modify the limit when system get error. 


How do you think? @jixuan1989

jt2594838:
Maybe we can ask the user like this:
```
./start-node.sh
Warning: System maximum open file number is too small, IoTDB wishes to set it to 65535.[y/n]
```

mychaow:
> Maybe we can ask the user like this:
> 
> ```
> ./start-node.sh
> Warning: System maximum open file number is too small, so IoTDB wishes to set it to 65535.[y/n]
> ```

Good idea.

Issue #2179, opened 2020-12-03
NullPointerException in Tablet.getValueBytesSize(Tablet.java:239)


qiaojialin:
`
    session = new Session("127.0.0.1", 6667, "root", "root");
    session.open();

    List<MeasurementSchema> schemaList = new ArrayList<>();
    schemaList.add(new MeasurementSchema("s1", TSDataType.INT64));
    schemaList.add(new MeasurementSchema("s2", TSDataType.DOUBLE));
    schemaList.add(new MeasurementSchema("s3", TSDataType.TEXT));

    Tablet tablet = new Tablet("root.sg.d", schemaList, 100);

    long timestamp = System.currentTimeMillis();

    for (long row = 0; row < 10; row++) {
      int rowIndex = tablet.rowSize++;
      tablet.addTimestamp(rowIndex, timestamp);
      tablet.addValue("s1", rowIndex, 1L);
      tablet.addValue("s2", rowIndex, 1D);
      tablet.addValue("s3", rowIndex, new Binary("1"));
      if (tablet.rowSize == tablet.getMaxRowNumber()) {
        session.insertTablet(tablet, true);
        tablet.reset();
      }
      timestamp++;
    }

    if (tablet.rowSize != 0) {
      session.insertTablet(tablet);
      tablet.reset();
    }`

Then, you will get a NullPointerException

qiaojialin:
Fixed in master and 0.11.1

PR #2180, opened 2020-12-03
fix insert partial tablet with binary NullPointer bug

qiaojialin:
When using insertTablet, we create a Binary[100], then insert 10 values.
The getValueBytesSize do not use the rowCount 10 when calculating the size.

#2179


jixuan1989 (approved):

qiaojialin:
has cherry-picked to rel/0.11

PR #2186, opened 2020-12-04
Fix desc batchdata count bug

Alima777:
When Batchdata is from mergeReader, the sequence of it will be in reverse. E.g. [3000, 2001], [2000, 1001], [1000, 1]

Then getMaxTimestamp should get the first but actually it got the last timestamp. That's the wrong place.

To keep it better to understand, I abstracted a new batchdata called DescReadWriteBatchdata. It's for mergeReader in `order by time desc`. And it will write data in reverse, i.e. the sequence will be [1, 1000], [1001, 2000], [2001, 3000]


tsfile/src/main/java/org/apache/iotdb/tsfile/read/common/DescReadWriteBatchData.java
```diff
@@ -0,0 +1,374 @@
...
+ * This class is for reading and writing batch data in reverse. The data source is from mergeReader.
+ * For example,
+ * the time sequence from mergeReader is 1000 -> 1, to keep the consistency that the timestamp
+ * should be ascending. It will be written in reverse, i.e. the timeRet will be [1, 1000].
+ * Then it can be handled the same as DescReadBatchData.
+ */
+public class DescReadWriteBatchData extends DescReadBatchData {
+
+  public DescReadWriteBatchData(TSDataType dataType) {
+    super();
```
jixuan1989, line 36: change to 
```
super(dataType);
this.writeCurArrayIndex = capacity - 1;
```

jixuan1989:
try to remove `Serializable` on BatchData

jixuan1989:
and submit a patch to rel/0.11

jixuan1989 (approved):

PR #2189, opened 2020-12-04
[Distributed] Make cluster configuration consensus with the single-node

jixuan1989:
I remove iotdb-engines.properties in cluster/src/assemby,
one more issue left:

`thrift_init_buffer_size=1024` is not used.


neuyilan (commented):
I think we can reserve the `thrift_init_buffer_size` parameter, as this is mainly used for the initial thrift frame buffer size, which should be used in the `org.apache.iotdb.rpc.RpcUtils`, I think this parameter is missed when developing this codes. 

besides, those code smells are better to solve

neuyilan (changes requested):
besides, those code smells are better to solve

mychaow (approved):
LGTM

jixuan1989:
> I think we can reserve the `thrift_init_buffer_size` parameter, as this is mainly used for the initial thrift frame buffer size, which should be used in the `org.apache.iotdb.rpc.RpcUtils`, I think this parameter is missed when developing this codes.

But currently there is no any function calling the parameter..

jt2594838:
That one is free to remove, as it makes little difference especially in long runs, and the default values should be enough.

PR #2209, opened 2020-12-07
Avoid overflow by converting to long

asdf2014:


server/src/main/java/org/apache/iotdb/db/utils/datastructure/TVList.java
```diff
@@ -82,11 +82,9 @@ public static TVList newList(TSDataType dataType) {
   public static long tvListArrayMemSize(TSDataType type) {
     long size = 0;
     // time size
-    size +=
-        PrimitiveArrayManager.ARRAY_SIZE * 8;
+    size += (long) PrimitiveArrayManager.ARRAY_SIZE * 8L;
```
jixuan1989, line 85: Hi, is this really needed? 
I think `PrimitiveArrayManager.ARRAY_SIZE * 8` is far less than  Integer.MAX_VALUE.
(even the overflow may occurs,  just changing `8` to `8L` is fine, right?)

JackieTien97 (approved):

PR #2215, opened 2020-12-08
[To rel/0.11] add max frame size configuration

neuyilan:


jixuan1989 (commented):
The client does not need any modifications, right?

neuyilan:
> The client does not need any modifications, right?

yes, I have tested using the benchmark.

jt2594838 (approved):

PR #2222, opened 2020-12-09
[Distributed] fix session cache leader interface and docs 

LebronAl:
As cluster_new is  merged into master, It is better not to open the leader cache by default and give an parameter interface for distributed IoTDB, otherwise it will cause unnecessary performance waste to the stand-alone IoTDB.


session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
```diff
@@ -75,6 +75,7 @@
   private long timeout; //ms
   private static int FINAL_RETRY = RETRY - 1;
   private boolean enableCompression = false;
+  private boolean enableCacheLeader = false;
```
jt2594838, line 78: Where is this set?

jt2594838 (approved):

session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
LebronAl: User can use `public Session(String host, int rpcPort, String username, String password,
      boolean isEnableCacheLeader)` interface to enable leader cache

LebronAl: OK, I will add this interfance for SessionPool~

session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
```diff
@@ -74,29 +74,36 @@
...
-  private boolean enableCompression = false;
+  private boolean enableCompression;
+  private boolean enableCacheLeader;
   private ZoneId zoneId;
 
   private boolean closed;//whether the queue is closed.
 
   public SessionPool(String ip, int port, String user, String password, int maxSize) {
-    this(ip, port, user, password, maxSize, Config.DEFAULT_FETCH_SIZE, 60_000, false, null);
+    this(ip, port, user, password, maxSize, Config.DEFAULT_FETCH_SIZE, 60_000, false, null, false);
```
jixuan1989, line 84: `false` -> `Config.DEFAULT_CACHE_LEADER_MODE`

session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
```diff
@@ -74,29 +74,36 @@
...
   }
 
   public SessionPool(String ip, int port, String user, String password, int maxSize,
       boolean enableCompression) {
     this(ip, port, user, password, maxSize, Config.DEFAULT_FETCH_SIZE, 60_000, enableCompression,
-        null);
+        null, false);
```
jixuan1989, line 90: `false` -> `Config.DEFAULT_CACHE_LEADER_MODE`

session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
```diff
@@ -74,29 +74,36 @@
...
+  public SessionPool(String ip, int port, String user, String password, int maxSize,
+      boolean enableCompression, boolean enableCacheLeader) {
+    this(ip, port, user, password, maxSize, Config.DEFAULT_FETCH_SIZE, 60_000, enableCompression,
+        null, enableCacheLeader);
   }
 
   public SessionPool(String ip, int port, String user, String password, int maxSize,
       ZoneId zoneId) {
-    this(ip, port, user, password, maxSize, Config.DEFAULT_FETCH_SIZE, 60_000, false, zoneId);
+    this(ip, port, user, password, maxSize, Config.DEFAULT_FETCH_SIZE, 60_000, false, zoneId, false);
```
jixuan1989, line 101: `false` -> `Config.DEFAULT_CACHE_LEADER_MODE`

session/src/test/java/org/apache/iotdb/session/pool/SessionPoolTest.java
```diff
@@ -164,7 +164,7 @@ private void correctQuery(SessionPool pool) {
 
   @Test
   public void tryIfTheServerIsRestart() {
-    SessionPool pool = new SessionPool("127.0.0.1", 6667, "root", "root", 3, 1, 6000, false, null);
+    SessionPool pool = new SessionPool("127.0.0.1", 6667, "root", "root", 3, 1, 6000, false, null, false);
```
jixuan1989, line 167: Is there some UT for set the enableCache as `true`?

jixuan1989 (changes requested):
some minor changes

session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
LebronAl: OK

session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
LebronAl: OK

session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
LebronAl: OK

session/src/test/java/org/apache/iotdb/session/pool/SessionPoolTest.java
LebronAl: I will add UT for this laterly

session/src/main/java/org/apache/iotdb/session/SessionConnection.java
```diff
@@ -67,6 +67,9 @@
   private ZoneId zoneId;
   private EndPoint endPoint;
 
+  //TestOnly
+  public SessionConnection(){}
```
jt2594838, line 71: It would be better to use an annotation, and please format the code.

jt2594838 (approved):

session/src/main/java/org/apache/iotdb/session/SessionConnection.java
LebronAl: Testonly's anonation is currently in the server module. The session module does not currently have a dependency on the server module, so I don't think it would be appropriate to introduce a dependency just for this purpose. As for creating a new testonly anotation in the session module, I think it's a bit redundant because it's the only one we've used so far. But if you think it's worth creating a new anonation in the Session Module, I'll do it. What your opinion?

jt2594838: That is the case, we can extract that into another module later.

PR #2229, opened 2020-12-09
[IOTDB-1049] Fix Nullpointer exception and a delete bug in Last query

wshao08:
This PR actually fixed two bugs related to Last Query:

- Nullpointer exception when querying Last: https://issues.apache.org/jira/browse/IOTDB-1049
- Querying Last of a timeseries after deletion gives incorrect result, because the invalid statistics is still used.

### DIFF
--- a/server/src/test/java/org/apache/iotdb/db/integration/IoTDBLastIT.java
+++ b/server/src/test/java/org/apache/iotdb/db/integration/IoTDBLastIT.java
@@ -42,8 +42,7 @@ import org.junit.Test;
 
 public class IoTDBLastIT {
 
-  private static String[] dataSet1 = new String[]{
-      "SET STORAGE GROUP TO root.ln.wf01.wt01",
+  private static final String[] dataSet1 = new String[]{
       "CREATE TIMESERIES root.ln.wf01.wt01.status WITH DATATYPE=BOOLEAN, ENCODING=PLAIN",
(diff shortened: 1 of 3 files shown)


server/src/test/java/org/apache/iotdb/db/integration/IoTDBLastIT.java
```diff
@@ -103,6 +100,30 @@ public void tearDown() throws Exception {
...
+            "root.ln.wf02.temperature,null",
+        };
+
+    try (Connection connection =
+             DriverManager.getConnection("jdbc:iotdb://127.0.0.1:6667/", "root", "root");
+         Statement statement = connection.createStatement()) {
+
+      statement.execute(
+          "CREATE TIMESERIES root.ln.wf02.temperature WITH DATATYPE=DOUBLE, ENCODING=PLAIN");
+      statement.execute("select last temperature from root.ln.wf02");
```
jixuan1989, line 115: how about select two timeseries in a sql, while one time series has not data and another one has data.

qiaojialin (approved):

server/src/test/java/org/apache/iotdb/db/integration/IoTDBLastIT.java
qiaojialin: could add one more test for this

PR #2237, opened 2020-12-10
add print tsfile specfic path

EJTTianYu:
when users like me would like to analyse the specfic measurement data distribution of the tsfile file, it may help


jixuan1989 (changes requested):
1. Doc needed
2. Test needed
3. Don't we have such a similar tool?

EJTTianYu:
> 1. Doc needed
> 2. Test needed
> 3. Don't we have such a similar tool?

1. Doc added
2. as the tool has just a main function, I think test can be skipped 
3. we have `TsFileSequenceRead`, while the class print all the data in the tsfile. As far as I know, no other tools

jixuan1989:
> as the tool has just a main function, I think test can be skipped

If no test, how to guarantee the main() is keeping maintained well?

jixuan1989:
> we have TsFileSequenceRead(the tool currently does not provide script）, while the class print all the data in the tsfile. As far as I know, no other tools

see `server/src/assembly/resources/tools/tsfileToolSet`.

EJTTianYu:
> > we have TsFileSequenceRead(the tool currently does not provide script）, while the class print all the data in the tsfile. As far as I know, no other tools
> 
> see `server/src/assembly/resources/tools/tsfileToolSet`.

Before I developed the tool, I had checked the scripts under the server dir. No tool can print out chunk data like this.

EJTTianYu:
> > as the tool has just a main function, I think test can be skipped
> 
> If no test, how to guarantee the main() is keeping maintained well?

test added

server/src/main/java/org/apache/iotdb/db/tools/TsFileSinglePathRead.java
```diff
@@ -0,0 +1,64 @@
...
+
+  public static void main(String[] args) throws IOException {
+    String filename = "test.tsfile";
+    String path = "root.vehicle.d0.s0";
+    if (args.length >= 2) {
+      filename = args[0];
+      path = args[1];
+    } else if (args.length >= 1){
+      filename = args[0];
+    }
```
JackieTien97, line 41: Maybe if the args.length is less than 2, we should throw an exception to tell users to give the right paremeters.

server/src/test/java/org/apache/iotdb/db/tools/TsFileSinglePathReadTest.java
```diff
@@ -0,0 +1,125 @@
...
+      for (ChunkMetadata metadata : chunkMetadataList) {
+        ChunkReader chunkReader = new ChunkReader(reader.readMemChunk(metadata), null);
+        ChunkDataIterator chunkDataIterator = new ChunkDataIterator(chunkReader);
+        while (chunkDataIterator.hasNextTimeValuePair()) {
+          chunkDataIterator.nextTimeValuePair();
+          numCnt ++;
+        }
+      }
+    }
+    assertEquals(numCnt, 2);
```
JackieTien97, line 123: I think you should not write the same code again, you can extract the code in TsFileSinglePathRead as a public method, and in this UT, you can just test that method. Or if the class has been modified, this test may be forgotten to be changed.

tsfile/src/main/java/org/apache/iotdb/tsfile/read/TsFileSequenceReader.java
```diff
@@ -1023,6 +1023,27 @@ public int getTotalChunkNum() {
...
   }
 
+  /**
+   * get ChunkMetaDatas of given path sorting by offset
+   *
+   * @param path         timeseries path
+   * @param sortByOffset whether to use offset to sort the ChunkMetadatas
+   * @return List of ChunkMetaData
+   */
+  public List<ChunkMetadata> getChunkMetadataList(Path path, boolean sortByOffset)
```
JackieTien97, line 1033: I've no idea whether the `sortByOffset` parameter is necessary, because as far as I known, the ChunkMetadataList has been sorted already in the disk.

JackieTien97 (changes requested):
Great Tool, there may be some places to be fixed.

server/src/main/java/org/apache/iotdb/db/tools/TsFileSinglePathRead.java
EJTTianYu: done

server/src/test/java/org/apache/iotdb/db/tools/TsFileSinglePathReadTest.java
EJTTianYu: done

tsfile/src/main/java/org/apache/iotdb/tsfile/read/TsFileSequenceReader.java
EJTTianYu: done

PR #2267, opened 2020-12-14
Update document new

samginshen:
add update web site ci yaml. when change docs files in master branch , will run this test. And change site/pom.xml. When run mvn complie, it will build compile-site-*.

### DIFF
--- a/.github/workflows/update_website.yml
+++ b/.github/workflows/update_website.yml
@@ -1,0 +1,27 @@
+name: Update website
+
+on:
+  push:
+    branches:
+      - master
+    paths:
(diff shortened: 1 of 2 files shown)


jixuan1989:
This workflow does not work.

`mvn scm:publish` only  accepts username and password that belong to some Apache committer.

You have to manually upload changes to the asf-site branch of github.com/apache/iotdb-website repo.

By the way, I have added a task to build the website everyday (no matter whether there is changes) using Jenkins. 
So, if we really can not implement it with github action, it is also OK.

PR #2272, opened 2020-12-16
fix file not found when restart

mychaow:


HTHou (approved):
I have tried this fix on `UpgradeToolV3` branch and it looks good!

PR #2276, opened 2020-12-16
[IOTDB-1060] Support full deletion for delete statement without where clause

wshao08:
This PR supports to delete all data in a timeseries if there is no where clause in delete statements.
```
delete from root.sg.d1.s0
```


JackieTien97 (approved):

qiaojialin:
has cherry-picked to rel/0.11

PR #2277, opened 2020-12-17
[Distributed] fix client is null cause NullPointerException

neuyilan:
if the node can not be connected, the server will cause NullPointerException, this pr try to solve the problem.


jt2594838 (approved):

jt2594838:
We have some trouble with the CI due to the Hive connector. Please manually run tests of cluster manually to confirm. Sorry for the trouble.

neuyilan:
> We have some trouble with the CI due to the Hive connector. Please manually run tests of cluster manually to confirm. Sorry for the trouble.

I have run tests using the[1] command, and all tests have passed. 
[1] `mvn clean package install -pl cluster,cli -am -Drat.skip=true`

PR #1721, opened 2020-09-10
[IOTDB-868] fix comma bug for mlog

mychaow:
After we import the double-quoted path, users may have any characters in the path that includes ",". 

If we still record the mlog using csv format, we could not operate correctly to paths like

root.changde.e34c49.struct@VALUE%PRESENT_ANGLE,_CUTTING_HEAD]

The mlog needs to be encoded.

https://issues.apache.org/jira/browse/IOTDB-868

PR #2309, opened 2020-12-22
Fix sync schema bug

fanhualta:
In pr [#1721](https://github.com/apache/iotdb/pull/1721), a bug of schema synchronization in the sync module is introduced. Which will lead to that the schema can not be loaded correctly on the receiver. This pr solves this problem.


HTHou:
There are two sonar bugs...

HTHou (approved):

PR #2315, opened 2020-12-23
[IOTDB-1069] restrict the flushing memtable number to avoid OOM when mem_control is disabled

HTHou:
To reproduce this bug, 
set JVM memory size to 32GB, 
tsfile_size_threshold=268435456
memtable_size_threshold=268435456
enable_mem_control=false
and insert 200 timeseries to IoTDB.

This pr restricts the flushing memtable size in one TsfileProcessor to one.


server/src/main/java/org/apache/iotdb/db/rescon/MemTableManager.java
```diff
@@ -0,0 +1,110 @@
...
+
+  /**
+   * Called when memory control is disabled
+   */
+  public synchronized IMemTable getAvailableMemTable(TsFileResource tsFileResource) {
+    if (!reachMaxMemtableNumber()) {
+      currentMemtableNumber++;
+      return new PrimitiveMemTable();
+    } else {
+      // wait until the size of flushingMemTables is less than 2
```
qiaojialin, line 56:
```suggestion
      // wait until the total number of memtable is less than the system capacity
```

qiaojialin (approved):

PR #2316, opened 2020-12-23
[IOTDB-1068] Fix Time series metadata cache bug

JackieTien97:
Currently, we add the entry size each time we put it into cache map. However, if the key already exists in cache, we will add the size again which will cause the cache used size is higher than the actual size.

Also, in this PR, I improve the process of deserializing time series metadata cache from disk.


samperson1997 (approved):

qiaojialin (approved):
Add  a  test

qiaojialin (approved):

qiaojialin (approved):

weizihan0110:
> Currently, we add the entry size each time we put it into cache map. However, if the key already exists in cache, we will add the size again which will cause the cache used size is higher than the actual size.
> 
> Also, in this PR, I improve the process of deserializing time series metadata cache from disk.

Hi, a PR should mainly  contain one changing intention. It's better to open another PR to improve the process of deserializing time series metadata cache from disk.

PR #2317, opened 2020-12-23
[To rel/0.11] [IOTDB-1069] restrict the flushing memtable number to avoid OOM when mem_control is disabled 

HTHou:
Same as https://github.com/apache/iotdb/pull/2315

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/rescon/MemTableManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/rescon/MemTableManager.java
@@ -79,7 +79,7 @@ public class MemTableManager {
     currentMemtableNumber++;
   }
 
-  public synchronized void resetMemtableNumber() {
+  public synchronized void decreaseMemtableNumber() {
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/apache/iotdb/db/rescon/MemTableManager.java
```diff
@@ -0,0 +1,110 @@
...
+
+  /**
+   * Called when memory control is disabled
+   */
+  public synchronized IMemTable getAvailableMemTable(TsFileResource tsFileResource) {
+    if (!reachMaxMemtableNumber()) {
+      currentMemtableNumber++;
+      return new PrimitiveMemTable();
+    } else {
+      // wait until the size of flushingMemTables is less than 2
```
qiaojialin, line 56:
```suggestion
      // wait until the total memtable number is less than the capacity of the system
```

qiaojialin (approved):

PR #2318, opened 2020-12-23
ci: JDK 8, JDK 11, JDK 15

sullis:


kr11:
Hi @sullis,  the failure in Java 15 is due to IoTDB using a too old version of Felix(4.2.1).
It has been reported at [FELIX-6259](https://issues.apache.org/jira/browse/FELIX-6259). 

A suggestion: upgrade Felix to at least `5.1.1` or higher at `$IoTDB_HOME/pom.xml:138`
```xml
        ...
        <metrics.version>3.2.6</metrics.version>
        <javax.xml.bind.version>2.4.0-b180725.0427</javax.xml.bind.version>
        <felix.version>4.2.1</felix.version> <!-- upgrade to 5.1.1 or higher-->
       ...
```
I have passed the compile in my Mac OS with Java 15.

PS:  the bug of Felix is caused by using an old version of bnd ([bnd-3903](https://github.com/bndtools/bnd/issues/3903)) .

HTHou:
Hi, since the JDK 15 is not a LTS version, I don't think it's necessary to add CI for this version. 
We may going to support the next LTS version, which is JDK 17.

jixuan1989:
Hi, thanks for raising this up. have done in #2603

PR #2331, opened 2020-12-24
[Security] Bump commons-collections4 from 4.0 to 4.4 to avoid a security bug

asdf2014:


asdf2014:
FYI, [https://github.com/advisories/GHSA-6hgm-866r-3cjv](https://github.com/advisories/GHSA-6hgm-866r-3cjv)

jixuan1989:
Hi @asdf2014 , would you like to upgrade all dependencies that CVE checker reports?
If so, we can have a test to see whether the system is still stable and the performance is kept after the upgrade.

asdf2014:
Hi, @jixuan1989 . Thanks for your comment. Normally, we should bump up a single dependency at a time to control the affected area. FYI, [https://github.com/apache/druid/pulls?q=is%3Apr+label%3A%22Area+-+Dependencies%22](https://github.com/apache/druid/pulls?q=is%3Apr+label%3A%22Area+-+Dependencies%22). For now, I only found this security bug in the ACC library, and don't know how to pull the entire list from CVE, so maybe we should raise up a new PR when we find anther security vulnerabilities.

HTHou (approved):

PR #2336, opened 2020-12-25
[IOTDB-1078] Save one unnecessary RPC call while the result is empty

JackieTien97:


jixuan1989 (approved):

jixuan1989:
cherry-pick to rel/0.11?

PR #2350, opened 2020-12-27
[To rel/0.11] [IOTDB-1069] restrict the flushing memtable number to avoid OOM when mem_control is disabled

HTHou:
See https://github.com/apache/iotdb/pull/2315

<hr>

This PR has:
- [x] been self-reviewed.
    - [x] concurrent read
    - [x] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/rescon/MemTableManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/rescon/MemTableManager.java
@@ -22,6 +22,7 @@ import org.apache.iotdb.db.conf.IoTDBConfig;
 import org.apache.iotdb.db.conf.IoTDBDescriptor;
 import org.apache.iotdb.db.engine.memtable.IMemTable;
 import org.apache.iotdb.db.engine.memtable.PrimitiveMemTable;
+import org.apache.iotdb.db.exception.WriteProcessException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -44,8 +45,10 @@ public class MemTableManager {
 
   /**
    * Called when memory control is disabled
+   * @throws WriteProcessException 
    */
-  public synchronized IMemTable getAvailableMemTable(String storageGroup) {
+  public synchronized IMemTable getAvailableMemTable(String storageGroup) 
+      throws WriteProcessException {
     if (!reachMaxMemtableNumber()) {
       currentMemtableNumber++;
       return new PrimitiveMemTable();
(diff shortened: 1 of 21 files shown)


server/src/main/java/org/apache/iotdb/db/rescon/MemTableManager.java
```diff
@@ -0,0 +1,109 @@
...
+    } else {
+      // wait until the total memtable number is less than the capacity of the system
+      int waitCount = 1;
+      while (true) {
+        if (!reachMaxMemtableNumber()) {
+          currentMemtableNumber++;
+          return new PrimitiveMemTable();
+        }
+        try {
+          wait(WAIT_TIME);
```
SilverNarcissus, line 61: We shouldn't use wait without notify. We should notifyAll at "decreaseMemtableNumber"

server/src/main/java/org/apache/iotdb/db/rescon/MemTableManager.java
```diff
@@ -0,0 +1,109 @@
...
+        } catch (InterruptedException e) {
+          logger.error("{} fails to wait for memtables {}, continue to wait", storageGroup, e);
+          Thread.currentThread().interrupt();
```
SilverNarcissus, line 64: After interrupt. We should also throw the exception. Or this will loop forever without wait. (When enter wait with interrupt, wait will just return without wait.)

server/src/main/java/org/apache/iotdb/db/rescon/MemTableManager.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  /**
+   * Called when memory control is disabled
+   */
+  public synchronized void addOrDeleteStorageGroup(int diff) {
+    int maxMemTableNum = CONFIG.getMaxMemtableNumber();
+    maxMemTableNum += MEMTABLE_NUM_FOR_EACH_PARTITION 
+        * CONFIG.getConcurrentWritingTimePartition() * diff;
+    CONFIG.setMaxMemtableNumber(maxMemTableNum);
```
SilverNarcissus, line 99: Also notifyAll at here

server/src/main/java/org/apache/iotdb/db/rescon/MemTableManager.java
```diff
@@ -0,0 +1,109 @@
...
+  public synchronized int getCurrentMemtableNumber() {
+    return currentMemtableNumber;
+  }
+
+  public synchronized void addMemtableNumber() {
+    currentMemtableNumber++;
+  }
+
+  public synchronized void decreaseMemtableNumber() {
+    currentMemtableNumber--;
```
SilverNarcissus, line 82: NotifyAll at here

SilverNarcissus (changes requested):
Nice work!  Only some concurrent issue. Please have a look

server/src/main/java/org/apache/iotdb/db/rescon/MemTableManager.java
HTHou: Make sense~

SilverNarcissus (approved):
Look good to me. It can be merged as soon as CI is passed.

PR #2358, opened 2020-12-28
[IOTDB-1084] Fix temporary memory of flushing may cause OOM

HTHou:
<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


SilverNarcissus (approved):
Look good to me~

server/src/main/java/org/apache/iotdb/db/engine/flush/MemTableFlushTask.java
```diff
@@ -170,94 +189,104 @@ private void writeOneSeries(TVList tvPairs, IChunkWriter seriesWriterImpl,
     @Override
     public void run() {
       long memSerializeTime = 0;
```
SilverNarcissus, line 191: You should use class variable `memSerializeTime` rather than this local variable

server/src/main/java/org/apache/iotdb/db/engine/flush/MemTableFlushTask.java
```diff
@@ -43,18 +46,23 @@
...
+          : new LinkedBlockingQueue<>();
+
   private String storageGroup;
 
   private IMemTable memTable;
 
-  private volatile boolean noMoreEncodingTask = false;
-  private volatile boolean noMoreIOTask = false;
+  private long memSerializeTime = 0L;
+  private long ioTime = 0L;
```
SilverNarcissus, line 65: These two variable will be write and read by different thread without locking. The situation is only one thread to write, so no locking is reasonable, but they should be volatile so they can be seen by another thread.

server/src/main/java/org/apache/iotdb/db/engine/flush/MemTableFlushTask.java
```diff
@@ -170,94 +189,104 @@ private void writeOneSeries(TVList tvPairs, IChunkWriter seriesWriterImpl,
...
+      
+      LOGGER.debug("Storage group {}, flushing memtable {} into disk: Encoding data cost "
+          + "{} ms.",
           storageGroup, writer.getFile().getName(), memSerializeTime);
     }
   };
 
   @SuppressWarnings("squid:S135")
   private Runnable ioTask = () -> {
     long ioTime = 0;
```
SilverNarcissus, line 246: You should use class variable ioTime rather than this local variable

SilverNarcissus (changes requested):
Fantastic work! But there are some issues, please have a look.

server/src/main/java/org/apache/iotdb/db/engine/flush/MemTableFlushTask.java
HTHou: Fixed~

server/src/main/java/org/apache/iotdb/db/engine/flush/MemTableFlushTask.java
HTHou: Fixed~

server/src/main/java/org/apache/iotdb/db/engine/flush/MemTableFlushTask.java
HTHou: Fixed~

SilverNarcissus (approved):
Look good to me~

PR #2384, opened 2020-12-30
refactor series reader to make it small

mychaow:
### Content3 ...

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/query/reader/series/SeriesReader.java
+++ b/server/src/main/java/org/apache/iotdb/db/query/reader/series/SeriesReader.java
@@ -18,16 +18,6 @@
  */
 package org.apache.iotdb.db.query.reader.series;
 
-import java.io.IOException;
-import java.io.Serializable;
-import java.util.Comparator;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Objects;
-import java.util.PriorityQueue;
-import java.util.Set;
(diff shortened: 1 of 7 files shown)


server/src/main/java/org/apache/iotdb/db/query/reader/series/SeriesReader.java
```diff
@@ -18,16 +18,6 @@
  */
 package org.apache.iotdb.db.query.reader.series;
```
jixuan1989, line 20: avoid more changes like this in the future.

jixuan1989:
Why for this improvement?

mychaow:
> Why for this improvement?

I think SeriesReader is too long,  move the utils class to separate class will make it looks more clear.

server/src/main/java/org/apache/iotdb/db/query/reader/series/SeriesReader.java
mychaow: Ok

Alima777:
Hi, From the perspective of OOP, TimeOrderUtils is only used in `SeriesReader` and should be better to be the inner class of it. There is no need to move it for the too long reason.

Issue #2419, opened 2021-01-05
Server test logs contains lots of useless content


Alima777:
Since lots of useless content (such as meaningless exception message, and even System.println) exsits in our server test logs, it would be better to clear it.

PR #2420, opened 2021-01-05
[ISSUE-2419] Server test logs contains lots of useless content

Alima777:
#2419 

After improvement, server test logs are reduced by about 5000 lines, from 24000+ lines to 19000+lines.


cluster/src/test/java/org/apache/iotdb/cluster/query/BaseQueryTest.java
```diff
@@ -68,6 +69,8 @@ protected static void checkAggregations(List<AggregateResult> aggregationResults
 
   @Before
   public void setUp() throws Exception {
+    IoTDBDescriptor.getInstance().getConfig()
+        .setCompactionStrategy(CompactionStrategy.NO_COMPACTION);
```
JackieTien97, line 73: Why changing the compaction strategy, if we don't change, will this test failed?

JackieTien97 (approved):

PR #2421, opened 2021-01-05
Make serialization pool in LogDispatcher static to reduece the number of pools

jt2594838:


mychaow (approved):
LGTM

LebronAl (approved):
LGTM

cluster/src/main/java/org/apache/iotdb/cluster/log/LogDispatcher.java
```diff
@@ -68,12 +69,13 @@
   private List<BlockingQueue<SendLogRequest>> nodeLogQueues =
       new ArrayList<>();
   private ExecutorService executorService;
```
neuyilan, line 71: LGTM!
only one question, why only change the _serializationService_ to use google's ThreadFactoryBuilder,
Why not change the _executorService_? Is there any special thinking?

neuyilan (approved):

cluster/src/main/java/org/apache/iotdb/cluster/log/LogDispatcher.java
jt2594838: Sure I can do this, but I was focusing on serializationService only and ignored executorService. I will modify it too.

jt2594838: I manually set the thread name in DispatcherThread, so the receiver's name can be included in the thread name, which is not convenient when using ThreadFactoryBuilder.

PR #2427, opened 2021-01-05
fix the case that column length is not right in CLI when the value contains Chinese characters

weizihan0110:
## Result

Before:
image

After:
image

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/cli/src/main/java/org/apache/iotdb/cli/AbstractCli.java
+++ b/cli/src/main/java/org/apache/iotdb/cli/AbstractCli.java
@@ -18,12 +18,15 @@
  */
 package org.apache.iotdb.cli;
 
-import static org.apache.iotdb.cli.utils.IoTPrinter.*;
+import static org.apache.iotdb.cli.utils.IoTPrinter.computeHANCount;
+import static org.apache.iotdb.cli.utils.IoTPrinter.printBlockLine;
+import static org.apache.iotdb.cli.utils.IoTPrinter.printCount;
+import static org.apache.iotdb.cli.utils.IoTPrinter.printRow;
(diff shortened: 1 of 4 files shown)


cli/src/main/java/org/apache/iotdb/cli/AbstractCli.java
```diff
@@ -18,10 +18,12 @@
  */
 package org.apache.iotdb.cli;
 
+import static org.apache.iotdb.cli.utils.IoTPrinter.*;
+
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
-import java.io.PrintStream;
+import java.lang.Character.UnicodeScript;
```
HTHou, line 26: This import seems unnecessary. I think it should be remove.

HTHou (approved):
It looks great! 
Screen Shot 2021-01-06 at 10 08 18 AM

cli/src/main/java/org/apache/iotdb/cli/AbstractCli.java
weizihan0110: > This `import java.lang.Character.UnicodeScript` seems unnecessary. I think it should be removed.

Hi, already optimize imports

weizihan0110:
running Travis CI failed, could you run it again?

weizihan0110:
> It looks great!
> Screen Shot 2021-01-06 at 10 08 18 AM

running Travis CI failed, could you run it again?

HTHou:
> running Travis CI failed, could you run it again?

Sure.

PR #2425, opened 2021-01-05
[IOTDB-1103] [To rel/0.11] Fix frame size larger than max length error

samperson1997:
When querying large amount size of data, `Frame size larger than max length` may thrown in client.

PR #2428, opened 2021-01-05
[IOTDB-1104]put all query error caused by incorrect sql  are not Server internal error any more and not printted in Tests

jixuan1989:
## Description

1. Now the CI print too many meaningless log, see https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-1104
2. Some error are just caused by incorrect SQL, they should not be marked as "Server Internal Error".

3. Some RestartTests failed because the compaction module does not allow clear the MManamger before it is stopped.
see #2426  https://github.com/apache/iotdb/pull/2426/checks?check_run_id=1650191463 and 
#2425  https://github.com/apache/iotdb/pull/2425/checks?check_run_id=1649494523

## How to fix

1. I define a new Logger in TsserviceImple, called `DETAILED_FAILUER_QUERY_TRACE`, and use it to print the errors except for the NullException, Exception, and AuthException.

2. The logger is in the `ERROR` level in Tests and the `WARN` level in `src/assembly`
 
3. I replace many `System.out` and `printStackTrace` in Tests with `logger.debug`.

4. Move  `IoTDB.metaManager.clear();` out of `IoTDB.shutdown`

<hr>

This PR has:
- [x] been self-reviewed.
    - [x] concurrent read
    - [x] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


jixuan1989:
@SteveYurongSu will resubmit a better PR.

PR #2442, opened 2021-01-07
[IOTDB-1106] Fix the bug MTree mistakenly deletes parent node when executing "delete timeseries" statement

wshao08:
JIRA issue: https://issues.apache.org/jira/browse/IOTDB-1106


wshao08:
Duplicated PR with #2441

PR #2445, opened 2021-01-07
Remove versionInfo in Tsfile and get rid of versions in memtable

wshao08:
IoTDB uses an incremental field "version" to keep track of the updating order of multiple Memtables, and this version is also persisted into TsFiles.
This PR targets to remove the "version" field in Memtables and avoid serializing/deserializing it into TsFiles.

Design docs：
https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=173082421


server/src/main/java/org/apache/iotdb/db/utils/QueryUtils.java
```diff
@@ -48,7 +48,7 @@ public static void modifyChunkMetaData(List<ChunkMetadata> chunkMetaData,
     for (int metaIndex = 0; metaIndex < chunkMetaData.size(); metaIndex++) {
       ChunkMetadata metaData = chunkMetaData.get(metaIndex);
       for (Modification modification : modifications) {
-        if (modification.getVersionNum() > metaData.getVersion()) {
+        if (modification.getFileOffset() > metaData.getOffsetOfChunkHeader()) {
```
SilverNarcissus, line 51: Can modification.getFileOffset() == metaData.getOffsetOfChunkHeader() ?   What should we do in this condition

SilverNarcissus (changes requested):
Fantastic work!  I think we need some test for a special situation. There are tests for deletion in memtable and tsfile but no test for **deletion in a flushing memtable**.  Please add some tests for it~

server/src/main/java/org/apache/iotdb/db/utils/QueryUtils.java
wshao08: 1. metaData.getOffsetOfChunkHeader() marks the byte offset of this chunk header.
2. modification.getFileOffset() returns the offset of any chunk in the file after the chunk has been flushed. 
So I consider it could be guaranteed that if any chunk is successfully flushed, offset in (2) > (1).
Thats why I didn't add the equality condition.

SilverNarcissus: Reasonable! So just add some comment here and add test which I said above.

SilverNarcissus: You can also check sonar code smell. There is an unused parameter.

HTHou (commented):
I checked the code related to Flushing and TsFile module, just some small log issues. Please fix them.

JackieTien97 (approved):

server/src/main/java/org/apache/iotdb/db/utils/QueryUtils.java
wshao08: Thanks, I'll add some more test scenarios.

server/src/main/java/org/apache/iotdb/db/engine/flush/MemTableFlushTask.java
```diff
@@ -67,8 +67,7 @@ public MemTableFlushTask(IMemTable memTable, RestorableTsFileIOWriter writer, St
     this.storageGroup = storageGroup;
     this.encodingTaskFuture = subTaskPoolManager.submit(encodingTask);
     this.ioTaskFuture = subTaskPoolManager.submit(ioTask);
-    logger.debug("flush task of Storage group {} memtable {} is created ",
-        storageGroup, memTable.getVersion());
+    logger.debug("flush task of Storage group {} memtable is created ", storageGroup);
```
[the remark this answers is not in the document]
wshao08: Fixed

HTHou (approved):

PR #2449, opened 2021-01-08
[To rel/0.11] [IOTDB-1108] Add error log to print file name while error happened

JackieTien97:


HTHou (approved):

JackieTien97:
Add error log to print file name while error happened

PR #2457, opened 2021-01-08
Fix delete bug in delete range judge

yuqi1129:
## Description

This PR try to fix the bug in the method `canSkipDelete` about the skip logic.

# Detail
if the TsFileResource has start with T1 and T2, that's [T1, T2],  when we try to delete a range [T3, T4]
if T3 <= T2 **or** T4 >= T1, we can make sure the two range has intersection, and `canSkipDelete` should return false

The reason why this bug has not been triggered is that the parameter of [T3, T4] is [Long.Min, Long.Max], and the expression T3 <= T2 and T4 > T1 is always true. but if the input value changed, error may occur

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
@@ -1541,7 +1541,7 @@ public class StorageGroupProcessor {
       }
 
       if (tsFileResource.getDevices().contains(deviceId) &&
-          (deleteEnd >= tsFileResource.getStartTime(deviceId) && deleteStart <= endTime)) {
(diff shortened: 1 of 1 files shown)


wshao08:
Hi. To make sure the two range has intersection, only T3 <= T2 or T4 >= T1 is not enough. 
e.g. 
[T1,T2] : [100,200]
[T3,T4] : [50, 80]
T3 <= T2 but the two ranges have no intersection.

It is easy to know that if T2<T3 or T4<T1, the two ranges cannot have intersection. Then go with the reverse logic the result should be !(T2<T3 or T4<T1) => (T2 >= T3 && T4>=T1)
Please check.

yuqi1129:
> It is easy to know that if T2<T3 or T4<T1, the two ranges cannot have intersection. Then go with the reverse logic the result should be !(T2<T3 or T4<T1) => (T2 >= T3 && T4>=T1)

Sorry, you are right, i miss something...

PR #2460, opened 2021-01-09
fix show devices

yanhongwangg:
before，when i create a timeseries with three nodes an then execute show devices
image2021-1-9 17_24_25

after，same operation
image

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
@@ -1362,11 +1362,6 @@ public class MTree implements Serializable {
         for (int i = depth + 1; i < nodes.length; i++) {
           pathWithKnownSG.append(IoTDBConstant.PATH_SEPARATOR).append(nodes[i]);
         }
-        if (depth >= nodes.length - 1 && currNode.equals(PATH_WILDCARD)) {
-          // the we find the sg at the last node and the last node is a wildcard (find "root
-          // .group1", for "root.*"), also append the wildcard (to make "root.group1.*")
-          pathWithKnownSG.append(IoTDBConstant.PATH_SEPARATOR).append(PATH_WILDCARD);
-        }
         paths.put(sgName, pathWithKnownSG.toString());
       } else if (!child.getChildren().isEmpty()) {
         // push it back so we can traver its children later


HTHou:
Good job, Yanhong!
Consider add an UT or IT?

yanhongwangg:
> Good job, Yanhong!
> Consider add an UT or IT?

OK, I'll add it

jixuan1989:
ping

jt2594838:
How this may affect data queries need to be further investigated. For example, if this change results in `root.group1.*` being replaced by `root.group1` in a data query, the result will be very different.
I suggest a modification in `show devices` rather than here.

yanhongwangg:
> How this may affect data queries need to be further investigated. For example, if this change results in `root.group1.*` being replaced by `root.group1` in a data query, the result will be very different.
> I suggest a modification in `show devices` rather than here.

Thanks for your advice,but this bug occurs before getting devices according to the path, so there is no way to change it when getting devices，meanwhile, I looked at it and "*" is not used.

jt2594838:
I do not think you get my point. I was saying adding ".*" to the paths is not necessarily a bug, as it may be needed in data queries. You should either convince me by showing that this change does not affect related data queries or allow `show device` and other queries to ignore the last ".*".

jt2594838:
For example, assuming we have 3 SGs: root.sg1, root.sg2, and root.sg3.
And now we execute `SELECT * FROM root`, the removed code will convert this SQL into "SELECT root.sg1.*, root.sg2.*, root.sg3.*", while after the removal, the result may be "SELECT root.sg1, root.sg2, root.s3", which is apparently a different one.

yanhongwangg:
> I do not think you get my point. I was saying adding `.*` to the paths is not necessarily a bug, as it may be needed in data queries. You should either convince me by showing that this change does not affect related data queries or allow `show device` and other queries to ignore the last `.*`.

Thanks for your guidence, I got it, I'll make changes in show devices to ignore the last one*

yanhongwangg:
> ignore the last `.*`.

Maybe i can fix it in show devices by ignoring the last .*., but there is another question 
before ignoring the last .*., when i execute show devices root.a.b.* ,the result is  root.a.b.c,
after ignoring the last .*., execute the same sql ,it will match root.a.b and root.a.b.c,
 result is inconsistent.

yanhongwangg:
> I do not think you get my point. I was saying adding `.*` to the paths is not necessarily a bug, as it may be needed in data queries. You should either convince me by showing that this change does not affect related data queries or allow `show device` and other queries to ignore the last `.*`.

Maybe i can fix it in show devices by ignoring the last .*., but there is another question 
before ignoring the last .*., when i execute show devices root.a.b.* ,the result is  root.a.b.c,
after ignoring the last .*., execute the same sql ,it will match root.a.b and root.a.b.c,
 result is inconsistent.

jt2594838:
> > I do not think you get my point. I was saying adding `.*` to the paths is not necessarily a bug, as it may be needed in data queries. You should either convince me by showing that this change does not affect related data queries or allow `show device` and other queries to ignore the last `.*`.
> 
> Maybe i can fix it in show devices by ignoring the last ._., but there is another question before ignoring the last ._., when i execute show devices root.a.b.* ,the result is root.a.b.c,
> after ignoring the last .*., execute the same sql ,it will match root.a.b and root.a.b.c,
> result is inconsistent.

How is it possible? Only one of `root.a.b` and `root.a.b.c` can be a device if you fix it correctly.

PR #2461, opened 2021-01-10
set the log is applied when the exception occurred during applying

neuyilan:
if an exception occurred during applying, the log would not be applied forever, and the server would be blocked.
this pr is to try to fix the problem.


jt2594838:
`log.setApplied(true);` is always invoked after `log.setException(e);` except
image
maybe it is better to add `log.setApplied(true);` there instead of adding unnecessary `synchronized`.

neuyilan:
> `log.setApplied(true);` is always invoked after `log.setException(e);` except
> image
> maybe it is better to add `log.setApplied(true);` there instead of adding unnecessary `synchronized`.

ok, I will add log.setApplied(true) here

LebronAl (approved):
LGTM

jt2594838 (approved):

jt2594838:
There are failing tests, please take a look.

mychaow (approved):
LGTM

neuyilan:
> There are failing tests, please take a look.

have fixed

PR #2495, opened 2021-01-14
Replace TreeSet with TimeSelector (heap-based) for better raw query performance

SteveYurongSu:
This PR improves the performance of the raw query by optimizing the data structure. By replacing TreeSet with TimeSelector, this PR reduces the Java boxing and unboxing operations and also avoids the rotation and colouring steps of the red-black tree.

**Performance report:**

Case | Reduced time
-- | --
1 | 17.3%
2 | 17.6%
3 | 4.7%

**Case1: single column query**
select s1 from root.sg1.d1;

**Case 2: multiple columns query (50 columns, 10 000 000 rows)**
select * from root.sg.d1;

Time | s1 | s2 | s3 | s4 | ... | s50
-- | -- | -- | -- | -- | -- | --
1 | 1 | 1 | 1 | 1 | ... | 1
2 | 2 | 2 | 2 | 2 | ... | 2
3 | 3 | 3 | 3 | 3 | ... | 3
... | ... | ... | ... | ... | ... | ...

**Case 3: multiple columns query (50 columns, 10 000 000 rows)**
select * from root.sg.d1;
If the original data has only 4 columns:

Time | s1 | s2 | s3 | s4
-- | -- | -- | -- | --
1 | 1 | null | null | null
2 | null | 2 | null | null
3 | null | null | 3 | null
4 | null | null | null | 4
5 | 5 | null | null | null
6 | null | 6 | null | null
7 | null | null | 7 | null
... | ... | ... | ... | ...


server/src/main/java/org/apache/iotdb/db/utils/datastructure/TimeSelector.java
```diff
@@ -0,0 +1,155 @@
...
+    timeHeap = new long[Math.max(defaultCapacity, MIN_DEFAULT_CAPACITY)];
+    heapSize = 0;
+    lastTime = Long.MIN_VALUE;
+  }
+
+  public boolean isEmpty() {
+    while (heapSize != 0 && timeHeap[0] == lastTime) {
+      timeHeap[0] = timeHeap[heapSize - 1];
+      percolateDown(0, timeHeap[0]);
+      --heapSize;
```
JackieTien97, line 45: The `--heapSize` statement should be moved before `percolateDown`.

JackieTien97 (approved):

PR #2400, opened 2020-12-31
[IOTDB-519] Apache IoTDB integration with Prometheus

yanhongwangg:
support prometheus write data to iotdb
https://cwiki.apache.org/confluence/display/IOTDB/Iotdb+integrated+Prometheus

PR #2497, opened 2021-01-14
Using openapi to generate rest framework (go and java) codes

jixuan1989:
## Description

OpenAPI is a uniform DSL for describing the restful interface. 
And, when developing a restful framework, we have to write many trivial codes (that will be also written in almost all restful services). So, generating such of codes is a good solution as we just need to develop the real logic and do not need to care to maintain the generated codes.

This PR defines several restful APIs for serving Grafana and Prometheus, and uses openapi-generator to generate the source codes of the restful framework in Go and Java. (The generator can also generate client-sdk, but I have not time to try it and we have no requirements).

So, just run `mvn generate-sources` in `openapi' module, and then develop your logic in 
`openapi/java-rest/src/main/java/org/apache/iotdb/openapi/gen/handler/impl/V1ApiServiceImpl.java`  (for java) and `openapi/go-rest/src/iotdbrestimpl/api_service_impl.go`(for Go).

The only 3 things should to know are, 

1. current openapi-generator has some bug for go-server codes, see: https://github.com/OpenAPITools/openapi-generator/issues/8448 and I have write how to fix it manually in go-rest/README.md

2. current openapi-generator can not generate java codes using native Netty. But it seems developing such a generator is not easy for openapi-generator.

3. I copy some generated codes to src folder.

The author of PR https://github.com/apache/iotdb/pull/2400 can consider to using this generated Go-server codes. (or, do some experiments to check whether the generated codes have an acceptable performance).


icloudsong:
Does it support REST API for https?

jixuan1989:
has been merged into https://github.com/apache/iotdb/pull/3299

PR #2500, opened 2021-01-15
[IOTDB-1091] add linear fill test case for sdt encoding

haimeiguo:
SDT algorithm uses a  parallelogram to filter raw data and its height is 2 * compression deviation. After SDT compression, if user uses linear fill, the resulting values will fall inside the parallelogram of two stored points. The absolute difference between filled value and original value is <= 2 * compression deviation.

### DIFF
--- a/server/src/test/java/org/apache/iotdb/db/integration/IoTDBSimpleQueryIT.java
+++ b/server/src/test/java/org/apache/iotdb/db/integration/IoTDBSimpleQueryIT.java
@@ -260,7 +260,7 @@ public class IoTDBSimpleQueryIT {
         }
       }
     } catch (SQLException e) {
-      e.printStackTrace();
+      fail();
     }
   }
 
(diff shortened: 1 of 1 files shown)


server/src/test/java/org/apache/iotdb/db/integration/IoTDBSimpleQueryIT.java
```diff
@@ -210,6 +213,57 @@ public void testSDTEncodingCompDev() throws ClassNotFoundException {
...
+          if (value == null) {
+            continue;
+          }
+          // sdt parallelogram's height is 2 * compDev, so after linear fill, the values will fall inside
+          // the parallelogram of two stored points
+          assertTrue(Math.abs(Integer.parseInt(value) - map.get(time)) <= 2 * compDev);
+        }
+      }
+    } catch (SQLException e) {
+      e.printStackTrace();
```
SteveYurongSu, line 263:
```suggestion
      fail();
```

SteveYurongSu (approved):
LGTM.

mychaow (approved):
LGTM

PR #2507, opened 2021-01-16
IOTDB-1112 some error occurs occasionally in the test case

wangchao316:
## Description

If test cases are run in parallel, a test case class uses the tool class TsFileGeneratorForTest.java to generate a data file. After the execution is complete, the file is deleted. However, another test class is running and the tool class TsFileGeneratorForTest is used to generate a the same name file. As a result, the file cannot be found.

This pr is used to generate different data files for different test case classes and allow them to be executed concurrently in parallel. two test case class be no influence between each other.
This is one of the features of the test case, that is independence.
This problem does not occur in Github because the current Github pr test uses a single thread to execute test cases one by one.

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


jixuan1989:
Could you describe the intention of the PR first?

wangchao316:
> Could you describe the intention of the PR first?

If test cases are run in parallel, a test case class uses the tool class TsFileGeneratorForTest.java to generate a data file. After the execution is complete, the file is deleted. However, another test class is running and the tool class TsFileGeneratorForTest is used to generate a the same name file. As a result, the file cannot be found.
1. This pr is used to generate different data files for different test case classes and allow them to be executed concurrently in parallel.
2. This is one of the features of the test case, that is independence.
3. This problem does not occur in Github because the current Github pr test uses a single thread to execute test cases one by one.

wangchao316:
@jixuan1989 @qiaojialin  hello , could you please review again? I add explanation

jixuan1989:
I am very interested in how to run tests in parallel.

wangchao316:
> I am very interested in how to run tests in parallel.

we could use TestNG in OpenSource softeware. 

I use my company private parallel-plug-in for iotdb in my local environment now.  

I find that it takes 20 minutes to build Maven in github, but 10 minutes to build and run tests in parallel in my local.
As the number of test cases increases, the execution time increases.
I recommend ：
1.modifying the test cases to be executed in parallel first.
2. and then switch to the parallel framework.

At the same time， if a single thread to execute test cases one by one,  failed to delete data file after a test finished, in this case, an error is reported when following test cases create the same name file .

wangchao316:
@HTHou @jt2594838 @neuyilan  hello , could you review this?
Thanks.

jixuan1989:
> we could use TestNG in OpenSource softeware.

TestNG is more powerful. 
Do you have plan to change current JUnit to TestNG? If not, I wonder the benefit that modifying current test codes.

> I use my company private parallel-plug-in for iotdb in my local environment now.

I am quite interested in parallel test. I know Maven 3 support parallel compiler by using `-T <number>`, but I do not know whether it runs all tests in parallel. 

In conclusion, the point is, what benefit the public repo will get from this PR? (No benefit does not means the pr can not be merged, as we welcome PRs that will making the project easier to be used in users' environment.)

wangchao316:
> get

yes, I plan to change TestNG..

PR #2509, opened 2021-01-18
[C++ Client] remove compiling boost for compling client-cpp module

wshao08:
we do not download and compile boost anymore. 
Instead, we require users to install the boost themselves before compiling this module.


pom.xml
```diff
@@ -119,7 +119,7 @@
         <common.io.version>2.5</common.io.version>
         <commons.collections4>4.4</commons.collections4>
         <!-- keep consistent with client-cpp/tools/thrift/pom.xml-->
-        <thrift.version>0.13.0</thrift.version>
+        <thrift.version>0.12.0</thrift.version>
```
jixuan1989, line 122: revert back to 0.13

PR #2511, opened 2021-01-18
[IOTDB-1113] Optimize the execution efficiency of UDF

SteveYurongSu:
There are several points that can be optimized:

1. Avoid Java boxing / unboxing in LRUCache and SafetyLine

2. Avoid unnecessary memory copy in SerializableRowReordList

3. Memory control:
  3.1 Disable memory control when there are no text fields
  3.2 Stricter eviction upper bound

4. Use lighter data structures: RowRecord -> Object[] 

JIRA: https://issues.apache.org/jira/browse/IOTDB-1113


SteveYurongSu:
Performance report:

Data: root.sg.d.s1, INT32, 100 000 000 points

SQL | Execution Time | Ratio
-- | -- | --
select s1 from root.sg.d; | 19636 ms | 100%
BEFORE OPTIMIZATION - select do_nothing(s1) from root.sg.d; | 76446 ms | 389.3%
AFTER OPTIMIZATION - select do_nothing(s1) from root.sg.d; | 35547 ms | 180.0%

After optimization, the execution time of the do_nothing function is reduced by (76446-35547) / 76446 = 53.5%.

PR #2513, opened 2021-01-18
[IOTDB-1077] [C++ Client] Add insertRecordsOfOneDevice() interface for C++ client

wshao08:
Add insertRecordsOfOneDevice() in C++ client to comply with Java Session interface


client-cpp/src/main/Session.cpp
```diff
@@ -408,12 +417,52 @@ void Session::sortIndexByTimestamp(int* index, int64_t* timestamps, int length)
     }
 }
 
-vector<string> Session::sortList(vector<string>& valueList, int* index, int indexLength) {
-    vector<string> sortedValues(valueList.size());
-    for (int i = 0; i < indexLength; i++) {
-        sortedValues[i] = valueList[index[i]];
+void Session::putValuesIntoBuffer(vector<TSDataType::TSDataType>& types, vector<char*>& values, string buf) {
```
Alima777, line 420:
```suggestion
void Session::putValuesIntoBuffer(vector<TSDataType::TSDataType>& types, vector<char*>& values, string& buf) {

```
Is reference here?

Alima777 (commented):
Are there any tests about this??

client-cpp/src/main/Session.cpp
wshao08: Fixed

Alima777 (approved):
LGTM

PR #2520, opened 2021-01-19
Add workload manager

THUMarkLau:
A new WorkloadManager is added to collect the workload of GroupBy queries and Aggregation queries, and then store them in the /data/workload folder **in the form of SQL statements with timestamps**. WorkloadManager has two adjustable parameters: one is RECORD_SIZE, which is the maximum number of query records stored in the memory. When this number is exceeded, the data in the memory will be written to the disk, and then the data in the memory will be cleared. The default value is 100. The other is RECORD_FILE_MAX_SIZE, which is the file used to store query records. If the size of the currently written file exceeds RECORD_FILE_MAX_SIZE, a new file will be opened for writing. The default value is 1GB.


server/src/main/java/org/apache/iotdb/db/query/workloadmanager/WorkloadManager.java
```diff
@@ -0,0 +1,155 @@
...
+        break;
+      }
+    }
+    // logger.info("Record: " + record.getSqlWithTimestamp());
+    logger.info("Adding record type of " + record.getRecordType());
+    logger.info("Aggregation count: " + aggregationCount + ", group by count: " + groupByCount);
+  }
+
+  private void flushRecords() {
+    // TODO: 转换成异步的
```
JackieTien97, line 102: It's better to avoid Chinese comments in the code.

server/src/main/java/org/apache/iotdb/db/query/workloadmanager/WorkloadManager.java
```diff
@@ -0,0 +1,155 @@
...
+        groupByCount++;
+        break;
+      }
+    }
+    // logger.info("Record: " + record.getSqlWithTimestamp());
+    logger.info("Adding record type of " + record.getRecordType());
+    logger.info("Aggregation count: " + aggregationCount + ", group by count: " + groupByCount);
+  }
+
+  private void flushRecords() {
```
JackieTien97, line 101: There is no need to choose what file you write into. You can define your own logger, and the LOGGER will do everything for you, like automatically switch to a new log file while the current file reach the thredshold. You can refer to the `SLOW_SQL_LOGGER` field in the `TSServiceImpl` class and its logger action is defined in `server/src/assembly/resources/conf/logback.xml`.

JackieTien97 (changes requested):

THUMarkLau:
Changed the way of file writing to use LOGGER.
Add "synchronized" to all the method in Workload Manager.
Changed the way of collecting query info.

PR #2533, opened 2021-01-21
fix the bug that when introduce the virtual sg, the path not consist with the tsfile's in snapshot

neuyilan:
<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

## Key changed/added classes (or packages if there are too many classes) in this PR

Cluster send snapshot failed, the error is as follows
image

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/log/snapshot/FileSnapshot.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/log/snapshot/FileSnapshot.java
@@ -379,7 +379,7 @@ public class FileSnapshot extends Snapshot implements TimeseriesSchemaSnapshot {
      * @param resource
      */
     private void loadRemoteResource(RemoteTsFileResource resource) throws IllegalPathException {
-      // the new file is stored at: remote/<nodeIdentifier>/${resource.getAbsolutePath()}
+      // the new file is stored at: remote/<nodeIdentifier>/resource.getTsFile().getAbsolutePath()
       PartialPath storageGroupName = new PartialPath(
           FilePathUtils.getLogicalStorageGroupName(resource));
       File remoteModFile =
--- a/cluster/src/main/java/org/apache/iotdb/cluster/log/snapshot/FileSnapshot.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/log/snapshot/FileSnapshot.java
@@ -435,11 +435,7 @@ public class FileSnapshot extends Snapshot implements TimeseriesSchemaSnapshot {
           node.getNodeIdentifier() + File.separator + FilePathUtils.getTsFilePrefixPath(resource)
               + File.separator + tempFileName;
       File tempFile = new File(REMOTE_FILE_TEMP_DIR, tempFilePath);
-      boolean success = tempFile.getParentFile().mkdirs();
-      if (!success) {
-        logger.error("mkdir failed when pull remote file, {}", tempFile);
-        return null;
-      }
+      tempFile.getParentFile().mkdirs();
       if (pullRemoteFile(resource.getTsFile().getAbsolutePath(), node, tempFile)) {
         // TODO-Cluster#353: implement file examination, may be replaced with other algorithm
         if (resource.isWithModification()) {
--- a/cluster/src/main/java/org/apache/iotdb/cluster/log/snapshot/FileSnapshot.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/log/snapshot/FileSnapshot.java
@@ -60,6 +60,7 @@ import org.apache.iotdb.db.exception.metadata.IllegalPathException;
 import org.apache.iotdb.db.metadata.PartialPath;
 import org.apache.iotdb.db.utils.FilePathUtils;
 import org.apache.iotdb.db.utils.SchemaUtils;
+import org.apache.iotdb.tsfile.utils.Pair;
 import org.apache.iotdb.tsfile.write.schema.TimeseriesSchema;
 import org.apache.thrift.TException;
 import org.slf4j.Logger;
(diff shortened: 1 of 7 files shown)


cluster/src/main/java/org/apache/iotdb/cluster/log/snapshot/FileSnapshot.java
```diff
@@ -430,24 +427,20 @@ private void loadRemoteResource(RemoteTsFileResource resource) throws IllegalPat
     private File pullRemoteFile(RemoteTsFileResource resource, Node node) throws IOException {
       logger.info("{}: pulling remote file {} from {}, plan index [{}, {}]", name, resource, node
           , resource.getMinPlanIndex(), resource.getMaxPlanIndex());
-
-      String[] pathSegments = FilePathUtils.splitTsFilePath(resource);
-      int segSize = pathSegments.length;
       // the new file is stored at:
-      // remote/<nodeIdentifier>/<storageGroupName>/<partitionNum>/<fileName>
+      // remote/<nodeIdentifier>/resource.getTsFile().getAbsolutePath()
```
mychaow, line 431: use logicalStorageGroupName/virtualStorageGroupName/timePartitionId/fileName is more better

mychaow (approved):
LGTM

cluster/src/main/java/org/apache/iotdb/cluster/log/snapshot/FileSnapshot.java
neuyilan: The path may change, so I put all the annotation in the` splitTsFilePath()` method in FilePathUtils to reduce dispersion so that such bugs occur less often

server/src/main/java/org/apache/iotdb/db/utils/FilePathUtils.java
```diff
@@ -50,36 +51,84 @@ private FilePathUtils() {
...
   }
 
+  /**
+   * IMPORTANT, when the path of TsFile changes, the following methods should be changed
+   * accordingly. The sequence TsFile is located at ${IOTDB_DATA_DIR}/data/sequence/. The unsequence
+   * TsFile is located at ${IOTDB_DATA_DIR}/data/unsequence/. Where different storage group's TsFile
+   * is located at <logicalStorageGroupName>/<virtualStorageGroupName>/<timePartitionId>/<fileName>.
+   * For example, one sequence TsFile may locate at /data/data/sequence/root.group_9/0/0/1611199237113-4-0.tsfile
+   *
+   * @param resource the tsFileResource
```
jixuan1989, line 73: UT is important for this class, as it is very important to guarantee it works on all platforms.

In some degree, IT is also useful. If we have IT, then we can guarantee other committers have to modify this function if they modified the position of the tsfiles.
E.g., start IoTDB -> write some data -> flush -> split the real data file path.

jixuan1989: By the way, is it necessary to get `getAbsolutePath`? (Seems you only use the last 3 level of folder names) (I am not sure)

neuyilan: Ok, I'll and some UT and IT.

> By the way, is it necessary to get getAbsolutePath? (Seems you only use the last 3 level of folder names) (I am not sure)
Yes, we need to get the last 4 level of the folder name, so `getAbsolutePath` maybe is the most appropriate method

server/src/main/java/org/apache/iotdb/db/utils/FilePathUtils.java
```diff
@@ -50,36 +51,84 @@ private FilePathUtils() {
...
   public static String[] splitTsFilePath(TsFileResource resource) {
     return resource.getTsFile().getAbsolutePath().split(PATH_SPLIT_STRING);
   }
 
+  public static String getLogicalStorageGroupName(TsFileResource resource) {
+    String[] pathSegments = splitTsFilePath(resource);
+    return pathSegments[pathSegments.length - 4];
+  }
+
+  public static String getVirtualStorageGroupName(TsFileResource resource) {
```
jt2594838, line 84: Better use `getVirtualStorageGroupId`.

server/src/test/java/org/apache/iotdb/db/integration/IoTDBFilePathUtilsIT.java
```diff
@@ -0,0 +1,101 @@
...
+      Assert.fail(e.getMessage());
+    }
+    Assert.assertNotNull(sgPath);
+    List<TsFileResource> tsFileResources = StorageEngine.getInstance().getProcessor(sgPath)
+        .getSequenceFileTreeSet();
+    Assert.assertNotNull(tsFileResources);
+
+    for (TsFileResource tsFileResource : tsFileResources) {
+      String sgName = FilePathUtils.getLogicalStorageGroupName(tsFileResource);
+      Assert.assertEquals(storageGroupName, sgName);
```
jt2594838, line 82: It would be better to also check other methods in `FilePathUtils`.

server/src/test/java/org/apache/iotdb/db/integration/IoTDBFilePathUtilsIT.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+  }
+
+  private static void close() {
+    if (Objects.nonNull(connection)) {
+      try {
+        connection.close();
+      } catch (Exception e) {
+        e.printStackTrace();
```
jt2594838, line 91: Use logger or throw the exception.

jt2594838 (approved):

server/src/main/java/org/apache/iotdb/db/utils/FilePathUtils.java
neuyilan: sure, I'll change the method name

server/src/test/java/org/apache/iotdb/db/integration/IoTDBFilePathUtilsIT.java
neuyilan: ok, I'll check other methods, but one difficulty is that we may only know the storage group name, and do not sure about the other values in the path.

PR #2534, opened 2021-01-21
Optimize VirtualStorageGroupManage class to make it more compact

rhmangmang:
Use storageGroupMNode as the param of VirtualStorageGroupManag's constructor. Because they have a corresponding relationship.


SilverNarcissus (changes requested):
Thanks for your contribution. However this change will add one reference on storage group mnode. In the future, we will persist mnode on disk, this may have an impact on evict mnode out of memory. As far as I'm concern, we should better keep the code as origin.

PR #2539, opened 2021-01-21
[IOTDB-1077] Add python test and insert one device interface

SilverNarcissus:
Add python test and insert one device interface

<hr>

This PR has:
- [x] been self-reviewed.
    - [x] concurrent read
    - [x] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


client-py/readme.md
```diff
@@ -72,3 +72,10 @@ session.open(False)
 zone = session.get_time_zone()
 session.close()
 ```
+
+## test file
+
+You can use `client-py/src/SessionTest.py` to test python session, if all test has been passed, it will return 0
+Otherwise it will return 1. You can use print message to locate failed operations and the reason of them
+
+Notice: you should start IoTDB server firstly and then run the test
```
HTHou, line 81:
```suggestion
You can use `client-py/src/SessionTest.py` to test python session, if the test has been passed, it will return 0. Otherwise it will return 1. You can use the printed message to locate failed operations and the reason of them.

Notice: you should start IoTDB server firstly and then run the test.
```

HTHou (approved):

PR #2543, opened 2021-01-21
[IOTDB-1091] fail to create timeseries when set invalid sdt parameters

haimeiguo:
Currently we check SDT parameters when flushing,  if the parameters are not valid we will only disable SDT compression. 

This pr moves SDT validation process to when we create timeseries. If SDT parameters are invalid, user will get an error and timeseries are not created. Also log info will record if user did not set compMinTime or compMaxTime.

```
IoTDB> create timeseries root.sg4.d1.s3 with datatype=double,loss=sdt,compdev=-2
Msg: 303: SDT compression deviation cannot be negative. Failed to create timeseries.

IoTDB> create timeseries root.sg4.d1.s4 with datatype=double,loss=sdt,compdev=2,compmintime=4,compmaxtime=1
Msg: 303: Error occurred SDT compression maximum needs to be greater than compression minimum. Failed to create timeseries.
```


server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
```diff
@@ -237,6 +242,58 @@ MeasurementMNode createTimeseries(PartialPath path, TSDataType dataType, TSEncod
     }
   }
 
+  //check if sdt parameters are valid
+  private void checkSDTFormat(PartialPath path, Map<String, String> props) throws IOException {
```
jixuan1989, line 246: Why IOexception...

jixuan1989 (changes requested):
create `IllegalParameterOfPathException` that `extends  MetadataException`

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
```diff
@@ -237,6 +242,58 @@ MeasurementMNode createTimeseries(PartialPath path, TSDataType dataType, TSEncod
...
+    String s = "Failed to create timeseries.";
+
+    if (!props.containsKey("compdev")) {
```
weizihan0110, line 249: Hi, something like "compdev" should not be hard-code. It's better to extract them to be a variable.

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
```diff
@@ -212,6 +212,11 @@ MeasurementMNode createTimeseries(PartialPath path, TSDataType dataType, TSEncod
       }
       cur = cur.getChild(nodeName);
     }
+
+    if (props != null && props.containsKey("loss") && props.get("loss").equals("sdt")) {
```
weizihan0110, line 216: Hi, something like "sdt" or "loss" should not be hard-code. It's better to extract them to be a variable.

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
```diff
@@ -237,6 +242,58 @@ MeasurementMNode createTimeseries(PartialPath path, TSDataType dataType, TSEncod
...
+      throw new IOException("Error occurred when formatting SDT compression deviation. " + s);
+    }
+
+    long compMaxTime = Long.MAX_VALUE;
+    long compMinTime = 0;
+    if (props.containsKey("compmintime")) {
+      try {
+        compMinTime = Long.parseLong(props.get("compmintime"));
+        if (compMinTime < 0) {
+          throw new NumberFormatException();
```
weizihan0110, line 268: should print something like "compmintime cannot be negative"

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
```diff
@@ -237,6 +242,58 @@ MeasurementMNode createTimeseries(PartialPath path, TSDataType dataType, TSEncod
...
+      }
+    } else {
+      logger.info(path + " enabled SDT but did not set compression minimum time.");
+    }
+
+    if (props.containsKey("compmaxtime")) {
+      try {
+        compMaxTime = Long.parseLong(props.get("compmaxtime"));
+        if (compMaxTime < 0) {
+          throw new NumberFormatException();
```
weizihan0110, line 281: Same as above

wangchao316 (changes requested):

tsfile/src/main/java/org/apache/iotdb/tsfile/write/chunk/ChunkWriterImpl.java
```diff
@@ -133,40 +133,15 @@ private void checkSdtEncoding() {
       }
 
       if (isSdtEncoding && measurementSchema.getProps().containsKey("compdev")) {
-        try {
-          sdtEncoder
-              .setCompDeviation(Double.parseDouble(measurementSchema.getProps().get("compdev")));
-        } catch (NumberFormatException e) {
-          logger.error("meet error when formatting SDT compression deviation");
-        }
-        if (sdtEncoder.getCompDeviation() < 0) {
```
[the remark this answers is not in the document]
jixuan1989: in general, we need to obey the rule you mentioned.
The exception here is SDT will be introduced from 0.12, that is, this feature is not released up to now.

server/src/main/java/org/apache/iotdb/db/conf/IoTDBConstant.java
```diff
@@ -102,6 +102,13 @@ private IoTDBConstant() {
   public static final String PATH_WILDCARD = "*";
   public static final String TIME = "time";
 
+  //sdt parameters
+  public static final String LOSS = "loss";
```
jixuan1989, line 106: The problem is... with Antlr, all these keywords will be converted to lowercase, but if users call thrift API, createTimeseries(), how they know they must write these keywords as lowercase.
(The example in your SQL grammar doc is uppercase)

jixuan1989: and make sure it works on the cluster mode.

haimeiguo: Hi, in CreateTimeSeriesPlan, I converted props to case insensitive. 
this.props = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);

jixuan1989: OK..

PR #2550, opened 2021-01-21
Optimize storage group recovery

rhmangmang:
It's better not to do recovery when Storage Group Directory doesn't exist or creating it failed


server/src/main/java/org/apache/iotdb/db/engine/StorageEngine.java
```diff
@@ -417,21 +417,13 @@ private StorageGroupProcessor getStorageGroupProcessorByPath(PartialPath deviceP
...
       // if finish recover
-      if (isAllSgReady.get()) {
-        waitAllSgReady(devicePath);
-        synchronized (storageGroupMNode) {
-          virtualStorageGroupManager = processorMap.get(storageGroupMNode.getPartialPath());
-          if (virtualStorageGroupManager == null) {
-            virtualStorageGroupManager = new VirtualStorageGroupManager();
-            processorMap.put(storageGroupMNode.getPartialPath(), virtualStorageGroupManager);
-          }
+      waitAllSgReady(devicePath);
```
jixuan1989, line 420: if sg is not ready, tell the user that message ASAP is a good design... 
In general, blocking user's call is not good

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/virtualSg/VirtualStorageGroupManager.java
```diff
@@ -159,12 +159,14 @@ public void run() {
...
-      recoverThread.start();
     }
 
+    Thread thread;
     for (int i = 0; i < partitioner.getPartitionCount(); i++) {
       try {
-        threadList.get(i).join();
+        thread = threadList.get(i);
+        thread.start();
+        thread.join();
```
jixuan1989, line 169: why change this from async to sync mode?

jixuan1989 (changes requested):
1. I think it is ok to reject recover the system if the SG folder creation failed, but if we can create it successfully, why you think we should not recover the system?

2. If `shouldDoRecovery=false`, I think exception should be thrown.

PR #2567, opened 2021-01-23
bug fix - skip illegal path when delete multi timeseries 

haimeiguo:
Currently when using session.deleteTimeseries(List<String> paths), if some of them are illegal path or do not exsit, it will throw exception and stop deleting timeseries in the list.

ie 
delete paths = "root.sg1.d1.s1", "", "root.sg1.d1.s3"
"root.sg1.d1.s1" will be deleted, when read "", it will throw exception,  "root.sg1.d1.s3" is ignored and not deleted. 

This pr catches the illegal path exception and continues deleting qualified timeseries. It will throw exception after iterating all timeseries. 
so in the above example,  "root.sg1.d1.s1", "root.sg1.d1.s3" will be deleted, and tell the user "" is illegal path.


server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
```diff
@@ -1258,37 +1256,20 @@ private boolean createMultiTimeSeries(CreateMultiTimeSeriesPlan multiPlan)
...
-        if (failedTimeseries != null) {
-          failedNames.add(failedTimeseries);
-        }
-      }
-      if (!nonExistPaths.isEmpty()) {
-        throw new PathNotExistException(nonExistPaths);
-      }
-      if (!failedNames.isEmpty()) {
-        throw new DeleteFailedException(String.join(",", failedNames));
+        IoTDB.metaManager.deleteTimeseries(path);
```
asdf2014, line 1263:
```suggestion
        IoTDB.metaManager.deleteTimeseries(path);
        deleteTimeSeriesPlan.getResults().put(i, RpcUtils.getStatus(TSStatusCode.SUCCESS_STATUS));
```

asdf2014: It would be better to record the status when the deletion is successful.

haimeiguo: added. thank you

cluster/src/main/java/org/apache/iotdb/cluster/log/applier/BaseApplier.java
```diff
@@ -73,6 +74,9 @@ void applyPhysicalPlan(PhysicalPlan plan, DataGroupMember dataGroupMember)
       } catch (QueryProcessException e) {
         if (e.getCause() instanceof StorageGroupNotSetException) {
           executeAfterSync(plan);
+        } else if (e instanceof BatchProcessException) {
+          String message = "Exception occurred while processing non-query. ";
+          logger.warn(message, e);
```
neuyilan, line 79:
```suggestion
          logger.warn("Exception occurred while processing non-query. ", e);
```

neuyilan (changes requested):
It's better and one IT or UT to cover the bug tests.

neuyilan (approved):

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
```diff
@@ -1258,37 +1256,21 @@ private boolean createMultiTimeSeries(CreateMultiTimeSeriesPlan multiPlan)
...
-  }
-
-  private String deleteTimeSeries(PartialPath path, List<String> nonExistPaths)
-      throws MetadataException {
-    String failedTimeseries = null;
-    try {
-      failedTimeseries = IoTDB.metaManager.deleteTimeseries(path);
-    } catch (PathNotExistException e) {
-      nonExistPaths.add(path.getFullPath());
+    if (!deleteTimeSeriesPlan.getResults().isEmpty()) {
```
jixuan1989, line 1270: seems this is always true..

jixuan1989 (changes requested):
btw, does this need to be cherry-picked to 0.11?

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
```diff
@@ -1261,46 +1259,26 @@ private boolean createMultiTimeSeries(CreateMultiTimeSeriesPlan multiPlan)
         multiPlan.getResults().put(i, RpcUtils.getStatus(e.getErrorCode(), e.getMessage()));
       }
     }
-    if (!multiPlan.getResults().isEmpty()) {
```
neuyilan, line 1264: Why throw BatchProcessException always? 
if  all CreateTimeSeriesPlans execute success,  we should return true.
besides, the createMultiTimeSeries's bug have been fix in https://github.com/apache/iotdb/pull/2698

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
```diff
@@ -1261,46 +1259,26 @@ private boolean createMultiTimeSeries(CreateMultiTimeSeriesPlan multiPlan)
...
-
-  private String deleteTimeSeries(PartialPath path, List<String> nonExistPaths)
-      throws MetadataException {
-    String failedTimeseries = null;
-    try {
-      failedTimeseries = IoTDB.metaManager.deleteTimeseries(path);
-    } catch (PathNotExistException e) {
-      nonExistPaths.add(path.getFullPath());
     }
-    return failedTimeseries;
```
neuyilan, line 1303: The same as above.
if no failure found, we should not throw the exception

neuyilan (changes requested):

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
```diff
@@ -1261,46 +1259,26 @@ private boolean createMultiTimeSeries(CreateMultiTimeSeriesPlan multiPlan)
         multiPlan.getResults().put(i, RpcUtils.getStatus(e.getErrorCode(), e.getMessage()));
       }
     }
-    if (!multiPlan.getResults().isEmpty()) {
-      throw new BatchProcessException(multiPlan.getResults().values().toArray(new TSStatus[0]));
-    }
-    return true;
+    throw new BatchProcessException(multiPlan.getResults().values().toArray(new TSStatus[0]));
   }
```
neuyilan, line 1263: Why throw BatchProcessException always?
if all CreateTimeSeriesPlans execute success, we should return true.
besides, the createMultiTimeSeries's bug have been fix in #2698

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
```diff
@@ -1261,46 +1259,26 @@ private boolean createMultiTimeSeries(CreateMultiTimeSeriesPlan multiPlan)
...
-    String failedTimeseries = null;
-    try {
-      failedTimeseries = IoTDB.metaManager.deleteTimeseries(path);
-    } catch (PathNotExistException e) {
-      nonExistPaths.add(path.getFullPath());
     }
-    return failedTimeseries;
+    throw new BatchProcessException(
+        deleteTimeSeriesPlan.getResults().values().toArray(new TSStatus[0]));
   }
```
neuyilan, line 1282: The same as above.
if no failure found, we should not throw the exception

neuyilan (changes requested):

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
```diff
@@ -1261,46 +1259,26 @@ private boolean createMultiTimeSeries(CreateMultiTimeSeriesPlan multiPlan)
...
-      }
-      if (!nonExistPaths.isEmpty()) {
-        throw new PathNotExistException(nonExistPaths);
+        IoTDB.metaManager.deleteTimeseries(path);
+        deleteTimeSeriesPlan.getResults().put(i, RpcUtils.getStatus(TSStatusCode.SUCCESS_STATUS));
+      } catch (StorageEngineException | MetadataException e) {
+        deleteTimeSeriesPlan
+            .getResults()
+            .put(i, RpcUtils.getStatus(e.getErrorCode(), e.getMessage()));
       }
```
neuyilan, line 1278: it seems we missed the failedTimeseries return by the `IoTDB.metaManager.deleteTimeseries` method,  the `deleteTimeseries` method may return some delete failed timeseries while not throw an exception, in this way, we may lose some error message.

I suggest adding those delete failed timeserise to the deleteTimeSeriesPlan.getResults().

neuyilan (changes requested):

HTHou (approved):

neuyilan (approved):
LGTM

PR #2571, opened 2021-01-25
bug fix - fetch size calculation error for group by month

haimeiguo:
currently when use group by month and starttime is in February, there is a calculation error. 

```
IoTDB> select count(*) from root.sg1 group by ([2020-02-13, 2020-03-13), 1mo)
+----+---------------------+---------------------+---------------------+
|Time|count(root.sg1.d1.s4)|count(root.sg1.d1.s5)|count(root.sg1.d1.s2)|
+----+---------------------+---------------------+---------------------+
+----+---------------------+---------------------+---------------------+

IoTDB> select count(*) from root.sg1 group by ([2020-04-01, 2020-05-01), 1mo)
+-----------------------------+---------------------+---------------------+---------------------+
|                         Time|count(root.sg1.d1.s4)|count(root.sg1.d1.s5)|count(root.sg1.d1.s2)|
+-----------------------------+---------------------+---------------------+---------------------+
|2020-04-01T00:00:00.000+08:00|                    0|                    0|                    0|
+-----------------------------+---------------------+---------------------+---------------------+
Total line number = 1
It costs 0.003s
```

but it should be
```
IoTDB> select count(*) from root.sg1 group by ([2020-02-01, 2020-03-01), 1mo)
+-----------------------------+---------------------+---------------------+---------------------+
|                         Time|count(root.sg1.d1.s4)|count(root.sg1.d1.s5)|count(root.sg1.d1.s2)|
+-----------------------------+---------------------+---------------------+---------------------+
|2020-02-01T00:00:00.000+08:00|                    0|                    0|                    0|
+-----------------------------+---------------------+---------------------+---------------------+
```

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
+++ b/server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
@@ -24,6 +24,7 @@ import java.sql.SQLException;
 import java.time.ZoneId;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Calendar;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedHashSet;
@@ -173,6 +174,7 @@ public class TSServiceImpl implements TSIService.Iface, ServerContext {
       .getQueryCacheSizeInMetric();
   private static final int DELETE_SIZE = 20;
   private static final int DEFAULT_FETCH_SIZE = 10000;
+  private static final long MS_TO_MONTH = 30 * 86400_000L;
 
   private final IoTDBConfig config = IoTDBDescriptor.getInstance().getConfig();
   private final boolean enableMetric = config.isEnableMetricService();
@@ -556,11 +558,7 @@ public class TSServiceImpl implements TSIService.Iface, ServerContext {
       }
 
       if (plan instanceof GroupByTimePlan) {
-        GroupByTimePlan groupByTimePlan = (GroupByTimePlan) plan;
-        // the actual row number of group by query should be calculated from startTime, endTime and interval.
-        fetchSize = Math.min(
-            (int) ((groupByTimePlan.getEndTime() - groupByTimePlan.getStartTime()) / groupByTimePlan
-                .getInterval()), fetchSize);
+        fetchSize = Math.min(getFetchSizeForGroupByTimePlan((GroupByTimePlan) plan), fetchSize);
       }
 
       // get deduplicated path num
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -668,6 +666,23 @@ private TSExecuteStatementResp internalExecuteQueryStatement(String statement, l
...
+  private int getFetchSizeForGroupByTimePlan(GroupByTimePlan groupByTimePlan) {
+    int rows =  (int) ((groupByTimePlan.getEndTime() - groupByTimePlan.getStartTime()) / groupByTimePlan
+        .getInterval());
+    // edge case for group by months when starttime is in February
+    // ie. time range = [2020-02-02, 2020-03-02), interval = 1mo, rows will get 0
+    if (rows == 0 && groupByTimePlan.isIntervalByMonth()) {
+      Calendar calendar = Calendar.getInstance();
+      calendar.setTimeInMillis(groupByTimePlan.getStartTime());
+      calendar.add(Calendar.MONTH, (int) (groupByTimePlan.getInterval() / MS_TO_MONTH));
+      rows = calendar.getTimeInMillis() <= groupByTimePlan.getEndTime() ? 1 : 0;
```
mychaow, line 681: we do not include the endTime, so it's not right to  use  <= here, change it to <

mychaow (changes requested):

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
haimeiguo: we add starttime by interval months to check if the group by interval is within starttime to endtime range which includes endtime itself. if we use < sign, the group by result is not correct for edge case 

ie starttime = 2/1, endtime = 3/1 and interval = 1mo, we add 1 mo to 2/1, then calendar.getTimeInMillis() will be 3/1, which is equal to groupByTimePlan.getEndTime(), if use < sign, the result will not include 2/1 to 3/1 interval

mychaow (approved):

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -668,6 +666,23 @@ private TSExecuteStatementResp internalExecuteQueryStatement(String statement, l
...
   }
 
+  /*
+  calculate fetch size for group by time plan
+   */
+  private int getFetchSizeForGroupByTimePlan(GroupByTimePlan groupByTimePlan) {
+    int rows =  (int) ((groupByTimePlan.getEndTime() - groupByTimePlan.getStartTime()) / groupByTimePlan
+        .getInterval());
+    // edge case for group by months when starttime is in February
+    // ie. time range = [2020-02-02, 2020-03-02), interval = 1mo, rows will get 0
```
jixuan1989, line 676: The comments is not good as it may cause some misleading.

rows gets 0 is caused by: the end time - the start time < the time interval.

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -668,6 +666,23 @@ private TSExecuteStatementResp internalExecuteQueryStatement(String statement, l
...
+    if (rows == 0 && groupByTimePlan.isIntervalByMonth()) {
+      Calendar calendar = Calendar.getInstance();
+      calendar.setTimeInMillis(groupByTimePlan.getStartTime());
+      calendar.add(Calendar.MONTH, (int) (groupByTimePlan.getInterval() / MS_TO_MONTH));
+      rows = calendar.getTimeInMillis() <= groupByTimePlan.getEndTime() ? 1 : 0;
+    }
```
jixuan1989, line 682: if rows==0 and !plan.isIntervalByMonth()

set rows = 1; 

e.g., select avg(*) from root group by ( [now() -1h, now()), 7200s) will return 1 row

jixuan1989 (changes requested):
Test needed

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
haimeiguo: rounding down issue is mentioned on https://github.com/apache/iotdb/issues/2600

jixuan1989 (approved):

PR #2576, opened 2021-01-25
[IOTDB-1091] SDT improvement store last point

haimeiguo:
store last point when use sdt encoding
SDT compression is trigger by flushing, in current implementation the last point before flushing is not stored. 
The pr fixes this issue.

### DIFF
--- a/tsfile/src/main/java/org/apache/iotdb/tsfile/write/chunk/ChunkWriterImpl.java
+++ b/tsfile/src/main/java/org/apache/iotdb/tsfile/write/chunk/ChunkWriterImpl.java
@@ -155,20 +155,26 @@ public class ChunkWriterImpl implements IChunkWriter {
 
   @Override
   public void write(long time, long value) {
-    if (!isSdtEncoding || isLastPoint) {
+    // store last point for sdtEncoding, it still needs to go through encoding process
+    // in case it exceeds compdev and needs to store second last point
+    if (!isSdtEncoding || sdtEncoder.encodeLong(time, value)) {
+      pageWriter.write(isSdtEncoding ? sdtEncoder.getTime() : time,
+          isSdtEncoding ? sdtEncoder.getLongValue() : value);
+    }
+    if (isSdtEncoding && isLastPoint) {
       pageWriter.write(time, value);
-    } else if (sdtEncoder.encodeLong(time, value)) {
-      pageWriter.write(sdtEncoder.getTime(), sdtEncoder.getLongValue());
     }
     checkPageSizeAndMayOpenANewPage();
   }
 
   @Override
   public void write(long time, int value) {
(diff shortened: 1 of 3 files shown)


jixuan1989:
Hi, could you add all PRs about the SDT to the SDT related jira ticket?

tsfile/src/main/java/org/apache/iotdb/tsfile/write/chunk/ChunkWriterImpl.java
```diff
@@ -174,26 +174,26 @@ private void checkSdtEncoding() {
...
-      if (isSdtEncoding) {
-        //store last read time and value
-        time = sdtEncoder.getTime();
-        value = sdtEncoder.getLongValue();
-      }
+      pageWriter.write(isSdtEncoding ? sdtEncoder.getTime() : time,
+          isSdtEncoding ? sdtEncoder.getLongValue() : value);
+    }
+    // store last point for sdtEncoding
+    if (isSdtEncoding && isLastPoint) {
```
mychaow, line 181: last point may store twice?

mychaow (changes requested):

mychaow (approved):

tsfile/src/main/java/org/apache/iotdb/tsfile/write/chunk/ChunkWriterImpl.java
jixuan1989: if (! isSdtEncoding || isLastPoint) {
  pageWriter.write(time, value);
} else if (sdtEncoder.encodeLong(time, value)) {
 pageWriter.write(sdtEncoder.getTime(), sdtEncoder.getLongValue());
}

tsfile/src/main/java/org/apache/iotdb/tsfile/write/chunk/ChunkWriterImpl.java
```diff
@@ -80,7 +80,7 @@
   private Statistics<?> statistics;
 
   private boolean isSdtEncoding;
-
+  private boolean isLastPoint;
```
jixuan1989, line 83: add comment: 
used for SDT. When the ChunkWriter WILL write the last data point in the chunk, set it to true to tell SDT saves the point.

jixuan1989 (changes requested):
leave some commets to let others know how to use fields.

tsfile/src/main/java/org/apache/iotdb/tsfile/write/chunk/ChunkWriterImpl.java
jixuan1989: any problem about my logic?

qiaojialin (approved):

PR #2577, opened 2021-01-25
IOTDB-1126 unseq tsfile delete due to merge

wangchao316:
reproduce:
set storage group to root.test
insert into root.test.wf02.wt02(timestamp,status) values(1,true)
select * from root.test.wf02.wt02;
DELETE PARTITION root.test 0;
select * from root.test.wf02.wt02;
insert into root.test.wf02.wt02(timestamp,status) values(1,true)
select * from root.test.wf02.wt02;
flush
select * from root.test.wf02.wt02;
this is null.

Reason:
After the partition is deleted, if the insert data is non-sequential data. when the flush operation, 1 Unseq-file and 0 seq-file are merged,
The MergeTask traverses sequential files. If the sequential files are empty, the merge is skipped.
After the execution is complete, delete the unordered files. However, the merged file is not generated. Therefore, the data in the unordered file is lost.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/merge/task/MergeTask.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/merge/task/MergeTask.java
@@ -111,9 +111,8 @@ public class MergeTask implements Callable<Void> {
   }
 
   private void doMerge() throws IOException, MetadataException {
-    if (resource.getSeqFiles().size() == 0) {
-      logger.info("{} starts to merge {} seqFiles, so will abort task.", taskName ,
-              resource.getSeqFiles().size());
+    if (resource.getSeqFiles().isEmpty()) {
+      logger.info("{} no sequence file to merge into, so will abort task.", taskName);
       abort();
       return;
     }
--- a/server/src/main/java/org/apache/iotdb/db/engine/merge/task/MergeTask.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/merge/task/MergeTask.java
@@ -111,6 +111,12 @@ public class MergeTask implements Callable<Void> {
   }
 
   private void doMerge() throws IOException, MetadataException {
+    if (resource.getSeqFiles().size() == 0) {
+      logger.info("{} starts to merge {} seqFiles, so will abort task.", taskName ,
+              resource.getSeqFiles().size());
+      abort();
+      return;
+    }
     if (logger.isInfoEnabled()) {
       logger.info("{} starts to merge {} seqFiles, {} unseqFiles", taskName,
           resource.getSeqFiles().size(), resource.getUnseqFiles().size());
--- a/server/src/main/java/org/apache/iotdb/db/engine/merge/task/MergeMultiChunkTask.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/merge/task/MergeMultiChunkTask.java
@@ -286,6 +286,7 @@ public class MergeMultiChunkTask {
       try {
         futures.get(i).get();
       } catch (InterruptedException e) {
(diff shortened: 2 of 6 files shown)


wangchao316:
https://issues.apache.org/jira/browse/IOTDB-1126

wangchao316:
@qiaojialin @jixuan1989 @HTHou  hello， could you please review this ?
thanks.

server/src/main/java/org/apache/iotdb/db/engine/merge/task/MergeTask.java
```diff
@@ -111,6 +111,12 @@ private void abort() throws IOException {
   }
 
   private void doMerge() throws IOException, MetadataException {
+    if (resource.getSeqFiles().size() == 0) {
+      logger.info("{} starts to merge {} seqFiles, so will abort task.", taskName ,
```
SilverNarcissus, line 115: Maybe log like "No sequence file to merge into, abort task" can describe the reason of abort more specifically.

SilverNarcissus (approved):
Excellent work and detailed test! Only some log issue, look good to me~

server/src/main/java/org/apache/iotdb/db/engine/merge/task/MergeTask.java
wangchao316: Thanks @SilverNarcissus , I have alter log.. please review...

qiaojialin (commented):
I wonder why the newly inserted data is written to the unsequence file.
Is it better to clear the related latestFlushTime when deleting a partition?

wangchao316:
> I wonder why the newly inserted data is written to the unsequence file.
> Is it better to clear the related latestFlushTime when deleting a partition?

because the newly inserted data time < lastestFlushTime, so it is unsequence file .
 how latest flush time should set after clear the related latestFlushTime.
I think if do not have sequence file, we can do not merge.  util have a sequence file , as time goes on， insert data will have a sequence file.

qiaojialin:
> > I wonder why the newly inserted data is written to the unsequence file.
> > Is it better to clear the related latestFlushTime when deleting a partition?
> 
> because the newly inserted data time < lastestFlushTime, so it is unsequence file .
> how latest flush time should set after clear the related latestFlushTime.
> I think if do not have sequence file, we can do not merge. util have a sequence file , as time goes on， insert data will have a sequence file.

When deleting a partition, its latestFlushTime could be set to Long.MIN_VALUE

wangchao316:
> > > I wonder why the newly inserted data is written to the unsequence file.
> > > Is it better to clear the related latestFlushTime when deleting a partition?
> > 
> > 
> > because the newly inserted data time < lastestFlushTime, so it is unsequence file .
> > how latest flush time should set after clear the related latestFlushTime.
> > I think if do not have sequence file, we can do not merge. util have a sequence file , as time goes on， insert data will have a sequence file.
> 
> When deleting a partition, its latestFlushTime could be set to Long.MIN_VALUE

Thanks, I will do it.

wangchao316:
> > > > I wonder why the newly inserted data is written to the unsequence file.
> > > > Is it better to clear the related latestFlushTime when deleting a partition?
> > > 
> > > 
> > > because the newly inserted data time < lastestFlushTime, so it is unsequence file .
> > > how latest flush time should set after clear the related latestFlushTime.
> > > I think if do not have sequence file, we can do not merge. util have a sequence file , as time goes on， insert data will have a sequence file.
> > 
> > 
> > When deleting a partition, its latestFlushTime could be set to Long.MIN_VALUE
> 
> Thanks, I will do it.

@qiaojialin hello , I update latestFlushTime , could you review this?

qiaojialin (approved):
LGTM~

Issue #2573, opened 2021-01-25
Create timeseries root.sg.d1\n.s1 should be disabled using native session api


HTHou:
When we create timeseries or insert data into timeseries like `root.sg.d1\n.s1` with session api, I find out this kind of operation is executed successfully! 

Here is my test code.
```java

  public static void main(String[] args)
      throws IoTDBConnectionException, StatementExecutionException {
    session = new Session("127.0.0.1", 6667, "root", "root");
    session.open(false);

    //set session fetchSize
    session.setFetchSize(10000);

    try {
      createTimeseries();
      insertRecord();
    } catch (Exception e) {
      e.printStackTrace();
    }


  }

  private static void createTimeseries()
      throws IoTDBConnectionException, StatementExecutionException {

    session.createTimeseries("root.sg888.d2"+"\n"+".s1", TSDataType.INT64, TSEncoding.RLE,
        CompressionType.SNAPPY);

  }

  private static void insertRecord() throws IoTDBConnectionException, StatementExecutionException {
    String deviceId = "root.sg888.d88"+"\n";
    List<String> measurements = new ArrayList<>();
    List<TSDataType> types = new ArrayList<>();
    measurements.add("s1");
    types.add(TSDataType.TEXT);

    for (long time = 0; time < 100; time++) {
      List<Object> values = new ArrayList<>();
      values.add("sssss");
      session.insertRecord(deviceId, time, measurements, types, values);
    }
  }

```

After insert these data successfully, I tried to query these data but I cannot get any result, even the timeseries themselves.

 - Version master, v0.11, v0.10

PR #2578, opened 2021-01-26
[ISSUE-2573] Create timeseries root.sg.d1\n.s1 should be disabled using native session api

HTHou:
#2573

<hr>

This PR has:
  for code coverage.

<hr>


HTHou:
The current `NODE_MATCHER` is not correct and will cause the tests failed. We need a updated one to make it works.

session/src/test/java/org/apache/iotdb/session/SessionUT.java
```diff
@@ -170,7 +170,7 @@ public void testInsertByStrAndSelectFailedData()
       RowRecord record = dataSet.next();
       int nullCount = 0;
       for (int j = 0; j < 4; ++j) {
-        if (record.getFields().get(j) == null) {
+        if (record.getFields().get(j) == null || record.getFields().get(j).getDataType() == null) {
```
SteveYurongSu, line 173: Why we need the statement `record.getFields().get(j).getDataType() == null` here?

SteveYurongSu (approved):
LGTM.

session/src/test/java/org/apache/iotdb/session/SessionUT.java
HTHou: I added it because this UT failed on my local machine if missing it. Let me try to remove this and see the GitHub actions test results.

HTHou: All right, the tests passed. It seems this statement is not necessary....

jixuan1989 (approved):

PR #2583, opened 2021-01-26
disable the merge task in PullSnapshotTaskTest

neuyilan:
this pr is trying to fix the testSync UT in PullSnapshotTaskTest.  we do not change the planIndex when the tsfile is generated, so it's probability changed by the merge task.
image


jt2594838 (approved):

LebronAl (approved):
LGTM

PR #2591, opened 2021-01-27
Add mods merge

zhanglingzhe0820:
Currently, the mods file will not be deleted although the compaction is finished. 
This has some problem when user delete timeseries which is in compaction process(the delete operation will not work in the merged file). 
This problem happens when user delete timeseries while a compaction is working.
So in this pr, I merge mods file to the target file.


HTHou (approved):

jt2594838 (approved):
Good job, but do not forget to add some tests to ensure that the merged file contains all necessary modifications.

PR #2592, opened 2021-01-27
[To rel/0.11] delete mods after merge

zhanglingzhe0820:
Currently, the mods file will not be deleted although the compaction is finished. This has some problem when user delete timeseries which is in compaction process(the delete operation will not work in the merged file). So in this pr, I merge mods file to the target file.


HTHou (approved):

qiaojialin (changes requested):
delete modifications

PR #2593, opened 2021-01-27
[IOTDB-1133] make the vulnerabilities check configurable

neuyilan:
please see the following link for detail
https://issues.apache.org/jira/browse/IOTDB-1133

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -144,6 +144,8 @@
         <!-- By default, the argLine is empty-->
         <gson.version>2.8.6</gson.version>
         <argLine/>
+        <!-- whether skip the vulnerabilities check or not -->
+        <iotdb.vulnerabilities.skip>false</iotdb.vulnerabilities.skip>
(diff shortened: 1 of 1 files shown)


mychaow (approved):
Good job!

jixuan1989 (changes requested):
Hi,
1. we do not need add any other properties, maven enforcer plugin has its property: just use: `-Denforcer.skip=true`
2. add the parameter to github actions.

jixuan1989:
> Hi,
> 
> 1. we do not need add any other properties, maven enforcer plugin has its property: just use: `-Denforcer.skip=true`
> 2. add the parameter to github actions.

And, if you want to skip the test by default, yes we can set the parameter as true by default.

Only the release manager needs to enable it.

qiaojialin:
+1 to set it to false by default

jixuan1989:
done in #2603

PR #2597, opened 2021-01-29
[IOTDB-1140] [IOTDB-1125] optimize regular data encoding

wangchao316:
current regular data encoding algorithm：

Calculate the difference between two adjacent values. The smallest difference is used as the equal-frequency frequency.
Determine the data range of this batch of data based on the difference between the last value and the first value.
Traverse this batch of data, use a BitSet, compare the difference between two adjacent values with the same frequency, and save the value true by default,
If the value is not equal to the equal frequency, calculate the number of equal frequency differences and set the value to false at the corresponding position, indicating that the point is a missing point.
 

this algorithm only can identity missing point,  if have error point , it will throw exception..

because BitSet only can do this thing,  indicates whether the same frequency exists in a segment of data

 

But there is some optimize point..

If there is an abnormal value in a column of values, the algorithm is deviated if the difference is directly obtained to the minimum value.

sample: 1000，1100，1800，1400，1500... 

current algorithm be do not use...

1800 is a error point,  we should identity error point,  revise data. 

revise data should be : 1000，1100，1300，1400，1500


wangchao316:
https://issues.apache.org/jira/browse/IOTDB-1125

wangchao316:
hi @HTHou @qiaojialin @jixuan1989 , could you please review this ? 
Thanks

jixuan1989:
Hi, I am confused by your description...

wangchao316:
> Hi, I am confused by your description...

定频编码要求数据为定频数据，存在几种情况：
1. 一次memtable 中数据不是定频数据，在flush的时候encoding会报错，直接导致本次flush的所有序列都会失败。（这是我测试的问题，本PR已解决，解决方法：将编码错误的数据过滤掉，不影响其他序列flush）
2. 单次memtable是定频的，但是两个tsfile文件中加起来的数据不是定频的，经过合并后，会导致在读取数据解码的时候，发生错误。（本PR已解决）（这是iotdb-1125中的问题）
3. 如果一旦flush失败，会设置成只读模式，没有恢复机制，我提了一个优化单（IOTDB-1136），将在这个单子里面解决。
Constant-frequency encoding requires constant-frequency data. There are several cases:
1. The data in the memtable is not constant-frequency data. During flushing, an error is reported during encoding. As a result, all sequences fail to be flushed. (This is a problem in the test and has been resolved in this PR.)
2. A single memtable is fixed-frequency, but the sum of data in the two TSfile files is not fixed-frequency. After the combination, an error occurs during data decoding. (This issue has been resolved in this PR.) (This issue is in iotdb-1125.)
3. If the flush fails, the system sets the data to the read-only mode and does not have a recovery mechanism. I have submitted an optimization ticket (IOTDB-1136) to resolve this issue.

jixuan1989:
OK.  Got it.

But IMO, I think it is a bug of "regular" encoding implementation, rather than solving it by rollback the WAL. 
see https://issues.apache.org/jira/browse/IOTDB-1136?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&focusedCommentId=17274179#comment-17274179

jixuan1989:
Ses the design https://lists.apache.org/thread.html/f612326ca8708981b5a2da29f52b826105bbb55dc840281c5bf6af4f%40%3Cdev.iotdb.apache.org%3E

wangchao316:
> OK. Got it.
> 
> But IMO, I think it is a bug of "regular" encoding implementation, rather than solving it by rollback the WAL.
> see https://issues.apache.org/jira/browse/IOTDB-1136?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&focusedCommentId=17274179#comment-17274179

Thanks ， I will try to alter regular" encoding...  if can  not , I will give reason... we discuss again...

jixuan1989:
Thanks. look forward to making the encoder really available~

wangchao316:
@jixuan1989 @jialin  hi, could you please review this code ?

jixuan1989:
Hi, Current description of the PR is clear.
But IMO,  the `regular` encoding is more suitable for the timestamp column... how do you think? (is there some cases that the value column obeys the pattern?)


> sample: 1000，1100，1800，1400，1500...
> current algorithm be do not use...
> 1800 is a error point, we should identity error point, revise data.
> revise data should be : 1000，1100，1300，1400，1500

This is ... dangerous, as we modify the raw data directly and do not tell the user in time.

wangchao316:
> Hi, Current description of the PR is clear.
> But IMO, the `regular` encoding is more suitable for the timestamp column... how do you think? (is there some cases that the value column obeys the pattern?)
> 
> > sample: 1000，1100，1800，1400，1500...
> > current algorithm be do not use...
> > 1800 is a error point, we should identity error point, revise data.
> > revise data should be : 1000，1100，1300，1400，1500
> 
> This is ... dangerous, as we modify the raw data directly and do not tell the user in time.

@jixuan1989  Thanks review.
（
是的，定频编码适合时间戳。
目前因为一个时间戳数据错误就导致整个ts flush失败，从而开启只读模式。

我觉得我们应该做一些事保证一个服务的稳定性。
因为
1. 我们不能确定用户指定了定频编码，就能按照时间戳数据顺序插入。
2.如果插入这种数据，需要有容错能力，最起码不能影响同一个flush中的其他序列。有以下几个方法：
（2.1） 要么修改这个数据。
（2.2） 要么删除这个数据。
（2.3） 或者报错，但是此时数据已经写了wal,和内存，只是在flush的时候出错，所以是否需要清理该数据wal.

我目前的解决办法是：
1. 尝试修复定频数据，并打印警告日志。
2.如果不能修复，输入的不是定频数据，那么我们抛出警告异常，然后删除错误数据，并且在日志中打印出来。）
Yes, constant frequency coding is suitable for timestamps.
However, the entire TS flush cannot fail because of a timestamp data error. In this case, the read-only mode cannot be enabled.

Because
1. If you cannot determine whether the code is specified by the user, the data can be inserted in the sequence of timestamp data.
2. If such data is inserted, the error tolerance capability is required.
(2.1) Modify the data.
(2.2) Delete the data.
(2.3) An error is reported. However, the data has been written to the wal and memory, and an error occurs only during flushing. Therefore, do you need to clear the wal?

My current solution is:
1. Restore the constant-frequency data.
2. If the fault cannot be rectified and the input data is not constant-frequency data, a warning exception is thrown, the error data is deleted, and the error data is printed in the log.

Jialin:
> @jixuan1989 @Jialin hi, could you please review this code ?

Just FYI, you guys might tag the wrong person for review :)

PR #2605, opened 2021-01-29
Add level merge to  "merge" command

zhanglingzhe0820:
Currently, the cli `merge` can just execute unseq compaction, it is not conveniently for user to execute compaction. So in this PR, I change the `merge` command to execute both level compaction and unseq merge. And also, this PR may fix the 'file not found' ci problem.
There is a windows ci problem now.
image
Every time this ci problem occurs, there will be a file not found error and it is all windows ci. Although the wrong test was not manually merged, I guessed that it was because the last test reported 'file not found' error affected this test
So I speculate that it is caused by an unseq compaction problem.
image
This error happens because we submit an unseq compaction task manually when a level compaction is working, the unseq compaction will get the current tsfile list and wait for the level compaction to be finished. However, the level compaction may merge the file which will be used in the unseq compaction process. The unseq compaction then cannot find the file to use. Then cause the ci problem.
So in this pr, we change the function of `merge` cli from the execution of unseq compaction to the execution of level compaction+unseq compaction. It just starts a normal compaction process which do not have this file-related problem.


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -1886,13 +1889,9 @@ private void loadUpgradedResources(List<TsFileResource> resources, boolean isseq
...
-    writeLock();
-    try {
-      this.tsFileManagement.merge(fullMerge, tsFileManagement.getTsFileList(true),
-          tsFileManagement.getTsFileList(false), dataTTL);
-    } finally {
-      writeUnlock();
+  public void merge(boolean isFullMerge) {
+    for (long timePartitionId : timePartitionIdVersionControllerMap.keySet()) {
+      executeCompaction(timePartitionId, isFullMerge);
     }
```
jt2594838, line 1895: Now that the write lock is removed, is it possible that if more than one client issues concurrent merge requests, the result of ` tsFileManagement.forkCurrentFileList` and `tsFileManagement.setForceFullMerge` will be messed up, when the merge requests contain different arguments?

jt2594838 (commented):
How you intend to solve the problem is a bit out of my expectation. Removing current locks may introduce other issues potentially.
I think the problem is actually one merge does not know that another merge is compacting its chosen files. Can we avoid this from the beginning somehow? For example, a merging mark for those files which are already chosen.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
zhanglingzhe0820: The param `compactionMergeWorking` in `StorageGroupProcessor` ensures that there is only one compaction thread.

zhanglingzhe0820: I solve the problem in this way because I think it is also necessary to update the function of the command `merge`.

jt2594838: Unfortunately, a simple if-else is barely enough for concurrency control, because the condition check and the flag set action are not atomic,  multiple threads may still enter that block concurrently.

jixuan1989:
> However, the level compaction may merge the file which will be used in the unseq compaction process. The unseq compaction then cannot find the file to use. Then cause the ci problem.

Why it affects query?

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
zhanglingzhe0820: okay, I will add lock in the function executeCompaction().

zhanglingzhe0820:
> > However, the level compaction may merge the file which will be used in the unseq compaction process. The unseq compaction then cannot find the file to use. Then cause the ci problem.
> 
> Why it affects query?

Every time this ci problem occurs, there will be a file not found error and it is all windows ci. Although the wrong test was not manually merged, I guessed that it was because the last test reported 'file not found' error affected this test
So I speculate that it is caused by an unseq compaction problem.
And Why it affects query?
When compaction meets error, it may destroy the current tearDown() process and leave some file that not be deleted. The data of these remaining files conflicts with the data inserted in the next test, resulting in query errors

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -1788,6 +1789,7 @@ private void executeCompaction(long timePartition, boolean fullMerge){
       logger.info("{} last compaction merge task is working, skip current merge",
           logicalStorageGroupName + "-" + virtualStorageGroupId);
     }
+    writeUnlock();
```
jt2594838, line 1792: Always do unlock in a `finally` block.

jt2594838 (approved):

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
zhanglingzhe0820: okay

server/src/main/java/org/apache/iotdb/db/engine/merge/manage/MergeManager.java
```diff
@@ -252,8 +252,7 @@ public ServiceType getID() {
 
   private void mergeAll() {
     try {
-      StorageEngine.getInstance()
-          .mergeAll(IoTDBDescriptor.getInstance().getConfig().isForceFullMerge());
+      StorageEngine.getInstance().mergeAll(IoTDBDescriptor.getInstance().getConfig().isForceFullMerge());
```
HTHou, line 255:
```suggestion
      StorageEngine.getInstance()
          .mergeAll(IoTDBDescriptor.getInstance().getConfig().isForceFullMerge());
```

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -1940,19 +1940,26 @@ private void closeUnsealedTsFileProcessorCallBack(TsFileProcessor tsFileProcesso
         "signal closing storage group condition in {}",
         logicalStorageGroupName + "-" + virtualStorageGroupId);
 
+    executeCompaction(
+        tsFileProcessor.getTimeRangeId(),
+        IoTDBDescriptor.getInstance().getConfig().isForceFullMerge());
+  }
+
+  private void executeCompaction(long timePartition, boolean fullMerge){
```
HTHou, line 1948:
```suggestion
  private void executeCompaction(long timePartition, boolean fullMerge) {
```

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -1940,19 +1940,26 @@ private void closeUnsealedTsFileProcessorCallBack(TsFileProcessor tsFileProcesso
...
           logicalStorageGroupName + "-" + virtualStorageGroupId);
       try {
         // fork and filter current tsfile, then commit then to compaction merge
-        tsFileManagement.forkCurrentFileList(tsFileProcessor.getTimeRangeId());
+        tsFileManagement.forkCurrentFileList(timePartition);
+        tsFileManagement.setForceFullMerge(fullMerge);
         CompactionMergeTaskPoolManager.getInstance()
             .submitTask(
                 tsFileManagement
                 .new CompactionMergeTask(
```
HTHou, line 1961:
```suggestion
                .new CompactionMergeTask(this::closeCompactionMergeCallBack,·timePartition));
```

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -1940,19 +1940,26 @@ private void closeUnsealedTsFileProcessorCallBack(TsFileProcessor tsFileProcesso
...
-                    this::closeCompactionMergeCallBack, tsFileProcessor.getTimeRangeId()));
+                    this::closeCompactionMergeCallBack, timePartition));
```
HTHou, line 1962:
```suggestion
```

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -1940,19 +1940,25 @@ private void closeUnsealedTsFileProcessorCallBack(TsFileProcessor tsFileProcesso
...
         // fork and filter current tsfile, then commit then to compaction merge
-        tsFileManagement.forkCurrentFileList(tsFileProcessor.getTimeRangeId());
+        tsFileManagement.forkCurrentFileList(timePartition);
+        tsFileManagement.setForceFullMerge(fullMerge);
         CompactionMergeTaskPoolManager.getInstance()
             .submitTask(
                 tsFileManagement
-                .new CompactionMergeTask(
-                    this::closeCompactionMergeCallBack, tsFileProcessor.getTimeRangeId()));
+                .new CompactionMergeTask(this::closeCompactionMergeCallBack,·timePartition));
```
HTHou, line 1961:
```suggestion
                .new CompactionMergeTask(this::closeCompactionMergeCallBack, timePartition));
```

PR #2608, opened 2021-02-01
[To NewTsFileV2] Change the way of checking page in Upgrade Tool

HTHou:
<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/tools/upgrade/TsFileOnlineUpgradeTool.java
+++ b/server/src/main/java/org/apache/iotdb/db/tools/upgrade/TsFileOnlineUpgradeTool.java
@@ -172,8 +172,12 @@ public class TsFileOnlineUpgradeTool implements AutoCloseable {
               dataInChunk.add(pageData);
               dataSize -= (Integer.BYTES * 2 // the bytes of uncompressedSize and compressedSize
                   + pageHeader.getStatistics().getSerializedSize()
+                  // count, startTime, endTime
+                  + 24
+                  // new boolean StatsSize is 8 bytes larger than old one
+                  + (pageHeader.getStatistics().getStatsSize() 
+                      - (dataType == TSDataType.BOOLEAN ? 8 : 0))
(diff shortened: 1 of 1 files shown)


wangchao316:
hi, HTHou, Do we need to process abnormal data during an online upgrade? Do we directly filter abnormal data or interrupt the upgrade?

HTHou:
> hi, HTHou, Do we need to process abnormal data during an online upgrade? Do we directly filter abnormal data or interrupt the upgrade?

Hi wangchao, normally if the tsfile isn't sealed, there won't be a tsfile resource with it. And the IoTDB will tell users to run `flush` before upgrading. 

Or if the tsfile is sealed and there is a resource file with it, but there are some bytes errors in TsFile. It may have different scenarios.

1. the error is in Magic String or version number
2. the error is in Metadata part
3. the error is in data part

For 1 and 3, the upgrade tool will throw the exception, keep this file and continue to upgrade the next file.
For 2, the upgrade tool will upgrade the file normally, since we don't need to read the old metadata for upgrading.

wangchao316:
> > hi, HTHou, Do we need to process abnormal data during an online upgrade? Do we directly filter abnormal data or interrupt the upgrade?
> 
> Hi wangchao, normally if the tsfile isn't sealed, there won't be a tsfile resource with it. And the IoTDB will tell users to run `flush` before upgrading.
> 
> Or if the tsfile is sealed and there is a resource file with it, but there are some bytes errors in TsFile. It may have different scenarios.
> 
> 1. the error is in Magic String or version number
> 2. the error is in Metadata part
> 3. the error is in data part
> 
> For 1 and 3, the upgrade tool will throw the exception, keep this file and continue to upgrade the next file.
> For 2, the upgrade tool will upgrade the file normally, since we don't need to read the old metadata for upgrading.

Thanks, this is ok..

PR #2612, opened 2021-02-02
[IOTDB-1138]Do not remove the overlap files which is merging when do snapshot

neuyilan:
After discussion with @zhanglingzhe0820 , there are three possible concurrency problems related to merge

1. After snapshot pulls the file, it will delete the local overlap file
2. The checkttl operation will also delete the expired data.
3. The user initiatively initiates the data deletion operation.

This PR only considers the first scenario. We need to comprehensively consider the concurrency problem to solve all the above problems.

I have opened an issue
https://github.com/apache/iotdb/issues/2618

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
@@ -2182,6 +2182,7 @@ public class StorageGroupProcessor {
       TsFileResource existingTsFile = iterator.next();
       if (newTsFile.isPlanRangeCovers(existingTsFile)
           && !newTsFile.getTsFile().equals(existingTsFile.getTsFile())
+          && !existingTsFile.isMerging()
           && existingTsFile.tryWriteLock()) {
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -2182,6 +2182,7 @@ private void removeFullyOverlapFiles(TsFileResource newTsFile, Iterator<TsFileRe
       TsFileResource existingTsFile = iterator.next();
       if (newTsFile.isPlanRangeCovers(existingTsFile)
           && !newTsFile.getTsFile().equals(existingTsFile.getTsFile())
+          && !existingTsFile.isMerging()
```
wangchao316, line 2185: hi ,  I have a qeustion.
if do not delete file which is merging in snapshot.  after merged finished, this tsfile will merge to a new tsfile..
this new tsfile will be overlap and snapshot tsfile..

neuyilan:
Close this PR, because the #2811   have fixed the problem

PR #2614, opened 2021-02-02
[IOTDB-1142] Fix deleting timeseries error

rhmangmang:
The error is described at https://issues.apache.org/jira/browse/IOTDB-1142


jixuan1989:
Hei, PR description (what it is, why it happens, and how to solve it) must be written.

weizihan0110:
I got it.

weizihan0110:
> Hi, as is discussed in mail list and JIRA, we intend to use `*` and `**`:
> 
> We could reference the usage of wild star in Linux :
> 
> ```
> /tmp/* matches any file in /tmp. // Substitutes for any number of characters, except /.
> /tmp/** matches all files and directories underneath /tmp. // Substitutes for any number of characters, including /.
> ```
> 
> So we should also consider that `**` also contains zero layer.
> For example, if we have some storage groups:
> 
> ```
> root.a.sg1
> root.a.b.sg2
> root.b.sg3
> ```
> 
> For "delete storage group root.a", we do not delete any sg.
> For "delete storage group root.a.*" we delete root.a.sg1
> For "delete storage group root.a.sg1.*" we do not delete any sg.
> For "delete storage group root.a.**" we delete root.a.sg1, root.a.b.sg2
> 
> If we also have another storage group `root.a`:
> 
> For "delete storage group root.a", we delete root.a
> For "delete storage group root.a.*" we delete root.a.sg1 (because * does not include zero layer)
> For "delete storage group root.a.**" we delete root.a, root.a.sg1, root.a.b.sg2 (because ** includes zero layer)

Hi, @samperson1997. I got it.

HTHou:
We disabled the usage of nested timeseries path, therefore it would not be a problem.

PR #2621, opened 2021-02-03
IOTDB-1140 optimize regular data encoding

wangchao316:
current regular data encoding algorithm：

Calculate the difference between two adjacent values. The smallest difference is used as the equal-frequency frequency.
Determine the data range of this batch of data based on the difference between the last value and the first value.
Traverse this batch of data, use a BitSet, compare the difference between two adjacent values with the same frequency, and save the value true by default,
If the value is not equal to the equal frequency, calculate the number of equal frequency differences and set the value to false at the corresponding position, indicating that the point is a missing point.

this algorithm only can identity missing point, if have error point , it will throw exception..

because BitSet only can do this thing, indicates whether the same frequency exists in a segment of data

But there is some optimize point..

If there is an abnormal value in a column of values, the algorithm is deviated if the difference is directly obtained to the minimum value.

sample: 1000，1100，1800，1400，1500...

current algorithm be do not use...

1800 is a error point, we should identity error point, revise data.

revise data should be : 1000，1100，1300，1400，1500

After discussion , solution:
1. The value cannot be in regular encoding.
2. time_encoder can not alter after service start.


wangchao316:
@jixuan1989 @qiaojialin  hi , could you please review this PR?
Thanks...

qiaojialin (approved):
Great!

server/src/test/java/org/apache/iotdb/db/integration/IoTDBCheckConfigIT.java
```diff
@@ -0,0 +1,149 @@
...
+import org.apache.iotdb.db.conf.IoTDBConfigCheck;
+import org.apache.iotdb.db.conf.IoTDBDescriptor;
+import org.apache.iotdb.db.engine.fileSystem.SystemFileFactory;
+import org.apache.iotdb.db.service.IoTDB;
+import org.apache.iotdb.db.utils.EnvironmentUtils;
+import org.apache.iotdb.tsfile.common.conf.TSFileConfig;
+import org.apache.iotdb.tsfile.common.conf.TSFileDescriptor;
+import org.junit.*;
+
+import java.io.*;
```
yuqi1129, line 30: It's no advice to import star in java

PR #2623, opened 2021-02-04
[IOTDB-1147] Fix the concurrent bug in FlushManager cause NoSuchElementException e…

neuyilan:
for more details, please see https://issues.apache.org/jira/browse/IOTDB-1147

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/flush/FlushManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/flush/FlushManager.java
@@ -136,10 +136,8 @@ public class FlushManager implements FlushManagerMBean, IService {
       if (LOGGER.isDebugEnabled()) {
         if (tsFileProcessor.isManagedByFlushManager()) {
           LOGGER.debug(
-              "{} is already in the flushPool, the first one: {}, the given processor flushMemtable number = {}",
+              "{} is already in the flushPool, the given processor flushMemtable number = {}",
               tsFileProcessor.getTsFileResource().getTsFile().getAbsolutePath(),
-              tsFileProcessorQueue.isEmpty() ? "empty now"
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/apache/iotdb/db/engine/flush/FlushManager.java
```diff
@@ -136,10 +136,8 @@ public void registerTsFileProcessor(TsFileProcessor tsFileProcessor) {
       if (LOGGER.isDebugEnabled()) {
         if (tsFileProcessor.isManagedByFlushManager()) {
           LOGGER.debug(
-              "{} is already in the flushPool, the first one: {}, the given processor flushMemtable number = {}",
+              "{} is already in the flushPool, the given processor flushMemtable number = {}",
               tsFileProcessor.getTsFileResource().getTsFile().getAbsolutePath(),
-              tsFileProcessorQueue.isEmpty() ? "empty now"
-                  : tsFileProcessorQueue.getFirst().getStorageGroupName(),
               tsFileProcessor.getFlushingMemTableSize());
```
qiaojialin, line 141: Thanks for bringing this up. Relook this process may have other problems...

Current thread setManagedByFlushManager(true).
FlushThread flush a memtable and setManagedByFlushManager(false)
Current thread check tsfileProcessor.isManagedByFlushManager() return false

So, it's better to put all logs before submit(FlushThread)   line 126.

jixuan1989: before line 126 or line 125?

neuyilan: Thank you all, I've reorganized the debug logs. PTAL

qiaojialin (approved):

PR #2632, opened 2021-02-04
one-click build script of N-node n-1 copy pseudo-distributed.

weizihan0110:


docs/zh/UserGuide/Server/Cluster Setup.md
```diff
@@ -69,6 +69,55 @@ or
 > nohup cluster\target\cluster-0.11.0-SNAPSHOT1\sbin\start-node.bat  -internal_meta_port 9005 -internal_data_port 40012 -cluster_rpc_port 55561
 > nohup cluster\target\cluster-0.11.0-SNAPSHOT2\sbin\start-node.bat  -internal_meta_port 9007 -internal_data_port 40014 -cluster_rpc_port 55562
 ```
+## n节点n-1副本伪分布式一键搭建脚本
```
neuyilan, line 72: Where to set the n-1 copy? At present, the default copy is 2 copies. I do not see the logic to modify the number of copies in the script?

docs/zh/UserGuide/Server/Cluster Setup.md
```diff
@@ -69,6 +69,55 @@ or
...
+#第一步（复制IoTDB集群包）
+cp -rf $CLUSTER_JAR_PATH/$CLUSTER_SNAPSHOT_NAME $CLUSTER_JAR_PATH/$temp_name
+
+#第二步（分别修改rpc与jmx端口）
+sed -i -e "s/${RPC_PORT}/${rpc_port}/g" $CLUSTER_JAR_PATH/$temp_name/conf/iotdb-engine.properties
+sed -i -e "s/${JMX_PORT}/${jmx_port}/g" $CLUSTER_JAR_PATH/$temp_name/conf/iotdb-env.sh
+done  
+
+#第三步（更改权限）
+chmod -R 777 $CLUSTER_JAR_PATH/
```
neuyilan, line 102: If there are other files under the current folder, the permissions of these files have been modified, which is a little too big? Just modify some script execution permissions of iotdb is ok.

neuyilan (changes requested):
Besides, there[1] are scripts that can deploy pseudo-distributed or distributed easily, maybe we can combine both?
[1] https://github.com/LebronAl/QuickDeployForIoTDBTesting

docs/zh/UserGuide/Server/Cluster Setup.md
weizihan0110: NODE_NUM can be any number.

docs/zh/UserGuide/Server/Cluster Setup.md
weizihan0110: current folder are ./cluster/target/.
I'll make it clearer

docs/zh/UserGuide/Server/Cluster Setup.md
weizihan0110: > Besides, there[1] are scripts that can deploy pseudo-distributed or distributed easily, maybe we can combine both?
> [1] https://github.com/LebronAl/QuickDeployForIoTDBTesting

Hi, @neuyilan:
   I'll be so glad if you can contribute scripts at[1] to IoTDB ：）

neuyilan: You may not understand me. The number of copies is `default_ replica_ num` parameter in iotdb-cluster.properties to control,  `NODE_NUM`  can not control the number of copies, it can only  control the number of nodes. For example, set NODE_NUM = 5, but the copy is still 2, not 4.

docs/zh/UserGuide/Server/Cluster Setup.md
neuyilan: The user may not want to deploy the cluster in the target folder, we can't take it for granted. Just modify some script execution permissions of IOTDB_PATH/sbin or IOTDB_PATH/conf is ok enough.

docs/UserGuide/Server/Cluster Setup.md
```diff
@@ -76,6 +76,55 @@ When both exist, the specified configuration item will overwrite the configurati
...
+## One-click build script of n nodes and n-1 replicas pseudo-distributed.
+```
+#!/bin/bash  
+
+#config the number of node
+NODE_NUM=3
+
+export CLUSTER_JAR_PATH=./
+#CLUSTER_SNAPSHOT_NAME is something like cluster-0.11.0-SNAPSHOT
+CLUSTER_SNAPSHOT_NAME=$(find $CLUSTER_JAR_PATH -name "cluster-*-SNAPSHOT" | sed 's#.*/##')
```
neuyilan, line 88: After 0.12 was released, the folder name will not contain SNAPSHOT, it may looks like `iotdb-cluster-0.12-bin.zip`, if the user wants to use this script to deploy the cluster,  it will not work.

neuyilan (changes requested):

PR #2635, opened 2021-02-05
[IOTDB-1148]fix the client leak of client pool&use remote schema cache when check timeseries exist or not

neuyilan:
for more details, please see https://issues.apache.org/jira/browse/IOTDB-1148

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
@@ -673,10 +673,12 @@ public class CMManager extends MManager {
         } else {
           SyncDataClient syncDataClient = null;
           try {
-            syncDataClient = metaGroupMember.getClientProvider().getSyncDataClient(node,
-                RaftServer.getReadOperationTimeoutMS());
-            result = syncDataClient
-                .getUnregisteredTimeseries(partitionGroup.getHeader(), seriesList);
+            syncDataClient =
+                metaGroupMember
+                    .getClientProvider()
+                    .getSyncDataClient(node, RaftServer.getReadOperationTimeoutMS());
+            result =
+                syncDataClient.getUnregisteredTimeseries(partitionGroup.getHeader(), seriesList);
           } finally {
             ClientUtils.putBackSyncClient(syncDataClient);
           }
@@ -855,8 +857,10 @@ public class CMManager extends MManager {
     } else {
       SyncDataClient syncDataClient = null;
       try {
-        syncDataClient = metaGroupMember.getClientProvider().getSyncDataClient(node,
-            RaftServer.getReadOperationTimeoutMS());
+        syncDataClient =
+            metaGroupMember
+                .getClientProvider()
+                .getSyncDataClient(node, RaftServer.getReadOperationTimeoutMS());
         PullSchemaResp pullSchemaResp = syncDataClient.pullTimeSeriesSchema(request);
         ByteBuffer buffer = pullSchemaResp.schemaBytes;
         int size = buffer.getInt();
@@ -867,7 +871,6 @@ public class CMManager extends MManager {
       } finally {
         ClientUtils.putBackSyncClient(syncDataClient);
       }
-
     }
 
     return schemas;
@@ -1093,8 +1096,10 @@ public class CMManager extends MManager {
     } else {
       SyncDataClient syncDataClient = null;
       try {
-        syncDataClient = metaGroupMember.getClientProvider().getSyncDataClient(node,
-            RaftServer.getReadOperationTimeoutMS());
+        syncDataClient =
+            metaGroupMember
+                .getClientProvider()
+                .getSyncDataClient(node, RaftServer.getReadOperationTimeoutMS());
         result = syncDataClient.getAllPaths(header, pathsToQuery, withAlias);
       } finally {
         ClientUtils.putBackSyncClient(syncDataClient);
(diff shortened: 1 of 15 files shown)


cluster/src/main/java/org/apache/iotdb/cluster/query/reader/RemoteSeriesReaderByTimestamp.java
```diff
@@ -86,20 +86,23 @@ private ByteBuffer fetchResultAsync(long timestamp) throws IOException {
...
-      curSyncClient.putBack();
-      return buffer;
     } catch (TException e) {
       //try other node
       if (!sourceInfo.switchNode(true, timestamp)) {
         return null;
       }
       return fetchResultSync(timestamp);
+    } finally {
+      if (curSyncClient != null) {
```
mychaow, line 103: ClientUtils.putBackSyncClient(client)   is better.

cluster/src/main/java/org/apache/iotdb/cluster/query/reader/RemoteSimpleSeriesReader.java
```diff
@@ -136,20 +136,23 @@ private ByteBuffer fetchResultAsync() throws IOException {
...
-      return buffer;
     } catch (TException e) {
       //try other node
       if (!sourceInfo.switchNode(false, lastTimestamp)) {
         return null;
       }
       return fetchResultSync();
+    } finally {
+      if (curSyncClient != null) {
+        curSyncClient.putBack();
```
mychaow, line 154: same.

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
```diff
@@ -623,13 +648,16 @@ public void setCoordinator(Coordinator coordinator) {
...
-          result = syncDataClient.getUnregisteredTimeseries(partitionGroup.getHeader(), seriesList);
-          ClientUtils.putBackSyncClient(syncDataClient);
+          SyncDataClient syncDataClient = null;
+          try {
+            syncDataClient = metaGroupMember.getClientProvider().getSyncDataClient(node,
+                RaftServer.getReadOperationTimeoutMS());
+            result = syncDataClient
+                .getUnregisteredTimeseries(partitionGroup.getHeader(), seriesList);
+          } finally {
+            ClientUtils.putBackSyncClient(syncDataClient);
```
yuqi1129, line 658: What about getSyncDataClient throws exception and syncDataClient is null and at this circumstance is still null?

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/PublicBAOS.java
```diff
@@ -61,7 +61,7 @@ public PublicBAOS(int size) {
    * @exception  IOException  if an I/O error occurs.
    */
   @Override
-  public void writeTo(OutputStream out) throws IOException {
+  public synchronized void writeTo(OutputStream out) throws IOException {
```
qiaojialin, line 64: Hi, this is removed on purpose, cause this will impact the Flush pipeline.

See: https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=173085039

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/PublicBAOS.java
```diff
@@ -76,12 +76,12 @@ public void writeTo(OutputStream out) throws IOException {
    * reusing the already allocated buffer space.
    */
   @Override
-  public void reset() {
+  public synchronized void reset() {
```
qiaojialin, line 79: the same, we remove the synchronized in all override method in PublicBAOS

wangchao316:
Generally, the connection pool is used to obtain connections from the connection pool. After the close connection is used, the close connection only puts the connection back to the connection pool. The connection pool maintains the lifecycle of the connection.

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
neuyilan: null client value can not be put back to the client pool, this is guaranteed in function putBackSyncClient in ClientUtils,
if the client is null, the two if in the judgment condition is not tenable, so it will not be put back.

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/PublicBAOS.java
neuyilan: Got it, I prefer to post this link here in this function, so that other developers know about the problem, otherwise, sonar will report a bug, and other developers may fix it.

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/PublicBAOS.java
neuyilan: Got it, I prefer to post this link here in this function, so that other developers know about the problem, otherwise, sonar will report a bug, and other developers may fix it.

neuyilan:
> Generally, the connection pool is used to obtain connections from the connection pool. After the close connection is used, the close connection only puts the connection back to the connection pool. The connection pool maintains the lifecycle of the connection.

Actually, it's not exactly right. Not all client life cycles are processed in the client pool. When getting a client for remote process use, there may be problems. For example, a client may fail to connect due to network delay or server downtime. At this time, we need to close the connection. When you put the client in the client pool, if you find that the connection is closed, you will not put it in the pool. The role of the client pool is to save some clients that have created connections. When you want to connect to a server, you don't need to create connections again, which saves the overhead of creating and closing connections frequently.

jt2594838 (approved):

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
```diff
@@ -393,6 +414,10 @@ public synchronized MeasurementMNode get(PartialPath key) {
         return null;
       }
     }
+
+    public synchronized boolean containsKey(PartialPath key) {
+      return cache.containsKey(key);
+    }
```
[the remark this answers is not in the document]
neuyilan: Since all methods in RemoteMetaCache are synchronized, it is safe to modify the cache. However, in CMManager, functions in RemoteMetaCache may be called many times in a function, which may be due to the serialization of cache operations in CMManager, so the lock is added?

Therefore, I think in CMManager, whether the cache operation is locked or not is OK. Locking can ensure that all cache operations in CMManager are serial.  However, since RemoteMetaCache itself is thread-safe, maybe it's also OK not to lock it.

Issue #2661, opened 2021-02-08
Statistics value in SyncClientPool/AsyncClientPool have concurrency problem


yuqi1129:
**Describe the bug**
Though field nodeClientNumMap and clientCaches is protected by object lock, but after wake up from monitor lock , this two field should also guarded by lock or concurrency problem will occur

**To Reproduce**
See code:
DataClientProviderTest
```java
@Test
  public void testSync() throws IOException, InterruptedException {
    Node node = new ClusterNode();
    node.setDataPort(9003).setIp("localhost");
    ServerSocket serverSocket = new ServerSocket(node.getDataPort());
    Thread listenThread = new Thread(() -> {
      while (!Thread.interrupted()) {
        try {
          serverSocket.accept();
        } catch (IOException e) {
          return;
        }
      }
    });
    listenThread.start();

    try {
      boolean useAsyncServer = ClusterDescriptor.getInstance().getConfig().isUseAsyncServer();
      ClusterDescriptor.getInstance().getConfig().setUseAsyncServer(false);
      ClusterDescriptor.getInstance().getConfig().setMaxClientPerNodePerMember(2);
      DataClientProvider provider = new DataClientProvider(new Factory());
      SyncDataClient client = null;
      try {
        client = provider.getSyncDataClient(node, 100);
      } catch (TException e) {
        Assert.fail(e.getMessage());
      }
      assertNotNull(client);

      //now try to test multi thread
      ExecutorService service = Executors.newFixedThreadPool(10);
      for (int i = 0; i < 4; i++) {
        service.submit(() -> provider.getSyncDataClient(node, 100));
      }

      //wait time should be great then 5000ms
      Thread.currentThread().sleep(10000);
      int totalNumber = provider.getDataSyncClientPool().getNodeClientNumMap().get(node);
      assertEquals(5, totalNumber);

      for (int i = 0; i < 4; i++) {
        service.submit(() -> provider.getSyncDataClient(node, 100));
      }

      Thread.currentThread().sleep(1000);
      //return one client to pool
      provider.getDataSyncClientPool().putClient(node, client);
      //wait all finish
      Thread.currentThread().sleep(10000);
      totalNumber = provider.getDataSyncClientPool().getNodeClientNumMap().get(node);

      //5 + 4 - 1
      assertEquals(8, totalNumber);

      ClusterDescriptor.getInstance().getConfig().setUseAsyncServer(useAsyncServer);
    } catch (Exception e) {
[...]

PR #2662, opened 2021-02-08
Fix statistics value in SyncClientPool/AsyncClientPool have concurren…

yuqi1129:
…cy problem

## Description
see https://github.com/apache/iotdb/issues/2661

### Content3 ...

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


cluster/src/main/java/org/apache/iotdb/cluster/client/async/AsyncClientPool.java
```diff
@@ -81,19 +82,19 @@ public AsyncClient getClient(Node node, boolean activatedOnly) throws IOExceptio
...
         if (nodeClientNum >= maxConnectionForEachNode) {
-          client = waitForClient(clientStack, clusterNode, nodeClientNum);
+          client = waitForClient(clientDeque, clusterNode);
         } else {
           nodeClientNumMap.put(clusterNode, nodeClientNum + 1);
           client = asyncClientFactory.getAsyncClient(clusterNode, this);
         }
       } else {
-        client = clientStack.pop();
+        client = clientDeque.getFirst();
```
LebronAl, line 97: why not use `pop()`? this client must be removed from this deque temporarily as it will be used in the caller

cluster/src/main/java/org/apache/iotdb/cluster/client/sync/SyncClientPool.java
```diff
@@ -125,25 +137,25 @@ private Client waitForClient(Deque<Client> clientStack, ClusterNode node, int no
...
+          clientDeque.push(syncClientFactory.getSyncClient(node, this));
           NodeStatusManager.getINSTANCE().activate(node);
+          this.notify();
         } catch (TTransportException e) {
           logger.error("Cannot open transport for client {}", node, e);
           nodeClientNumMap.computeIfPresent(clusterNode, (n, oldValue) -> oldValue - 1);
           NodeStatusManager.getINSTANCE().deactivate(node);
         }
       }
-      this.notifyAll();
```
LebronAl, line 146: I can understand why you change `notifyAll` to `notify`, but I don't know why remove this line here to above?

LebronAl (commented):
Hi, thanks for your contribution. It's really a careful finding!
After reading your issue description, I see how this bug happens. In my opinion, the key reason is that we use the outdated nodeClientNum to update the nodeClientNumMap. As nodeClientNumMap is a concurrentMap, we can use it's atomic function to simply fix this bug acually.
I try to change `nodeClientNumMap.put(node, nodeClientNum + 1);` to `nodeClientNumMap.computeIfPresent(node, (n, oldValue) -> oldValue + 1);` in `waitForClient`, and change `nodeClientNumMap.put(node, nodeClientNum);` to `nodeClientNumMap.computeIfPresent(node, (n, oldValue) -> oldValue - 1);` in `putClient`, and then I can pass your newly concurrent test below. So maybe just change these two line will be enough? 
What's your opinion?

cluster/src/main/java/org/apache/iotdb/cluster/client/async/AsyncClientPool.java
yuqi1129: Sorry, this is a mistake i make here, `getFirst()` will not remove element from queue

cluster/src/main/java/org/apache/iotdb/cluster/client/sync/SyncClientPool.java
yuqi1129: > but I don't know why remove this line here to above
My intention is just to replace notifyAll() with notify(), it's a mistake to move to the inner try block

yuqi1129:
@LebronAl Yes, we can just use `nodeClientNumMap.computeIfPresent(node, (n, oldValue) -> oldValue + 1)` to solve the atomic update problem in map.  nice advice~

cluster/src/main/java/org/apache/iotdb/cluster/client/async/AsyncClientPool.java
```diff
@@ -195,12 +198,12 @@ void recreateClient(Node node) {
           clientCaches.computeIfAbsent(clusterNode, n -> new ArrayDeque<>());
       try {
         AsyncClient asyncClient = asyncClientFactory.getAsyncClient(node, this);
-        clientStack.push(asyncClient);
+        clientStack.addLast(asyncClient);
```
LebronAl, line 201: what's the benefit of changing stack to queue?

LebronAl (commented):
Actually we considered a lot in this module.

The reason why we use `while` in `waitForClient` and `notifyAll` to wakeup all waiting thread in `PutClient` is that we want to control the client generating speed to avoid a client leak as well as not to block upper application for more than `WAIT_CLIENT_TIMEOUT_MS `. 

After your change, a bad situation just like `Thread Starvation` will be introduced, which means the upper application may preform a `getClient` and be blocked for more than  `WAIT_CLIENT_TIMEOUT_MS`, this could cause the upper application to trigger the re-election of the leader and so on, and this is not very well. 

So, my advice is just not to change these logic and just use the atomic function in map to solve this concurrent bug.

What's your opinion?

PR #2667, opened 2021-02-09
[IOTDB-1152] optimize regular data size in traversing

wangchao316:
data[] storage data in RegularDataEncoder, data[] default-size is 128,  if real data  < 128,  but traversing 128 times in data2Diff  method, which can use real data size...


wangchao316:
@qiaojialin @jixuan1989 , hi , could you please review this?

tsfile/src/main/java/org/apache/iotdb/tsfile/encoding/encoder/RegularDataEncoder.java
```diff
@@ -220,7 +223,7 @@ private void data2Diff(int[] missingPointData) {
       bitmap = new BitSet(newBlockSize);
       bitmap.flip(0, newBlockSize);
       int offset = 0;
-      for (int i = 1; i < missingPointData.length; i++) {
+      for (int i = 1; i < dataTotal; i++) {
```
HTHou, line 226: Just a small suggestion, would you like adding some the comments here?

HTHou (approved):
LGTM.

PR #2670, opened 2021-02-10
[IOTDB-1151]fix dependency error because some jars are removed from maven's central repo

jixuan1989:
we use jaxb-runtime (v2.4) to solve compatibility problem of JDK 8 and 11, but its dependency, com.sun.xml.fastinfoset:FastInfoset:jar:1.2.14, is removed from maven's central repo. 

image

So, we have to upgrade jaxb-runtime.


jixuan1989:
has cherry-picked.

PR #2674, opened 2021-02-11
Remove `final` if there are no subclasses by design.(Happy New Year)

sunjincheng121:
Remove `final` if there are no subclasses by design, In this PR would like to using `static` take over `static final`.


qiaojialin (approved):
Happy new year!

sunjincheng121:
Happy new year and Thanks for reviewing the PR @qiaojialin !
I will merge this PR and track the check error of Ubuntu 11 in IOTDB-1155.

PR #2680, opened 2021-02-14
upgrade sonar plugin version and testing the permission 

jixuan1989:
This PR upgrades the sonar plugin version
And, it is for checking whether non-apache repo has the permission to upload data to sonarcloud.


jixuan1989:
does not work.

Issue #2688, opened 2021-02-17
LRULinkedHashMap does not work as an LRU Cache


JunChen404:
master:
LRULinkedHashMap.java.
LinkedHashMap is used incorrectly.
To make LinkedHashMap an LRU cache, a true boolean should be provided for the param accessOrder of the constructor.

JunChen00:
I want to do this issue.

PR #2689, opened 2021-02-17
[ISSUE-2688] LRULinkedHashMap does not work as an LRU Cache

chenjun40:
Fixes the issue [#2688](https://github.com/apache/iotdb/issues/2688)
**Describe the bug**

LRULinkedHashMap.java.
LinkedHashMap is used incorrectly.
To make LinkedHashMap an LRU cache, a true boolean should be provided for the param accessOrder of the constructor.


server/src/main/java/org/apache/iotdb/db/engine/cache/LRULinkedHashMap.java
```diff
@@ -55,7 +55,7 @@
   protected long averageSize = 0;
 
   public LRULinkedHashMap(long maxMemory) {
-    this.linkedHashMap = new LinkedHashMap<>(INITIAL_CAPACITY, LOAD_FACTOR_MAP);
```
jixuan1989, line 58: Hi, 
I am not sure whether it is enough by just enabling `accessOrder = true`.

If you read the `put()` method, you will find it removes elements at the head of the set, rather than the tail...

Do we need to use `removeEldestEntry ()` method instead? @JackieTien97

JunChen00: I read the `put()` method, and I think it is ok to just enable `accessOrder = true`.
1、At line 63 of the current file, the removal of elements is realized outside of the LinkedHashMap.
2、Everytimes after an element is accessed, the element is put at the tail, so the head is the eldest one.
3、As the default value returned by `removeEldestEntry()` method is false, the `put()` method of LinkedHashMap does not remove any elements.

JunChen00: I read the `put()` method, and I think it is ok to just enable `accessOrder = true`.
1、At line 63 of the current file, the removal of elements is realized outside of the LinkedHashMap.
2、Everytimes after an element is accessed (including the `put()` method and the `get()` method), the element is put at the tail, so the head is the eldest one.
3、As the default value returned by `removeEldestEntry()` method is false, the `put()` method of LinkedHashMap does not remove any elements.

jixuan1989: 1. but why not use `removeEldestEntry `? 
Can calculating the size in `put` bring any benefit ?

2. About added UT:

The UT does not take effect. Even you do not modify `LRULinkedHashMap.java` and only add your UT codes, the UT can also pass, which means the UT does not cover the incorrection of the original `LRULinkedHashMap.java`.

JunChen00: 1、extend `removeEldestEntry()` method
2、improve UT

server/src/main/java/org/apache/iotdb/db/engine/cache/LRULinkedHashMap.java
```diff
@@ -21,71 +21,56 @@
...
 
-import java.util.Iterator;
 import java.util.LinkedHashMap;
-import java.util.Map.Entry;
-import java.util.Set;
+import java.util.Map;
 
 /** This class is an LRU cache. <b>Note: It's not thread safe.</b> */
-public abstract class LRULinkedHashMap<K extends Accountable, V> {
+public abstract class LRULinkedHashMap<K extends Accountable, V> extends LinkedHashMap<K, V> {
```
JackieTien97, line 28: I think we should keep LinkedHashMap as a field instead of extending LinkedHashMap, because we don't need to expose all the methods of LinkedHashMap to users.

server/src/main/java/org/apache/iotdb/db/engine/cache/LRULinkedHashMap.java
```diff
@@ -21,71 +21,56 @@
...
-
-  public boolean containsKey(K key) {
-    return linkedHashMap.containsKey(key);
+  @Override
+  protected boolean removeEldestEntry(Map.Entry<K, V> eldest) {
+    boolean shouldRemove = usedMemory > maxMemory;
+    if (shouldRemove) {
+      usedMemory -= eldest.getKey().getRamSize();
+    }
+    return shouldRemove;
```
JackieTien97, line 61: The previous logic is when the LRUMap is full, we should evict 10% object, your implementation is inconsistent with that.

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/engine/cache/LRULinkedHashMap.java
JunChen00: Agree & Retract @JackieTien97

JackieTien97 (approved):

chenjun40:
How to merge the commits? @JackieTien97

HTHou:
> How to merge the commits? @JackieTien97

Hi, we are working on the performance test of this PR in order to make sure the improvement before the commits merged. We'll share the result once we get it. Thanks!

HTHou:
Test result:
https://cwiki.apache.org/confluence/display/IOTDB/PR-2689+LRULinkedHashMap+does+not+work+as+an+LRU+Cache

Issue #2666, opened 2021-02-09
The time column of the resultset "select sensor from device  limit n offset idx  "  is not in ascending order


liuzhen1207:
**Describe the bug**
The time column of the resultset "select sensor from device  limit n offset idx  "  is not in ascending order.

**To Reproduce**
1. iotdb master : b6586ff 
2. iotdb benchmark  config.properties:
CLIENT_NUMBER=10
GROUP_NUMBER=2
DEVICE_NUMBER=50
SENSOR_NUMBER=500
IS_SENSOR_TS_ALIGNMENT=true
BATCH_SIZE=100
LOOP=10000
IS_OVERFLOW=true
OVERFLOW_MODE=0
OVERFLOW_RATIO=0.5
OPERATION_PROPORTION=1:0:0:0:0:0:0:0:0:0:0
3.  run benchmark insert operation
4. when step 3 is over,iotdb execute flush
5. iotdb execute query
IoTDB> select s_88 from  root.group_1.d_11 limit 10 offset 35655;
+-----------------------------+----------------------+
|                         Time|root.group_1.d_11.s_88|
+-----------------------------+----------------------+
|2018-09-20T09:44:49.425+08:00|                   364|
|2018-09-20T09:44:51.233+08:00|                  1013|
|2018-09-20T09:44:54.756+08:00|                  1013|
|2018-09-20T08:55:02.274+08:00|                  1090|
|2018-09-20T08:55:09.767+08:00|                   219|
|2018-09-20T08:55:11.651+08:00|                   795|
|2018-09-20T08:55:15.753+08:00|                  1027|
|2018-09-20T08:55:21.270+08:00|                   937|
|2018-09-20T08:55:26.960+08:00|                    44|
|2018-09-20T08:55:31.867+08:00|                  1404|
+-----------------------------+----------------------+
Total line number = 10
It costs 0.223s

The time column is not in ascending order,the 4th row's time is less than the 3rd row's.

PR #2693, opened 2021-02-18
[DRAFT] Fix chunk unseq merge bug

zhanglingzhe0820:
#2666 
The unseq merge now has a bug, which cause sequence pages overlapped.
image

There are two possible causes of the phenomena:
1、unseq merge page write overlap
2、unseq merge chunk write overlap, level merge and then merge chunk again.

The second case is the most likely.


jixuan1989:
how about 0.11?

zhanglingzhe0820:
This pr just 

> how about 0.11?

This PR is only for testing, it is not completely certain that it is the bug here。

PR #2698, opened 2021-02-19
[IOTDB-1163]optimize the insertRecords session interface for cluster version

neuyilan:
https://issues.apache.org/jira/browse/IOTDB-1163


server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
```diff
@@ -1097,6 +1101,27 @@ public void insert(InsertRowsOfOneDevicePlan insertRowsOfOneDevicePlan)
...
+        continue;
+      }
+      try {
+        insert(plan.getInsertRowPlanList().get(i));
+        plan.getResults().put(i, RpcUtils.getStatus(TSStatusCode.SUCCESS_STATUS));
+      } catch (QueryProcessException e) {
+        plan.getResults().put(i, RpcUtils.getStatus(e.getErrorCode(), e.getMessage()));
+        allSuccess = false;
+      }
+      if (!allSuccess) {
```
mychaow, line 1119: do we need to insert the left records while one record is failed? @qiaojialin

mychaow (approved):
Good Job!

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
neuyilan: Thanks for your review, in my opinion, I support inserting the remaining records in case of one record insertion failure. After all, we support partial insertion success. Moreover, this implementation is consistent with the previous semantics of composing each record into an InsertRowPlan.

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -1158,23 +1158,46 @@ public TSStatus insertRecords(TSInsertRecordsReq req) {
...
+        allSuccess, tsStatus, insertRowsPlan.getResults(), req.deviceIds.size());
+  }
+
+  private TSStatus judgeFinalTsStatus(
+      boolean allSuccess,
+      TSStatus executeTsStatus,
+      Map<Integer, TSStatus> checkTsStatus,
+      int totalRowCount) {
+
+    if (allSuccess) {
```
mychaow, line 1188: Changing allSuccess to allCheckSuccess is better.

mychaow (approved):

PR #2702, opened 2021-02-20
[IOTDB-1136] Improved reliability in flush error

wangchao316:
The flushing fails due to OOM or insufficient disk space etc.... As a result, the cluster is in read-only mode.
We need a mechanism to restore read-only mode.
If the fault is caused by a common error
1. If the flush fails, a retry mechanism can be added. By default, the retry mechanism is used for three times at an interval of 1s.
2. After the read-only mode is enabled, the system can recover the read-only mode and pull the data every 10 minutes.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/common/RetryCounter.java
+++ b/server/src/main/java/org/apache/iotdb/db/common/RetryCounter.java
@@ -1,0 +1,129 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iotdb.db.common;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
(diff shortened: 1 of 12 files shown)


server/src/main/java/org/apache/iotdb/db/writelog/manager/ReadOnlyRecoverManager.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/** read only mode recover default: 10 min once, retry 3 times */
+public class ReadOnlyRecoverManager {
+  private static final Logger logger = LoggerFactory.getLogger(ReadOnlyRecoverManager.class);
+
+  private static final int READ_ONLY_RECOVER_DEFAULT_RETRY_ATTEMPTS = 3;
+
+  private static final long READ_ONLY_RECOVER_DEFAULT_RETRY_SLEEP_IINTERVAL = 10 * 60 * 1000L;
```
HTHou, line 40: Hi, I think these two fields could be configured and users might like to change them.

wangchao316: thanks @HTHou ,   I will add configuration in config...

server/src/main/java/org/apache/iotdb/db/common/RetryCounter.java
```diff
@@ -0,0 +1,129 @@
...
+  public static class BackoffPolicy {
+    public long getBackoffTime(RetryConfig config, int attempts) {
+      return addJitter(config.getSleepInterval(), config.getJitter());
+    }
+  }
+
+  public static class ExponentialBackoffPolicy extends BackoffPolicy {
+    @Override
+    public long getBackoffTime(RetryConfig config, int attempts) {
+      long backoffTime = (long) (config.getSleepInterval() * Math.pow(2, attempts));
```
neuyilan, line 125: `long backoffTime = (long) (config.getSleepInterval() * Math.pow(2, attempts));`

The above code line seems not used?

wangchao316: thanks @neuyilan , this code is called in sleepUntilNextRetry - > getBackoffTime  method...

wangchao316:
@qiaojialin @jixuan1989  hi, could you please review this? 
Thanks...

PR #2707, opened 2021-02-21
[ISSUE-2709,IOTDB-1178] Fix cache not cleared after unseq compaction bug, Fix windows 70,10 ci bug in unseq compaction ci

zhanglingzhe0820:
# ISSUE-2709
## Behavior
User's query operation get invalid chunk after compaction.

## Reason
After compaction, we need to delete a file.
However, the chunk cache and timeseriesMetadata cache is also active which may be queried.
So user can get invalid chunk.

## Solution
We will clear the chunkMetadata cache and timeseriesMetadata cache after a compaction is done.

# IOTDB-1178
## Behavior
A windows ci problem about unseq compaction.
The error behaves like below (The 70,10 problem).
image

## Reason
In unseq compaction, when delete tsfile,  if tsfile is reading now,  will throw exception and then cannot delete the old file in windows system.

## Solution
As we have get the file lock, we just have to clear the TsFileSequenceReader cache before delete a tsfile.


server/src/main/java/org/apache/iotdb/db/engine/compaction/level/LevelCompactionTsFileManagement.java
```diff
@@ -155,7 +157,9 @@ private void deleteLevelFilesInList(
   private void deleteLevelFile(TsFileResource seqFile) {
     seqFile.writeLock();
     try {
+      ChunkCache.getInstance().clear();
```
jixuan1989, line 160: clear all Cache?

HTHou:
Don't forget merge master branch....

JackieTien97 (approved):

example/udf/pom.xml
```diff
@@ -73,7 +73,7 @@
                         <importOrder>
                             <order>org.apache.iotdb,,javax,java,\#</order>
                         </importOrder>
-                        <removeUnusedImports/>
```
HTHou, line 76: Don't change this...

pom.xml
```diff
@@ -559,7 +559,6 @@
                             <exclude>**/.git/**</exclude>
                             <exclude>**/.mvn/**</exclude>
                             <exclude>**/.gitignore</exclude>
-                            <exclude>**/.git-blame-ignore-revs</exclude>
```
HTHou, line 562: Same

example/udf/pom.xml
```diff
@@ -73,7 +73,7 @@
...
+                        <removeUnusedImports />
```
HTHou, line 76:
```suggestion
                        <removeUnusedImports/>
```

pom.xml
```diff
@@ -559,7 +559,6 @@
                             <exclude>**/.git/**</exclude>
                             <exclude>**/.mvn/**</exclude>
                             <exclude>**/.gitignore</exclude>
```
HTHou, line 561:
```suggestion
                            <exclude>**/.gitignore</exclude>
                            <exclude>**/.git-blame-ignore-revs</exclude>
```

pom.xml
```diff
@@ -613,7 +612,7 @@
                             <importOrder>
                                 <order>org.apache.iotdb,,javax,java,\#</order>
                             </importOrder>
-                            <removeUnusedImports/>
+                            <removeUnusedImports />
```
HTHou, line 615:
```suggestion
                            <removeUnusedImports/>
```

HTHou:
Need to merge master branch again..

zhanglingzhe0820:
cherry-pick to 0.11 [PR](https://github.com/apache/iotdb/pull/2749)

PR #2738, opened 2021-02-25
[IOTDB-1177] Trigger module: ANTLR, logical operators, physical plans and authorization 

SteveYurongSu:
It is the entrance to the trigger management module, and should mainly contain the following parts of code:
1. Trigger management SQL (ANTLR, logical operators, physical plans and corresponding execution interfaces)
2. User authentication for trigger management

User Manual: 
中文 
https://github.com/apache/iotdb/blob/0f48ba2de83d2b1c0da00bba8485baa1c60abe28/docs/zh/UserGuide/Operation%20Manual/Triggers.md

English: https://github.com/apache/iotdb/blob/0f48ba2de83d2b1c0da00bba8485baa1c60abe28/docs/UserGuide/Operation%20Manual/Triggers.md


server/src/main/java/org/apache/iotdb/db/qp/constant/SQLConstant.java
```diff
@@ -170,7 +170,13 @@ private SQLConstant() {
   public static final int TOK_QUERY_PROCESSLIST = 97;
   public static final int TOK_KILL_QUERY = 98;
 
-  public static final int TOK_CHILD_NODES = 99;
```
SilverNarcissus, line 173: You change it from 99 to 104, may this cause wal inconsistent problem? More specifically if token 99 has written into wal and I update the IoTDB, may this cause a problem?

SilverNarcissus (changes requested):
A fantastic beginning of trigger and I'm looking forward to use this in the future. Only one issue and I think you should add some test for sql analysis. Currently may add some ITs for sql parse and test if it can execute without exception. In later PR you can improve these tests for testing the function of trigger.

HTHou (commented):
The code looks good, but does this module need to add some tests?

server/src/main/java/org/apache/iotdb/db/qp/constant/SQLConstant.java
HTHou: TOK_CHILD_NODES was added into master two days ago. So I don't think it's a big problem.
But it's still better to keep TOK_CHILD_NODES = 99.

SteveYurongSu:
Thanks for reviewing @HTHou @SilverNarcissus . I will add some UTs.

server/src/main/java/org/apache/iotdb/db/qp/constant/SQLConstant.java
SteveYurongSu: Thank you all. I'll keep TOK_CHILD_NODES = 99.

HTHou (approved):
LGTM

HTHou:
Hi, I think it's better to add some explanation about the reason of hiding user guide for now..

SteveYurongSu:
> Hi, I think it's better to add some explanation about the reason of hiding user guide for now..

Added. :)

PR #2739, opened 2021-02-26
[ISSUE-2746] Fix data overlapped bug after unseq compaction

zhanglingzhe0820:
## Behavior
Unseq compaction causes the overlap of seq files which lead to out-of-order query.
The query behaves like below.
image

## Reason
The unseq compaction selector tries to find unseq files which overlap with some unsealed files.

However, the unsealed file's endTime is always Long.MIN_VALUE which will be ignored because [Long.MIN_VALUE, Long.MIN_VALUE) never overlap with other files .

As a result, the unseq compaction can not select an overlapped file if the file is unsealed. In this case, the unseq file's remaining data will be appended to the end of the last sealed seq file which is overlapped with the unsealed file.

## Solution
When selecting the overlapped seq files with an unseq file, if we encounter an unsealed seq file, the unseq file will be excluded from this unseq merge task.


server/src/main/java/org/apache/iotdb/db/engine/merge/selector/MaxFileMergeFileSelector.java
```diff
@@ -228,7 +227,7 @@ private void selectOverlappedSeqFiles(TsFileResource unseqFile) {
         if (seqSelected[i] || !seqFile.getDevices().contains(deviceId)) {
           continue;
         }
-        long seqEndTime = seqFile.getEndTime(deviceId);
+        long seqEndTime = seqFile.isClosed() ? seqFile.getEndTime(deviceId) : Long.MAX_VALUE;
```
qiaojialin, line 230: add a javadoc for TsFileResource.getEndTime()

jixuan1989:
> The unseq merge selector try to filter unseq files which overlapped with some open file. However, the open file's endTime is always Long.MIN_VALUE which will be filtered in pervious logic. So unseq files which overlapped with some open file cannot be excluded.

## Behavior

## Reasom

The unseq merge selector tries to find unseq files which overlap with some unsealed files. 

However, the unsealed file's endTime is always Long.MIN_VALUE which will be ignored because [Long.MIN_VALUE, Long.MIN_VALUE) never overlap with other files . 

As a result,  the unseq files which overlapped with some unsealed file cannot be excluded. (what is this meaning????)

## Solution

.....

server/src/main/java/org/apache/iotdb/db/engine/merge/selector/MaxFileMergeFileSelector.java
wangchao316: hi ,I have a question.
if a seqfile do not closed. and then  is there a conflict?  flushing and merging...

wangchao316:
hi  @zhanglingzhe0820 @jixuan1989 @qiaojialin ,I have a question.
if a seqfile do not closed. and then is there a conflict? flushing and merging...
1. merging finished, need delete old tsfile , if this seqfile do not closed,  can this delete ?
2. if flushing data  when merging...  , maybe have a chunk data do not write tsfile, but this tsfile merged..

server/src/main/java/org/apache/iotdb/db/engine/merge/selector/MaxFileMergeFileSelector.java
zhanglingzhe0820: > hi @zhanglingzhe0820 @jixuan1989 @qiaojialin ,I have a question.
> if a seqfile do not closed. and then is there a conflict? flushing and merging...
> 
> 1. merging finished, need delete old tsfile , if this seqfile do not closed,  can this delete ?
> 2. if flushing data  when merging...  , maybe have a chunk data do not write tsfile, but this tsfile merged..
hi
1. unsealed seq file will not be selected to a merge task so it will not be deleted.
2. a file which is flushing is also an unsealed file.

wangchao316: that is ok...

qiaojialin (approved):

zhanglingzhe0820:
cherry-pick to 0.11 [PR](https://github.com/apache/iotdb/pull/2750)

PR #2744, opened 2021-02-26
remove redundant call of sealCurrentPage().

weizihan0110:
Call sealCurrentPage() method twice:

image
image

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/flush/MemTableFlushTask.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/flush/MemTableFlushTask.java
@@ -237,7 +237,6 @@ public class MemTableFlushTask {
                   (Pair<TVList, MeasurementSchema>) task;
               IChunkWriter seriesWriter = new ChunkWriterImpl(encodingMessage.right);
               writeOneSeries(encodingMessage.left, seriesWriter, encodingMessage.right.getType());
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/apache/iotdb/db/engine/flush/MemTableFlushTask.java
```diff
@@ -237,7 +237,6 @@ public void run() {
                   (Pair<TVList, MeasurementSchema>) task;
               IChunkWriter seriesWriter = new ChunkWriterImpl(encodingMessage.right);
               writeOneSeries(encodingMessage.left, seriesWriter, encodingMessage.right.getType());
-              seriesWriter.sealCurrentPage();
```
wangchao316, line 240: this code can not delete.  because this  method make encode data...
this will lead to encode problem...

server/src/main/java/org/apache/iotdb/db/engine/flush/MemTableFlushTask.java
```diff
@@ -271,7 +270,7 @@ public void run() {
             storageGroup,
             writer.getFile().getName());
         while (true) {
-          Object ioMessage = null;
```
wangchao316, line 274: local var init...this do not have problem...

wangchao316 (changes requested):

PR #2745, opened 2021-03-01
Make the Wal buffer pool parameters to be configurable

JackieTien97:
## Problem
After adding the Wal Buffer Pool, the writing test with unseq data became 50% slower than before.

## Reasons
Before, as the descriptions of the design document(https://cwiki.apache.org/confluence/display/IOTDB/DirectByteBuffer+Pool+for+WAL) the default number of wal buffer for each time partition in each storage group is 4. 2 for seq and another 2 for unseq. However, if we have unseq data, then seq occupy 2 wal buffer, unseq occupy 2 wal buffer, now the wal buffer pool wil be empty. While one tsfile is closing,  it won't put the wal buffer back to the pool until the closing is finished. So the writing operation coming later will be blocked until the closing tsfile put its wal buffer back.

## Solutions
Now, I make the max wal buffer number for each time partition to be configurable in iotdb-engine.properties and it can be modified without need to restart iotdb, just change the config file and execute the load configuration command.
The default number of wal buffer for each time partition is changed from 4 to 6 in this version. Even if new writing operation comes while the tsfile is closing, it won't be blocked, because there are still wal buffer in the pool.

Actually, if the file closing operation is very slow, it will happen that more than one tsfile are closing together, the later writing will still be blocked. In this case, you will see the 'Waiting {} ms for wal direct byte buffer.' logs printed. If you have enough memory, you can change the `max_wal_bytebuffer_num_for_each_partition` to be higher, like 8, 10, 12 or 16.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
+++ b/server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
@@ -18,6 +18,11 @@
  */
 package org.apache.iotdb.db.conf;
 
+import static org.apache.iotdb.tsfile.common.constant.TsFileConstant.PATH_SEPARATOR;
+
+import java.io.File;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import org.apache.iotdb.db.conf.directories.DirectoryManager;
 import org.apache.iotdb.db.engine.compaction.CompactionStrategy;
 import org.apache.iotdb.db.engine.merge.selector.MergeFileStrategy;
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
```diff
@@ -167,6 +167,10 @@
    */
   private int walBufferSize = 16 * 1024 * 1024;
 
+  private int maxWalBytebufferNumForEachPartition = 8;
```
HTHou, line 170: The default value should be 6?

JackieTien97: fix

qiaojialin (approved):

PR #2753, opened 2021-03-01
Extract the common codes to one utils class  & Reduce the content of debug log printing of the InsertRowsPlan

neuyilan:
1. Extract the common codes to one utils class (which is used to judge the final status when whose plan have multi rows)
2. Reduce the content of debug log printing of the InsertRowsPlan(which is too big when the bug log level is enabled, reduce the write throughput. In our test, we use the insertRecords interface to the performance,  the performance was 300,000 points/s before modification and it became 3.4 million points/s after modification).

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
+++ b/server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
@@ -1143,7 +1143,7 @@ public class TSServiceImpl implements TSIService.Iface, ServerContext {
           req.deviceIds.get(0),
           req.getTimestamps().get(0));
     }
-    boolean checkAllSuccess = true;
+    boolean allCheckSuccess = true;
     InsertRowsPlan insertRowsPlan = new InsertRowsPlan();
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -1156,11 +1156,11 @@ public TSStatus insertRecords(TSInsertRecordsReq req) {
         TSStatus status = checkAuthority(plan, req.getSessionId());
         if (status != null) {
           insertRowsPlan.getResults().put(i, status);
-          allSuccess = false;
```
mychaow, line 1159: Even if the check fails, the insertRowsPlan will continue.

neuyilan: it will not be executed, this is done in the insert(InsertRowsPlan plan)  method in PlanExecutor.

mychaow (approved):
LGTM

LebronAl (approved):
LGTM

Issue #2741, opened 2021-02-26
getObject method in JDBC should return an Object


HTHou:
**Describe the bug**
Currently when using `getObject` interface in JDBC ResultSet, it will return a String result, which doesn't match the JDBC standard. 

See https://github.com/apache/iotdb/blob/6eda5e478beca32bfc042d9313420423cb97a2d9/jdbc/src/main/java/org/apache/iotdb/jdbc/AbstractIoTDBJDBCResultSet.java#L436-L447

https://github.com/apache/iotdb/blob/6eda5e478beca32bfc042d9313420423cb97a2d9/jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBJDBCResultSet.java#L99-L105

PR #2757, opened 2021-03-02
[ISSUE-2741] getObject method in JDBC should return an Object

HTHou:
#2741
Currently, the getObject method in JDBC returns a String which does not match the JDBC standard.

In this PR, I re-implement this method by calling `IoTDBRpcDataSet.getObjectByName` and `IoTDBRpcDataSet.getObject` method directly to get the Object result.

### DIFF
--- a/jdbc/src/test/java/org/apache/iotdb/jdbc/IoTDBJDBCResultSetTest.java
+++ b/jdbc/src/test/java/org/apache/iotdb/jdbc/IoTDBJDBCResultSetTest.java
@@ -370,34 +370,34 @@ public class IoTDBJDBCResultSetTest {
 
   private void constructObjectList(List<Object> standardObject) {
     Object[][] input = {
-        {
-          2L, 2.22F, 40000L, null, 2.22F,
-        },
-        {
-          3L, 3.33F, null, null, 3.33F,
-        },
-        {
-          4L, 4.44F, null, null, 4.44F,
-        },
-        {
-          50L, null, 50000L, null, null,
-        },
-        {
-          100L, null, 199L, null, null,
-        },
-        {
-          101L, null, 199L, null, null,
-        },
-        {
-          103L, null, 199L, null, null,
-        },
-        {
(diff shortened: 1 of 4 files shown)


HTHou:
Adding test....

jdbc/src/test/java/org/apache/iotdb/jdbc/IoTDBJDBCResultSetTest.java
```diff
@@ -233,6 +236,12 @@ public void testQuery() throws Exception {
               + "105,11.11,199,33333,11.11,\n"
               + "1000,1000.11,55555,22222,1000.11,\n"; // Note the LIMIT&OFFSET clause takes effect
       Assert.assertEquals(standard, resultStr.toString());
+      List<Object> standardObject = new ArrayList<>();
+      constructObjectList(standardObject);
+      Assert.assertEquals(standardObject.size(), resultObjectList.size());
+      for (int i = 0; i < standardObject.size(); i++) {
+        Assert.assertEquals(standardObject.get(i), resultObjectList.get(i));
+      }
```
Alima777, line 244: Hi, I don't think the testing way is ok because the assertEquals() method just compares their value. If `result.getObject(i)` returns a String, the test result is still right.

Alima777 (changes requested):

jdbc/src/test/java/org/apache/iotdb/jdbc/IoTDBJDBCResultSetTest.java
HTHou: Screen Shot 2021-03-02 at 2 42 29 PM

If I change `result.getObject(i)` to `result.getString(i)`, the test failed....

Alima777: My fault...Sorry about that

Alima777 (approved):
LGTM

Issue #2758, opened 2021-03-02
NullPointerException in QueryTimeManager.checkQueryAlive()


Alima777:
![image](https://user-images.githubusercontent.com/34242296/109595912-429b3480-7b50-11eb-8204-06ab03dbc8ea.png)

It's caused by the operation of multi threads, when the first thread passes the not null check and move to the next check, the second thread unregisters the object, then the next check for the first thread will throw the null pointer exception.

![image](https://user-images.githubusercontent.com/34242296/109595940-4fb82380-7b50-11eb-9453-71d1ebc7b0bb.png)

PR #2759, opened 2021-03-02
[ISSUE-2758] NullPointerException in QueryTimeManager.checkQueryAlive()

Alima777:
The situation is described in #2758.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/query/control/QueryTimeManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/query/control/QueryTimeManager.java
@@ -24,6 +24,7 @@ import org.apache.iotdb.db.conf.IoTDBDescriptor;
 import org.apache.iotdb.db.exception.query.QueryTimeoutRuntimeException;
 import org.apache.iotdb.db.service.IService;
 import org.apache.iotdb.db.service.ServiceType;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/apache/iotdb/db/query/control/QueryTimeManager.java
```diff
@@ -90,20 +90,18 @@ public AtomicBoolean unRegisterQuery(long queryId) {
...
-        });
-    queryScheduledTaskMap.computeIfPresent(
-        queryId,
-        (k, v) -> {
-          queryScheduledTaskMap.get(queryId).cancel(false);
+          ScheduledFuture<?> scheduledFuture = queryScheduledTaskMap.remove(queryId);
+          if (scheduledFuture != null) {
+            scheduledFuture.cancel(false);
+          }
           return null;
```
wangchao316, line 97: maybe return null.

server/src/main/java/org/apache/iotdb/db/query/control/QueryTimeManager.java
```diff
@@ -90,20 +90,18 @@ public AtomicBoolean unRegisterQuery(long queryId) {
...
         });
     return successRemoved;
   }
 
   public static void checkQueryAlive(long queryId) {
-    if (getInstance().queryInfoMap.get(queryId) != null
-        && getInstance().queryInfoMap.get(queryId).isInterrupted()) {
+    QueryInfo queryInfo = getInstance().queryInfoMap.get(queryId);
+    if (queryInfo != null && queryInfo.isInterrupted()) {
       if (getInstance().unRegisterQuery(queryId).get()) {
```
wangchao316, line 105: hi,   getInstance().unRegisterQuery(queryId) maybe return null, so #get() will throw nullpointerexception

Alima777: It may return false but never return null. Am i wrong?

Alima777: It may return false but never return null. Am i wrong?

qiaojialin:
add more description and resubmit

server/src/main/java/org/apache/iotdb/db/query/control/QueryTimeManager.java
wangchao316: sorry, I see fault..   map k-v return null...  
unRegisterQuery return boolean..
you are right...

server/src/main/java/org/apache/iotdb/db/query/control/QueryTimeManager.java
```diff
@@ -90,20 +90,18 @@ public AtomicBoolean unRegisterQuery(long queryId) {
         queryId,
         (k, v) -> {
           successRemoved.set(true);
-          return null;
-        });
-    queryScheduledTaskMap.computeIfPresent(
-        queryId,
-        (k, v) -> {
-          queryScheduledTaskMap.get(queryId).cancel(false);
+          ScheduledFuture<?> scheduledFuture = queryScheduledTaskMap.remove(queryId);
```
wangchao316, line 93: remove queryId from Map, this is solution..
good...

wangchao316 (approved):

PR #2760, opened 2021-03-02
[To rel/0.11] cherry-pick getObject method in JDBC should return an Object

HTHou:
cherry pick #2757


jdbc/src/main/java/org/apache/iotdb/jdbc/AbstractIoTDBJDBCResultSet.java
```diff
@@ -425,7 +425,7 @@ public Object getObject(int columnIndex) throws SQLException {
 
   @Override
   public Object getObject(String columnName) throws SQLException {
-    return getValueByName(columnName);
+    return getObjectByName(columnName);
```
wangchao316, line 428: +1

wangchao316 (approved):

PR #2766, opened 2021-03-02
[IOTDB-1118] [Distributed] Limit query log lag in mid consistency level

wangchao316:
We provide a middle level of read consistency, which is a best-effort strategy, and will try to synchronize with the leader and uses the local data when the synchronization times out. However, the maximum log lag is undefined, which means the node can be arbitrarily far from the leader (missing arbitrarily length of logs) and incurs inconvenience when the user wants to control the data freshness.

To resolve this, a configuration should be provided like `max_read_log_lag`, which only takes effect when the consistency level is set to mid, and queries will fail if the log lag exceeds this configuration after a time out.


wangchao316:
@jixuan1989 @qiaojialin @mychaow @jt2594838 ,hi , could you please review this?
Thanks..

cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterConfig.java
```diff
@@ -147,14 +147,13 @@
    */
   private boolean waitForSlowNode = true;
 
+  /** when consistency level set to mid, query failed if the log lag exceeds maxReadLogLog. */
```
mychaow, line 150:
```suggestion
  /** when consistency level set to mid, query failed if the log lag exceeds maxReadLogLag. */
```

cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterConfig.java
```diff
@@ -147,14 +147,13 @@
...
+  private long maxReadLogLag = 1000L;
+
   public int getSelectorNumOfClientPool() {
     return selectorNumOfClientPool;
   }
 
-  public void setSelectorNumOfClientPool(int selectorNumOfClientPool) {
```
mychaow, line 154: Why remove the method?

cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterConfig.java
```diff
@@ -423,7 +422,11 @@ public boolean isWaitForSlowNode() {
     return waitForSlowNode;
   }
 
-  public void setWaitForSlowNode(boolean waitForSlowNode) {
```
mychaow, line 426: same as above

cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterConfig.java
wangchao316: this method do not use..

cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterConfig.java
wangchao316: this method do not use..

cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterConfig.java
wangchao316: Thanks, I will alter...

cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
```diff
@@ -731,28 +730,70 @@ public TSStatus executeNonQueryPlan(ExecutNonQueryReq request)
...
-          }
+          syncLeader(new StrongCheckConsistency());
           return;
         case MID_CONSISTENCY:
-          // do not care success or not
-          syncLeader();
+          // if leaderCommitId bigger than localAppliedId a value,
+          // will throw CHECK_MID_CONSISTENCY_EXCEPTION
+          syncLeader(
+              new CheckConsistency() {
```
mychaow, line 743: why not provide a midCheckConsistency like strongCheckConsistency?

cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
```diff
@@ -731,28 +730,70 @@ public TSStatus executeNonQueryPlan(ExecutNonQueryReq request)
...
+                @Override
+                public void postCheckConsistency(long leaderCommitId, long localAppliedId)
+                    throws CheckConsistencyException {
+                  if (leaderCommitId == Long.MAX_VALUE
+                      || leaderCommitId == 0
```
mychaow, line 748: 0 or Long.MIN_VALUE?

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
```diff
@@ -1199,7 +1207,11 @@ private boolean tryPullTimeSeriesSchemas(Node node, PullSchemaRequest request) {
       if (partitionGroup.contains(metaGroupMember.getThisNode())) {
         // this node is a member of the group, perform a local query after synchronizing with the
         // leader
-        metaGroupMember.getLocalDataMember(partitionGroup.getHeader()).syncLeader();
+        try {
+          metaGroupMember.getLocalDataMember(partitionGroup.getHeader()).syncLeader(null);
```
mychaow, line 1211: maybe a dummy checkconsistent is better than null.

mychaow (commented):
Good job! There are some minor issues, PTAL.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
wangchao316: strongcheckConsistency have two place using, due to Code Reusability..
midcheckconsistency only use in a place,  do not extracts a class, for Code readability..

cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
wangchao316: > 0 or Long.MIN_VALUE?

should be Long.MIN_VALUE,   commitIindexId is Long.Min_value for default..
I will alter..

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
wangchao316: If the value is null, the program skips it.
If it is a virtual method, the program needs to enter it.
This increases the stack call of the program jvm.

cluster/src/test/java/org/apache/iotdb/cluster/server/member/MemberTest.java
```diff
@@ -343,121 +356,211 @@ public AsyncClient getAsyncClient(Node node) {
...
+        .setConsistencyLevel(ConsistencyLevel.STRONG_CONSISTENCY);
     try {
-      dataGroupMemberWithWriteStrongConsistencyFalse.syncLeaderWithConsistencyCheck(true);
+      dataGroupMemberWithWriteStrongConsistencyFalse.waitUntilCatchUp(
+          new RaftMember.CheckConsistency() {
+            @Override
+            public void postCheckConsistency(long leaderCommitId, long localAppliedId)
+                throws CheckConsistencyException {
+              if (leaderCommitId == Long.MAX_VALUE
+                  || leaderCommitId == 0
```
mychaow, line 411: better to keep consistent with previous.

cluster/src/test/java/org/apache/iotdb/cluster/server/member/MemberTest.java
```diff
@@ -343,121 +356,211 @@ public AsyncClient getAsyncClient(Node node) {
...
+      Mockito.when(partitionedSnapshotLogManager.getMaxHaveAppliedCommitIndex()).thenReturn(-2L);
+      dataGroupMemberWithWriteStrongConsistencyFalse.setLogManager(partitionedSnapshotLogManager);
+
+      dataGroupMemberWithWriteStrongConsistencyFalse.waitUntilCatchUp(
+          new RaftMember.CheckConsistency() {
+            @Override
+            public void postCheckConsistency(long leaderCommitId, long localAppliedId)
+                throws CheckConsistencyException {
+              if (leaderCommitId == Long.MAX_VALUE
+                  || leaderCommitId == 0
```
mychaow, line 477: same

cluster/src/test/java/org/apache/iotdb/cluster/server/member/MemberTest.java
```diff
@@ -343,121 +356,211 @@ public AsyncClient getAsyncClient(Node node) {
...
+      Mockito.when(partitionedSnapshotLogManager.getMaxHaveAppliedCommitIndex()).thenReturn(600L);
+      dataGroupMemberWithWriteStrongConsistencyTrue.setLogManager(partitionedSnapshotLogManager);
+
+      dataGroupMemberWithWriteStrongConsistencyTrue.waitUntilCatchUp(
+          new RaftMember.CheckConsistency() {
+            @Override
+            public void postCheckConsistency(long leaderCommitId, long localAppliedId)
+                throws CheckConsistencyException {
+              if (leaderCommitId == Long.MAX_VALUE
+                  || leaderCommitId == 0
```
mychaow, line 511: same

cluster/src/test/java/org/apache/iotdb/cluster/server/member/MemberTest.java
```diff
@@ -311,15 +319,20 @@ public AsyncDataClient getAsyncDataClient(Node node, int timeout) throws IOExcep
...
 
           @Override
-          public boolean syncLeader() {
+          public boolean syncLeader(RaftMember.CheckConsistency checkConsistency) {
             return syncLeader;
           }
 
+          @Override
+          protected long requestCommitIdAsync() {
+            return 0;
```
mychaow, line 333: return long.min_value?

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
mychaow: ok

wangchao316:
@neuyilan @LebronAl  hi , could you please review?

cluster/src/assembly/resources/conf/iotdb-cluster.properties
```diff
@@ -148,4 +148,8 @@ enable_use_persist_log_on_disk_to_catch_up=true
 
 # The number of logs read on the disk at one time, which is mainly used to control the memory usage.
 # This value multiplied by the log size is about the amount of memory used to read logs from the disk at one time.
-max_number_of_logs_per_fetch_on_disk=1000
\ No newline at end of file
+max_number_of_logs_per_fetch_on_disk=1000
+
+# When consistency level set to mid, query failed if the log lag exceeds max_read_log_lag
+# This default value is 1000
+max_read_log_lag=1000
```
neuyilan, line 155: Perhaps we need to describe in detail what this parameter represents, such as with the `insertRecord()` interface, where a single log represents a row data, whereas if the `insertTablet()` interface is used, where a single log may represent several thousand rows of data (depending on the size of the tablet set by the user), these things need to be known to users to set this parameter reasonably according to their business logic.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
```diff
@@ -806,32 +850,42 @@ public void waitLeader() {
...
     try {
-      leaderCommitId =
-          ClusterDescriptor.getInstance().getConfig().isUseAsyncServer()
-              ? requestCommitIdAsync()
-              : requestCommitIdSync();
-      if (leaderCommitId == Long.MAX_VALUE) {
-        // Long.MAX_VALUE representing there is a network issue
-        return false;
-      }
+      leaderCommitId = config.isUseAsyncServer() ? requestCommitIdAsync() : requestCommitIdSync();
```
neuyilan, line 860: Maybe we should discuss whether we need to use the leaderCommitId in mid consistency, because, the applied index is updated in one separate thread, which means that in a very short time of one node, the applied index must be smaller than the commit index. 

In the process of writing, the applied index of the local node is always smaller than the commit index, not to mention the comparison between the local applied index and the commit index of the leader. This leads to the need to wait for a heartbeat time before judging whether the local applied index is catching up with the leader commit index.

cluster/src/test/java/org/apache/iotdb/cluster/server/member/MemberTest.java
```diff
@@ -343,121 +356,211 @@ public AsyncClient getAsyncClient(Node node) {
...
+            @Override
+            public void postCheckConsistency(long leaderCommitId, long localAppliedId)
+                throws CheckConsistencyException {
+              if (leaderCommitId == Long.MAX_VALUE
+                  || leaderCommitId == Long.MIN_VALUE
+                  || leaderCommitId > localAppliedId) {
+                throw CheckConsistencyException.CHECK_STRONG_CONSISTENCY_EXCEPTION;
+              }
+            }
+          });
```
neuyilan, line 416: Why not just use `StrongCheckConsistency`?
The same as others.

cluster/src/test/java/org/apache/iotdb/cluster/server/member/MemberTest.java
```diff
@@ -109,6 +111,9 @@ public void setUp() throws Exception {
     IoTDBDescriptor.getInstance().getConfig().setEnableWal(false);
     RaftMember.setWaitLeaderTimeMs(10);
 
+    RaftServer.setSyncLeaderMaxWaitMs(100);
+    RaftServer.setHeartBeatIntervalMs(100);
```
neuyilan, line 115: Maybe we need to save the previous settings and then reset them after the test is finished.

cluster/src/test/java/org/apache/iotdb/cluster/server/member/MemberTest.java
```diff
@@ -190,6 +195,9 @@ public void tearDown() throws Exception {
     ClusterDescriptor.getInstance().getConfig().setRaftLogBufferSize(preLogBufferSize);
     ClusterDescriptor.getInstance().getConfig().setUseAsyncApplier(prevUseAsyncApplier);
     IoTDBDescriptor.getInstance().getConfig().setEnableWal(prevEnableWAL);
+
+    RaftServer.setSyncLeaderMaxWaitMs(20 * 1000);
+    RaftServer.setHeartBeatIntervalMs(1000L);
```
neuyilan, line 200: It's better to use the saved configuration. We can't guarantee that the default configuration will not be modified.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
wangchao316: Thanks houliang
When a leader synchronizes logs to more than half of followers, the current log is set to commitid. Therefore, commitId indicates the maximum number of logs synchronized from the leader to followers. We should use this commitid more appropriately.

cluster/src/test/java/org/apache/iotdb/cluster/server/member/MemberTest.java
wangchao316: Thanks, houliang,
done..

cluster/src/test/java/org/apache/iotdb/cluster/server/member/MemberTest.java
wangchao316: Thanks 
done

cluster/src/test/java/org/apache/iotdb/cluster/server/member/MemberTest.java
wangchao316: Thanks 
done

cluster/src/test/java/org/apache/iotdb/cluster/server/member/MemberTest.java
wangchao316: done

cluster/src/test/java/org/apache/iotdb/cluster/server/member/MemberTest.java
wangchao316: done

cluster/src/test/java/org/apache/iotdb/cluster/server/member/MemberTest.java
wangchao316: done

cluster/src/test/java/org/apache/iotdb/cluster/server/member/MemberTest.java
wangchao316: done

cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterConfig.java
wangchao316: done

cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
jt2594838: But I do not think it is necessary to `new` each time the method is called, as they are not changed throughout the process. Maybe you can make them static members of a class.

jt2594838 (approved):

neuyilan (approved):
LGTM

cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
LebronAl: +1 for jt2594838

LebronAl (approved):
LGTM
Only minor issues

cluster/src/assembly/resources/conf/iotdb-cluster.properties
```diff
@@ -153,4 +153,8 @@ enable_use_persist_log_on_disk_to_catch_up=true
 
 # The number of logs read on the disk at one time, which is mainly used to control the memory usage.
 # This value multiplied by the log size is about the amount of memory used to read logs from the disk at one time.
-max_number_of_logs_per_fetch_on_disk=1000
\ No newline at end of file
+max_number_of_logs_per_fetch_on_disk=1000
+
+# When consistency level set to mid, query failed if the log lag exceeds max_read_log_lag
```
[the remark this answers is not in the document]
wangchao316: done

cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterConfig.java
```diff
@@ -147,16 +147,15 @@
    */
   private boolean waitForSlowNode = true;
 
+  /** when consistency level set to mid, query failed if the log lag exceeds maxReadLogLag. */
```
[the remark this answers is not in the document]
wangchao316: done

mychaow (approved):
LGTM

PR #2768, opened 2021-03-03
[IOTDB-1181] Upgrade jetty jar to fix CVE-2020-27216

wangchao316:
current  jetty version is 9.4.24.v20191120,  which have a high loopholes CVE-2020-27216， we can use 9.4.35.v20201120 version


wangchao316:
@HTHou @jixuan1989 @qiaojialin  hi,  could you review this?
Thanks...

jixuan1989 (approved):

jixuan1989:
now we have many dependencies that having problem of CVEs. 
It is time to upgrade them but we must guarantee the stability and the peroformance not decreasing.

PR #2771, opened 2021-03-04
Change the default max JVM heap size from 8GB to 64GB

HTHou:
## Description
This PR is changing the default Maximum JVM heap size setting from 8GB to 64GB.

### DIFF
--- a/server/src/assembly/resources/conf/iotdb-env.sh
+++ b/server/src/assembly/resources/conf/iotdb-env.sh
@@ -66,9 +66,9 @@ calculate_heap_sizes()
     fi
 
     # set max heap size based on the following
-    # max(min(1/2 ram, 1024MB), min(1/4 ram, 8GB))
+    # max(min(1/2 ram, 1024MB), min(1/4 ram, 64GB))
     # calculate 1/2 ram and cap to 1024MB
-    # calculate 1/4 ram and cap to 8192MB
+    # calculate 1/4 ram and cap to 65536MB
     # pick the max
     half_system_memory_in_mb=`expr $system_memory_in_mb / 2`
     quarter_system_memory_in_mb=`expr $half_system_memory_in_mb / 2`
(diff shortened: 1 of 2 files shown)


server/src/assembly/resources/conf/iotdb-env.sh
```diff
@@ -66,19 +66,19 @@ calculate_heap_sizes()
     fi
 
     # set max heap size based on the following
-    # max(min(1/2 ram, 1024MB), min(1/4 ram, 8GB))
+    # max(min(1/2 ram, 1024MB), min(1/4 ram, 64GB))
     # calculate 1/2 ram and cap to 1024MB
```
samperson1997, line 70: Should this be changed?
```suggestion
    # calculate 1/2 ram and cap to 8192MB
```

server/src/assembly/resources/conf/iotdb-env.sh
```diff
@@ -66,19 +66,19 @@ calculate_heap_sizes()
...
-    # calculate 1/4 ram and cap to 8192MB
+    # calculate 1/4 ram and cap to 65536MB
     # pick the max
     half_system_memory_in_mb=`expr $system_memory_in_mb / 2`
     quarter_system_memory_in_mb=`expr $half_system_memory_in_mb / 2`
     if [ "$half_system_memory_in_mb" -gt "1024" ]
     then
         half_system_memory_in_mb="1024"
     fi
-    if [ "$quarter_system_memory_in_mb" -gt "8192" ]
```
wangchao316, line 79: hi, haonan
We can extract a parameter and modify it to control the memory size.
This facilitates user modification.

HTHou: https://github.com/apache/iotdb/blob/dc67d245740fd13b46fca4d95a33419b979405a2/server/src/assembly/resources/conf/iotdb-env.sh#L175-L80

We already have it. I users like to change it, just uncomment this line.

wangchao316: ok， good..

wangchao316 (approved):

PR #2776, opened 2021-03-05
[IOTDB-1190]Fully support HTTP URL char set in IoTDB path

wshao08:
## Description

JIRA Issue: https://issues.apache.org/jira/browse/IOTDB-1190

A couple of special characters are not allowed in timeseries path, e.g. '=', '!', '~'. 
Screen Shot 2021-03-05 at 5 32 30 PM

Sometimes the timeseries paths are designed to contain these characters. At least we need to support all HTTP Url characters.


jixuan1989:
JIRA/ISSUE

weizihan0110:
Hi, @wshao08. I think the dot character in double quotations(like root.a.b."com.edu.cn") should also be considered. Plus, some UTs will be needed.

jixuan1989:
Will merge this PR now.
But the feature will be rewrite in the next version (using antlr).

PR #2781, opened 2021-03-08
[IOTDB-1188][To rel/0.11]Fix IoTDB 0.11 unable to delete data bug

wshao08:
## Description
Check JIRA issue: https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-1188?filter=allopenissues
This bug has been fixed in master branch. So only 0.11 release versions have this bug.

## Reason 
The bug occurs when the delete path is not a full timeseries path but a device path.
If any device path contains no data, delete code will falsely return. In this way, the devices that truly contain data will not be deleted.


HTHou:
Add some test?

PR #2782, opened 2021-03-09
[IOTDB-1189] [Distributed]optimize thread pool client close

wangchao316:
Currently, the thread pool is used to manage inter-cluster communication access. The connection is obtained from the connection pool. After the connection is used, the connection is returned to the connection pool. After the connection is used, the interface must be called to put the connection back to the connection pool. This requires that the connection must be closed in finally.
Therefore, the connection usage is optimized here. Use try() to obtain the closed connection（push to thread pool）.

1. interface Closeable and close method. if call close , will put client to pool.
2. use try() syntax,  will put client to pool.This is the common method of closing connections for thread pools. For example, common-pool.jar.
3. In case the connection is not put back into the connection pool after the connection is completed.

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/client/sync/SyncClientPool.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/client/sync/SyncClientPool.java
@@ -107,7 +107,7 @@ public class SyncClientPool {
           logger.warn(
               "Cannot get an available client after {}ms, create a new one",
               WAIT_CLIENT_TIMEOUT_MS);
-          nodeClientNumMap.put(node, nodeClientNum + 1);
+          nodeClientNumMap.computeIfPresent(node, (n, oldValue) -> oldValue + 1);
           return createClient(node, nodeClientNum);
         }
       } catch (InterruptedException e) {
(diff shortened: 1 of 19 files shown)


LebronAl (approved):
LGTM
BTW, would you please fix the concurrent bug for clientpool mentioned in [2662](https://github.com/apache/iotdb/pull/2662) in this pr?

wangchao316:
> LGTM
> BTW, would you please fix the concurrent bug for clientpool mentioned in [2662](https://github.com/apache/iotdb/pull/2662) in this pr?

ok, hi , I alter 2662....

wangchao316:
> > I try to change nodeClientNumMap.put(node, nodeClientNum + 1); to nodeClientNumMap.computeIfPresent(node, (n, oldValue) -> oldValue + 1); in waitForClient, and change nodeClientNumMap.put(node, nodeClientNum); to nodeClientNumMap.computeIfPresent(node, (n, oldValue) -> oldValue - 1); in putClient, and then I can pass your newly-added concurrent test below. So maybe just change these two line will be enough?
> 
> It seems we should change these two places at least, and after these changes we may remove the unused parameter 'NodeClientNum' in some functions, last but not least , maybe we can copy the UT there too.

hi all,  I submit a new PR  for 2662, this PR need merged first...

neuyilan (approved):
LGTM

PR #2785, opened 2021-03-09
[IOTDB-1187] Fix unseq compaction loss data bug after delete operation

zhanglingzhe0820:
## Behavior
(1) Use the folowing code to write data into 0.11.2:
(2) delete from root.sg_1
(3) Then rerun the following code
(4) select count( * ) from root
The count is less than 1000000.

``` java
static List<Record> mockData(int loop) {
    List<Record> recordList = new ArrayList<>();
    int mockSize=10000;
    long timestamp = mockSize*loop;
    for (int r = 0; r < mockSize; r++) {
        Record record = new Record();
        String storageGroup = "root.sg_1";
        String deviceId = String.format("%s.device_1", storageGroup);
        record.setStorageGroup(storageGroup);
        record.setDeviceId(deviceId);
        record.setTimestamp(new Timestamp(timestamp+r));
        Record.Column column = new Record.Column("pressure", "Float", r%999.0f);
        record.addColumn(column);
        column = new Record.Column("temperature", "Float", r%37.0f);
        record.addColumn(column);
        recordList.add(record);
    }
    return recordList;
}

public static void main(String[] args) throws IoTDBConnectionException, StatementExecutionException {

    IOTDBReader iotdbReader = new IOTDBReader();
    Session session = iotdbReader.openSession();
    for (int loop = 0; loop < 100; loop++) {
        List<Record> recordList = mockData(loop);
        iotdbReader.insertTablet(session, recordList, loop);
        System.out.println(loop);
    }
    session.close();
}
```

## Reason

The unseq compaction does not update the end time of the involved TsFileResource, even if the end time after unseq compaction is smaller than the end time before.
This case can be caused by deletion operation as the deleted data will be removed in a compaction.

Suppose a deletion (delete all data) is marked on TsFile 1 (time range: 0-100) and TsFile 2 (time range: 101-200).
On the 2nd write loop, all data will be written as unseq files, e.g., generating two unseq files, unseq file 3 (time range: 0-50), and unseq file 4 (time range: 51-150).

The merge process may run like this:
1. unseq file 3 is overlapped with tsfile 1. After merge, the time range of the ChunkMetadata in the new TsFile 1 is 0-50, while the time range in the tsfile resource is still 0-100.
2. Unseq file 4 is overlapped with tsfile 1 and 2 (check condition: according to the tsfile resource, rather than the chunkmetadatas, for performance considerison). First, we check the chunk metadata in (new) Tsfile 1, and find the last chunk metadata is 0-50, which is less than 51. Then, we check TsFile2, and merge the data into tsfile2. 
As a result, the time range in the new TsFile 1 is 0-50, while its tsfile resource reports the range as 0-100;
the time range in the new tsfile 2 is 51-150, while its tsfile resource reports the range as 101-200.

As the query process will select files according to the tsfile resource, the incorrect tsfile resource will cause some bug:
e.g., select * .... where time < 100, then all data in TsFile 2 looks lost.

## Solution
1. Maintain the correct start time and the end time of each involved devices in the merge process.
2. After finish writing a tsfile, update the start time and the end time in its tsfile resource: 
 - use putEndTime instead of updateEndTime
 - use putStartTime instead of updateStartTime


server/src/main/java/org/apache/iotdb/db/engine/merge/task/MergeFileTask.java
```diff
@@ -208,15 +208,27 @@ private void moveMergedToOld(TsFileResource seqFile) throws IOException {
...
+          if (chunkMetadata.getStartTime() < minStartTime) {
+            minStartTime = chunkMetadata.getStartTime();
+          }
+          if (chunkMetadata.getEndTime() > maxEndTime) {
+            maxEndTime = chunkMetadata.getEndTime();
+          }
+        }
       }
+      seqFile.putStartTime(device, minStartTime);
+      seqFile.putEndTime(device, maxEndTime);
```
jt2594838, line 231: I do not think the range of the new file can be directly put into the resource.
Consider a compaction with a seq file containing three chunks, [0, 7] [10, 15] [20, 29], and an unseq file with one chunk, [17,28]. The new file (merged file) will contain only a chunk, [17, 29] if a full merge is not required, and the chunk will be appended to the old seq file as merged chunks are the minority. If you just use the time range from the new file, which is [17, 29], to put into the new TsFileResource,  the result will not be accurate.

jt2594838 (changes requested):
Good, but I think you are work another way around instead of resolving the core problem, which I think would better be fixed.
By the way, please add corresponding tests.

server/src/main/java/org/apache/iotdb/db/engine/merge/task/MergeFileTask.java
```diff
@@ -192,8 +192,8 @@ private void moveMergedToOld(TsFileResource seqFile) throws IOException {
             return;
           }
         }
+        updateStartTimeAndEndTime(seqFile, newFileReader);
       }
-      updateStartTimeAndEndTime(seqFile, oldFileWriter);
```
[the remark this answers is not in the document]
zhanglingzhe0820: I am also curios about it as the oldFileWriter cannot get the metadata.

server/src/main/java/org/apache/iotdb/db/engine/merge/task/MergeFileTask.java
zhanglingzhe0820: okay

server/src/main/java/org/apache/iotdb/db/engine/merge/manage/MergeResource.java
```diff
@@ -263,4 +266,28 @@ public void setMeasurementSchemaMap(Map<PartialPath, MeasurementSchema> measurem
...
+
+  public void updateStartTime(TsFileResource tsFileResource, String device, long startTime) {
+    Map<String, Pair<Long, Long>> deviceStartEndTimePairMap =
+        startEndTimeCache.getOrDefault(tsFileResource, new HashMap<>());
+    Pair<Long, Long> startEndTimePair =
+        deviceStartEndTimePairMap.getOrDefault(device, new Pair<>(Long.MAX_VALUE, Long.MIN_VALUE));
+    long newStartTime = startEndTimePair.left > startTime ? startTime : startEndTimePair.left;
+    deviceStartEndTimePairMap.put(device, new Pair<>(newStartTime, startEndTimePair.right));
+    startEndTimeCache.put(tsFileResource, deviceStartEndTimePairMap);
+  }
```
jt2594838, line 278: Use `computeIfAbsent` then you may avoid calling extra `put` and you can directly modify the time pairs.

jt2594838 (approved):

PR #2787, opened 2021-03-09
[IOTDB-1193] Remove redundant sync meta leader in CManager for cluster module

LebronAl:
see [1193](https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-1193?filter=allopenissues)

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
@@ -153,6 +153,11 @@ public class CMManager extends MManager {
     return CMManager.MManagerHolder.INSTANCE;
   }
 
+  /**
+   * sync meta leader to get the newest partition table and storage groups.
+   *
+   * @throws MetadataException throws MetadataException if necessary
+   */
   public void syncMetaLeader() throws MetadataException {
     try {
       metaGroupMember.syncLeaderWithConsistencyCheck(false);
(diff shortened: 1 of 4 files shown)


cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
```diff
@@ -1335,12 +1332,6 @@ private boolean tryPullTimeSeriesSchemas(Node node, PullSchemaRequest request) {
    * @return all paths after removing wildcards in the path
    */
   public List<PartialPath> getMatchedPaths(PartialPath originPath) throws MetadataException {
-    // make sure this node knows all storage groups
-    try {
-      metaGroupMember.syncLeaderWithConsistencyCheck(false);
-    } catch (CheckConsistencyException e) {
-      throw new MetadataException(e);
-    }
```
neuyilan, line 1343: Hi, @LebronAl I see there are so many places use this method, how to make sure that this modification has no effect on the other logic that calls this function?

The same as other methods.

LebronAl: Manually review their call stacks one by one to ensure that a query request only performs synchronization once.

Maybe we need to analyse the current synchronization of CMManager and fix it further.

wangchao316 (approved):
LGTM

mychaow (approved):
LGTM

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
neuyilan: Ok

PR #2794, opened 2021-03-10
Simplify the Avg formula on removing redundant operation

yifuzhou:
## Description
I did serval improvements of the AVG function:
1. The formula to calculate AVG function is complex, it can be simplified. I removed reduncdant operations.
2. When we calculate the AVG, the result value will transfer into Double type. in Java, the calculation of double type will always cause the loss of accuracy. And in our test case, there are lots of values which are loss of accuracy. I recommend to use BigDecimal to solve it which is a popular solution.
3.  If we use BigDecimal, we have to set  the scale and roundingMode, otherwise once the value is an infinite loop, it would got the error. I set scale is 5 and roundingMode is HALF_UP and I think it is enough, we can discuss here. 
Thanks


LebronAl (changes requested):
Acutally, performance is better than aggregation result's accuracy in our user's recent feedbacks. So we had better create a new datatype if there are really some user cases caring about the accuracy very much in the future.

BTW, I do see something inappropriate with the current code in this PR. Maybe you can fix it in this PR, there are two main problems:
1. some redundant operation such as `((double) statistics.getCount() / cnt) * sum / statistics.getCount();`
2. the test should be performed as a delta way, such as if `|a - 11.93333| < 0.000001` then test passed or test failed.

What's your opinion?

yifuzhou:
> Acutally, performance is better than aggregation result's accuracy in our user's recent feedbacks. So we had better create a new datatype if there are really some user cases caring about the accuracy very much in the future.
> 
> BTW, I do see something inappropriate with the current code in this PR. Maybe you can fix it in this PR, there are two main problems:
> 
> 1. some redundant operation such as `((double) statistics.getCount() / cnt) * sum / statistics.getCount();`
> 2. the test should be performed as a delta way, such as if `|a - 11.93333| < 0.000001` then test passed or test failed.
> 
> What's your opinion?

Hello LebronAl,
Thanks for your reply, I am confused about what's your meaning of "performance is better than aggregation result's accuracy"?
If we do not use Bigdecimal way, it is a good idea only to solve this issue by simplifing and using delta test value. However, why don't we use Bigdecimal to calculate the avg which also can guarantee the accuracy instead of creating a new datatype?

LebronAl:
> > Acutally, performance is better than aggregation result's accuracy in our user's recent feedbacks. So we had better create a new datatype if there are really some user cases caring about the accuracy very much in the future.
> > BTW, I do see something inappropriate with the current code in this PR. Maybe you can fix it in this PR, there are two main problems:
> > 
> > 1. some redundant operation such as `((double) statistics.getCount() / cnt) * sum / statistics.getCount();`
> > 2. the test should be performed as a delta way, such as if `|a - 11.93333| < 0.000001` then test passed or test failed.
> > 
> > What's your opinion?
> 
> Hello @LebronAl,
> Thanks for your reply, I am confused about what's your meaning of "performance is better than aggregation result's accuracy"?
> If we do not use Bigdecimal way, it is a good idea only to solve this issue by simplifing and using delta test value. However, why don't we use Bigdecimal to calculate the avg which also can guarantee the accuracy instead of creating a new datatype?

Sorry, I missed a key sentence in last reviews. 
> BigDecimal maybe a significant performance penalty.

As you can benchmark the performance on your own or search out some [results](https://www.iteye.com/problems/63215),
There is no need to compromise performance for accuracy issues that users do not respond to.

Alima777:
Hi. First, I want to apologize for the bad code style in test.

Using this method `assertEquals(double expected, double actual, double delta)` to test inaccurate result will be better.

yifuzhou:
> Hi. First, I want to apologize for the bad code style in test.
> 
> Using this method `assertEquals(double expected, double actual, double delta)` to test inaccurate result will be better.

Thank you for your advice. I will do it in the next pr.

PR #2797, opened 2021-03-10
[IOTDB-1204] set parameter in iotdb-cluster.properties

wangchao316:
The WAIT_CLIENT_TIMEOUT_MS and maxConnectionForEachNode
parameter controls the connection generation speed and affects cluster link access. Therefore, it can be configured in the iotdb-cluster file.

### DIFF
--- a/cluster/src/assembly/resources/conf/iotdb-cluster.properties
+++ b/cluster/src/assembly/resources/conf/iotdb-cluster.properties
@@ -162,5 +162,7 @@ max_read_log_lag=1000
 # Max number of clients in a ClientPool of a member for one node.
 max_client_pernode_permember_number=1000
 
-# If the number of connections created for a node exceeds  `max_client_pernode_permember_number`, we need to wait so much time for other connections to be released until timeout, or a new connection will be created.
+# If the number of connections created for a node exceeds  `max_client_pernode_permember_number`,
+# we need to wait so much time for other connections to be released until timeout,
+# or a new connection will be created.
 wait_client_timeout_ms=5000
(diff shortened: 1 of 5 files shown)


jixuan1989 (approved):

cluster/src/assembly/resources/conf/iotdb-cluster.properties
```diff
@@ -157,4 +157,11 @@ max_number_of_logs_per_fetch_on_disk=1000
...
 # This default value is 1000
-max_read_log_lag=1000
\ No newline at end of file
+max_read_log_lag=1000
+
+# Max number of clients in a ClientPool of a member for one node.
+max_client_pernode_permember_number=1000
+
+# Wait for some timeout, if the old client is not released
+# A new client is created
```
neuyilan, line 166: I think the annotation could be more clearer if it can be configurated by the user.
```suggestion
# If the number of connections created for a node exceeds  `max_client_pernode_permember_number`, you need to wait so much time for other connections to be released until timeout, or a new connection will be created.
```

cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterConfig.java
```diff
@@ -71,6 +71,9 @@
   /** max number of clients in a ClientPool of a member for one node. */
   private int maxClientPerNodePerMember = 1000;
 
+  /** Wait for some timeout, if the old client is not released A new client is created */
```
neuyilan, line 74: The same as above

neuyilan (approved):
LGTM.
Only minor issues which could be better.

cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterConfig.java
wangchao316: done

LebronAl (approved):
LGTM

PR #2798, opened 2021-03-10
[IOTDB-1192] Fix sql end with semicolon 

HTHou:
When execute sql statement like `select * from root;` using JDBC, an exception thrown.

```
Exception in thread "main" java.sql.BatchUpdateException: Error occurred while parsing SQL to physical plan: line 1:47 token recognition error at: ';'

        at org.apache.iotdb.jdbc.IoTDBStatement.executeBatchSQL(IoTDBStatement.java:293)

        at org.apache.iotdb.jdbc.IoTDBStatement.executeBatch(IoTDBStatement.java:255)
```


antlr/src/main/antlr4/org/apache/iotdb/db/qp/sql/SqlBase.g4
```diff
@@ -20,7 +20,7 @@
 grammar SqlBase;
 
 singleStatement
-    : statement EOF
+    : statement (';')? EOF
```
mychaow, line 23: need to update docs?

HTHou: Actually, it is the doc shows we can support the sql statement end with semicolon...

See
https://github.com/apache/iotdb/blob/master/docs/UserGuide/Client/Programming%20-%20JDBC.md#examples

jixuan1989 (approved):

PR #2800, opened 2021-03-10
[IOTDB-1194]Fix a bug of checking time partition in DeviceTimeIndex

fanhualta:
# Behavior

The link of the corresponding Jira issue is: https://issues.apache.org/jira/browse/IOTDB-1194 

# Reason

The reason is `startTimes` and `endTimes` array may be larger than the number of devices in the loaded tsfile, and they are assigned default values for those positions which are larger than `the number of devices`.

# Solution

So, either traversing the `startTimes`   until  `the number of devices`, or traversing the `deviceToIndex` and using its index to access `startTimes`.

### DIFF
--- a/server/src/test/java/org/apache/iotdb/db/integration/IoTDBLoadExternalTsfileWithTimePartitionIT.java
+++ b/server/src/test/java/org/apache/iotdb/db/integration/IoTDBLoadExternalTsfileWithTimePartitionIT.java
@@ -18,13 +18,6 @@
  */
 package org.apache.iotdb.db.integration;
 
-import java.io.File;
-import java.io.IOException;
-import java.sql.Connection;
-import java.sql.DriverManager;
-import java.sql.SQLException;
-import java.sql.Statement;
-import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


Ring-k (approved):

mychaow (approved):
LGTM

neuyilan (approved):
Good Work~

jixuan1989:
could you explain why it is correct by changing the traversing order  in the device order?

jixuan1989:
An IT needed

Ring-k (changes requested):
Hi, I added IT for this pull request. It seems the tsfile could be added to the dir in IoTDB using load statement, but the time partition didn't work. There are only 1 directory in a virtual partition directory.

Ring-k (approved):

Ring-k:
The failure is casued by time unit transformation in IT. The unit of the parameter in config.setPartitionInterval() is second, while I treated it as milisecond before.

samperson1997 (approved):
Hi @Ring-k , thanks for your detailed explanation and careful test : )

PR #2801, opened 2021-03-10
[To 0.11] fix unseq compaction loss data bug after delete operation

zhanglingzhe0820:
## Behavior
(1) Use the folowing code to write data into 0.11.2:
(2) delete from root.sg_1
(3) Then rerun the following code
(4) select count( * ) from root
The count is less than 1000000.

``` java
static List<Record> mockData(int loop) {
    List<Record> recordList = new ArrayList<>();
    int mockSize=10000;
    long timestamp = mockSize*loop;
    for (int r = 0; r < mockSize; r++) {
        Record record = new Record();
        String storageGroup = "root.sg_1";
        String deviceId = String.format("%s.device_1", storageGroup);
        record.setStorageGroup(storageGroup);
        record.setDeviceId(deviceId);
        record.setTimestamp(new Timestamp(timestamp+r));
        Record.Column column = new Record.Column("pressure", "Float", r%999.0f);
        record.addColumn(column);
        column = new Record.Column("temperature", "Float", r%37.0f);
        record.addColumn(column);
        recordList.add(record);
    }
    return recordList;
}

public static void main(String[] args) throws IoTDBConnectionException, StatementExecutionException {

    IOTDBReader iotdbReader = new IOTDBReader();
    Session session = iotdbReader.openSession();
    for (int loop = 0; loop < 100; loop++) {
        List<Record> recordList = mockData(loop);
        iotdbReader.insertTablet(session, recordList, loop);
        System.out.println(loop);
    }
    session.close();
}
```

Detail analysis is in PR #2785


server/src/main/java/org/apache/iotdb/db/engine/merge/manage/MergeResource.java
```diff
@@ -265,4 +266,27 @@ public void clearChunkWriterCache() {
...
 
+  public void updateStartTime(TsFileResource tsFileResource, String device, long startTime) {
+    Map<String, Pair<Long, Long>> deviceStartEndTimePairMap =
+        startEndTimeCache.getOrDefault(tsFileResource, new HashMap<>());
+    Pair<Long, Long> startEndTimePair =
+        deviceStartEndTimePairMap.getOrDefault(device, new Pair<>(Long.MAX_VALUE, Long.MIN_VALUE));
+    long newStartTime = startEndTimePair.left > startTime ? startTime : startEndTimePair.left;
+    deviceStartEndTimePairMap.put(device, new Pair<>(newStartTime, startEndTimePair.right));
+    startEndTimeCache.put(tsFileResource, deviceStartEndTimePairMap);
+  }
```
jt2594838, line 277: Try using `Map.computeIfAbsent` to avoid redundant  `put` and `new`.

jt2594838 (approved):

server/src/main/java/org/apache/iotdb/db/engine/merge/manage/MergeResource.java
zhanglingzhe0820: okay

jt2594838 (approved):

PR #2809, opened 2021-03-11
[IOTDB-1207]Open time partition causes stackoverflow

yanhongwangg:
https://issues.apache.org/jira/browse/IOTDB-1207

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
@@ -321,17 +321,11 @@ public class CMManager extends MManager {
   @Override
   public MNode getSeriesSchemasAndReadLockDevice(InsertPlan plan) throws MetadataException {
     MeasurementMNode[] measurementMNodes = new MeasurementMNode[plan.getMeasurements().length];
-    try {
-      MNode deviceNode = getDeviceNode(plan.getDeviceId());
-
-      int nonExistSchemaIndex =
-          getMNodesLocally(plan.getDeviceId(), plan.getMeasurements(), measurementMNodes);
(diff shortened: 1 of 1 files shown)


mychaow (approved):
LGTM

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
```diff
@@ -321,17 +321,11 @@ public TimeValuePair getLastCache(PartialPath seriesPath) {
...
-        plan.setMeasurementMNodes(measurementMNodes);
-        return deviceNode;
-      }
-    } catch (PathNotExistException exception) {
-      // ignore, so we could try local MTree
+    int nonExistSchemaIndex =
+        getMNodesLocally(plan.getDeviceId(), plan.getMeasurements(), measurementMNodes);
+    if (nonExistSchemaIndex == -1) {
+      plan.setMeasurementMNodes(measurementMNodes);
+      return new MNode(null, plan.getDeviceId().getDevice());
```
wangchao316, line 328: new MNode, which could add to cache?   Does the memory overhead increase if new is used each time?

neuyilan (approved):
LGTM

LebronAl (approved):
LGTM.
Looking forward to your solution to the deadlock problem described in this issue in next PR.

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
LebronAl: In cluster module, the return value of this function is not cached and is only used to detect whether an exception will be thrown.
In addition, an MNode instance is not very memory intensive, and the variables are null by default, so the overhead in memory should not be very large

PR #2811, opened 2021-03-11
Fix compaction with cluster snapshot deletion

zhanglingzhe0820:
## Behavior
Compaction run with cluster snapshot together and meet error.

## Reason
If run as below:
(1) Start compaction task;
(2) Cluster pull snapshot;
(3) The snapshot pulled is fully overlapped with some merging local file and delete them.
(4) The compaction cannot find the file to be read and exit.
The temp file will not be deleted.

## Solution
1、Add restore logic when meet exception.


neuyilan (commented):
LGTM

server/src/main/java/org/apache/iotdb/db/engine/compaction/utils/CompactionUtils.java
```diff
@@ -271,6 +271,9 @@ public static void merge(
       for (TsFileResource tsFileResource : tsFileResources) {
         TsFileSequenceReader reader =
             buildReaderFromTsFileResource(tsFileResource, tsFileSequenceReaderMap, storageGroup);
+        if (reader == null) {
+          throw new IOException();
```
jt2594838, line 275: Add some message in this exception.

jt2594838 (approved):

jt2594838:
There is a failed test, please look into it.

zhanglingzhe0820:
> There is a failed test, please look into it.

This test has nothing to do with this pr.

server/src/main/java/org/apache/iotdb/db/engine/compaction/utils/CompactionUtils.java
zhanglingzhe0820: okay

PR #2817, opened 2021-03-12
[IOTDB-1221] Compaction module: chunk metadata lists returned by the getMeasurementChunkMetadataListMapIterator method are not lexicographically ordered by the measurement names

SteveYurongSu:
Chunk metadata lists returned by the `getMeasurementChunkMetadataListMapIterator` method are not lexicographically ordered by the measurement names, which will cause compaction errors.

JIRA: https://issues.apache.org/jira/browse/IOTDB-1221


LebronAl (approved):
LGTM

PR #2822, opened 2021-03-12
optimize WAL has three ways to be flashed to disk

jxlgzwh:
optimize code  and fix doc


server/src/test/java/org/apache/iotdb/db/writelog/io/LogWriterReaderTest.java
```diff
@@ -26,16 +26,13 @@
...
-
 import org.junit.Before;
 import org.junit.Test;
-
 import java.io.File;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.List;
-
```
wangchao316, line 38: we use command "mvn spotless:apply",  which can alter code fomart.

wangchao316 (changes requested):

jixuan1989 (changes requested):
run `mvn spotless:apply` and then submit codes

jixuan1989 (approved):
Many thanks!

Issue #2827, opened 2021-03-13
Add a new method "addChild" in MNode


weizihan0110:
This method will not take the child's name as one of the inputs and will also make this Mnode be the child node's parent.
   All is to reduce the probability of mistaken by users and be more convenient for users to use.
   And the return of this method is used to conveniently construct a chain of time series for users, such as:
![image](https://user-images.githubusercontent.com/15262869/111018495-0e4b3200-83f4-11eb-9452-4e5d4ff7a4a8.png)

 It can be very helpful when we wanna write test cases about different time series.

PR #2828, opened 2021-03-13
[ISSUE-2827] add an overrided addChild() method in MNode.

weizihan0110:
the related issue is at https://github.com/apache/iotdb/issues/2827

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/mnode/MNode.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/mnode/MNode.java
@@ -102,6 +102,35 @@ public class MNode implements Serializable {
     children.putIfAbsent(name, child);
   }
 
+  /**
+   * Add a child to the current mnode.
+   *
+   * This method will not take the child's name as one of the inputs and will also make this Mnode be child node's parent.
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/iotdb/db/metadata/mnode/MNode.java
```diff
@@ -102,6 +102,36 @@ public void addChild(String name, MNode child) {
...
+      // double check, children is volatile
+      synchronized (this) {
+        if (children == null) {
+          children = new ConcurrentHashMap<>();
+        }
+      }
+    }
+
+    children.putIfAbsent(child.getName(), child);
+    child.parent = this;
```
jixuan1989, line 131: this is inconsist with existing addChild(String name, MNode node);

Either set child.parent in the method or not.

jixuan1989: Besides, if you set child.parent in this method, better to set its parent before putting it in children map.

jixuan1989 (changes requested):
some minor issue.

PR #2834, opened 2021-03-14
Improve aggregation with value filter

SilverNarcissus:
Improve aggregation with value filter. Avoid duplicate series reader.


Alima777:
This PR has been moved to ClusterQueryOpt branch.

PR #2857, opened 2021-03-16
[IOTDB-1234] fix ImportCsv tool cannot parse empty string

HelKim:


server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -47,29 +47,9 @@
 import org.apache.iotdb.db.qp.executor.PlanExecutor;
 import org.apache.iotdb.db.qp.logical.Operator.OperatorType;
 import org.apache.iotdb.db.qp.physical.PhysicalPlan;
-import org.apache.iotdb.db.qp.physical.crud.AggregationPlan;
-import org.apache.iotdb.db.qp.physical.crud.AlignByDevicePlan;
+import org.apache.iotdb.db.qp.physical.crud.*;
```
wangchao316, line 50: general,  do not use import  *...

HelKim: > general, do not use import *...

Thanks

jixuan1989:
thanks for contribution!

jixuan1989:
Hi, though the PR is merged, it left one thing... 
You'd better add a Test for the case.

PR #2864, opened 2021-03-18
[IOTDB-1230]Support cross multi time partition when loading one TsFile

neuyilan:
When load one tsfile which cross multi time partition interval, we first read the tsfile and rewrite it to some new tsfiles according to its belongs time partition.


example/tsfile/src/main/java/org/apache/iotdb/tsfile/TsFileRead.java
```diff
@@ -39,7 +39,8 @@
  * TsFileWriteWithTSRecord or TsFileWriteWithTablet to generate the test.tsfile first
  */
 public class TsFileRead {
-  private static final String DEVICE1 = "device_1";
+
+  private static final String DEVICE1 = Constant.STORAGE_GROUP + ".device_1";
```
HTHou, line 43: Actually it's not necessary to indicate the storage group for TsFile module. The storage group is a concept of IoTDB only. We are supporting to use TsFile module stand alone and this example is for this purpose.

neuyilan: This is because in the TsFileWriteWithTablet class, the code write tsfile use root.sg info. if we write use the class and read use TsFileRead class, it will not work.
So in this case, I will remove the storage info in TsFileWriteWithTablet class.

neuyilan:
@HTHou  PTAL

server/src/main/java/org/apache/iotdb/db/tools/upgrade/TsFileOnlineUpgradeTool.java
```diff
@@ -18,92 +18,56 @@
...
-import java.nio.file.Files;
 import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 
-public class TsFileOnlineUpgradeTool implements AutoCloseable {
+public class TsFileOnlineUpgradeTool extends TsFileRewriteTool implements AutoCloseable {
```
sunjincheng121, line 50: Remove `implements AutoCloseable`?

server/src/main/java/org/apache/iotdb/db/tools/TsFileRewriteTool.java
```diff
@@ -0,0 +1,478 @@
...
+    }
+  }
+
+  /**
+   * Rewrite an old file to the latest version
+   *
+   * @param resourceToBeRewritten the tsfile which to be rewrite
+   * @param rewrittenResources the rewritten files
+   */
+  public static void rewriteTsfile(
```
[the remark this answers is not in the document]
neuyilan: Thanks, solved

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
```diff
@@ -975,7 +981,23 @@ private void loadFile(File file, OperateFilePlan plan) throws QueryProcessExcept
         createSchemaAutomatically(chunkGroupMetadataList, schemaMap, plan.getSgLevel());
       }
 
-      StorageEngine.getInstance().loadNewTsFile(tsFileResource);
+      List<TsFileResource> splitResources = new ArrayList();
+      try {
+        tsFileResource.getTimePartitionWithCheck();
```
sunjincheng121, line 986: How about we add a method that  judgment  whether it contains multiple partitions？

server/src/main/java/org/apache/iotdb/db/tools/upgrade/TsFileOnlineUpgradeTool.java
```diff
@@ -152,7 +111,7 @@ private void upgradeFile(List<TsFileResource> upgradedResources)
             if (newChunkGroup) {
```
sunjincheng121, line 111: Why do we need to check if it's a new chunk？i.e., Where is the purpose of this variable check?

server/src/main/java/org/apache/iotdb/db/tools/TsFileRewriteTool.java
```diff
@@ -0,0 +1,478 @@
...
+    List<List<ByteBuffer>> pageDataInChunkGroup = new ArrayList<>();
+    List<List<Boolean>> needToDecodeInfoInChunkGroup = new ArrayList<>();
+    byte marker;
+    List<MeasurementSchema> measurementSchemaList = new ArrayList<>();
+    String lastChunkGroupDeviceId = null;
+    try {
+      while ((marker = reader.readMarker()) != MetaMarker.SEPARATOR) {
+        switch (marker) {
+          case MetaMarker.CHUNK_HEADER:
+          case MetaMarker.ONLY_ONE_PAGE_CHUNK_HEADER:
```
sunjincheng121, line 153: Not sure, do we need deal with `CHUNK_HEADER`?

sunjincheng121 (commented):
Thanks for the nice PR! @neuyilan  

I only left some suggestions and questions. any feedback is welcome :)

Best,
Jincheng

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
```diff
@@ -975,7 +981,23 @@ private void loadFile(File file, OperateFilePlan plan) throws QueryProcessExcept
...
+      } catch (PartitionViolationException e) {
+        logger.info("try to split the tsfile={}", tsFileResource.getTsFile().getPath());
+        TsFileRewriteTool.rewriteTsFile(tsFileResource, splitResources);
```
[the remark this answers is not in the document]
neuyilan: The purpose here is to split and rewrite a tsFile only when it spans multiple partitions. When the tsFile is not partitioned across time, there is no need to split and rewrite it. Otherwise, we may need a function to determine whether the tsFile spans multiple time partitions. But that's exactly what the `getTimePartitionWithCheck()` function does, so I set the split logic here.

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
neuyilan: That is another method, I will add another method to judge whether it contains multiple partitions.

server/src/main/java/org/apache/iotdb/db/tools/upgrade/TsFileOnlineUpgradeTool.java
neuyilan: I see this code is not used, I think this is legacy code, I will remove this variable. Thanks.

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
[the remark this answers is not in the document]
sunjincheng121: I see, after we add a method to judgment whether it contains multiple partitions, then we do not need do the process logic in catch block.

server/src/main/java/org/apache/iotdb/db/tools/upgrade/TsFileOnlineUpgradeTool.java
```diff
@@ -18,92 +18,56 @@
...
 
   /**
-   * upgrade a single TsFile
+   * upgrade a single TsFile.
    *
-   * @param tsFileName old version tsFile's absolute path
+   * @param resourceToBeUpgraded the old file's resource which need to be upgrade.
    * @param upgradedResources new version tsFiles' resources
    */
   public static void upgradeOneTsfile(
```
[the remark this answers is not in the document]
neuyilan: Thanks, done

server/src/main/java/org/apache/iotdb/db/tools/upgrade/TsFileOnlineUpgradeTool.java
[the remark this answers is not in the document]
neuyilan: Yes, we can remove the AutoCloseable as it already extends TsFileRewriteTool.

server/src/main/java/org/apache/iotdb/db/tools/TsFileRewriteTool.java
neuyilan: In the new tsfile format, CHUNK_HEADER and ONLY_ONE_PAGE_CHUNK_HEADER both indicts the start of one chunk. @HTHou is an expert on tsfile format, please take a look at.

HTHou (approved):
I have tested the UpgradeTool with some small 0.11 files, this PR doesn't change the correctness of the upgrade feature.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/timeindex/FileTimeIndex.java
```diff
@@ -153,15 +153,25 @@ public long getTimePartition(String tsfilePath) {
...
     }
     return startPartitionId;
   }
 
+  @Override
+  public boolean isSpanMultiTimePartitions(String tsFilePath) {
+    try {
+      getTimePartitionWithCheck(tsFilePath);
+    } catch (PartitionViolationException e) {
+      return false;
```
sunjincheng121: Frankly, I don't recommend returning method values in a catch block. :)
I have an idea to do the following：

```
  @Override
  public long getTimePartition(String tsFilePath) throws PartitionViolationException {
    Pair<Boolean, Long> pair = getTimePartitionWithCheck();
    if(pair.left) {
      throw new PartitionViolationException(tsFilePath);
    }else {
      return pair.right;
    }
  }

  public Pair<Boolean, Long> getTimePartitionWithCheck() {
    Pair<Boolean, Long> pair = new Pair(true, -1);
    for (int index : deviceToIndex.values()) {
      long p = StorageEngine.getTimePartition(startTimes[index]);
      if ( pair.right == -1) {
        pair.right = p;
      } else {
        if ( pair.right != p) {
          return pair;
        }
      }
      p = StorageEngine.getTimePartition(endTimes[index]);
      if ( pair.right != p) {
        return pair;
      }
    }
    if ( pair.right == -1) {
      return pair;
    }
    pair.left = false;
    return pair;
  }
```

Of course，I know that there is already an implementation of 'gettimepartition(..)'. We can take advantage of this opportunity to refactor it. What do you think？

neuyilan: Hi,  I think the new `getTimePartition()` method will cost more time compare to the original one, because the new method above will iterate through all `deviceToIndex` arrays and calculate `deviceToIndex.size()` times to get the time partition id,  however in the original method, it will only calculate once. So I prefer to reserve the original method. And I have rearranged the code when checking or getting the time partition id, please take a look at it again.
Thanks~

server/src/main/java/org/apache/iotdb/db/tools/TsFileRewriteTool.java
jixuan1989: if "only one page in the chunk", then there is no pageStatistics in the page header. 

Line 170 and 171 has indicated that in PageHeader's deserialize method: 
```
PageHeader pageHeader =
                  reader.readPageHeader(dataType, header.getChunkType() == MetaMarker.CHUNK_HEADER);
```

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/timeindex/FileTimeIndex.java
sunjincheng121: Yes, the  code above is just the sample code. What we most concerned about here is about removing the business logic in catch block. i.e., my suggestion is that  return `Pair<Boolean, Long> ` in `getTimePartitionWithCheck()` for remove the logic in catch block. About 'getTimePartition(..)',  both keep the original implementation or make a little improvement, such as' getTimePartition (string tsfilepath, Boolean needCheck) throws partitionviationexception `. I respect your opinion. What do you think? :)

neuyilan: Yes, thanks for your careful review, I agree with your opinion, the new commit here removed the business logic in the catch block and use -1 to indict the tsFile that may cross multiple time partitions. 
As for `getTimePartition()` method ,  I prefer reserve the original two methods, one the `getTimePartition()` and the other is `getTimePartitionWithCheck() throws PartitionViolationException `, as almost method in the engine called is `getTimePartition()` method(Like the following picture shows, the method is called 37 times), the  `getTimePartitionWithCheck() throws PartitionViolationException` method only used when load one tsfile(Just 2 times ). If we combine the two method together such as `getTimePartition (string tsfilepath, Boolean needCheck) throws PartitionViolationException`, the caller may need to put the `needCheck` parameter all the time. it may be more complicated.
What do you think?
image
image

sunjincheng121 (approved):
Thanks for the update @neuyilan 

The code about `getTimePartitionWithCheck()`  is just the sample code. What we most concerned about here is about removing the business logic in catch block. i.e., my suggestion is that do not throw exception in getTimePartitionWithCheck() , e.g., we can return Pair<Boolean, Long> (I notice that your new commit return -1 as the cross partitions).
About 'getTimePartition(..)', both keep the original implementation or make a little improvement, such as' getTimePartition (string tsFilePath, Boolean needCheck) ......`, Feel free to make your decision and I respect your opinion due to some designs are related to personal code tastes.

And the last suggestion is keep the expression of abnormal Exception information `cross partitions` in `PartitionViolationException` consistent with `SPANS_MULTI_TIME_PARTITIONS_FLAG_ID`.  i.e.: `cross partitions` vs `spans partitions`. 

// BTW: I don't mean to find fault on purpose. I just share my ideas. Our common goal is to make iotdb better.  

So, overall，from my point of view, PR looks very good to me, and great thanks for your nice PR @neuyilan :)

Cheers,
Jincheng Sun

neuyilan:
> Our common goal is to make iotdb better
@sunjincheng121 
Thank you for your sincere suggestions. Your suggestions are very good. Yes, we all hope iotdb will be better and better :).
BTW, I've unified the vocabulary to use spans time partition to represent one tsfile spans multiple time partitions.
Thank you again for your careful review.

sunjincheng121:
The PR will be merged after the CI completed...

PR #2866, opened 2021-03-18
[IOTDB-1240]support show node command in cluster version

yifuzhou:
## Description

show child nodes query is only supported in standalone version. In cluster version, when we use show child nodes query, it will only display local node's data.


wangchao316:
good..
I have some small suggestion..
yifu ,you could add some UT or IT for your function..

wangchao316 (commented):
good..
I have some small suggestion..
yifu ,you could add some UT or IT for your function..

neuyilan (approved):
Good Job!

yifuzhou:
> good..
> I have some small suggestion..
> yifu ,you could add some UT or IT for your function..

Yes, I will add it later. Thanks~

wangchao316 (approved):
LGTM

LebronAl (approved):
LGTM~
BTW, could you please take a look at is there still some `show node command` not supported in cluster module? If so, you can fix it in next PR.

PR #2867, opened 2021-03-18
IOTDB-1241 support redirect query for cluster

mychaow:
## Description

see https://issues.apache.org/jira/browse/IOTDB-1241.

3 nodes, replica num 1.

one benchmark only sends query to one node.

Query performance improved by 37%.

disable redirect

image

enable redirect

image

### Content3 ...

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -452,7 +457,28 @@ public long getTimeout() {
    */
   public SessionDataSet executeQueryStatement(String sql)
       throws StatementExecutionException, IoTDBConnectionException {
-    return defaultSessionConnection.executeQueryStatement(sql, timeout);
+    try {
+      logger.info("{} execute sql {}", defaultSessionConnection.getEndPoint(), sql);
+      return defaultSessionConnection.executeQueryStatement(sql, timeout);
+    } catch (RedirectException e) {
+      handleQueryRedirection(e.getEndPoint());
+      if (enableQueryRedirection) {
```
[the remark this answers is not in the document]
mychaow: ok

jt2594838 (changes requested):
Despite your effort that should be thanked, the current implementation is not very efficient, because the decision of redirect or not is made **AFTER** readers are created, which means a lot of resources have already been wasted by the time the decision is made. Moreover, I wonder if you have released the resources properly after the decision.

I would recommend you do the redirection check **BEFORE** readers are created, and maybe it is done right after the physical plan is generated with the help of the partition table.

session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -585,6 +641,14 @@ private void handleRedirection(String deviceId, EndPoint endpoint)
     }
   }
 
+  private void handleQueryRedirection(EndPoint endPoint) throws IoTDBConnectionException {
+    if (enableQueryRedirection) {
+      defaultSessionConnection.close();
+      defaultSessionConnection = constructSessionConnection(this, endPoint, zoneId);
+      defaultSessionConnection.setEnableRedirect(enableQueryRedirection);
+    }
+  }
```
jt2594838, line 650: Maybe it would be better to cache a certain number of connections for different endpoints.
If the client queries root.sg1, then root.sg2, then root.sg1, then root.sg2... a new connection will be created for each query if the two SGs are located in different data groups.

mychaow: yes, we could cache a certain number of connections for different node.

mychaow:
> Despite your effort that should be thanked, the current implementation is not very efficient, because the decision of redirect or not is made **AFTER** readers are created, which means a lot of resources have already been wasted by the time the decision is made. Moreover, I wonder if you have released the resources properly after the decision.
> 
> I would recommend you do the redirection check **BEFORE** readers are created, and maybe it is done right after the physical plan is generated with the help of the partition table.

Ok, I will think about whether the plan could be redirected to the right node with the help of the partition table.

wangchao316:
hi chao, there is 4 Bugs and 8 code smell, which maybe deal it.
Thanks.

mychaow:
> hi chao, there is 4 Bugs and 8 code smell, which maybe deal it.
> Thanks.

hi, 8  code smell are system.out, I think it's not necessary to deal with it.

wangchao316:
> > hi chao, there is 4 Bugs and 8 code smell, which maybe deal it.
> > Thanks.
> 
> hi, 8 code smell are system.out, I think it's not necessary to deal with it.

ok, no problem.

session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -643,9 +643,24 @@ private void handleRedirection(String deviceId, EndPoint endpoint)
...
+                  return sessionConnection;
+                } catch (IoTDBConnectionException ex) {
+                  tmp.set(ex);
+                  return null;
+                }
+              });
+      if (connection == null) {
+        throw new IoTDBConnectionException(tmp.get());
+      }
+      defaultSessionConnection = endPointToSessionConnection.get(endPoint);
```
jt2594838, line 663: `defaultSessionConnection = connection` should be enough.

jt2594838 (approved):

jt2594838:
Maybe you can suppress these warnings.

jt2594838 (approved):

wangchao316 (approved):
LGTM

PR #2868, opened 2021-03-18
simplify the Avg formula on removing redundant operations

yifuzhou:
## Description

I did serval improvements of the AVG function:

The formula to calculate AVG function is complex, it can be simplified. I removed reduncdant operations.
When we calculate the AVG, the result value will transfer into Double type. in Java, the calculation of double type will always cause the loss of accuracy. And in our test case, there are lots of values which are loss of accuracy. I try to fix it by using assertEquals(double expected, double actual, double delta)


Alima777 (approved):
LGTM :D

HTHou:
There are some tests failed. Please take a look...

yifuzhou:
> There are some tests failed. Please take a look...

Thanks HTHou. Yes, I have already fixed it but it still has a weird fail test case. Last time I saw it and just rerun ci again without any change and it pass...

LebronAl (approved):
LGTM

PR #2725, opened 2021-02-23
[IOTDB-1164]Optimize the executeBatch interface in JDBC

yanhongwangg:
https://issues.apache.org/jira/browse/IOTDB-1164

PR #2872, opened 2021-03-19
[IOTDB-1236] Improve jdbc performance for creating timeseries in cluster module

jxlgzwh:
Jira: 
https://issues.apache.org/jira/browse/IOTDB-1236

Improve jdbc performance for creating timeseries in cluster module

Refer to InsertRowsPlan to judge the type of operation specifically, and execute CreateMultiTimeSeriesPlan without affecting the execution sequence
1. Determine the type of operation
2. Assemble batch objects
3. If it is not INSERT and CREATE_TIMESERIES operations, start to execute the operation queue in order

reference
https://github.com/apache/iotdb/pull/2725

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
+++ b/server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
@@ -553,14 +553,8 @@ public class TSServiceImpl implements TSIService.Iface, ServerContext {
         }
 
         if (physicalPlan.getOperatorType().equals(OperatorType.INSERT)) {
-          if (lastOperatorType == null || OperatorType.INSERT == lastOperatorType) {
-            insertRowsPlan =
-                executeList.size() > 0
-                    ? (InsertRowsPlan) executeList.get(executeList.size() - 1)
-                    : new InsertRowsPlan();
-            if (lastOperatorType == null) {
-              executeList.add(insertRowsPlan);
-            }
+          if (OperatorType.INSERT == lastOperatorType) {
+            insertRowsPlan = (InsertRowsPlan) executeList.get(executeList.size() - 1);
           } else {
             insertRowsPlan = new InsertRowsPlan();
             executeList.add(insertRowsPlan);
@@ -576,14 +570,8 @@ public class TSServiceImpl implements TSIService.Iface, ServerContext {
             executeList = new ArrayList();
           }
         } else if (physicalPlan.getOperatorType().equals(OperatorType.CREATE_TIMESERIES)) {
-          if (lastOperatorType == null || OperatorType.CREATE_TIMESERIES == lastOperatorType) {
-            multiPlan =
-                executeList.size() > 0
-                    ? (CreateMultiTimeSeriesPlan) executeList.get(executeList.size() - 1)
-                    : new CreateMultiTimeSeriesPlan();
-            if (lastOperatorType == null) {
-              executeList.add(multiPlan);
-            }
+          if (OperatorType.CREATE_TIMESERIES == lastOperatorType) {
+            multiPlan = (CreateMultiTimeSeriesPlan) executeList.get(executeList.size() - 1);
           } else {
             multiPlan = new CreateMultiTimeSeriesPlan();
             executeList.add(multiPlan);
--- a/server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
+++ b/server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
@@ -461,6 +461,16 @@ public class TSServiceImpl implements TSIService.Iface, ServerContext {
 
     InsertRowsPlan insertRowsPlan = new InsertRowsPlan();
     int index = 0;
+
+    CreateMultiTimeSeriesPlan multiPlan = new CreateMultiTimeSeriesPlan();
+    List<PartialPath> paths = new ArrayList<>();
+    List<TSDataType> tsDataTypes = new ArrayList<>();
+    List<TSEncoding> tsEncodings = new ArrayList<>();
+    List<CompressionType> tsCompressionTypes = new ArrayList<>();
+    List<Map<String, String>> tagsList = new ArrayList<>();
+    List<Map<String, String>> attributesList = new ArrayList<>();
+    List<String> aliasList = new ArrayList<>();
+
     for (int i = 0; i < req.getStatements().size(); i++) {
       String statement = req.getStatements().get(i);
       try {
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -461,6 +461,16 @@ public TSStatus executeBatchStatement(TSExecuteBatchStatementReq req) {
 
     InsertRowsPlan insertRowsPlan = new InsertRowsPlan();
     int index = 0;
+
+    CreateMultiTimeSeriesPlan multiPlan = new CreateMultiTimeSeriesPlan();
```
jixuan1989, line 465: you cannot create so many Object instances before you confirm some statements can be wrapped by a CreateMultiTimeSeriesPlan.

jixuan1989 (changes requested):
1. Add Tests for new codes
2. Describe your design or idea in the PR.

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -506,6 +534,30 @@ public TSStatus executeBatchStatement(TSExecuteBatchStatementReq req) {
...
+        for (Entry<Integer, TSStatus> entry : multiPlan.getResults().entrySet()) {
+          result.set(startIndex + entry.getKey(), entry.getValue());
+        }
+        isAllSuccessful = false;
+      }
+    }
+
     Measurement.INSTANCE.addOperationLatency(Operation.EXECUTE_JDBC_BATCH, t1);
     return isAllSuccessful
         ? RpcUtils.getStatus(TSStatusCode.SUCCESS_STATUS, "Execute batch statements successfully")
```
neuyilan, line 563: We can not do the meta operation in the last step, for example, we have the following SQL statements:

example1: 

> delete timeseries root.ln.wf01.wt01.status
> insert into root.ln.wf01.wt01(timestamp,status) values(1,true)

example2:

> create timeseries root.ln.wf01.wt01.temperature with datatype=FLOAT,encoding=RLE
> create timeseries root.ln.wf01.wt01.status with datatype=BOOLEAN,encoding=PLAIN
> // insert one record with wrong type
> insert into root.ln.wf01.wt01(timestamp,status) values(1,"this is wrong type, can not be insert success")

For example 1, if we first execute the insert statement, then execute the delete statement, all data will be deleted.

For example 2, if we first execute the insert statement, IoTDB will create **TEXT** type for root.ln.wf01.wt01.status  automatically, which is in conflict with the  user given type(**BOOLEN**).

neuyilan (changes requested):
I guess the function of your PR is to improve the speed when all the executed statements are `create timeseries ...,` but please note that the user's execution order cannot be disturbed.

BTW, the operation of `creating time series` is not very frequent. Is it worth processing separately for this metadata operation?

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
jxlgzwh: already fixed

jxlgzwh:
> 1. Add Tests for new codes
> 2. Describe your design or idea in the PR.

Changes completed

jxlgzwh:
I found an error in the executeInsertRowsPlan method of https://github.com/apache/iotdb/pull/2725

tsStatus.equals(RpcUtils.SUCCESS_STATUS) 

This method will always return false because
The message of tsStatus has a set value, so isSetMessage is true
The message of RpcUtils.SUCCESS_STATUS is not set, so isSetMessage is false

So I modified it to compare the code and fix it together with pr this time

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
jxlgzwh: Changes completed

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -447,7 +447,85 @@ private boolean executeInsertRowsPlan(InsertRowsPlan insertRowsPlan, List<TSStat
...
+    multiPlan.getCompressors().add(compressor);
+    multiPlan.getTags().add(tags);
+    multiPlan.getAttributes().add(attributes);
+    multiPlan.getAlias().add(alias);
+  }
+
+  private boolean executeBatchList(ArrayList executeList, List<TSStatus> result) {
+    boolean isAllSuccessful = true;
+    if (executeList.size() > 0) {
+      for (int j = 0; j < executeList.size(); j++) {
```
jt2594838, line 515: `if (executeList.size() > 0) {` can be removed.

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -459,8 +537,11 @@ public TSStatus executeBatchStatement(TSExecuteBatchStatementReq req) {
       return RpcUtils.getStatus(TSStatusCode.NOT_LOGIN_ERROR);
     }
 
-    InsertRowsPlan insertRowsPlan = new InsertRowsPlan();
+    InsertRowsPlan insertRowsPlan;
     int index = 0;
+    ArrayList executeList = new ArrayList();
```
jt2594838, line 542: It would be better to define a variable using interface type.

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -472,19 +553,47 @@ public TSStatus executeBatchStatement(TSExecuteBatchStatementReq req) {
...
           insertRowsPlan.addOneInsertRowPlan((InsertRowPlan) physicalPlan, index);
           index++;
-          if (i == req.getStatements().size() - 1
-              && !executeInsertRowsPlan(insertRowsPlan, result)) {
-            isAllSuccessful = false;
+          if (i == req.getStatements().size() - 1) {
+            if (!executeBatchList(executeList, result)) {
+              isAllSuccessful = false;
+            }
+            executeList = new ArrayList();
```
jt2594838, line 570: Is it necessary to create a new list now that you have processed the last statement?
And maybe this branch can be extracted from the loop.

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -472,19 +553,47 @@ public TSStatus executeBatchStatement(TSExecuteBatchStatementReq req) {
...
-          if (insertRowsPlan.getRowCount() > 0) {
-            if (!executeInsertRowsPlan(insertRowsPlan, result)) {
+          lastOperatorType = physicalPlan.getOperatorType();
+          if (executeList.size() > 0) {
+            if (!executeBatchList(executeList, result)) {
               isAllSuccessful = false;
             }
-            index = 0;
-            insertRowsPlan = new InsertRowsPlan();
+            executeList = new ArrayList();
```
jt2594838, line 596: `List.clear` should be enough here.

server/src/test/java/org/apache/iotdb/db/integration/IoTDBExecuteBatchIT.java
```diff
@@ -26,11 +26,7 @@
 import org.junit.Before;
 import org.junit.Test;
 
-import java.sql.Connection;
-import java.sql.DriverManager;
-import java.sql.ResultSet;
-import java.sql.SQLException;
-import java.sql.Statement;
+import java.sql.*;
```
jt2594838, line 29: I think we prefer not to use wildcards in imports.

jt2594838 (changes requested):

mychaow (approved):
LGTM

jxlgzwh:
@neuyilan @jixuan1989 @jt2594838  PTAL again. Thanks!

neuyilan (approved):
LGTM

jt2594838 (approved):

PR #2879, opened 2021-03-20
IOTDB-854  Limit the memory foorprint of the committed log cache

mychaow:
## Description
see https://issues.apache.org/jira/browse/IOTDB-854

limit the total  size of raftlog in memory.

### Content3 ...

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterConfig.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterConfig.java
@@ -65,6 +65,7 @@ public class ClusterConfig {
   /** max number of committed logs in memory */
   private int maxNumOfLogsInMem = 1000;
 
+  /** max memory size of committed logs in memory, default 512M */
   private long maxMemorySizeForRaftLog = 536870912;
 
   /** deletion check period of the submitted log */
--- a/cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterConfig.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterConfig.java
@@ -65,6 +65,8 @@ public class ClusterConfig {
   /** max number of committed logs in memory */
   private int maxNumOfLogsInMem = 1000;
 
+  private long maxMemorySizeForRaftLog = 536870912;
+
   /** deletion check period of the submitted log */
   private int logDeleteCheckIntervalSecond = -1;
 
@@ -379,6 +381,14 @@ public class ClusterConfig {
 
   public void setMaxRaftLogIndexSizeInMemory(int maxRaftLogIndexSizeInMemory) {
     this.maxRaftLogIndexSizeInMemory = maxRaftLogIndexSizeInMemory;
+  }
+
+  public long getMaxMemorySizeForRaftLog() {
+    return maxMemorySizeForRaftLog;
+  }
+
+  public void setMaxMemorySizeForRaftLog(long maxMemorySizeForRaftLog) {
+    this.maxMemorySizeForRaftLog = maxMemorySizeForRaftLog;
   }
 
   public int getMaxRaftLogPersistDataSizePerFile() {
--- a/cluster/src/main/java/org/apache/iotdb/cluster/log/manage/CommittedEntryManager.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/log/manage/CommittedEntryManager.java
@@ -41,6 +41,8 @@ public class CommittedEntryManager {
   // memory cache for logs which have been persisted in disk.
   private List<Log> entries;
 
+  private long entryTotalMemSize;
+
   /**
    * Note that it is better to use applyingSnapshot to update dummy entry immediately after this
    * instance is created.
@@ -48,6 +50,7 @@ public class CommittedEntryManager {
   CommittedEntryManager(int maxNumOfLogInMem) {
     entries = Collections.synchronizedList(new ArrayList<>(maxNumOfLogInMem));
     entries.add(new EmptyContentLog(-1, -1));
+    entryTotalMemSize = 0;
   }
 
   /**
@@ -209,6 +212,9 @@ public class CommittedEntryManager {
         0,
         new EmptyContentLog(
             entries.get(index).getCurrLogIndex(), entries.get(index).getCurrLogTerm()));
+    for (int i = 1; i <= index; i++) {
+      entryTotalMemSize -= entries.get(i).getByteSize();
+    }
     entries.subList(1, index + 1).clear();
   }
 
@@ -225,6 +231,9 @@ public class CommittedEntryManager {
     }
     long offset = appendingEntries.get(0).getCurrLogIndex() - getDummyIndex();
     if (entries.size() - offset == 0) {
+      for (int i = 0; i < appendingEntries.size(); i++) {
+        entryTotalMemSize += appendingEntries.get(i).getByteSize();
+      }
       entries.addAll(appendingEntries);
     } else if (entries.size() - offset > 0) {
       throw new TruncateCommittedEntryException(
@@ -245,5 +254,30 @@ public class CommittedEntryManager {
   @TestOnly
   List<Log> getAllEntries() {
     return entries;
+  }
+
+  public long getEntryTotalMemSize() {
+    return entryTotalMemSize;
+  }
+
+  public void setEntryTotalMemSize(long entryTotalMemSize) {
+    this.entryTotalMemSize = entryTotalMemSize;
+  }
+
+  /**
+   * check how many logs could be reserved in memory.
+   *
+   * @param maxMemSize the max memory size for old committed log
+   * @return max num to reserve old committed log
+   */
+  public int maxLogNumShouldReserve(long maxMemSize) {
+    long totalSize = 0;
+    for (int i = entries.size() - 1; i >= 1; i--) {
+      if (totalSize + entries.get(i).getByteSize() > maxMemSize) {
+        return entries.size() - 1 - i;
+      }
+      totalSize += entries.get(i).getByteSize();
+    }
+    return entries.size();
   }
 }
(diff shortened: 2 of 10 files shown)


cluster/src/main/java/org/apache/iotdb/cluster/log/manage/RaftLogManager.java
```diff
@@ -582,14 +585,30 @@ public void commitTo(long newCommitIndex) throws LogExecutionException {
           .clear();
     }
     try {
-      int currentSize = committedEntryManager.getTotalSize();
-      int deltaSize = entries.size();
-      if (currentSize + deltaSize > maxNumOfLogsInMem) {
-        int sizeToReserveForNew = maxNumOfLogsInMem - deltaSize;
+      boolean needCompactedLog = false;
```
LebronAl, line 588: typo? verb + adj + noun? maybe `needToCompactLog`?

cluster/src/main/java/org/apache/iotdb/cluster/log/manage/CommittedEntryManager.java
```diff
@@ -246,4 +255,29 @@ public void append(List<Log> appendingEntries) throws TruncateCommittedEntryExce
...
+   */
+  public int maxLogNumShouldReserve(long maxMemSize) {
+    long totalSize = 0;
+    for (int i = entries.size() - 1; i >= 1; i--) {
+      if (totalSize + entries.get(i).getByteSize() > maxMemSize) {
+        return entries.size() - 1 - i;
+      }
+      totalSize += entries.get(i).getByteSize();
+    }
+    return entries.size();
```
LebronAl, line 281: `entries.size() - 1` ?

LebronAl (commented):
Great work! 
I have left some reviews, PTAL.

cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterConfig.java
```diff
@@ -65,6 +65,8 @@
   /** max number of committed logs in memory */
   private int maxNumOfLogsInMem = 1000;
 
+  private long maxMemorySizeForRaftLog = 536870912;
```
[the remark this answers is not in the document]
mychaow: yes

cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
```diff
@@ -529,12 +531,15 @@ public long appendEntries(AppendEntriesRequest request) throws UnknownLogTypeExc
 
     long response;
     List<Log> logs = new ArrayList<>();
+    int logByteSize = 0;
     long startTime = Timer.Statistic.RAFT_RECEIVER_LOG_PARSE.getOperationStartTime();
     for (ByteBuffer buffer : request.getEntries()) {
       buffer.mark();
       Log log;
+      logByteSize = buffer.limit() - buffer.position();
```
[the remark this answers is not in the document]
mychaow: yes, you are right.

cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterConfig.java
```diff
@@ -65,6 +65,8 @@
   /** max number of committed logs in memory */
   private int maxNumOfLogsInMem = 1000;
```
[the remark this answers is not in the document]
mychaow: seems not very useful.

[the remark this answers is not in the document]
LebronAl: Maybe we can just eliminate this parameter in this PR as you have just mentioned there are many unuseful parameters in IoTDB by mailing list.

wangchao316 (approved):
LGTM

cluster/src/main/java/org/apache/iotdb/cluster/log/manage/RaftLogManager.java
```diff
@@ -582,14 +585,30 @@ public void commitTo(long newCommitIndex) throws LogExecutionException {
...
-      int deltaSize = entries.size();
-      if (currentSize + deltaSize > maxNumOfLogsInMem) {
-        int sizeToReserveForNew = maxNumOfLogsInMem - deltaSize;
+      boolean needToCompactLog = false;
+      int numToReserveForNew = minNumOfLogsInMem;
+      if (committedEntryManager.getTotalSize() + entries.size() > maxNumOfLogsInMem) {
+        needToCompactLog = true;
+        numToReserveForNew = maxNumOfLogsInMem - entries.size();
+      }
+
```
neuyilan, line 594: if `entries.size()=2,minNumOfLogsInMem=100,maxNumOfLogsInMem=200`, `committedEntryManager.getTotalSize()=199`, then the calculated `numToReserveForNew =200-2=198?` I think it should be `min(198, 100) = 100`

cluster/src/main/java/org/apache/iotdb/cluster/log/manage/RaftLogManager.java
```diff
@@ -582,14 +585,30 @@ public void commitTo(long newCommitIndex) throws LogExecutionException {
...
+      for (Log entry : entries) {
+        newEntryMemSize += entry.getByteSize();
+      }
+      int sizeToReserveForNew = minNumOfLogsInMem;
+      if (newEntryMemSize + committedEntryManager.getEntryTotalMemSize() > maxLogMemSize) {
+        needToCompactLog = true;
+        sizeToReserveForNew =
+            committedEntryManager.maxLogNumShouldReserve(maxLogMemSize - newEntryMemSize);
+      }
+
```
neuyilan, line 605: What about if  `maxLogMemSize <  newEntryMemSize ` ? 
In this case, this implementation will remove all commit logs?

neuyilan (approved):
LGTM

cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
```diff
@@ -948,6 +953,7 @@ TSStatus processPlanLocally(PhysicalPlan plan) {
 
       log.setPlan(plan);
       plan.setIndex(log.getCurrLogIndex());
+      log.setByteSize(log.serialize().array().length);
```
LebronAl, line 956: According to my previous test results, the `serialize()` step doesn't take very little overhead. Since we need serialization to calculate the memory, the leader should try to reuse the serialization here with the one sent to the followers to reduce overhead. What's your opinion?

cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
```diff
@@ -981,6 +987,7 @@ private TSStatus processPlanLocallyV2(PhysicalPlan plan) {
       log.setCurrLogIndex(logManager.getLastLogIndex() + 1);
       log.setPlan(plan);
       plan.setIndex(log.getCurrLogIndex());
+      log.setByteSize(log.serialize().array().length);
```
LebronAl, line 990: same as above

cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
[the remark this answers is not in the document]
LebronAl: Sorry, I don't see that `for` loop. Then it seems  that  `logByteSize = buffer.limit() - buffer.position();` is right ...

Just lack of leader logic here

cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
mychaow: ok, I will try reuse it.

cluster/src/main/java/org/apache/iotdb/cluster/log/manage/RaftLogManager.java
```diff
@@ -582,14 +585,35 @@ public void commitTo(long newCommitIndex) throws LogExecutionException {
...
+      if (committedEntryManager.getTotalSize() + entries.size() > maxNumOfLogsInMem) {
+        needToCompactLog = true;
+        numToReserveForNew = maxNumOfLogsInMem - entries.size();
+      }
+
+      long newEntryMemSize = 0;
+      for (Log entry : entries) {
+        if (entry.getByteSize() == 0) {
+          logger.warn("{} should not go here, must be send to the follower, so the log has been serialized", entry);
+          newEntryMemSize += entry.serialize().array().length;
```
jt2594838, line 599: It always goes here when there is only one replica.
By the way, it is wasteful to serialize the log here when there is only one replica. Maybe we can use a rough estimation or leave it to the restriction of the number of entries.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
```diff
@@ -529,12 +531,15 @@ public long appendEntries(AppendEntriesRequest request) throws UnknownLogTypeExc
 
     long response;
     List<Log> logs = new ArrayList<>();
+    int logByteSize = 0;
     long startTime = Timer.Statistic.RAFT_RECEIVER_LOG_PARSE.getOperationStartTime();
     for (ByteBuffer buffer : request.getEntries()) {
       buffer.mark();
       Log log;
+      logByteSize = buffer.array().length;
```
jt2594838, line 539: ByteBuffers from thrift may only be a slice of a larger array, so `array.length` may not refer to its real size. Use `buffer.limit() - buffer.pos()` instead.

jt2594838 (approved):

cluster/src/main/java/org/apache/iotdb/cluster/log/manage/RaftLogManager.java
mychaow: 1. I serialize the log before entering commitTo when there is only one replica. So it's not right to here. 
2. Yes, it's wasteful to serialize the log when there is only one replica.  Maybe we could pass it.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
mychaow: So, the buffer must be in read state.

LebronAl (approved):
LGTM

PR #2881, opened 2021-03-21
[IOTDB-1249][To Vector] Serialize of device MNode with device template

samperson1997:
Device MNode with device template must be serialized so that template could be set when recovering.


server/src/main/java/org/apache/iotdb/db/metadata/logfile/MLogWriter.java
```diff
@@ -366,6 +371,7 @@ public void operation(String cmd) throws IOException, MetadataException {
         createTimeseries(plan);
         break;
       case MetadataOperationType.CREATE_ALIGNED_TIMESERIES:
+      case MetadataOperationType.AUTO_CREATE_DEVICE_MNODE:
```
SilverNarcissus, line 374: Maybe add a "TODO" here for later polishing

server/src/test/java/org/apache/iotdb/db/qp/physical/InsertTabletPlanTest.java
```diff
@@ -186,29 +208,50 @@ public void testInsertTabletPlanWithDeviceTemplate()
...
     Assert.assertEquals(3, dataSet.getPaths().size());
     while (dataSet.hasNext()) {
       RowRecord record = dataSet.next();
       Assert.assertEquals(6, record.getFields().size());
     }
+
+    // test recover
+    // we want to recover
+    EnvironmentUtils.stopDaemon();
+    // wait for close
```
SilverNarcissus, line 239: You should use 
```
IoTDB.metaManager.clear();
```
to clear meta manager for testing recovery

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -390,6 +391,10 @@ public void operation(PhysicalPlan plan) throws IOException, MetadataException {
         SetDeviceTemplatePlan setDeviceTemplatePlan = (SetDeviceTemplatePlan) plan;
         setDeviceTemplate(setDeviceTemplatePlan);
         break;
+      case AUTO_CREATE_DEVICE_MNODE:
```
SilverNarcissus, line 394: please add
```
this.templateMap.clear();
```
in the clear method to ensure complete cleanness

SilverNarcissus (changes requested):
Fantastic work! Now we can recovery from template! Only 3 issues, please have a look~

server/src/main/java/org/apache/iotdb/db/metadata/logfile/MLogWriter.java
samperson1997: Since `AUTO_CREATE_DEVICE_MNODE` could never appear in the old mlog, so we do not need to support this case forever : )

server/src/test/java/org/apache/iotdb/db/qp/physical/InsertTabletPlanTest.java
samperson1997: Added, thank you

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
samperson1997: Added

SilverNarcissus (approved):
Nice work!

PR #2882, opened 2021-03-22
Use switch case in enums class 

yifuzhou:
## Description
Fix bad smell code, it is not necessary to use for loop to find enums type. It can use switch case to reduce complexity from O(n) ->O(1)...


wangchao316 (approved):
LGTM

LebronAl (approved):
LGTM

PR #2886, opened 2021-03-23
[IOTDB-1251] optimize route intervals

wangchao316:
When data is not partitioned, all data of a sequence is stored in partition 0. Therefore, during the query, we only need to route to partition 0 regardless of the interval. Therefore, partition calculation is not performed.
This reduces the time spent on partition calculation and routing to other partitions.
Current problems:
1. When
firstLB == Long.MIN_VALUE || lastUB == Long.MAX_VALUE
route all partiongroup.
2. For other time conditions, partition calculation is performed.


wangchao316:
@LebronAl @neuyilan @mychaow @jt2594838 , hi could you please review?
Thanks

LebronAl (approved):
LGTM~
It's a very necessary optimization!

mychaow (approved):
LGTM

neuyilan (approved):
LGTM

PR #2887, opened 2021-03-23
[IOTDB-1252] optimize test mult-times perform

wangchao316:
Currently, we use MemberTest as the base class and reuse some of its methods. If one test-class extends the MemberTest , the test case of the MemberTest class is executed again when the test case of the subclass is executed. As a result, some tests in the MemberTest class are executed multiple times, increasing the test case execution time
image
Solution:
1. test case extract in MemberTest.
2. MemberTest rename a basemember class，which can not add new test case.


wangchao316:
@LebronAl @neuyilan @mychaow  hi, could you please review?
Thanks..

LebronAl (approved):
LGTM

neuyilan (approved):
LGTM

PR #2888, opened 2021-03-23
[IOTDB-1253] [To Vector] Support null value in aligned time series

HTHou:
## Support the nullable aligned timeseries data by using bitmaps to recode the null value position of the tablet.

Some subtasks shows as below:

- [x] Use the bitmaps to indicate the null-value position in the tablet on the client side
- [x] Serialize and send the nullable tablet to the server side
- [x] Use the bitmaps in VectorTVList to show the null values in the memtable
- [x] Support nullable aligned timeseries value in TsFile (already done in Vector branch)
- [x] Support query memory data
- [x] Support query disk data
- [x] Add UT / IT
- [x] Fix serialize insertTablePlan with bitmap bug 

## Support inserting nullable aligned timeseries value with sql statement

```
IoTDB> insert into root.t1.wf01.wt01(time, (status, temperature)) values (4000, (true, 17.1))
IoTDB> insert into root.t1.wf01.wt01(time, (status, temperature)) values (4001, (false, null))
IoTDB> insert into root.t1.wf01.wt01(time, (status, temperature)) values (4002, (null, 3.9))
IoTDB> select * from root
+-----------------------------+------------------------+-----------------------------+
|                         Time|root.t1.wf01.wt01.status|root.t1.wf01.wt01.temperature|
+-----------------------------+------------------------+-----------------------------+
|1970-01-01T08:00:04.000+08:00|                    true|                         17.1|
|1970-01-01T08:00:04.001+08:00|                   false|                         null|
|1970-01-01T08:00:04.002+08:00|                    null|                          3.9|
+-----------------------------+------------------------+-----------------------------+
Total line number = 3
It costs 0.207s
IoTDB> show timeseries
+-----------------------------+-----+-------------+--------+--------+-----------+----+----------+
|                   timeseries|alias|storage group|dataType|encoding|compression|tags|attributes|
+-----------------------------+-----+-------------+--------+--------+-----------+----+----------+
|     root.t1.wf01.wt01.status| null|      root.t1| BOOLEAN|     RLE|     SNAPPY|null|      null|
|root.t1.wf01.wt01.temperature| null|      root.t1|   FLOAT| GORILLA|     SNAPPY|null|      null|
+-----------------------------+-----+-------------+--------+--------+-----------+----+----------+
Total line number = 2
It costs 0.049s
```


server/src/main/java/org/apache/iotdb/db/utils/TypeInferenceUtils.java
```diff
@@ -81,6 +81,8 @@ public static TSDataType getPredictedDataType(Object value, boolean inferType) {
           return floatingStringInferType;
         }
         // "NaN" is returned if the NaN Literal is given in Parser
+      } else if ("null".equals(strValue)) {
```
SilverNarcissus, line 84: What if user pass a "NULL" as value? Should we think "NULL" also a null?

session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -1453,6 +1453,10 @@ private int calculateLength(List<TSDataType> types, List<Object> values)
   private void putValues(List<TSDataType> types, List<Object> values, ByteBuffer buffer)
       throws IoTDBConnectionException {
     for (int i = 0; i < values.size(); i++) {
+      if (values.get(i) == null) {
+        ReadWriteIOUtils.write((byte) -2, buffer);
```
SilverNarcissus, line 1457: Maybe this "-2" can be a constant with a understandable name.

SilverNarcissus (changes requested):
Nice work and abundant tests!  Only minor issues, please have a look~

session/src/main/java/org/apache/iotdb/session/Session.java
HTHou: Sure

server/src/main/java/org/apache/iotdb/db/utils/TypeInferenceUtils.java
HTHou: I'm going to let antlr transfer `NULL` to `null`. 
But this strValue may also come from session client by insertStringRecord, right? Yes, I think consider `NULL` here is necessary.

SilverNarcissus (approved):
Look good to me~  Only need to pass CI before we merge.

PR #2892, opened 2021-03-23
[IOTDB-1255] refactor cluster package and jar name

wangchao316:
We should use statistical naming throughout the project, for jar names, and zip names.

current， cluster package-name is cluster-*.zip, and cluster-*.jar

solution：
image


LebronAl (approved):
LGTM

PR #2897, opened 2021-03-24
Add the method size() to Row in UDF

543202718:
Currently, UDF supports multiple input series but the method to return the number of input series is lacked.
+ All of the input series are organized as a table-like structrue.
+ For each row, we can get a specifc value with index.
+ When the index is larger than the number of columns, UDF throws an exception and crashes. 

Thus, we need the following method:
```
int size(); // return the number of columns
```  

This pull request implements this method.


SteveYurongSu (approved):
LGTM. Thanks for your contribution! :D

Issue #2900, opened 2021-03-24
Run IoTDB main meets 'Wrong max_degree_of_index_node, please set as'


yuqi1129:
**Describe the bug**

When start IoTDB Main in IDE, the following error occur

```java
19:16:55.203 [main] WARN org.apache.iotdb.db.conf.IoTDBDescriptor - Cannot find IOTDB_HOME or IOTDB_CONF environment variable when loading config file iotdb-engine.properties, use default configuration
19:16:55.215 [main] WARN org.apache.iotdb.tsfile.common.conf.TSFileDescriptor - Failed to find config file iotdb-engine.properties at classpath, use default configuration
19:16:55.219 [main] WARN org.apache.iotdb.db.conf.IoTDBDescriptor - Couldn't load the configuration from any of the known sources.
19:16:55.232 [main] INFO org.apache.iotdb.db.conf.IoTDBConfigCheck - Starting IoTDB UNKNOWN
19:16:55.233 [main] ERROR org.apache.iotdb.db.conf.IoTDBConfigCheck - Wrong max_degree_of_index_node, please set as: 1024 !

```

**To Reproduce**
1. checkout master branch 
2. run IoTDB main  

**Expected behavior**
Run OK

**Screenshots**

![image](https://user-images.githubusercontent.com/15794564/112302172-bbb42480-8cd5-11eb-92d4-986b3020999b.png)

**Desktop (please complete the following information):**
 - Mac OSX

**Additional context**

PR #2901, opened 2021-03-24
Fix Run IoTDB main meets 'Wrong max_degree_of_index_node, please set as'

yuqi1129:
## Description

please see https://github.com/apache/iotdb/issues/2900

### Content3 ...

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

Just change a config

### DIFF
--- a/tsfile/src/main/java/org/apache/iotdb/tsfile/common/conf/TSFileConfig.java
+++ b/tsfile/src/main/java/org/apache/iotdb/tsfile/common/conf/TSFileConfig.java
@@ -75,7 +75,7 @@ public class TSFileConfig implements Serializable {
   /** The maximum number of data points in a page, default value is 1024 * 1024. */
   private int maxNumberOfPointsInPage = 1024 * 1024;
   /** The maximum degree of a metadataIndex node, default value is 1024 */
-  private int maxDegreeOfIndexNode = 256;
+  private int maxDegreeOfIndexNode = 1024;
   /** Data type for input timestamp, TsFile supports INT32 or INT64. */
   private String timeSeriesDataType = "INT64";
   /** Max length limitation of input string. */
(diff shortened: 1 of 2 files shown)


tsfile/src/main/java/org/apache/iotdb/tsfile/common/conf/TSFileConfig.java
```diff
@@ -75,7 +75,7 @@
   /** The maximum number of data points in a page, default value is 1024 * 1024. */
   private int maxNumberOfPointsInPage = 1024 * 1024;
   /** The maximum degree of a metadataIndex node, default value is 1024 */
-  private int maxDegreeOfIndexNode = 256;
+  private int maxDegreeOfIndexNode = 1024;
```
wangchao316, line 78: hi , maxDegreeOfIndexNode should be Consistent as config item.  you should alter this config too. 
https://github.com/apache/iotdb/blob/4f4ee906c6a10915d86ce900ede7174ab161fbfa/server/src/assembly/resources/conf/iotdb-engine.properties#L570

wangchao316 (changes requested):

tsfile/src/main/java/org/apache/iotdb/tsfile/common/conf/TSFileConfig.java
yuqi1129: OK, Thanks for your careful review~

wangchao316 (approved):
LGTM

JackieTien97:
The change of this default config value aims to adapt to the new Tsfile format, because the new tsfile format has moved the chunk metadata list next to the timeseries metadata which locates in the leaf node of metadata tree, so we reduce the default number of maxDegreeOfIndexNode.
If you occur the error, it means that you use the master version which is not merged with NewTsfileV2 to generate the tsfile, and then use the current master version to restart with the stale data.

PR #2912, opened 2021-03-25
[ISSUE-2905] Fix Files.deleteIfExists() doesn't work for HDFS file

samperson1997:
Cluster module introduce `Files.deleteIfExists(file.toPath())` in tsfile resource, but `toPath()` method returns `java.nio.file.Path` which doesn't work for HDFS file, and result in failure of storing tsfiles in HDFS.


samperson1997:
Hi @jt2594838 @LebronAl , do you think this change is acceptable? Instead of using `Files. deleteIfExists()`, we use an if statement to judge ... so that HDFS files could work : )

HTHou (approved):

PR #2920, opened 2021-03-26
Fix unseq compaction file selector conflicts with time partition bug

zhanglingzhe0820:
## Behavior
A ci bug behave like below:
image

## Reason

The problem happens in LevelCompaction. The compaction process get all seq files to merge with unseq files in one time partition t1. And the unseq compaction may select seq files which is not in time partition t1. This operation may destroy the time partition structure.

## Solution 1
Just submit seq files in time partition t1 for unseq compaction in `LevelCompactionTsFileManagement`.
We changed the management of the `TsFileResourceList` to groups by time partition in `NoCompactionTsFileManagement`.
We deprecated the `getTsFileList()` method.

## Solution 2 (not implemented)
Make unseq compaction not select seq files exceed end time of each unseq file.
We cannot implement this by adding condition `unseqEndTime >= seqStartTime` in `selectOverlappedFile`. Reason: 
Case there are
-  a seq file file1 with time range (0-6099),
- a seq file file2 with time range (6300-6399), 
- a seq file file3 with time range (7300-7399), 
- a seq file file4 with time range (8300-8399), 
- a seq file file5 with time range (9300-9399), 
- an unseq file file6 with time range (7000-7099), 
- an unseq file file7 with time range (8000-8099), 
- an unseq file file8 with time range (9000-9099), 
- an unseq file file9 with time range (0-100). 

For unseq file file6, it does not satisfy the condition `unseqStartTime <= seqEndTime` for file1, file2.
And for file3, the unseq file file6 will satisfy the condition `unseqEndTime >= seqStartTime` which is filtered.
So we can just select only an unseq file file6. 
This also applies to the next two files.
And when we check unseq file file9, seq file file1 will be selected.
Then we just select seq file file1, unseq file file6 file7 file8 file9. When they merge, will get a file with time range (0-9099), which is wrong.


server/src/main/java/org/apache/iotdb/db/engine/merge/selector/MaxFileMergeFileSelector.java
```diff
@@ -229,10 +229,13 @@ private void selectOverlappedSeqFiles(TsFileResource unseqFile) {
         }
         // the open file's endTime is Long.MIN_VALUE, this will make the file be filtered below
         long seqEndTime = seqFile.isClosed() ? seqFile.getEndTime(deviceId) : Long.MAX_VALUE;
+        long seqStartTime = seqFile.getStartTime(deviceId);
         if (unseqEndTime <= seqEndTime) {
           // the unseqFile overlaps current seqFile
-          tmpSelectedSeqFiles.add(i);
-          tmpSelectedNum++;
+          if (seqStartTime <= seqEndTime) {
```
jixuan1989, line 235: seqStartTime<= seqEndTime?

server/src/main/java/org/apache/iotdb/db/engine/compaction/no/NoCompactionTsFileManagement.java
```diff
@@ -28,41 +28,53 @@
...
-                      Long.parseLong(o1.getTsFile().getParentFile().getName()),
-                      Long.parseLong(o2.getTsFile().getParentFile().getName()));
-              return rangeCompare == 0
-                  ? compareFileName(o1.getTsFile(), o2.getTsFile())
-                  : rangeCompare;
-            } catch (NumberFormatException e) {
-              return compareFileName(o1.getTsFile(), o2.getTsFile());
-            }
-          });
+  private final Map<Long, TreeSet<TsFileResource>> sequenceFileTreeSetMap =
```
SilverNarcissus, line 41: You use ConcurrentSkipListMap to improve the concurrency. But the value of this map is treeset, which is not thread safe. In your later code, you use synchronize in "add" method but never use synchronize in "contains method". That will cause a concurrency problem. As far as I'm concerned, I recommend to use synchronize on this complex structure. Use TreeMap and tree set and use synchronize at every read and write.

server/src/main/java/org/apache/iotdb/db/engine/compaction/no/NoCompactionTsFileManagement.java
```diff
@@ -28,41 +28,53 @@
...
+  @Deprecated
   @Override
   public List<TsFileResource> getTsFileList(boolean sequence) {
+    List<TsFileResource> result = new ArrayList<>();
     if (sequence) {
-      return new ArrayList<>(sequenceFileTreeSet);
+      synchronized (sequenceFileTreeSetMap) {
+        for (long timePartition : sequenceFileTreeSetMap.keySet()) {
+          result.addAll(getTsFileListByTimePartition(true, timePartition));
+        }
```
jt2594838, line 60: How about just iterating the `valueSet`?

server/src/main/java/org/apache/iotdb/db/engine/compaction/no/NoCompactionTsFileManagement.java
```diff
@@ -28,41 +28,53 @@
...
+    }
+    return result;
+  }
+
+  @Override
+  public List<TsFileResource> getTsFileListByTimePartition(boolean sequence, long timePartition) {
+    if (sequence) {
+      return new ArrayList<>(sequenceFileTreeSetMap.get(timePartition));
+    } else {
+      return new ArrayList<>(unSequenceFileListMap.get(timePartition));
```
jt2594838, line 77: Is it possible that this `get` returns a null? If so, better to use `getOrDefault`.

SilverNarcissus (changes requested):
Nice work! Only one concurrency problem, please have a look.

server/src/main/java/org/apache/iotdb/db/engine/compaction/no/NoCompactionTsFileManagement.java
```diff
@@ -74,27 +86,48 @@ public NoCompactionTsFileManagement(String storageGroupName, String storageGroup
...
     if (sequence) {
-      sequenceFileTreeSet.removeAll(tsFileResourceList);
+      for (Set<TsFileResource> sequenceFileTreeSet : sequenceFileTreeSetMap.values()) {
+        sequenceFileTreeSet.removeAll(tsFileResourceList);
+      }
     } else {
-      unSequenceFileList.removeAll(tsFileResourceList);
+      for (List<TsFileResource> unSequenceFileList : unSequenceFileListMap.values()) {
+        unSequenceFileList.removeAll(tsFileResourceList);
+      }
```
jt2594838, line 112: This contains potentially redundant operations, here is my advice:
1. sort `tsFileResourceList` according to their partitions
2. for each partition `p` existing in `tsFileResourceList`, find its first position `min` and last position `max` in `tsFileResourceList`
3. call `sequenceFileTreeSetMap.get(p).removeAll(tsFileResourceList.subList(min, max + 1))`

jt2594838: By the way, is not concurrent safety a concern?

server/src/main/java/org/apache/iotdb/db/engine/compaction/no/NoCompactionTsFileManagement.java
```diff
@@ -105,44 +138,61 @@ public void addRecover(TsFileResource tsFileResource, boolean sequence) {
...
     }
   }
 
   @Override
   public boolean contains(TsFileResource tsFileResource, boolean sequence) {
     if (sequence) {
-      return sequenceFileTreeSet.contains(tsFileResource);
+      return sequenceFileTreeSetMap
+          .computeIfAbsent(tsFileResource.getTimePartition(), this::newSequenceTsFileResources)
+          .contains(tsFileResource);
```
jt2594838, line 151: It may be better to use `getOrDefault(partition, Collections.emptyList())` here since it is not necessary to create entries for reads.

server/src/main/java/org/apache/iotdb/db/engine/compaction/no/NoCompactionTsFileManagement.java
```diff
@@ -159,4 +209,25 @@ public void forkCurrentFileList(long timePartition) {
...
+                ? compareFileName(o1.getTsFile(), o2.getTsFile())
+                : rangeCompare;
+          } catch (NumberFormatException e) {
+            return compareFileName(o1.getTsFile(), o2.getTsFile());
+          }
+        });
+  }
+
+  private List<TsFileResource> newUnSequenceTsFileResources(Long k) {
+    return new CopyOnWriteArrayList<>();
```
jt2594838, line 231: Now that you have used `synchronized`, is it necessary to use `CopyOnWriteArrayList` too?

jt2594838 (changes requested):

jt2594838 (approved):

PR #2938, opened 2021-03-29
[Distributed] Fix stackoverflow bug caused by InsertMultiTabletPlan or InsertRowsPlan 

LebronAl:
Suppose the cluster has just started without any data, and client perferm a InsertTablets including two Tablets with same deviceId, the cluster will try to auto-create schema for these two tablets in a `for` loop.For the first tablet, cluster will auto-create related timeseries, but for the second tablet, the `unregisteredSeriesList` is empty, and if we don't return true at this time, the cluster will create a emply createMultiTimeseriesPlan to execute, and this will cause a endless loop which finally cause stackoverflow 

image
image
image
image


mychaow (approved):
LGTM

neuyilan (approved):
image

wangchao316 (approved):
LGTM

PR #2942, opened 2021-03-29
[IOTDB-1262] The test fails to be executed independently

wangchao316:
The test fails to be executed independently

image

Reason:

This test uses an empty `queryPlan`. However, `SeriesRawDataBatchReader` may need to use the path from the queryPlan.

```
return new SeriesRawDataBatchReader(
        path,
        queryPlan.getAllMeasurementsInDevice(path.getDevice()), \\called `allSensors`
        dataType,
        context,
        queryDataSource,
        timeFilter,
        valueFilter,
        null,
        queryPlan.isAscending());
```

The current logic is: if we can find the `path` in `TimeSeriesMetadataCache`, then we do not need to use `allSensors`.  If not, we have to check `allSensors`.

When running the whole IT test, `TimeSeriesMetadataCache` is not empty and the `path` exists in it, so even though  `allSensors` is empty, it does not take any side effect. 
But when run the IT independently,  `TimeSeriesMetadataCache` is empty and then we will search the `path` in `allSensors`, which is also empty...

But for this case, it is clear that we should not generate an empty queryPlan, and this PR fixes it using this way.

More, in `public ServerTimeGenerator(IExpression expression, QueryContext context, RawDataQueryPlan queryPlan)` , we should remove `IExpression expression`, and use `queryPlan.getExpression` instead.

After change:

image

### DIFF
--- a/server/src/test/java/org/apache/iotdb/db/integration/IoTDBEngineTimeGeneratorIT.java
+++ b/server/src/test/java/org/apache/iotdb/db/integration/IoTDBEngineTimeGeneratorIT.java
@@ -198,8 +198,12 @@ public class IoTDBEngineTimeGeneratorIT {
 
     SingleSeriesExpression singleSeriesExpression =
         new SingleSeriesExpression(pd0s0, FilterFactory.and(valueGtEq, timeGt));
+    RawDataQueryPlan queryPlan = new RawDataQueryPlan();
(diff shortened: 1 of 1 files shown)


wangchao316:
@HTHou hi hannan, could you please reivew this PR?
Thanks

PR #2958, opened 2021-04-01
[IOTDB-1266]SHOW TIMESERIES will only display 2000 timeseries

yifuzhou:
## Description
In current cluster version, if we have more than 2,000 timeseries, when we commend SHOW TIMESERIES, it will only display 2,000 timeseries, the same as the SHOW DEVICES...

In the default settings, if we do not set any limit, the default plan.limit will be 1000(set as the fetchsize). This is weird, because if we do not set the limit, it means that I want to fetch all the data instead of fetchsize. So I remove it.

In my opinion, hasNextWithoutConstraint() method is not correct: after I have already fetched all the timeseries, it is not necessary to do fetch again... 
For example, 
if we have 10,000 timeseries, so the result.size() should be 10,000. 
After the first display(fetch size==1000), it will only display the first 1000 timeseries, 
now the index is 1000, so it is still index < result.size() and it will keep display the rest. 

And there is the only extreme condition we need this method, if we do not set any limit value, it means that we want to fetch all the data, but if the data size is bigger than Integer.MAX_VALUE, the exceed part cannot be displayed. Then we need to do second fetch.. 

I think it is unnecessary because the user may not want to display more than Integer.MAX_VALUE timeseries...
Please correct me if I am wrong, Thanks a lot!

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
@@ -1697,7 +1697,11 @@ public class CMManager extends MManager {
     // do not use limit and offset in sub-queries unless offset is 0, otherwise the results are
     // not combinable
     if (offset != 0) {
-      plan.setLimit(0);
+      if (limit > Integer.MAX_VALUE - offset) {
+        plan.setLimit(0);
+      } else {
+        plan.setLimit(limit + offset);
+      }
       plan.setOffset(0);
     }
 
--- a/cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
@@ -1649,6 +1649,11 @@ public class CMManager extends MManager {
     // do not use limit and offset in sub-queries unless offset is 0, otherwise the results are
     // not combinable
     if (offset != 0) {
+      if (limit > Integer.MAX_VALUE - offset) {
+        plan.setLimit(0);
+      } else {
+        plan.setLimit(limit + offset);
+      }
       plan.setOffset(0);
     }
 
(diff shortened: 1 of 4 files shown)


cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
```diff
@@ -1519,7 +1519,6 @@ public MNode getMNode(MNode deviceMNode, String measurementName) {
     // do not use limit and offset in sub-queries unless offset is 0, otherwise the results are
     // not combinable
     if (offset != 0) {
-      plan.setLimit(0);
```
jt2594838, line 1522: This should not be removed. Consider this case: 
The offset is 10000, the limit is 10, and the results are distributed on 3 nodes. If you keep the limit, each of the 3 nodes will only return 10 results, 30 results in total, which will all be filtered when the offset is applied.
You may set the limit to `limit + offset` so the unnecessary fetch can be reduced.

jt2594838 (changes requested):

jt2594838 (approved):

jixuan1989:
Hi,  no test... could you add a screenshot to show the correction?
BTW, once PR https://github.com/apache/iotdb/pull/3024 is merged, we can add IT for the cluster.

yifuzhou:
Hi all. This PR is still in progress actually. In this PR, first, if we do not set the 'limit', it will fetch as large as it can, it may cause the problem of Memory Explosion. Second, the sequence of fetch data is not in order, I think it violates the purpose of 'offset' and 'limit' field because we can find what we want by SHOW TIMESERIES OFFSET ? LIMIT ?.
I will fix it later and  after the IT for cluster merged, I will add the test case of this part, thanks!

qiaojialin:
Hi, to make the offset and limit meaningful in show timeseries, we need to manage timeseries in some order in the MTree firstly.
Besides, I remember that hasNextWithoutConstraint() does not consider the offset and limit. Only hasNext considers the offset and limit.

LebronAl:
> Hi all. This PR is still in progress actually. In this PR, first, if we do not set the 'limit', it will fetch as large as it can, it may cause the problem of Memory Explosion. Second, the sequence of fetch data is not in order, I think it violates the purpose of 'offset' and 'limit' field because we can find what we want by SHOW TIMESERIES OFFSET ? LIMIT ?.
> I will fix it later and after the IT for cluster merged, I will add the test case of this part, thanks!

ping, you can add IT tests now~

PR #2959, opened 2021-04-01
fix group by month bug

haimeiguo:
"select first_value(s0) from root.xinsaike.wf01.wt01 group by ([1609430400000,1640967000000), 10m, 1mo)"
image


Alima777 (commented):
1. I think creating a new sub-class `GroupByMonthFilter` which extends `GroupByFilter` and rewrite it's `satisfy` and other methods will be better.
2. Since the `safisfy` method will be executed for every single data point, so if it has to be calculated by Calendar every time, I am afraid maybe there is a lot performance loss.

Alima777:
I've submitted a new PR, please have a review :D

PR #2964, opened 2021-04-01
Fix group by month bug

Alima777:
The condition is: the sliding step and time interval whose unit is month has to be increased using natural month but not 30 days.


haimeiguo (commented):
looks good to me!

JackieTien97 (approved):

PR #2972, opened 2021-04-06
[DRAFT][IOTDB-1281] Tsfile check and restore tool

zhanglingzhe0820:


server/src/assembly/resources/tools/tsfileToolSet/check-tsfile-chunk.bat
```diff
@@ -0,0 +1,62 @@
...
+
+
+@echo off
+echo ````````````````````````
+echo Starting Checking the TsFile Chunk
+echo ````````````````````````
+
+if "%OS%" == "Windows_NT" setlocal
+
+pushd %~dp0..\..
```
wangchao316, line 28: hi , Good Job.
I have a little...    if you can add some doc of tool using,  this will be perfect.

Issue #2919, opened 2021-03-26
C++ client 'class SessionDataSet' will memory leak


tanhangbo:
通过valgrind检测，发现这里存在一些内存泄露，会导致时间久了爆内存

```
==20709== 576 bytes in 8 blocks are definitely lost in loss record 47 of 49
==20709==    at 0x4C3217F: operator new(unsigned long) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==20709==    by 0x4F36B12: SessionDataSet::constructOneRow() (in /home/tan/iotdb/client-cpp-0.12.0-SNAPSHOT-cpp-linux-x86_64/lib/libiotdb_session.so)
==20709==    by 0x4F3676C: SessionDataSet::hasNext() (in /home/tan/iotdb/client-cpp-0.12.0-SNAPSHOT-cpp-linux-x86_64/lib/libiotdb_session.so)
==20709==    by 0x120739: query1() (in /home/tan/iotdb/client-cpp-0.12.0-SNAPSHOT-cpp-linux-x86_64/a.out)
==20709==    by 0x123B51: main (in /home/tan/iotdb/client-cpp-0.12.0-SNAPSHOT-cpp-linux-x86_64/a.out)
==20709==


==20709== 1,218 (376 direct, 842 indirect) bytes in 1 blocks are definitely lost in loss record 48 of 49
==20709==    at 0x4C3217F: operator new(unsigned long) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==20709==    by 0x4F3CC71: Session::executeQueryStatement(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) (in /home/tan/iotdb/client-cpp-0.12.0-SNAPSHOT-cpp-linux-x86_64/lib/libiotdb_session.so)
==20709==    by 0x120610: query1() (in /home/tan/iotdb/client-cpp-0.12.0-SNAPSHOT-cpp-linux-x86_64/a.out)
==20709==    by 0x123B51: main (in /home/tan/iotdb/client-cpp-0.12.0-SNAPSHOT-cpp-linux-x86_64/a.out)


==20709== 3,535 (376 direct, 3,159 indirect) bytes in 1 blocks are definitely lost in loss record 49 of 49
==20709==    at 0x4C3217F: operator new(unsigned long) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
[...]

PR #2974, opened 2021-04-07
[IOTDB-1282] fix C++ class SessionDataSet mem-leak

jamber001:
[IOTDB-1282] fix C++ class SessionDataSet mem-leak


jixuan1989:
https://github.com/apache/iotdb/issues/2919

PR #2987, opened 2021-04-09
optimize sync leader for meta

mychaow:
## Description

In the distributed version, metadata synchronization will be performed when parsing SQL statements, which will lead to parsing time consuming up to 1 second. The main reason is that the heartbeat of the master is not received in time, which leads to the delay of the synchronization time. The root cause is that metadata and data belong to the same raft replication group, which affects the judgment of whether metadata is consistent or not.

The main idea of this pr is to reduce the time of synchronization point by getting the submitted information earlier. With weak consistency, we can get 20~30% query performance improvement.

### Content3 ...

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
@@ -47,6 +47,7 @@ import org.apache.iotdb.cluster.rpc.thrift.HeartBeatResponse;
 import org.apache.iotdb.cluster.rpc.thrift.Node;
 import org.apache.iotdb.cluster.rpc.thrift.RaftService.AsyncClient;
 import org.apache.iotdb.cluster.rpc.thrift.RaftService.Client;
+import org.apache.iotdb.cluster.rpc.thrift.RequestCommitIndexResponse;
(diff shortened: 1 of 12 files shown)


cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
```diff
@@ -439,6 +417,31 @@ public HeartBeatResponse processHeartbeatRequest(HeartBeatRequest request) {
     }
   }
 
+  private void tryUpdateCommitIndex(long leaderTerm, long commitIndex, long commitTerm) {
+    if (term.get() == leaderTerm && logManager.getCommitLogIndex() < commitIndex) {
```
neuyilan, line 421:
```suggestion
    if (leaderTerm >= term.get() && logManager.getCommitLogIndex() < commitIndex) {
```

neuyilan: Maybe, we can commit the log as long as the leader's term larger than the follower's term?

neuyilan (approved):
Good Job!

PR #3001, opened 2021-04-13
[Distributed] minor improve the concurrency problem caused by the checking apply thread and the compact entries thread

neuyilan:
image

There are so many checks apply thread warnings, even though, we have handled the issue, it is better to reduce it.
Reduce the concurrency problem caused by the check apply thread and the compact entries thread.


mychaow (approved):
LGTM

wangchao316 (approved):

LebronAl (approved):
LGTM

PR #3005, opened 2021-04-15
[Distributed]Lowering the logging level from debug to trace when serializing meta 

neuyilan:
The debug log outputs some useless log when serializing meta, which make it harder to find the usage info when debugging the code, so I think we can lower the debug level to trace.

image


wangchao316 (approved):
LGTM

mychaow (approved):
LGTM

PR #3006, opened 2021-04-15
[IOTDB-1301] snappy compatible arm64

wangchao316:
Snappy must be compatible with the ARM server. 1.8.0 and later versions are compatible with ARM.


wangchao316:
@jixuan1989 @LebronAl , could you please review? 
Thanks

qiaojialin (approved):

PR #3007, opened 2021-04-15
[Distributed] fix the plan index miss 

neuyilan:
**[phenomenon]**
The plan index is always zero when we use insertRecords interface to test the cluster.
image

**[After resloved]**
After the bug was resolved, we can see that the plan index is not zero.
image

 **[Reason]**
the plan index is not set in the sub plans when we split the InsertRowsPlan, InertMultiTabletPlans and InsertRowsOfOneDevicePlan


LebronAl (commented):
Why call `setIndex` until the log is serialized? What if the coordinator node executes log locally? Maybe we should overload the `setIndex` function in `PhysicalPlan` for these plans so that the raft index will be set successfully at the beginning.

neuyilan:
> Why call `setIndex` until the log is serialized? What if the coordinator node executes log locally? Maybe we should overload the `setIndex` function in `PhysicalPlan` for these plans so that the raft index will be set successfully at the beginning.

Of course, thanks for your reivew.

LebronAl (approved):
LGTM

PR #3008, opened 2021-04-15
[REL/0.12, Distributed] lowering the logging level from debug to trace when serialize meta

neuyilan:
https://github.com/apache/iotdb/pull/3005


LebronAl (approved):
LGTM

PR #3010, opened 2021-04-15
[IOTDB-1306] Fix DeadLock in MemControl module

HTHou:
In current version, the insert thread will sleep with the sg write lock until the flush thread release some memory and the total memtable memory size is lower than the reject threshold in SystemInfo. 

But if there is a flush thread causes another memtable of this sg flushing, it would require the write lock that holden by the insert thread. 

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
@@ -1102,8 +1102,8 @@ public class StorageGroupProcessor {
           && !closingUnSequenceTsFileProcessor.contains(tsFileProcessor)) {
         fileFlushPolicy.apply(this, tsFileProcessor, tsFileProcessor.isSequence());
       }
-    } finally {
       rejectConditionSignal();
+    } finally {
       writeUnlock();
     }
   }
--- a/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
@@ -1104,6 +1104,7 @@ public class StorageGroupProcessor {
       }
     } finally {
       writeUnlock();
+      rejectConditionSignal();
     }
   }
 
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -979,6 +983,7 @@ private boolean insertTabletToTsFileProcessor(
 
     try {
       tsFileProcessor.insertTablet(insertTabletPlan, start, end, results);
+      writeLock();
```
mychaow, line 986: Why do you add writeLock here? We have get write lock before call the function.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileProcessor.java
```diff
@@ -387,14 +385,15 @@ private void updateMemoryInfo(
...
-        StorageEngine.blockInsertionIfReject();
-      } catch (WriteProcessRejectException e) {
-        storageGroupInfo.releaseStorageGroupMemCost(memTableIncrement);
-        tsFileProcessorInfo.releaseTSPMemCost(unsealedResourceIncrement + chunkMetadataIncrement);
-        SystemInfo.getInstance().resetStorageGroupStatus(storageGroupInfo, false);
-        throw e;
+      long startTime = System.currentTimeMillis();
+      while (SystemInfo.getInstance().isRejected()) {
+        try {
+          storageGroupInfo.getStorageGroupProcessor().rejectConditionAwait();
```
mychaow, line 391: So, if we wait to timeout, we could not return now?

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
HTHou: Yes, it is a mistake......

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileProcessor.java
HTHou: Yes, I remove the In-WriteLock timeout exception. 
There are still the out-writeLock timeout exception in there. How do you think?

HTHou: Ok, I add the timeout back....

JackieTien97 (approved):

PR #3013, opened 2021-04-15
[IOTDB-1294] Compaction mods for new mods structure

zhanglingzhe0820:
## Behavior
Delete operation become invalid after compaction.

## Reason
1. In `append merge`, the mods modify of chunkMetadata will not change the data of the chunk corresponding to it. However, we just merge the chunk data by BufferBytes and then delete the modification record.
2. As we change the structure of mods file in 0.12 (delete `version` and change it to `tsfileOffset` which means that only if the offset of the chunk <= `tsfileOffset`, then the modification record is valid). However, the position of chunk in tsfile may change after compaction.

## Solution
1. Do not use mods file in `append merge` and keep the modification record.
2. Update the `tsfileOffset` of remain modification record to Long.MAX_VALUE, which means that it will be valid for the whole tsfile.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/compaction/TsFileManagement.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/compaction/TsFileManagement.java
@@ -346,6 +346,9 @@ public abstract class TsFileManagement {
       seqFile.removeModFile();
       if (mergingModification != null) {
         for (Modification modification : mergingModification.getModifications()) {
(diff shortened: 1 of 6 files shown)


HTHou (approved):
Don't forget to cherry pick to rel/0.12.

SilverNarcissus (approved):
Look good to me~

zhanglingzhe0820:
cherry pick to 0.12 by [PR](https://github.com/apache/iotdb/pull/3046)

PR #3024, opened 2021-04-17
[IOTDB-1310] Enable docker, docker-compose and testcontainer for End to end test

jixuan1989:
This PR fixes:

1. in the cluster module, set the default number of nodes as 1, and the default replica number as 1.

2. users can skip to set the value of `internal_ip`. If `internal_ip` is commented, then the hostname of the server will be used rather than `127.0.0.1` to be used.

3. add a new module `testcontainer`. The module will be added if your PC/Mac has installed Docker.

4. the `testcontainer` module will build two docker images: `apache/iotdb:maven-development` and `apache/iotdb:cluster-maven-development` when running `mvn pre-integration-test`

5. Two E2E test (end with *IT.java) with TestContainer framework examples are provided, in `testcontainer/src/test/java/`, while one is for single node and another is for the cluster with 3 nodes.

6. remove duplicated `verify` phash in `integration-test` phase.

7. To disable the `testcontainer` module, use `-P !testcontainer` in your `mvn` command.

8. In this PR, `spotless:check` is moved to a profile, which is enabled by default. To skip the step (for faster compiling, use `-P !spotless`).


cluster/src/main/java/org/apache/iotdb/cluster/ClusterMain.java
```diff
@@ -117,7 +117,7 @@ public static void main(String[] args) {
     } else if (MODE_ADD.equals(mode)) {
       try {
         metaServer = new MetaClusterServer();
-        // preStartCustomize();
+        preStartCustomize();
```
mychaow, line 120: preStratCustomize is just for debug, should not used in deployment.

icloudsong (approved):

testcontainer/Readme.md
```diff
@@ -0,0 +1,66 @@
...
+
+You have to install Docker before you runn this module.
+See [Docker Version Requirement](https://www.testcontainers.org/supported_docker_environment/).
+
+IoTDB project will detect whether a Docker is installed (but do not check the docker's version).
+
+The logic is, for Unix-like system, it checks whether `/var/run/docker.sock` exists.
+For Window system, it checks whether `C:\Program Files\Docker\Docker\resources\bin\docker.exe` exists.
+
+If you are sure you have installed the Docker but `testcontainer` module is not loaed, use `-P unixDockerCheck`
```
icloudsong, line 36: Could "-P unixDockerCheck" be used in Windows environment check?

icloudsong (approved):

cluster/src/main/java/org/apache/iotdb/cluster/ClusterMain.java
```diff
@@ -157,6 +157,7 @@ private static void startServerCheck() throws StartupException {
               config.getSeedNodeUrls().size(), quorum);
       throw new StartupException(metaServer.getMember().getName(), message);
     }
+    logger.info("seed urls in start Server Check 3:" + config.getSeedNodeUrls());
```
neuyilan, line 160: What this log is used for?

cluster/src/main/java/org/apache/iotdb/cluster/ClusterMain.java
```diff
@@ -170,7 +171,7 @@ private static void startServerCheck() throws StartupException {
       seedNodes.add(node);
     }
 
-    // assert this node is in all nodes when restart
+    // assert this node is in all nodes when restart 172.19.0.2:9003, 172.19.0.3:9003,
```
neuyilan, line 174: Why add the magic IPs?

cluster/src/assembly/resources/conf/iotdb-cluster.properties
```diff
@@ -51,7 +52,8 @@ internal_data_port=40010
 # nodes that already in the cluster, unnecessary to be the nodes that were used to build the
 # initial cluster by start-node.sh(.bat). Several nodes will be picked randomly to send the
 # request, the number of nodes picked depends on the number of retries.
-seed_nodes=127.0.0.1:9003,127.0.0.1:9005,127.0.0.1:9007
+#seed_nodes=127.0.0.1:9003,127.0.0.1:9005,127.0.0.1:9007
+seed_nodes=127.0.0.1:9003
```
neuyilan, line 56: Is it unfriendly for users to configure only one IP in distributed mode?
Why change the default parameter?

neuyilan (changes requested):

testcontainer/Readme.md
jixuan1989: yes.

cluster/src/main/java/org/apache/iotdb/cluster/ClusterMain.java
jixuan1989: It is a very very strange error... I will remove it after double checking.

cluster/src/main/java/org/apache/iotdb/cluster/ClusterMain.java
jixuan1989: my fault.

cluster/src/assembly/resources/conf/iotdb-cluster.properties
jixuan1989: Actually THIS is  "friendly".

What we want:
1. let user download the zip (iotdb-cluster.zip)
2. unzip the file
3. call `start-node.sh`
4. begin to have a try.

If we set 3 ips here, then the process is.... very very difficult for users:
1. let user download the zip (iotdb-cluster.zip)
2. unzip the file
3. copy the file into 3 parts
4. modiy `iotdb-cluster.properties` one by one (very hard, as they have to read the file to know "ah, in the first file, I need modify the parameter as 9003, and modify the rpc port and internal_data port as .... In the second file, I need ...."
5. call `start-node.sh`
6. begin to have a try.

It is a disaster for fresh man.

If you see other distributed systems, all their default setting is just one node.

jixuan1989:
The code coverage is hard to be promoted anymore 
(I do not know how to simulate exception when calling InetAddress.getLocalHost()).

cluster/src/assembly/resources/conf/iotdb-cluster.properties
neuyilan: Sure, your consideration is reasonable.

neuyilan (approved):

mychaow (approved):
LGTM

PR #3029, opened 2021-04-19
[IOTDB-1317] Log CatchUp always failed du to not check the follower's match index

neuyilan:
https://issues.apache.org/jira/browse/IOTDB-1317

When the follower lags behind the leader, it will catch up with the leader through the log. When it can't find the logs to catch up with in the memory of the leader, it will go to the file on the local disk to find the satisfied logs. The logs found are between the old match index and the commit index saved by the leader. However, at this time, the match index of the follower may have been updated and the old match index of the follower may have been deleted (in order to prevent a large number of logs from being kept on disk, there is a log deletion mechanism). So at this time, if you send the above logs to the follower, the follower will fail to find the matching log. At this time, this storage group will appear to write stuck phenomenon.

The solution is to re-check the match index of the follower after reading the log from the disk.


mychaow (approved):
LGTM

wangchao316 (approved):
LGTM

PR #3040, opened 2021-04-21
[To rel/0.12][Distributed] update cluster-setup-example docs in config.js

LebronAl:


HTHou:
I think we don't need to update this in rel/0.12. Updating master branch is enough.

PR #3041, opened 2021-04-21
[To rel/0.12][IOTDB-1323] IoTDB will return a success message when it encounters RuntimeException during the insertion process

SteveYurongSu:
JIRA: https://issues.apache.org/jira/browse/IOTDB-1323


qiaojialin (approved):
create an issue

PR #3044, opened 2021-04-22
Fix CI failed caused by updating unnecessary vector last cache

HTHou:
## Description

The InsertRowPlanTest failed due to the data type mismatching when update the last cache for an aligned timeseries.

Eventually, we don't need to update the last cache for aligned timeseries. Therefore, I add a check to fix it.

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
@@ -73,6 +73,7 @@ import org.apache.iotdb.rpc.RpcUtils;
 import org.apache.iotdb.rpc.TSStatusCode;
 import org.apache.iotdb.service.rpc.thrift.TSStatus;
 import org.apache.iotdb.tsfile.file.metadata.ChunkMetadata;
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -1028,24 +1029,32 @@ private void tryToUpdateBatchInsertLastCache(InsertTabletPlan plan, Long latestF
...
       } else {
-        // measurementMNodes[i] is null, use the path to update remote cache
-        IoTDB.metaManager.updateLastCache(
-            plan.getDeviceId().concatNode(plan.getMeasurements()[i]),
-            plan.composeLastTimeValuePair(i),
-            true,
-            latestFlushedTime,
-            null);
+        if (plan.getColumns()[i] == null) {
+          columnIndex++;
```
SteveYurongSu, line 1039: It is better to take out `columnIndex++` from the if-else.

SteveYurongSu (approved):
LGTM

PR #3050, opened 2021-04-23
[To rel/0.11] use memory control module to remove old partition

SilverNarcissus:
Use memory control module to remove old partition


jixuan1989:
Is `IoTDBDescriptor.getInstance().getConfig().getConcurrentWritingTimePartition()` used in other places?

SilverNarcissus:
I have removed this parameter from doc and commented it as deprecated in config file

PR #3052, opened 2021-04-23
[To rel/0.12] make version change easy

mychaow:
## Description

the pr solve two problems:

1. make default value of config file consistent with config in code.
2. make version change easy, only need to change the iotdb-parent version when we want to change project version. Some explain is in https://segmentfault.com/a/1190000019280804.

### Content3 ...

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/cluster/src/assembly/resources/conf/iotdb-cluster.properties
+++ b/cluster/src/assembly/resources/conf/iotdb-cluster.properties
@@ -86,7 +86,7 @@ default_replica_num=3
 # so the leader will send logs(snapshot) to the follower,
 # NOTICE, it may cost minutes of time to send a snapshot,
 # so this parameter should be larger than the snapshot cost time.
-# catch_up_timeout_ms=600000
+# catch_up_timeout_ms=300000
 
 # whether to use batch append entries in log catch up
 # use_batch_in_catch_up=true
--- a/cluster/src/assembly/resources/conf/iotdb-cluster.properties
+++ b/cluster/src/assembly/resources/conf/iotdb-cluster.properties
@@ -86,7 +86,7 @@ default_replica_num=3
 # so the leader will send logs(snapshot) to the follower,
 # NOTICE, it may cost minutes of time to send a snapshot,
 # so this parameter should be larger than the snapshot cost time.
-# catch_up_timeout_ms=300000
+# catch_up_timeout_ms=600000
 
 # whether to use batch append entries in log catch up
 # use_batch_in_catch_up=true
--- a/cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterConfig.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterConfig.java
@@ -60,10 +60,10 @@ public class ClusterConfig {
   private boolean useBatchInLogCatchUp = true;
 
   /** max number of committed logs to be saved */
-  private int minNumOfLogsInMem = 100;
+  private int minNumOfLogsInMem = 1000;
 
   /** max number of committed logs in memory */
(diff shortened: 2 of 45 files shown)


server/src/assembly/resources/conf/iotdb-engine.properties
```diff
@@ -43,7 +43,7 @@ rpc_port=6667
 # rpc_max_concurrent_client_num=65535
 
 # thrift max frame size, 64MB by default
```
wangchao316, line 45: hi chao, comments need alter.

service-rpc/src/main/java/org/apache/iotdb/rpc/RpcUtils.java
```diff
@@ -38,7 +38,7 @@
 public class RpcUtils {
 
   /** How big should the default read and write buffers be? Defaults to 64KB */
```
wangchao316, line 40: the same...

server/src/assembly/resources/conf/iotdb-engine.properties
mychaow: fixed.

cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterConfig.java
```diff
@@ -60,10 +60,10 @@
   private boolean useBatchInLogCatchUp = true;
```
neuyilan, line 60: Hi, the `catchUpTimeoutMS` is not consistent with the iotdb-cluster.properites, in the ClusterConfig.java, the value is 60 seconds, while in the iotdb-cluster.properites, the value is 600 seconds, may be we can change to 300 seconds as the snapshot may take a long time, how do you think?

cluster/src/assembly/resources/conf/iotdb-cluster.properties
```diff
@@ -86,7 +86,7 @@ default_replica_num=3
 # so the leader will send logs(snapshot) to the follower,
 # NOTICE, it may cost minutes of time to send a snapshot,
 # so this parameter should be larger than the snapshot cost time.
-# catch_up_timeout_ms=300000
+# catch_up_timeout_ms=600000
```
neuyilan, line 89: Hi, the catchUpTimeoutMS is not consistent with the iotdb-cluster.properites, in the ClusterConfig.java, the value is 60 seconds, while in the iotdb-cluster.properites, the value is 600 seconds, may be we can change to 300 seconds as the snapshot may take a long time, how do you think?

mychaow: ok

wangchao316 (approved):
LGTM

neuyilan (approved):
LGTM

jixuan1989:
about the version management: I can not evaluate the impaction... 
When releasing a new version, we use: 
`mvn release:prepare -DautoVersionSubmodules=true -P site -P compile-cpp`  
and 
`mvn release:perform  -P site -P compile-cpp`

How the command will be after using this PR?

jixuan1989:
current `mvn release:prepare` will change the version of sub modules from ${relativeversion} to a specific version...
So, unless we change the plugins in `release:prepare`, the solution of ${relativeversion} does not take effect..

mychaow:
> current `mvn release:prepare` will change the version of sub modules from ${relativeversion} to a specific version...
> So, unless we change the plugins in `release:prepare`, the solution of ${relativeversion} does not take effect..

ok, I will revert the change.

PR #3073, opened 2021-04-27
Fixed openid bug

CloudWise-Lukemiao:
Fix the exception of starting report when the configuration openid_url is HTTP and using keyclock to login without password abnormal bug

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/auth/authorizer/OpenIdAuthorizer.java
+++ b/server/src/main/java/org/apache/iotdb/db/auth/authorizer/OpenIdAuthorizer.java
@@ -99,7 +99,7 @@ public class OpenIdAuthorizer extends BasicAuthorizer {
     logger.debug("Using Provider Metadata: {}", providerMetadata);
 
     try {
-      URL url = new URI(providerMetadata.getJWKSetURI().toString().replace("http", "http")).toURL();
+      URL url = new URI(providerMetadata.getJWKSetURI().toString()).toURL();
       logger.debug("Using url {}", url);
       return getProviderRSAJWK(url.openStream());
     } catch (IOException e) {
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/apache/iotdb/db/auth/authorizer/OpenIdAuthorizer.java
```diff
@@ -99,8 +99,7 @@ private static JSONObject getJWKFromProvider(String providerUrl)
     logger.debug("Using Provider Metadata: {}", providerMetadata);
 
     try {
-      URL url =
-          new URI(providerMetadata.getJWKSetURI().toString().replace("http", "https")).toURL();
+      URL url = new URI(providerMetadata.getJWKSetURI().toString().replace("http", "http")).toURL();
```
JulianFeinauer, line 102: Why do we generally need the replace here?

server/src/main/java/org/apache/iotdb/db/auth/authorizer/OpenIdAuthorizer.java
```diff
@@ -189,7 +188,7 @@ private Claims validateToken(String token) {
         .setAllowedClockSkewSeconds(Long.MAX_VALUE / 1000)
         // .setSigningKey(DatatypeConverter.parseBase64Binary(secret))
         .setSigningKey(providerKey)
-        .parseClaimsJws(token)
+        .parseClaimsJws(token.replaceAll("access_token:", ""))
```
JulianFeinauer, line 191: What does this line do?

CloudWise-Lukemiao: I'm sorry, I take the return value( eg:{"access_token":xxx, "token_type": "Bearer","not-before-policy": 0, "session_state": "xxx","scope": "profile email","expires_in": 300,"refresh_expires_in": 1800,"refresh_token": ""} )as the token.

CloudWise-Lukemiao: I'm sorry, I take the return value( eg:{"access_token":xxx, "token_type": "Bearer","not-before-policy": 0, "session_state": "xxx","scope": "profile email","expires_in": 300,"refresh_expires_in": 1800,"refresh_token": ""} )as the token.

JulianFeinauer (approved):

JulianFeinauer:
LGTM. Is the failing test systematic or should we care about it @jixuan1989 or @qiaojialin ? Otherwise this could be merged IMHO.

PR #3087, opened 2021-04-28
[Cluster] Fix query multiple subSensors for multiple device timeout bug in cluster

LebronAl:
This bug is because we don't set FullPath for MeasurementMnode to distinguish them. 
The logic here is new since the aligned timeseries pr were merged and may require further refactoring later


wangchao316:
> refactoring

Good , this need refactor.

wangchao316:
> refactoring

Good , this need refactor.

HTHou (approved):

wangchao316 (approved):
LGTM

PR #3095, opened 2021-04-29
[IOTDB-1341]Print too many useless logs when snapshot occurred loading one tsfile

neuyilan:
https://issues.apache.org/jira/browse/IOTDB-1341


mychaow (approved):
LGTM

PR #3100, opened 2021-04-29
Add TsFileSequenceReader getChunkMetadataList method return empty if path not exists

zhanglingzhe0820:
## Behavior
Unseq compaction failed if some path not in the file, throw error like below.
4111617881357_ pic_hd

## Reason
The getChunkMetadataList now throw Exception if there is not the path in the file.

## Solution
Add a method getChunkMetadataListReturnEmptyIfPathNotExists and use it.


HTHou:
Please don't forget to cherry pick to rel/0.12....

samperson1997 (approved):

qiaojialin:
add a parameter to getChunkMetadataList

PR #3102, opened 2021-04-29
[To rel/0.12][IOTDB-1212] The given error message is not right when executing `select sin(non_existence) from root.sg1.d1`

SteveYurongSu:
JIRA: https://issues.apache.org/jira/browse/IOTDB-1212

Let's ignore the eval of udf which has at least one nonexistant series as input.


mychaow (approved):
There are some ut/it failed.

qiaojialin (approved):

PR #3105, opened 2021-04-29
[IoTDB-1331] Add transformPath for CMManger

sunjincheng121:
Fix the following bug:
```
IoTDB> select avg(temperature) from root.ln.wf01.wt01;
Msg: 411: Error occurred in query process: Path [root.ln.wf01.wt01.temperature] does not exist
```
In this PR only add overide the method `transformPath` in CMManger.

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
@@ -81,6 +81,7 @@ import org.apache.iotdb.tsfile.file.metadata.enums.TSEncoding;
 import org.apache.iotdb.tsfile.read.TimeValuePair;
 import org.apache.iotdb.tsfile.utils.Pair;
 import org.apache.iotdb.tsfile.write.schema.IMeasurementSchema;
+import org.apache.iotdb.tsfile.write.schema.MeasurementSchema;
 import org.apache.iotdb.tsfile.write.schema.TimeseriesSchema;
 import org.apache.iotdb.tsfile.write.schema.VectorMeasurementSchema;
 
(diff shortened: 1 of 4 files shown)


irvine0109:
Looks good to me.

irvine0109 (approved):
Looks good to me.

jixuan1989:
hi, add a test in testcontainer?

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
```diff
@@ -265,6 +266,20 @@ public IMeasurementSchema getSeriesSchema(PartialPath fullPath) throws MetadataE
     return super.getSeriesSchema(fullPath, getMeasurementMNode(fullPath));
   }
 
+  /**
+   * Transform the PartialPath to VectorPartialPath if it is a sub sensor of one vector. otherwise,
+   * we don't change it.
+   */
+  @Override
+  public PartialPath transformPath(PartialPath partialPath) throws MetadataException {
+    MeasurementMNode node = getMeasurementMNode(partialPath);
```
jixuan1989, line 275: I think `getMeasurementMNode`  should be consist with `getNodeByPath`... 
However, this is a problem carried over from the past..

irvine0109: > I think `getMeasurementMNode` should be consist with `getNodeByPath`...
To be consistent, do you mean that method `MManager.getNodeByPath()` should have the function of getting schema both locally and remotely? Or method `CMManager.getMeasurementMNode()`should only have the function of getting schema locally?

jixuan1989: The diff of`transformPath`  between CMManager and MManager is the `getMeasurementMNode`.
Therefore, I think a better way is to re-write  `getMeasurementMNode` in MManager, which just retrieves the Node locally, rather than to use  `getNodeByPath ` in MManager's `transformPath` and  use `getMeasurementMNode` in CMManger.

sunjincheng121:
> hi, add a test in testcontainer?
Nice, reminder... 
I have add the test case both in ClusterIT and SingleNodeIT. 
NOTE: will add a little bit improvement for Cluster and SingleNode test in https://issues.apache.org/jira/browse/IOTDB-1351

jixuan1989 (approved):

PR #3114, opened 2021-04-30
[IOTDB-1349]Fix reduce frequent log printing.

wangchao316:
The report log  is frequently printed. As a result, the log is updated quickly. Therefore, the log level is changed to debug.

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
@@ -458,9 +458,9 @@ public class MetaGroupMember extends RaftMember {
 
   private void generateNodeReport() {
     try {
-      if (logger.isInfoEnabled()) {
+      if (logger.isDebugEnabled()) {
         NodeReport report = genNodeReport();
-        logger.info(report.toString());
+        logger.debug(report.toString());
       }
     } catch (Exception e) {
       logger.error("{} exception occurred when generating node report", name, e);


neuyilan (commented):
Hi， the log is reported every 10 seconds, and I think it's useful to see the cluster status like the raft groups and the log incremental, I'd like it as info level, What your opition？

wangchao316:
> Hi， the log is reported every 10 seconds, and I think it's useful to see the cluster status like the raft groups and the log incremental, I'd like it as info level, What your opition？
hi , thanks houliang.
I feel like 10s are more frequent, and there are two ways to alleviate it.
1. Parameters can be added to set the print frequency. Hot modification is supported. If the cluster is unstable, you can set the frequency to a high value. In common scenarios, you can set the frequency to a low value.
2. Change the log level to debug.
do you feel How about option 1?

neuyilan:
> > Hi， the log is reported every 10 seconds, and I think it's useful to see the cluster status like the raft groups and the log incremental, I'd like it as info level, What your opition？
> 
> hi , thanks houliang.
> I feel like 10s are more frequent, and there are two ways to alleviate it.
> 
> 1. Parameters can be added to set the print frequency. Hot modification is supported. If the cluster is unstable, you can set the frequency to a high value. In common scenarios, you can set the frequency to a low value.
> 2. Change the log level to debug.
>    do you feel How about option 1?

Thanks for your remind, It suddenly occurred to me that the logback has support that when 60 seconds rescan the log configuration, I think changing the log level to debug maybe is enough.😁

wangchao316:
> > > Hi， the log is reported every 10 seconds, and I think it's useful to see the cluster status like the raft groups and the log incremental, I'd like it as info level, What your opition？
> > 
> > 
> > hi , thanks houliang.
> > I feel like 10s are more frequent, and there are two ways to alleviate it.
> > 
> > 1. Parameters can be added to set the print frequency. Hot modification is supported. If the cluster is unstable, you can set the frequency to a high value. In common scenarios, you can set the frequency to a low value.
> > 2. Change the log level to debug.
> >    do you feel How about option 1?
> 
> Thanks for your remind, It suddenly occurred to me that the logback has support that when 60 seconds rescan the log configuration, I think changing the log level to debug maybe is enough.😁

yes 

> > > Hi， the log is reported every 10 seconds, and I think it's useful to see the cluster status like the raft groups and the log incremental, I'd like it as info level, What your opition？
> > 
> > 
> > hi , thanks houliang.
> > I feel like 10s are more frequent, and there are two ways to alleviate it.
> > 
> > 1. Parameters can be added to set the print frequency. Hot modification is supported. If the cluster is unstable, you can set the frequency to a high value. In common scenarios, you can set the frequency to a low value.
> > 2. Change the log level to debug.
> >    do you feel How about option 1?
> 
> Thanks for your remind, It suddenly occurred to me that the logback has support that when 60 seconds rescan the log configuration, I think changing the log level to debug maybe is enough.😁, when we encountered some problems with the cluster, we may change the info level to debug is enough.

yes, we alter log level when we need

mychaow:
I think it's better to set the frequency, because the problem didn't happen when you found it, but happened before. If there is a more accurate fault time point, it will help us locate the problem faster.

jixuan1989:
As I want to release v0.12.1. how about accepting the log level change first?
If there is good solution to change the frequency, let's do it.

mychaow:
> As I want to release v0.12.1. how about accepting the log level change first?
> If there is good solution to change the frequency, let's do it.

OK

mychaow (approved):
LGTM

PR #3120, opened 2021-04-30
[IOTDB-1350]FIx log Compression time

wangchao316:
Logs are compressed every hour. There are many compressed packages in a day, and the size of each compressed package is small. Therefore, logs are compressed by day and size.


wangchao316:
@neuyilan @mychaow @qiaojialin @jixuan1989   hi , could you please review?
Thanks.

PR #3126, opened 2021-05-03
[IOTDB-1354] fix errors in SlotPartitionTableTest

jixuan1989:
When running the Tests using `mvn test`, SlotPartitionTableTest will be skipped:
```
[INFO] Running org.apache.iotdb.cluster.partition.SlotPartitionTableTest
[WARNING] Tests run: 1, Failures: 0, Errors: 0, Skipped: 1, Time elapsed: 0.014 s - in org.apache.iotdb.cluster.partition.SlotPartitionTableTest
```
I do not know why, but that is why we never find this failed test...

When running the test in IDEA, all tests will fail, because the snapets:

```       
 for (int i = 0; i < 20; i++) {
          mManager[i] = MManagerWhiteBox.newMManager("target/schemas/mlog_" + i);
          initMockMManager(i, mManager[i], storageNames, nodeSGs[i]);
          Whitebox.setInternalState(tables[i], "mManager", mManager[i]);
        }
```

`assertGetHeaderGroup` has another error:
```
    assertEquals(
        new Node(
            "localhost", 30000 + last, last, 40000 + last, Constants.RPC_PORT + start, "localhost"),
        group.get(replica_size - 1));
```
It is obvious that the PRC_PORT should be `Constants.RPC_PORT + last`

### DIFF
--- a/cluster/src/test/java/org/apache/iotdb/cluster/partition/SlotPartitionTableTest.java
+++ b/cluster/src/test/java/org/apache/iotdb/cluster/partition/SlotPartitionTableTest.java
@@ -63,7 +63,6 @@ import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Ignore;
 import org.junit.Test;
-import org.mockito.internal.util.reflection.Whitebox;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


LebronAl (changes requested):
The whole test will be skipped too as you don't remove `@Ignore`.
BTW, as `metaManager` has become a static variable, which is quite different from when we added this test, we may need to re-implement this test now

neuyilan:
Ping, any updates?

PR #3127, opened 2021-05-04
[IOTDB-1316] Fix input format check

kydxh:
## Fix input format check for import csv files
   This is used to fix the problem show in issue 1316. The new code will check the input format before import the data. If there is wrong data format, this part will be null. And the other data can import successfully.
  
  Specifically, I modified issue 1316 (https://issues.apache.org/jira/browse/IOTDB-1316)，“A sensor point value is invalid and this row can be imported successfully，but the following rows in this file are not executed”.

  In the original code, when there is a format error, such as in the example of issue 1316, in dump_ts3.csv, the format of "china" in the column of root.db1.tab1.other is wrong, and in dump_ts4.csv the format of id and age is wrong (it should be int32, but it is written as float type), the file import will stop and an error will be reported. 

  After modifying the code, when the wrong data format is encountered, the wrong data location will be replaced by null, and the data of this line and other lines can continue to be imported successfully.

### DIFF
--- a/cli/src/main/java/org/apache/iotdb/tool/ImportCsv.java
+++ b/cli/src/main/java/org/apache/iotdb/tool/ImportCsv.java
@@ -110,7 +110,7 @@ public class ImportCsv extends AbstractCsvTool {
   public static Connection getConnection(String ip, String port, String username, String password) {
     // JDBC driver name and database URL
     String driver = "org.apache.iotdb.jdbc.IoTDBDriver";
-    String url = "jdbc:iotdb://"+ip+":"+port+"/";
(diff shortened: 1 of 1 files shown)


cli/src/main/java/org/apache/iotdb/tool/ImportCsv.java
```diff
@@ -44,6 +44,7 @@
 import java.io.InputStreamReader;
 import java.io.LineNumberReader;
 import java.nio.charset.StandardCharsets;
+import java.sql.*;
```
wangchao316, line 47: generate we do not use import *.

PR #3133, opened 2021-05-06
[ISSUE-3132] Replace the dependence repository Bintray with JFrog in compile-tool

HTHou:
## Description

The compile-tool has a dependence `boost`, which needs to be downloaded from Bintray repository.

But since May 1st 2021, Bintray was not available and moved to JFrog. See https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/. 

Therefore, we have to update these URLs to new ones.


HTHou:
This PR need to be cherry picked to rel/0.12

jixuan1989 (approved):

PR #3139, opened 2021-05-07
[ISSUE-3116] Bug when using natural month unit in time interval in group by query

Alima777:
#3116 

Bug is caused by, when natural month unit is used in time interval in group by query, the flag `isIntervalByMonth` and `isSlidingStepByMonth` will be changed incorrectly.

### DIFF
--- a/server/src/test/java/org/apache/iotdb/db/integration/IoTDBGroupByMonthIT.java
+++ b/server/src/test/java/org/apache/iotdb/db/integration/IoTDBGroupByMonthIT.java
@@ -244,7 +244,8 @@ public class IoTDBGroupByMonthIT {
             cnt++;
           }
         }
-        Assert.assertEquals(30, cnt);
+        Assert.assertTrue(cnt >= 28);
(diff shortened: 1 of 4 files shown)


server/src/test/java/org/apache/iotdb/db/integration/IoTDBGroupByMonthIT.java
```diff
@@ -224,6 +224,34 @@ public void groupByNaturalMonth5() {
...
+      Assert.assertTrue(hasResultSet);
+      int cnt = 0;
+      try (ResultSet resultSet = statement.getResultSet()) {
+        while (resultSet.next()) {
+          String ans = resultSet.getString(sum("root.sg1.d1.temperature"));
+          if (ans.equals("0.0")) {
+            cnt++;
+          }
+        }
+        Assert.assertEquals(30, cnt);
```
jixuan1989, line 247: You use `[now() - 1mo, now())` in the tests...

30? natural month?

HTHou (approved):

PR #3141, opened 2021-05-07
[To rel/0.12] [ISSUE-3116] Bug when using natural month unit in time interval in group by query

Alima777:
#3116

Bug is caused by, when natural month unit is used in time interval in group by query, the flag isIntervalByMonth and isSlidingStepByMonth will be changed incorrectly.


Alima777:
It may be able to cherry pick from master, so i will not update it until the pr to master is merged.

PR #3153, opened 2021-05-10
[IOTDB-1321] Filter RowRecord automatically if any column in it is null or all columns are null

JackieTien97:
## Description
Currently
1. IoTDB will join all the sensor value by its time, and if some sensors don't have values in that timestamp, we will fill it with null. In some analysis scenarios, we only need the row if all the columns of it have value.

2. in group by query, we will fill null for any group by interval if the columns don't have values in that group by interval. However, if all columns in that group by interval are null, maybe users don't need that RowRecord, so we want IoTDB to support this feature.

## Syntax
`select * from root.sg1.d1 WITHOUT NULL ANY` will fix scenario 1 and filter the rows if any column of which is null.
`select last_value(*) from root.testWithoutAllNull.d1 GROUP BY([1, 21), 5ms) WITHOUT NULL ALL` will fix scenario 2 and filter the rows if all columns of which are null.


antlr/src/main/antlr4/org/apache/iotdb/db/qp/sql/SqlBase.g4
```diff
@@ -106,7 +106,8 @@ statement
     | SELECT topClause? selectElements
     fromClause
     whereClause?
-    specialClause? #selectStatement
+    specialClause?
+    (WITHOUT NULL (ALL | ANY))? #selectStatement
     ;
```
Alima777, line 111: To keep the structure, rename this statement as `withoutNullClause` and put it into specialClause.

server/src/main/java/org/apache/iotdb/db/qp/logical/crud/QueryOperator.java
```diff
@@ -62,6 +62,10 @@
 
   private IndexType indexType;
 
+  private boolean withoutAnyNull;
+
+  private boolean withoutAllNull;
```
Alima777, line 67: Add more java doc please

server/src/main/java/org/apache/iotdb/db/utils/QueryDataSetUtils.java
```diff
@@ -79,6 +67,16 @@ public static TSQueryDataSet convertQueryDataSetByFetchSize(
...
         RowRecord rowRecord = queryDataSet.next();
+        // filter the row that all columns are null
+        if (queryDataSet.isWithoutAllNull() && rowRecord.isAllNull()) {
+          i--;
+          continue;
+        }
+        // filter the row that any column is null
+        if (queryDataSet.isWithoutAnyNull() && rowRecord.hasNullField()) {
+          i--;
+          continue;
```
Alima777, line 78: Combine these two `if` using `or`

tsfile/src/main/java/org/apache/iotdb/tsfile/read/query/dataset/QueryDataSet.java
```diff
@@ -97,7 +103,15 @@ public boolean hasNext() throws IOException {
...
-        nextWithoutConstraint(); // DO NOT use next()
+        RowRecord rowRecord = nextWithoutConstraint(); // DO NOT use next()
+        // filter the row that all columns are null
+        if (withoutAllNull && rowRecord.isAllNull()) {
+          continue;
+        }
+        // filter the row that any column is null
+        if (withoutAnyNull && rowRecord.hasNullField()) {
+          continue;
+        }
```
Alima777, line 114: Combine these two `if` using `or`

tsfile/src/main/java/org/apache/iotdb/tsfile/read/common/RowRecord.java
```diff
@@ -36,14 +41,31 @@ public RowRecord(long timestamp) {
   public RowRecord(long timestamp, List<Field> fields) {
     this.timestamp = timestamp;
     this.fields = fields;
+    for (Field field : fields) {
+      if (field == null || field.getDataType() == null) {
+        hasNullField = true;
+      } else {
+        allNull = false;
+      }
+    }
```
Alima777, line 50: If we doesn't use the `without null` clause, we also have to check each field again, which may have an impact on performance.

server/src/main/java/org/apache/iotdb/db/query/dataset/RawQueryDataSetWithoutValueFilter.java
```diff
@@ -465,6 +454,60 @@ public TSQueryDataSet fillBuffer(int fetchSize, WatermarkEncoder encoder)
     return tsQueryDataSet;
   }
 
+  /** if any column in the row record, we filter it. */
```
Alima777, line 457: javadoc typo

server/src/main/java/org/apache/iotdb/db/query/dataset/RawQueryDataSetWithoutValueFilter.java
```diff
@@ -256,6 +256,10 @@ public TSQueryDataSet fillBuffer(int fetchSize, WatermarkEncoder encoder)
 
       long minTime = timeHeap.pollFirst();
 
+      if (withoutAnyNull && filterRowRecord(seriesNum, minTime)) {
+        continue;
+      }
```
Alima777, line 261: Why not making a judge on `withoutAllNull`?

Alima777:
I set the syntax order of `withoutNull` before `limit` and after the functional declaration. I think it's more reasonable.
Because `withoutNull` filters the result **from the origin result set,** and `limit` restrict the number of **final** result we have get.

Alima777:
Please add the user guide describing this function and its syntax rule.

Alima777 (approved):

PR #3156, opened 2021-05-11
[IOTDB-1352]fix cluster_info_public_port  parameter not set in cluster/src/test/resources/

jixuan1989:
after add `cluster_info_public_port`, we need to set them as different ports in our cluster/src/test/resources.


LebronAl (approved):
LGTM

PR #3164, opened 2021-05-12
[IOTDB-1371] Fix NPE when creating aligned timeseries and inserting with mismatched data type

samperson1997:
When creating aligned timeseries and inserting with mismatched data type, the MNode will be null and a NullPointerException would be thrown.


HTHou (approved):
LGTM

SilverNarcissus (approved):
Nice work!

PR #3181, opened 2021-05-13
Implement some JDBC methods to support DataGrip

liutaohua:
The `getMaxRow` and `getMoreResults` methods are used to connect IoTDB using `DataGrip`.

`IoTDB-JDBC` is not currently supported. 

I implemented two methods in which `getMoreResults` are always returning false instead of throwing an exception.

It looks OK in my local tests so far.

image

### DIFF
--- a/jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBStatement.java
+++ b/jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBStatement.java
@@ -19,6 +19,18 @@
 
 package org.apache.iotdb.jdbc;
 
+import java.nio.ByteBuffer;
+import java.sql.BatchUpdateException;
+import java.sql.Connection;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.SQLWarning;
+import java.sql.Statement;
+import java.time.ZoneId;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
 import org.apache.iotdb.rpc.RpcUtils;
 import org.apache.iotdb.rpc.StatementExecutionException;
 import org.apache.iotdb.rpc.TSStatusCode;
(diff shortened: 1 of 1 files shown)


jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBStatement.java
```diff
@@ -581,18 +590,21 @@ public void setMaxFieldSize(int arg0) throws SQLException {
...
+    if (num <= 0) {
+      throw new SQLException(String.format("maxRows %d must be > 0!", num));
+    }
+    this.maxRows = num;
   }
 
   @Override
   public boolean getMoreResults() throws SQLException {
-    throw new SQLException("Not support getMoreResults");
+    return false;
```
jixuan1989, line 607: sounds incorrect...
`getMoreResults` should return false either:
1. the number of fetched rows >= maxRows
2. no more results from resultSet.

liutaohua: When a statement executes multiple `SELECT`s, the `result` returned becomes a result with multiple results.
`getMoreResults` are used to obtain the next result set in this case.

```
ResultSet resultSet = preparedStatement.executeQuery("select * from root.group_9.d_9 slimit 10;select * from root.group_8.d_8 slimit 10;");
```

I tested whether the server supports multiple `SELECT` execution at the same time, and I got this error:

```
org.apache.iotdb.jdbc.IoTDBSQLException: 401: Error occurred while parsing SQL to physical plan: line 1:41 mismatched input 'select' expecting <EOF>
```

Obviously the server doesn't support it, so here I think it's correct to always return `false`.

HTHou: Copied from JDBC javaDoc. I agree with @liutaohua .

```
boolean getMoreResults()
                throws SQLException
Moves to this Statement object's next result, returns true if it is a ResultSet object, and implicitly closes any current ResultSet object(s) obtained with the method getResultSet.
There are no more results when the following is true:


     // stmt is a Statement object
     ((stmt.getMoreResults() == false) && (stmt.getUpdateCount() == -1))
 
Returns:
true if the next result is a ResultSet object; false if it is an update count or there are no more results
Throws:
SQLException - if a database access error occurs or this method is called on a closed Statement
```

HTHou (approved):

qiaojialin (approved):

PR #3195, opened 2021-05-14
[To rel/0.12] Update force full merge config

zhanglingzhe0820:
cherry pick #3194
Currently, we use inplace merge to speed up unseq compaction. However, there are some case that too much unseq data appended to one seq file which may make some seq file too large. So we change the default merge strategy to full merge instead.


mychaow:
There are some test fails, please take a look.

HTHou (approved):

PR #3202, opened 2021-05-17
Fix continuous compaction run into dead loop when unseq compaction cannot select candidates

zhanglingzhe0820:
Currently, if the unseq compaction cannot select candidates for merge, it will always run again in the same unseq file list in the continuous compaction process. So we have to avoid continuous compaction if the unseq file cannot be selected to be merged.


SteveYurongSu (approved):
LGTM

PR #3208, opened 2021-05-17
fix index out of bound when VectorTimeSeriesMetadata is not find in memory

liutaohua:
1. start  the `master` branch IoTDBServer
2. use `VectorSessionExample` to insert data:
```
    createTemplate();
    insertTabletWithAlignedTimeseriesMethod1();
    insertTabletWithAlignedTimeseriesMethod2();
```
3. use cli to select .
```
select * from root.*;
```

4.the error result like this:
```
[[INTERNAL_SERVER_ERROR] Exception occurred while executing executeQueryStatement. Index: 0][500] 500: [INTERNAL_SERVER_ERROR] Exception occurred while executing executeQueryStatement. Index: 0
```


liutaohua:
It does not appear that the error is related to this modification

JackieTien97 (approved):

PR #3209, opened 2021-05-17
Modify the getAggrResult method, don't destroy polymorphism too early

liutaohua:
The current cluster version cannot perform `count` because the `VectorPartialPath` type was missing prematurely when determining the type


neuyilan (approved):
image

I try to use the aligned timeseries feature to test this PR in the cluster module, however, this seems to have a lot of problems, maybe the aligned timeseries feature of the cluster module is not ready. 
But this PR LGTM.

HTHou:
spotless...

jixuan1989:
add a test to testcontainer

liutaohua:
> add a test to testcontainer

I tried to write the test, but I encountered a large number of errors. So much so that I need to fully understand all of the Vector and distribution related work if I want to do this. @jixuan1989 

Maybe we can do this when the distribution is fully compatible with Vector.

LebronAl (commented):
Hi, In fact, this may be a bit complicated after the alignedTimeseries function is merged, but the test sample may need to add more tests, such as counting a subbensor or two subsensors, which I suspect may not work properly.
such as:
```
select count(vm1) from root.template.device1
select count(vm1,vm2) from root.template.device1
```

LebronAl (approved):
LGTM

PR #3217, opened 2021-05-18
[To rel/0.12] Filter RowRecord automatically if any column in it is null or all columns are null

JackieTien97:
## Description
Currently
1. IoTDB will join all the sensor value by its time, and if some sensors don't have values in that timestamp, we will fill it with null. In some analysis scenarios, we only need the row if all the columns of it have value.

2. in group by query, we will fill null for any group by interval if the columns don't have values in that group by interval. However, if all columns in that group by interval are null, maybe users don't need that RowRecord, so we want IoTDB to support this feature.

## Syntax
`select * from root.sg1.d1 WITHOUT NULL ANY` will fix scenario 1 and filter the rows if any column of which is null.
`select last_value(*) from root.testWithoutAllNull.d1 GROUP BY([1, 21), 5ms) WITHOUT NULL ALL` will fix scenario 2 and filter the rows if all columns of which are null.


Alima777 (approved):
LGTM

PR #3218, opened 2021-05-18
[IOTDB-1403] Dictionary encoding for TEXT 

liujiajun:
## Motivation

Currently, TEXT is not encoded in IoTDB. While this approach works fine for data with low repetitive patterns, it may result in storage inefficiency and I/O overhead for low-cardinality data, where repetitions happen frequently. To this end, we propose supporting dictionary encoding for TEXT data type, which could provide higher compression rate and faster deserialeraliziation for low-cardinality data.

### Dictionary building

**Choice 1: Character aligned**

We scan the page data character by character to build the dictionary incrementally. 

- [+] The idea has been implemented in LZ encoding and its variants (LZ4, LZW, etc.)
- [-] IoTDB already supports LZ4 compression along with other compresion algorithms at compression level, so it may not be necessary to implement it again at encoder level.
- [-] Incurs relatively large compression overhead in the dictionary building phase.

**Choice 2 (current choice): Datapoint aligned**

This method is based on the key assumption that repetition happens frequently at data point level (e.g. in the extreme, all data points have value "Beijing") and does not care about whether repetitions happen across data points. We can then scan the page data point by data point to build the dictionary. Similar idea was implemented in TimeScaleDB ([https://github.com/timescale/timescaledb/blob/eace5ea23b299681a45efa312cf6f51f587d561c/tsl/src/compression/dictionary.h](https://github.com/timescale/timescaledb/blob/eace5ea23b299681a45efa312cf6f51f587d561c/tsl/src/compression/dictionary.h))

- [+] High compression rate when cardinality of data point values are low.
- [+] Smaller overhead for compression and decompression compared with choice 1
- [-] Does not capture repetition larger than the length of a data point.

## Encoding format

Dictionary encoding supports arbitrary binary data, and works well with data with low cardinality. 

The encoding format is shown below:

```
<map> <values>
  <map> := <map size> <map data>
    <map data> := [<entry size><entry data>]...
  <values> := [<value>]...
```

- `map` is an array of strings. During the encoding phase, we can use an in-memory hash table to speed up look-up.
- `encodedValues` is  an array of indexes corresponding to the indexes of `map`. This allows us to apply (RLE encoding + bit packing) on it during serialisation to further reduce space usage.

### Encoding

```markup
func encode(values[]):
	map[] <- []
	encodedValues[] <- []
	for value in values:
		if value in map:
			encodedValues.push(map.indexOf(value))
	  else:
	    map.push(value)
	return [map, encodedValues.push]
```

### Decoding

```markup
func decode(map, encodedValues):
	values[] <- []
  for value in encodedValues:
	   values.push(map[value])
  return values
```

### ~~Adaptive encoding~~

~~Not all binary data are suitable for dictionary encoding. This, I propose to implemented the ability to dynamically choose whether to use dictionary or plain encoding based on the cardinality of the page (i.e. number of unique values).~~

### Setup

Total number of points = 100, 000

Point size = 32B

Max page size = 64 * 1024 B

### Results
Screenshot_2021-05-19_at_1 07 23_PM

- **Observation 1.** Compression rate (lower the better) increases as the repetitions of values decreases
- **Observation 2.** Slowdown (Time taken to for dictionary encoding / time taken for plain encoding) increases as the repetitions of values decreases
**Conclusion:** Dictionary encoding works well for data with many repetitions, giving up to 100x smaller space usage with little encoding/decoding overhead. However, it is not suitable for data with few repetitions.

This PR has:
- [x] been self-reviewed.
    - [x] concurrent read
    - [x] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


antlr/src/main/antlr4/org/apache/iotdb/db/qp/sql/SqlBase.g4
```diff
@@ -695,7 +695,7 @@ dateExpression
     ;
 
 encoding
-    : PLAIN | PLAIN_DICTIONARY | RLE | DIFF | TS_2DIFF | GORILLA | REGULAR
+    : PLAIN | PLAIN_DICTIONARY | RLE | DIFF | TS_2DIFF | GORILLA | REGULAR | DICTIONARY
```
liutaohua, line 698: I suggest renaming the original `PLAIN_DICTIONARY` as `DICTIONARY`, and modify the relevant documents, including `Format.md` and so on...

client-cpp/src/main/Session.h
```diff
@@ -112,7 +112,8 @@ namespace TSEncoding {
         BITMAP = 5,
         GORILLA_V1 = 6,
         REGULAR = 7,
-        GORILLA = 8
+        GORILLA = 8,
+        DICTIONARY = 9
```
liutaohua, line 116: reuse the number of `PLAIN_DICTIONARY`

client-py/iotdb/utils/IoTDBConstants.py
```diff
@@ -40,6 +40,7 @@ class TSEncoding(Enum):
     GORILLA_V1 = 6
     REGULAR = 7
     GORILLA = 8
+    DICTIONARY = 9
```
liutaohua, line 43: same

tsfile/src/main/java/org/apache/iotdb/tsfile/encoding/decoder/DictionaryDecoder.java
```diff
@@ -0,0 +1,86 @@
...
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class DictionaryDecoder extends Decoder {
+  private static final Logger logger = LoggerFactory.getLogger(DictionaryDecoder.class);
+
+  private List<Binary> map;
```
liutaohua, line 37: rename to `indexEntry`

tsfile/src/main/java/org/apache/iotdb/tsfile/encoding/decoder/DictionaryDecoder.java
```diff
@@ -0,0 +1,86 @@
...
+    if (map == null) {
+      initMap(buffer);
+    }
+    int code = valueDecoder.readInt(buffer);
+    return map.get(code);
+  }
+
+  private void initMap(ByteBuffer buffer) {
+    map = new ArrayList<>();
+    int length = ReadWriteForEncodingUtils.readVarInt(buffer);
```
liutaohua, line 72: initialize the array with a fixed length to avoid redundant copies

tsfile/src/main/java/org/apache/iotdb/tsfile/encoding/encoder/DictionaryEncoder.java
```diff
@@ -0,0 +1,113 @@
...
+ * <map> <indexes>
+ * <map> := <map length> <map data>
+ * <map data> := [<entry size><entry data>]...
+ * <indexes> := [<index>]...
+ * }</pre>
+ */
+public class DictionaryEncoder extends Encoder {
+  private static final Logger logger = LoggerFactory.getLogger(DictionaryEncoder.class);
+
+  private HashMap<Binary, Integer> valueToCode;
```
liutaohua, line 48: rename to entryIndex

tsfile/src/main/java/org/apache/iotdb/tsfile/encoding/encoder/DictionaryEncoder.java
```diff
@@ -0,0 +1,113 @@
...
+  private List<Binary> map;
```
liutaohua, line 49: rename to IndexEntry

liutaohua (commented):
Thank you for the great contribution.
But there are some small problems that need to be improved。

HTHou:
Hi, thanks for you contribution. What an obviously good feature! 
Would you like to create a jira issue and update the PR title with the issue number?

liujiajun:
> Hi, thanks for you contribution. What an obviously good feature!
> Would you like to create a jira issue and update the PR title with the issue number?

Thank you! Updated accordingly.

tsfile/src/main/java/org/apache/iotdb/tsfile/encoding/encoder/TSEncodingBuilder.java
```diff
@@ -296,4 +298,22 @@ public void initFromProps(Map<String, String> props) {
...
+  public static class Dictionary extends TSEncodingBuilder {
+
+    @Override
+    public Encoder getEncoder(TSDataType type) {
+      switch (type) {
+        case TEXT:
+          return new DictionaryEncoder();
+        default:
+          throw new UnSupportedDataTypeException("DICTIONARY doesn't support data type: " + type);
+      }
```
HTHou, line 311: How about using if-else here?

tsfile/src/main/java/org/apache/iotdb/tsfile/encoding/encoder/DictionaryEncoder.java
```diff
@@ -0,0 +1,113 @@
...
+    mapSize = 0;
+  }
+
+  @Override
+  public void encode(Binary value, ByteArrayOutputStream out) {
+    if (!entryIndex.containsKey(value)) {
+      entryIndex.put(value, entryIndex.size());
+      indexEntry.add(value);
+      mapSize += value.getLength();
+    }
```
HTHou, line 68: How about using `entryIndex.computeIfAbsent()`?

tsfile/src/main/java/org/apache/iotdb/tsfile/encoding/encoder/TSEncodingBuilder.java
liujiajun: Updated

tsfile/src/main/java/org/apache/iotdb/tsfile/encoding/encoder/DictionaryEncoder.java
liujiajun: Updated

liutaohua (approved):
thank you for the great contribution

HTHou (approved):

PR #3219, opened 2021-05-18
add query last data interface 

ijihang:
## Description
When the user queries for the latest value,we hope skip sql sql parser,
i add last data query interface in session according to raw data query.
This is jira : https://issues.apache.org/jira/browse/IOTDB-670


ijihang:
insert 100w timeseries
Test:   no sql parse:45ms,have sql parse:125ms
image

mychaow (approved):
LGTM

session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -589,6 +589,47 @@ public SessionDataSet executeRawDataQuery(List<String> paths, long startTime, lo
     }
   }
 
+  /**
+   * only: select last status from root.ln.wf01.wt01 where time >= 1621326244168; <PrefixPath> +
+   * <suffixPath> = <TimeSeries>
+   *
+   * @param suffixPath root.ln.wf01.wt01
+   * @param prefixPath status
```
neuyilan, line 597: Change the comment to a reasonable one, we should first explain the param's meaning and then we can give some examples.

session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -589,6 +589,47 @@ public SessionDataSet executeRawDataQuery(List<String> paths, long startTime, lo
...
+        throw new StatementExecutionException(MSG_DONOT_ENABLE_REDIRECT);
+      }
+    }
+  }
+
+  /**
+   * query eg. select last status from root.ln.wf01.wt01; <PrefixPath> + <suffixPath> = <TimeSeries>
+   *
+   * @param suffixPath query statement
+   * @param prefixPath query statement
```
neuyilan, line 625: The same as above,  if we provide the params, we should explain the params correctly.

neuyilan (changes requested):
Only some minor issues, PTAL
Besides, have you tested in the cluster version?

ijihang:
> Only some minor issues, PTAL
> Besides, have you tested in the cluster version?

only test stand-alone ,i think  should be  same

session/src/main/java/org/apache/iotdb/session/Session.java
ijihang: ok ，Has been modified and submitted

Alima777 (changes requested):

neuyilan:
> > Only some minor issues, PTAL
> > Besides, have you tested in the cluster version?
> 
> only test stand-alone ,i think should be same

Please add one cluster IT test in the testcontainer module, it's better to use automated testing to ensure correctness than we think. In this way, if the subsequent developers modify the code by mistake, they can have automatic tests to check it out.

ijihang:
add cluster test lastDataQuery ,and  modify Dockerfile, Because the package is compiled in Windows system and then run into docker, the file format needs to be modified

thrift/src/main/thrift/rpc.thrift
```diff
@@ -291,6 +291,16 @@ struct TSRawDataQueryReq {
   7: optional bool enableRedirectQuery;
 }
 
+struct TSLastDataQueryReq {
+  1: required i64 sessionId
+  2: required list<string> suffixPath
+  3: required list<string> prefixPath
```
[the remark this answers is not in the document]
ijihang: Yes, this is to increase user convenience. If we write the full path, we will write more

testcontainer/src/test/java/org/apache/iotdb/db/sql/Cases.java
```diff
@@ -220,4 +220,71 @@ public void vectorCountTest() throws IoTDBConnectionException, StatementExecutio
...
+    prefixPath.add("root.sg1.d2");
+
+    SessionDataSet sessionDataSet = session.executeLastDataQuery(suffixPath, prefixPath);
+    sessionDataSet.setFetchSize(1024);
+
+    int count = 0;
+    while (sessionDataSet.hasNext()) {
+      count++;
+    }
+    Assert.assertEquals(2, count);
```
neuyilan, line 249: The same as above.

session/src/test/java/org/apache/iotdb/session/IoTDBSessionComplexIT.java
```diff
@@ -594,6 +608,25 @@ private void rawDataQuery() throws StatementExecutionException, IoTDBConnectionE
...
+    prefixPath.add("root.sg1.d2");
+
+    SessionDataSet sessionDataSet = session.executeLastDataQuery(suffixPath, prefixPath);
+    sessionDataSet.setFetchSize(1024);
+
+    int count = 0;
+    while (sessionDataSet.hasNext()) {
+      count++;
+    }
+    Assert.assertEquals(2, count);
```
neuyilan, line 626: As we control the insert of the records, it would be better to add one assert that can make sure the value of the two measurements.

thrift/src/main/thrift/rpc.thrift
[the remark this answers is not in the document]
Alima777: Sorry, I don't quite get your meaning. 
1. Now you define the user interface which need suffixPath and prefixPath both, the user will have to give two arrays, which is more complex and not intuitive enough. 
For example, I want query `s1` of device `root.sg.d1` and `root.sg.d2`, the sql will be like:
`select last s1 from root.sg.d1, root.sg.d2`.
suffixPath = {s1}, prefixPath = {root.sg.d1, root.sg.d2}.
But intuitively speaking, when calling the session interface, I just give any path I want to query. e.g.
completePath = {root.sg.d1.s1, root.sg.d2.s1}.

2. And it's limited in function.
For exmaple, if I want to get `s1` and `s2` of device `root.sg.d1`, and just `s1` of device `root.sg.d2`.
You can't write the correct suffix array directly in your way, right?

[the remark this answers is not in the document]
ijihang: Yes, you are right, so I submitted a new commit to modify the time series to full path

testcontainer/src/test/java/org/apache/iotdb/db/sql/Cases.java
ijihang: ok,the new commit add

thrift/src/main/thrift/rpc.thrift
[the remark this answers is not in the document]
Alima777: Thank you~ :D

server/src/main/java/org/apache/iotdb/db/qp/Planner.java
```diff
@@ -139,6 +140,40 @@ public PhysicalPlan rawDataQueryReqToPhysicalPlan(
...
+    BasicFunctionOperator basicFunctionOperator =
+        new BasicFunctionOperator(SQLConstant.GREATERTHANOREQUALTO, timePath, Long.toString(time));
+    queryOp.setFilterOperator(basicFunctionOperator);
+
+    int maxDeduplicatedPathNum = Integer.MAX_VALUE - 1;
+    Operator op = logicalOptimize(queryOp, maxDeduplicatedPathNum);
+
+    PhysicalGenerator physicalGenerator = new PhysicalGenerator();
+    return physicalGenerator.transformToPhysicalPlan(op, req.fetchSize);
+  }
```
Alima777, line 175: A easy-fix problem is: you can see `rawDataQueryReqToPhysicalPlan`, the code logic is abstracted and very clear, we hope you can keep this code structure.

server/src/main/java/org/apache/iotdb/db/qp/Planner.java
```diff
@@ -139,6 +140,40 @@ public PhysicalPlan rawDataQueryReqToPhysicalPlan(
...
+      throws QueryProcessException, IllegalPathException {
+    List<String> paths = req.getPaths();
+    long time = req.getTime();
+
+    SelectOperator selectOp = new SelectOperator(SQLConstant.TOK_SELECT, zoneId);
+    FromOperator fromOp = new FromOperator(SQLConstant.TOK_FROM);
+    QueryOperator queryOp = new QueryOperator(SQLConstant.TOK_QUERY);
+
+    selectOp.setLastQuery();
+    selectOp.addSelectPath(new PartialPath(""));
```
Alima777, line 154: The data structure of select opeartor was changed, since a PR modified it was merged before. You can see `generate(TSRawDataQueryReq rawDataQueryReq, ZoneId zoneId)` for reference.

```suggestion
    selectOp.setLastQuery();
    selectOp.addResultColumn(new ResultColumn(new TimeSeriesOperand(new PartialPath(""))));
```

Alima777 (commented):
Look good to me now. :D Please fix the small issues~

server/src/main/java/org/apache/iotdb/db/qp/Planner.java
ijihang: Yes, I saw the master modify part of the code ,so i submit a new one

cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanner.java
```diff
@@ -60,4 +62,18 @@ public PhysicalPlan rawDataQueryReqToPhysicalPlan(
     return new ClusterPhysicalGenerator()
         .transformToPhysicalPlan(operator, rawDataQueryReq.fetchSize);
   }
+
+  @Override
+  public PhysicalPlan lastDataQueryReqToPhysicalPlan(
```
mychaow, line 67: Why do you need to override the method?

ijihang: ClusterPlanner extends Planner,i  add lastDataQueryReqToPhysicalPlan in Planner

cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanner.java
```diff
@@ -60,4 +62,18 @@ public PhysicalPlan rawDataQueryReqToPhysicalPlan(
...
+      TSLastDataQueryReq lastDataQueryReq, ZoneId zoneId)
+      throws QueryProcessException, IllegalPathException {
+    // from TSLastDataQueryReq to logical operator
+    Operator operator = LogicalGenerator.generate(lastDataQueryReq, zoneId);
+    // check if there are logical errors
+    LogicalChecker.check(operator);
+    // optimize the logical operator
+    operator = logicalOptimize(operator, lastDataQueryReq.fetchSize);
+    // from logical operator to physical plan
+    return new PhysicalGenerator().transformToPhysicalPlan(operator, lastDataQueryReq.fetchSize);
```
Alima777, line 77:
```suggestion
    return new ClusterPhysicalGenerator().transformToPhysicalPlan(operator, lastDataQueryReq.fetchSize);
```

Alima777 (approved):
It looks good to me now :D Thank you~

mychaow (approved):
LGTM

PR #3242, opened 2021-05-21
[IOTDB-1384] [To rel/0.11] Fix group by bug

JackieTien97:
## Description
we can see in the jira.

## Reasons
Forgot to reset the `lastReadCurArrayIndex` and `lastReadCurListIndex` while generating new BatchData which cause the new BatchData will be read from the previous `lastReadCurArrayIndex` and `lastReadCurListIndex`.

## Solution
reset the `lastReadCurArrayIndex` and `lastReadCurListIndex` while generating new BatchData.

### DIFF
--- a/server/src/test/java/org/apache/iotdb/db/integration/IoTDBGroupByUnseqIT.java
+++ b/server/src/test/java/org/apache/iotdb/db/integration/IoTDBGroupByUnseqIT.java
@@ -1,0 +1,130 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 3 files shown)


server/src/test/java/org/apache/iotdb/db/integration/IoTDBGroupByUnseqIT.java
```diff
@@ -0,0 +1,130 @@
...
+import org.apache.iotdb.db.utils.EnvironmentUtils;
+import org.apache.iotdb.jdbc.Config;
+import org.apache.iotdb.tsfile.common.conf.TSFileDescriptor;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+public class IoTDBGroupByUnseqIT {
+
```
Alima777, line 38: Better add more javadoc for this IT to explain the situation, since lots of `GroupByIT` exsits now that we can't see why we need this test.

Alima777 (approved):
LGTM

PR #3246, opened 2021-05-21
[IOTDB-1387] [To rel/0.12] Fix Without Null ALL doesn't take effect in align by device clause

JackieTien97:
## Description

The details can be seen in JIRA.

## Reasons

Align by device always have one device column which will never be null. While using WITHOUT NULL clause, we shouldn't take device column into consideration.

## Solution

Reset the null state in RowRecord after add the device column. Also decrease the alreadyReturnedRowNum field in QueryDataset while the current RowRecord doesn't satisify the without null constraint.


server/src/main/java/org/apache/iotdb/db/query/dataset/AlignByDeviceDataSet.java
```diff
@@ -235,6 +235,7 @@ public RowRecord nextWithoutConstraint() throws IOException {
     Field deviceField = new Field(TSDataType.TEXT);
     deviceField.setBinaryV(new Binary(currentDevice.getFullPath()));
     rowRecord.addField(deviceField);
+    rowRecord.resetNullFlag();
```
Alima777, line 238: Add comment please

Alima777 (approved):

PR #3247, opened 2021-05-21
pseudo distributed no logs error

haimeiguo:
logback.xml not added in the conf folder. 
now each node's logs are moved to node1/ node2/ node3/ folders
image


neuyilan (approved):
Many thanks.
BTW, does the lockback file[1] is needed? if not, we can delete it.
[1] https://github.com/apache/iotdb/blob/master/cluster/src/test/resources/logback.xml

HTHou:
> BTW, does the lockback file[1] is needed? if not, we can delete it.
> [1] https://github.com/apache/iotdb/blob/master/cluster/src/test/resources/logback.xml

Is it for starting one node?

HTHou (approved):

neuyilan:
> > BTW, does the lockback file[1] is needed? if not, we can delete it.
> > [1] https://github.com/apache/iotdb/blob/master/cluster/src/test/resources/logback.xml
> 
> Is it for starting one node by `nohup ./cluster/target/iotdb-cluster-0.13.0-SNAPSHOT/sbin/start-node.sh >/dev/null 2>&1 & `

I think if we only start one node cluster using the command `nohup ./cluster/target/iotdb-cluster-0.13.0-SNAPSHOT/sbin/start-node.sh >/dev/null 2>&1 & `, the logback in the `iotdb-cluster-0.13.0-SNAPSHOT/conf` folder  is used.

cluster/src/test/resources/node2conf/logback.xml
```diff
@@ -0,0 +1,282 @@
...
+        <filter class="ch.qos.logback.classic.filter.LevelFilter">
+            <level>error</level>
+            <onMatch>ACCEPT</onMatch>
+            <onMismatch>DENY</onMismatch>
+        </filter>
+    </appender>
+    <appender class="ch.qos.logback.core.rolling.RollingFileAppender" name="FILEWARN">
+        <file>${IOTDB_HOME}/logs_node2/log_warn.log</file>
+        <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
+            <fileNamePattern>${IOTDB_HOME}/logs_node2/log-warn-%d{yyyyMMdd-HH}.%i.log.gz</fileNamePattern>
```
mychaow, line 48: why use logs_node2?

haimeiguo: fixed

PR #3249, opened 2021-05-21
Support JDBC placeholder when set sensor or device by set string

yanhongwangg:
before，we can not set sensor or device in sql by set string
now we support it but you must handle double quotes yourself


neuyilan (approved):
Decent work!

HTHou (approved):
LGTM

PR #3250, opened 2021-05-21
[IOTDB-1384] [To rel/0.12] Fix group by bug

JackieTien97:
## Description
we can see in the jira.

## Reasons
Forgot to reset the `lastReadCurArrayIndex` and `lastReadCurListIndex` while generating new BatchData which cause the new BatchData will be read from the previous `lastReadCurArrayIndex` and `lastReadCurListIndex`.

## Solution
reset the `lastReadCurArrayIndex` and `lastReadCurListIndex` while generating new BatchData.


mychaow (approved):
LGTM

PR #3253, opened 2021-05-24
fix import csv split by comma bug

haimeiguo:
previously when ImportCSV tool read csv files, single or double quotes are not allowed inside a string (ie ad"bcd). 
This pr changes the csv splitting method. now it allows quotes or commas inside a string value. 
 
csv file as following:
time,root.fit.d6.s5,root.fit.d6.s6
1,True,
4,True,aapovpba=","akava'',vava=z//z'nnn
6,True,aapovpba=','akavavava=z//z'nnn
7,True,"aapovpba=,,,akavavava=z//z'nnn"

after import csv, query result:
image


mychaow (approved):
LGTM

PR #3255, opened 2021-05-24
[IOTDB-1390][To rel/0.12] Fix unseq compaction loss data bug

zhanglingzhe0820:
## Behavior
set compaction_strategy = LEVEL_COMPACTION

use the `data` to recover the IoTDB. After Recovery, it will merge the unseq data automatically. After unseq compaction, the data of some sensor in seq file loss.

use the following SQL to check data status before and after unseq compaction.

select Channel96 from root.*.J053102.T034.V1 where time >= 2021-05-18 22:32:27 and time <= 2021-05-18 22:32:37;
9FC2321FEC32B5D6D27ECF043C05AF93
73AA84209157A758BAF4153BB569FB1D

## Reason
Unseq compaction filter sensor by lexicographic order. However, the input sensor list may be not in lexicographic order.

## Solution
Ignore the condition of lexicographic order and use traditional filter means.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/merge/task/MergeMultiChunkTask.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/merge/task/MergeMultiChunkTask.java
@@ -195,6 +195,16 @@ public class MergeMultiChunkTask {
     mergeLogger.logTSEnd();
   }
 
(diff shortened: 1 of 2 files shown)


mychaow (approved):
LGTM

PR #3270, opened 2021-05-26
Bug Fix: Overlapped data should be consumed first

Alima777:
When judging whether a page is overlapped with merge reader, '>=' should be used rather than '>'.


Alima777:
Do not forget to check pick to 0.11/0.12.

JackieTien97 (approved):

PR #3271, opened 2021-05-26
[To rel/0.11][IOTDB-1356] Separate unseq_file_num_in_each_level from selecting candidate file in unseq compaction

zhanglingzhe0820:
Currently, unseq_file_num_in_each_level will impact the max selected file num in unseq compaction. This is not intuition.

Need to optimize the compaction candidate selecting process, according to the merge_memory_budget, max_selecting_merge_file_num, max open file num in system.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/compaction/TsFileManagement.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/compaction/TsFileManagement.java
@@ -72,6 +72,8 @@ public abstract class TsFileManagement {
   protected boolean isMergeExecutedInCurrentTask = false;
 
   protected boolean isForceFullMerge = IoTDBDescriptor.getInstance().getConfig().isForceFullMerge();
+  private final int maxOpenFileNumInEachUnseqCompaction =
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/apache/iotdb/db/engine/compaction/TsFileManagement.java
```diff
@@ -214,6 +216,16 @@ public boolean merge(
...
 
+    if (unSeqMergeList.size() > maxOpenFileNumInEachUnseqCompaction) {
+      logger.info(
+          "{} too much unseq files to be merged, reduce it to {}",
+          storageGroupName,
+          maxOpenFileNumInEachUnseqCompaction);
+      unSeqMergeList =
+          unSeqMergeList.subList(
+              unSeqMergeList.size() - maxOpenFileNumInEachUnseqCompaction, unSeqMergeList.size());
+    }
```
jt2594838, line 227: Older files should be merged first, otherwise, the overwritten data may re-emerge.

zhanglingzhe0820: The new unseq file may be more important?

PR #3275, opened 2021-05-27
cluster bug - fix privilege check

haimeiguo:
This PR fixed the following 2 bugs:
1. For user who does not have CREATE_TIMESERIES privilege, JDBC executeBatch() interface shows execution success, though the unauthorized timeseries are not created.
2. if enabled auto-create schema, CMManager creates SGs for all paths ignoring privilege check. 

Now fixed as:
```
Exception in thread "main" java.sql.BatchUpdateException: 
No permissions for this operation CREATE_TIMESERIES for SQL: "CREATE TIMESERIES root.sg4.d1.s1 WITH DATATYPE=INT64"
No permissions for this operation CREATE_TIMESERIES for SQL: "CREATE TIMESERIES root.sg7.d1.s1 WITH DATATYPE=INT64"
No permissions for this operation CREATE_TIMESERIES for SQL: "CREATE TIMESERIES root.sg8.d1.s1 WITH DATATYPE=INT64"

	at org.apache.iotdb.jdbc.IoTDBStatement.executeBatchSQL(IoTDBStatement.java:355)
	at org.apache.iotdb.jdbc.IoTDBStatement.executeBatch(IoTDBStatement.java:308)
	at org.apache.iotdb.JDBCExample.main(JDBCExample.java:43)
```

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/coordinator/Coordinator.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/coordinator/Coordinator.java
@@ -46,6 +46,7 @@ import org.apache.iotdb.db.exception.metadata.PathNotExistException;
 import org.apache.iotdb.db.exception.metadata.StorageGroupNotSetException;
 import org.apache.iotdb.db.exception.query.QueryProcessException;
 import org.apache.iotdb.db.metadata.PartialPath;
+import org.apache.iotdb.db.qp.physical.BatchPlan;
 import org.apache.iotdb.db.qp.physical.PhysicalPlan;
 import org.apache.iotdb.db.qp.physical.crud.InsertMultiTabletPlan;
 import org.apache.iotdb.db.qp.physical.crud.InsertPlan;
@@ -228,14 +229,9 @@ public class Coordinator {
       return StatusUtils.PARTITION_TABLE_NOT_READY;
     }
 
-    // not allowed to create sgs
-    if (plan instanceof CreateMultiTimeSeriesPlan) {
-      CreateMultiTimeSeriesPlan createMultiTimeSeriesPlan = (CreateMultiTimeSeriesPlan) plan;
-      if (createMultiTimeSeriesPlan.getResults().size()
-          == createMultiTimeSeriesPlan.getPaths().size()) {
-        return concludeFinalStatus(
-            plan, plan.getPaths().size(), true, null, false, null, Collections.emptyList());
-      }
+    if (!checkPrivilegeForBatchExecution(plan)) {
+      return concludeFinalStatus(
+          plan, plan.getPaths().size(), true, null, false, null, Collections.emptyList());
     }
 
     // split the plan into sub-plans that each only involve one data group
@@ -269,6 +265,20 @@ public class Coordinator {
     }
     logger.debug("{}: The data groups of {} are {}", name, plan, planGroupMap);
     return forwardPlan(planGroupMap, plan);
+  }
+
+  /**
+   * check if batch execution plan has privilege on any sg
+   *
+   * @param plan
+   * @return
+   */
+  private boolean checkPrivilegeForBatchExecution(PhysicalPlan plan) {
+    if (plan instanceof BatchPlan) {
+      return ((BatchPlan) plan).getResults().size() != plan.getPaths().size();
+    } else {
+      return true;
+    }
   }
 
   /**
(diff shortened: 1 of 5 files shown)


mychaow:
add ut

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
```diff
@@ -590,7 +590,17 @@ public void createSchema(PhysicalPlan plan) throws MetadataException, CheckConsi
               Collections.singletonList(
                   new PartialPath(((SetDeviceTemplatePlan) plan).getPrefixPath()))));
     } else {
-      storageGroups.addAll(getStorageGroups(plan.getPaths()));
+      List<PartialPath> paths = new ArrayList<>();
```
LebronAl, line 593: why not judge in this way?
```
} else if (plan instanceof CreateMultiTimeSeriesPlan) {
   xxx
} else {
storageGroups.addAll(getStorageGroups(paths));
}
```

cluster/src/main/java/org/apache/iotdb/cluster/coordinator/Coordinator.java
```diff
@@ -334,7 +345,8 @@ private TSStatus forwardPlan(List<PartitionGroup> partitionGroups, PhysicalPlan
   private TSStatus forwardPlan(Map<PhysicalPlan, PartitionGroup> planGroupMap, PhysicalPlan plan) {
     // the error codes from the groups that cannot execute the plan
     TSStatus status;
-    if (planGroupMap.size() == 1) {
+    // need to create substatus for multiPlan
+    if (planGroupMap.size() == 1 && !(plan instanceof CreateMultiTimeSeriesPlan)) {
```
LebronAl, line 349: I think these specializations make the code difficult to maintain, so why not do permission checking in `TSServiceImpl`?

LebronAl (changes requested):
PTAL

jixuan1989 (changes requested):
Test needed.

cluster/src/main/java/org/apache/iotdb/cluster/coordinator/Coordinator.java
haimeiguo: TSServiceImpl did privilege check and mark it in plan.getResults() map, so server mode privilege works fine. 
But in cluster mode, if current user has no privilege and cannot create any timeseries, in Coordinator planGroupMap is always empty, then the processPartitionedPlan() method will always call upon itself to try split the plan. its an endless loop

so i think we need a check here to break out the loop when planGroupMap is empty.
please let me know if you have any workaround

jt2594838: Still not quite clear of here.
Why does this branch not create substatus for multiPlans? And how about other multiPlans other than creating timeseries?

cluster/src/test/java/org/apache/iotdb/cluster/integration/SingleNodeTest.java
```diff
@@ -93,4 +98,34 @@ public void testDeleteNonExistTimeSeries()
...
+            System.lineSeparator()
+                + "No permissions for this operation CREATE_TIMESERIES for SQL: \"create timeseries root.sg1.d1.s1 with datatype=int32\""
+                + System.lineSeparator()
+                + "No permissions for this operation CREATE_TIMESERIES for SQL: \"create timeseries root.sg2.d1.s1 with datatype=int32\""
+                + System.lineSeparator(),
+            e.getMessage());
+        System.out.println(e.getMessage());
+      }
+    } catch (Exception e) {
+      e.printStackTrace();
```
jt2594838, line 128: Fail it.

cluster/src/test/java/org/apache/iotdb/cluster/integration/SingleNodeTest.java
```diff
@@ -93,4 +98,34 @@ public void testDeleteNonExistTimeSeries()
...
+        userStatement.executeBatch();
+      } catch (Exception e) {
+        assertEquals(
+            System.lineSeparator()
+                + "No permissions for this operation CREATE_TIMESERIES for SQL: \"create timeseries root.sg1.d1.s1 with datatype=int32\""
+                + System.lineSeparator()
+                + "No permissions for this operation CREATE_TIMESERIES for SQL: \"create timeseries root.sg2.d1.s1 with datatype=int32\""
+                + System.lineSeparator(),
+            e.getMessage());
+        System.out.println(e.getMessage());
```
jt2594838, line 125: Do not use sout if not necessary.

jt2594838 (approved):

jt2594838 (changes requested):
Good job, but something needs further discussion.

cluster/src/main/java/org/apache/iotdb/cluster/coordinator/Coordinator.java
haimeiguo: in processPartitionedPlan method, it checks if (planGroupMap == null || planGroupMap.isEmpty()) first, when user has no privilege, the planGroupMap is empty, so processPartitionedPlan method calls on itself directly. It skips calling on forwardPlan(). thus cannot create substatus.

jt2594838 (approved):

cluster/src/main/java/org/apache/iotdb/cluster/coordinator/Coordinator.java
```diff
@@ -250,6 +258,20 @@ public TSStatus processPartitionedPlan(PhysicalPlan plan) throws UnsupportedPlan
...
+   *
+   * @param plan
+   * @return
+   */
+  private boolean checkPrivilegeForBatchExecution(PhysicalPlan plan) {
+    if (plan instanceof BatchPlan) {
+      return ((BatchPlan) plan).getResults().size() != plan.getPaths().size();
+    } else {
+      return true;
+    }
```
neuyilan, line 272: IMO, If` ((BatchPlan) plan).getResults().size()>0`, which means some plan not have the privilege on some sg. 
I do not clearly know what's the above code mean.

neuyilan (changes requested):

cluster/src/main/java/org/apache/iotdb/cluster/coordinator/Coordinator.java
haimeiguo: batch execution can operate on multiple storage groups, if auser has privilege on any of the sg, he can continue in processPartitionedPlan. on the other hand if he has no privilege on any sgs when getResults.size() == paths.size(). it will jump to concludeFinalStatus(...)

neuyilan: ok

neuyilan (approved):
LGTM

SteveYurongSu:
To rel/0.12: #3476

PR #3292, opened 2021-05-31
[IOTDB-1407] Filtering time series based on tags query fails Occasion…

wangchao316:
https://issues.apache.org/jira/browse/IOTDB-1407
5 node , 3 rep.

1. Connect to the client.

2. Run the show timeseries root.turbine where owner=user1 command.
throw :   The key owner is not  a tag.

reason：
When the number of nodes is greater than the number of replicas, querying a sequence label is broadcast to all DataGroups. Some DataGroups do not contain the sequence. Therefore, an error is reported when querying the label. As a result, an error is reported for the entire query.

solved：
When the root sequence is queried, all DataGroup queries are broadcast.
When a subsequence is queried, the data group of the corresponding subsequence is queried.


wangchao316:
@LebronAl @neuyilan @mychaow , hi ,could you please reivew?
Thanks..

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
```diff
@@ -1690,7 +1690,15 @@ public MNode getMNode(MNode deviceMNode, String measurementName) {
...
         new ThreadPoolExecutor(
             THREAD_POOL_SIZE, THREAD_POOL_SIZE, 0, TimeUnit.SECONDS, new LinkedBlockingDeque<>());
-    List<PartitionGroup> globalGroups = metaGroupMember.getPartitionTable().getGlobalGroups();
+
+    List<PartitionGroup> globalGroups = new ArrayList<>();
+    if (IoTDBConstant.PATH_ROOT.equals(plan.getPath().getFullPath())) {
+      globalGroups = metaGroupMember.getPartitionTable().getGlobalGroups();
+    } else {
+      PartitionGroup partitionGroup =
+          metaGroupMember.getPartitionTable().partitionByPathTime(plan.getPath(), 0);
```
mychaow, line 1699: show timeseries root.* is not right?

wangchao316: Thanks @mychaow .  this is have a situation,  root.a.*
I  will alter this.

LebronAl:
Hi, it's better to add some ITs in testContainer for this function

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
```diff
@@ -1690,7 +1690,16 @@ public MNode getMNode(MNode deviceMNode, String measurementName) {
     ExecutorService pool =
         new ThreadPoolExecutor(
             THREAD_POOL_SIZE, THREAD_POOL_SIZE, 0, TimeUnit.SECONDS, new LinkedBlockingDeque<>());
-    List<PartitionGroup> globalGroups = metaGroupMember.getPartitionTable().getGlobalGroups();
+
+    List<PartitionGroup> globalGroups = new ArrayList<>();
+    if (IoTDBConstant.PATH_ROOT.equals(plan.getPath().getFullPath())
+        || plan.getPath().getFullPath().contains(IoTDBConstant.PATH_WILDCARD)) {
```
neuyilan, line 1696: As we have supported special characters, how about  the timeseries contains asterisk like the following:
`root.sg."device_1*".sensor1`

wangchao316: This grammar....  I alter this, support this grammar...

wangchao316:
> Hi, it's better to add some ITs in testContainer for this function

hi , I add some IT in test Container, please review...  thanks.

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
```diff
@@ -1717,7 +1726,7 @@ public MNode getMNode(MNode deviceMNode, String measurementName) {
               () -> {
                 try {
                   showTimeseries(group, plan, resultSet, context);
-                } catch (CheckConsistencyException e) {
+                } catch (CheckConsistencyException | MetadataException e) {
```
mychaow, line 1729: LGTM. But there is one minor issues.  This will ignore some errors, resulting in query results have problems. Like we execute  show timeseries, some data group failed with other errors exclude tag not found, we will return result.

wangchao316: 1. We send requests to all DataGroups only for sequences that are not found in partitiontable.
2. The general metadata exception is because the DataGroup does not contain the sequence. Therefore, you can skip the metadata exception by default. add only the queried sequences.

LebronAl: I feel like the previous design here was to ignore all MetaExceptions and read as many as we could.
If we had no intention of changing the design, it would seem pointless to catch this exception?

testcontainer/src/test/java/org/apache/iotdb/db/sql/Cases.java
```diff
@@ -220,4 +220,61 @@ public void vectorCountTest() throws IoTDBConnectionException, StatementExecutio
...
+  @Test
+  public void showTimeseriesTagsTest() throws SQLException {
+    String createTimeSeries1 =
+        "create timeseries root.ln.wf01.wt1 WITH DATATYPE=DOUBLE, ENCODING=RLE, compression=SNAPPY tags(tag1=v1, tag2=v2)";
+    String createTimeSeries2 =
+        "create timeseries root.ln.wf01.wt2 WITH DATATYPE=DOUBLE, ENCODING=RLE, compression=SNAPPY tags(tag1=v1, tag2=v2)";
+    writeStatement.execute(createTimeSeries1);
+    writeStatement.execute(createTimeSeries2);
+    // try to read data on each node. select .*
+    for (Statement readStatement : readStatements) {
```
LebronAl, line 234: what about adding some corner cases,such as `SHOW TIMESERIES root.ln.wf01.* where tag1=v3`, `HOW TIMESERIES root.ln.wf01.* where tag3=v1`

wangchao316: Thanks, done.

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
wangchao316: Thanks , I alter this exception , Same as before.

mychaow (approved):
LGTM

neuyilan (approved):
LGTM

PR #3300, opened 2021-06-01
[IOTDB-1405] Refactor MManager for new vector interfaces

samperson1997:
Now that we have new interfaces of vector module, MManager module must be refactored.


example/session/src/main/java/org/apache/iotdb/VectorSessionExample.java
```diff
@@ -178,6 +180,23 @@ private static void selectWithLastTest()
     dataSet.closeOperationHandle();
   }
 
+  private static void createAlignedTimeseries()
+      throws StatementExecutionException, IoTDBConnectionException {
```
HTHou, line 184: How about changing the class name to `AlignedTimeseriesSessionExample`?

example/session/src/main/java/org/apache/iotdb/VectorSessionExample.java
```diff
@@ -217,9 +239,12 @@ private static void insertTabletWithAlignedTimeseriesMethod1()
...
     schemaList.add(
         new VectorMeasurementSchema(
-            new String[] {"s1", "s2"}, new TSDataType[] {TSDataType.INT64, TSDataType.INT32}));
+            "vector",
+            new String[] {"s1", "s2"},
+            new TSDataType[] {TSDataType.INT64, TSDataType.INT32}));
 
-    Tablet tablet = new Tablet(ROOT_SG1_D1, schemaList);
+    Tablet tablet = new Tablet(ROOT_SG1_D1 + ".vector", schemaList);
+    tablet.setAligned(true);
```
HTHou, line 247: @qiaojialin I remember we discussed before. We may use normal MeasurementSchema here. Like the code I shows below.
```java
List<IMeasurementSchema> schemaList = new ArrayList<>();
schemaList.add(
    new MeasurementSchema("s1",TSDataType.INT64));
schemaList.add(
    new MeasurementSchema("s2",TSDataType.INT32));

Tablet tablet = new Tablet(ROOT_SG1_D1 + ".vector", schemaList);
tablet.setAligned(true);
```

server/src/main/java/org/apache/iotdb/db/engine/memtable/AbstractMemTable.java
```diff
@@ -194,7 +194,7 @@ public void write(InsertTabletPlan insertTabletPlan, int start, int end) {
       }
       IWritableMemChunk memSeries =
           createIfNotExistAndGet(
-              insertTabletPlan.getDeviceId().getFullPath(),
+              insertTabletPlan.getPrefixPath().getFullPath(),
               insertTabletPlan.getMeasurementMNodes()[i].getSchema());
       if (insertTabletPlan.getMeasurementMNodes()[i].getSchema().getType() == TSDataType.VECTOR) {
         VectorMeasurementSchema vectorSchema =
```
HTHou, line 200: From this PR, there can be only one aligned timeseries in an insertTabletPlan, right? I think we can use the `isAligned` in insertTabletPlan to know if the type is vector or not. Using `MeasurementMNodes` is not a good choice.

example/session/src/main/java/org/apache/iotdb/VectorSessionExample.java
SilverNarcissus: This will be refactored by Zhao Xin later.

example/session/src/main/java/org/apache/iotdb/VectorSessionExample.java
SilverNarcissus: fixed~

server/src/main/java/org/apache/iotdb/db/engine/memtable/AbstractMemTable.java
SilverNarcissus: fixed~

HTHou (approved):

PR #3308, opened 2021-06-02
[IOTDB-1415] Fix OOM caused by ChunkCache

JackieTien97:
## Description
The detailed description could be seen in [JIRA](https://issues.apache.org/jira/browse/IOTDB-1415)

## Analysis
The details could be found in [Confluence](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=181307964)

## Solution
Use Caffeine Cache to replace the previous LinkedHashMap.

Caffeine is a high performance Java caching library providing a near optimal hit rate. Caffeine provides an in-memory cache using a Google Guava inspired API. The improvements draw on their experience designing Guava's cache and ConcurrentLinkedHashMap.

The [Tests](https://github.com/ben-manes/caffeine/wiki/Benchmarks) in their website shows that it's much more better than Guava Cache.

In our own [weekly test](http://ise.thss.tsinghua.edu.cn/confluence/pages/viewpage.action?pageId=57212546), Caffeine Cache's performance is same as the previous wrong implementation of master.


server/src/main/java/org/apache/iotdb/db/engine/cache/ChunkCache.java
```diff
@@ -49,39 +51,64 @@
...
-              count = 1;
-              currentSize = averageSize;
-            }
-            return currentSize;
-          }
-        };
+        Caffeine.newBuilder()
+            .maximumWeight(MEMORY_THRESHOLD_IN_CHUNK_CACHE)
+            .weigher(
+                new Weigher<ChunkMetadata, Chunk>() {
```
wangchao316, line 66: hello,  this is good job.
I have a little question. Whether to add an idle release. If a chunk is not used for a long time, it can be cleared. Kind of like guava expireAfterAccess(long, TimeUnit)

JackieTien97: It chould be added, but the memory allocated for the cache is fixed, e.g. if we allocate 5MB for cache and even if we can release some part of these 5MB that are not used for a long time, the released memory won't be used by other part of the IoTDB. On the other hand, the expireAfterAccess will cause some overhead.

wangchao316: The data that is not used for a long time is cold data. If the data is not deleted at this time, the data can be replaced only when the threshold is reached. In this case, the memory data replacement time during data reading is prolonged. We can eliminate this data at our leisure.

HTHou (approved):

PR #3304, opened 2021-06-01
[IOTDB-1414]NPE occurred when call getStorageGroupNodeByPath() method using not exist path

neuyilan:

PR #3311, opened 2021-06-02
[To rel/0.12][IOTDB-1414]NPE occurred when call getStorageGroupNodeByPath() method using not exist path

neuyilan:
https://github.com/apache/iotdb/pull/3304/files


mychaow (approved):
LGTM

Issue #3316, opened 2021-06-02
Query result with the same time range is inconsistent in group by query


Alima777:
select count(Channel27) from root.group7.J053102.T034.V2 group by ([2021-06-01 22:39:00,2021-06-01 22:40:00),1m)

select count(Channel27) from root.group7.J053102.T034.V2 group by ([2021-06-01 22:39:00,2021-06-01 22:41:00),1m)

![image](https://user-images.githubusercontent.com/34242296/120471428-4781b580-c3d7-11eb-96d9-84d50a3107f0.png)

PR #3317, opened 2021-06-02
[ISSUE-3316] Query result with the same time range is inconsistent in group by query 

Alima777:
The bug is caused by: when overlapped chunks are unpacked, firstChunk should be updated. #3316


Alima777:
DO NOT forget to cherry pick to 0.11/0.12

JackieTien97 (approved):

PR #3296, opened 2021-06-01
[IOTDB-1412] Unclear exception message thrown when executing empty InsertTabletPlan

SteveYurongSu:
When the system executes an empty InsertTabletPlan, the system will throw the following error which is unclear:

image

Empty insertTabletPlan / insertRowsOfOneDevicePlan are now handled properly in PlanExecutor.

PR #3320, opened 2021-06-03
[To rel/0.12][IOTDB-1412] Unclear exception message thrown when executing empty InsertTabletPlan (#3296)

SteveYurongSu:
See #3296 .


LebronAl (commented):
How about adding these logic to `plan.checkIntegrity`

PR #3216, opened 2021-05-18
[To rel/0.12]Upgrade libthrift from 0.12.0/0.13.0 to 0.14.1 and updated related classes or files

neuyilan:
Cherry-pick the following two PRs
https://github.com/apache/iotdb/pull/3163
https://github.com/apache/iotdb/pull/3198

PR #3328, opened 2021-06-03
enable the client library test for all changes

jixuan1989:
The client library CI was enabled only when some folders have some changes:
- client-*
- thrift-*
- service-rpc

However, it is not enough because even though the client library does not change, the server's change may occur some errors.

So, in this PR, I remove the trigger constraint.

### DIFF
--- a/.github/workflows/client.yml
+++ b/.github/workflows/client.yml
@@ -89,7 +89,7 @@ jobs:
         run:  choco install openssl
       - name: Add Flex and Bison Path and OpenSSL
         shell: bash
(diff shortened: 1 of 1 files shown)


neuyilan:
Excellent!!!!
cherry-pick to rel/0.12 and I think the PR[1] can be merged
[1] https://github.com/apache/iotdb/pull/3216

HTHou:
> Excellent!!!!
> cherry-pick to rel/0.12 and I think the PR[1] can be merged
> [1] #3216

OK, I'm cherry-picking.

PR #3364, opened 2021-06-07
Remove the broken cached leader connection & optimize the insertRecords method in session

neuyilan:
1. remove the broken session when enable cache leader
2. optimize the `insertRecords` method in session, which we can insert all the devices of data in one session connection, in the previous degien, if some of the devices belong to one session connection, we send to the server per device per connection.

### DIFF
--- a/service-rpc/src/main/java/org/apache/iotdb/rpc/RpcUtils.java
+++ b/service-rpc/src/main/java/org/apache/iotdb/rpc/RpcUtils.java
@@ -95,6 +95,22 @@ public class RpcUtils {
     }
   }
 
+  public static void verifySuccessWithRedirectionForMultiDevices(
+      TSStatus status, List<String> devices) throws StatementExecutionException, RedirectException {
+    verifySuccess(status);
+    if (status.getCode() == TSStatusCode.MULTIPLE_ERROR.getStatusCode()) {
+      Map<String, EndPoint> deviceEndPointMap = new HashMap<>();
+      List<TSStatus> statusSubStatus = status.getSubStatus();
+      for (int i = 0; i < statusSubStatus.size(); i++) {
+        TSStatus subStatus = statusSubStatus.get(i);
+        if (subStatus.isSetRedirectNode()) {
+          deviceEndPointMap.put(devices.get(i), subStatus.getRedirectNode());
+        }
+      }
+      throw new RedirectException(deviceEndPointMap);
+    }
+  }
+
   public static void verifySuccessWithRedirectionForInsertTablets(
       TSStatus status, TSInsertTabletsReq req)
       throws StatementExecutionException, RedirectException {
--- a/service-rpc/src/main/java/org/apache/iotdb/rpc/RpcUtils.java
+++ b/service-rpc/src/main/java/org/apache/iotdb/rpc/RpcUtils.java
@@ -22,7 +22,6 @@ import org.apache.iotdb.service.rpc.thrift.EndPoint;
 import org.apache.iotdb.service.rpc.thrift.TSExecuteStatementResp;
 import org.apache.iotdb.service.rpc.thrift.TSFetchResultsResp;
 import org.apache.iotdb.service.rpc.thrift.TSIService;
-import org.apache.iotdb.service.rpc.thrift.TSInsertTabletsReq;
 import org.apache.iotdb.service.rpc.thrift.TSStatus;
 
 import java.lang.reflect.Proxy;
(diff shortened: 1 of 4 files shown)


mychaow (approved):
LGTM

session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -1257,13 +1283,15 @@ private void insertTabletsWithLeaderCache(Map<String, Tablet> tablets, boolean s
...
     Map<SessionConnection, TSInsertTabletsReq> tabletGroup = new HashMap<>();
+    Map<SessionConnection, EndPoint> sessionConnectionEndPointMap = new HashMap<>();
     for (Entry<String, Tablet> entry : tablets.entrySet()) {
       endPoint = deviceIdToEndpoint.get(entry.getKey());
       if (endPoint != null) {
         connection = endPointToSessionConnection.get(endPoint);
       } else {
         connection = defaultSessionConnection;
       }
+      sessionConnectionEndPointMap.putIfAbsent(connection, endPoint);
```
LebronAl, line 1294: As `IoTDBConnectionException` is rarely thrown, I doubt whether this is worth recording even without errors, maybe we can traverse `endPointToSessionConnection` to delete expired sessionConnection when `IoTDBConnectionException` is thrown.

session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -1096,6 +1118,10 @@ private void insertRecordsWithLeaderCache(
         handleRedirection(entry.getKey(), e.getEndPoint());
       } catch (StatementExecutionException e) {
         errMsgBuilder.append(e.getMessage());
+      } catch (IoTDBConnectionException e) {
```
LebronAl, line 1121: Maybe we can optimize insertRecords interface to perfrom one request per sessionConnection rather than per deviceID, just like insertTablets.

LebronAl (commented):
How about adding some UTs for `removeBrokenSessionConnection` in `SessionCacheLeaderUT`.

session/src/main/java/org/apache/iotdb/session/Session.java
neuyilan: That's another solution. the original though is as the number of connections is raly small, so I think it ok add one map here and trade space for time.
As you say, maybe `IoTDBConnectionException` is rarely thrown, so I agree with you.

neuyilan:
> How about adding some UTs for `removeBrokenSessionConnection` in `SessionCacheLeaderUT`.

Sure

session/src/main/java/org/apache/iotdb/session/Session.java
neuyilan: Ok I'll optimize later in this PR.

session/src/test/java/org/apache/iotdb/session/SessionCacheLeaderUT.java
```diff
@@ -738,6 +1073,8 @@ private void addLine(
 
   static class MockSession extends Session {
 
+    private MockSessionConnection mockSessionConnection;
```
LebronAl, line 1076: The name of this `mockSessionConnection` is misleading, I didn't understand the meaning until after debugging. I think we can change this name to a better one, such as `lastConstructedSessionConnection`

service-rpc/src/main/java/org/apache/iotdb/rpc/RpcUtils.java
```diff
@@ -95,6 +95,22 @@ public static void verifySuccessWithRedirection(TSStatus status)
...
+        TSStatus subStatus = statusSubStatus.get(i);
+        if (subStatus.isSetRedirectNode()) {
+          deviceEndPointMap.put(devices.get(i), subStatus.getRedirectNode());
+        }
+      }
+      throw new RedirectException(deviceEndPointMap);
+    }
+  }
+
   public static void verifySuccessWithRedirectionForInsertTablets(
```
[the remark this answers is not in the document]
neuyilan: sure

session/src/test/java/org/apache/iotdb/session/SessionCacheLeaderUT.java
neuyilan: have changed

session/src/test/java/org/apache/iotdb/session/SessionCacheLeaderUT.java
```diff
@@ -709,6 +712,338 @@ public void testInsertTablets() throws IoTDBConnectionException, StatementExecut
...
+        tablet2.reset();
+        tablet3.reset();
+      }
+      timestamp++;
+    }
+
+    // set the session connection as broken
+    ((MockSession) session).getLastConstructedSessionConnection().setConnectionBroken(true);
+    // set connection as broken, due to we enable the cache leader, when we called
+    // ((MockSession) session).getMockSessionConnection(), the session's endpoint have changed to
```
LebronAl, line 993:
```suggestion
// ((MockSession) session). getLastConstructedSessionConnection(), the session's endpoint has been changed to
```

session/src/test/java/org/apache/iotdb/session/SessionCacheLeaderUT.java
```diff
@@ -709,6 +712,338 @@ public void testInsertTablets() throws IoTDBConnectionException, StatementExecut
...
+        }
+        deviceIds.clear();
+        measurementsList.clear();
+        valuesList.clear();
+        timestamps.clear();
+      }
+    }
+
+    // set connection as broken, due to we enable the cache leader, when we called
+    // ((MockSession) session).getMockSessionConnection(), the session's endpoint have changed to
```
LebronAl, line 839:
```suggestion
// ((MockSession) session). getLastConstructedSessionConnection(), the session's endpoint has been changed to
```

LebronAl (approved):
LGTM

PR #3365, opened 2021-06-07
Fix compaction lock with query

JackieTien97:
## Problem

When Level compaction is ending(delete file in resource list and delete file in disk). The query may get a tsfile resource from resource list. However, it may be deleted by the compaction ending process later, then the query process cannot read data from the deleted resource. Whichmay cause a ci problem like below.
3501622863945_ pic_hd

## Solution

Get the tsfileResource list's read lock in TsFileManagament before trying to lock each tsfile and after locking all the tsfiles, we then unlock the tsfileResource list's read lock in TsFileManagament.


zhanglingzhe0820 (approved):
LGTM

qiaojialin (approved):

PR #3367, opened 2021-06-07
set TTL on any prefix path

bigreybear:
SET TTL can be applied to root now


HTHou (commented):
Consider adding some UT/IT?

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
```diff
@@ -1152,8 +1152,16 @@ private void operateMoveFile(OperateFilePlan plan) throws QueryProcessException
 
   private void operateTTL(SetTTLPlan plan) throws QueryProcessException {
     try {
-      IoTDB.metaManager.setTTL(plan.getStorageGroup(), plan.getDataTTL());
-      StorageEngine.getInstance().setTTL(plan.getStorageGroup(), plan.getDataTTL());
+      List<PartialPath> storageGroupPaths = new ArrayList<>();
+      if (plan.getStorageGroup().equals("root")) {
```
wangchao316, line 1156: IoTDBConstant.PATH_ROOT    replace "root"

wangchao316 (changes requested):

jixuan1989 (changes requested):
see others' comments.

PR #3369, opened 2021-06-08
[IOTDB-1426] Restructure physicalGenerator

Alima777:
Move the logic of generating physical plan in physicalGenerator to each operator inside, mainly query operator, which is more maintainable.


server/src/main/java/org/apache/iotdb/db/qp/logical/Operator.java
```diff
@@ -64,22 +68,17 @@ public String toString() {
...
   /** If you want to add new OperatorType, you must add it in the last. */
   public enum OperatorType {
-    GROUPBYTIME,
     NULL,
-    INSERT,
-    BATCHINSERT,
-    DELETE,
-    QUERY,
-    AGGREGATION,
+
```
Alima777, line 79: Just modify the order of these type to make them more maintainable.

PR #3372, opened 2021-06-08
[IOTDB-1425]Privilege controller does not support some operators

yifuzhou:
### Content3 ...

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/auth/AuthorityChecker.java
+++ b/server/src/main/java/org/apache/iotdb/db/auth/AuthorityChecker.java
@@ -117,6 +117,8 @@ public class AuthorityChecker {
         return PrivilegeType.REVOKE_USER_ROLE.ordinal();
       case SET_STORAGE_GROUP:
         return PrivilegeType.SET_STORAGE_GROUP.ordinal();
+      case DELETE_STORAGE_GROUP:
+        return PrivilegeType.DELETE_STORAGE_GROUP.ordinal();
       case CREATE_TIMESERIES:
         return PrivilegeType.CREATE_TIMESERIES.ordinal();
       case DELETE_TIMESERIES:
@@ -138,6 +140,7 @@ public class AuthorityChecker {
       case INSERT:
       case LOADDATA:
       case CREATE_INDEX:
+      case ALTER_TIMESERIES:
         return PrivilegeType.INSERT_TIMESERIES.ordinal();
       case LIST_ROLE:
       case LIST_ROLE_USERS:
--- a/server/src/main/java/org/apache/iotdb/db/auth/entity/PrivilegeType.java
+++ b/server/src/main/java/org/apache/iotdb/db/auth/entity/PrivilegeType.java
@@ -21,6 +21,7 @@ package org.apache.iotdb.db.auth.entity;
 /** This enum class contains all available privileges in IoTDB. */
 public enum PrivilegeType {
   SET_STORAGE_GROUP,
+  DELETE_STORAGE_GROUP,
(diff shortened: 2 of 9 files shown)


server/src/main/java/org/apache/iotdb/db/auth/AuthorityChecker.java
```diff
@@ -138,6 +140,7 @@ private static int translateToPermissionId(Operator.OperatorType type) {
       case INSERT:
       case LOADDATA:
       case CREATE_INDEX:
+      case ALTER_TIMESERIES:
         return PrivilegeType.INSERT_TIMESERIES.ordinal();
```
jt2594838, line 144: This should be pointed out in the document that altering a timeseries requires a write privilege.

server/src/main/java/org/apache/iotdb/db/auth/entity/PrivilegeType.java
```diff
@@ -46,6 +45,7 @@
   DROP_TRIGGER,
   START_TRIGGER,
   STOP_TRIGGER,
+  TTL,
   ALL;
```
jt2594838, line 49: All new privileges should be appended to the tail as they are serialized according to their ordinals, changing the order may introduce compatibility problems.

jt2594838: By the way, have you mentioned the new privileges in the document?

jt2594838 (approved):

server/src/main/java/org/apache/iotdb/db/auth/entity/PrivilegeType.java
yifuzhou: I mentioned this new privilege in docs/UserGuide/Administration-Management/Administration.md.

server/src/main/java/org/apache/iotdb/db/auth/AuthorityChecker.java
yifuzhou: Can you tell me the place that I can mention it?

server/src/main/java/org/apache/iotdb/db/auth/entity/PrivilegeType.java
jixuan1989: > All new privileges should be appended to the tail as they are serialized according to their ordinals, changing the order may introduce compatibility problems.

So why you approve the PR though it is not changed?

jixuan1989 (changes requested):
Test needed. (coverage of new codes >=80%)

yifuzhou:
> Test needed. (coverage of new codes >=80%)

Done.

jixuan1989 (changes requested):
> All new privileges should be appended to the tail as they are serialized according to their ordinals, changing the order may introduce compatibility problems.

Why this is omitted..

yifuzhou:
> > All new privileges should be appended to the tail as they are serialized according to their ordinals, changing the order may introduce compatibility problems.
> 
> Why this is omitted..

You mean add after 'ALL'? I see other pr added the privilege type at the end but before 'ALL', I think it is more reasonable...

yifuzhou:
> > All new privileges should be appended to the tail as they are serialized according to their ordinals, changing the order may introduce compatibility problems.
> 
> Why this is omitted..

Done. I move new privileges to the end.

PR #3373, opened 2021-06-08
[To Rel/0.12]Remove the broken cached leader connection & optimize the insertRecords method in session 

neuyilan:
https://github.com/apache/iotdb/pull/3364


mychaow (approved):
LGTM

PR #3382, opened 2021-06-10
fix iotdb-env.bat not working

SiasDoming:
## Description
iotdb-env.bat doesn't work as expected on Windows systems due to local variable settings. This bug cause all environment variables cannot be assessed in start-server.bat, such as heap memory settings and JMX options. Besides, it's not able to calculate memory size correctly as iotdb-env.sh does due to syntax errors in lines 70, 71, 73, and 82.
### Variables Passing
Delete SETLOCAL clause to make sure all variables are passing to start-server.bat. Also, system information(number of cores and physical memory size) is acquired in another way because of this modification.
### Value Comparisons
Modify variable citations in GTR clauses to arithmetic operations instead of string comparisons.

### DIFF
--- a/server/src/assembly/resources/conf/iotdb-env.bat
+++ b/server/src/assembly/resources/conf/iotdb-env.bat
@@ -18,7 +18,6 @@
 @REM
 
 @echo off
-setlocal enabledelayedexpansion
 @REM true or false
 @REM DO NOT FORGET TO MODIFY THE PASSWORD FOR SECURITY (%IOTDB_CONF%\jmx.password and %{IOTDB_CONF%\jmx.access)
 set JMX_LOCAL="true"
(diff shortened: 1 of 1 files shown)


HTHou:
Hi, thanks so much for your contribution!

I find out there are some lines about `SetEnvironmentVariable` in Windows CI workflow. Is it the reason why the Windows tests failed? 

https://github.com/apache/iotdb/blob/fe983ed81b97aa8b232bbdde57df08f5e9c32b0d/.github/workflows/main-win.yml#L85-L88

HTHou (approved):
LGTM

SiasDoming:
> Hi, thanks so much for your contribution!
> 
> I find out there are some lines about `SetEnvironmentVariable` in Windows CI workflow. Is it the reason why the Windows tests failed?
> 
> https://github.com/apache/iotdb/blob/fe983ed81b97aa8b232bbdde57df08f5e9c32b0d/.github/workflows/main-win.yml#L85-L88

I'm not familiar with CI workflows, but it seems these lines are working with the Hadoop environment and I prefer that these have nothing to do with my modifications. All I've done is within the cope of iotdb-env.bat, which means even if the SETLOCAL clause is deleted, all variables are only valid in the lifecycle of the single command line. 
By the way, although iotdb-env.bat may modify environment variables within the command line, thus causing errors when some variables with the same names are vital for other commands or scripts, it's mostly called by start-server.bat and the latter has a SETLOCAL clause before calling it and this will strict all variables to be valid within the start-server.bat.

PR #3389, opened 2021-06-10
[To rel/0.12]TTL can be set to any path

bigreybear:


qiaojialin (commented):
Hi, the path in SetTTLPlan may be an arbitrary prefix path, not only root.

qiaojialin (approved):

PR #3391, opened 2021-06-11
[IOTDB-1433] Fix bug in getMetadataAndEndOffset when querying non-exist device

samperson1997:
When querying non-exist device, getMetadataAndEndOffset() may give a wrong result.

For example, this method may present results.
```
Iterator<Map<String, List<ChunkMetadata>>> iterator =
          fileReader.getMeasurementChunkMetadataListMapIterator("dd");
      Assert.assertFalse(iterator.hasNext());
```


SteveYurongSu (approved):
LGTM

PR #3407, opened 2021-06-15
Complete syntax sugar for createSchemaTemplate

bigreybear:


HTHou:
Any api javaDoc? 
It's a little bit hard to understand the design and the usage of these interfaces....

HTHou (changes requested):
I got the design doc.
```
void createSchemaTemplate(
String templateName,
List<IMeasurementSchema>
)

VectorMeasurementSchema(String prefixPath, List<String> measurementNames, List<TSDataType> datatypes, List<TSEncoding> encodings)

MeasurementSchema(Sting path, TSDatatype datatype, TSEncoding encoding)
```
I think you may misunderstand the purpose of this PR.

What you should do in this PR is creating only one new interface `createSchemaTemplate`.
The `IMeasurementSchema` is an interface that could be a VectorMeasurementSchema or a `MeasurementSchema`. 

`VectorMeasurementSchema(String prefixPath, List<String> measurementNames, List<TSDataType> datatypes, List<TSEncoding> encodings)` and `MeasurementSchema(Sting path, TSDatatype datatype, TSEncoding encoding)` are trying to tell you what informations are in these classes.

You cannot use the singletonList for datatypes and encodings directly.

session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -1790,6 +1790,81 @@ public void createSchemaTemplate(
...
+    List<List<String>> measurements = new ArrayList<>();
+    List<List<TSDataType>> dataTypes = new ArrayList<>();
+    List<List<TSEncoding>> encodings = new ArrayList<>();
+
+    List<CompressionType> compressors = new ArrayList<>();
+    for (IMeasurementSchema measurementSchema : measurementSchemas) {
+      schemaNames.add(measurementSchema.getMeasurementId());
+      measurements.add(Collections.singletonList(measurementSchema.getMeasurementId()));
+      dataTypes.add(Collections.singletonList(measurementSchema.getType()));
+      encodings.add(Collections.singletonList(measurementSchema.getEncodingType()));
```
HTHou, line 1805: When the measurementSchemas is a VectorMeasurementSchema, the `Collections.singletonList` seems not correct.

session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -1790,6 +1790,81 @@ public void createSchemaTemplate(
...
+    encodings.add(Collections.singletonList(encoding));
+    compressors.add(compressor);
+
+    TSCreateSchemaTemplateReq request =
+        getTSCreateSchemaTemplateReq(
+            name, schemaNames, measurements, dataTypes, encodings, compressors);
+    defaultSessionConnection.createSchemaTemplate(request);
+  }
+
+  public void createVectorSchemaTemplate(
```
HTHou, line 1841: There two methods are not necessary.

bigreybear: deleted

session/src/main/java/org/apache/iotdb/session/Session.java
bigreybear: edited as required

HTHou (commented):
Almost good to me. 
I think we can also add a usage example in alignedTimeseriesSessionExample.

PR #3434, opened 2021-06-22
[IOTDB-1399]Add a session interface to connect multiple nodes

ijihang:
While the cluster is running, the coordinator node may go down, which may cause the session to be unable to read or write properly.i add a session interface to connect multiple nodes.Users can now define a list that contains multiple servers that can be connected. When one of the servers goes down, try again. If it fails, connect to the next one. If all the nodes in the list fail, an exception will be thrown.

### DIFF
--- a/session/src/main/java/org/apache/iotdb/session/Session.java
+++ b/session/src/main/java/org/apache/iotdb/session/Session.java
@@ -584,16 +584,12 @@ public class Session {
   private SessionDataSet executeStatementMayRedirect(String sql, long timeoutInMs)
       throws StatementExecutionException, IoTDBConnectionException {
     try {
-      logger.info("{} execute sql {}", defaultSessionConnection.getEndPoint(), sql);
+      logger.info("{} execute sql {}", defaultEndPoint, sql);
       return defaultSessionConnection.executeQueryStatement(sql, timeoutInMs);
     } catch (RedirectException e) {
       handleQueryRedirection(e.getEndPoint());
       if (enableQueryRedirection) {
-        logger.debug(
-            "{} redirect query {} to {}",
-            defaultSessionConnection.getEndPoint(),
-            sql,
-            e.getEndPoint());
+        logger.debug("{} redirect query {} to {}", defaultEndPoint, sql, e.getEndPoint());
         // retry
         try {
           return defaultSessionConnection.executeQueryStatement(sql, timeout);
--- a/session/src/main/java/org/apache/iotdb/session/Session.java
+++ b/session/src/main/java/org/apache/iotdb/session/Session.java
@@ -253,6 +253,11 @@ public class Session {
         Config.DEFAULT_CACHE_LEADER_MODE);
   }
 
+  /**
+   * Multiple nodeUrl,If one node down, connect to the next one
+   *
+   * @param nodeUrls List<String> Multiple ip:rpcPort eg.127.0.0.1:9001
+   */
   public Session(List<String> nodeUrls, String username, String password, int fetchSize) {
     this(
         nodeUrls,
@@ -589,7 +594,7 @@ public class Session {
     } catch (RedirectException e) {
       handleQueryRedirection(e.getEndPoint());
       if (enableQueryRedirection) {
-        logger.debug("{} redirect query {} to {}", defaultEndPoint, sql, e.getEndPoint());
+        logger.debug("redirect query {} to {}", sql, e.getEndPoint());
         // retry
         try {
           return defaultSessionConnection.executeQueryStatement(sql, timeout);
--- a/session/src/main/java/org/apache/iotdb/session/Session.java
+++ b/session/src/main/java/org/apache/iotdb/session/Session.java
@@ -589,12 +589,16 @@ public class Session {
   private SessionDataSet executeStatementMayRedirect(String sql, long timeoutInMs)
       throws StatementExecutionException, IoTDBConnectionException {
     try {
-      logger.info("{} execute sql {}", defaultEndPoint, sql);
+      logger.info("{} execute sql {}", defaultSessionConnection.getEndPoint(), sql);
       return defaultSessionConnection.executeQueryStatement(sql, timeoutInMs);
     } catch (RedirectException e) {
       handleQueryRedirection(e.getEndPoint());
       if (enableQueryRedirection) {
-        logger.debug("redirect query {} to {}", sql, e.getEndPoint());
+        logger.debug(
+            "{} redirect query {} to {}",
+            defaultSessionConnection.getEndPoint(),
+            sql,
+            e.getEndPoint());
         // retry
         try {
           return defaultSessionConnection.executeQueryStatement(sql, timeout);
--- a/session/src/main/java/org/apache/iotdb/session/Session.java
+++ b/session/src/main/java/org/apache/iotdb/session/Session.java
@@ -75,6 +75,7 @@ public class Session {
   public static final String MSG_UNSUPPORTED_DATA_TYPE = "Unsupported data type:";
   public static final String MSG_DONOT_ENABLE_REDIRECT =
       "Query do not enable redirect," + " please confirm the session and server conf.";
+  protected List<String> nodeUrls;
   protected String username;
   protected String password;
   protected int fetchSize;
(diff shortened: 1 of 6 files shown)


session/src/main/java/org/apache/iotdb/session/SessionConnection.java
```diff
@@ -722,20 +751,27 @@ protected void testInsertTablets(TSInsertTabletsReq request)
   private boolean reconnect() {
     boolean flag = false;
     for (int i = 1; i <= Config.RETRY_NUM; i++) {
```
mychaow, line 753: Always try the first RETRY_NUM clients, it's not good. Maybe you can record a nextHostIndex to try next hostId.

session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -240,6 +241,70 @@ public Session(
...
+    this.username = username;
+    this.password = password;
+    this.fetchSize = fetchSize;
+    this.zoneId = zoneId;
+    this.thriftDefaultBufferSize = thriftDefaultBufferSize;
+    this.thriftMaxFrameSize = thriftMaxFrameSize;
+    this.enableCacheLeader = enableCacheLeader;
+  }
+
+  public synchronized void clusterOpen() throws IoTDBConnectionException {
```
mychaow, line 275: I think it's no need to create a new open function, you could put the logic which just create SessionConnection who use the nodeUrl in the contructSessionConn.

mychaow (changes requested):

session/src/main/java/org/apache/iotdb/session/SessionConnection.java
ijihang: in new commit,Instead of trying the first node in the list every time, try the current node again, and then try the next node in the list

session/src/main/java/org/apache/iotdb/session/Session.java
ijihang: ok,i modify in new commit,using the old open method

session/src/main/java/org/apache/iotdb/session/SessionConnection.java
```diff
@@ -722,20 +748,28 @@ protected void testInsertTablets(TSInsertTabletsReq request)
...
-      } catch (Exception e) {
-        try {
-          Thread.sleep(Config.RETRY_INTERVAL_MS);
-        } catch (InterruptedException e1) {
-          logger.error("reconnect is interrupted.", e1);
-          Thread.currentThread().interrupt();
+      if (transport != null) {
+        transport.close();
+        int currHostIndex = endPointList.indexOf(session.defaultEndPoint);
+        for (int j = currHostIndex; j < endPointList.size(); j++) {
```
mychaow, line 754: not right, can not try the previous endpoint.

mychaow (changes requested):

session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -525,16 +584,12 @@ public SessionDataSet executeQueryStatement(String sql, long timeoutInMs)
   private SessionDataSet executeStatementMayRedirect(String sql, long timeoutInMs)
       throws StatementExecutionException, IoTDBConnectionException {
     try {
-      logger.info("{} execute sql {}", defaultSessionConnection.getEndPoint(), sql);
+      logger.info("{} execute sql {}", defaultEndPoint, sql);
```
neuyilan, line 587: `defaultSessionConnection.getEndPoint() `may not equal `defaultEndPoint` as long as we enable redirection

session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -525,16 +584,12 @@ public SessionDataSet executeQueryStatement(String sql, long timeoutInMs)
...
       return defaultSessionConnection.executeQueryStatement(sql, timeoutInMs);
     } catch (RedirectException e) {
       handleQueryRedirection(e.getEndPoint());
       if (enableQueryRedirection) {
-        logger.debug(
-            "{} redirect query {} to {}",
-            defaultSessionConnection.getEndPoint(),
-            sql,
-            e.getEndPoint());
+        logger.debug("{} redirect query {} to {}", defaultEndPoint, sql, e.getEndPoint());
```
neuyilan, line 592: the same as above

session/src/main/java/org/apache/iotdb/session/SessionUtils.java
```diff
@@ -149,4 +158,32 @@ private static void getValueBufferOfDataType(
...
   }
+
+  public static List<EndPoint> parseSeedNodeUrls(List<String> nodeUrls) {
+    if (nodeUrls == null) {
+      return Collections.emptyList();
+    }
+    List<EndPoint> endPointsList = new ArrayList<>();
+    for (String nodeUrl : nodeUrls) {
+      EndPoint endPoint = parseNodeUrl(nodeUrl);
+      endPointsList.add(endPoint);
```
neuyilan, line 169: `parseNodeUrl`  method may return `null`, we can not add one `null` endpoint to the `endPointsList`

session/src/main/java/org/apache/iotdb/session/SessionUtils.java
```diff
@@ -149,4 +158,32 @@ private static void getValueBufferOfDataType(
...
+    }
+    String ip = split[0];
+    try {
+      int rpcPort = Integer.parseInt(split[1]);
+      return endPoint.setIp(ip).setPort(rpcPort);
+    } catch (NumberFormatException e) {
+      logger.warn("parse url fail {}", nodeUrl);
+    }
+    return endPoint;
+  }
```
neuyilan, line 188: If the user given one wrong URL format,  in the code implement, the `NumberFormatException` may occur, you will return one `empty` endPoint.
In my opinion, if the seed URL format error, It's better rethrow the exception to the user.

session/src/main/java/org/apache/iotdb/session/SessionConnection.java
```diff
@@ -145,6 +153,24 @@ private void init(EndPoint endPoint) throws IoTDBConnectionException {
     }
   }
 
+  private void initClusterConn() throws IoTDBConnectionException {
+    for (EndPoint endPoint : endPointList) {
+      if (endPoint == null) {
+        continue;
+      }
```
neuyilan, line 160: Its better not allow `null` endPoint added in the `endPointList` in the previous `parseSeedNodeUrls` method. In other words, abnormal judgment should be made as soon as possible.

session/src/main/java/org/apache/iotdb/session/SessionConnection.java
neuyilan: If `currHostIndex` is the last index of `endPointList`, in your implementation, it can not try the others nodes again.

neuyilan (changes requested):

session/src/main/java/org/apache/iotdb/session/SessionConnection.java
ijihang: If all the endpoints in the list fail, I will try to connect again with the head node, traversing the list RETRY_NUM times in total

session/src/main/java/org/apache/iotdb/session/SessionConnection.java
```diff
@@ -721,20 +746,28 @@ protected void testInsertTablets(TSInsertTabletsReq request)
...
     boolean flag = false;
-    for (int i = 1; i <= Config.RETRY_NUM; i++) {
-      try {
-        if (transport != null) {
-          close();
-          init(endPoint);
-          flag = true;
+    if (transport != null) {
+      transport.close();
+      int currHostIndex = endPointList.indexOf(session.defaultEndPoint);
```
mychaow, line 751: currHostIndex maybe -1, so you should handle it.

session/src/main/java/org/apache/iotdb/session/SessionConnection.java
```diff
@@ -722,20 +747,38 @@ protected void testInsertTablets(TSInsertTabletsReq request)
...
+              "endPoint:{} not in endPointList,Try the first one in the endPointList", endPoint);
+          currHostIndex = 0;
         }
-      } catch (Exception e) {
-        try {
-          Thread.sleep(Config.RETRY_INTERVAL_MS);
-        } catch (InterruptedException e1) {
-          logger.error("reconnect is interrupted.", e1);
-          Thread.currentThread().interrupt();
+        int tag = 0;
```
mychaow, line 758: tag->tryHostNum

session/src/main/java/org/apache/iotdb/session/SessionConnection.java
```diff
@@ -722,20 +747,38 @@ protected void testInsertTablets(TSInsertTabletsReq request)
   private boolean reconnect() {
     boolean flag = false;
```
mychaow, line 748: flag -> connectedSuccess

mychaow (changes requested):

neuyilan (approved):
LGTM

session/src/main/java/org/apache/iotdb/session/SessionConnection.java
ijihang: ok, in new commit,connect nodes randomly

mychaow (approved):
LGTM

PR #3435, opened 2021-06-22
Fix Bug: delete timeseries with multiple time interval

Alima777:
When delete timeseries with multiple time interval, only the first delete interval is checked, which causes chunks that have been deleted are read.

#3436 

You can see the IT to reappear this bug.


HTHou:
Do not forget to cherry pick to rel/0.11 or rel/0.12.

PR #3441, opened 2021-06-23
MergeReader overlap check logic in SeriesReader

LeiRui:
As shown in the figure below, the original mergeReader overlap check logic in SeriesReader seems not to include the desc time order case. So this pr aims to fix this potential problem.
image
TODO: IT/UT is needed to expose this potential problem and to verify this pr solution. The design of the IT/UT is underway.

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


Alima777:
We still think there is potential problem though we have not found it. So we decided to fix it in #4819 . I will close this PR.

PR #3452, opened 2021-06-25
[IOTDB-1453]Fix result set when the server query time filtered is And

ijihang:
jira:https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-1453?filter=allissues
Test Query statement is：select * from root.ln.wf01.wt01 where time <5 and time > 10
cluster version result set is：
image
server version result set is:
image

I think the cluster result set is true.


jixuan1989:
Hi, IMO, the server version is better...  considering you are using a relational database, Oracel or MySQL, and query using the SQL: `select * from table where id >5 and id <3`, will you get an exception?

But, we indeed can optimize the execution logic, to return empty resultset ASAP.

ijihang:
exactly,if Oracle or MySQL ,the  return is empty,now  I tried to return the result set to null

server/src/main/java/org/apache/iotdb/db/utils/TimeValuePairUtils.java
```diff
@@ -107,4 +115,228 @@ public static TimeValuePair getEmptyTimeValuePair(TSDataType dataType) {
         throw new UnsupportedOperationException("Unrecognized datatype: " + dataType);
     }
   }
+
```
neuyilan, line 118: Is those code duplicated with the `org.apache.iotdb.cluster.utils.PartitionUtils.Intervals` ? if yes, why repeat coding in two places？
Don't repeat yourself.

neuyilan (changes requested):

server/src/main/java/org/apache/iotdb/db/utils/TimeValuePairUtils.java
```diff
@@ -107,4 +115,228 @@ public static TimeValuePair getEmptyTimeValuePair(TSDataType dataType) {
...
+        }
+      }
+
+      long lastUB = getUpperBound(result.getIntervalSize() - 1);
+      if (lastUB != Long.MAX_VALUE) {
+        result.addInterval(lastUB + 1, Long.MAX_VALUE);
+      }
+      return result;
+    }
+  }
```
neuyilan, line 341: Is those code duplicated with the org.apache.iotdb.cluster.utils.PartitionUtils.Intervals ? if yes, why repeat coding in two places？
Don't repeat yourself.

neuyilan (changes requested):

server/src/main/java/org/apache/iotdb/db/utils/TimeValuePairUtils.java
ijihang: I'll make some improvements

neuyilan (approved):
LGTM

mychaow (approved):
LGTM. And please cherry-pick to 0.12

PR #3459, opened 2021-06-28
fix monitor bug

yanhongwangg:
There are two problems;

The number of successful requests is counted when the request fails
Start statistics of performance monitoring does not take effect


mychaow (approved):
LGTM

PR #3269, opened 2021-05-26
[IOTDB-1292] Versioned MTree

neuyilan:
https://issues.apache.org/jira/browse/IOTDB-1292

design doc
https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=177050789

PR #3474, opened 2021-06-30
[IOTDB-1456] Fix Error occurred while executing delete timeseries statement

cmlmakahts:
Error was thrown in cluster mode when executing delete timeseries statement.
It happended frequently in the condition that ReplicateNum greater than node count.
It seems that batch exception is not handled correctly, timeseries does not exist exception should be skipped when delete timeseries plan is executing.

More details in https://issues.apache.org/jira/browse/IOTDB-1456

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/log/applier/BaseApplier.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/log/applier/BaseApplier.java
@@ -38,6 +38,7 @@ import org.apache.iotdb.db.qp.executor.PlanExecutor;
 import org.apache.iotdb.db.qp.physical.BatchPlan;
 import org.apache.iotdb.db.qp.physical.PhysicalPlan;
 import org.apache.iotdb.db.qp.physical.crud.InsertPlan;
+import org.apache.iotdb.db.qp.physical.sys.DeleteTimeSeriesPlan;
 import org.apache.iotdb.db.utils.SchemaUtils;
 import org.apache.iotdb.rpc.TSStatusCode;
 import org.apache.iotdb.service.rpc.thrift.TSStatus;
@@ -94,6 +95,7 @@ abstract class BaseApplier implements LogApplier {
   private void handleBatchProcessException(BatchProcessException e, PhysicalPlan plan)
       throws QueryProcessException, StorageEngineException, StorageGroupNotSetException {
     TSStatus[] failingStatus = e.getFailingStatus();
+    boolean needThrow = false;
     for (int i = 0; i < failingStatus.length; i++) {
       TSStatus status = failingStatus[i];
       // skip succeeded plans in later execution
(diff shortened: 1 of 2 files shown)


cluster/src/main/java/org/apache/iotdb/cluster/log/applier/BaseApplier.java
```diff
@@ -102,6 +104,16 @@ private void handleBatchProcessException(BatchProcessException e, PhysicalPlan p
           && plan instanceof BatchPlan) {
         ((BatchPlan) plan).setIsExecuted(i);
       }
+
+      if (plan instanceof DeleteTimeSeriesPlan) {
+        if (status != null && status.getCode() != TSStatusCode.SUCCESS_STATUS.getStatusCode()) {
+          if (status.getCode() != TSStatusCode.TIMESERIES_NOT_EXIST.getStatusCode()) {
```
LebronAl, line 110: 'status.getCode() == TSStatusCode.TIMESERIES_NOT_EXIST.getStatusCode()' ?

cluster/src/main/java/org/apache/iotdb/cluster/log/applier/BaseApplier.java
```diff
@@ -102,6 +104,16 @@ private void handleBatchProcessException(BatchProcessException e, PhysicalPlan p
...
+            logger.info("{} doesn't exist, it may has been deleted.", plan.getPaths().get(i));
+          } else {
+            needThrow = true;
```
LebronAl, line 113: As delete operation doesn't need to be performd once again, maybe we can directly throw exception here ?

LebronAl (commented):
Great work!
Actually, there are some more compliated situations for delete operations in [PR3269](https://github.com/apache/iotdb/pull/3269)
BTW, Please add some ITs in testcontainer for cluster to convince your changes !

cluster/src/main/java/org/apache/iotdb/cluster/log/applier/BaseApplier.java
cmlmakahts: 1. First mistake has been modified.(status.getCode() == TSStatusCode.TIMESERIES_NOT_EXIST.getStatusCode())
2.Delete operation won't be performed once again, but if exeception is thrown directly here, batch exception won't be catched correctly in coordinator.
image

wangchao316 (approved):
LGTM

cluster/src/main/java/org/apache/iotdb/cluster/log/applier/BaseApplier.java
LebronAl: Sorry I don't get your meaning.
It seems once we set `needThrow` to `true` here, It will always be thrown in the last?  below codes are irrelevant with  delete operations, so I still don't know why we can not throw this exception here or before below codes?
```
boolean needRetry = false;
    for (int i = 0, failingStatusLength = failingStatus.length; i < failingStatusLength; i++) {
      TSStatus status = failingStatus[i];
      if (status != null
          && (status.getCode() == TSStatusCode.STORAGE_GROUP_NOT_EXIST.getStatusCode()
              || status.getCode() == TSStatusCode.UNDEFINED_TEMPLATE.getStatusCode())
          && plan instanceof BatchPlan) {
        ((BatchPlan) plan).unsetIsExecuted(i);
        needRetry = true;
      }
    }
    if (needRetry) {
      executeAfterSync(plan);
      return;
    }
```

cmlmakahts: Delete operations indeed succeed, but the client may receives a error message which may confuse user. 


image
2. 506

LebronAl: OK, I see your meaning : DeleteOpearion can not throw exception here.

LebronAl (approved):
LGTM

PR #3475, opened 2021-06-30
[To rel/0.12] [IOTDB-1456] Fix Error occurred while executing delete timeseries statement

cmlmakahts:
Error was thrown in cluster mode when executing delete timeseries statement.
It happended frequently in the condition that ReplicateNum greater than node count.
It seems that batch exception is not handled correctly, timeseries does not exist exception should be skipped when delete timeseries plan is executing.

More details in https://issues.apache.org/jira/browse/IOTDB-1456


wangchao316 (approved):
LGTM

PR #3478, opened 2021-07-01
Fix Windows shell error on JDK11

HTHou:
## Description
When users use the JDK11 on Windows, the start-up script will show an error message like this.

5841625100755_ pic_hd

### DIFF
--- a/server/src/assembly/resources/conf/iotdb-env.bat
+++ b/server/src/assembly/resources/conf/iotdb-env.bat
@@ -78,9 +78,9 @@ if %desired_yg_in_mb% GTR %max_sensible_yg_in_mb% (
 ) else set HEAP_NEWSIZE=%desired_yg_in_mb%M
 
 @REM Maximum heap size
-@REM MAX_HEAP_SIZE="2G"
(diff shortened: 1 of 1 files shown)


server/src/assembly/resources/conf/iotdb-env.bat
```diff
@@ -77,6 +77,11 @@ if %desired_yg_in_mb% GTR %max_sensible_yg_in_mb% (
 	set HEAP_NEWSIZE=%max_sensible_yg_in_mb%M
 ) else set HEAP_NEWSIZE=%desired_yg_in_mb%M
 
+@REM Maximum heap size
+@REM MAX_HEAP_SIZE="2G"
+@REM Minimum heap size
+@REM HEAP_NEWSIZE="2G"
```
HTHou, line 83:
```suggestion
@REM Maximum heap size
@REM set MAX_HEAP_SIZE="2G"
@REM Minimum heap size
@REM set HEAP_NEWSIZE="2G"
```

HTHou:
Already tested on a Windows PC.

PR #3483, opened 2021-07-01
Session timeout

liujiajun:
Users may forget to close sessions. Such idle connections lead to a waste of server resource and in the worst case, take up all the memory of server. To solve this problem, we would like to introduce session timeout.

# Design
`SessionTimeoutManager` manages the lifecycle of all sessions. It keeps track of the last active time of the sessions 
and **periodically** checks if the current running sessions have timed out.  Specifically, if a session has been idle for a configurable period of time, `SessionTimeoutManager` will consider it as timed out and ends the session forcibly.
 
Session timeouts are refreshed under the following scenarios:
- `checkLogin` is called. Effectively, session timeout is refreshed whenever a new query/non-query operations comes in.
- long query is unregistered in `QueryTimeManager`.


server/src/main/java/org/apache/iotdb/db/query/control/SessionTimeoutManager.java
```diff
@@ -30,14 +30,18 @@
...
 
 public class SessionTimeoutManager {
-  private static final long CLEANUP_PERIOD_MILLIS = 10 * 1000; // TODO: Use a more reasonable value
+  private static final long SESSION_TIMEOUT =
+      IoTDBDescriptor.getInstance().getConfig().getSessionTimeoutThreshold();
+  private static final long CLEANUP_PERIOD_MILLIS = Math.max(5000, SESSION_TIMEOUT / 5);
   private static final Logger LOGGER = LoggerFactory.getLogger(SessionTimeoutManager.class);
 
-  private Map<Long, SessionInfo> idToSessionInfo;
+  private Map<Long, Long> idToLastActiveTime;
```
Alima777: Rename `sessionIdToLastActiveTime`?

liujiajun: Lol I was about to change to the exact same name

Alima777: hhh It's cool :D

server/src/main/java/org/apache/iotdb/db/query/control/SessionTimeoutManager.java
```diff
@@ -0,0 +1,116 @@
...
+  private static final Logger LOGGER = LoggerFactory.getLogger(SessionTimeoutManager.class);
+  private static final IoTDBConfig CONFIG = IoTDBDescriptor.getInstance().getConfig();
+  private static final long MINIMUM_CLEANUP_PERIOD = 2000;
+
+  private long sessionTimeout;
+  private Map<Long, Long> sessionIdToLastActiveTime;
+  private ScheduledExecutorService executorService;
+
+  private SessionTimeoutManager() {
+    this.sessionTimeout = CONFIG.getSessionTimeoutThreshold();
```
Alima777, line 44: Move to the above

Alima777 (approved):

server/src/main/java/org/apache/iotdb/db/query/control/SessionTimeoutManager.java
liujiajun: Thanks! Changed accordingly

PR #3500, opened 2021-07-03
Bug fix: time interval is parsed incorrectly

Alima777:
For example: time interval is [now() - 1d, now())

The first now() and second now() are parsed seperately, which is wrong.


HTHou (approved):

Alima777:
DO NOT forget to cherry pick to 0.11, 0.12

PR #3457, opened 2021-06-27
Add user document of time zone 

Alima777:

PR #3502, opened 2021-07-04
[IOTDB-1463] Implement builder pattern for Session and SessionPool

liujiajun:
Providing a more fluent way to construct Session and SessionPool

### DIFF
--- a/session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
+++ b/session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
@@ -1225,8 +1225,8 @@ public class SessionPool {
     }
 
     public Builder maxSize(int maxSize) {
-     this.maxSize = maxSize;
-     return this;
+      this.maxSize = maxSize;
+      return this;
     }
 
     public Builder user(String user) {
--- a/session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
+++ b/session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
@@ -1161,4 +1161,105 @@ public class SessionPool {
     // never go here
     return null;
   }
+
+  public int getMaxSize() {
+    return maxSize;
+  }
+
+  public String getIp() {
+    return ip;
+  }
+
+  public int getPort() {
+    return port;
+  }
+
+  public String getUser() {
+    return user;
+  }
+
+  public String getPassword() {
+    return password;
+  }
+
+  public int getFetchSize() {
+    return fetchSize;
+  }
+
+  public long getTimeout() {
+    return timeout;
+  }
+
+  public boolean isEnableCompression() {
+    return enableCompression;
+  }
+
+  public boolean isEnableCacheLeader() {
+    return enableCacheLeader;
+  }
+
+  public ZoneId getZoneId() {
+    return zoneId;
+  }
+
+  public static class Builder {
+    private final String ip;
+    private final int port;
+    private final String user;
+    private final String password;
+    private final int maxSize;
+
+    private int fetchSize = Config.DEFAULT_FETCH_SIZE;
+    private long timeout = 60_000;
+    private boolean enableCompression = false;
+    private ZoneId zoneId = null;
+    private boolean enableCacheLeader = Config.DEFAULT_CACHE_LEADER_MODE;
+
+    public Builder(String ip, int port, String user, String password, int maxSize) {
+      this.ip = ip;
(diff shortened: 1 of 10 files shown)


session/src/test/java/org/apache/iotdb/session/SessionUT.java
```diff
@@ -43,6 +43,7 @@
 
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 public class SessionUT {
```
wangchao316, line 49: generate, Generally, UT ends with Test,
IT ends with IT.

session/src/test/java/org/apache/iotdb/session/SessionUT.java
```diff
@@ -244,4 +245,25 @@ public void createDeviceTemplate() throws IoTDBConnectionException, StatementExe
         "template1", schemaNames, measurementList, dataTypeList, encodingList, compressionTypes);
     session.setSchemaTemplate("template1", "root.sg.1");
   }
+
+  @Test
+  public void testBuilder() {
+    Session.Builder builder = new Session.Builder("localhost", 1234);
```
wangchao316, line 251: hi
I found this implementation to be inconsistent with the use cases on jira.
We need to return a session through sessionbuild.
Session session = new SessionBuilder().build()

wangchao316 (changes requested):

session/src/test/java/org/apache/iotdb/session/SessionUT.java
liujiajun: Hi, thanks for the review! I think the implementation is consistent (see https://dzone.com/articles/design-patterns-the-builder-pattern). When using the pattern, however, we usually do not keep the interim builder object. I have changed the test case to make this more explicit.

wangchao316: default of constructors is no parameter.
// default
Session session = new SessionBuilder().build()
SessionPool sessionpool = new SessionPoolBuilder().build()

liujiajun: Correct me if I am wrong: `host` and `rpcPort` are compulsory fields for session and has no default values, so they are placed in the constructor. Same goes for sessionpool. If they are not put in constructor, users might forget to set values for them.

LebronAl (commented):
Great Job！
Only minor issues!
1. As we can directly use `sh sbin/start-cli.sh` to connect server with default host "127.0.0.1" and default port "6667" using default user "root" with default password "root", So I think the host and port of session/sessionPool can also have default values, so that on the one hand, it is more convenient to test IoTDB locally, on the other hand, it makes the session interface consistent with CLI. BTW, it's my fault to make a wrong example. Actually, A built-in-class builder creation must be performed just like this "Session.Builder.build()" rather than 'SessionBuilder.build()'
2. Please update the docs.
3. After this PR has been merged, please send an email to the community telling everyone how the new session/sessionPool can be created now.

liujiajun:
Thanks for the comments! Have changed accordingly. 

I removed the old construction method in the docs and also changed examples, so that new users can use builder as a preferred way.

Will blast out an email after this is merged~

session/src/test/java/org/apache/iotdb/session/SessionUT.java
wangchao316: ok

session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
```diff
@@ -1161,4 +1161,121 @@ public SessionDataSetWrapper executeRawDataQuery(List<String> paths, long startT
...
+
+  public boolean isEnableCompression() {
+    return enableCompression;
+  }
+
+  public boolean isEnableCacheLeader() {
+    return enableCacheLeader;
+  }
+
+  public ZoneId getZoneId() {
```
LebronAl, line 1201: Currently, the setup of zondId needs a rpc `setTimeZone` which you can refer to [PR3457](https://github.com/apache/iotdb/pull/3457/files), so maybe we can not initialized zoneId here, what do you think?

LebronAl (commented):
Last minor issue!

LebronAl (approved):
LGTM

wangchao316 (approved):
LGTM

PR #3511, opened 2021-07-06
add audit log when execute delete and set sg for tracing

yanhongwangg:


server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
```diff
@@ -1445,6 +1452,7 @@ private boolean createMultiTimeSeries(CreateMultiTimeSeriesPlan multiPlan)
 
   protected boolean deleteTimeSeries(DeleteTimeSeriesPlan deleteTimeSeriesPlan)
       throws QueryProcessException {
+    AUDIT_LOGGER.info("delete timeseries {}", deleteTimeSeriesPlan.getPaths());
```
HTHou, line 1455: deleteTimeSeriesPlan.getPaths() returns a list...

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
```diff
@@ -1512,6 +1520,7 @@ private boolean alterTimeSeries(AlterTimeSeriesPlan alterTimeSeriesPlan)
 
   public boolean setStorageGroup(SetStorageGroupPlan setStorageGroupPlan)
       throws QueryProcessException {
+    AUDIT_LOGGER.info("set storage group to {}", setStorageGroupPlan.getPaths());
     PartialPath path = setStorageGroupPlan.getPath();
```
HTHou, line 1524: getPath or getPaths?

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
```diff
@@ -959,6 +961,11 @@ private void addRowRecordForShowQuery(
 
   @Override
   public void delete(DeletePlan deletePlan) throws QueryProcessException {
+    AUDIT_LOGGER.info(
+        "delete data from {} in [{},{}]",
+        deletePlan.getPaths(),
+        deletePlan.getDeleteStartTime(),
+        deletePlan.getDeleteEndTime());
```
HTHou, line 968: Same as above.
getPaths returns a list. Can the logger print the paths in this list directly?

yanhongwangg: it can print the array elements directly as follows
image

yanhongwangg: it can print  a list as follows
image

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
yanhongwangg: The return value is same, but getPath will be better. I'll change it

mychaow (approved):
LGTM

PR #3514, opened 2021-07-06
[To rel/0.12] Use StringCachedPool in TsFileResource to reduce the memory size

HTHou:


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
```diff
@@ -315,20 +315,24 @@ private byte readVersionNumber(InputStream inputStream) throws IOException {
   }
 
   public void updateStartTime(String device, long time) {
+    device = device.intern();
```
wangchao316, line 318: LGTM.
hi , haonan , I have a minor question.
Because using the internal method has a performance impact. If intern() is performed after a new string is generated, the query performance will be affected，This needs to be considered. If the memory consumption and gc time are longer than intern, I agree with using intern().

HTHou: Hi, this pr is under testing and won't be merged until we get the performance result.

HTHou: By the way, it is a temporary way to reduce the memory size of TsFileResource. Actually, we are considering to remove the Devices set in FileTimeIndex.

wangchao316: ok. no problem..

qiaojialin (approved):

SteveYurongSu (approved):

PR #3525, opened 2021-07-07
Fix template restart bug

LebronAl:
In the current implementation, the template-based time series is restored by calling the getMNodes function after a restart. However, the getMNodes function currently does not support template lookup and this needs to be fixed.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
@@ -809,6 +809,7 @@ public class MManager {
       try {
         mNodes[i] = (MeasurementMNode) getNodeByPath(deviceId.concatNode(measurements[i]));
       } catch (PathNotExistException ignored) {
+        logger.warn("{} does not exist in {}", measurements[i], deviceId);
       }
       if (mNodes[i] == null && !IoTDBDescriptor.getInstance().getConfig().isEnablePartialInsert()) {
         throw new MetadataException(measurements[i] + " does not exist in " + deviceId);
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -804,10 +804,12 @@ public TSDataType getSeriesType(PartialPath path) throws MetadataException {
...
   public MeasurementMNode[] getMNodes(PartialPath deviceId, String[] measurements)
       throws MetadataException {
-    MNode deviceNode = getNodeByPath(deviceId);
     MeasurementMNode[] mNodes = new MeasurementMNode[measurements.length];
     for (int i = 0; i < mNodes.length; i++) {
-      mNodes[i] = ((MeasurementMNode) deviceNode.getChild(measurements[i]));
+      try {
+        mNodes[i] = (MeasurementMNode) getNodeByPath(deviceId.concatNode(measurements[i]));
+      } catch (PathNotExistException ignored) {
+      }
```
wangchao316, line 812: hi,  I have a minor..
If an exception occurs, a line of logs must be recorded to facilitate fault tracing.

LebronAl: OK

HTHou (approved):
lgtm

wangchao316 (approved):
LGTM

LebronAl:
> LGTM

Hi, it seems that the failure of E2E test has nothing to do with this PR, but with [PR 3292](https://github.com/apache/iotdb/pull/3292). Could you please check what the problem is?
image

SteveYurongSu (approved):

PR #3537, opened 2021-07-12
[IOTDB-1454] Fix wrong figure for index tree

samperson1997:
Fix wrong figure for index tree: the root node for the first and second examples should be LEAF_ENTITY type instead of INTERNAL_MEASUREMENT type.


SteveYurongSu (approved):
LGTM

PR #3538, opened 2021-07-12
[To rel/0.12][IOTDB-1488] Fix metaMember's forwarding clientPool timeout in cluster module

LebronAl:
see [jira1488](https://issues.apache.org/jira/browse/IOTDB-1488) and it's parent [jira1487](https://issues.apache.org/jira/browse/IOTDB-1487)

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
@@ -482,22 +482,22 @@ public class MetaGroupMember extends RaftMember {
         CLEAN_HARDLINK_INTERVAL_SEC,
         TimeUnit.SECONDS);
     dataClientRefresher.scheduleAtFixedRate(
-        this::RefreshClientOnce, REFRESH_CLIENT_SEC, REFRESH_CLIENT_SEC, TimeUnit.SECONDS);
+        this::refreshClientOnce, REFRESH_CLIENT_SEC, REFRESH_CLIENT_SEC, TimeUnit.SECONDS);
   }
 
-  private void RefreshClientOnce() {
+  private void refreshClientOnce() {
     for (Node receiver : allNodes) {
       if (!receiver.equals(thisNode)) {
         if (ClusterDescriptor.getInstance().getConfig().isUseAsyncServer()) {
-          RefreshClientOnceAsync(receiver);
+          refreshClientOnceAsync(receiver);
         } else {
-          RefreshClientOnceSync(receiver);
+          refreshClientOnceSync(receiver);
         }
       }
     }
   }
 
-  private void RefreshClientOnceSync(Node receiver) {
+  private void refreshClientOnceSync(Node receiver) {
     RaftService.Client client;
     try {
       client =
@@ -518,7 +518,7 @@ public class MetaGroupMember extends RaftMember {
     }
   }
 
-  private void RefreshClientOnceAsync(Node receiver) {
+  private void refreshClientOnceAsync(Node receiver) {
     RaftService.AsyncClient client;
     try {
       client =
--- a/cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
@@ -56,6 +56,8 @@ import org.apache.iotdb.cluster.rpc.thrift.CheckStatusResponse;
 import org.apache.iotdb.cluster.rpc.thrift.HeartBeatRequest;
 import org.apache.iotdb.cluster.rpc.thrift.HeartBeatResponse;
 import org.apache.iotdb.cluster.rpc.thrift.Node;
+import org.apache.iotdb.cluster.rpc.thrift.RaftService;
+import org.apache.iotdb.cluster.rpc.thrift.RefreshReuqest;
 import org.apache.iotdb.cluster.rpc.thrift.SendSnapshotRequest;
 import org.apache.iotdb.cluster.rpc.thrift.StartUpStatus;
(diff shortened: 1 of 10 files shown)


cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -163,6 +165,12 @@
    */
   private static final int REPORT_INTERVAL_SEC = 10;
 
+  /**
+   * every "REFRESH_CLIENT_SEC" seconds, a dataClientRefresher thread will try to refresh one thrift
+   * connection for each nodes other than itself.
+   */
+  private static final int REFRESH_CLIENT_SEC = 1;
```
mychaow, line 172: Why do you set default value to one? It seems too frequent.

mychaow: And could you please add some descriptions in jira 1488?

LebronAl: > And could you please add some descriptions in jira 1488?
added
> Why do you set default value to one? It seems too frequent.
This is an empirical number, and it can be increased. In a recent user scenario, the number of concurrent users may be 40 or 50, and the user perceiving a timeout is about 15 minutes, so if the background thread can check all the clients in a one-minute loop, almost all timeouts can be avoided. This PR is just to support this user's launch on 0.12 as soon as possible. More elaborate policies need to be developed in Issue1487, including but not limited to TTL for all ClientPools.

thrift-cluster/src/main/thrift/cluster.thrift
```diff
@@ -336,7 +338,9 @@ service RaftService {
   * When a follower finds that it already has a file in a snapshot locally, it calls this
   * interface to notify the leader to remove the associated hardlink.
   **/
-  void removeHardLink(1: string hardLinkPath)
+  void removeHardLink(1:string hardLinkPath)
+
+  void RefreshConnection(1:RefreshReuqest request)
```
wangchao316, line 343: hi , I have a little question,  what does  RefreshConnection() do?    which is an empty implementation in MetaClusterServer and DataClusterServer.   if have a use, we can use refreshConnection. generate, Method Name Camel Command Format.

LebronAl: For the `refreshConnection` rpc, It's just a simple implementation of application layer heartbeat preservation mechanism for long TCP connections.  For the detailed reason, you can refer to [jira1487](https://issues.apache.org/jira/browse/IOTDB-1487).

For the Camel Command Format, It's my fault and I will fix this. thanks for your reminding!

wangchao316: Thanks. I get it.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -460,6 +481,57 @@ private void threadTaskInit() {
...
+    } catch (TException e) {
+      logger.warn("encounter refreshing client timeout, throw broken connection", e);
+      // the connection may be broken, close it to avoid it being reused
+      client.getInputProtocol().getTransport().close();
+    } finally {
+      ClientUtils.putBackSyncClient(client);
+    }
+  }
+
+  private void RefreshClientOnceAsync(Node receiver) {
```
wangchao316, line 521: the same..

LebronAl: fixed

wangchao316 (approved):
LGTM

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -460,6 +481,57 @@ private void threadTaskInit() {
...
+    for (Node receiver : allNodes) {
+      if (!receiver.equals(thisNode)) {
+        if (ClusterDescriptor.getInstance().getConfig().isUseAsyncServer()) {
+          refreshClientOnceAsync(receiver);
+        } else {
+          refreshClientOnceSync(receiver);
+        }
+      }
+    }
+  }
```
neuyilan, line 498: Good job.
But I have some issues,  I think the `refreshClientOnce` can not **refresh** all clients in the client pool, because for one node receiver, there will be some clients, so this method can just fresh one client for one receiver node? for example, if the receiver node1 has 5 clients in the client pool, this method can only refresh one client, the other 4 clients can not be refreshed.

neuyilan (changes requested):

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
LebronAl: >this method can only refresh one client, the other 4 clients can not be refreshed.
No. Normal `getClient` will get a client from deque's tail and then push back to the tail, so the client in deque's tail is most likely to be used a lot but the client in deque's head is mostly likely to be not used for a long time, so `refreshClientOnce` will get a client from deque's head and then push back to deque's tail.

LebronAl: It looks like I might have misunderstood `pop` and `poll` in deque, I will check this :)

LebronAl: You are right. I have checked the source code in `ArrayDeque`. I should use `pollLast` rather than `poll`. I'll go back to the oven and make myself again

neuyilan (approved):
LGTM

PR #3544, opened 2021-07-13
[IOTDB-1482]Fix timeseries count and device count with Template

zyk990424:
### Bug Description
Apache IoTDB v0.12 cannot count timeseries and device correctly after introducing template.

### Bug Reproduction
When we create a template t1(s1, s2) and then set it to device root.sg1.d1, we expect value 2 produced by count timeseries and value 1 produced by count device. However, the actual result is 0 timeseries and 0 device.

### Bug Fix Solution
If a MNode is using template, it should be considered as device and the schema in its template will be considered as timeseries. However, the current code only consider a MeasurementMNode as timeseries and a MNode with MeasurementMNode child as device. Therefore, the method getAllTimeseriesCount and getAllTimeseriesCount in MTree.java should be modified and the judging criteria will be extended with template usage.

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
-  added unit tests or modified existing tests to cover new code paths, ensuring the threshold 
  for code coverage.

<hr>

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
@@ -1018,7 +1018,7 @@ public class MTree implements Serializable {
         MNode child = node.getChild(nodes[idx]);
         if (child == null) {
           if (node.isUseTemplate()
-                  && node.getUpperTemplate().getSchemaMap().containsKey(nodes[idx])) {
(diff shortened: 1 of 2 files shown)


HTHou:
Hi, you can use `mvn spotless:apply` to format the code.

HTHou (approved):

HTHou (approved):

PR #3553, opened 2021-07-13
improve mlog compatibility from 0.12.2 to 0.13.0

HTHou:
## Description
Since 0.12.2 introduce some new PhysicalPlans about template, the mlog will record some template information. 
Therefore, we need to reorder the PhysicalPlanType in master branch to keep the compatibility of mlog with 0.12.2.


HTHou:
Already tested by a 0.12.2 mlog which contains template.

qiaojialin (approved):

PR #3554, opened 2021-07-13
[IOTDB-1478]The whole IoTDB can not read/write if any one sg is not ready

HeimingZ:
When recovering IoTDB, if one storage group is not ready, all read/write requests will be rejected, which is not our respected.
To make each virtual storage group independent,  i change the ready unit of recovering IoTDB from the whole db to each independent virtual storage group, so each virtual storage group can resolve read/write requests if it has recovered.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/StorageEngine.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/StorageEngine.java
@@ -81,6 +81,7 @@ import java.util.ConcurrentModificationException;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/apache/iotdb/db/engine/StorageEngine.java
```diff
@@ -493,14 +443,14 @@ public StorageGroupProcessor buildNewStorageGroupProcessor(
   }
 
   private void waitAllSgReady(PartialPath storageGroupPath) throws StorageEngineException {
```
qiaojialin, line 445: This method could be removed

PR #3555, opened 2021-07-13
Using cached pool to reduce the String memory size in TsFileResourse

HTHou:
## Description
Cherry pick #3550


wangchao316 (approved):
LGTM

PR #3556, opened 2021-07-13
[IOTDB-1509] show timeseries throw a bug

wangchao316:
In a whole new environment, it has to be a whole new environment.

1. create timeseries root.ln.wf01.wt21 WITH DATATYPE=DOUBLE, ENCODING=RLE, compression=SNAPPY tags(tag1=v1, tag2=v2)

2. SHOW TIMESERIES root where tag1=v1

Msg: 303: The key tag1 is not a tag

Reason:
The show timeseries root... command is used to query all DataGroup groups. If a DataGroup group does not have a sequence, an error is reported. As a result, other DataGroups that have a sequence cannot obtain the result.
Solution:
we should catch a remote show-timeseries.

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
@@ -1565,7 +1565,7 @@ public class CMManager extends MManager {
               () -> {
                 try {
                   showTimeseries(group, plan, resultSet, context);
-                } catch (CheckConsistencyException e) {
+                } catch (CheckConsistencyException | MetadataException e) {
                   logger.error("Cannot get show timeseries result of {} from {}", plan, group);
                 }
                 return null;
--- a/testcontainer/src/test/java/org/apache/iotdb/db/sql/Cases.java
+++ b/testcontainer/src/test/java/org/apache/iotdb/db/sql/Cases.java
@@ -333,6 +333,9 @@ public abstract class Cases {
         "create timeseries root.ln.wf01.wt2 WITH DATATYPE=DOUBLE, ENCODING=RLE, compression=SNAPPY tags(tag1=v1, tag2=v2)";
     writeStatement.execute(createTimeSeries1);
     writeStatement.execute(createTimeSeries2);
+
+    Thread.sleep(200);
+
     // try to read data on each node. select .*
     for (Statement readStatement : readStatements) {
       ResultSet resultSet =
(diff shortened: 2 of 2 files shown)


neuyilan (commented):
I think this is just one probability problem :), when one docker container or VM  stop more than the time you set, it still fails :) 
so... I still do not know whether is there a better solution.....

wangchao316:
> I think this is just one probability problem :), when one docker container or VM stop more than the time you set, it still fails :)
> so... I still do not know whether is there a better solution.....

this pr is sloved for next error:
org.apache.thrift.TException: org.apache.iotdb.db.exception.metadata.MetadataException: The key tag1 is not a tag. 
 
This error occurs because the creation sequence is not synchronized to the follower. As a result, the query sequence cannot be found. This problem is increasing timeout and requires additional sleep between creation and query.

neuyilan:
> This error occurs because the creation sequence is not synchronized to the follower.
Yeah, I know, but how long does it take for the follower to synchronize the leader? Is it affected by the state of the virtual machine where the follower is located? After all, our tests are all in the cloud environment.

LebronAl:
> This error occurs because the creation sequence is not synchronized to the follower. As a result, the query sequence cannot be found. This problem is increasing timeout and requires additional sleep between creation and query.

If we query the followers of the data group, should we synchronize with the leader of the data group?

wangchao316:
> > This error occurs because the creation sequence is not synchronized to the follower. As a result, the query sequence cannot be found. This problem is increasing timeout and requires additional sleep between creation and query.
> 
> If we query the followers of the data group, should we synchronize with the leader of the data group?

no , we can add retry for this error. because this situation does not often

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
```diff
@@ -1565,7 +1565,7 @@ public MNode getMNode(MNode deviceMNode, String measurementName) {
               () -> {
                 try {
                   showTimeseries(group, plan, resultSet, context);
-                } catch (CheckConsistencyException e) {
+                } catch (CheckConsistencyException | MetadataException e) {
```
wangchao316, line 1568: The show timeseries root... command is used to query all DataGroup groups. If a DataGroup group does not have a sequence, an error is reported. As a result, other DataGroups that have a sequence cannot obtain the result.

LebronAl (commented):
Great Job~Finally we figured out what the problem was！
Please add same logic for `showTimeseries` and some other similar method if there has in `CMManager`

LebronAl:
ping, any updates?

LebronAl:
> Great Job~Finally we figured out what the problem was！
> Please add same logic for `showTimeseries` and some other similar method if there has in `CMManager`

Since the CI of master has been dead for a long time, I will merge this PR first. @wangchao316 will fix this review later this week.

PR #3569, opened 2021-07-14
[IoTDB-1449] path already exist error in IoTDBSink

Ring-k:
## Description
After starting IoTDB, executing `FlinkIoTDBSink.java` would result in an exception "... path already exists". It turns out `IoTDBSink.initSession()` would be executed in parallel. At the very beginning, `pool.checkTimeseriesExists(option.getPath())` retruns `true` among them. After one of them creating the path successfully, the others would fail. 

What is mentioned in JIRA is not an issue, which is reported in server and it is allowed. The client should receive the feedback (storage already exists), catch the exception and ignore it.


qiaojialin (approved):

mychaow (approved):
LGTM

PR #3574, opened 2021-07-15
[IoTDB-1499] Remove series registeration using IoTDBSink -> master

Ring-k:
The storage group and time series registerations are not necessary in IoTDBSink, which would introduce inconvenience for users.


qiaojialin (approved):

wangchao316:
1 Security Hotspot   , root is   Plaintext password， But This is in the sample code and does not affect

wangchao316 (approved):
LGTM

PR #3578, opened 2021-07-15
[IOTDB-1495]Remove Nested Measurement Function

zyk990424:
###  Forbide add child to MeasurementMNode
By overriding child operation in MeasurementMNode and checking parent node type while adding child, there will be no child under MeasurementNode.

### Template and ChildName Conflict
If a schema in template has the same name as a child of current MNode, the template should not be set to the MNode or the  child should not be add to the  MNode. Otherwise, there will be the  following conflicts.
1. There is a timeseries root.sg.d.s1.m1 and a template t1(s1, s2) is going to be set to root.sg.d. Then root.sg.d.s1 will either be a measurement or a device.
2. There is a root.sg.d with template t1(s1, s2) and a timeseries root.sg.d.s1.m1 is going to be created. Then root.sg.d.s1 will either be a measurement or a device.

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


qiaojialin:
https://github.com/apache/iotdb/pull/3589

PR #3580, opened 2021-07-15
[To rel/0.12]Fix thrift out of sequence in cluster module

LebronAl:
The Thrift Client will assign a serial number to each RPC. SocketTimeoutException is a behavior of the client, after which we should discard it. If it is used to send RPC again, it will increase the serial number when sending, but it may receive the last packet when receiving data. This causes the out of sequence bug.

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/query/last/ClusterLastQueryExecutor.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/query/last/ClusterLastQueryExecutor.java
@@ -30,7 +30,6 @@ import org.apache.iotdb.cluster.rpc.thrift.Node;
 import org.apache.iotdb.cluster.server.RaftServer;
 import org.apache.iotdb.cluster.server.member.DataGroupMember;
 import org.apache.iotdb.cluster.server.member.MetaGroupMember;
-import org.apache.iotdb.cluster.utils.ClientUtils;
 import org.apache.iotdb.cluster.utils.ClusterQueryUtils;
 import org.apache.iotdb.db.exception.StorageEngineException;
 import org.apache.iotdb.db.exception.query.QueryProcessException;
@@ -245,6 +244,7 @@ public class ClusterLastQueryExecutor extends LastQueryExecutor {
                 .getClientProvider()
                 .getAsyncDataClient(node, RaftServer.getReadOperationTimeoutMS());
       } catch (IOException e) {
+        logger.warn("can not get client for node= {}", node);
         return null;
       }
       buffer =
@@ -259,12 +259,10 @@ public class ClusterLastQueryExecutor extends LastQueryExecutor {
     }
 
     private ByteBuffer lastSync(Node node, QueryContext context) throws TException {
-      SyncDataClient client = null;
-      try {
-        client =
-            metaGroupMember
-                .getClientProvider()
-                .getSyncDataClient(node, RaftServer.getReadOperationTimeoutMS());
+      try (SyncDataClient client =
+          metaGroupMember
+              .getClientProvider()
+              .getSyncDataClient(node, RaftServer.getReadOperationTimeoutMS())) {
         return client.last(
             new LastQueryRequest(
                 PartialPath.toStringList(seriesPaths),
@@ -274,15 +272,8 @@ public class ClusterLastQueryExecutor extends LastQueryExecutor {
                 group.getHeader(),
                 client.getNode()));
       } catch (IOException e) {
+        logger.warn("can not get client for node= {}", node);
         return null;
-      } catch (TException e) {
-        // the connection may be broken, close it to avoid it being reused
-        client.getInputProtocol().getTransport().close();
-        throw e;
-      } finally {
-        if (client != null) {
-          ClientUtils.putBackSyncClient(client);
-        }
       }
     }
   }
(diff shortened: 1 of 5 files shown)


cluster/src/main/java/org/apache/iotdb/cluster/query/reader/DataSourceInfo.java
```diff
@@ -156,10 +156,15 @@ private Long applyForReaderIdSync(Node node, boolean byTimestamp, long timestamp
...
-    try (SyncDataClient client =
-        this.metaGroupMember
-            .getClientProvider()
-            .getSyncDataClient(node, RaftServer.getReadOperationTimeoutMS())) {
+    try {
+      SyncDataClient client =
+          this.metaGroupMember
+              .getClientProvider()
+              .getSyncDataClient(node, RaftServer.getReadOperationTimeoutMS())
+    }catch (IOException e){
```
HTHou, line 164:
```suggestion
    } catch (IOException e) {
```

cluster/src/main/java/org/apache/iotdb/cluster/query/last/ClusterLastQueryExecutor.java
```diff
@@ -258,19 +259,30 @@ private ByteBuffer lastAsync(Node node, QueryContext context)
...
             new LastQueryRequest(
                 PartialPath.toStringList(seriesPaths),
                 dataTypeOrdinals,
                 context.getQueryId(),
                 queryPlan.getDeviceToMeasurements(),
                 group.getHeader(),
-                syncDataClient.getNode()));
+                client.getNode()));
+      } catch (IOException e) {
+        return null;
```
wangchao316, line 277: 1. if have a exception, we need print a log.

cluster/src/main/java/org/apache/iotdb/cluster/query/last/ClusterLastQueryExecutor.java
```diff
@@ -258,19 +259,30 @@ private ByteBuffer lastAsync(Node node, QueryContext context)
...
+      } catch (TException e) {
+        // the connection may be broken, close it to avoid it being reused
+        client.getInputProtocol().getTransport().close();
+        throw e;
+      } finally {
+        if (client != null) {
```
wangchao316, line 283: we should use try()  --> close() --> putBack()

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
```diff
@@ -498,23 +498,22 @@ private void refreshClientOnce() {
...
-    } catch (TException e) {
-      return;
-    }
-    try {
       RefreshReuqest req = new RefreshReuqest();
       client.refreshConnection(req);
+    } catch (IOException ignored) {
     } catch (TException e) {
-      logger.warn("encounter refreshing client timeout, throw broken connection", e);
+      logger.info("encounter refreshing client timeout, throw broken connection", e);
```
wangchao316, line 510: why use info replace warn?   generate,  exception deal bettwen warn and error.

cluster/src/main/java/org/apache/iotdb/cluster/coordinator/Coordinator.java
```diff
@@ -735,7 +730,7 @@ public Node getThisNode() {
    * @param node the node to be connected
    * @param timeout timeout threshold of connection
    */
-  public SyncDataClient getSyncDataClient(Node node, int timeout) throws TException {
+  public SyncDataClient getSyncDataClient(Node node, int timeout) throws IOException {
```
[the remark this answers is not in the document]
LebronAl: To be consistent with `getAsyncDataClient`

cluster/src/main/java/org/apache/iotdb/cluster/query/last/ClusterLastQueryExecutor.java
LebronAl: Just to make consistent with `lastAsync`.I will add logs for both

cluster/src/main/java/org/apache/iotdb/cluster/query/last/ClusterLastQueryExecutor.java
LebronAl: OK

cluster/src/main/java/org/apache/iotdb/cluster/server/member/MetaGroupMember.java
LebronAl: fixed

cluster/src/main/java/org/apache/iotdb/cluster/query/fill/ClusterPreviousFill.java
```diff
@@ -240,19 +242,28 @@ private ByteBuffer remoteAsyncPreviousFill(
...
-    } catch (Exception e) {
+    SyncDataClient client = null;
+    try {
+      client =
+          metaGroupMember
+              .getClientProvider()
+              .getSyncDataClient(node, RaftServer.getReadOperationTimeoutMS());
+      byteBuffer = client.previousFill(request);
+    } catch (IOException e) {
+      logger.warn("{}: Cannot connect to {} during previous fill", metaGroupMember, node);
```
wangchao316, line 253: hi , xinyu , I have a little question.
why does IOException  not deal?  only print warn

LebronAl: To maintain compatibility. Before this PR, this function will print a row of log and return null when it fails to get the client

wangchao316 (approved):
LGTM

qiaojialin (approved):

PR #3601, opened 2021-07-20
[To rel/0.11] [IoTDB-1513]fix recover TsFileResource time bug in restart

qiaojialin:
https://issues.apache.org/jira/browse/IOTDB-1513

Suppose we have the following ChunkMetadata in a unclosed tsfile:

d1.s1: ChunkMetadata(INT32, starttime=1, endtime=2)，ChunkMetadata(INT32, starttime=1, endtime=2)
d1.s2: ChunkMetadata(BOOLEAN, starttime=1, endtime=2)，ChunkMetadata(INT32, starttime=1, endtime=2)

To solve this problem, we need to group the Chunkmetadata by measurement, and only remove the first ChunkMetadata of d1.s2, because the d1.s2 may be deleted and recreated.


HTHou (approved):
LGTM

PR #3605, opened 2021-07-20
[IOTDB-1515]Fix binary convertion in LastQueryExecutor

ericpai:
Please see JIRA https://issues.apache.org/jira/browse/IOTDB-1515 .

Here are another two questions. Any guidance or discussion is welcomed :).

1. What is the `Binary` type designed for? Any bytes or only valid UTF-8 strings?
2. If any bytes is allowed,  the implementation of `AbstractIoTDBJDBCResultSet.getBytes` is needed. As we cannot restore orignal binary data from a malformed UTF-8 string.

If this PR is accepted. I think it should be cherry-picked to release branches.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/query/executor/LastQueryExecutor.java
+++ b/server/src/main/java/org/apache/iotdb/db/query/executor/LastQueryExecutor.java
@@ -45,17 +45,13 @@ import org.apache.iotdb.tsfile.read.filter.basic.Filter;
 import org.apache.iotdb.tsfile.read.query.dataset.QueryDataSet;
 import org.apache.iotdb.tsfile.utils.Binary;
 import org.apache.iotdb.tsfile.utils.Pair;
+import org.apache.iotdb.tsfile.utils.TsPrimitiveType;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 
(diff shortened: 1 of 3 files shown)


HTHou:
Hi, currently we can only insert any bytes with Binary by insertTablet interface, right? InsertRecord interface only support insert String record. If we consider to support any bytes, a new interface of insertRecord is needed I think…

ericpai:
> some modification of insertRecord is needed I think…

@HTHou Yes, InsertRecord treats the input `TEXT` type data as a `String`, not `byte[]`. But user will be confused if it uses InsertTablet API to insert Binary objects and gets unexcepted ones.

Maybe we can indicate that only valid UTF-8 String supported in TEXT type in `InsertRecord()`, and support any bytes in the future. It's backward compatiable I think. But if we want to support any bytes. The JDBC interface should support `getBytes()`. I could help to do this.

However this PR is the superset of  UTF-8 string support(I can add another test case for valid utf8 bytes). Do you think it should be merged?

qiaojialin:
Hi, the data type of Binary is TEXT, so it's before only store UTF-8 String.

If we want to store bytes, we could import another data type called BINARY.

This could be discussed in the mail list :)

ericpai:
> Hi, the data type of Binary is TEXT, so it's before only store UTF-8 String.
> 
> If we want to store bytes, we could import another data type called BINARY.
> 
> This could be discussed in the mail list :)

It makes sense! Currently I will use base64 encoding as a workaround solution to store arbitrary bytes. This PR and JIRA will be closed.

jixuan1989:
@ericpai , we indeed need a new binary (better to be called as Blob) data type, which stores bytes and has no statistics like sum etc (first value and last value are also not needed by default)..

Then, current binary can be renamed as Clob or Text directly (but we need to consider compatibility)

PR #3609, opened 2021-07-22
[IOTDB-1041]modify startUpCheck

ijihang:
Solve the fourth problem of JIRA：https://issues.apache.org/jira/browse/IOTDB-1041
When more than half of seednodes are satisfied, the check is considered correct

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/utils/ClusterUtils.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/utils/ClusterUtils.java
@@ -252,18 +252,13 @@ public class ClusterUtils {
 
   public static boolean analyseStartUpCheckResult(
       int consistentNum, int inconsistentNum, int totalSeedNum) throws ConfigInconsistentException {
-    if (consistentNum == totalSeedNum) {
-      // break the loop and establish the cluster
-      return true;
-    } else if (inconsistentNum > 0) {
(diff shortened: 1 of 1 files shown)


LebronAl:
I remember I made the limit this way before because it is possible that some key parameters (such as the number of copies) are inconsistent with the node can also be added. After this change, can you manually check whether the node with inconsistent parameters will fail to be added? If so, can you successfully add the node after modifying the configuration again

ijihang:
> I remember I made the limit this way before because it is possible that some key parameters (such as the number of copies) are inconsistent with the node can also be added. After this change, can you manually check whether the node with inconsistent parameters will fail to be added? If so, can you successfully add the node after modifying the configuration again

I think the function of adding and deleting nodes should be to synchronously modify the configuration of each node in the cluster after successfully adding and deleting nodes

ijihang:
Maybe need to think about it

LebronAl:
Sorry, I mean `started`, not `added`, this change has nothing to do with adding or deleting nodes.
> I remember I made the limit this way before because it is possible that some key parameters (such as the number of copies) are inconsistent with the node can also be started. After this change, can you manually check whether the node with inconsistent parameters will fail to be started? If so, can you successfully start the node after modifying the configuration again

So maybe your change is better, I just want you to manually check if this change will affect the logic I just described. If it does not, then the startup logic is obviously more attractive

PR #3610, opened 2021-07-22
[IOTDB-1496] Timed flush memtable

HeimingZ:
Because some memtables are accumulated slowly, so add memtable_flush_interval to flush inactive memtables.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
@@ -1556,7 +1556,8 @@ public class StorageGroupProcessor {
     writeLock("timedFlushMemTable");
     try {
       // only check unsequence tsfiles' memtables
-      List<TsFileProcessor> tsFileProcessors = new ArrayList<>(workUnsequenceTsFileProcessors.values());
+      List<TsFileProcessor> tsFileProcessors =
+          new ArrayList<>(workUnsequenceTsFileProcessors.values());
       long timestampBaseline = System.currentTimeMillis() - config.getUnseqMemtableFlushInterval();
       for (TsFileProcessor tsFileProcessor : tsFileProcessors) {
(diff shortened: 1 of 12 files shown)


wangchao316 (commented):
Good Job.
There are some problems. If some caches are in use and we just get cleaned up, it's not going to affect its performance.

HeimingZ:
> Good Job.
> There are some problems. If some caches are in use and we just get cleaned up, it's not going to affect its performance.

1. Users can use unseq_memtable_flush_interval param to specify the alive time of memtables according to their use cases.
2. Change the timed flush level from all memtables to unsequence memtables because zombie memtables often oocurs in unsequence tsfiles.
3. Add enable_timed_flush_unseq_memtable param, so users can decide whether to enable this function.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -1552,6 +1552,29 @@ private void checkFileTTL(TsFileResource resource, long ttlLowerBound, boolean i
     }
   }
 
+  public void timedFlushMemTable() {
+    List<TsFileProcessor> tsFileProcessors = new ArrayList<>();
+    // only check unsequence tsfiles' memtables
+    tsFileProcessors.addAll(workUnsequenceTsFileProcessors.values());
+    long timestampBaseline = System.currentTimeMillis() - config.getUnseqMemtableFlushInterval();
```
qiaojialin, line 1559: put into writeLock

qiaojialin (changes requested):

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
HeimingZ: fixed

qiaojialin (approved):

PR #3624, opened 2021-07-26
[IOTDB-1528] CQ IT unstable: sometimes fails on `checkCQExecutionResult()`

SteveYurongSu:
### **JIRA:** 
https://issues.apache.org/jira/browse/IOTDB-1528

### **In this PR, I did two things:**
1. Improve code style in IoTDBContinuousQueryIT.java 
2. Try to make CQ IT stable

### **Why the CQ IT is unstable?**
CQ is asynchronously executed, the startup time and completion time of the CQ task are very dependent on the test environment. The IT sometimes failed on Github Actions but always passed on my MBP. 

At present, CQ IT fails mostly because the wait time is not long enough and the database does not produce the expected data points. We can make the test wait long enough for the database to produce more data points than expected and then do a follow-up check on the data points.

The core changes are as follows:

```java
long waitSeconds = 0;
List<Pair<Long, String>> result;
do {
  Thread.sleep(waitSeconds);
  waitSeconds += 1000;

  statement.execute("select temperature_avg from root.ln.wf01");
  result = collectQueryResult();
} while (result.size() < expectedSize);
```

### DIFF
--- a/server/src/test/java/org/apache/iotdb/db/integration/IoTDBContinuousQueryIT.java
+++ b/server/src/test/java/org/apache/iotdb/db/integration/IoTDBContinuousQueryIT.java
@@ -43,10 +43,6 @@ import java.util.stream.Collectors;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
-/**
- * Notice that, all test begins with "IoTDB" is integration test. All test which will start the
- * IoTDB server should be defined as integration test.
- */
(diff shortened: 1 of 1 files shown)


SteveYurongSu:
Please rerun the CI several times before merging to check my changes. :D

SteveYurongSu:
The 1st pass recorded.

SteveYurongSu:
The 2nd pass was recorded.

PR #3625, opened 2021-07-26
Fix TestContainer  IT

LebronAl:
Remove unnecessary tests due to there reasons:
- Since each test case is currently tested in seven clusters started by TestContainer, the time to add a test CI increases by approximately 2 mins, so we need to minimize the number of test functions if possible.
- I don't think it makes any sense to put three operations functions in testContainer without checking them, and these functions should be added to applier's IT which is already done.
- Currently, master ci often fails because of the presence of leader Not found for these three functions. Although this may require us to add the corresponding retry logic, it is better to kill these unrelated functions first.

### DIFF
--- a/testcontainer/src/test/java/org/apache/iotdb/db/sql/Cases.java
+++ b/testcontainer/src/test/java/org/apache/iotdb/db/sql/Cases.java
@@ -213,79 +213,6 @@ public abstract class Cases {
     }
   }
 
-  //  @Test
-  //  public void vectorCountTest() throws IoTDBConnectionException, StatementExecutionException {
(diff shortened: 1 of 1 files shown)


HTHou (approved):

wangchao316 (approved):
LGTM

jixuan1989:
> Since each test case is currently tested in seven clusters started by TestContainer, the time to add a test CI increases by approximately 2 mins, so we need to minimize the number of test functions if possible.

If some Tests have no conflict, you can put them into one test function() dominated by @Test annocation.

PR #3638, opened 2021-07-27
[IOTDB-1484]Auto create schema functionality with e2e testing in cluster

MrQuansy:
## Description
Auto create schema functionality in cluster requires careful design combined with the stand-alone partialInsert functionality and the BatchPlan interface, you can refer to the design https://shimo.im/docs/TTJDkQPkX3YYjpyR .
After the reorganization, the corresponding E2E tests need to be added for continuous integration.

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


cluster/src/main/java/org/apache/iotdb/cluster/coordinator/Coordinator.java
```diff
@@ -145,7 +145,10 @@ public TSStatus executeNonQueryPlan(PhysicalPlan plan) {
   private TSStatus executeNonQueryLocally(PhysicalPlan plan) {
     boolean execRet;
     try {
-      execRet = metaGroupMember.getLocalExecutor().processNonQuery(plan);
+      if (planExecutor == null) {
```
wangchao316, line 148: This is the singleton mode, and the concurrency problem needs to be considered.

cluster/src/main/java/org/apache/iotdb/cluster/log/applier/SingleReplicaApplier.java
```diff
@@ -0,0 +1,56 @@
...
+import org.apache.iotdb.db.exception.StorageEngineException;
+import org.apache.iotdb.db.exception.metadata.StorageGroupNotSetException;
+import org.apache.iotdb.db.exception.query.QueryProcessException;
+import org.apache.iotdb.db.qp.physical.PhysicalPlan;
+import org.apache.iotdb.db.qp.physical.crud.*;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SingleReplicaApplier extends DataLogApplier {
```
wangchao316, line 34: why this replica applier ?  could you please say reason?  thanks.

MrQuansy: Single replica applier is used for cluster mode with one replica. When data group member executes non-query plans, it does not need to execute the plan through raft protocol. Instead, it can directly execute the plan thorugh single replica applier.

chengjianyun:
Love the refine logic of error handling  in the PR very much. For auto creating schema, I have a suggestion. Is there any way that put the creation logic in executor instead of coordinator as I have done in https://github.com/apache/iotdb/pull/3770? 

Executor takes charge of schema auto creation could reduce a RTT between coordinator and simplify the error handling process. When an exception is thrown to coordinator, then it must not be resolved by retry or something like. So coordinator only needs to properly sends the error to user and/or record some log.

PR #3642, opened 2021-07-28
[IOTDB-1531] Check tsfile creation time when recovering

HeimingZ:
Check tsfile's time when recovering, make sure all tsfile time are less than System.currentTimeMillis(). If the tsfile's time is larger than System.currentTimeMillis(), the recover of corresponding virtual storage group will fail, users cannot read/write this virtual storage group until the time of the tsfile is correct.

Before fixing time：
before fix
After fixing time：
after fix

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
@@ -635,7 +635,7 @@ public class StorageGroupProcessor {
 
   @SuppressWarnings("squid:S3776") // Suppress high Cognitive Complexity warning
   private Pair<List<TsFileResource>, List<TsFileResource>> getAllFiles(List<String> folders)
-      throws IOException {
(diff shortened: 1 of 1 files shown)


wangchao316:
I have a little question. why all tsfile time are less than System.currentTimeMillis()?  
if tsfile time are better than System.currentTimeMillis(), Does iotdb-server used?
The time of real data may change due to a fault in the collection device (data is later than the current time). In this case, the entire cluster cannot be used.

HeimingZ:
> I have a little question. why all tsfile time are less than System.currentTimeMillis()?
> if tsfile time are better than System.currentTimeMillis(), Does iotdb-server used?
> The time of real data may change due to a fault in the collection device (data is later than the current time). In this case, the entire cluster cannot be used.

tsfile name time
This pr aims to make sure server time zone is correct and all tsfiles' created time is less than system current time. It only compares the created time of tsfile, which is determined by the server, so data from devices won't affect the use of iotdb.

qiaojialin (approved):

PR #3671, opened 2021-08-03
[IOTDB-1542] Cpp client segment fault: char[] buffer overflow caused by long exception message

SteveYurongSu:
JIRA: https://issues.apache.org/jira/browse/IOTDB-1542

image

I made 2 changes in this PR:
1. format code
2. remove `char buf[[111]`


HTHou (approved):
LGTM

PR #3686, opened 2021-08-05
close compaction by default

qiaojialin:
We'd better close compaction before it is stable


mychaow:
Greate job!

neuyilan (commented):
Great！
The IoTDB has suffered from compaction for a long time :)

PR #3699, opened 2021-08-06
[To rel/0.12] MeasurementId check while create timeseries or template

zyk990424:
### Problem
When users use session API to insert data and auto create time series, there's no syntax check on deviceIds or measurementIds before insert plan generating like SQL  parse. If there's a id contains symbol ".", like device id as root.sg.d and measurement id as a.b, the measurement id is obviously illegal and the meaning of path root.sg.d.a.b is ambiguous.  

Also solved https://issues.apache.org/jira/browse/IOTDB-1475 .

The schema names in template is also a loophole.

### Solution
Currently, a syntax check is added to mtree.checkTimeseries while it's going to auto create time series.
Method checkTemplateSchemaNames is  added to mmanager  while create template

### Suggestion
The illegal ids should be filtered when the API is invoked or a query is parsed as soon as possible before further operation.  

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
@@ -96,6 +96,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
@@ -2038,7 +2039,10 @@ public class MManager {
 
   public void createDeviceTemplate(CreateTemplatePlan plan) throws MetadataException {
     try {
+      checkTemplateSchemaNames(plan.getSchemaNames());
+
       Template template = new Template(plan);
+
       if (templateMap.putIfAbsent(plan.getName(), template) != null) {
         // already have template
         throw new DuplicatedTemplateException(plan.getName());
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -1901,6 +1901,12 @@ public MNode getSeriesSchemasAndReadLockDevice(InsertPlan plan)
           if (!config.isAutoCreateSchemaEnabled()) {
             throw new PathNotExistException(deviceId + PATH_SEPARATOR + measurementList[i]);
           } else {
+            // check measurementId syntax
+            if (measurementList[i].contains(".")
+                && !(measurementList[i].startsWith("\"") && measurementList[i].endsWith("\""))) {
+              throw new MetadataException(
+                  String.format("%s is an illegal measurementId", measurementList[i]));
+            }
```
HTHou, line 1909: I see a method in MTree called `checkTimeseries`. Should we use that?

zyk990424: That's may be better encapsulation. I've updated my code.

Here's some explanation. In this problem, once the path is transformed from PartialPath to String, like root.sg.d.a.b with measurementId a.b, the regex is unable to check the problem. Only when the path is in format of name list, String[] nodeNames, can the problem be detected. 

The string concat operation on path may exist widely in iotdb's code. That's why I suggested above, filter the illegal syntax before create query plan.

HTHou (approved):
LGTM

HTHou (changes requested):
Wait, plz let me review this PR again...

server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
```diff
@@ -283,11 +283,18 @@ MeasurementMNode createTimeseries(
...
       throw new IllegalPathException(
           String.format("The timeseries name contains unsupported character. %s", timeseries));
     }
+    String measurementId = timeseries.getMeasurement();
+    // check measurementId syntax
+    // only measurementId may be named separately from fullPath by user via API
+    if (measurementId.contains(".")
+        && !(measurementId.startsWith("\"") && measurementId.endsWith("\""))) {
+      throw new MetadataException(String.format("%s is an illegal measurementId", measurementId));
+    }
```
HTHou, line 297: I think we can also solve the issue https://issues.apache.org/jira/browse/IOTDB-1475 in this PR.

zyk990424: Fixed, plz check.

HTHou (approved):
LGTM

qiaojialin (approved):

PR #3707, opened 2021-08-10
[To rel/0.12][IOTDB-1556] Abort auto create device while setStorageGroup exception

zyk990424:
### Case
After operating the following operations: 
1. set storage group to root.a.sg
2. insert into root.a.d(time, s) values(1, 1)

Timeseries root.a.d.s won't be created for no storage group, but device node root.a.d will be created without storage group.

If a device node without any timeseries is created without storageGroup, then, in current code, it's unable to set any storage group on its ancestors and create any timeseries under it. Only set storage group to its descendants could be operated successfully. 

Since we don't have device node deletion oeperation, most of users will be confused and don't know how to handle this device node.

### Cause
In MManager.getDeviceNodeWithAutoCreate, if there's a storageGroupAlreadyExistException, the device node will be created with out storage group check. If the exception is caused by child node existence, then there will be no storage group above the device.

In current setStorageGroup logic, it is assumed that any timeseries or device is working under some storage group, which means a path existence implies a storage group existence.

### Solution
Modify StorageGroupAlreadyExistException with boolean hasChild added, thus mmanager could know whether the exception is caused by child. 

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/exception/metadata/StorageGroupAlreadySetException.java
+++ b/server/src/main/java/org/apache/iotdb/db/exception/metadata/StorageGroupAlreadySetException.java
@@ -25,12 +25,19 @@ public class StorageGroupAlreadySetException extends MetadataException {
 
   private static final long serialVersionUID = 9110669164701929779L;
 
+  private boolean hasChild = false;
+
   public StorageGroupAlreadySetException(String path) {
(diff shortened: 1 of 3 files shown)


HTHou (approved):

HTHou:
Plz do not forget to cherry-pick to master branch.

PR #3719, opened 2021-08-11
[IOTDB-1539] Fix delete operation with value filter is abnormal

xinzhongtianxia:
## Description
https://issues.apache.org/jira/browse/IOTDB-1539

For now, deletion do not support values filter, but, all conditions in where clause will be treated as time condition, thus will product two affects:

1. If the data type in condition could be parsed to Long, the statement will be executed successfully, it's dangerous!!
    For example, the original statement is "delete from root.vehicle.d1.s1 where s1 > 100", but the actually statement be 
    executed is "delete from root.vehicle.d1.s1 where time > 100".  Then, data may be deleted abnormally.

2. If the data type in condition could not be parsed to Long, we will get a 500 error because of a NumberFormatException.

## Solution
Before the value filter feature is implemented, I temporally fixed this by return an error message like "For delete statement, where clause can only contain time expressions, value filter is not currently supported." if any non-time-expression appears in where clause.

### DIFF
--- a/server/src/test/java/org/apache/iotdb/db/integration/IoTDBDeletionIT.java
+++ b/server/src/test/java/org/apache/iotdb/db/integration/IoTDBDeletionIT.java
@@ -94,6 +94,19 @@ public class IoTDBDeletionIT {
           "303: Check metadata error: For delete statement, where clause can only"
               + " contain time expressions, value filter is not currently supported.";
 
+      String errorMsg2 =
+          "303: Check metadata error: For delete statement, where clause can only contain"
+              + " atomic expressions like : time > XXX, time <= XXX,"
+              + " or two atomic expressions connected by 'AND'";
+
+      try {
+        statement.execute(
+            "DELETE FROM root.vehicle.d0.s0  WHERE s0 <= 300 AND time > 0 AND time < 100");
+        fail("should not reach here!");
+      } catch (SQLException e) {
+        assertEquals(errorMsg2, e.getMessage());
+      }
+
       try {
         statement.execute("DELETE FROM root.vehicle.d0.s0  WHERE s0 <= 300 AND s0 > 0");
         fail("should not reach here!");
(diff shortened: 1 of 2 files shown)


wangchao316:
Good Job
Thanks for your contribution.

server/src/test/java/org/apache/iotdb/db/integration/IoTDBDeletionIT.java
```diff
@@ -73,6 +74,73 @@ public void tearDown() throws Exception {
...
+      statement.execute("insert into root.vehicle.d0(time,s0) values (10,310)");
+      statement.execute("insert into root.vehicle.d0(time,s3) values (10,'text')");
+      statement.execute("insert into root.vehicle.d0(time,s4) values (10,true)");
+
+      String errorMsg =
+          "303: Check metadata error: For delete statement, where clause can only"
+              + " contain time expressions, value filter is not currently supported.";
+
+      try {
+        statement.execute("DELETE FROM root.vehicle.d0.s0  WHERE s0 <= 300 AND s0 > 0");
```
wangchao316, line 98: hi, add a test case,  DELETE FROM root.vehicle.d0.s0  WHERE s0 <= 300 and time > *** and time < ***.
what will this situation return ?

wangchao316 (changes requested):

server/src/test/java/org/apache/iotdb/db/integration/IoTDBDeletionIT.java
xinzhongtianxia: It will return"For delete statement, where clause can only contain atomic expressions like : time > XXX, time <= XXX, or two atomic expressions connected by 'AND'"

In situations expressions number more than 2, will always return this message.

Related code:

> List<FilterOperator> children = filterOperator.getChildren();  
    FilterOperator lOperator = children.get(0);
    FilterOperator rOperator = children.get(1);
    if (!lOperator.isLeaf() || !rOperator.isLeaf()) {
      throw new SQLParserException(DELETE_RANGE_ERROR_MSG);
    }

wangchao316 (approved):
LGTM

wangchao316:
after this pr merged, you need cherry pick 0.12-branch.

HTHou:
> after this pr merged, you need cherry pick 0.12-branch.

Let me do it.

HTHou (approved):

PR #3724, opened 2021-08-12
[To rel/0.12] Fix CreateTimeseriesIT runs slow

HTHou:
When I run this IT on my computer, it costs as long as 1 min to stop the rpc service. 
And I got an error message `org.apache.thrift.server.TThreadPoolServer:144 - Shutdown is not done after 60 SECONDS`.

The reason is this test doesn't close the statement and connection before execute `EnvironmentUtils.stopDaemon()`.


wangchao316 (approved):
LGTM.
We need to check other places to see if this is the case.

PR #3755, opened 2021-08-16
[IOTDB-1473] Fix the bug that run SessionExample.java will be failed

ghost:
JIRA issue: https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-1473

The issue happens because auto create schema doesn't work in cluster mode. The insertTables() tries to insert timeseries to `root.sg1.d2.s1`, `root.sg1.d2.s2`, `root.sg1.d2.s3` but which aren't created until the insertion. The schema create logic fail in the execution path which cause the issue.

Actually, there are three errors of the auto create schema fail in the cluster as far as I can know (comment out the create timeseries related code in `SessionExample.java`). 

```
    ...
    // createTimeseries();
    // createMultiTimeseries();
    // insertRecord();
    insertTablet();   // can't auto create schema
    insertTablets(); // can't auto create schema
    insertRecords(); // can't auto create schema
    selectInto();
   .....
```

This PR fixed the first two of situations. Has filed a [JIRA issue](https://issues.apache.org/jira/browse/IOTDB-1570) to track the third situation which need some more discussion.

### Fix logic

`insertTablet()` fails because this kind of insertion will call `forwardMultiSubPlan()` which will return a wrapper result instead of 304 error directly which cause  the error handling logic doesn't work.

For `insertTablets()`, it also need to apply measurements according to failed measurements for all sub `insertTablet()`.

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/coordinator/Coordinator.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/coordinator/Coordinator.java
@@ -461,20 +461,6 @@ public class Coordinator {
     } else {
       status = forwardToMultipleGroup(planGroupMap);
     }
-    boolean hasCreated = false;
-    try {
-      if (plan instanceof InsertPlan
-          && status.getCode() == TSStatusCode.TIMESERIES_NOT_EXIST.getStatusCode()
-          && ClusterDescriptor.getInstance().getConfig().isEnableAutoCreateSchema()) {
-        hasCreated = createTimeseriesForFailedInsertion(((InsertPlan) plan));
-      }
-    } catch (MetadataException | CheckConsistencyException e) {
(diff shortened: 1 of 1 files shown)


wangchao316:
you need perform "mvn spotless:apply"

wangchao316 (changes requested):
you need perform "mvn spotless:apply"

chengjianyun:
> you need perform "mvn spotless:apply"

Done.

chengjianyun:
Please help close the PR as it doesn't fix the bug clearly. I will start a discussion to talk about the auto schema creating feature. @wangchao316

chengjianyun:
Hi, there

The PR should be abandoned. I have submitted a new PR https://github.com/apache/iotdb/pull/3770 to replace this one. Sorry for the inconvenience.

PR #3758, opened 2021-08-16
[IOTDB-1562]incorrect exception processing in insertXXX() API

ShuanglinWu:
catch IOTDBException to fix issue_1562


wangchao316 (changes requested):
Good Job.Thanks for your contribution.
I have a little question, you need add some IT.

cigarl:
In fact, I think the exception here only throws an `IllegalPathException` on `new PartialPath()`
So we just need to update the current handling of the corresponding exception catch.
In my view，perhaps we should use more specific exception types when handling exceptions ,what do you think？

ShuanglinWu:
Thanks for reminding, I will add some IT tonight.
------------------------------------------------------------------
发件人：wangchao316 ***@***.***>
发送时间：2021年8月17日(星期二) 09:43
收件人：apache/iotdb ***@***.***>
抄　送：ShuanglinWu ***@***.***>; Author ***@***.***>
主　题：Re: [apache/iotdb] [IOTDB-1562]incorrect exception processing in insertXXX() API (#3758)

@wangchao316 requested changes on this pull request.
Good Job.Thanks for your contribution.
 I have a little question, you need add some IT.
—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

ShuanglinWu:
I think now it's just IllegalPathException, but maybe there will throws an MisMatchDataTypeException or an PathNOtExistException.
        It seems to me that we don't need to catch each exception, but we should get the error message and error code and return them to the client.
------------------------------------------------------------------
发件人：lisijia ***@***.***>
发送时间：2021年8月17日(星期二) 10:31
收件人：apache/iotdb ***@***.***>
抄　送：ShuanglinWu ***@***.***>; Author ***@***.***>
主　题：Re: [apache/iotdb] [IOTDB-1562]incorrect exception processing in insertXXX() API (#3758)

In fact, I think the exception here only throws an IllegalPathException on new PartialPath()
 So we just need to update the current handling of the corresponding exception catch.
 In my view，perhaps we should use more specific exception types when handling exceptions ,what do you think？
—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -2155,6 +2189,19 @@ private TSStatus onNPEOrUnexpectedException(
     return RpcUtils.getStatus(statusCode, message + e.getMessage());
   }
 
+  private TSStatus onIOTDBException(Exception e, String operation, int errorCode) {
```
qiaojialin, line 2192:
```suggestion
  private TSStatus onIoTDBException(Exception e, String operation, int errorCode) {
```

server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -1372,6 +1372,11 @@ public TSStatus insertRecords(TSInsertRecordsReq req) {
           allCheckSuccess = false;
         }
         insertRowsPlan.addOneInsertRowPlan(plan, i);
+      } catch (IoTDBException e) {
+        allCheckSuccess = false;
+        insertRowsPlan
+            .getResults()
+            .put(i, onIOTDBException(e, "inserting records", e.getErrorCode()));
```
qiaojialin, line 1379: Better to move all these "inserting records" to a IoTDBConstant or a n Enum class

qiaojialin (approved):

session/src/test/java/org/apache/iotdb/session/IoTDBSessionSimpleIT.java
```diff
@@ -726,6 +725,278 @@ public void testInsertOneDeviceRecordsWithIncorrectOrder()
...
+          tsEncodings,
+          CompressionType.SNAPPY,
+          Arrays.asList("alias1", "alias2", "alias3"));
+      fail("Exception expected");
+    } catch (StatementExecutionException e) {
+      assertTrue(
+          e.getMessage()
+              .contains(
+                  String.format(
+                      "[%s] Exception occurred: %s failed. %s is not a legal path",
```
qiaojialin, line 943: extract a constant string in this class

qiaojialin (approved):

wangchao316 (approved):
LGTM

HTHou (approved):

Issue #3786, opened 2021-08-19
Data file is deleted while file handle is not released


Alima777:
![image](https://user-images.githubusercontent.com/34242296/130008934-dcf50266-8362-4588-83a9-cbdb9c7038ca.png)

![image](https://user-images.githubusercontent.com/34242296/130008954-6ea75d60-0e6a-4468-8a7c-d34ca9ee12cf.png)

HTHou:
Fixed

PR #3787, opened 2021-08-19
[IOTDB-1574] [ISSUE-3786] Data file is deleted while file handle is not released

Alima777:
#3786 

This problem is caused by the reference number of query reader is not decreased corrcetly, which leads to the reference number is not 0 and the object can not be closed.


HTHou:
jira https://issues.apache.org/jira/browse/IOTDB-1574

HTHou (approved):

PR #3788, opened 2021-08-19
[To rel/0.12] [ISSUE-3786] Data file is deleted while file handle is not released

Alima777:
#3786

This problem is caused by the reference number of query reader is not decreased corrcetly, which leads to the reference number is not 0 and the object can not be closed.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,4 @@
+
 <!--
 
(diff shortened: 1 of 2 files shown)


HTHou:
Jira https://issues.apache.org/jira/browse/IOTDB-1574
Don't forget to update RELEASE_NOTES.md.

HTHou (approved):

PR #3795, opened 2021-08-19
[To rel/0.12] Fix load with time partition enable

yschengzi:
IoTDBLoadExternalTsfileIT.loadUnsequenceTsfile fails on CI for unix(16, ubuntu-latest) sometimes, 
make this branch to find what leads this results.


wangchao316 (changes requested):
Thanks for your contribution.
but this hava a little question.              
+ " "  + "subsequenceTime: "     ====       "       subsequenceTime:"

yschengzi:
> Thanks for your contribution.
> but this hava a little question.
> 
> * " "  + "subsequenceTime: "     ====       "       subsequenceTime:"

i see, thanks for reminding

server/src/main/java/org/apache/iotdb/db/engine/compaction/level/LevelCompactionTsFileManagement.java
```diff
@@ -191,13 +191,13 @@ private void deleteLevelFile(TsFileResource seqFile) {
       List<TsFileResource> result = new ArrayList<>();
       if (sequence) {
         List<SortedSet<TsFileResource>> sequenceTsFileList =
-            sequenceTsFileResources.get(timePartition);
+            sequenceTsFileResources.getOrDefault(timePartition, new ArrayList<>());
```
HTHou, line 194: Should we use computeIfAbsent to set the ArrayList back?

yschengzi: ok, computeIfAbsent is a better choice

HTHou: It seem that the tests failed if change here…

qiaojialin: this only for a check, he will add the tsfile in another place

HTHou (approved):

PR #3797, opened 2021-08-19
[IOTDB-1564] Make hearbeat and election timeout parameters be configurable

ericpai:
See JIRA  https://issues.apache.org/jira/browse/IOTDB-1564

The commits include:

1. Two new parameters of hearbeat and election and its docs.
2. Fix heartbeat  client pool timeout issue. The timeout should not be larger than heartbeat interval.
3. Some logger issues which don't print the stacktrace.

BTW, did we need to change the default value? I think we should be compatible with v0.12.


ericpai:
@jixuan1989 @LebronAl this pr is ready to be reviewed. I have tested in my local environment with a 3 nodes cluster. To set heartbeat_interval_ms=100, election_timeout_ms=10000, connection_timeout_ms=20000.  The fastest successful election will be finished in 200ms, from the leader shutting down, and about 1000ms in average.

ericpai:
@jixuan1989 I make a brief test with 3 nodes cluster without any read/write pressure. And use Arthas tool to analyze the CPU usage. Here is my conclusion:

1. With heartbeat_interval_ms=100, the MetaHeartbeatThread and DataHeartbeatThread costs <0.1% CPU. And with heartbeat_interval_ms=1000, they costs <0.01% CPU. There's no significant impaction if we decrease the heartbeat interval.
2. Under no outer pressure environment, with heartbeat_interval_ms=100, the headertbeatThread CPU costs about 34.09% of the whole IoTDB program cost. And with heartbeat_interval_ms=1000, the percentage is 1.27%. Because CPU cost in total is very low, the percentage difference is huge.

I think the side effect of CPU is acceptable.

Attachments for reference:
1. hb-dashboard-100.png: The snapshot of dashboard of CPU of one leader node under heartbeat_interval_ms=100 
2. hb-dashboard-1000.png: The snapshot of dashboard of CPU of one leader node under heartbeat_interval_ms=1000 
3. 1. hb -100.svg: The flame graph of CPU of one leader node by sampling at least 10 seconds under heartbeat_interval_ms=100 
4. 2. hb -1000.svg: The flame graph of CPU of one leader node by sampling at least 30 seconds under heartbeat_interval_ms=1000
[CPU-performance.zip](https://github.com/apache/iotdb/files/7096637/CPU-performance.zip)

cluster/src/main/java/org/apache/iotdb/cluster/client/sync/SyncDataHeartbeatClient.java
```diff
@@ -38,14 +38,19 @@
...
+    // establishing connection with some slow or dead nodes.
     super(
         protocolFactory.getProtocol(
             RpcTransportFactory.INSTANCE.getTransport(
                 new TSocket(
                     TConfigurationConst.defaultTConfiguration,
                     node.getInternalIp(),
                     node.getDataPort() + ClusterUtils.DATA_HEARTBEAT_PORT_OFFSET,
-                    RaftServer.getConnectionTimeoutInMS()))));
+                    Math.min(
```
jixuan1989, line 51:
```Math.min(
                        (int) RaftServer.getHeartbeatIntervalMs(),
                        RaftServer.getConnectionTimeoutInMS()))))```
can be saved in `RaftServer` as a new field.

Comment (LebronAl) [reply cluster/src/main/java/org/apache/iotdb/cluster/client/sync/SyncDataHeartbeatClient.java:51]: +1

File: cluster/src/main/java/org/apache/iotdb/cluster/client/sync/SyncMetaHeartbeatClient.java [hunk 2]
```diff
@@ -37,15 +37,20 @@
...
+    // establishing connection with some slow or dead nodes.
     super(
         protocolFactory.getProtocol(
             RpcTransportFactory.INSTANCE.getTransport(
                 new TSocket(
                     TConfigurationConst.defaultTConfiguration,
                     node.getInternalIp(),
                     node.getMetaPort() + ClusterUtils.META_HEARTBEAT_PORT_OFFSET,
-                    RaftServer.getConnectionTimeoutInMS()))));
+                    Math.min(
```
LebronAl, line 51: same

cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterConstant.java
```diff
@@ -47,8 +48,9 @@ private ClusterConstant() {
   }
 
   /**
-   * a failed election will restart in 2s~5s, this should be at least as long as a heartbeat
-   * interval, or a stale node may frequently issue elections and thus makes the leader step down
+   * a failed election will restart in [2*heartbeatInterval, 2*heartbeatInterval + 50ms). If this
```
LebronAl, line 51: Acutally, I think the lower bound should be 0. 
With current logic, all candidates will wait this extra time for `2*heartbeatInterval`, which does not help to avoid live locks. It also does not help that stale nodes initiate elections first. 
What do you think?

cluster/src/main/java/org/apache/iotdb/cluster/server/heartbeat/HeartbeatThread.java
```diff
@@ -230,10 +235,15 @@ void sendHeartbeatSync(Node node) {
                   heartbeatHandler.onComplete(heartBeatResponse);
                 } catch (TTransportException e) {
                   logger.warn(
-                      "{}: Cannot send heart beat to node {} due to network", memberName, node, e);
+                      memberName
```
LebronAl, line 238: Although there's nothing wrong with this, I am curious for doing this.

cluster/src/main/java/org/apache/iotdb/cluster/server/heartbeat/HeartbeatThread.java
```diff
@@ -230,10 +235,15 @@ void sendHeartbeatSync(Node node) {
...
+                          + ": Cannot send heartbeat to node "
+                          + node.toString()
+                          + " due to network",
+                      e);
                   client.getInputProtocol().getTransport().close();
                 } catch (Exception e) {
-                  logger.warn("{}: Cannot send heart beat to node {}", memberName, node, e);
+                  logger.warn(
+                      memberName + ": Cannot send heart beat to node " + node.toString(), e);
```
LebronAl, line 246: same

cluster/src/main/java/org/apache/iotdb/cluster/server/heartbeat/HeartbeatThread.java
```diff
@@ -413,7 +422,11 @@ private void requestVoteSync(Node node, ElectionHandler handler, ElectionRequest
                 } catch (TException e) {
                   client.getInputProtocol().getTransport().close();
                   logger.warn(
-                      "{}: Cannot request a vote from {} due to network", memberName, node, e);
+                      memberName
```
LebronAl, line 425: same

cluster/src/main/java/org/apache/iotdb/cluster/server/heartbeat/HeartbeatThread.java
```diff
@@ -86,15 +84,17 @@ public void run() {
...
             hasHadLeader = true;
             break;
           case FOLLOWER:
             // check if heartbeat times out
-            long heartBeatInterval =
+            long heartbeatInterval =
                 System.currentTimeMillis() - localMember.getLastHeartbeatReceivedTime();
-            if (heartBeatInterval >= RaftServer.getConnectionTimeoutInMS()) {
+            long randomElectionTimeout =
+                RaftServer.getElectionTimeoutMs() + getElectionRandomTimeOutMs();
```
LebronAl, line 96: In fact, I don't think we need to calculate a randomElectionTimeout every time we receive a heartbeat, we just need to save a calculated randomElectionTimeout for every term. 
But as it's trivil, I can accept this way for now.

LebronAl (commented):
Great Work!
I have left some comments, PTAL

cluster/src/main/java/org/apache/iotdb/cluster/server/heartbeat/HeartbeatThread.java
ericpai: Because the `logger.warn` has the following overload signatures with the `Marker` string:

```java
    void warn(Marker var1, String var2);
    void warn(Marker var1, String var2, Object var3);
    void warn(Marker var1, String var2, Object var3, Object var4);
    void warn(Marker var1, String var2, Object... var3);
    void warn(Marker var1, String var2, Throwable var3);
```

We can see that only the last one can print the stacktrace from the last parameter. However, the original code here satisfies `void warn(Marker var1, String var2, Object var3, Object var4);` , 3 parameters with only two placeholders. The execption message and stacktrace are lost.

cluster/src/main/java/org/apache/iotdb/cluster/server/heartbeat/HeartbeatThread.java
ericpai: This has been explained

cluster/src/main/java/org/apache/iotdb/cluster/server/heartbeat/HeartbeatThread.java
ericpai: This has been explained

cluster/src/main/java/org/apache/iotdb/cluster/config/ClusterConstant.java
ericpai: Yes, it's useless to set a minimum waiting time before starting election. After discussed with @jixuan1989 , we think that a constant maximum waiting time is not a good practice in real environment, but adding another configure parameter makes the cluster config more complicated. So we just set it as `1*heartbeatInterval`, which is lack of experiment now. The 50ms is just a protection if the heartbeatInterval is very small.

cluster/src/assembly/resources/conf/iotdb-cluster.properties
```diff
@@ -79,7 +79,9 @@ multi_raft_factor=1
 # all node's cluster_name in one cluster are the same
 # cluster_name=default
 
-# connection time out (ms) among raft nodes
+# Thrift socket and connection timeout between raft nodes, in milliseconds.
+# NOTE: the timeout of connection used for sending heartbeats and requesting votes
+# will be adjust to min(heartbeat_interval_ms, connection_timeout_ms).
```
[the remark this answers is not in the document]
ericpai: Thanks, this grammer issue has been fixed.

cluster/src/assembly/resources/conf/iotdb-cluster.properties
```diff
@@ -90,6 +92,13 @@ multi_raft_factor=1
 # not for the whole operation.
 # read_operation_timeout_ms=30000
 
+# the time interval (ms) between two rounds of heartbeat broadcast of one raft group leader.
+# Recommend to set it as 1/10 of election_timeout_ms, but larger than 1 RTT between each two nodes.
+# heartbeat_interval_ms=1000
+
+# The election timeout in follower, or the time waiting for request votes in elector, in milliseconds.
```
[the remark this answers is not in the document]
ericpai: Thanks, this grammer issue has been fixed.

cluster/src/main/java/org/apache/iotdb/cluster/client/sync/SyncDataHeartbeatClient.java
ericpai: A new static field is added to RaftServer.

cluster/src/main/java/org/apache/iotdb/cluster/server/heartbeat/HeartbeatThread.java
ericpai: OK, this is only a call of a random number.

cluster/src/main/java/org/apache/iotdb/cluster/server/heartbeat/HeartbeatThread.java
LebronAl: got it

LebronAl (approved):
LGTM

PR #3804, opened 2021-08-22
[IOTDB-1581] Consider deletions when recovering tsFileResource of incomplete tsfile

liuminghui233:
Jira: https://issues.apache.org/jira/browse/IOTDB-1581

When recovering a incomplete TsFile:

- Call `recoverResourceFromWriter` to recover the resource information through the ChunkMetadata information in `RestorableTsFileIOWriter`
- Call the `redoLogs` method to write the data in one or more wal files corresponding to this file to a temporary Memtable and persist to this incomplete TsFile

The deletions ( Recorded in the .mods file ) are not considered in the two aspects, resulting in the recovery time range being wider than actual.


HTHou (approved):
LGTM

PR #3806, opened 2021-08-23
[To rel/0.12] [ISSUE-3805]  OOM caused by Chunk cache

Alima777:
We used  to fuzzy computing to calculate the cache size, which led to some large caches being incorrectly counted, resulting in oom. And some concurrent problems may be caused by this.

Therefore, we try to calculate the cache size exactly to avoid these problems.


Alima777:
We need to test the performance comparing with the previous.

HTHou:
Plz also update the release note.

server/src/main/java/org/apache/iotdb/db/engine/cache/ChunkCache.java
```diff
@@ -63,50 +62,28 @@ private ChunkCache() {
...
-                      count = 1;
-                      currentSize = averageSize;
-                      entryAverageSize.set(averageSize);
-                    }
-                    return currentSize;
-                  }
-                })
+                      if (entrySize > 1024 * 1024 * 10) {
+                        logger.warn(String.format("Chunk size is %d over 10M", entrySize));
+                      }
```
qiaojialin, line 73:
```suggestion
```

qiaojialin, line 73:
```suggestion
```

qiaojialin (approved):

HTHou:
Don't forget to cherry pick it to master.

PR #3808, opened 2021-08-23
[IOTDB-1585][To rel/0.12] ModificationFile‘s write interface blocking

HeimingZ:
ModificationFile‘s write interface reads modifications from .mods file before writing modification, this causes deleting data costing too much time. After fixing, ModificationFile‘s write interface will write modifications directly into .mods file.

Before fixing:

before

After fixing:

after


HTHou (approved):
Update the release note?

HeimingZ:
> Update the release note?

Upadted.

PR #3810, opened 2021-08-24
[To rel/0.12] [IOTDB-1584] Doesn't support order by time desc in cluster mode

Alima777:
The type of batchdata is not serialized in cluster which leads to the information is lost.

WE NEED TO TEST Whether this modification is right.


Alima777:
Correctness check:
image

HTHou:
release note...

qiaojialin (approved):

PR #3818, opened 2021-08-24
[To rel/0.12] [IOTDB-1580] Error result of order by time desc when enable time partition

Alima777:
The bug is caused by we didn't traverse the tsfile seqeunce in time partition order.


qiaojialin (approved):

HTHou:
release note...

PR #3823, opened 2021-08-24
[IOTDB-1587] SessionPool optimization: a more aggressive Session creation strategy

SteveYurongSu:
https://issues.apache.org/jira/browse/IOTDB-1587

In this PR, I do the following things:

1. rename timeout to queryTimeoutInMs in Session
2. allow users to set connectionTimeoutInMs in SessionPool
3. new Session creation strategy in SessionPool: removeSession -> tryConstructNewSession
4. new Session creation strategy in SessionPool: when the Session cannot be obtained and the thread enters the waiting loop, it will always judge whether a new Session can be created first in each loop, instead of waiting in each loop after judging whether a new Session can be created only once.


HTHou (approved):
LGTM

PR #3832, opened 2021-08-25
[IOTDB-1583] Raft log failed to be committed in cluster version

cigarl:
The bug is caused by a raft log is oversize. 

When the size of a log is bigger than buffer (default 16MB),it can not be persistent due to the `BufferOverflowException`. But the log has committed in memory, we will commit the same log again. 

So,we should check the size of the log at the entrance to avoid that leader will send this log to followers.

BTW, this case may cause the next plan failed in server mode. Because we didn't clear the buffer, it still contains some information from the previous log.


cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
```diff
@@ -1032,6 +1032,15 @@ TSStatus processPlanLocally(PhysicalPlan plan) {
       log.setCurrLogTerm(getTerm().get());
       log.setCurrLogIndex(logManager.getLastLogIndex() + 1);
 
+      // if a single log exceeds the threshold
+      // we need to return error code to the client as in server mode
+      if (log.serialize().capacity() + Integer.BYTES
+          >= ClusterDescriptor.getInstance().getConfig().getRaftLogBufferSize()) {
+        logger.error(
+            "Log cannot fit into buffer, please increase raft_log_buffer_size;"
+                + "or reduce the size of requests you send.");
```
wangchao316, line 1041: 1. We need to control the log size. We can't let him have this anomaly.
2. If an exception occurs, the server obtains the exception and then splits the log file or uses other methods to apply the log file.

cigarl: > 1. We need to control the log size. We can't let him have this anomaly.
> 2. If an exception occurs, the server obtains the exception and then splits the log file or uses other methods to apply the log file.

When this problem occurs, the service can only be restarted. And the current modification method is consistent with the single server mode. So the operation of splitting logs involves several modules, we may need to control the plan size but not only log.
We may need to spend some time on this part in the future to make it better.

neuyilan: I think this needs further discussion, is some error happened during raft log `apply`, maybe only one raft group failed, it will lose some data, and cause the raft group not consistent with each other, how to deal with the exception?

For now, I think it ok just throw the exception and not let the log be committed.

wangchao316: ok, Thanks for submit a new issues for further discussion in jira, this pr could merged.

neuyilan: Thanks for all your suggesions，I will merge the pr, further discussion can be found here https://github.com/apache/iotdb/issues/3856

PR #3847, opened 2021-08-26
Remove timestamp datatype config

HTHou:
## Description

Currently we have a meaningless config that changes the datatype of timestamp from INT64 to INT32. 
It also may cause bug if users change the default value to INT32.

It's better to remove it.


wangchao316 (approved):
LGTM

PR #3870, opened 2021-08-30
Fix load TsFile failed when time partition enabled caused by TsFileReWriteTool NPE

HTHou:
Cherry pick 
- #3864


wangchao316 (approved):
LGTM

PR #3879, opened 2021-08-31
[IOTDB-1484]Auto create schema functionality with e2e testing in cluster

MrQuansy:
## Description
Auto create schema functionality in cluster requires careful design combined with the stand-alone partialInsert functionality and the BatchPlan interface, you can refer to the design https://shimo.im/docs/TTJDkQPkX3YYjpyR .
After the reorganization, the corresponding E2E tests need to be added for continuous integration.

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


testcontainer/src/test/java/org/apache/iotdb/db/sql/Cases.java
```diff
@@ -610,4 +614,222 @@ public void SetSystemReadOnlyWritableTest() throws SQLException {
...
+        tablet.addValue("s2", rowIndex, 2L);
+        tablet.addValue("s3", rowIndex, 3L);
+      }
+      session.insertTablet(tablet);
+      tablet.setPrefixPath(String.format("root.sg%s.d4", i));
+      tabletMap.put(String.format("root.sg%s.d4", i), tablet);
+    }
+
+    session.insertTablets(tabletMap);
+
```
LebronAl, line 752: where are the cases that sg exists but device does not exist?

testcontainer/src/test/java/org/apache/iotdb/db/sql/Cases.java
```diff
@@ -610,4 +614,222 @@ public void SetSystemReadOnlyWritableTest() throws SQLException {
...
+            Assert.assertEquals(resultSet.getLong(1), t);
+            Assert.assertEquals(resultSet.getString(2), "1");
+            Assert.assertEquals(resultSet.getString(3), "2");
+            Assert.assertEquals(resultSet.getString(4), "3");
+          }
+        }
+      }
+    }
+
+    // test create time series without sg
```
LebronAl, line 769: test create time series with sg

testcontainer/src/test/java/org/apache/iotdb/db/sql/Cases.java
```diff
@@ -610,4 +614,222 @@ public void SetSystemReadOnlyWritableTest() throws SQLException {
...
+    List<List<String>> measurementsList = new ArrayList<>();
+    for (int i = 0; i < 5; i++) {
+      time_list.add((long) i);
+      List<String> measurements = new ArrayList<>();
+      measurements.add(String.format("s%d", i));
+      measurements.add(String.format("s%d", i + 5));
+      measurements.add(String.format("s%d", i + 10));
+      measurementsList.add(measurements);
+    }
+    session.insertRecordsOfOneDevice(
```
LebronAl, line 809: Why don't you put them in front like all the others?

LebronAl (changes requested):
Please take a look at your test first as they lacks the most important case for auto-creating aschema

cluster/src/main/java/org/apache/iotdb/cluster/server/member/DataGroupMember.java
```diff
@@ -696,40 +699,35 @@ public TSStatus executeNonQueryPlan(PhysicalPlan plan) {
...
+              if (e instanceof BatchProcessException) {
+                for (TSStatus status : ((BatchProcessException) e).getFailingStatus()) {
+                  if (status.getCode() == TSStatusCode.TIMESERIES_NOT_EXIST.getStatusCode()) {
+                    hasCreated = createTimeseriesForFailedInsertion(((InsertPlan) plan));
+                    ((BatchPlan) plan).getResults().clear();
+                    break;
+                  }
+                }
+              }
+            } else if (cause instanceof PathNotExistException) {
```
LebronAl, line 721: looks like the `isEnableAutoCreateSchema` parameter in this if branch is turned off?

cluster/src/main/java/org/apache/iotdb/cluster/server/member/DataGroupMember.java
```diff
@@ -816,6 +828,22 @@ private boolean createTimeseriesForFailedInsertion(InsertPlan plan)
...
+        if (insertPlan.getFailedMeasurements() != null) {
+          insertPlan.getPlanFromFailed();
+        }
+      }
+    }
+
+    if (plan instanceof InsertRowsOfOneDevicePlan) {
+      for (InsertRowPlan insertPlan : ((InsertRowsOfOneDevicePlan) plan).getRowPlans()) {
+        if (insertPlan.getFailedMeasurements() != null) {
+          insertPlan.getPlanFromFailed();
```
LebronAl, line 842: What's the difference for `getPlanFromFailed` and `recoverFromFailure`? I remember we used `recoverFromFailure` in `Coordinator` before.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/DataGroupMember.java
```diff
@@ -776,14 +774,28 @@ private TSStatus executeNonQueryPlanWithKnownLeader(PhysicalPlan plan) {
       boolean hasCreated = false;
       try {
         if (plan instanceof InsertPlan
-            && status.getCode() == TSStatusCode.TIMESERIES_NOT_EXIST.getStatusCode()
             && ClusterDescriptor.getInstance().getConfig().isEnableAutoCreateSchema()) {
-          hasCreated = createTimeseriesForFailedInsertion(((InsertPlan) plan));
+          if (plan instanceof InsertRowsPlan || plan instanceof InsertMultiTabletPlan) {
```
LebronAl, line 778: How about put these 10 lines to a new function?

server/src/main/java/org/apache/iotdb/db/qp/physical/crud/InsertRowPlan.java
```diff
@@ -323,7 +323,7 @@ public void markFailedMeasurementInsertion(int index, Exception e) {
     }
     failedValues.add(values[index]);
     values[index] = null;
-    dataTypes[index] = null;
+    // dataTypes[index] = null;
```
LebronAl, line 326: Do not comment codes.Just remove it

server/src/main/java/org/apache/iotdb/db/qp/physical/crud/InsertRowPlan.java
```diff
@@ -414,27 +414,31 @@ private void putValues(DataOutputStream outputStream) throws QueryProcessExcepti
...
-          case DOUBLE:
-            ReadWriteIOUtils.write((Double) values[i], outputStream);
-            break;
-          case TEXT:
-            ReadWriteIOUtils.write((Binary) values[i], outputStream);
-            break;
-          default:
-            throw new QueryProcessException("Unsupported data type:" + dataTypes[i]);
+        if (isNeedInferType) {
+          ReadWriteIOUtils.write(values[i].toString(), outputStream);
```
LebronAl, line 418: Why add this? Can we just infer types once? It look like the type needs to be inferred again after the change, whether it is restarted or sent to other nodes?

server/src/main/java/org/apache/iotdb/db/qp/physical/sys/CreateMultiTimeSeriesPlan.java
```diff
@@ -144,6 +144,11 @@ public void setIndexes(List<Integer> indexes) {
     return results;
   }
 
+  @Override
+  public List<PartialPath> getPrefixPaths() {
+    return null;
```
LebronAl, line 149: Never return `null`! You can use `Collections.emptyList()`.BTW, why not return true `prefixPaths` for `CreateMultiTimeSeriesPlan` ?

LebronAl (commented):
Great Work!
I have finished my review.PTAL

cluster/src/main/java/org/apache/iotdb/cluster/server/member/DataGroupMember.java
MrQuansy: The method getPlanFromFailed can only get failed measurements without successful ones, so fewer time series will be processed afterwards.

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
```diff
@@ -705,6 +700,22 @@ public boolean createTimeseries(InsertRowsPlan insertRowsPlan)
     return allSuccess;
   }
 
+  public boolean createTimeseries(InsertRowsOfOneDevicePlan insertRowsOfOneDevicePlan)
```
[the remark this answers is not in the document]
MrQuansy: Each InsertRowPlan in InsertRowsOfOneDevicePlan is related with different measurements. It is necessary to  create time series for each InsertRowPlan.

server/src/main/java/org/apache/iotdb/db/qp/physical/sys/CreateMultiTimeSeriesPlan.java
MrQuansy: The method get prefixPaths is defined for batch plan. But CreateMultiTimeSeriesPlan stores each path of time series, I think it is no need to implement it.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/DataGroupMember.java
MrQuansy: Actually, there is only 5 lines to put together as a function, I think it is of no need.

server/src/main/java/org/apache/iotdb/db/qp/physical/crud/InsertRowPlan.java
MrQuansy: You can image this situation: at first the data type has been inferred (data type is double but value is string type), then when the plan needs to be serialized, it will throw the exception like "String can not cast to Double".

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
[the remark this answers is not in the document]
LebronAl: Sorry, My mistake. I mean you can use `CreateMultiTimeSeries` once rather than `createTimeSeries` many times

cluster/src/main/java/org/apache/iotdb/cluster/server/member/DataGroupMember.java
```diff
@@ -206,6 +208,7 @@ public DataGroupMember(PartitionGroup nodes) {
     if (ClusterDescriptor.getInstance().getConfig().isUseAsyncApplier()) {
       applier = new AsyncDataLogApplier(applier, name);
     }
+    singleReplicaApplier = new SingleReplicaDataApplier(metaGroupMember, this);
```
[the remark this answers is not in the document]
MrQuansy: Maybe I can achieve this using another pr.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/DataGroupMember.java
LebronAl: Can you be sure to return the correct error message(the correct index for a column) when a column is recovered with getPlanFromFailed and the execution fails again?

server/src/main/java/org/apache/iotdb/db/qp/physical/sys/CreateMultiTimeSeriesPlan.java
LebronAl: So please return `Collections.emptyList()` rather than `null`.There is no doubt that returning`null` is not elegant.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/DataGroupMember.java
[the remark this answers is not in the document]
LebronAl: OK, there are also other things we can do under the single replica, such as closing thread pools that are not needed.

cluster/src/main/java/org/apache/iotdb/cluster/server/member/DataGroupMember.java
LebronAl: It doesn't matter how many lines of code you have. In my opinion, these ten or so lines of code are lower than processPlanLocally, createTimeseries, etc. If you don't think so, then it's okay with me. It's about everyone's code style.

server/src/main/java/org/apache/iotdb/db/qp/physical/crud/InsertRowPlan.java
LebronAl: This is an optimization that can be done on our standalone version as well, otherwise it would be a waste of time to have to refer it again after restarting from WAL. In that case, that doesn't have much to do with the PR. You can do this logic for now, but you need to make it complete. For example, implement the same logic for "private void putValues(ByteBuffer buffer)", plus serialize/deserialize UT for your case.

cluster/src/main/java/org/apache/iotdb/cluster/log/applier/BaseApplier.java
```diff
@@ -219,10 +219,9 @@ private void processPlanWithTolerance(InsertPlan plan, DataGroupMember dataGroup
           pullTimeseriesSchema(plan, dataGroupMember.getHeader());
           plan.recoverFromFailure();
           getQueryExecutor().processNonQuery(plan);
-        } else {
-          throw e;
         }
-      } else throw e;
+      }
+      throw e;
```
LebronAl, line 224: Maybe these two `throws` is difficult to merge, but there's definitely something wrong with this `throws` ? so will it be thrown again when it executes correctly?

cluster/src/main/java/org/apache/iotdb/cluster/server/member/DataGroupMember.java
```diff
@@ -204,11 +202,11 @@ public DataGroupMember(PartitionGroup nodes) {
     allNodes = nodes;
     setQueryManager(new ClusterQueryManager());
     slotManager = new SlotManager(ClusterConstant.SLOT_NUM, getMemberDir(), getName());
+    singleReplicaApplier = new DataLogApplier(metaGroupMember, this);
```
LebronAl, line 205: can we use `applier` directly? What's their difference?

LebronAl (commented):
PTAL

cluster/src/main/java/org/apache/iotdb/cluster/server/member/DataGroupMember.java
MrQuansy: I extract the process of executing physical plan as a function. We can directly call the DataLogApplier.exexcutePhysicalPlan to skip log processing.

LebronAl (approved):
LGTM

PR #3910, opened 2021-09-04
[IOTDB-1088][WIP] Add API executeStatement

LeicongLi:
Add function `executeStatement` Sesssion and SessionPool, referring to java.sql.Statement#execute(java.lang.String).

Need help to point out how to test. 

Ticket link
https://issues.apache.org/jira/browse/IOTDB-1088

### DIFF
--- a/session/src/main/java/org/apache/iotdb/session/Session.java
+++ b/session/src/main/java/org/apache/iotdb/session/Session.java
@@ -559,6 +559,21 @@ public class Session {
   /**
    * execute query sql
    *
+   * @param sql any sql statement
+   * @return <code>true</code> if the first result is a <code>SessionDataset</code> object; <code>
+   *     false</code> if it is an update count or there are no results
+   * @throws IoTDBConnectionException if a database access error occurs.
+   * @throws StatementExecutionException when the statement failed.
+   */
+  public boolean executeStatement(String sql)
+      throws StatementExecutionException, IoTDBConnectionException {
+    SessionDataSet dataSet = executeStatementMayRedirect(sql, queryTimeoutInMs);
+    return dataSet.hasNext();
+  }
+
+  /**
+   * execute query sql
+   *
    * @param sql query statement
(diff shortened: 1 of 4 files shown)


HTHou:
> Need help to point out how to test.

Thanks for your contribution!
To add test cases, you can see the following IT classes.
https://github.com/apache/iotdb/blob/master/session/src/test/java/org/apache/iotdb/session/IoTDBSessionSimpleIT.java
https://github.com/apache/iotdb/blob/master/session/src/test/java/org/apache/iotdb/session/pool/SessionPoolTest.java

HTHou:
Hi, I checked the execute(String sql) method on IoTDB JDBC. It seems that it keeps a static ResultSet in this class. 
I think we should do the same thing.

https://github.com/apache/iotdb/blob/3333f3b2a6a5035b7f42a33dd23b3e841b0768b1/jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBStatement.java#L251-L287

I also checked the clickhouse and shardingsphere JDBC code and they all keep the ResultSet.

LeicongLi:
> Hi, I checked the execute(String sql) method on IoTDB JDBC. It seems that it keeps a static ResultSet in this class.
> I think we should do the same thing.
> 
> https://github.com/apache/iotdb/blob/3333f3b2a6a5035b7f42a33dd23b3e841b0768b1/jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBStatement.java#L251-L287
> 
> I also checked the clickhouse and shardingsphere JDBC code and they all keep the ResultSet.
> 
> [IOTDB-1088](https://issues.apache.org/jira/browse/IOTDB-1088) also states
> 
> ```
> You must then use the methods * <code>getSessionDataSet</code> to retrieve the result.
> ```
> 
> Therefore, I think we also need a method `getSessionDataSet`.

Thanks.
Will reopen the PR when the changes are made.

PR #3915, opened 2021-09-06
[cluster-refactor]split DataGroupServiceImpls into engine and impls

cigarl:
Split DataGroupServiceImpls into DataGroupEngine and DataGroupServiceImpls.


cigarl:
@jixuan1989 
If we want to split `DataGroupServiceImpls` like `StorageEngine` . Requests from clients _(in the entrance of RPC)_ should be encapsulated, and send wrappers to engine, engine try to route to their processors _(Impls)_.This change may extend the scope of modification.

So,I'm just splitting out the parts that belong to Engine from `DataGroupServiceImpls`. The entrance of RPC is still in _Impls_.

Please review the code if you have any time,thanks.

LebronAl (approved):
LGTM

cigarl:
apply spotless to trigger _Checks_

PR #3919, opened 2021-09-07
[cluster-refactor] refactor client pool

chengjianyun:
Closed the https://github.com/apache/iotdb/pull/3886 to merge and resolve conflicts.

The main idea of the refactor:

- Apply `Apache commons pool` replace ourself implementation.
- Remove unnecessary classes.
- Add new class `ClientManager` to help manage client instances instead of letting user get client from pool directly.

The PR will contain:

- [x] keep the interface unchanged as much as possible
- [x] Close client means close the connection after refactor.
- [x] replace `ClientPool` with `KeyedObjectPool` in Apache Commons Pool
- [x] fix the test cases error
- [ ] self test


chengjianyun:
Should be push to cluster- instead of master.

PR #3926, opened 2021-09-07
[ISSUE-3927] Inefficient Usage of JCF

DittoTool:
Hi,

We find that there are several inefficient usages of Java Collections:

1. The contains method is invoked upon a list object in a loop. We recommend replacing it with a HashSet.
2. Random access can occur at several linkedlist objects, which run in linear time complexity. We recommend replacing them with Arraylist objects.
3. There is no iteration occurring upon a LinkedHashMap, thus the insertion order does not matter. We recommend replacing it with a HashMap.
4. ArrayList is inserted before an iteration, while multiple memory reallocation might occur when the size of the list exceeds its capacity. We recommend replacing it with a LinkedList.

We discovered the above inefficient usage of containers by our tool Ditto. The patch is submitted. Could you please check and accept it? We have tested the patch on our PC. The patched program works well.

Bests

Ditto


DittoTool:
I appended several other transformations just now, in which a list used for ownership checking is transformed to HashSet. I think these transformations are meaningful both in the theoretical and practical manner. Besides, here is another case as follows. https://github.com/apache/iotdb/blob/860c8de52ced88b4440f10bc0b07cd8a9c7441bd/jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBDatabaseMetadata.java#L1395-L1397
I did not include it and submit the transformation for it because the size of the list is only 4. However, I still think it would be better if a set is constructed to maintain the four elements rather than list during the initialization.

tsfile/src/main/java/org/apache/iotdb/tsfile/read/reader/chunk/ChunkReader.java
```diff
@@ -55,7 +55,7 @@
 
   protected Filter filter;
 
-  private List<IPageReader> pageReaderList = new LinkedList<>();
+  private List<IPageReader> pageReaderList = new ArrayList<>();
```
Alima777, line 58: In chunkReader Line 118, `return pageReaderList.remove(0).getAllSatisfiedPageData();`

tsfile/src/main/java/org/apache/iotdb/tsfile/read/reader/chunk/VectorChunkReader.java
```diff
@@ -59,7 +59,7 @@
 
   protected Filter filter;
 
-  private final List<IPageReader> pageReaderList = new LinkedList<>();
+  private final List<IPageReader> pageReaderList = new ArrayList<>();
```
Alima777, line 62: In vectorChunkReader Line 268, `return pageReaderList.remove(0).getAllSatisfiedPageData();`

cluster/src/main/java/org/apache/iotdb/cluster/log/manage/FilePartitionedSnapshotLogManager.java
```diff
@@ -130,11 +133,11 @@ private void collectTsFilesAndFillTimeseriesSchemas(List<Integer> requiredSlots)
 
     // 2.register the measurement
     boolean slotExistsInPartition;
-    List<Integer> slots = null;
+    HashSet<Integer> slots = null;
```
Alima777, line 136: It's ok

Alima777 (commented):
Hi, thank you for the contribution. I checked all the change place, most changes are good while some are not reasonable. Please check it.

Alima777:
And you need to run `mvn:spotless apply` to modify the code style.

tsfile/src/main/java/org/apache/iotdb/tsfile/read/reader/chunk/ChunkReader.java
DittoTool: We think the following part actually performs random access upon the list object frequently in a loop. It would be quite expensive. Compare to the remove(0) you mentioned, I still think it would be better to change it to ArrayList. Remove(0) would invoke the copy of a successive memory, which can be conducted in a relatively efficient way.

https://github.com/apache/iotdb/blob/aaa96b974becf7cc57ef5ac36c3fa9e199423010/server/src/main/java/org/apache/iotdb/db/query/executor/fill/LastPointReader.java#L171-L173

tsfile/src/main/java/org/apache/iotdb/tsfile/read/reader/chunk/VectorChunkReader.java
DittoTool: Similarly, this list object is also randomly accessed in a loop

tsfile/src/main/java/org/apache/iotdb/tsfile/read/reader/chunk/ChunkReader.java
Alima777: Hi, I didn't notice that. If that, maybe use a iterator here to traverse the list is better. In this way we can do these two methods both efficiently.

DittoTool: Wow. It is the optimal solution.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileProcessor.java
```diff
@@ -746,7 +747,8 @@ public void syncFlush() throws IOException {
     synchronized (tmpMemTable) {
       try {
         long startWait = System.currentTimeMillis();
-        while (flushingMemTables.contains(tmpMemTable)) {
+        HashSet<IMemTable> flushingMemSet = new HashSet<>(flushingMemTables);
+        while (flushingMemSet.contains(tmpMemTable)) {
```
SteveYurongSu, line 751: I think the change here may cause concurrent issues: `flushingMemTables` may be changed by other working threads.

DittoTool: Ditto has not modeled the semantics of multi-threading. It is true that the transformation here is not sound. Ditto just finds that there exists a performance issue here. The method contains of a list is quite inefficient. Moreover, it is invoked frequently in this loop.

SteveYurongSu: Theoretically speaking, the statement `flushingMemSet.contains(tmpMemTable)` is inefficient, so at least your finding is correct. 😊

Alima777:
Hi, thank you for the discovery of the inefficient usage of jcf, almost all of them are correct. Since we often pay less attention to these details, if one analysis tool can be used to do this work, it will be great!
Thank you for the contribution! :D

SteveYurongSu:
#3927

SteveYurongSu (approved):
LGTM. Your findings are correct overall.

SteveYurongSu (changes requested):
Oh ... Plz fix the compile errors

LebronAl (approved):
LGTM for cluster related changes

server/src/main/java/org/apache/iotdb/db/query/executor/fill/LastPointReader.java
```diff
@@ -169,8 +170,9 @@ private TimeValuePair getChunkLastPoint(IChunkMetadata chunkMetaData) throws IOE
           chunkStatistics.getEndTime(), chunkStatistics.getLastValue(), dataType);
     }
     List<IPageReader> pageReaders = FileLoaderUtils.loadPageReaderList(chunkMetaData, timeFilter);
-    for (int i = pageReaders.size() - 1; i >= 0; i--) {
-      IPageReader pageReader = pageReaders.get(i);
+    Iterator it = pageReaders.descendingIterator();
```
HTHou, line 173:
```suggestion
    Iterator it = ((LinkedList<IPageReader>)pageReaders).descendingIterator();
```

server/src/main/java/org/apache/iotdb/db/query/executor/fill/LastPointReader.java
```diff
@@ -35,6 +35,7 @@
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Iterator;
```
HTHou, line 38:
```suggestion
import java.util.Iterator;
import java.util.LinkedList;
```

server/src/main/java/org/apache/iotdb/db/query/executor/fill/LastPointReader.java
```diff
@@ -169,8 +171,9 @@ private TimeValuePair getChunkLastPoint(IChunkMetadata chunkMetaData) throws IOE
           chunkStatistics.getEndTime(), chunkStatistics.getLastValue(), dataType);
     }
     List<IPageReader> pageReaders = FileLoaderUtils.loadPageReaderList(chunkMetaData, timeFilter);
-    for (int i = pageReaders.size() - 1; i >= 0; i--) {
-      IPageReader pageReader = pageReaders.get(i);
+    Iterator it = ((LinkedList<IPageReader>) pageReaders).descendingIterator();
```
HTHou, line 174:
```suggestion
    if (pageReaders instanceof LinkedList) {
      Iterator it = ((LinkedList<IPageReader>) pageReaders).descendingIterator();
    } else {
      // for singleton list
      Iterator it = pageReaders.iterator();
    }
```

server/src/main/java/org/apache/iotdb/db/query/executor/fill/LastPointReader.java
```diff
@@ -169,8 +171,14 @@ private TimeValuePair getChunkLastPoint(IChunkMetadata chunkMetaData) throws IOE
...
-    for (int i = pageReaders.size() - 1; i >= 0; i--) {
-      IPageReader pageReader = pageReaders.get(i);
+    if (pageReaders instanceof LinkedList) {
+      Iterator it = ((LinkedList<IPageReader>) pageReaders).descendingIterator();
+    } else {
+      // for singleton list
+      Iterator it = pageReaders.iterator();
+    }
+    while (it.hasNext()) {
+      IPageReader pageReader = (IPageReader) it.next();
```
HTHou, line 181:
```suggestion
    Iterator it;
    if (pageReaders instanceof LinkedList) {
      it = ((LinkedList<IPageReader>) pageReaders).descendingIterator();
    } else {
      // for singleton list
      it = pageReaders.iterator();
    }
    while (it.hasNext()) {
      IPageReader pageReader = (IPageReader) it.next();
```

SteveYurongSu (approved):
Finally ... :D

PR #3930, opened 2021-09-08
[IOTDB-1619] There is an error msg when I restart iotdb-cluster

cigarl:
When i try to figure out why this issue occurs, i find that the result of last raft log file is different with my guess.Here are the result of my test environment _(ps.I ran into the same problem.)_

We find these files of raft log:
```
/.../system/raftLog/-1869016501/49923372-9223372036854775807-244-idx, 
/.../system/raftLog/-1869016501/48875442-49224751-241-idx, 
/.../system/raftLog/-1869016501/49224752-49574061-242-idx, 
/.../system/raftLog/-1869016501/49574062-49923371-243-idx

/.../system/raftLog/-1869016501/49574062-49923371-243-data, 
/.../system/raftLog/-1869016501/49923372-9223372036854775807-244-data, 
/.../system/raftLog/-1869016501/48875442-49224751-241-data, 
/.../system/raftLog/-1869016501/49224752-49574061-242-data
```
and process these files _(242-index and 243-data?)_ :
image

Obviously, we have a problem with where the files are sorted.That may be a bomb in `recover` module.

_footnote_  Here's meta info
```
commitLogTerm=1, 
commitLogIndex=49923370, 
lastLogIndex=49923370, 
lastLogTerm=1, 
maxHaveAppliedCommitIndex=49923357
```

PR #3939, opened 2021-09-09
init dummyIndex after restart cluster

cigarl:
In current logic, if we restart cluster, and `applyIndex` is smaller than `commitIndex`,here are three problems occur.

- First, we cannot append un-apply logs to `commitManager`.Like 103 and 104 in the picture,they can not be appended to `commitEntries` in `commitManager`.
- Second, un-commitManager can not append a new log. Like 0 can not be appended to `unCommitEntries`  in `unCommitManager`.
-  Third, un-applied logs can not be re-applied.That means 103 and 104 could be lost. 

This is all because `dummyIndex` is initialized to the default value,it cannot return to its pre-reboot state.

_ps.The information in the picture assumes none of these questions exist_
6211a9f5a25a465958c041d145f6dd5

This PR try to resolve this problem,but we may also need a processor to avoid logIndex is too big, dunmmyIndex need to re-initialize if logIndex will be bigger than our expect.

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/log/manage/RaftLogManager.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/log/manage/RaftLogManager.java
@@ -56,19 +56,13 @@ public abstract class RaftLogManager {
 
   private static final Logger logger = LoggerFactory.getLogger(RaftLogManager.class);
 
-  /**
-   * manage uncommitted entries
-   */
+  /** manage uncommitted entries */
   private UnCommittedEntryManager unCommittedEntryManager;
 
-  /**
-   * manage committed entries in memory as a cache
-   */
+  /** manage committed entries in memory as a cache */
   private CommittedEntryManager committedEntryManager;
 
-  /**
-   * manage committed entries in disk for safety
-   */
+  /** manage committed entries in disk for safety */
   private StableEntryManager stableEntryManager;
 
   private long commitIndex;
@@ -91,9 +85,7 @@ public abstract class RaftLogManager {
 
   protected LogApplier logApplier;
 
-  /**
-   * to distinguish managers of different members
-   */
+  /** to distinguish managers of different members */
   private String name;
 
   private ScheduledExecutorService deleteLogExecutorService;
@@ -102,15 +94,11 @@ public abstract class RaftLogManager {
   private ExecutorService checkLogApplierExecutorService;
   private Future<?> checkLogApplierFuture;
 
-  /**
-   * minimum number of committed logs in memory
-   */
+  /** minimum number of committed logs in memory */
   private int minNumOfLogsInMem =
       ClusterDescriptor.getInstance().getConfig().getMinNumOfLogsInMem();
 
-  /**
-   * maximum number of committed logs in memory
-   */
+  /** maximum number of committed logs in memory */
   private int maxNumOfLogsInMem =
       ClusterDescriptor.getInstance().getConfig().getMaxNumOfLogsInMem();
 
@@ -294,7 +282,7 @@ public abstract class RaftLogManager {
    *
    * @param index request entry index
    * @return throw EntryCompactedException if index < dummyIndex, -1 if index > lastIndex or the
-   * entry is compacted, otherwise return the entry's term for given index
+   *     entry is compacted, otherwise return the entry's term for given index
    * @throws EntryCompactedException
    */
   public long getTerm(long index) throws EntryCompactedException {
@@ -362,11 +350,11 @@ public abstract class RaftLogManager {
    * Used by follower node to support leader's complicated log replication rpc parameters and try to
    * commit entries.
    *
-   * @param lastIndex    leader's matchIndex for this follower node
-   * @param lastTerm     the entry's term which index is leader's matchIndex for this follower node
+   * @param lastIndex leader's matchIndex for this follower node
+   * @param lastTerm the entry's term which index is leader's matchIndex for this follower node
    * @param leaderCommit leader's commitIndex
-   * @param entries      entries sent from the leader node Note that the leader must ensure
-   *                     entries[0].index = lastIndex + 1
+   * @param entries entries sent from the leader node Note that the leader must ensure
+   *     entries[0].index = lastIndex + 1
    * @return -1 if the entries cannot be appended, otherwise the last index of new entries
    */
   public long maybeAppend(long lastIndex, long lastTerm, long leaderCommit, List<Log> entries) {
@@ -408,10 +396,10 @@ public abstract class RaftLogManager {
    * Used by follower node to support leader's complicated log replication rpc parameters and try to
    * commit entry.
    *
-   * @param lastIndex    leader's matchIndex for this follower node
-   * @param lastTerm     the entry's term which index is leader's matchIndex for this follower node
+   * @param lastIndex leader's matchIndex for this follower node
+   * @param lastTerm the entry's term which index is leader's matchIndex for this follower node
    * @param leaderCommit leader's commitIndex
-   * @param entry        entry sent from the leader node
+   * @param entry entry sent from the leader node
    * @return -1 if the entries cannot be appended, otherwise the last index of new entries
    */
   public long maybeAppend(long lastIndex, long lastTerm, long leaderCommit, Log entry) {
@@ -486,7 +474,7 @@ public abstract class RaftLogManager {
    * Used by leader node to try to commit entries.
    *
    * @param leaderCommit leader's commitIndex
-   * @param term         the entry's term which index is leaderCommit in leader's log module
+   * @param term the entry's term which index is leaderCommit in leader's log module
    * @return true or false
    */
   public synchronized boolean maybeCommit(long leaderCommit, long term) {
@@ -540,7 +528,7 @@ public abstract class RaftLogManager {
    * then whichever log has the larger lastIndex is more up-to-date. If the logs are the same, the
    * given log is up-to-date.
    *
-   * @param lastTerm  candidate's lastTerm
+   * @param lastTerm candidate's lastTerm
    * @param lastIndex candidate's lastIndex
    * @return true or false
    */
@@ -553,7 +541,7 @@ public abstract class RaftLogManager {
    * Pack entries from low through high - 1, just like slice (entries[low:high]). firstIndex <= low
    * <= high <= lastIndex.
    *
-   * @param low  request index low bound
+   * @param low request index low bound
    * @param high request index upper bound
    */
   public List<Log> getEntries(long low, long high) {
@@ -687,7 +675,7 @@ public abstract class RaftLogManager {
   /**
    * Returns whether the index and term passed in match.
    *
-   * @param term  request entry term
+   * @param term request entry term
    * @param index request entry index
    * @return true or false
    */
@@ -735,7 +723,7 @@ public abstract class RaftLogManager {
    * Check whether the parameters passed in satisfy the following properties. firstIndex <= low <=
    * high.
    *
-   * @param low  request index low bound
+   * @param low request index low bound
    * @param high request index upper bound
    * @throws EntryCompactedException
    * @throws GetEntriesWrongParametersException
@@ -883,9 +871,7 @@ public abstract class RaftLogManager {
     return maxHaveAppliedCommitIndex;
   }
 
-  /**
-   * check whether delete the committed log
-   */
+  /** check whether delete the committed log */
   void checkDeleteLog() {
     try {
       synchronized (this) {
@@ -1041,9 +1027,7 @@ public abstract class RaftLogManager {
     this.blockAppliedCommitIndex = blockAppliedCommitIndex;
   }
 
-  /**
-   * Apply the committed logs that were previously blocked by `blockAppliedCommitIndex` if any.
-   */
+  /** Apply the committed logs that were previously blocked by `blockAppliedCommitIndex` if any. */
   private void reapplyBlockedLogs() {
     if (!blockedUnappliedLogList.isEmpty()) {
       applyEntries(blockedUnappliedLogList);
--- a/cluster/src/main/java/org/apache/iotdb/cluster/log/manage/RaftLogManager.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/log/manage/RaftLogManager.java
@@ -30,6 +30,7 @@ import org.apache.iotdb.cluster.log.Log;
 import org.apache.iotdb.cluster.log.LogApplier;
 import org.apache.iotdb.cluster.log.Snapshot;
 import org.apache.iotdb.cluster.log.StableEntryManager;
+import org.apache.iotdb.cluster.log.manage.serializable.LogManagerMeta;
 import org.apache.iotdb.cluster.server.monitor.Timer.Statistic;
 import org.apache.iotdb.db.utils.TestOnly;
 import org.apache.iotdb.tsfile.utils.RamUsageEstimator;
@@ -55,13 +56,19 @@ public abstract class RaftLogManager {
 
   private static final Logger logger = LoggerFactory.getLogger(RaftLogManager.class);
 
-  /** manage uncommitted entries */
+  /**
+   * manage uncommitted entries
+   */
   private UnCommittedEntryManager unCommittedEntryManager;
 
-  /** manage committed entries in memory as a cache */
+  /**
+   * manage committed entries in memory as a cache
+   */
   private CommittedEntryManager committedEntryManager;
 
-  /** manage committed entries in disk for safety */
+  /**
+   * manage committed entries in disk for safety
+   */
   private StableEntryManager stableEntryManager;
 
   private long commitIndex;
@@ -84,7 +91,9 @@ public abstract class RaftLogManager {
 
   protected LogApplier logApplier;
 
-  /** to distinguish managers of different members */
+  /**
+   * to distinguish managers of different members
+   */
   private String name;
 
   private ScheduledExecutorService deleteLogExecutorService;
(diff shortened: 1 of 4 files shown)


cluster/src/main/java/org/apache/iotdb/cluster/log/manage/serializable/SyncLogDequeSerializer.java
```diff
@@ -236,6 +237,24 @@ public LogManagerMeta getMeta() {
     return getLogs(meta.getMaxHaveAppliedCommitIndex(), meta.getCommitLogIndex());
   }
 
+  /**
+   * Recover un-committed logs but already persistent. Maybe,we can extract
```
LebronAl, line 241: This persistence strategy needs to be improved, such as persisting uncommitted logs to ensure that some of raft's corner cases are correct. But in currrent logic, this function doesn't seem to be working?

cluster/src/main/java/org/apache/iotdb/cluster/log/manage/RaftLogManager.java
```diff
@@ -125,6 +127,8 @@ protected RaftLogManager(StableEntryManager stableEntryManager, LogApplier appli
     long first = getCommittedEntryManager().getDummyIndex();
     long last = getCommittedEntryManager().getLastIndex();
     this.setUnCommittedEntryManager(new UnCommittedEntryManager(last + 1));
+    this.getUnCommittedEntryManager()
```
LebronAl, line 130: same as below

LebronAl (commented):
To solve these problems completely, I think `raftLogManager` needs a complete refactoring, including but not limited to better concurrency control, better persistence strategy, etc.

When I first joined the community in my senior year, the first thing I did was to implement `raftLogManager` with reference to `storage` and `unstable` interfaces in `etcd`. Because I didn't have a very deep understanding of `raft` and `etcd` at that time,  after a long period, I finally realized that this was probably an awful implementation and I apologize for that. There are two main reasons:
* For `raft`, uncommitted logs also need to be persisted (i.e., logs on disk may also need to be truncated) in order to ensure correctness after reboot.
* `etcd` is an event-driven architecture, so its interior is all unlocked. However, in our architecture, `raftLogManager` can be accessed concurrently by multiple threads. So we've added a lot of patches for concurrency control right now, but this is actually an area we can think about implementing better.

Thank you very much for your great contributions. This week I will focus on the PR you raised about `cluster-` branch and a bug at hand. After that, I would like to discuss the refactoring of `raftLogManager` with you guys and community. And of course you're welcome to start to design the new  `raftLogManager` with us right now.

cigarl:
> To solve these problems completely, I think `raftLogManager` needs a complete refactoring, including but not limited to better concurrency control, better persistence strategy, etc.
> 
> When I first joined the community in my senior year, the first thing I did was to implement `raftLogManager` with reference to `storage` and `unstable` interfaces in `etcd`. Because I didn't have a very deep understanding of `raft` and `etcd` at that time, after a long period, I finally realized that this was probably an awful implementation and I apologize for that. There are two main reasons:
> 
> * For `raft`, uncommitted logs also need to be persisted (i.e., logs on disk may also need to be truncated) in order to ensure correctness after reboot.
> * `etcd` is an event-driven architecture, so its interior is all unlocked. However, in our architecture, `raftLogManager` can be accessed concurrently by multiple threads. So we've added a lot of patches for concurrency control right now, but this is actually an area we can think about implementing better.
> 
> Thank you very much for your great contributions. This week I will focus on the PR you raised about `cluster-` branch and a bug at hand. After that, I would like to discuss the refactoring of `raftLogManager` with you guys and community. And of course you're welcome to start to design the new `raftLogManager` with us right now if you are free.

IMO，we need the following two approaches in parallel:

- First,we should solve this problem in the current code structure and submit to `master` and `rel/0.12.x` branch. Because when a user has a problem like mine, it's hard to solve immediately, this means that some data could be loss, and cluster may enter an unstable state.This is a disaster for the production environment.We need to provide users with a patch to avoid this problem.
- Second,we can refactor this part on `cluster-` branch.Once we have solved the first problem, we can spend more time refactoring this part better.  And we don't have to push this part of the content to the user immediately.It can be put into the next release as a better implementation.

I'd be happy to do it together   : )

LebronAl:
> I'd be happy to do it together : )

OK, then that's my reviews~At current, it seems that the `getAllEntriesAfterCommittedIndex()` will always return empty list in current design? What's your opinion?

cluster/src/main/java/org/apache/iotdb/cluster/log/manage/serializable/SyncLogDequeSerializer.java
cigarl: > This persistence strategy needs to be improved, such as persisting uncommitted logs to ensure that some of raft's corner cases are correct. But in currrent logic, this function doesn't seem to be working?

In fact, this part of the change is due to the meta information was not flushed in time. When we flush the log file, we don't flush the `LogManagerMeta` at the same time(see [#3930](https://github.com/apache/iotdb/pull/3930)).And we can't be sure that log and meta files are in perfect sync.Therefore, we need to start up with meta information as a baseline for recovery

cigarl:
> > I'd be happy to do it together : )
> 
> OK, then that's my reviews~At current, it seems that the `getAllEntriesAfterCommittedIndex()` will always return empty list in current design? What's your opinion?

see [#3930](https://github.com/apache/iotdb/pull/3930)
In my environment, that is not empty.  It is caused by meta information not being flushed synchronously.

cluster/src/main/java/org/apache/iotdb/cluster/log/manage/serializable/SyncLogDequeSerializer.java
LebronAl: If so, I think it would be better to adjust the function name and comment?

cigarl: That can be adjusted.  But when I name it, I base it on the meta's information, because these logs are uncommitted for the meta(although its info is staled).Maybe,i can add some comments to illustrate the point.

cluster/src/main/java/org/apache/iotdb/cluster/log/manage/RaftLogManager.java
```diff
@@ -125,6 +127,8 @@ protected RaftLogManager(StableEntryManager stableEntryManager, LogApplier appli
...
+        .truncateAndAppend(stableEntryManager.getAllEntriesAfterCommittedIndex());
```
LebronAl, line 131: Actually, these entries are committed at a high level. So we need to `commit` it before we expose them to the outside world, which means we need to revocer them in `commitEntryManager`, and `unCommitEntryManager` should be empty at current design.

cigarl: > which means we need to revocer them in `commitEntryManager`,

That means,we need to update meta's info when recover logs file in this case(meta's info is stale)? that may be difficult _(because we need to consider the meta recovery process in all situations to ensure that leaders and followers are completely consistent)_.

Of course,there is another way to fix it. Every time the log file is flushed, the meta is flushed with it. Except in special 
 situations _(log file flushed successfully but meta failed when we restart cluster)_, it doesn't have any problems.

How do you think?

cluster/src/main/java/org/apache/iotdb/cluster/log/manage/serializable/SyncLogDequeSerializer.java
```diff
@@ -236,6 +237,30 @@ public LogManagerMeta getMeta() {
...
+   * When raft log files flushed,meta would not be flushed synchronously.So data has flushed to disk
+   * is uncommitted for persistent LogManagerMeta(meta's info is stale).We need to recover these
+   * already persistent logs.
+   *
+   * <p>For example,commitIndex is 5 in persistent LogManagerMeta,But the log file has actually been
+   * flushed to 7,when we restart cluster,we need to recover 6 and 7.
+   *
+   * <p>Maybe,we can extract getAllEntriesAfterAppliedIndex and getAllEntriesAfterCommittedIndex
+   * into getAllEntriesByIndex,but now there are too many test cases using it.
+   */
```
neuyilan, line 250: In this comment, it seems due to that the log's meta file does not flush synchronously with the raft log, however, the persistent commit raft log maybe has been applied.
We should not commit the entry again when restart.

cluster/src/main/java/org/apache/iotdb/cluster/log/manage/RaftLogManager.java
LebronAl: If we update the log and meta information of all nodes to the status of all logs persisted and applied after restart, will there be any inconsistency between the leader and follower?

cluster/src/main/java/org/apache/iotdb/cluster/log/manage/serializable/SyncLogDequeSerializer.java
cigarl: The problem here is that we don't know the actual `applyIndex`.
Fortunately, in IOTDB, repeating _apply_ operation doesn't matter too much, but losing data can be unacceptable.
That is only my views.

cluster/src/main/java/org/apache/iotdb/cluster/log/manage/RaftLogManager.java
cigarl: This could not cause consistency problems, but some data may be lost. 
For example, `applyIndex` in persistent meta file is 5, and `commitIndex` is 7.But the actual index flushed to 10,and apply to 7,we may lost 8~10.

chengjianyun:
> To solve these problems completely, I think `raftLogManager` needs a complete refactoring, including but not limited to better concurrency control, better persistence strategy, etc.
> 
> When I first joined the community in my senior year, the first thing I did was to implement `raftLogManager` with reference to `storage` and `unstable` interfaces in `etcd`. Because I didn't have a very deep understanding of `raft` and `etcd` at that time, after a long period, I finally realized that this was probably an awful implementation and I apologize for that. There are two main reasons:
> 
> * For `raft`, uncommitted logs also need to be persisted (i.e., logs on disk may also need to be truncated) in order to ensure correctness after reboot.
> * `etcd` is an event-driven architecture, so its interior is all unlocked. However, in our architecture, `raftLogManager` can be accessed concurrently by multiple threads. So we've added a lot of patches for concurrency control right now, but this is actually an area we can think about implementing better.
> 
> Thank you very much for your great contributions. This week I will focus on the PR you raised about `cluster-` branch and a bug at hand. After that, I would like to discuss the refactoring of `raftLogManager` with you guys and community. And of course you're welcome to start to design the new `raftLogManager` with us right now if you are free.

Thanks for your implementation so that we can enjoy the first cluster version of IoTDB  :). Raft algorithm is really hard to implementation, there are too many corner cases to consideration. We could make the algorithm work correct in most of cases in a short time which is terrific.

In my opinion, we need to spend too much effort to guarantee the correctness of Raft if we want to keep working on the current cluster implementation. Cluster module lack a whole system design according to my feeling. 

Another my suggestion is we can gradually separate raft implementation and raft client(application) just like what `etcd` did. The goal of this is, finally, we can involve `Ratis` to help manage the Raft status. Maybe it's time to put `Ratis` on the agenda. I strongly suggest someone spend some time to investigate `Ratis` and evaluate the cost of integration. Instead of Raft correctness, We could focus on improving the core engine of IoTDB and ecosystem after that.

cluster/src/main/java/org/apache/iotdb/cluster/log/manage/RaftLogManager.java
LebronAl: OK, I see. This is really a matter of ensuring atomicity between apply and commit. We can talk about it later during refactoring, I have no problem with the current implementation

neuyilan:
> Another my suggestion is we can gradually separate raft implementation and raft client(application) just like what `etcd` did. The goal of this is, finally, we can involve `Ratis` to help manage the Raft status. Maybe it's time to put `Ratis` on the agenda. I strongly suggest someone spend some time to investigate `Ratis` and evaluate the cost of integration. Instead of Raft correctness, We could focus on improving the core engine of IoTDB and ecosystem after that.

Actually, I have the same idea as you, I think we can integrate `Apache Ratis` with our project,  so at least we can guarantee the correctness of the raft, and at present, and our energy can be used for some functions of `IoTDB`.  And at present, I think the correctness is far greater than the performance.
But it's really a time-consuming thing. We need to work together to finish it.

cluster/src/main/java/org/apache/iotdb/cluster/log/manage/serializable/SyncLogDequeSerializer.java
neuyilan: As we have so many plans, I am not sure all the plan is idempotent.

cluster/src/main/java/org/apache/iotdb/cluster/log/manage/RaftLogManager.java
chengjianyun: In my opinion, it doesn't matter where the entries to be recovered. The recover progress won't affect the consistency of raft group. `CommitIndex` to be determined until the next new log appending the follower. If my understanding right, the `committed`  logs committed again in the same sequence won't break the correctness of IoTDB. We have no choice even though this is not a good one.

cluster/src/main/java/org/apache/iotdb/cluster/log/manage/serializable/SyncLogDequeSerializer.java
chengjianyun: According to my understanding, re-apply the same sequence(guaranteed by log index) of the plan won't break the correctness of IoTDB. And looks like we don't have a better choice.

chengjianyun:
> > Another my suggestion is we can gradually separate raft implementation and raft client(application) just like what `etcd` did. The goal of this is, finally, we can involve `Ratis` to help manage the Raft status. Maybe it's time to put `Ratis` on the agenda. I strongly suggest someone spend some time to investigate `Ratis` and evaluate the cost of integration. Instead of Raft correctness, We could focus on improving the core engine of IoTDB and ecosystem after that.
> 
> Actually, I have the same idea as you, I think we can integrate `Apache Ratis` with our project, so at least we can guarantee the correctness of the raft, and our energy can be used for some functions of `IoTDB`. And at present, I think the correctness is far greater than the performance.
> But it's really a time-consuming thing. We need to work together to finish it.

Very glad to work together. As I will work on prometheus integration in the later of this month, I only have small bandwidth for this. But I'd like to propose some design sooner or later to discuss together after investigation.

LebronAl:
> Another my suggestion is we can gradually separate raft implementation and raft client(application) just like what `etcd` did. The goal of this is, finally, we can involve `Ratis` to help manage the Raft status. Maybe it's time to put `Ratis` on the agenda. I strongly suggest someone spend some time to investigate `Ratis` and evaluate the cost of integration. Instead of Raft correctness, we can focus on improving the core engine of IoTDB and ecosystem after that.

In fact, In the last few months I have studied Raft algorithm carefully and completed [6.824 lab](https://github.com/LebronAl/MIT6.824-2021). This is just an instructional Raft implementation, which makes me deeply appreciate how difficult it is to implement a correct consensus algorithm, while Raft algorithm at production level requires much more. So I think maybe it's time for us to weigh in and make a decision. Choosing a mature implementation of the raft algorithm liberates our productivity, which may take a long time at first, but can be hugely beneficial once implemented.

But I also want to list two concerns:

- Production-level raft algorithms generally guarantee linearizability, which is the strictest consistency in distributed systems. Our current Raft algorithm may not be complete, so strictly speaking it does not guarantee linearizability, but at the same time it's performance may be better. What would we say if we migrated raft implementation and performance dropped? Furthermore, is linearizability really necessary for OLAP databases like time-series databases?

- At the moment we have Raft implementation and business logic mixed together, but there is a certain amount of performance improvement. For example, we currently put data from multiple storage groups into one Raft group to be executed synchronously. In fact, at the bottom, they can be executed in parallel. So we implemented parallel asynchronous apply optimization so that plans from the same Raft group but different storage groups can be applied in parallel. The performance gains from this optimization are significant. How should we handle this case after we migrate raft implementation? Add parallel apply feature to raft library Or do we change our partitioning pattern so that a Raft group serves only one storage group? Or just let them execute inefficiently?

Any discussions are welcomed~

neuyilan:
> is linearizability really necessary for OLAP databases like time-series databases?

As I see the raft is one consensus algorithm, which may do not have many relations with the linearizability, just put the post[1] for reference.
> we currently put data from multiple storage groups into one Raft group to be executed synchronously.

As far as I know, the apply function is user-defined, we can still implement a parallel apply function according to different storage groups in the one raft log.

What is certain is that using a raft library will definitely limit our optimization work compared with the current implementation (mixing the raft framework and business logic), but I think the availability and correctness are far greater than the performance for now.
[1] https://zhuanlan.zhihu.com/p/47117804

chengjianyun:
> > is linearizability really necessary for OLAP databases like time-series databases?
> 
> As I see the raft is one consensus algorithm, which may do not have many relations with the linearizability, just put the post[1] for reference.
> 
> > we currently put data from multiple storage groups into one Raft group to be executed synchronously.
> 
> As far as I know, the apply function is user-defined, we can still implement a parallel apply function according to different storage groups in the one raft log.
> 
> What is certain is that using a raft library will definitely limit our optimization work compared with the current implementation (mixing the raft framework and business logic), but I think the availability and correctness are far greater than the performance for now.
> [1] https://zhuanlan.zhihu.com/p/47117804

Let's move Ratis related discussion to https://github.com/apache/iotdb/issues/3954 so that more people could join.

cigarl:
> but we may also need a processor to avoid logIndex is too big, dunmmyIndex need to re-initialize if logIndex will be bigger than our expect.

cigarl:
> but we may also need a processor to avoid logIndex is too big, dunmmyIndex need to re-initialize if logIndex will be bigger than our expect.

This may not seem like a problem.I tried to calculate that it would handle 100 million requests per second and the cluster would run for 2924 years.

Sorry, I closed this PR due to my misoperation. I have reopened it.

PR #3913, opened 2021-09-06
[IOTDB-1629] fix the NPE when using value fill in cluster mode

cornmonster:
For more detailed descriptions: https://issues.apache.org/jira/browse/IOTDB-1629

PR #3957, opened 2021-09-14
[To rel/0.12][IOTDB-1629] fix the NPE when using value fill in cluster mode (#3913)

cornmonster:
Cherry-pick changes


LebronAl (approved):
LGTM

PR #3953, opened 2021-09-13
[To rel/0.12][ISSUE-3945] Fuzzy query support multiDevices and alignByDevice Dataset

ijihang:
## Description
ISSUE : #3945

Fuzzy query support multiDevice and alignByDevice

PR #3958, opened 2021-09-14
[ISSUE-3945] Fuzzy query support multiDevices and alignByDevice Dataset

ijihang:
## Description
cherry-pick #3953

Since there is no operator.getType() method, use else if instead of switch


HTHou (approved):
LGTM

PR #3962, opened 2021-09-14
[IOTDB-1678] client-cpp session bug: can cause connection leak.

jamber001:
## Description
[IOTDB-1678] client-cpp session bug: can cause connection leak.


SteveYurongSu (approved):
LGTM. Plz cherry-pick it to rel/0.12 :D

HTHou (approved):

PR #3968, opened 2021-09-15
[To rel/0.12] change sync version check

yschengzi:
change version checking rule for sync process
before:
- Check full version number between sender and receiver.
- Stop sync process if full version number does not match

Now:
- Check major version number between sender and receiver
- Stop sync process if major version number does not match


HTHou (approved):
Cherry pick to master branch.

yschengzi:
> Cherry pick to master branch.

got it

PR #3970, opened 2021-09-15
Change the way of calculating vector avg points

JackieTien97:
The current way of calculating vector's avg points is wrong, we should also take the number of vector's subsensors into account.


HTHou (approved):
LGTM

HTHou (commented):
Plz merge the latest master branch to fix the tests :>

HTHou (approved):

PR #3971, opened 2021-09-15
[IOTDB-1690] Fix align by device type cast error

jun0315:
Use align by device function:

select * from root.database.student(any similar sql) align by device

Will print error:

Msg: 500: [INTERNAL_SERVER_ERROR] Exception occurred: "select * from root.database.student align by device". executeStatement failed. org.apache.iotdb.db.metadata.mnode.EntityMNode cannot be cast to org.apache.iotdb.db.metadata.mnode.IMeasurementMNode

Reason:

Children may be path or root node. ("country" and "score" belong path, and "anhui" belongs path)if it belong to path node, they don't need to type cast and add their schema to the res. Otherwise, if we don't make type  judgment , will be throw type cast exception.

image

The actual stored time series is shown in the figure below.

image

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
@@ -1241,7 +1241,7 @@ public class MManager {
       Template template = node.getUpperTemplate();
 
       for (IMNode child : node.getChildren().values()) {
-        if (child instanceof IMeasurementMNode){
+        if (child instanceof IMeasurementMNode) {
           IMeasurementMNode measurementMNode = (IMeasurementMNode) child;
           res.add(measurementMNode.getSchema());
         }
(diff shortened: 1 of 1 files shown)


HTHou:
@zyk990424 Can you plz take a look?

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -1241,8 +1241,10 @@ public IMNode getDeviceNodeWithAutoCreate(PartialPath path)
       Template template = node.getUpperTemplate();
 
       for (IMNode child : node.getChildren().values()) {
-        IMeasurementMNode measurementMNode = (IMeasurementMNode) child;
-        res.add(measurementMNode.getSchema());
+        if (child instanceof IMeasurementMNode) {
```
zyk990424, line 1244: It's better use override method child.isMeasurement.
Java instansof violates OO design and program.

jun0315: Thanks for your reply!
I have changed to is measurement function.

HTHou (approved):

PR #3979, opened 2021-09-16
[IOTDB-1691] Wrong way of calculating the vector's memory footprint

JackieTien97:
[JIRA-1691](https://issues.apache.org/jira/browse/IOTDB-1691)


server/src/main/java/org/apache/iotdb/db/utils/MemUtils.java
```diff
@@ -101,40 +101,32 @@ public static long getRecordSize(
...
-    int columnCount = 0;
+    boolean hasVector = false;
     for (int i = 0; i < insertTabletPlan.getMeasurementMNodes().length; i++) {
       if (insertTabletPlan.getMeasurementMNodes()[i] == null) {
-        columnCount++;
         continue;
       }
       IMeasurementSchema schema = insertTabletPlan.getMeasurementMNodes()[i].getSchema();
+      TSDataType valueType;
       if (schema.getType() == TSDataType.VECTOR) {
```
HTHou, line 111: Can we use `if (insertTabletPlan.isAlign) ` here?

HTHou, line 111:
```suggestion
      if (insertTabletPlan.isAligned()) {
```
```suggestion
      if (schema.getType() == TSDataType.VECTOR) {
```

HTHou (approved):

PR #3986, opened 2021-09-16
[IOTDB-1577] Alias supports any Unicode character in double quotation

jun0315:
Alias supports any Unicode character in double quotation


SteveYurongSu:
The feature was already supported. Thanks for your contributions!

Issue #3857, opened 2021-08-27
add "show now()" or "select now()"  DML


jixuan1989:
### Discussed in https://github.com/apache/iotdb/discussions/3829

<div type='discussions-op-text'>

<sup>Originally posted by **jiangxuewusheng** August 25, 2021</sup>
iotdb-cli add "show now()" or "select now()" DML.</div>

wqx-229:
I want to join IoTDB community, I want to implement this function. Thanks

PR #3987, opened 2021-09-16
[ISSUE-3857] Support show now() function

wqx-229:
## Description
ISSUE #3857 
Implementation returns the current time of the system. 
Format: "yyyy-MM-dd'T'HH:mm:ss.SSSXXX"
e.g :1970-01-01T08:00:00.001+08:00

### DIFF
--- a/antlr/src/main/antlr4/org/apache/iotdb/db/qp/sql/SqlBase.g4
+++ b/antlr/src/main/antlr4/org/apache/iotdb/db/qp/sql/SqlBase.g4
@@ -109,8 +109,11 @@ statement
     fromClause
     whereClause?
     specialClause? #selectStatement
+    |SHOW timeClause #showNow
     ;
-
+timeClause
+    : NOW LR_BRACKET RR_BRACKET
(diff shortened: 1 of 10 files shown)


HTHou:
Hi, thanks for your contribution!
Can you plz submit this PR to master branch? We actually won't add new feature to the rel/0.12 branch any more.

cornmonster:
Hi, it is really nice to have a way to check the current server time. 

The relational DBMS(s) like MySQL or PostgreSQL use built-in UDF to handle this need. But I since IoTDB doesn't support syntax like `select now()` (a from clause is required), I can totally understand the way you are doing it.

But since adding a new top-level statement is quite have, I suggest we make most of it and make the SQL API more extendable. Instead of simply shows a system time, maybe we could provide more information like current workload, storage usage, and so on.

In addition, in cluster mode, we should display the information for every single node in the cluster.

neuyilan:
How is this operator defined in the cluster module? Because the time of cluster nodes may be inconsistent?

PR #3988, opened 2021-09-16
[IOTDB-1037] parse url param and add ut

jun0315:
parse url param and add UT


HTHou (approved):
LGTM
Should we update the JDBC user guide to introduce this new feature?

jun0315:
> LGTM
> Should we update the JDBC user guide to introduce this new feature?

Thanks for your review, If it needs to be updated, I will update it to the user manual recently.

jun0315:
> LGTM
> Should we update the JDBC user guide to introduce this new feature?

Hi, I have updated the JDBC user guide to introduce this new feature. :D

wangchao316 (approved):
LGTM

PR #3993, opened 2021-09-17
[cluster-refactor] update start class name in shell/bat script and apply spotless

chengjianyun:
As main class name is changed from `ClusterMain` to `ClusterIoTDB`, update the names in scripts.


chengjianyun:
Please help merge the PR to unblock most of IT.

PR #3994, opened 2021-09-17
[IOTDB-1338][To rel/0.12] The content of DDL viewing sub-nodes in the user guide module is wrong

jun0315:
fix user guide doc v0.12


HTHou (approved):
LGTM

PR #3995, opened 2021-09-17
[IOTDB-1514] Support null in insertTablet

HeimingZ:
In the putXxxxs() methods of basic types' TVList, i drop null values by moving null values to the end of time array and value array, so user can insert null in insertTablet method.

Experiment result:
https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=191332453


HTHou:
I just come up with another idea on the train. :) Is it necessary to maintain bitmaps in normal TVList? 

The insertTablet logic in TVList is actually calls the System.arrayCopy to copy the timestamps and values array into TVList. If we check the bitmap passed from outside here and split the arrays accordingly, skip copying the range that the bitmap marked. Then we don’t even need to change the sort, encoding and query logic anymore. 

How do you think? Should we do some experiment to compare the current implement in this pr and the idea above?

HeimingZ:
> I just come up with another idea on the train. :) Is it necessary to maintain bitmaps in normal TVList?
> 
> The insertTablet logic in TVList is actually calls the System.arrayCopy to copy the timestamps and values array into TVList. If we check the bitmap passed from outside here and split the arrays accordingly, skip copying the range that the bitmap marked. Then we don’t even need to change the sort, encoding and query logic anymore.
> 
> How do you think? Should we do some experiment to compare the current implement in this pr and the idea above?

I think it is ok to remove bitmaps in normal TVList, the write performance may a little bit worse than current implement, but more space is saved and the sort performance is much more better.

Should I implement this in another PR?

HTHou:
How do you think? @qiaojialin

example/session/src/main/java/org/apache/iotdb/SessionExample.java
```diff
@@ -433,6 +435,55 @@ private static void insertTablet() throws IoTDBConnectionException, StatementExe
...
+        // mark null value
+        if (row % 3 == s) {
+          bitMaps[s].mark((int) row);
+        }
+        tablet.addValue(schemaList.get(s).getMeasurementId(), rowIndex, value);
+      }
+      if (tablet.rowSize == tablet.getMaxRowNumber()) {
+        session.insertTablet(tablet, true);
+        tablet.reset();
+      }
```
HTHou, line 477: This is the method 1 of adding tablet data in insertTablet() method, I think we should also show the method 2 here.

HTHou (commented):
Almost look good to me! Just minor issue.
1. add the insert Tablet with null value in SessionExample by method 2.
2. consider the last cache when insertTabletPlan is executing

example/session/src/main/java/org/apache/iotdb/SessionExample.java
HeimingZ: Already fixed these two issues.

server/src/main/java/org/apache/iotdb/db/utils/datastructure/TVList.java
```diff
@@ -165,27 +167,27 @@ public void putVector(long time, Object[] value) {
...
 
-  public void putInts(long[] time, int[] value, int start, int end) {
+  public void putInts(long[] time, BitMap bitMap, int[] value, int start, int end) {
     throw new UnsupportedOperationException(ERR_DATATYPE_NOT_CONSISTENT);
   }
 
-  public void putFloats(long[] time, float[] value, int start, int end) {
+  public void putFloats(long[] time, BitMap bitMap, float[] value, int start, int end) {
     throw new UnsupportedOperationException(ERR_DATATYPE_NOT_CONSISTENT);
   }
```
HTHou, line 180: How about change the order of the params?
Eg. `public void putFloats(long[] time, float[] value, BitMap bitMap, int start, int end) `

HTHou (approved):
LGTM

server/src/main/java/org/apache/iotdb/db/utils/datastructure/TVList.java
HeimingZ: Already Fixed. I also change params  order of relevant methods.

HTHou (approved):

Issue #2177, opened 2020-12-03
There is a vulnerability in Apache Hadoop 2.7.3 ,upgrade recommended 


QiAnXinCodeSafe:
https://github.com/apache/iotdb/blob/d951e82f9509b3772ff9a9fe3bf1444f406d4afc/pom.xml#L108

CVE-2017-15718  CVE-2018-1296 CVE-2017-3166 CVE-2018-8009

Recommended upgrade version：
2.10.0

PR #3998, opened 2021-09-17
[IOTDB-1713] Run IoTDB CI with JDK17

HTHou:
## Description

The LTS version JDK 17 is released on 2021/09/14. This PR is checking if the CI works with JDK17 and update the JDK16 Github action workflow to JDK17.

This PR also fixed the #2177, because the hadoop-tsfile test cases would fail if using the hadoop 2.7.3 with JDK17.


.github/workflows/main-win.yml
```diff
@@ -14,7 +14,7 @@ on:
     branches:
       - master
       - 'rel/*'
-      - cluster_new
+      - JDK17
```
HTHou, line 17:
```suggestion
      - cluster_new
```

SteveYurongSu (approved):
LGTM

PR #3999, opened 2021-09-17
[To rel/0.12] [IOTDB-1484] auto create schema on 0.12

MrQuansy:


neuyilan:
Thanks for your effort, I find you have post https://github.com/apache/iotdb/pull/3879 in the master branch, however, when I use benchmark insertRecords interface with auto-create schame to test, the cluster is stuck and no timeseries is inserted.

LebronAl:
> Thanks for your effort, I find you have post #3879 in the master branch, however, when I use benchmark insertRecords interface with auto-create schame to test, the cluster is stuck and no timeseries is inserted.

We've added detailed test cases for this [enhancement](https://github.com/apache/iotdb/pull/3879/files#diff-f4565abb8dbfb5577127c4c2e8704031c6f632a07bb5f35084b57f0688f9c0fe), and it works in most scenarios. The problem you raised should be one that will recur in a special scene. @MrQuansy  has already established an [issue](https://issues.apache.org/jira/browse/IOTDB-1724) and will continue to follow up this problem after this PR's merging. In order not to block the 0.12.3 release, I will merge the PR.

PR #4002, opened 2021-09-18
[IOTDB-1650] Rename the sql command `move`

HeimingZ:
MOVE command is confusing, it should move some TsFile to the target directory, but it also unloads the TsFile from IoTDB and the data moved cannot be searched any more. So, i think UNLOAD is more suitable than MOVE.


HTHou (approved):
LGTM

HTHou:
Consider add some user guide of `unload` and `remove` command?

HeimingZ:
> Consider add some user guide of `unload` and `remove` command?

Already modify user guide 'Load External TsFile Tool' under 'System Tools'. :)

HTHou (approved):

PR #4007, opened 2021-09-22
[IOTDB-1718] correct the position of "chmod + cmake"

jamber001:
[IOTDB-1718]  compile-tools/thrift/pom.xml, make-cmake-executable use wrong phase

compile-tools/thrift/pom.xml:62

```xml
<plugin>
<groupId>org.codehaus.mojo</groupId>
<artifactId>exec-maven-plugin</artifactId>
<version>1.6.0</version>
<executions>
<execution>
<id>make-cmake-executable</id>
<phase>process-sources</phase> //here, this phase is too late, becasue "cmake-generate" begin to run cmake file in <phase>generate-sources</phase>
<goals>
<goal>exec</goal>
</goals>
<configuration>
<basedir>${cmake.root.dir}/bin</basedir>
<executable>${chmod.command}</executable>
<commandlineArgs>${chmod.command.para}</commandlineArgs>
</configuration>
</execution>
</executions>
</plugin>
```

 "chmod +x cmake" should be run before running cmake.

### DIFF
--- a/compile-tools/thrift/pom.xml
+++ b/compile-tools/thrift/pom.xml
@@ -54,25 +54,6 @@
             <build>
                 <plugins>
                     <plugin>
-                        <groupId>org.codehaus.mojo</groupId>
-                        <artifactId>exec-maven-plugin</artifactId>
-                        <version>1.6.0</version>
-                        <executions>
-                            <execution>
-                                <id>make-cmake-executable</id>
-                                <phase>process-sources</phase>
-                                <goals>
-                                    <goal>exec</goal>
-                                </goals>
-                                <configuration>
-                                    <basedir>${cmake.root.dir}/bin</basedir>
-                                    <executable>${chmod.command}</executable>
-                                    <commandlineArgs>${chmod.command.para}</commandlineArgs>
-                                </configuration>
-                            </execution>
-                        </executions>
-                    </plugin>
-                    <plugin>
                         <groupId>com.googlecode.maven-download-plugin</groupId>
                         <artifactId>download-maven-plugin</artifactId>
                         <version>1.3.0</version>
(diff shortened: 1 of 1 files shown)


jamber001:
@HTHou    Can this PR  be merged ?

HTHou:
> @HTHou Can this PR be merged ?

Sorry, I don't get the reason why this change can solve the problem you mentioned. Can you plz explain more?

jamber001:
> > @HTHou Can this PR be merged ?
> 
> Sorry, I don't get the reason why this change can solve the problem you mentioned. Can you plz explain more?

I  try to give 1 summary for this part.  :-)
In orginal IoTDB maven design, client-cpp's compiling procedure is as below:
1) download cmake-3.17.3-Linux-x86_64.tar.gz and untar it.  (in maven phase <**generate-sources**>)
2) download thrif 0.14.1 src code tar and untar it  (in maven phase <**generate-sources**>)
3) use cmake to process thrift 0.14.1 src codes. (in maven phase <**generate-sources**>)
4)  chmod +x cmake     (in maven phase <**process-sources**>)
5) compile thrift 0.14.1  src code.  (in maven phase <**compile**>)  
6)  more thrift generating work.   

so, in above procedure, the step-4's postion should be 1 bug. 
I guess that step-4 's orginal goal is to avoid new-downloaded cmake file has no executable permission, and step-4 should be run before step-3.
But in origin codes, the step-4 was put in wrongly maven phase **process-sources**,  which casues it is run after step-3.

**Note:** Of course, if the new-downloaded cmake file is origianly executable, the step-4 is uesless.

My codes change is that:
1) change the step-4's maven phase to **generate-sources**
2) adjust step-4's position in pom.xml.
so let "chmod +x cmake"  happen just after step-2.

HTHou (approved):
Thanks for your explanation. It makes sense.

PR #4015, opened 2021-09-23
[to rel/0.12] Fix concurrency issue in compaction selection

THUMarkLau:
Currently, there are two submission entries for compaction tasks in IoTDB: one is triggered periodically, and the other is to manually execute the `merge` statement in the CLI. Under normal circumstances, these two portals will not submit merge tasks at the same time. But in some cases, these two entries are triggered together, it is possible to submit the same merge task, causing concurrency problems.
1ef0deaf8a07b11cec0a360c1b68980


server/src/test/java/org/apache/iotdb/db/engine/merge/ConcurrentMergeTest.java
```diff
@@ -0,0 +1,131 @@
...
+    Thread mergeThreadTwo = new Thread(() -> processor.merge());
+    mergeThreadOne.start();
+    mergeThreadTwo.start();
+    try {
+      Thread.sleep(1000);
+    } catch (Exception e) {
+
+    }
+    while (processor.getTsFileManagement().isSeqMerging) {
+      // wait
```
qiaojialin, line 107: add  a sleep and timeout mechanism

server/src/test/java/org/apache/iotdb/db/integration/IoTDBMergeIT.java
```diff
@@ -380,7 +380,7 @@ public void testShowMergeStatus() throws SQLException {
       }
       // it is uncertain whether the sub tasks are created at this time point, and we are only
       // sure that the main task is created
-      assertEquals(4, cnt);
+      assertEquals(0, cnt);
```
qiaojialin, line 383: why modify this

qiaojialin (approved):

PR #4026, opened 2021-09-24
[IOTDB-1733] Fix dropping built-in function

ericpai:
See JIRA https://issues.apache.org/jira/browse/IOTDB-1733

PR #4027, opened 2021-09-24
[To rel/0.12][IOTDB-1733] Fix dropping built-in function

ericpai:
This is the cherry-pick of #4026 

Related JIRA: https://issues.apache.org/jira/browse/IOTDB-1733

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/query/udf/service/UDFRegistrationService.java
+++ b/server/src/main/java/org/apache/iotdb/db/query/udf/service/UDFRegistrationService.java
@@ -203,7 +203,7 @@ public class UDFRegistrationService implements IService {
 
   public void deregister(String functionName) throws UDFRegistrationException {
     functionName = functionName.toUpperCase();
-    UDFRegistrationInformation information = registrationInformation.remove(functionName);
+    UDFRegistrationInformation information = registrationInformation.get(functionName);
     if (information == null) {
       String errorMessage = String.format("UDF %s does not exist.", functionName);
       logger.warn(errorMessage);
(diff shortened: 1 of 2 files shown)


ericpai:
@SteveYurongSu this cherry-pick has a little difference of the master commit because of the `TEMPORARY` special logics.

server/src/main/java/org/apache/iotdb/db/query/udf/service/UDFRegistrationService.java
```diff
@@ -216,7 +216,7 @@ public void deregister(String functionName) throws UDFRegistrationException {
       logger.error(errorMessage);
       throw new UDFRegistrationException(errorMessage);
     }
-
+    registrationInformation.remove(functionName);
```
SteveYurongSu, line 219: I think it's better to move the `remove` statement to the last. What do you think?  :D

```suggestion
    if (!information.isTemporary()) {
      try {
        appendDeregistrationLog(functionName);
      } catch (IOException e) {
        registrationInformation.put(functionName, information);
        String errorMessage =
            String.format(
                "Failed to append UDF log when deregistering UDF %s, because %s", functionName, e);
        logger.error(errorMessage);
        throw new UDFRegistrationException(errorMessage, e);
      }
    }
    
    registrationInformation.remove(functionName);
```

wangchao316: @SteveYurongSu  good,  code should be the same as master branch.  it's better to move the remove statement to the last.

ericpai: @SteveYurongSu That's the best solution! A new commit is pushed.

SteveYurongSu (approved):
Good job!

wangchao316 (approved):
LGTM

PR #4033, opened 2021-09-26
[IOTDB-1739] Constant timeseries generating functions

SteveYurongSu:
Using constants in nested expressions is a common requirement.

With constant timeseries generating functions, a lot of calculations are going to be very convenient.


SteveYurongSu:
@ericpai Take a look? 😁

ericpai:
> @ericpai Take a look? 😁

Sure~

server/src/test/java/org/apache/iotdb/db/integration/IoTDBUDTFBuiltinFunctionIT.java
```diff
@@ -295,4 +295,39 @@ public void testVariationTrendCalculationFunction(String functionName, double ex
...
+          actual.append(resultSet.getString(1 + j)).append(", ");
+        }
+        assertEquals(expected[i], actual.toString());
+      }
+
+      assertFalse(resultSet.next());
+    } catch (SQLException throwable) {
+      fail(throwable.getMessage());
+    }
+  }
```
ericpai, line 332: May add some more test cases.
1. Invalid queries to test the validate rule.
2. Wildcard paths as the UDF parameter.
3. Invalid paths as the UDF parameter.

And I'm not very sure whether the nested function query is available currently. If so, is this working well in complicated queries lessing IT cases to prove, or the following queries are not allowed by design? 

Such as `SELECT const(avg(s1), 'value'='1024', 'type'='int64')`, `SELECT const(sin(avg(s1)), 'value'='1024', 'type'='int64')` and so on.

docs/zh/UserGuide/IoTDB-SQL-Language/DML-Data-Manipulation-Language.md
```diff
@@ -432,6 +432,43 @@ Total line number = 5
...
 
+#### 常序列生成函数
+
+常序列生成函数用于生成所有数据点的值都相同的时间序列。
+
+常序列生成函数接受一个或者多个时间序列输入，其输出的数据点的时间戳集合是这些输入序列时间戳集合的并集。
+
+目前 IoTDB 支持如下常序列生成函数：
+
+| 函数名 | 输入序列类型                                    | 必要的属性参数                                               | 输出序列类型               | 功能描述                                                     |
```
ericpai, line 443: As we only care about the union of the timestamps of each input timeseries, the data type doesn't matter. So the docs can remove the 'Allowed Input Series Data Types' column. What do you think?

SteveYurongSu: Good idea!

server/src/test/java/org/apache/iotdb/db/integration/IoTDBUDTFBuiltinFunctionIT.java
SteveYurongSu: Thanks for your review!

> May add some more test cases.
> 1. Invalid queries to test the validate rule.

I will add tests to test the validation rule.

> 2. Wildcard paths as the UDF parameter.
> 3. Invalid paths as the UDF parameter.

These 2 are well tested in other ITs, so I think those tests are not necessary here.

> And I'm not very sure whether the nested function query is available currently. If so, is this working well in complicated queries lacking IT cases to prove, or the following queries are not allowed by design?

The nested expression (#4012) is available now! :D

Please see the user docs for more detail: https://github.com/apache/iotdb/blob/master/docs/UserGuide/IoTDB-SQL-Language/DML-Data-Manipulation-Language.md#nested-query-subquery

Functions of all kinds (multi inputs, single inputs, row-by-row iteration, window-by-window iteration, etc), arithmetic operators, and time series are proved to work well in nested expression by several ITs. The code coverage of the expression framework is about 90%. (See https://github.com/apache/iotdb/blob/master/server/src/test/java/org/apache/iotdb/db/integration/IoTDBNestedQueryIT.java)

So I think we don't need to test all functions in nested expressions. The only thing we need to do is to check the basic behavior of the function, which will simplify our work.

ericpai: > So I think we don't need to test all functions in nested expressions. The only thing we need to do is to check the basic behavior of the function, which will simplify our work.
Agreed.

SteveYurongSu:
@ericpai Updated~ 😆😆😆

docs/UserGuide/IoTDB-SQL-Language/DML-Data-Manipulation-Language.md
```diff
@@ -423,6 +423,43 @@ Total line number = 5
...
+
+Currently, IoTDB supports the following constant timeseries generating functions:
+
+| Function Name | Required Attributes                                          | Output Series Data Type                      | Description                                                  |
+| ------------- | ------------------------------------------------------------ | -------------------------------------------- | ------------------------------------------------------------ |
+| CONST         | `value`: the value of the output data point <br />`type`: the type of the output data point, it can only be INT32 / INT64 / FLOAT / DOUBLE / BOOLEAN / TEXT | Determined by the required attribute  `type` | Output the user-specified constant timeseries according to the  attributes `value` and `type`. |
+| PI            | None                                                         | DOUBLE                                       | Data point value: a double value that is closer than any other to `pi`, the ratio of the circumference of a circle to its diameter. |
```
ericpai, line 437: This statement `a double value that is closer than any other to` is a little tricky and hard to understand. How about 
`a double value of `pi`, the ratio of the circumference of a circle to its diameter, with errors less than 1e-6`?
And I think we can use the letter π instead of `pi` in the description,  everyone may know this letter :)

SteveYurongSu: > a double value that is closer than any other to

Well, I copied the statement from the Java Standard Library... :D

I will modify the statement referring to yours, thanks for your suggestions!

SteveYurongSu:
@ericpai Please take a look again :D

ericpai:
> @ericpai Please take a look again :D

The code is OK now. Here I have another one question about the design.

_Does the result of wildcard path query of `const` function really satisfy user's actual requirement?_ 

   For example there're 2 timeseries root.sg.d1.s1, root.sg.d1.s2. And we have a query `SELECT const(*, type='int32', value='10') FROM root.sg.d1`.  In our logic, the function will be separated into two columns `const(root.sg.d1.s1, type='int32', 'value'='10')` and `const(root.sg.d1.s2, type='int32', 'value'='10')` right? But maybe a user want's to a single column with the union timestamps of all the measurements of `root.sg.d1`.  Moreover the user can't specify every timeseries in the parameters as there're so many ones.

SteveYurongSu:
@ericpai 

> For example there're 2 timeseries root.sg.d1.s1, root.sg.d1.s2. And we have a query `SELECT const(*, type='int32', value='10') FROM root.sg.d1`. In our logic, the function will be separated into two columns `const(root.sg.d1.s1, type='int32', 'value'='10')` and `const(root.sg.d1.s2, type='int32', 'value'='10')` right?

You are right.

> _Does the result of wildcard path query of `const` function really satisfy user's actual requirement?_
> 
>  Maybe a user wants a single column with the union timestamps of all the measurements of `root.sg.d1`. Moreover the user can't specify every timeseries in the parameters as there're so many ones.

I understand your concern, and I've been aware of the issue for a long time.

I think this is a semantic issue of the * symbol in the UDF parameters, not just a semantic issue of the `const` function.

In fact, we already have a JIRA to record the issue, we can move the discussion on it: 

[IOTDB-1397] https://issues.apache.org/jira/browse/IOTDB-1397

Welcome to re-define (and re-implement :D) the semantic of the * symbol in the UDF parameters!

ericpai (approved):

ericpai:
> @ericpai
> 
> > For example there're 2 timeseries root.sg.d1.s1, root.sg.d1.s2. And we have a query `SELECT const(*, type='int32', value='10') FROM root.sg.d1`. In our logic, the function will be separated into two columns `const(root.sg.d1.s1, type='int32', 'value'='10')` and `const(root.sg.d1.s2, type='int32', 'value'='10')` right?
> 
> You are right.
> 
> > _Does the result of wildcard path query of `const` function really satisfy user's actual requirement?_
> > Maybe a user wants a single column with the union timestamps of all the measurements of `root.sg.d1`. Moreover the user can't specify every timeseries in the parameters as there're so many ones.
> 
> I understand your concern, and I've been aware of the issue for a long time.
> 
> I think this is a semantic issue of the * symbol in the UDF parameters, not just a semantic issue of the `const` function.
> 
> In fact, we already have a JIRA to record the issue, we can move the discussion on it:
> 
> [[IOTDB-1397](https://issues.apache.org/jira/browse/IOTDB-1397)] https://issues.apache.org/jira/browse/IOTDB-1397
> 
> Welcome to re-define (and re-implement :D) the semantic of the * symbol in the UDF parameters!

Yes, the wildcard in function parameter should be more clear and closer to our user business scenarios. I will watch out the real requirements of our users. if I'm certain about some usages, I will discuss in that JIRA thread.

I have approved this PR.

ericpai:
@SteveYurongSu BTW, I'm glad to join the following discussions about UDF design, bugfix and code reviews. If anything is kicked off, it's better to send a mail or wechat group message to let me get the notice. Or I may lack of some context :D

wangchao316 (approved):
LGTM

ericpai (approved):

SteveYurongSu:
@ericpai

> I will watch out the real requirements of our users. if I'm certain about some usages, I will discuss in that JIRA thread.
 
This issue has been mentioned by many users, it's time to start the discussion! Looking forwards to your ideas! 

> @SteveYurongSu BTW, I'm glad to join the following discussions about UDF design, bugfix and code reviews. If anything is kicked off, it's better to send a mail or wechat group message to let me get the notice. Or I may lack of some context :D

Sure! Let's make the UDF function better!

PR #4034, opened 2021-09-26
[IOTDB-1738] Cache paths list in batched insert plan

ericpai:
See JIRA https://issues.apache.org/jira/browse/IOTDB-1738

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
@@ -52,7 +52,13 @@ import org.apache.iotdb.db.metadata.utils.MetaUtils;
 import org.apache.iotdb.db.qp.constant.SQLConstant;
 import org.apache.iotdb.db.qp.physical.BatchPlan;
 import org.apache.iotdb.db.qp.physical.PhysicalPlan;
-import org.apache.iotdb.db.qp.physical.crud.*;
+import org.apache.iotdb.db.qp.physical.crud.InsertMultiTabletPlan;
+import org.apache.iotdb.db.qp.physical.crud.InsertPlan;
+import org.apache.iotdb.db.qp.physical.crud.InsertRowPlan;
+import org.apache.iotdb.db.qp.physical.crud.InsertRowsOfOneDevicePlan;
+import org.apache.iotdb.db.qp.physical.crud.InsertRowsPlan;
+import org.apache.iotdb.db.qp.physical.crud.InsertTabletPlan;
+import org.apache.iotdb.db.qp.physical.crud.SetSchemaTemplatePlan;
 import org.apache.iotdb.db.qp.physical.sys.CreateAlignedTimeSeriesPlan;
 import org.apache.iotdb.db.qp.physical.sys.CreateMultiTimeSeriesPlan;
 import org.apache.iotdb.db.qp.physical.sys.CreateTimeSeriesPlan;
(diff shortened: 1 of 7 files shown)


cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
```diff
@@ -563,10 +569,11 @@ public void createSchema(PhysicalPlan plan) throws MetadataException, CheckConsi
 
   private List<PartialPath> getValidStorageGroups(PhysicalPlan plan) {
     List<PartialPath> paths = new ArrayList<>();
-    for (int i = 0; i < plan.getPaths().size(); i++) {
+    List<PartialPath> originalPaths = plan.getPaths();
+    for (int i = 0; i < originalPaths.size(); i++) {
```
ericpai, line 573: Here I have some questions about this method. The key of getResults()  is the index of the insertRowPlan, not the path. Is this statement  a bug?

neuyilan: The `(BatchPlan) plan).getResults().containsKey(i)` here are those plans that do not have authority to operate, so here we do not process those plans.

neuyilan (approved):
LGTM

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
ericpai: The variable `i` is the index of plan.getPaths(), not the sub insert plans. 
Let's take an InsertRowsPlan as example, which equals `insert into root.sg.d1(time, s1, s2, s3) values(1, 1, 2, 3); insert into(time, s1, s2, s3) values(2, 4, 5, 6);`
1. The plan.getPaths() returns `[root.sg.d1.s1, root.sg.d1.s2, root.d1.s3]` (the order is not deterministic as it's a set to list). The `insertRowPlanList` contains two elements `(InsertRowPlan, [s1,s2,s3], [1,2,3])` and `(InsertRowPlan, [s1,s2,s3], [4,5,6])`.
2. The getResults() map may only contain key `0` and `1` as there're only 2 InsertRowPlan.
3. `i` is the path index generated from getPaths() list, and there's no relationship with the plan index.

neuyilan: Thanks for point out.
Yes, this is one bug, it should be used `getPrefixPath()` rather than the` plan.getPaths()`
`for (int i = 0; i < plan.getPrefixPath().size(); i++)
...
 `

ericpai: That makes sense! And I think prefixPaths should be cached as well. What do you think?

ericpai: However, in InsertRowsPlan's getPrefixPaths(), it uses a set to deduplicate the prefix paths, which losing the prefixPath indexing information for using here. I think it can't resolve this issue.

neuyilan: Yes, I find only `private void pullTimeseriesSchema(InsertPlan plan, RaftNode ignoredGroup)` method in `BaseAppIier` class used the `getPrefixPaths() `methods and it have logic combine the same sg when pull timeseries, so I think we can just change the `Set` to `List`, how do you think?

ericpai: That's right! And I think the prefixPaths should be cached in the plan as well

neuyilan: > And I think the prefixPaths should be cached in the plan as well

It's ok for me

ericpai: I have pushed a new commit. It includes:
1. Fix this issue.
2. Cache getPrefixPath.
3. Leave and move the de-duplicate logic to BasicApplier, which may help to speed up the process by former design.

ericpai:
@neuyilan Please review these changes again. I find another issue in `CreateMultiTimeseriesPlan.getPrefixPaths()` and have fixed it.

neuyilan (approved):

wangchao316 (approved):
LGTM

ericpai:
@neuyilan Is this ready to be merged? I think maybe rel/0.12 have the same issue

neuyilan:
> @neuyilan Is this ready to be merged? I think maybe rel/0.12 have the same issue

Yes, better merged to the rel/0.12 branch.

PR #4037, opened 2021-09-27
[To rel/0.12] [IOTDB-1741] Avoid double close in level compaction execution

THUMarkLau:
## Description
image
Sometimes in level compaction, the output stream of target file could be closed twice, which will lead to IOException. In `TsFileIOWriter.endFile`, the stream will be close the first time. In `TsFileIOWriter.close`, the stream will be close the second time. The program should check the status of writer before the second close.
image


qiaojialin (approved):

server/src/main/java/org/apache/iotdb/db/engine/compaction/utils/CompactionUtils.java
```diff
@@ -431,7 +431,10 @@ public static void merge(
       writer.endFile();
       targetResource.close();
     } finally {
-      writer.close();
+      if (writer.canWrite()) {
+        // avoid double close
```
wangchao316, line 435: I feel we should set "canWrite" to volatile.
Because if two threads close concurrency， this will be not find.
This mistake is inevitable.
do you think ? @qiaojialin

PR #4042, opened 2021-09-27
[IOTDB-1749] sync-tool's lockInstance() dose not take effect

jamber001:
[IOTDB-1749] sync-tool's lockInstance() dose not take effect

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/sync/sender/transfer/SyncClient.java
+++ b/server/src/main/java/org/apache/iotdb/db/sync/sender/transfer/SyncClient.java
@@ -171,7 +171,9 @@ public class SyncClient implements ISyncClient {
    * @param lockFile lock file
    */
   private boolean lockInstance(File lockFile) {
-    try (final RandomAccessFile randomAccessFile = new RandomAccessFile(lockFile, "rw")) {
+    RandomAccessFile randomAccessFile = null;
+    try {
+      randomAccessFile = new RandomAccessFile(lockFile, "rw");
       final FileLock fileLock = randomAccessFile.getChannel().tryLock();
       if (fileLock != null) {
         Runtime.getRuntime()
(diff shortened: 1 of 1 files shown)


HTHou:
@yschengzi

server/src/main/java/org/apache/iotdb/db/sync/sender/transfer/SyncClient.java
```diff
@@ -171,7 +171,8 @@ public void verifySingleton() throws IOException {
    * @param lockFile lock file
    */
   private boolean lockInstance(File lockFile) {
-    try (final RandomAccessFile randomAccessFile = new RandomAccessFile(lockFile, "rw")) {
+    try {
+      final RandomAccessFile randomAccessFile = new RandomAccessFile(lockFile, "rw");
```
wangchao316, line 175: randomAccessFile.close() will be not call?

jamber001: > randomAccessFile.close() will be not call?

Please review next 15 line of codes below this change.
randomAccessFile will be released when Process exit.

jamber001: image

jamber001: @wangchao316   Has my reply answer your question?

wangchao316: The first thread acquires the lock, addShutdownHook.
The second thread executes new RandomAccessFile, but tryLock() fails. Do you want to close RandomAccessFile?
We can't just rely on close in the hook. If there is no hook, there is no close file.
@jamber001

jamber001: > The first thread acquires the lock, addShutdownHook.
> The second thread executes new RandomAccessFile, but tryLock() fails. Do you want to close RandomAccessFile?
> We can't just rely on close in the hook. If there is no hook, there is no close file.
> @jamber001

I understand your thought, now.  
Though this will not affect sync-tool process, but from function view, it is better to close RandomAccessFile.
I will add codes for this.

PR #4044, opened 2021-09-27
[To rel/0.12] [IOTDB-1726] Wrong hashCode() and equals() method in ChunkMetadata

liuminghui233:
We can see that chunkMetadata use version to differentiate, however, in different storage groups, different file's version may be same. That means root.sg1.d1.s1 and root.sg2.d1.s1 may share the same cache entry in ChunkCache.


tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/ChunkMetadata.java
```diff
@@ -224,16 +228,12 @@ public boolean equals(Object o) {
     ChunkMetadata that = (ChunkMetadata) o;
     return offsetOfChunkHeader == that.offsetOfChunkHeader
         && version == that.version
-        && Objects.equals(measurementUid, that.measurementUid)
-        && tsDataType == that.tsDataType
-        && Objects.equals(deleteIntervalList, that.deleteIntervalList)
-        && Objects.equals(statistics, that.statistics);
+        && partialFilePathForCache.equals(that.partialFilePathForCache);
```
wangchao316, line 231: hi,  equals and hashcode can not only use for cache,  ChunkMetadata have other place for judgments. 
test case have some fail.

PR #4063, opened 2021-09-30
[To influxdb-protocal] Compatible influxdb protocal

jun0315:
At present, many users use influxdb as a time series database. If users change to another database, there will be a relatively high migration cost. The PR goal is to develop an adapter that is compatible with the protocol of influxdb.

Project requirements come from https://summer.iscas.ac.cn/#/org/prodetail/210070151

The objectives actually achieved include:

Support influxdb compatible insertion function.

Support the basic query function compatible with incluxdb, including tag value filtering, file value filtering and time value filtering.

Support some advanced functions compatible with incluxdb, including common count, first, last, Max, mean, median, min, mode, spread, StdDev and sum functions.

The work completed at the code design level includes:

Firstly, the performance of two relational mapping schemes is tested (including large-scale 20W data insertion and query test code), and finally a better scheme is determined for development.

Complete the data schema mapping of the two protocols at the code level.

Complete the data structure conversion of query results.

Complete the g4 syntax of antlr4 of the query SQL of inluxdb and the generation of the corresponding traversal data structure.


SteveYurongSu (approved):
LGTM. Thanks for your contribution!

PR #4083, opened 2021-10-07
[IOTDB-1742] Fix test bug caused by restart error of virtual storage groups

HeimingZ:
Recently, i see following error message in the IoTDBCompleteIT's log:

IoTDBCompleteIT error

To resolve this error, there are two places to modify:

Firstly, It is dangerous that IoTDBCompleteIT uses HashVirtualPartitioner's setStorageGroupNum method to set the number of virtual storage groups, because VirtualStorageGroupManager initializes StorageGroupProcessor array when it's constructed and the call of setStorageGroupNum method will cause sg num conflicts between HashVirtualPartitioner and VirtualStorageGroupManager, then throw ArrayIndexOutOfBoundsException. So, i remove the setStorageGroupNum method of HashVirtualPartitioner.

Secondly, the restart of StorageEngine in EnvironmentUtils.restartDaemon() is confusing because the start method only start timed service except the initialization of storage groups. So, i move the initialization of storage groups in StorageEngine from constructor to start method.


SteveYurongSu:
Closes issue [IOTDB-1742](https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-1742?filter=allopenissues).

PR #4091, opened 2021-10-08
[IOTDB-1798] Compatibility of Apache IoTDB with InfluxDB - `flush` / `ping` / `version`

jun0315:
complete close, ping and version func


jun0315:
Thanks for your review. I have fix it. :D

PR #4077, opened 2021-10-05
[IOTDB-1792]remove tomcat-embed dependency and make all transitive dependencies versions consistent

jixuan1989:
This PR upgrades the following dependencies:

httpclient: 4.5.12,4.5.2,4.3.5 -> 4.5.13
jetty-server: 9.4.35 -> 11.0.6
junit: 4.12 -> 4.13.2
httpcore: xxx -> 4.4.13
tomcat-embed-core: removed
net.java.dev.jna: 5.5.0
zookeeper: 3.4.9
commons-beanutils: 1.9.4
commons-compress:1.21
error_prone_annotations:2.5.1 -> 2.7.1

PR #4103, opened 2021-10-11
[To rel/0.12][IOTDB-1792] remove tomcat-embed dependency and make all transitive dependencies versions consistent

SteveYurongSu:
Cherry-picked from #4077


SteveYurongSu:
@LIU-WEI-git

PR #4146, opened 2021-10-13
[To rel/0.12][IOTDB-1837] Fix tagIndex rebuild failure after upgrade mlog from mlog.txt to mlog.bin

MarcosZyk:
## Description
When upgrade mlog from mlog.txt to mlog.bin, the data of tags and attributes are not read to the mlog.bin which results tagIndex rebuild failure. Thus add tagFile read operation while upgrade mlog.


HTHou:
Cherry pick to master.

PR #4162, opened 2021-10-14
[IOTDB-1848] Failed to initialize pool: Does not support setReadOnly

xinzhongtianxia:
see jira
https://issues.apache.org/jira/browse/IOTDB-1848

### DIFF
--- a/jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBConnection.java
+++ b/jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBConnection.java
@@ -327,8 +327,10 @@ public class IoTDBConnection implements Connection {
   }
 
   @Override
-  public void setReadOnly(boolean arg0) throws SQLException {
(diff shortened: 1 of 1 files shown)


HTHou:
Hi, we do support setReadOnly from v0.12.2. Can we implement this method by the way?

https://iotdb.apache.org/UserGuide/Master/IoTDB-SQL-Language/Maintenance-Command.html#set-ststem-to-readonly-writable

xinzhongtianxia:
> Hi, we do support setReadOnly from v0.12.2. Can we implement this method by the way?
> 
> https://iotdb.apache.org/UserGuide/Master/IoTDB-SQL-Language/Maintenance-Command.html#set-ststem-to-readonly-writable

That's are two concepts. JDBC's readonly is a connection scope, in this connection, user can only perform read operations and some databases will also do some optimization to this conneciton like transactional and so on. In other words, out iotdb can work in normal mode both support read and write, which has no relation with readonly conneciton.

PR #4167, opened 2021-10-15
[IOTDB-1851] Optimize iot-grafana code and fix some bugs

xinzhongtianxia:
see jira
https://issues.apache.org/jira/browse/IOTDB-1851


grafana/src/main/java/org/apache/iotdb/web/grafana/controller/DatabaseConnectController.java
```diff
@@ -69,15 +68,12 @@ public void testDataConnection(HttpServletResponse response) throws IOException
...
    * get metrics numbers in JSON string structure.
    *
-   * @param request http request
-   * @param response http response
    * @return metrics numbers in JSON string structure
    */
   @RequestMapping(value = "/search")
   @ResponseBody
-  public String metricFindQuery(HttpServletRequest request, HttpServletResponse response) {
+  public String metricFindQuery() {
```
SteveYurongSu, line 75: Why can we make this change here?

grafana/src/main/java/org/apache/iotdb/web/grafana/controller/DatabaseConnectController.java
```diff
@@ -88,58 +84,52 @@ public String metricFindQuery(HttpServletRequest request, HttpServletResponse re
...
       Pair<ZonedDateTime, ZonedDateTime> timeRange = getTimeFromAndTo(jsonObject);
-      JsonArray array = (JsonArray) jsonObject.get("targets"); // []
+      JsonArray array = jsonObject.getAsJsonArray("targets");
       JsonArray result = new JsonArray();
       for (int i = 0; i < array.size(); i++) {
-        JsonObject object = array.get(i).getAsJsonObject(); // {}
+        JsonObject object = array.get(i).getAsJsonObject();
         if (!object.has(targetStr)) {
-          return "[]";
+          continue;
```
SteveYurongSu, line 110: It's different from `return "[]"`, right？

SteveYurongSu: Ok, I understand now :D 

Forgot to see your JIRA ticket before reviewing it.

grafana/src/main/java/org/apache/iotdb/web/grafana/controller/DatabaseConnectController.java
xinzhongtianxia: In method body, we have not used the `request` param and the response's default code is 200, so we do not need these two params.

SteveYurongSu (approved):
LGTM.

Thanks!

PR #4206, opened 2021-10-21
[To rel/0.12] [IOTDB-1872 IOTDB-1902 IOTDB-1903] Fix data increases abnormally after IoTDB restarts

THUMarkLau:


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -695,6 +687,10 @@ private void recoverTsFiles(List<TsFileResource> tsFiles, boolean isSeq) throws
         logger.warn(
             "Skip TsFile: {} because of error in recover: ", tsFileResource.getTsFilePath(), e);
         continue;
+      } finally {
+        if (writer == null) {
```
qiaojialin, line 691:
```suggestion
        if (writer != null) {
```

THUMarkLau: resolve

tsfile/src/main/java/org/apache/iotdb/tsfile/read/TsFileSequenceReader.java
```diff
@@ -1142,6 +1145,9 @@ public long selfCheck(
           this.position(),
           e.getMessage());
     }
+    if (loadLastChunkMetadata) {
+      chunkGroupMetadataList.add(new ChunkGroupMetadata(lastDeviceId, chunkMetadataList));
+    }
```
JackieTien97, line 1150: What if the last ChunkGroupMetadata has already been added before.

JackieTien97 (changes requested):

tsfile/src/main/java/org/apache/iotdb/tsfile/read/TsFileSequenceReader.java
THUMarkLau: resolve

PR #4134, opened 2021-10-12
[IOTDB-1280] Rewrite the Antlr grammar file

liuminghui233:
The current ANTLR g4 file is not elegant.

For more details, please refer to: https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=191334171

PR #4234, opened 2021-10-26
[IOTDB-1280] Solve the performance reduction caused by syntax error in SLL prediction mode

liuminghui233:
After merge #4134, the parsing performance of SQL has dropped significantly. 

> ANTLR can use two different prediction modes: SLL and LL*. They are also known as parsing strategies. Basically, these dictate how much effort the parser will put to parse the input. The first one is simpler and quicker, the second one is more powerful but a bit slower. By more powerful we mean that it can parse more complicated inputs that would result as ambiguous using just the first strategy.

Due to some modifications in #4134, SLL cannot parse the grammar correctly. The parser had to switch to LL mode, wasting a lot of performance.


liuminghui233:
@SteveYurongSu Hi, I did some experiments. The performance of SQL parsing is approximately unchanged (perhaps slightly improved) compared with before rewrite.

cornmonster:
@liuminghui233 Hi, I have some changes on the ANTLR part. And I am wondering about the way you find the causes of the performance issue. Could you explain a little bit? 

Thanks!

liuminghui233:
> @liuminghui233 Hi, I have some changes on the ANTLR part. And I am wondering about the way you find the causes of the performance issue. Could you explain a little bit?
> 
> Thanks!

Hi, you can refer to this [blog](https://tomassetti.me/improving-the-performance-of-an-antlr-parser) to learn more about the performance of Antlr4. Before merging your PR, you can conduct tests on the parsing performance of some critical statements (such as insert and query statements). Good luck!

PR #4244, opened 2021-10-27
[IOTDB-1745] Metagroup snapshots support template

stormbroken:
Snapshots of the meta group in the current cluster module support storage groups and TTL information, etc, but do not support templates. Templates need to be supported so that newly added nodes can register templates during cluster expansion.

I add test in cluster/log/snapshot/MetaSimpleSnapshotTest(testSerialize and testInstall) and cluster/server/member/MetaGroupMemberTest(testSendSnapshot)


cluster/src/test/java/org/apache/iotdb/cluster/log/snapshot/MetaSimpleSnapshotTest.java
```diff
@@ -44,8 +49,7 @@
 import java.util.HashMap;
 import java.util.Map;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
```
neuyilan, line 52: In general, we don't need to reference packages using .*

neuyilan (approved):
LGTM, only minor issue.

cluster/src/test/java/org/apache/iotdb/cluster/log/snapshot/MetaSimpleSnapshotTest.java
stormbroken: I fix this problems

cluster/src/main/java/org/apache/iotdb/cluster/log/snapshot/MetaSimpleSnapshot.java
```diff
@@ -115,6 +125,13 @@ public ByteBuffer serialize() {
         dataOutputStream.write(entry.getValue().serialize().array());
       }
 
+      dataOutputStream.writeInt(templateMap.size());
+      for (Map.Entry<String, Template> entry : templateMap.entrySet()) {
+        SerializeUtils.serialize(entry.getKey(), dataOutputStream);
+        logger.info("A template into snapshot: {}", entry.getValue());
```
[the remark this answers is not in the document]
stormbroken: I fix them all

PR #4257, opened 2021-10-28
[IOTDB-1886] Feature/const expr

cornmonster:
This feature introduces the following backward incompatibilities:

1. Align by device queries no longer support constant measurements
2. Node name cannot start with 0-9
3. Arithmetic operators like +-*/% are no longer allowed in a node name


SteveYurongSu:
Great to see this fantastic feature implemented.

Please resolve conflicts and add some user docs. :D

server/src/main/java/org/apache/iotdb/db/query/udf/core/transformer/ArithmeticBinaryTransformer.java
```diff
@@ -36,6 +36,11 @@ protected ArithmeticBinaryTransformer(
     this.rightPointReader = rightPointReader;
   }
 
+  @Override
+  public boolean isConstantPointReader() {
+    return leftPointReader.isConstantPointReader() && rightPointReader.isConstantPointReader();
```
SteveYurongSu, line 41: It's better to cache the result of isConstantPointReader as a class member of Expression. Otherwise, a cachetime() call will cause multiple chain isConstantPointReader() calls.

server/src/main/java/org/apache/iotdb/db/query/udf/core/reader/ConstantLayerPointReader.java
```diff
@@ -0,0 +1,117 @@
...
+
+  @Override
+  public long currentTime() throws IOException {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override
+  public int currentInt() throws IOException {
+    if (TSDataType.INT32.equals(expression.getDataType())) {
+      return (int) expression.getValue();
```
SteveYurongSu, line 67: It's better to avoid the unboxing process, which will cause heavy performance issues. You can refer to org/apache/iotdb/db/query/udf/core/transformer/Transformer.java to find the solution. :D

server/src/main/java/org/apache/iotdb/db/query/expression/unary/ConstantExpression.java
```diff
@@ -0,0 +1,111 @@
...
+
+import java.time.ZoneId;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/** Constant expression */
+public class ConstantExpression extends Expression {
+
+  private final Object value;
```
SteveYurongSu, line 44: It's better to avoid the unboxing process, which will cause heavy performance issues. You can refer to org/apache/iotdb/db/query/udf/core/transformer/Transformer.java to find the solution. :D

server/src/main/java/org/apache/iotdb/db/query/expression/unary/ConstantExpression.java
```diff
@@ -0,0 +1,111 @@
...
+      UDTFPlan udtfPlan,
+      RawQueryInputLayer rawTimeSeriesInputLayer,
+      Map<Expression, IntermediateLayer> expressionIntermediateLayerMap,
+      Map<Expression, TSDataType> expressionDataTypeMap,
+      LayerMemoryAssigner memoryAssigner) {
+    expressionDataTypeMap.put(this, this.getDataType());
+    IntermediateLayer intermediateLayer =
+        new ConstantIntermediateLayer(this, queryId, memoryAssigner.assign());
+    expressionIntermediateLayerMap.put(this, intermediateLayer);
+    return intermediateLayer;
```
SteveYurongSu, line 104: For 

select sin(1 + s1), cos(1 + s1) from root.sg.d

we'd better not construct the IntermediateLayer twice.

server/src/main/java/org/apache/iotdb/db/query/expression/unary/ConstantExpression.java
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  @Override
+  public boolean isPureConstantExpression() {
+    return true;
+  }
+
+  @Override
+  public void concat(List<PartialPath> prefixPaths, List<Expression> resultExpressions) {
+    resultExpressions.add(this);
```
SteveYurongSu, line 67: This will not handle the following case:

select sin(1 + s1) from root.sg.d1, root.sg.d2

server/src/main/java/org/apache/iotdb/db/query/expression/unary/TimeSeriesOperand.java
```diff
@@ -104,7 +109,7 @@ public IntermediateLayer constructIntermediateLayer(
 
       expressionIntermediateLayerMap.put(
           this,
-          memoryAssigner.getReference(this) == 1
+          memoryAssigner.getReference(this) == 1 || isPureConstantExpression()
```
SteveYurongSu, line 112: Why we construct a SingleInputColumnSingleReferenceIntermediateLayer when isPureConstantExpression() is true?

SteveYurongSu (commented):
Very good try, very good implementation!

server/src/test/java/org/apache/iotdb/db/integration/IoTDBAlignByDeviceIT.java
```diff
@@ -788,79 +788,6 @@ public void unusualCaseTest2() throws ClassNotFoundException {
     }
   }
 
-  @Test
-  public void selectConstantTest() throws ClassNotFoundException {
-    String[] retArray =
-        new String[] {
-          "1,root.vehicle.d0,101,1101,null,null,null,'11',",
```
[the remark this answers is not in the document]
cornmonster: Sure.

server/src/main/java/org/apache/iotdb/db/query/expression/unary/ConstantExpression.java
cornmonster: Make sense. I will have the constant intermediate cached in the constant operand so that the unboxing happens only once for each different constant operand.

server/src/main/java/org/apache/iotdb/db/query/udf/core/reader/ConstantLayerPointReader.java
cornmonster: Good point. I will use the same way as the Transformer to avoid the performance issue.

server/src/main/java/org/apache/iotdb/db/query/udf/core/transformer/ArithmeticBinaryTransformer.java
cornmonster: Make sense. Will cache the result.

server/src/main/java/org/apache/iotdb/db/query/expression/unary/ConstantExpression.java
```diff
@@ -0,0 +1,111 @@
...
+import org.apache.iotdb.db.utils.CommonUtils;
+import org.apache.iotdb.tsfile.file.metadata.enums.TSDataType;
+
+import java.time.ZoneId;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/** Constant expression */
+public class ConstantExpression extends Expression {
```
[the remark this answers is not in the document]
cornmonster: Will change to ConstantOperand

server/src/main/java/org/apache/iotdb/db/query/expression/unary/ConstantExpression.java
cornmonster: Will cache the intermediate layer for the same const expression.

server/src/main/java/org/apache/iotdb/db/query/expression/unary/ConstantExpression.java
cornmonster: I believe the path concat is handled in TimeSeriesOperand, right?

```
create timeseries root.sg1.d1.a with datatype=int32
create timeseries root.sg1.d1.b with datatype=int32
create timeseries root.sg1.d2.a with datatype=int32
create timeseries root.sg1.d2.b with datatype=int32
insert into root.sg1.d1(time, a) values(0, 1)
insert into root.sg1.d1(time, b) values(0, 2)
insert into root.sg1.d2(time, a) values(0, 3)
insert into root.sg1.d2(time, b) values(0, 4)

select a + b - 1 from root.sg1.d1, root.sg1.d2

result:
+-----------------------------+-----------------------------------+-----------------------------------+-----------------------------------+-----------------------------------+
|                         Time|(root.sg1.d1.a + root.sg1.d1.b) - 1|(root.sg1.d1.a + root.sg1.d2.b) - 1|(root.sg1.d2.a + root.sg1.d1.b) - 1|(root.sg1.d2.a + root.sg1.d2.b) - 1|
+-----------------------------+-----------------------------------+-----------------------------------+-----------------------------------+-----------------------------------+
|1970-01-01T08:00:00.000+08:00|                                2.0|                                4.0|                                4.0|                                6.0|
+-----------------------------+-----------------------------------+-----------------------------------+-----------------------------------+-----------------------------------+

```

server/src/main/java/org/apache/iotdb/db/query/expression/unary/TimeSeriesOperand.java
cornmonster: That's a little bit tricky.

Since the constant point reader cannot provide a reasonable timestamp, using ElasticSerializableTVList with a constant point reader causes troubles. 

Also, I believe using a multi reference intermediate layer is mainly for avoiding duplicated IO and memory control, but unlike a time series reader, the constant reader won't cause any IO and the memory usage is little. 

Thus, I think SingleInputColumnSingleReferenceIntermediateLayer is a proper choice. What do you think?

server/src/main/java/org/apache/iotdb/db/query/dataset/UDTFAlignByTimeDataSet.java
```diff
@@ -80,6 +80,9 @@ public UDTFAlignByTimeDataSet(
   protected void initTimeHeap() throws IOException, QueryProcessException {
     timeHeap = new TimeSelector(transformers.length << 1, true);
     for (LayerPointReader reader : transformers) {
+      if (reader.isConstantPointReader()) {
+        continue;
+      }
```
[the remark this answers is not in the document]
cornmonster: Will remove it.

cornmonster:
@SteveYurongSu Hi, PTAL

SteveYurongSu (changes requested):

antlr/src/main/antlr4/org/apache/iotdb/db/qp/sql/IoTDBSqlLexer.g4
```diff
@@ -915,6 +915,19 @@ fragment NAME_CHAR
     | CN_CHAR
     ;
 
+fragment FIRST_NAME_CHAR
```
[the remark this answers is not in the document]
cornmonster: No problem. PTAL.

server/src/main/java/org/apache/iotdb/db/query/expression/unary/ConstantExpression.java
SteveYurongSu: You are right :D

SteveYurongSu (approved):
I have to say this is a fantastic job! Thanks for your contributions!!!

PR #4278, opened 2021-10-29
set backlog of transport server

cigarl:
default backlog size is too small, that might occur `Cannot send heartbeat to node xxx  due to network`


neuyilan (commented):
The heartbeat thread will send hearts to the followers once per second, the backlog default is 50, from this point, I think the default value is enough.
Maybe the root cause of the error you pointed out has not been found yet?

cigarl:
> The heartbeat thread will send hearts to the followers once per second, the backlog default is 50, from this point, I think the default value is enough. Maybe the root cause of the error you pointed out has not been found yet?

Yes, I'm still looking for the root cause. This is an evasive operation on my own environment. BTW, port 6667 May have this risk, and we need to treat each port differently.

This PR is a draft, I will close it later and create a new PR when the issue is resolved. Thanks.

PR #4279, opened 2021-10-30
[IOTDB-1609] Print the actual size of plan when it exceeds the wal_buffer_size

Cpaulyz:
## Contributions:
* Print the actual size of plan when it exceeds the wal_buffer_size
  * Use `PhysicalPlan.serialize(DataOutputStream stream)`  to get actual size of plan after serialization. 
  * Print infomation in IOException message
* Add some test in `testOverSizedWAL` to verify the actual size of plan
  * If wal_buffer_size is less than actual size, it would throw IOException.
  * If wal_buffer_size is more than or equal to actual size, it would not throw IOException.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/writelog/node/ExclusiveWriteLogNode.java
+++ b/server/src/main/java/org/apache/iotdb/db/writelog/node/ExclusiveWriteLogNode.java
@@ -33,9 +33,7 @@ import org.apache.commons.io.FileUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.File;
-import java.io.FileNotFoundException;
-import java.io.IOException;
+import java.io.*;
 import java.nio.BufferOverflowException;
 import java.nio.ByteBuffer;
 import java.nio.channels.ClosedChannelException;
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/apache/iotdb/db/writelog/node/ExclusiveWriteLogNode.java
```diff
@@ -115,8 +113,15 @@ public void write(PhysicalPlan plan) throws IOException {
     } catch (BufferOverflowException e) {
       // if the size of a single plan bigger than logBufferWorking
       // we need to clear the buffer to drop something wrong that has written.
+      DataOutputStream dos = new DataOutputStream(new ByteArrayOutputStream());
+      plan.serialize(dos);
+      int neededSize = dos.size();
+      dos.close();
```
JackieTien97, line 119: It seems that there is no need to really do the serialization to get the actual size, maybe you can add a new method like `getSerializedSize()` in `PhysicalPlan`, each actual plan class to override that method to calculate its own size. In this way, you can avoid allocate extra memory in this `BufferOverflowException` situation.

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/writelog/node/ExclusiveWriteLogNode.java
Cpaulyz: > It seems that there is no need to really do the serialization to get the actual size, maybe you can add a new method like `getSerializedSize()` in `PhysicalPlan`, each actual plan class to override that method to calculate its own size. In this way, you can avoid allocate extra memory in this `BufferOverflowException` situation.

Yes, I totally agree with what you said. In fact, I wanted to do the same thing at first. 

However, I noticed that there are quite a lot of classes extends `PhysicalPlan`. It needs to modify code massively. Considering that `getSerializedSize()` is only used for this exception and exception may be occurred rarely, I allocate extra memory to get the actual size. 

Do you think it's necessary to do what you said? If so, I may need more time to implement `getSerializedSize()` in each actual plan class.

PR #4285, opened 2021-11-01
[To rel/0.12] Print the file path while meeting error in case of reading chunk

JackieTien97:
Print the file path while meeting error in case of reading chunk


HTHou (approved):

liuzhen1207:
0.12.3 rc2  / commit a8c5dc8a13a4b90303fec4b29e40a1b33e6b737d
PASS

image

PR #4298, opened 2021-11-02
Add compaction version in cache key

JackieTien97:
In current IoTDB, after compaction finished, it will clear all the query cache, because the old file and the new file share the same key in query cache and there is no way to remove the exact old file entry in query cache.

However, we can add the compaction version into the cache entry key so that the new file generated by compaction will has a new entry instead of sharing the same key as the old file. In this way, we can avoid clearing all query cache after each compaction.

### DIFF
--- a/tsfile/src/main/java/org/apache/iotdb/tsfile/utils/FilePathUtils.java
+++ b/tsfile/src/main/java/org/apache/iotdb/tsfile/utils/FilePathUtils.java
@@ -19,14 +19,12 @@
 
 package org.apache.iotdb.tsfile.utils;
 
+import static org.apache.iotdb.tsfile.common.constant.TsFileConstant.TSFILE_SUFFIX;
+
+import java.io.File;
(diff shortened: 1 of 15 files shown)


HTHou:
Hi, you can merge the master branch to fix the CI now.

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/FilePathUtils.java
```diff
@@ -103,6 +111,23 @@ public static long splitAndGetTsFileVersion(String tsFileName) {
     return Long.parseLong(names[1]);
   }
 
+  /**
+   * @return a long array whose length is 2, the first long value is tsfile version, second long
+   *     value is compaction version, high 32 bit is compaction level, low 32 bit is merge count
```
qiaojialin, line 116:
```suggestion
   *     value is compaction version, high 32 bit is in-space compaction count, low 32 bit is cross-space compaction count
```

qiaojialin (approved):

PR #4309, opened 2021-11-03
Fix the bug that election in term n may increase to n+1 before send e…

chengjianyun:
The log shows behaviors like below in meta group:

1. Node 09 run as follower in term 3.
2. Node 09 heartbeat timeout in term 3.
3. Node 09 update its term to 4
4. Node 09 update `voteFor` to null
5. Node 09 receive an election request from Node 08 in term 4, accept the election, become follower of the group in term 4.
6. Node 09 continue to election. but term is 5.

It's not right in the step 6 as Node 09 timed out in term 3, it should only start the election in term 4 instead of term 5. I think the behavior isn't by design although it didn't run into some wrong status.

[Attach log: meta.log](https://github.com/apache/iotdb/files/7466363/meta.log)


chengjianyun:
Need more thinking, close temporary.

PR #4334, opened 2021-11-08
[IOTDB-1613] Recover mods file if a delete write modification failed

HeimingZ:
When we write a Deletion to a modification file, the Deletion may crash because of power-off. So, the modifications after this broken Deletion will be incorrect, so we need to truncate them to make the mods file correct.

To trace the read position, i implement TracedBufferedReader based on java.io.BufferedReader by modifying the fill() method.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/modification/utils/TracedBufferedReader.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/modification/utils/TracedBufferedReader.java
@@ -1,0 +1,393 @@
+package org.apache.iotdb.db.engine.modification.utils;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.Reader;
+import java.io.UncheckedIOException;
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import java.util.Spliterator;
+import java.util.Spliterators;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+/**
+ * Copied from {@link java.io.BufferedReader}, trace the position by modifying the fill() method.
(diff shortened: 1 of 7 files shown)


server/src/main/java/org/apache/iotdb/db/engine/modification/utils/TracedBufferedReader.java
```diff
@@ -0,0 +1,412 @@
...
+  }
+
+  /**
+   * Returns this reader's file position.
+   *
+   * @return This reader's file position, a non-negative integer counting the number of bytes from
+   *     the beginning of the file to the current position
+   */
+  public long position() {
+    return position - nChars + nextChar;
```
HTHou, line 410: Add some comments here？

server/src/main/java/org/apache/iotdb/db/engine/modification/utils/TracedBufferedReader.java
```diff
@@ -0,0 +1,412 @@
...
+  /** If the next character is a line feed, skip it */
+  private boolean skipLF = false;
+
+  /** The skipLF flag when the mark was set */
+  private boolean markedSkipLF = false;
+
+  private static int defaultCharBufferSize = 8192;
+  private static int defaultExpectedLineLength = 80;
+
+  private long position = 0;
```
HTHou, line 56: Add comments

HeimingZ: Already added.

qiaojialin (approved):

HTHou (approved):

PR #4336, opened 2021-11-08
[To rel/0.12][IOTDB-1958] Add storage group not ready exception

HeimingZ:
There are too many logs like "sg may not ready now ", so i add StorageGroupNotReadyException and exclude it from server log.


server/src/main/java/org/apache/iotdb/db/service/TSServiceImpl.java
```diff
@@ -1944,10 +1943,12 @@ private TSStatus tryCatchNonQueryException(Exception e) {
       LOGGER.warn(message, e);
       return RpcUtils.getStatus(Arrays.asList(((BatchProcessException) e).getFailingStatus()));
     } else if (e instanceof IoTDBException) {
-      if (((IoTDBException) e).isUserException()) {
-        LOGGER.warn(message + e.getMessage());
-      } else {
-        LOGGER.warn(message, e);
+      if (!(e instanceof StorageGroupNotReadyException)) {
```
HTHou, line 1946: } else if (e instanceof IoTDBException && !(e instanceof StorageGroupNotReadyException)) {

HeimingZ: fixed.

HTHou (approved):
LGTM

PR #4347, opened 2021-11-09
[To rel/0.12][IOTDB-1985]Fix Sync Error in parse sg path between different os

yschengzi:
sync sender use "_" to replace File.separator to make sync between different os correct.
such as:
    sender(windows): root.sg\0\0\1.tsfile -> root.sg_0_0_1.tsfile
    receiver(linux): root.sg_0_0_1.tsfile -> root.sg/0/0/1.tsfile

but some sg name contains "_", so the parse for receiver is not correct.
such as:
    sg name: root.test.g_7
    receiver parse: root.test.g_7_0_0_1.tsfile -> root.test.g/7/0/0/1.tsfile
So the receiver can not got correct sg name from folder path

now, i change the chart to replace File.separator to "!", a chart that is not exist in sg's name and a tsfile'name according to anltr4.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/sync/conf/SyncConstant.java
+++ b/server/src/main/java/org/apache/iotdb/db/sync/conf/SyncConstant.java
@@ -36,6 +36,8 @@ public class SyncConstant {
 
   public static final String SYNC_DIR_NAME_SEPARATOR = "_";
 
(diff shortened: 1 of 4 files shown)


yschengzi:
add UT to test encode for fileInfo in initSyncData() and syncDeletedFileName()

HTHou (approved):

PR #4354, opened 2021-11-10
[IOTDB-1986] Support `select UDF as alisa` clauses

jun0315:
see https://issues.apache.org/jira/browse/IOTDB-1986


SteveYurongSu (approved):
LGTM

PR #4358, opened 2021-11-11
[IOTDB-1990] Fix uncheck null result by calling IReaderByTimestamp

ericpai:
See JIRA: https://issues.apache.org/jira/browse/IOTDB-1990

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/query/dataset/RawQueryDataSetWithValueFilter.java
+++ b/server/src/main/java/org/apache/iotdb/db/query/dataset/RawQueryDataSetWithValueFilter.java
@@ -118,7 +118,7 @@ public class RawQueryDataSetWithValueFilter extends QueryDataSet implements UDFI
 
       // 3. use values in results to fill row record
       for (int j = 0; j < cachedTimeCnt; j++) {
-        if (results[j] == null) {
+        if (results == null || results[j] == null) {
           rowRecords[j].addField(null);
         } else {
           hasField[j] = true;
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/apache/iotdb/db/query/dataset/RawQueryDataSetWithValueFilter.java
```diff
@@ -118,7 +118,7 @@ private boolean cacheRowRecords() throws IOException {
 
       // 3. use values in results to fill row record
       for (int j = 0; j < cachedTimeCnt; j++) {
-        if (results[j] == null) {
+        if (results == null || results[j] == null) {
```
JackieTien97, line 121: It seems that only when `cachedTimeCnt` equals to zero, `getValuesInTimestamps` will return null. However, if `cachedTimeCnt` equals to zero, we will `return false` before here.

server/src/main/java/org/apache/iotdb/db/query/dataset/RawQueryDataSetWithValueFilter.java
```diff
@@ -199,7 +199,7 @@ private boolean cacheRowInObjects() throws IOException {
 
       // 3. use values in results to fill row record
       for (int j = 0; j < cachedTimeCnt; j++) {
-        if (results[j] != null) {
+        if (results != null && results[j] != null) {
```
JackieTien97, line 202: Same as above.

JackieTien97 (changes requested):
Could you provide an IT for this kind of case

ericpai:
@JackieTien97 I replayed this test case `IoTDBUDTFAlignByTimeQueryIT.queryWithValueFilter9` in a cluster with 3 nodes and 2 replicas. It can reappear in 100%. This is the debug screenshot. I guess the reason is that although the `cachedTimeCnt` is not empty, there's no data matched the `cachedTimeCnt` and a null result returns. This only happens in cluster query.

image

PR #4357, opened 2021-11-11
[IOTDB-1983] Fix DescReadWriteBatchData serializing bug

ericpai:
See JIRA: https://issues.apache.org/jira/browse/IOTDB-1983

Fix bugs:
1. Incorrect order of serializing `DescReadWriteBatchData`.
2. Unsupport **VECTOR** type of `DescReadWriteBatchData`.
3. Add unit test cases to cover the `BatchData` and its children classes.
4. Add some comments to make the `getXXXByIndex()` interfaces more clear

PR #4361, opened 2021-11-11
[To rel/0.12][IOTDB-1983] Fix DescReadWriteBatchData serializing bug

ericpai:
A cherry-pick of #4357 
JIRA: https://issues.apache.org/jira/browse/IOTDB-1983

As 0.12 doesn't support Vector type, I remove those codes.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,4 +1,3 @@
-
 <!--
 
     Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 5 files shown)


ericpai:
@JackieTien97 I think this one should be fixed to 0.12 as well.

JackieTien97 (changes requested):
LGTM, you can merge the latest rel/0.12, then CI will pass.

PR #4369, opened 2021-11-12
[rel/0.12][IOTDB-1996] Decrease the default max-size of direct memory

xiaoyekanren:
Decrease the default max-size of direct memory to one third of MADecrease the default max-size of direct memory to one third of MAX_HEAP_SIZEX_HEAP_SIZE


FLOW4215:
MAX_DIRECT_MEMORY_SIZE="2G"直接确定在2G？不在按照服务器资源计算？

PR #4380, opened 2021-11-14
Fixed flaky tests

alany9552:
## Description
Fixed a flaky test causes by non-deterministic concurrent map

### Reason of change
Folllowing tests are flaky and failing:
```
org.apache.iotdb.cluster.log.snapshot.PullSnapshotTaskTest#testWithRetry
org.apache.iotdb.cluster.log.snapshot.MetaSimpleSnapshotTest#testSerialize
org.apache.iotdb.cluster.query.reader.mult.AssignPathManagedMergeReaderTest#testMultManagerMergeRemoteSeriesReader
org.apache.iotdb.cluster.query.reader.mult.RemoteMultSeriesReaderTest#testSyncMultSeriesReader 
org.apache.iotdb.cluster.query.reader.mult.RemoteMultSeriesReaderTest#testAsyncMultSeriesReader
```
when running 
`mvn -pl cluster edu.illinois:nondex-maven-plugin:1.1.2:nondex -Dtest=*TESTNAME*
`
The tests are flaky because in `serialize()` method, it uses a concurrent map to store `nodeSlot` which returns
non deterministic results for each run. To fix this, I changed the concurrent map to `Collections.synchronizedMap(new LinkedHashMap())` to ensure the orders and the threads' safety instead of just using `LinkedHashMap()`.

<hr>

This PR has:
- [X] been self-reviewed.
    - [X] concurrent read
    - [X] concurrent write
    - [X] concurrent read and write 
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/partition/slot/SlotPartitionTable.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/partition/slot/SlotPartitionTable.java
@@ -48,6 +48,7 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -76,7 +77,8 @@ public class SlotPartitionTable implements PartitionTable {
 
   // The following fields are used for determining which node a data item belongs to.
   // the slots held by each node
-  private Map<RaftNode, List<Integer>> nodeSlotMap = new ConcurrentHashMap<>();
+  private Map<RaftNode, List<Integer>> nodeSlotMap =
+      Collections.synchronizedMap(new LinkedHashMap());
   // each slot is managed by whom
   private RaftNode[] slotNodes = new RaftNode[ClusterConstant.SLOT_NUM];
   // the nodes that each slot belongs to before a new node is added, used for the new node to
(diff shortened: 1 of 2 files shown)


neuyilan (changes requested):
Thanks for your job.
However, I can not get your point, are you going to keep the order of map elements? 
if we use `Collections.synchronizedMap` relpace the `ConcurrentHashMap`, which may occur `ConcurrentModificationException`.

alany9552:
> Thanks for your job. However, I can not get your point, are you going to keep the order of map elements? if we use `Collections.synchronizedMap` relpace the `ConcurrentHashMap`, which may occur `ConcurrentModificationException`.

Yeah I understand your concerns. The order of the map elements keeps the same and pass for the regular runs, but the order can't be guaranteed when running `NonDex` tests because `ConcurrentHashMap`. If using `ConcurrentHashMap` is the best in this situation as you mentioned the exception, it would be better to change the test code in that doesn't require orders, otherwise the tests will be keep flaky when running.

alany9552:
> Thanks for your job. However, I can not get your point, are you going to keep the order of map elements? if we use `Collections.synchronizedMap` relpace the `ConcurrentHashMap`, which may occur `ConcurrentModificationException`.

Originally I was thinking about using just `LinkedHashMap` but it not thread safe, so that's why I choose to use `Collections.synchronizedMap`.

PR #4384, opened 2021-11-15
[IOTDB-2008] Fix null result bug of GROUP BY FILL query

ericpai:
See JIRA: https://issues.apache.org/jira/browse/IOTDB-2008

Because the `GroupByFillDataSet` has been refined totally, I don't know whether BUG1 described in JIRA has been fixed.

If so, I will open a new PR to fix it in rel/0.12 branch.

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterQueryRouter.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterQueryRouter.java
@@ -21,6 +21,7 @@ package org.apache.iotdb.cluster.query;
 
 import org.apache.iotdb.cluster.query.aggregate.ClusterAggregateExecutor;
 import org.apache.iotdb.cluster.query.fill.ClusterFillExecutor;
+import org.apache.iotdb.cluster.query.groupby.ClusterGroupByFillNoVFilterDataSet;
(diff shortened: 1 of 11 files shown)


ericpai:
This PR has fixed the following issues:
1. Add `ClusterGroupByFillNoVFilterDataSet` to handle GROUP BY FILL query in cluster mode. As the query with value filter is not supported yet, the implementation is left to furture developers. Please do not forget :)
2. Refining the constructor of `GroupByWithValueFilterDataSet` and its subclasses, by separating `initGroupBy` to an individual method which is needed to call explicitly. As in `initGroupBy`, we should init underlying readers of timeseries. In cluster mode, we must use a new parameter `MetaGroupMember` to init `readerFactory`, which passed in the constructor. However, the `ClusterGroupByFillNoVFilterDataSet` extends `GroupByWithValueFilterDataSet` and the `super()` call must be the first line in the constructor, and we have no opportunity to set `MetaGroupMember` as a class field.
3. Fix a null result issue which casing thrift execption thrown when calling peekNextNotNullValue().

@JackieTien97 @CRZbulabula Thanks for reviewing the codes.

JackieTien97 (approved):
@CRZbulabula As in the new group by fill pr, `GroupByFillWithoutValueFilter` doesn't extends `GroupByWithoutValueFilter`, remember to move the `initGroupBy()` method back to the constructor of these classes. It's better to keep the init codes in the constructor.

PR #4388, opened 2021-11-15
[IOTDB-2010] fix incomplete show timeseries result

cornmonster:
See: https://issues.apache.org/jira/browse/IOTDB-2010

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/server/handlers/caller/ShowTimeSeriesHandler.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/server/handlers/caller/ShowTimeSeriesHandler.java
@@ -29,14 +29,33 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 /** Handler for getting the schemas from each data group concurrently. */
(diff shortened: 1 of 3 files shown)


cornmonster:
@JackieTien97 Hi, could u please help review this PR? Thanks!

cornmonster:
@LebronAl Hi, could u please help review this PR? Thanks!

cluster/src/main/java/org/apache/iotdb/cluster/server/handlers/caller/ShowTimeSeriesHandler.java
```diff
@@ -0,0 +1,133 @@
...
+
+  private static final Logger logger = LoggerFactory.getLogger(ShowTimeSeriesHandler.class);
+
+  /** String representation of a partial path for logging */
+  private final String path;
+
+  private final CountDownLatch countDownLatch;
+  private final long startTimeInMs;
+
+  private final Map<String, ShowTimeSeriesResult> timeSeriesNameToResult = new HashMap<>();
```
neuyilan, line 66: We should use a thread-safe map, as this may be used by multi threads when called `onComplete()` method.

neuyilan (changes requested):

cluster/src/main/java/org/apache/iotdb/cluster/server/handlers/caller/ShowTimeSeriesHandler.java
cornmonster: `onComplete` and `onError` are decorated with keyword `synchronized`. Thus this handler should be thread-safe.

neuyilan: Sorry for didn't pay attention to the method signature

neuyilan (approved):

LebronAl (approved):
LGTM

PR #4409, opened 2021-11-16
[To rel/0.12] [IOTDB-2018] Query dataset memory leak on server caused by python client

noorall:
I made the following changes in this PR:

 1. fix leak caused by python client: statementId not set in TSCloseOperationReq

JIRA: https://issues.apache.org/jira/browse/IOTDB-2018


SteveYurongSu (approved):
LGTM

PR #4414, opened 2021-11-17
[IOTDB-2023] Fix serializing and deserializing bugs of Filters

ericpai:
See JIRA: https://issues.apache.org/jira/browse/IOTDB-2023

The following bugs have been fixed:
1. `GroupByMonthFilter`: It inherits `GroupByMonth` filter but doesn't override `getSerializeId()`, which returns `FilterSerializeId.GROUP_BY` and causes the deserializing instance being `GroupByFilter`.
2. `In`: Not supported in `FilterFactory`.
3. `In`: In the serializing and  deserializing method, we should use `ReadWriteUtils.write()` and `ReadWriteUtils.readInt()`. As `buffer.get()` and `stream.write()` handle only 1 byte.


ericpai:
@JackieTien97 Thanks for the review

PR #4425, opened 2021-11-18
[IOTDB-2016][metric] Fix ClassCastException when create new metric types

xinzhongtianxia:
## Description
1. Create Gauge with metric name = 'a' and tags = ['b' = 'c']
2. Create Timer with metric name = 'a' and tags = ['b' = 'c']
3. Then we will get a ClassCastException

I fixed it like dropwizard or micrometer does:
1. If IMetric exist in the map, check if it instanceof the right metric type.
2. If not, throw an IllegalArgumentException to test user the name and tags are already used for a different type of metric

jira:
https://issues.apache.org/jira/browse/IOTDB-2016


SteveYurongSu:
Thanks for the contribution! Please take a look at the failed tests.

@stormbroken PTAL :D

stormbroken:
Thanks for your contribution! I think you need to fix related test at the same time. Example id as follows:
image

stormbroken:
Thank you for your continued attention! From my respective,  LGTM @SteveYurongSu

PR #4431, opened 2021-11-19
[IOTDB-2031] Fix incorrect result of descending query with value filter in cluster

ericpai:
See JIRA: https://issues.apache.org/jira/browse/IOTDB-2031

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/query/reader/series/SeriesReader.java
+++ b/server/src/main/java/org/apache/iotdb/db/query/reader/series/SeriesReader.java
@@ -40,6 +40,7 @@ import org.apache.iotdb.tsfile.file.metadata.statistics.Statistics;
 import org.apache.iotdb.tsfile.read.TimeValuePair;
 import org.apache.iotdb.tsfile.read.common.BatchData;
 import org.apache.iotdb.tsfile.read.common.BatchDataFactory;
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/apache/iotdb/db/query/reader/series/SeriesReader.java
```diff
@@ -231,6 +232,10 @@ protected DescPriorityMergeReader getDescPriorityMergeReader() {
     return new DescPriorityMergeReader();
   }
 
+  public static Filter defaultTimeFilter(boolean ascending) {
+    return ascending ? TimeFilter.gtEq(Long.MIN_VALUE) : TimeFilter.ltEq(Long.MAX_VALUE);
+  }
+
```
JackieTien97, line 238: You can move the method into `TimeFilter` class and add java doc for this method to indicate that it will return a timefilter thta will always satisfy any time.

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/query/reader/series/SeriesReader.java
ericpai: Thanks for the comment, it has been fixed.

ericpai:
@JackieTien97 Hi, are there any other issues to fix?

Alima777 (approved):

PR #4439, opened 2021-11-21
[IOTDB-2006] avoid NPE while while calling the equals method of Object

dcoliversun:
## Description
NullPointerException can possibly be thrown while calling the equals method of Object maybe null.


dcoliversun:
/assign @qiaojialin  
Thanks for reviewing these code for avoiding NPE while calling the equals method of Object

qiaojialin (approved):
LGTM~

PR #4440, opened 2021-11-21
[IOTDB-2038] MqttService accesses to BasicServiceProvider

jun0315:
see : https://issues.apache.org/jira/browse/IOTDB-2038


server/src/main/java/org/apache/iotdb/db/mqtt/PublishHandler.java
```diff
@@ -64,6 +71,26 @@ public String getID() {
     return "iotdb-mqtt-broker-listener";
```
SteveYurongSu, line 71: We can add a session id suffix to the id name :D

SteveYurongSu (approved):
Well done!

SteveYurongSu (approved):

PR #4457, opened 2021-11-24
[IOTDB-1899] Fix stream closed exception during compaction

THUMarkLau:
# Description
In the process of IoTDB running, the compaction thread sometimes throws `IOException: Stream Closed`.
# Cause
Before the compaction task is executed, it will be temporarily stored in the cache queue of the CompactionTaskManager. Before joining the compaction queue, we will judge whether there are the same compaction tasks in the queue to avoid duplication. But if a task is being executed in the thread pool, at this time an identical compaction task may be allowed to be submitted to the queue (because the same task does not exist in the queue). Therefore, a compaction task may be executed at the same time.


server/src/main/java/org/apache/iotdb/db/engine/compaction/CompactionTaskManager.java
```diff
@@ -240,10 +244,18 @@ public void abortCompaction(String fullStorageGroupName) {
     }
   }
 
-  public int getTaskCount() {
+  public int getExecutingTaskCount() {
     return taskExecutionPool.getActiveCount() + taskExecutionPool.getQueue().size();
   }
 
+  public int getTotalTaskCount() {
+    return getExecutingTaskCount() + compactionTaskQueue.size();
```
qiaojialin, line 252:
```suggestion
    return getExecutingTaskCount() + candidateCompactionTaskQueue.size();
```

THUMarkLau: resolve

qiaojialin (approved):

PR #4482, opened 2021-11-26
Add annotation to compaction

choubenson:
Add some annotation to Compaction. Make it easier to read and understand code.


server/src/main/java/org/apache/iotdb/db/engine/compaction/CompactionTaskManager.java
```diff
@@ -32,26 +32,16 @@
...
-import java.util.Map;
-import java.util.Set;
-import java.util.concurrent.Callable;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Future;
-import java.util.concurrent.RejectedExecutionException;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
+import java.util.*;
```
qiaojialin, line 35: do not import *

PR #4506, opened 2021-12-01
[IoTDB-2017][metric] Add a new gauge implement

xinzhongtianxia:
## Description
I did two things in this pr
1. Add a new gauge implement
    In some instance, say, we want to monitor a list or a queue's size, it's not convenient and has low performance to set the value every time the list or queue's size changed.
    The new gauge implement is designed for this instance.
2. Add incr and decr method for gauge interface.
   In some cases, we may need to get the current value, add or decrease , then put it back. 
   Since the current `setValue` method cannot guarantee the atomicity of this operation, I added incr and decr method to the gauge interface.
    
related jira : https://issues.apache.org/jira/browse/IOTDB-2017


SteveYurongSu:
Thanks for your contribution! 

@stormbroken PTAL

stormbroken:
From my perspective, LGTM! Thanks for your contribution.
@SteveYurongSu

SteveYurongSu (approved):

PR #4514, opened 2021-12-02
[IOTDB-1971] Improve the parallelism of UDF framework execution in stand-alone mode

SteveYurongSu:
WIP.

I will give a performance report later. :D


SteveYurongSu:
@ericpai @cornmonster PTAL :D

server/src/main/java/org/apache/iotdb/db/query/udf/core/layer/LayerBuilder.java
```diff
@@ -19,18 +19,28 @@
...
+import org.apache.iotdb.tsfile.read.query.dataset.QueryDataSet;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
-public class DAGBuilder {
+public class LayerBuilder {
```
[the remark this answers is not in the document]
SteveYurongSu: Sure~

server/src/main/java/org/apache/iotdb/db/query/udf/core/layer/LayerBuilder.java
```diff
@@ -67,18 +82,30 @@ public DAGBuilder(
...
     memoryAssigner.build();
     return this;
   }
 
-  public DAGBuilder buildResultColumnPointReaders() throws QueryProcessException, IOException {
-    for (int i = 0; i < resultColumnExpressions.length; ++i) {
+  public LayerBuilder buildResultColumnPointReaders() throws QueryProcessException, IOException {
+    for (int i = 0, n = resultColumnExpressions.length; i < n; ++i) {
+      // resultColumnExpressions[i] -> the index of the fragment it belongs to
+      Integer fragmentDataSetIndex =
```
cornmonster, line 101: Maybe it is more intuitive to figure out the way of splitting the computation during the planning stage. And I believe that the parallel execution should be a framework for all the other splittable plans.

Thus, in the standalone mode, the parallel execution framework could be like this:

UDTF plan -- split --> Several sub-plans -- parallel execution (multi-threaded) --> Several datasets -- merge --> merged dataset

And in the cluster mode, the sub-plans could be distributed across the cluster (according to data locality) to achieve better performance:

UDTF plan -- split --> Several sub-plans -- parallel execution (distributed execution) --> Several datasets -- merge --> merged dataset

server/src/main/java/org/apache/iotdb/db/query/dataset/udf/UDTFFragmentDataSetTask.java
```diff
@@ -0,0 +1,84 @@
...
+public class UDTFFragmentDataSetTask extends WrappedRunnable {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(UDTFFragmentDataSetTask.class);
+
+  private final int fetchSize;
+  private final QueryDataSet queryDataSet;
+
+  // there are 3 elements in Object[].
+  // [0]: RowRecord[] or Throwable.
+  // [2]: Integer. actual length of produced row records in [0]. note that the element is -1 when
```
ericpai, line 40: It seems that this comment is mismatched with the data actually returned: [1] is an integer and [2] is a boolean.

server/src/main/java/org/apache/iotdb/db/query/dataset/udf/UDTFFragmentDataSet.java
```diff
@@ -0,0 +1,122 @@
...
+  public RowRecord nextWithoutConstraint() {
+    RowRecord rowRecord = rowRecords[++rowRecordsIndexConsumed];
+    if (rowRecordsIndexConsumed == rowRecordsLength - 1) {
+      rowRecords = null;
+    }
+    return rowRecord;
+  }
+
+  private void submitTask() {
+    if (productionBlockingQueue.remainingCapacity() > 0) {
```
ericpai, line 109: What's the consquence of running a query when `productionBlockingQueue.remainingCapacity() == 0`? It's more likely that the task will be dropped and the next time we call `tryToTakeNextRowRecords()`, it will be blocked in `productionBlockingQueue.take()` forever(`hasNextRoweRecords == true` and the previous call of `tryToTakeNextRowRecords()` is true). I think it's better to throw an exception when the queue is full.

ericpai (changes requested):

server/src/main/java/org/apache/iotdb/db/query/dataset/udf/UDTFFragmentDataSetTask.java
SteveYurongSu: Nice catch! I've fixed it.

PR #4530, opened 2021-12-07
[To rel/0.12] [IOTDB-2072] Remove TVListAllocator to reduce the TVList mem cost

HTHou:
## Description

Method 2 to reduce the TVList mem cost.

After release the TVList, the empty Object array in arrayList is not trimmed but put back to the cache of TVListAllocator. If a longer TVList reused by a shorter TVList, the memory is wasted.

Therefore, removing TVListAllocator and avoid reusing the TVList is another way to reduce TVList mem cost.


qiaojialin (approved):

HTHou:
44081638928863_ pic_hd

Removing TVListAllocator also improved the write performance. The reason is we don't need to hold the lock to apply and release TVList.

mychaow:
So, we do not reuse the tvlist, every time we allocate a new one when insert?

HTHou:
> So, we do not reuse the tvlist, every time we allocate a new one when insert?

Yes. And it seems that it doesn't have negative effect.

mychaow (approved):
LGTM

PR #4534, opened 2021-12-07
[IOTDB-2102] Push down limit to ReadTask in RawDataSetWithoutValueFilter

Alima777:
If we use limit 1 while select ** from root, the sub thread is not noticed that we only need one result, so it works until the blockingQueue is filled, which wastes lots of resource.
Therefore, we push down limit to ReadTask in RawDataSetWithoutValueFilter.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/query/dataset/RawQueryDataSetWithoutValueFilter.java
+++ b/server/src/main/java/org/apache/iotdb/db/query/dataset/RawQueryDataSetWithoutValueFilter.java
@@ -58,7 +58,7 @@ public class RawQueryDataSetWithoutValueFilter extends QueryDataSet
     private final ManagedSeriesReader reader;
     private final String pathName;
     private final BlockingQueue<BatchData> blockingQueue;
-    private List<Integer> batchDataLengthList;
+    private int[] batchDataLengthList;
     private final int seriesIndex;
     private final int fetchLimit;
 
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/apache/iotdb/db/query/dataset/RawQueryDataSetWithoutValueFilter.java
```diff
@@ -58,12 +58,23 @@
     private final ManagedSeriesReader reader;
     private final String pathName;
     private final BlockingQueue<BatchData> blockingQueue;
+    private List<Integer> batchDataLengthList;
```
JackieTien97, line 61: use `int[]` instead of `List<Integer>`

server/src/main/java/org/apache/iotdb/db/query/dataset/RawQueryDataSetWithoutValueFilter.java
```diff
@@ -530,6 +560,10 @@ protected void fillCache(int seriesIndex) throws IOException, InterruptedExcepti
       synchronized (seriesReaderList.get(seriesIndex)) {
         // we only need to judge whether to submit another task when the queue is not full
         if (blockingQueueArray[seriesIndex].remainingCapacity() > 0) {
+          if (batchDataLengthList != null
+              && batchDataLengthList.get(seriesIndex) >= rowLimit + rowOffset) {
+            return;
+          }
```
JackieTien97, line 566: You can safely delete this `if` block. It seems that we don't need to do this judgement here, because we will set `hasRemaining` to false in sub thread.

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/query/dataset/RawQueryDataSetWithoutValueFilter.java
Alima777: You are right~

PR #4539, opened 2021-12-08
[IOTDB-2124] the filtering condition does not take efffect for last query in cluster

wangchao316:
After enable_last_cache is disabled, the filter criteria are not pushed to the remote node. As a result, the query result is incorrect.

5 node , 3 rep in cluster 

 1. enable_last_cache = false

2. insert into root.sg1.d1(time,s1) values(1,1);

3. flush;

4. select last s1 from root.sg1.d1 where time > 1;

the query result is incorrect.

reason : 
the filter criteria are not pushed to the remote node. As a result, the query result is incorrect.


neuyilan (approved):
LGTM

mychaow (approved):
LGTM

PR #4545, opened 2021-12-09
[IOTDB-2116] Fix all client connections are stuck bug caused by logback bug

JackieTien97:


wangchao316 (approved):
LGTM

mychaow:
what is the bug?

wangchao316:
> what is the bug?

Frequent log printing may cause lock requisition, affecting performance. We've met before

mychaow:
> > what is the bug?
> 
> Frequent log printing may cause lock requisition, affecting performance. We've met before

Is audit log frequenct?

wangchao316:
> > > what is the bug?
> > 
> > 
> > Frequent log printing may cause lock requisition, affecting performance. We've met before
> 
> Is audit log frequenct?

yes

HTHou (approved):

mychaow:
> > > > what is the bug?
> > > 
> > > 
> > > Frequent log printing may cause lock requisition, affecting performance. We've met before
> > 
> > 
> > Is audit log frequenct?
> 
> yes

I have searched the code, and found that the debug audit log is frequent, but info audit log is not frequent, only delete and create meta is log. So, this bug is info log or debug log?

qiaojialin (approved):

PR #4550, opened 2021-12-09
[IOTDB-2131] Change the fill syntax

CRZbulabula:


JackieTien97 (changes requested):
change the doc meanwhile

PR #4547, opened 2021-12-09
[IOTDB-1969] Remove old metric framework from iotdb-server

stormbroken:
I remove old metric framework from iotdb-server, and all parameters about metric are in iotdb-metric.yml

PR #4551, opened 2021-12-10
[IOTDB-1969] Remove statistic resources of old metric framework from iotdb-server 

stormbroken:
Remove statistic resources of old metric framework from iotdb-server


SteveYurongSu:
#4547

PR #4556, opened 2021-12-11
[IOTDB-1970] Migrate the org.apache.iotdb.db.cost.statistic module to the new metric framework

stormbroken:
Migrate the org.apache.iotdb.db.cost.statistic module to the new metric framework


stormbroken:
If you want to use query stat, please set `enable_performance_stat=true` in `iotdb-engine.properties` and set `enableMetric: false` in `iotdb-metric.yml`
I use iotdb-benchmark to have a try, and set `OPERATION_PROPORTION=1:1:1:1:1:1:1:1:1:1:1`. take jmx for example, some related pictures are as follows.

count of query:

image

count of insert:
image


histogram of latency of query:

image

histogram of latency of insert:

image

SteveYurongSu (approved):

PR #4564, opened 2021-12-13
[IOTDB-2122] MQTT / REST module migrate to the correct path

jun0315:
see https://issues.apache.org/jira/browse/IOTDB-2122


SteveYurongSu (approved):
Well done!

PR #4565, opened 2021-12-13
Fix IT framework

cornmonster:
1. disable Influx DB RPC service to avoid port conflict


HTHou (approved):
LGTM

neuyilan (changes requested):
This pr LGTM but whether it is necessary to enable `enable_influxdb_rpc_service=true` as default value in `iotdb-engine.properties`?

SteveYurongSu:
> This pr LGTM but whether it is necessary to enable `enable_influxdb_rpc_service=true` as default value in `iotdb-engine.properties`?

Thanks for pointing this out. @neuyilan  

After the development of the InfluxDB adapter is completed, the parameter `enable_influxdb_rpc_service` will be set to `false` by default. We now set this parameter to true just to make it easier to test using the pipeline. @jun0315

neuyilan:
> We now set this parameter to true just to make it easier to test using the pipeline
OK, got it, thanks.

PR #4566, opened 2021-12-13
[To rel/0.12] Fix connection refused using session when users forget to set client ip

MrQuansy:
Users only set internal ip in iotdb-cluster.properties, and they often forget to set the rpc_address (public ip) in iotdb-engine.properties. So the rpc address is the default value(0.0.0.0). Session will cache the 0.0.0.0 endpoint when an insertPlan is redirected. 
We want internal ip as private ip and rpc_address as public ip. So if the rpc_address is not set, set it the same with internal ip.

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/ClusterMain.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/ClusterMain.java
@@ -96,6 +96,13 @@ public class ClusterMain {
       return;
     }
 
+    // if client ip is the default address, set it same with internal ip
+    if (IoTDBDescriptor.getInstance().getConfig().getRpcAddress().equals("0.0.0.0")) {
+      IoTDBDescriptor.getInstance()
+          .getConfig()
+          .setRpcAddress(ClusterDescriptor.getInstance().getConfig().getInternalIp());
+    }
+
     String mode = args[0];
 
     logger.info("Running mode {}", mode);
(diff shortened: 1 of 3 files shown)


wangchao316 (commented):
hi , I have a little question. why do you alter this?
could you please say reason?

LebronAl (approved):
LGTM

wangchao316:
> public ip

For security purposes, we generally do not set public IP addresses. We need to set specific IP addresses and enable the network firewall and filtering.

LebronAl:
> For security purposes, we generally do not set public IP addresses. We need to set specific IP addresses and enable the network firewall and filtering.

We were only concerned about helping users reduce the confusion of writing errors, not security before. Your suggestion is very meaningful. Do you have any suggestions for helping users in this scenario while avoiding security issues?

wangchao316:
> We were only concerned about helping users reduce the confusion of writing errors, not security before. Your suggestion is very meaningful. Do you have any suggestions for helping users in this scenario while avoiding security issues?

this pr  could merge, I only suggest that the customer also need to configure the IP address in the production environment.

wangchao316 (approved):

LebronAl:
> this pr could merge, I only suggest that the customer also need to configure the IP address in the production environment.

Ok, welcome to tell us more about your thoughts on security, so we may need to change our assumptions about public IP.

PR #4567, opened 2021-12-13
Fix connection refused using session when users forget to set client ip

MrQuansy:
Users only set internal ip in iotdb-cluster.properties, and they often forget to set the rpc_address (public ip) in iotdb-engine.properties. So the rpc address is the default value(0.0.0.0). Session will cache the 0.0.0.0 endpoint when an insertPlan is redirected. 
We want internal ip as private ip and rpc_address as public ip. So if the rpc_address is not set, set it the same with internal ip.


neuyilan (commented):
What is the purpose of this? Could you please describe it in detail? Thanks.

LebronAl (approved):
LGTM

neuyilan (approved):
LGTM

PR #4569, opened 2021-12-14
[IOTDB-2139] Move example udfs to src/java package to pass cluster IT

ericpai:
See JIRA: https://issues.apache.org/jira/browse/IOTDB-2139

As in cluster test, the test and server are separated in different processes, UDF registeration must be done under UDF class pre-complied. So moving those claasses to src/java package can be involved to integration.jar by executing `mvn verify`.


ericpai:
@SteveYurongSu PTAL~

SteveYurongSu (approved):
LGTM

PR #4570, opened 2021-12-14
[IOTDB-2143] fix wrong precision in cluster mode

cornmonster:
See: https://issues.apache.org/jira/browse/IOTDB-2143

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
@@ -1188,7 +1188,13 @@ public class CMManager extends MManager {
     return paths;
   }
 
-  /** Similar to method getAllTimeseriesPath(), but return Path with alias alias. */
+  /**
+   * Similar to method getAllTimeseriesPath(), but return Path with alias.
+   *
+   * <p>Please note that for a returned measurement path, the name, alias and datatype are
+   * guaranteed to be accurate, while the compression type, encoding and other fields are not. See
+   * {@link GetAllPathsResult}
+   */
   @Override
   public Pair<List<MeasurementPath>, Integer> getMeasurementPathsWithAlias(
       PartialPath pathPattern, int limit, int offset) throws MetadataException {
@@ -1213,7 +1219,11 @@ public class CMManager extends MManager {
   }
 
   /**
-   * Get all paths after removing wildcards in the path
+   * Get all paths after removing wildcards in the path.
+   *
+   * <p>Please note that for a returned measurement path, the name, alias and datatype are
+   * guaranteed to be accurate, while the compression type, encoding and other fields are not. See
+   * {@link GetAllPathsResult}.
    *
    * @param originPath a path potentially with wildcard
    * @return all paths after removing wildcards in the path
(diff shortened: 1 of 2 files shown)


HTHou (approved):

LebronAl (commented):
Thank you for your contribution.
Although the current changes will work, in my opinion, a more elegant way to solve this problem would be to address the source of the problem rather than tinkering with it. It is also very likely that the metadata will be refetched at the data node using the CMManager each time the query is performed.
I think we should make the schema returned by the CMManager interface more precise (including encoding and other variables in GetAllPathsResult), and also build SingleSeriesQueryRequest, MultSeriesQueryRequest and so on with the corresponding structure to solve this problem. Not only will this solve the problem at the source, but it may also avoid some potential bugs in the future.
What's your opinion?

cornmonster:
>
1. Adding schemas to GetAllPathsResult may double its size, which may cause performance issues. So my suggestion is to provide 2 RPCs, one for basic information like pathname, data type, alias, and the other one for extended information like encoding and compression type.

1. For SingleSeriesQueryRequest , since the coordinator may not have the metadata, it is reasonable to let the data group member set the schema. Same for MultSeriesQueryRequest.

LebronAl:
> 1. Adding schemas to GetAllPathsResult may double its size, which may cause performance issues. So my suggestion is to provide 2 RPCs, one for basic information like pathname, data type, alias, and the other one for extended information like encoding and compression type.

I don't think adding the Encoding will double the size of GetAllPathsResult. In my opinion, most of the space in this structure is taken up by paths and aliasList of type List<String>, and List<byte> should not take up much space.

If you really want to save it, it's possible to maintain an additional RPC interface, but you can simply add parameters to the current RPC interface to indicate whether additional fields are needed.

In the current cluster architecture, our CMManager will transfer a large number of paths in RPC, which is relatively performance adverse. Our recently designed new cluster architecture will try to avoid this process, welcome to join the discussion.

> 2. For SingleSeriesQueryRequest , since the coordinator may not have the metadata, it is reasonable to let the data group member set the schema. Same for MultSeriesQueryRequest.

If the coordinator node doesn't need to retrieve metadata to build queryPlan, then I don't think there's anything wrong with your statement. However, in the current architecture, the coordinator node already needs to fetch all the metadata when parsing SQL into a Queryplan. In my opinion, rather than having the data node send the RPC again to fetch the metadata, it would be better to have the RPC transmit a few more bytes (most of which are occupied by the string variable).

What do you think?

neuyilan:
I have the same idea with @LebronAl , 
It's better to add the metadata of the measurement when returning the `GetAllPathsResult` or  change the interface `public List<MeasurementPath> getMatchedPaths(PartialPath originPath, boolean fetchMetadata)` in one RPC to indict whether we need to return all the metadata of the measurement, after all, one more RPC is unnecessary

LebronAl:
any updates?

cornmonster:
> any updates?

Sorry for not getting back to you sooner, I am working on IOTDB-2153, which is a much server issue, for the last few days. 

I will update this PR today or tomorrow.

cornmonster:
@LebronAl PTAL

LebronAl (approved):
LGTM~
The design of the new cluster architecture will minimize the transfer of large amounts of metadata

neuyilan (approved):
LGTM

PR #4571, opened 2021-12-14
[IOTDB-2141] Fix some issues in cluster test

ericpai:
See JIRA: https://issues.apache.org/jira/browse/IOTDB-2141

Some IT cases related issues have been fixed.

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
@@ -307,7 +307,12 @@ public class ClusterPlanExecutor extends PlanExecutor {
     // if path doesn't end with **.
     // e.g. we have SG root.sg.a and root.sg.b, the query path is root.sg, we should return the map
     // with key root.sg.a and root.sg.b instead of an empty one.
-    PartialPath wildcardPath = path.concatNode(IoTDBConstant.MULTI_LEVEL_PATH_WILDCARD);
+    PartialPath wildcardPath;
+    if (!path.getMeasurement().equals(IoTDBConstant.MULTI_LEVEL_PATH_WILDCARD)) {
+      wildcardPath = path.concatNode(IoTDBConstant.MULTI_LEVEL_PATH_WILDCARD);
+    } else {
+      wildcardPath = path;
+    }
     Map<String, String> sgPathMap = IoTDB.metaManager.groupPathByStorageGroup(wildcardPath);
     if (sgPathMap.isEmpty()) {
       throw new PathNotExistException(path.getFullPath());
--- a/cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
@@ -307,11 +307,9 @@ public class ClusterPlanExecutor extends PlanExecutor {
     // if path doesn't end with **.
     // e.g. we have SG root.sg.a and root.sg.b, the query path is root.sg, we should return the map
     // with key root.sg.a and root.sg.b instead of an empty one.
-    PartialPath wildcardPath;
-    if (!path.getMeasurement().equals(IoTDBConstant.MULTI_LEVEL_PATH_WILDCARD)) {
-      wildcardPath = path.concatNode(IoTDBConstant.MULTI_LEVEL_PATH_WILDCARD);
-    } else {
-      wildcardPath = path;
+    PartialPath wildcardPath = path;
+    if (!wildcardPath.getMeasurement().equals(IoTDBConstant.MULTI_LEVEL_PATH_WILDCARD)) {
+      wildcardPath = wildcardPath.concatNode(IoTDBConstant.MULTI_LEVEL_PATH_WILDCARD);
     }
     Map<String, String> sgPathMap = IoTDB.metaManager.groupPathByStorageGroup(wildcardPath);
     if (sgPathMap.isEmpty()) {
--- a/cluster/src/main/java/org/apache/iotdb/cluster/query/aggregate/ClusterAggregator.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/query/aggregate/ClusterAggregator.java
@@ -244,12 +244,16 @@ public class ClusterAggregator {
               results);
           return results;
         }
-      } catch (TException | IOException e) {
+      } catch (TException e) {
         logger.error(
-            "{}: Cannot query aggregation {} from {}", metaGroupMember.getName(), path, node, e);
+            metaGroupMember.getName() + " query aggregation error " + path + " from " + node, e);
+        throw new StorageEngineException(e.getMessage());
+      } catch (IOException e) {
+        logger.error(
+            metaGroupMember.getName() + " cannot query aggregation " + path + " from " + node, e);
       } catch (InterruptedException e) {
         Thread.currentThread().interrupt();
-        logger.error("{}: query {} interrupted from {}", metaGroupMember.getName(), path, node, e);
+        logger.error(metaGroupMember.getName() + " query interrupted " + path + " from " + node, e);
       }
     }
     throw new StorageEngineException(
--- a/cluster/src/main/java/org/apache/iotdb/cluster/query/aggregate/ClusterAggregator.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/query/aggregate/ClusterAggregator.java
@@ -46,6 +46,7 @@ import org.apache.iotdb.tsfile.file.metadata.enums.TSDataType;
 import org.apache.iotdb.tsfile.read.common.Path;
 import org.apache.iotdb.tsfile.read.filter.basic.Filter;
 
+import org.apache.thrift.TApplicationException;
 import org.apache.thrift.TException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -244,11 +245,11 @@ public class ClusterAggregator {
               results);
           return results;
         }
-      } catch (TException e) {
+      } catch (TApplicationException e) {
         logger.error(
             metaGroupMember.getName() + " query aggregation error " + path + " from " + node, e);
         throw new StorageEngineException(e.getMessage());
-      } catch (IOException e) {
+      } catch (TException | IOException e) {
         logger.error(
             metaGroupMember.getName() + " cannot query aggregation " + path + " from " + node, e);
       } catch (InterruptedException e) {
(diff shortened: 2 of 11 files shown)


ericpai:
@LebronAl PTAL~

cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
```diff
@@ -307,7 +307,12 @@ protected int getNodesNumInGivenLevel(PartialPath path, int level) throws Metada
     // if path doesn't end with **.
     // e.g. we have SG root.sg.a and root.sg.b, the query path is root.sg, we should return the map
     // with key root.sg.a and root.sg.b instead of an empty one.
-    PartialPath wildcardPath = path.concatNode(IoTDBConstant.MULTI_LEVEL_PATH_WILDCARD);
+    PartialPath wildcardPath;
+    if (!path.getMeasurement().equals(IoTDBConstant.MULTI_LEVEL_PATH_WILDCARD)) {
+      wildcardPath = path.concatNode(IoTDBConstant.MULTI_LEVEL_PATH_WILDCARD);
+    } else {
+      wildcardPath = path;
+    }
```
Alima777, line 315:
```suggestion
    PartialPath wildcardPath = path;
    if (!wildcardPath.getMeasurement().equals(IoTDBConstant.MULTI_LEVEL_PATH_WILDCARD)) {
      wildcardPath = wildcardPath.concatNode(IoTDBConstant.MULTI_LEVEL_PATH_WILDCARD);
    }
```

Alima777 (approved):
LGTM~
And a little advice: try to avoid fix multi issues in one PR, it will be difficult to test and revert if something gets wrong.

cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
ericpai: Thanks for your comments. The codes have been refined. Does this PR need to be separeted to different ones? If so, I will close this one.

Alima777: Pay attention to it next time is ok~

integration/src/test/java/org/apache/iotdb/db/integration/IoTDBTtlIT.java
```diff
@@ -39,7 +39,7 @@
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
-@Category({LocalStandaloneTest.class, ClusterTest.class})
+@Category({LocalStandaloneTest.class})
```
LebronAl, line 42: why delete ClusterTest?

cluster/src/main/java/org/apache/iotdb/cluster/query/aggregate/ClusterAggregator.java
```diff
@@ -244,12 +244,16 @@ public ClusterAggregator(MetaGroupMember metaGroupMember) {
               results);
           return results;
         }
-      } catch (TException | IOException e) {
+      } catch (TException e) {
+        logger.error(
+            metaGroupMember.getName() + " query aggregation error " + path + " from " + node, e);
+        throw new StorageEngineException(e.getMessage());
```
LebronAl, line 250: any reasons ?

cluster/src/main/java/org/apache/iotdb/cluster/query/reader/ClusterReaderFactory.java
```diff
@@ -897,11 +897,14 @@ private GroupByExecutor getRemoteGroupByExecutor(
           logger.debug("{}: no data for {} from {}", metaGroupMember.getName(), path, node);
           return new EmptyReader();
         }
-      } catch (TException | IOException e) {
-        logger.error("{}: Cannot query {} from {}", metaGroupMember.getName(), path, node, e);
+      } catch (TException e) {
+        logger.error(metaGroupMember.getName() + ": Cannot query " + path + " from " + node, e);
+        throw new StorageEngineException(e.getMessage());
```
LebronAl, line 902: any reasons ?

LebronAl (commented):
Could you please add some comments or simply describe the reasons for the following changes in the PR? This will make it easier to follow up later

integration/src/test/java/org/apache/iotdb/db/integration/IoTDBTtlIT.java
ericpai: The test case testDefaultTTL() (without any test categories) relies on the change of default_ttl configure parameter. The change is not supported yet in ConfigFactory. New configuration parameters support in test will be submitted in other PRs.

cluster/src/main/java/org/apache/iotdb/cluster/query/aggregate/ClusterAggregator.java
ericpai: In my opinion, the retrying to other nodes only happends when there're timeout exception.
If some exception, like 'UnSupportedDataTypeException' throws, there's no need to retry in other nodes, and we should tell the caller the true error message, not the outer 'TimeoutExecption'.

cluster/src/main/java/org/apache/iotdb/cluster/query/reader/ClusterReaderFactory.java
ericpai: Same as above.

cluster/src/main/java/org/apache/iotdb/cluster/server/service/DataSyncService.java
```diff
@@ -352,8 +355,12 @@ public ByteBuffer getAllMeasurementSchema(RaftNode header, ByteBuffer planBinary
...
     try {
       return dataGroupMember.getLocalQueryExecutor().getAggrResult(request);
-    } catch (StorageEngineException | QueryProcessException | IOException e) {
-      throw new TException(e);
+    } catch (StorageEngineException
+        | QueryProcessException
+        | IOException
+        | StatisticsClassException
+        | UnSupportedDataTypeException e) {
+      throw new TApplicationException(e.getMessage());
```
ericpai, line 363: The raw exception message will be shadowed by TException as it doesn't serialize the root cause to other nodes.

integration/src/test/java/org/apache/iotdb/db/integration/IoTDBMetadataFetchIT.java
```diff
@@ -52,7 +52,6 @@
  * Notice that, all test begins with "IoTDB" is integration test. All test which will start the
  * IoTDB server should be defined as integration test.
  */
-@Category({LocalStandaloneTest.class, ClusterTest.class, RemoteTest.class})
```
ericpai, line 55: The test case `databaseMetaDataTest()` needs to dump the whole MTree as a json string. However in a cluster, each node doesn't have the whole tree. And I think the metadata interface in JDBC is not called regularly.

server/src/main/java/org/apache/iotdb/db/utils/ErrorHandlingUtils.java
```diff
@@ -106,6 +107,9 @@ public static TSStatus tryCatchQueryException(Exception e) {
     } else if (e instanceof IoTDBException && !(e instanceof StorageGroupNotReadyException)) {
       DETAILED_FAILURE_QUERY_TRACE_LOGGER.warn(INFO_QUERY_PROCESS_ERROR, e);
       return RpcUtils.getStatus(((IoTDBException) e).getErrorCode(), getRootCause(e));
+    } else if (e instanceof TsFileRuntimeException) {
+      DETAILED_FAILURE_QUERY_TRACE_LOGGER.warn(INFO_QUERY_PROCESS_ERROR, e);
```
ericpai, line 111: If an aggregation result query throws an UnSupportedException, we should return a proper error code instead of 500.

integration/src/test/java/org/apache/iotdb/db/integration/IoTDBTtlIT.java
LebronAl: got it

cluster/src/main/java/org/apache/iotdb/cluster/query/aggregate/ClusterAggregator.java
LebronAl: > the retrying to other nodes only happends when there're timeout exception.
Agree with you.

However, I recorded that if a timeout occurs in an RPC call, TException is still thrown, and the new implementation is still not the way we want it to be. Do you have any ideas on this? Or we could refer to existing [implementations](https://github.com/apache/iotdb/blob/master/cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java#L1377), but I'm not sure whether just specifying `socketTimeoutException` is enough.

ericpai: > > the retrying to other nodes only happends when there're timeout exception.
> 
> Agree with you.
> 
> However, I recorded that if a timeout occurs in an RPC call, TException is still thrown, and the new implementation is still not the way we want it to be. Do you have any ideas on this? Or we could refer to existing [implementations](https://github.com/apache/iotdb/blob/master/cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java#L1377), but I'm not sure whether just specifying `socketTimeoutException` is enough.

Maybe we can check whether it's a TApplicationiException first, and then TException? A TApplicationException usally means there's some application error in the server side, the request-response is not timeout.

I have found another picture about the exception tree in Thrift. 
image

ericpai: I have rewrited these exception handling codes.

LebronAl: good idea!

LebronAl (approved):
LGTM

cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
```diff
@@ -307,7 +307,10 @@ protected int getNodesNumInGivenLevel(PartialPath path, int level) throws Metada
     // if path doesn't end with **.
     // e.g. we have SG root.sg.a and root.sg.b, the query path is root.sg, we should return the map
     // with key root.sg.a and root.sg.b instead of an empty one.
-    PartialPath wildcardPath = path.concatNode(IoTDBConstant.MULTI_LEVEL_PATH_WILDCARD);
+    PartialPath wildcardPath = path;
+    if (!wildcardPath.getMeasurement().equals(IoTDBConstant.MULTI_LEVEL_PATH_WILDCARD)) {
+      wildcardPath = wildcardPath.concatNode(IoTDBConstant.MULTI_LEVEL_PATH_WILDCARD);
+    }
```
neuyilan, line 313: Thanks for your job, however, I have one issue, It's a bug or not？if so, why not add one UT to cover it?

ericpai: `IoTDBMetadataFetchIT.showCountTimeSeries` can test this in the cluster test. I have just found that the ClusterPlanExecutor has its own UT file. And I will try to add UTs in following PRs.

PR #4575, opened 2021-12-15
[IOTDB-2162] Simplify the recovery merge process

cigarl:
Simply the `MergeRecoveryTask`, we may just need to clean last info.


wangchao316:
I  suggest we raise a jira firstly, and then raise pr.. So we can get a good tally of.

cigarl:
> I suggest we raise a jira firstly, and then raise pr.. So we can get a good tally of.

Thanks for your reminding, I have created a [issue](https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-2162?filter=allopenissues).

server/src/main/java/org/apache/iotdb/db/engine/compaction/cross/inplace/task/CleanLastMergeTask.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.util.List;
+
+/**
+ * RecoverCrossMergeTask is an extension of MergeTask, which resumes the last merge progress by
+ * scanning merge.log using LogAnalyzer and continue the unfinished merge.
+ */
+public class CleanLastMergeTask extends CrossSpaceMergeTask {
```
qiaojialin, line 35: In master, we'd better rename merge to cross space compaction.
Another is inner space compaction.

cigarl: ok~ the related names have been changed.

qiaojialin (approved):

PR #4581, opened 2021-12-15
[IOTDB-2153][IOTDB-2157] fix wildcard issues

cornmonster:
See: https://issues.apache.org/jira/browse/IOTDB-2153 and https://issues.apache.org/jira/browse/IOTDB-2157

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
@@ -920,7 +920,7 @@ public class CMManager extends MManager {
    * @return all paths after removing wildcards in the path
    */
   public Set<PartialPath> getMatchedDevices(PartialPath originPath) throws MetadataException {
-    Map<String, String> sgPathMap = groupPathByStorageGroup(originPath);
+    Map<String, List<PartialPath>> sgPathMap = groupPathByStorageGroup(originPath);
     Set<PartialPath> ret = getMatchedDevices(sgPathMap);
     logger.debug("The devices of path {} are {}", originPath, ret);
     return ret;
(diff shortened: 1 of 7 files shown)


cornmonster:
@zyk990424 Please have a look at this PR. Thanks!

I decided to remove `PartialPath#alterPrefixPath` for 2 reasons:
1. the meaning of the method is not clear
2. returning multiple possible results means the caller need to query (or delete) each of the returned path, which is a waste of time.

zyk990424:
> @zyk990424 Please have a look at this PR. Thanks!
> 
> I decided to remove `PartialPath#alterPrefixPath` for 2 reasons:
> 
> 1. the meaning of the method is not clear
> 2. returning multiple possible results means the caller need to query (or delete) each of the returned path, which is a waste of time.

If you decide to remove this method, there should be a method that satisfy the requirements of the scenarios.  
This method currently is mainly used for dividing logic path set, represented by the given path pattern, into several buckets identified by storage group. 
Maybe using both the origin path pattern and the prefix path as parameters to filter series is fine.

zyk990424 (approved):

zyk990424:
@qiaojialin @LebronAl please have a look

zyk990424 (commented):
LGTM~

cornmonster:
I checked the result of the cluster IT.

1. The failure of aligned-by-device tests is expected.
2. IoTDBFillIT failed because the env is unstable. 
3. The failure of the precision issue of float numbers is expected. Please see: https://github.com/apache/iotdb/pull/4570
4. The failure of tag tests is expected. Please see: https://github.com/apache/iotdb/pull/4613

ericpai:
@qiaojialin @LebronAl Does this PR have any other issues to solve? If not, I think we should merge it as soon as possible because the query with wildcards in cluster is a high frequency operation. Other refining works about MTree can be done later.

qiaojialin (approved):

PR #4587, opened 2021-12-16
[IOTDB-2163] Fix unexpected amount of columns in a cluster slimit query

ericpai:
See JIRA: https://issues.apache.org/jira/browse/IOTDB-2163


ericpai:
@SteveYurongSu PTAL

@JackieTien97 Although the root cause is clear. Why the test case can be passed in standalone test? Are there any bugs in `MeasurementCollector` with curOffset updating? I compared the implementations between `CMManger.getMeasurementPathsWithAlias` with `MManager.getMeasurementPathsWithAlias`. I'm certain about the impelmentation of  `CMManger`‘s correct.

ericpai:
@SteveYurongSu @JackieTien97 Are there any comments about this fix?

SteveYurongSu (approved):
LGTM

PR #4594, opened 2021-12-16
[To rel/0.12][IOTDB-1408]Statement with 'as' executes incorrectly in …

cmlmakahts:
One alias should only match with exactly one timeseries

【operations】
set storage group to root.auto1.wf02.wt02;
set storage group to root.auto1.wf01.wt02;
create timeseries root.auto1.wf01.wt02.status with datatype=boolean, encoding=plain
create timeseries root.auto1.wf01.wt02.temperature with datatype=float, encoding=plain, compression=snappy,max_point_number=3
create timeseries root.auto1.wf02.wt02.status with datatype=float, encoding=plain, compression=snappy,max_point_number=3
create timeseries root.auto1.wf02.wt02.temperature with datatype=float, encoding=plain, compression=snappy,max_point_number=3
insert into root.auto1.wf01.wt02(timestamp, status) values (now(), false);
insert into root.auto1.wf01.wt02(timestamp, temperature) values (now(), 66.7);
insert into root.auto1.wf02.wt02(timestamp, status) values (now(), 22.6);
insert into root.auto1.wf02.wt02(timestamp, temperature) values (now(), 33.7); 

SELECT status as speed FROM root.auto1.wf02.wt02,root.auto1.wf01.wt02
【result】
image

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/qp/strategy/optimizer/ConcatPathOptimizer.java
+++ b/server/src/main/java/org/apache/iotdb/db/qp/strategy/optimizer/ConcatPathOptimizer.java
@@ -45,9 +45,11 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 /** concat paths in select and from clause. */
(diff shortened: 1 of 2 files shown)


wangchao316 (approved):
LGTM

Alima777:
Hi, there are some faults in the method `checkAndSetTsAlias()` 

For example:
IoTDB> SELECT status as speed, status as speed  FROM root.auto1.wf02.wt02
Msg: 411: Error occurred in query process: alias 'speed' can only be matched with one time series
which should not throw an exception.

I think we can just define **different time series cannot has the same alias**. 
Under this definition, you can define aliasMap like: alias(String) -> partitalPath
If one alias can be corresponding to different paths, we throw an exception.

Alima777 (changes requested):
Please modify checkAndSetTsAlias() method

Alima777 (approved):
LGTM~

PR #4601, opened 2021-12-17
[IOTDB-2174] Fix Regexp filter serializing and deserializing error

ericpai:
See JIRA: https://issues.apache.org/jira/browse/IOTDB-2174


HTHou (approved):
LGTM

PR #4602, opened 2021-12-17
[IOTDB-2171, To Rel/0.12] Check the timeseries legitimacy when create timeseries

neuyilan:
https://issues.apache.org/jira/browse/IOTDB-2171

Performance Test:
create 1 million timeseries in one session client, calculate the total cost time.

Testing environment:
macOS
The time unit is milliseconds

  | 1000000 | 100000 | 10000
-- | -- | -- | --
new | 54835 | 6274 | 1186
old | 46704 | 5270 | 1118

From the test result, we can see that the new version will be slower than the old version, which also makes sense because, in this pr, we use antlr to check the grammar, which will cost more time.  However, the benefit is that when we check ourselves， we always miss some situations and are inaccurate but antlr can ensure this.

Any ideas about this?

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
@@ -43,6 +43,7 @@ import org.apache.iotdb.db.qp.physical.sys.MeasurementMNodePlan;
 import org.apache.iotdb.db.qp.physical.sys.ShowDevicesPlan;
 import org.apache.iotdb.db.qp.physical.sys.ShowTimeSeriesPlan;
 import org.apache.iotdb.db.qp.physical.sys.StorageGroupMNodePlan;
+import org.apache.iotdb.db.qp.strategy.LogicalGenerator;
 import org.apache.iotdb.db.query.context.QueryContext;
 import org.apache.iotdb.db.query.control.QueryResourceManager;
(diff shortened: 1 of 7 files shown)


server/src/main/java/org/apache/iotdb/db/metadata/MTree.java
```diff
@@ -222,7 +223,7 @@ MeasurementMNode createTimeseries(
     if (nodeNames.length <= 2 || !nodeNames[0].equals(root.getName())) {
       throw new IllegalPathException(path.getFullPath());
     }
-    checkTimeseries(path);
+    LogicalGenerator.checkCreateTimeseriesGrammar(path.getFullPath());
```
wangchao316, line 226: hi ,  I have a little question. Can we perform verification when the SQL (string) is converted into a plan instead of during metadata operations?

neuyilan: Because there are many situations that lead to the creation of `CreateTimesPlan`, but in the end, this method will be called, so I wrote here to prevent some situations from being missed.

neuyilan:
As anltr itself still has some problems, please see[1], and the solution has some performance issue, so this may not be the best solution, so I turn off this PR.
[1] https://issues.apache.org/jira/browse/IOTDB-2175

PR #4613, opened 2021-12-20
 [IOTDB-2181] Fix cluster tag

cornmonster:
See: https://issues.apache.org/jira/browse/IOTDB-2181


LebronAl (commented):
LGTM~But we might need to talk a little bit more about whether we should return 0 or an error if the tagkey doesn't exist.
@qiaojialin cc

qiaojialin:
Please update the UserGuide, add the behavior when the tag does not exist.

cornmonster:
@qiaojialin Sorry for taking so long to update the PR. Please take another look at it.

LebronAl (approved):
LGTM

qiaojialin (approved):

PR #4614, opened 2021-12-20
[IOTDB-2183] Fix the config problems in cluster mode

stormbroken:
Because of the development of iotdb server, there are some problems in cluster/src/resources(for metric framework and influxdb rpc service), it will fail when you use these config to start iotdb cluster.


LebronAl (approved):
LGTM

neuyilan (approved):
LGTM

PR #4616, opened 2021-12-21
[IOTDB-2043] refactor: remove haveSorted param from Session

SzyWilliam:
## Description
For some methods in Session, we have a haveSorted parameter to indicate whether multi records are in ascending order of time.

This parameter is useless, we could check this by ourselves.
(1) checkSorted
(2) if not sorted, we could sort it.

## What
1. Delete `haveSorted` param in `insertRecordsOfOneDevice, insertAlignedRecordsOfOneDevice`. Since currently there are two overloaded version of `insertRecordsOfOneDevice, insertAlignedRecordsOfOneDevice` and they only differ in `haveSorted` param, so I merged two overloaded version into a single interface
2. Change the caller usages of above methods

### DIFF
--- a/session/src/main/java/org/apache/iotdb/session/Session.java
+++ b/session/src/main/java/org/apache/iotdb/session/Session.java
@@ -1189,7 +1189,7 @@ public class Session {
    *
    * <p>Each row could have same deviceId but different time, number of measurements
    *
-   * @param haveSorted whether the times have been sorted
+   * @param haveSorted deprecated, whether the times have been sorted
    * @see Session#insertTablet(Tablet)
    */
   public void insertRecordsOfOneDevice(
@@ -1242,7 +1242,7 @@ public class Session {
    *
    * <p>Each row could have same prefixPath but different time, number of measurements
    *
-   * @param haveSorted whether the times have been sorted
+   * @param haveSorted deprecated, whether the times have been sorted
    * @see Session#insertTablet(Tablet)
    */
   public void insertAlignedRecordsOfOneDevice(
@@ -1290,12 +1290,7 @@ public class Session {
           "times, measurementsList and valuesList's size should be equal");
     }
 
-    if (haveSorted) {
-      if (!checkSorted(times)) {
-        throw new BatchExecutionException(
-            "Times in InsertOneDeviceRecords are not in ascending order");
-      }
-    } else {
+    if (!checkSorted(times)) {
       // sort
       Integer[] index = new Integer[times.size()];
       for (int i = 0; i < times.size(); i++) {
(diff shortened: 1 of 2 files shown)


session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
```diff
@@ -692,42 +658,7 @@ public void insertOneDeviceAlignedRecords(
       Session session = getSession();
       try {
         session.insertAlignedRecordsOfOneDevice(
-            deviceId, times, measurementsList, typesList, valuesList, false);
```
SzyWilliam, line 695: 这儿是将两个overload method给合并成了一个

session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -1179,35 +1179,14 @@ public void insertRecordsOfOneDevice(
       List<List<TSDataType>> typesList,
       List<List<Object>> valuesList)
       throws IoTDBConnectionException, StatementExecutionException {
-    insertRecordsOfOneDevice(deviceId, times, measurementsList, typesList, valuesList, false);
```
SzyWilliam, line 1182: two overload function which only differs in haveSorted param, thus merged to one

session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -1231,28 +1210,6 @@ public void insertAlignedRecordsOfOneDevice(
       List<List<TSDataType>> typesList,
       List<List<Object>> valuesList)
       throws IoTDBConnectionException, StatementExecutionException {
```
SzyWilliam, line 1212: two overload function which only differs in `haveSorted param`, thus merged to one

session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -1179,35 +1179,14 @@ public void insertRecordsOfOneDevice(
...
```
SzyWilliam, line 1181: two overload function which only differs in `haveSorted param`, thus merged to one

session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
```diff
@@ -625,41 +625,7 @@ public void insertOneDeviceRecords(
       Session session = getSession();
       try {
         session.insertRecordsOfOneDevice(
```
SzyWilliam, line 627: two overload function which only differs in `haveSorted param`, thus merged to one

qiaojialin (approved):

PR #3289, opened 2021-05-28
[IOTDB-1391]  Add a new Aggregation function ext

lyssom:
## Description
Add a new aggregation function: ext

### Content3 ...

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
    - [ ] concurrent read and write 
- [ ] added documentation for new or modified features or behaviors.
- [ ] added Javadocs for most classes and all non-trivial methods. 
- [ ] added or updated version, __license__, or notice information
- [ ] added comments explaining the "why" and the intent of the code wherever would not be obvious 
  for an unfamiliar reader.
- [ ] added unit tests or modified existing tests to cover new code paths, ensuring the threshold 
  for code coverage.
- [ ] added integration tests.
- [ ] been tested in a test IoTDB cluster.

<hr>

PR #4626, opened 2021-12-23
[To rel/0.12] [IOTDB-2192] Support extreme function

liuminghui233:
Support extreme function in v0.12
cherry-pick from [#3289](https://github.com/apache/iotdb/pull/3289)


ericpai:
Is it good to add new features to v0.12? It's better to make every patch version contain only bug fixes and performance  improvements.

liuminghui233:
> Is it good to add new features to v0.12? It's better to make every patch version contain only bug fixes and performance improvements.

Yeah, I agree with you. However, `extreme` is a commonly used aggregate function. Maybe some users need to use this function before v0.13.

PR #4627, opened 2021-12-23
[To rel/0.12][IOTDB-2194] Fix 0.12 show timeseries

stormbroken:
In current cluster version, if we have more than 2,000 timeseries, when we commend SHOW TIMESERIES, it will only display 2,000 timeseries, the same as the SHOW DEVICES...

This problem is fixed on master branch, but not cherry-pick to rel/0.12.

Related PR: [IOTDB-1266]SHOW TIMESERIES will only display 2000 timeseries
Related Commit ID: 55d9260


LebronAl (approved):
LGTM

PR #4617, opened 2021-12-21
[IOTDB-1960] Fix count timeseries in cluster mode.

stormbroken:
## The problem
count timeseries Msg: 304: Path [root.**] does not exist ，expect ：0

## After fix

image

PR #4633, opened 2021-12-24
[To rel/0.12][IOTDB-1960] Fix count timeseries in cluster mode.

stormbroken:
The problem
count timeseries Msg: 304: Path [root.**] does not exist ，expect ：0

Related PR: #4617


LebronAl (approved):
LGTM

PR #4645, opened 2021-12-28
[IOTDB-2217] Provide an iterator to return timeseries Path in dictionary order

samperson1997:
* Provide an iterator to return timeseries Path in dictionary order
* Add unit tests accordingly with all occasions

### DIFF
--- a/tsfile/src/main/java/org/apache/iotdb/tsfile/read/TsFileSequenceReader.java
+++ b/tsfile/src/main/java/org/apache/iotdb/tsfile/read/TsFileSequenceReader.java
@@ -649,19 +649,32 @@ public class TsFileSequenceReader implements AutoCloseable {
   }
 
   /**
-   * this function return all timeseries names in dictionary order
+   * this function return all timeseries names
    *
    * @return list of Paths
    * @throws IOException io error
    */
   public List<Path> getAllPaths() throws IOException {
-    if (tsFileMetaData == null) {
-      readFileMetadata();
-    }
     List<Path> paths = new ArrayList<>();
+    for (String device : getAllDevices()) {
+      Map<String, TimeseriesMetadata> timeseriesMetadataMap = readDeviceMetadata(device);
+      for (String measurementId : timeseriesMetadataMap.keySet()) {
+        paths.add(new Path(device, measurementId));
+      }
+    }
+    return paths;
+  }
+
+  /**
+   * @return an iterator of timeseries list, in which names of timeseries are ordered in dictionary order
+   * @throws IOException io error
+   */
+  public Iterator<List<Path>> getPathsIterator() throws IOException {
+    readFileMetadata();
 
     MetadataIndexNode metadataIndexNode = tsFileMetaData.getMetadataIndex();
     List<MetadataIndexEntry> metadataIndexEntryList = metadataIndexNode.getChildren();
+    Queue<Pair<String, Pair<Long, Long>>> queue = new LinkedList<>();
     for (int i = 0; i < metadataIndexEntryList.size(); i++) {
       MetadataIndexEntry metadataIndexEntry = metadataIndexEntryList.get(i);
       long endOffset = tsFileMetaData.getMetadataIndex().getEndOffset();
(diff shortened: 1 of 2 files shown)


tsfile/src/test/java/org/apache/iotdb/tsfile/write/MetadataIndexConstructorTest.java
```diff
@@ -49,11 +49,7 @@
 
 import java.io.File;
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
ericpai, line 52: Please use single class import instead of wildcard

tsfile/src/main/java/org/apache/iotdb/tsfile/read/TsFileSequenceReader.java
```diff
@@ -665,6 +665,100 @@ protected int binarySearchInTimeseriesMetadataList(
     return paths;
   }
 
+  /**
+   * @return an iterator of timeseries list, in which names of timeseries are ordered in dictionary order
+   * @throws IOException io error
+   */
+  public Iterator<List<Path>> getPathsIterator() throws IOException {
```
ericpai, line 672: Where is this method called, or this one will be used in later PRs?

tsfile/src/test/java/org/apache/iotdb/tsfile/write/MetadataIndexConstructorTest.java
```diff
@@ -237,6 +235,19 @@ private void test(String[] devices, int[][] vectorMeasurement, String[][] single
...
     }
+    try (TsFileSequenceReader reader = new TsFileSequenceReader(FILE_PATH)) {
+      Iterator<List<Path>> iterator = reader.getPathsIterator();
+      int idx = 0;
+      while (iterator.hasNext()) {
+        for (Path actualPath : iterator.next()) {
+          assertEquals(actualPath.getFullPath(), correctPaths.get(idx));
+          idx++;
+        }
+      }
```
ericpai, line 246: I think we may need to assert the `idx` value finally, as the paths size returned by the iterator may be less than the size of `correctPaths` and the test can't find this bug if it really happens.
Add `Assert.assertEquals(correctPaths.size(), idx);`

ericpai (changes requested):

tsfile/src/test/java/org/apache/iotdb/tsfile/write/MetadataIndexConstructorTest.java
samperson1997: Sorry I'm trying to fix the code format in my new laptop today. This is fixed in the latest commit : )

tsfile/src/main/java/org/apache/iotdb/tsfile/read/TsFileSequenceReader.java
samperson1997: This method will be called by compaction module. Thanks for your kindly remind.

tsfile/src/test/java/org/apache/iotdb/tsfile/write/MetadataIndexConstructorTest.java
samperson1997: Great point! Thanks a lot.

tsfile/src/main/java/org/apache/iotdb/tsfile/read/TsFileSequenceReader.java
```diff
@@ -665,6 +665,95 @@ protected int binarySearchInTimeseriesMetadataList(
...
+      public List<Path> next() {
+        if (!hasNext()) {
+          throw new NoSuchElementException();
+        }
+        Pair<String, Pair<Long, Long>> startEndPair = queue.remove();
+        List<Path> paths = new ArrayList<>();
+        try {
+          List<TimeseriesMetadata> timeseriesMetadataList = new ArrayList<>();
+          ByteBuffer nextBuffer = readData(startEndPair.right.left, startEndPair.right.right);
+          while (nextBuffer.hasRemaining()) {
```
ericpai, line 704: If the local variable `timeseriesMetadataList` is used only for one-time iteration, maybe this is more efficient without this list?
```suggestion
          while (nextBuffer.hasRemaining()) {
              paths.add(
                new Path(startEndPair.left, 
                  TimeseriesMetadata.deserializeFrom(nextBuffer, false).getMeasurementId()));
          }
```

samperson1997: Great! Fixed

ericpai:
LGTM

ericpai (approved):

JackieTien97 (approved):

PR #4648, opened 2021-12-28
[To rel/0.12] [IOTDB-2219] Bug fix: query in-memory data is incorrect in cluster mode

liuminghui233:
image

### DIFF
--- a/cluster/src/test/java/org/apache/iotdb/cluster/query/ClusterDataQueryExecutorTest.java
+++ b/cluster/src/test/java/org/apache/iotdb/cluster/query/ClusterDataQueryExecutorTest.java
@@ -48,11 +48,12 @@ import org.junit.Before;
 import org.junit.Test;
 
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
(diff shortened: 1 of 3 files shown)


neuyilan (changes requested):
Could you please explain why? 
if remove the code here, will the single node not benefit from the cache?  and  the `cachedQueryDataSourcesMap` attribute seems do not used anywhere.

liuminghui233:
> Could you please explain why? if remove the code here, will the single node not benefit from the cache? and the `cachedQueryDataSourcesMap` attribute seems do not used anywhere.

In [IOTDB-2101](https://issues.apache.org/jira/browse/IOTDB-2101), we refactor the construction process of `QueryDataSource`. We let all sensors in one virtual sg share the same `QueryDataSource` to save memory.

In the standalone mode, we will call the `initQueryDataSource()` method to construct all the required `QueryDataSource` in this query and save it in `cachedQueryDataSourcesMap`.

But in the cluster mode, for some reason, we must construct a `QueryDataSource` for each series.

I will post more details about the construction and use of `QueryDataSource` to a confluence page later.

PR #4650, opened 2021-12-28
[To rel/0.12] [IOTDB-2221] Accelerate query by reducing the number of files in QueryDataSource

liuminghui233:
In this PR,
1. When all the selected series are under the same device, we use the time range of the device to filter QueryDataSource.
2. In query with value filter, we construct time filter from expression to filter QueryDataSource.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
@@ -644,7 +644,7 @@ public class TsFileResource {
   }
 
   /** @return true if the TsFile lives beyond TTL */
-  public boolean isSatisfied(Filter timeFilter, boolean isSeq, long ttl, boolean debug) {
+  private boolean isSatisfied(Filter timeFilter, boolean isSeq, long ttl, boolean debug) {
     long startTime = getFileStartTime();
     long endTime = closed || !isSeq ? getFileEndTime() : Long.MAX_VALUE;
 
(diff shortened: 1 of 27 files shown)


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
```diff
@@ -608,6 +608,41 @@ public boolean stillLives(long timeLowerBound) {
...
+        DEBUG_LOGGER.info(
+            "Path: {} file {} is not satisfied because of time filter!", deviceId, fsFactory);
+      }
+      return res;
+    }
+    return true;
+  }
+
   /** @return true if the TsFile lives beyond TTL */
   public boolean isSatisfied(Filter timeFilter, boolean isSeq, long ttl, boolean debug) {
```
JackieTien97, line 647:
```suggestion
  private boolean isSatisfied(Filter timeFilter, boolean isSeq, long ttl, boolean debug) {
```

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
```diff
@@ -632,12 +667,7 @@ public boolean isSatisfied(Filter timeFilter, boolean isSeq, long ttl, boolean d
 
   /** @return true if the device is contained in the TsFile and it lives beyond TTL */
```
JackieTien97, line 668:
```suggestion
  /** @return true if the device is contained in the TsFile */
```

JackieTien97 (changes requested):

JackieTien97 (approved):

PR #4657, opened 2021-12-29
[IOTDB-2224] Accelerate tests by using batch insert

ericpai:
See JIRA: https://issues.apache.org/jira/browse/IOTDB-2224

### DIFF
--- a/integration/src/test/java/org/apache/iotdb/db/integration/IoTDBRestartIT.java
+++ b/integration/src/test/java/org/apache/iotdb/db/integration/IoTDBRestartIT.java
@@ -434,7 +434,6 @@ public class IoTDBRestartIT {
 
     try {
       CompactionTaskManager.getInstance().waitAllCompactionFinish();
-      Thread.sleep(10000);
       EnvironmentUtils.restartDaemon();
     } catch (Exception e) {
       Assert.fail();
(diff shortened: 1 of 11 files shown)


ericpai:
@HTHou PTAL. This PR increases some unexpected slow tests.

integration/src/test/java/org/apache/iotdb/db/integration/IoTDBRestartIT.java
```diff
@@ -434,7 +434,6 @@ public void testRestartCompaction()
 
     try {
       CompactionTaskManager.getInstance().waitAllCompactionFinish();
-      Thread.sleep(10000);
```
HTHou, line 437: Why do you remove it? I remember the tests may fail if not sleep here...

ericpai: In the `waitAllCompactionFinish()` it will wait until `taskExecutionPool.getActiveCount() == 0 && taskExecutionPool.getQueue().size() == 0`. So I don't know what the `sleep()` are waiting for?

ericpai: @THUMarkLau This is relatived to your work. Could you explain the design of this test case :)

THUMarkLau: `testRestartCompaction` can be removed in 0.13, because we redesigned and implemented the Compaction module. The amount of data inserted in this test does not trigger a compaction. There are more detailed tests of restarting a Compaction than this one.

ericpai: Thanks! I have deleted this test case in the current PR

HTHou (approved):
lgtm

PR #4659, opened 2021-12-29
[To rel/0.12][IOTDB-1297] Refactor the memory control when enabling time partitions

HeimingZ:
wait awhile for bytebuffers ready when creating wal node and meeting OOM

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
@@ -252,7 +252,7 @@ public class StorageGroupProcessor {
 
   private final Deque<ByteBuffer> walByteBufferPool = new LinkedList<>();
 
-  private AtomicInteger currentWalPoolSize = new AtomicInteger(0);
+  private int currentWalPoolSize = 0;
 
   // this field is used to avoid when one writer release bytebuffer back to pool,
   // and the next writer has already arrived, but the check thread get the lock first, it find the
@@ -268,7 +268,10 @@ public class StorageGroupProcessor {
 
   private volatile boolean compacting = false;
 
-  /** get the direct byte buffer from pool, each fetch contains two ByteBuffer */
+  /**
+   * get the direct byte buffer from pool, each fetch contains two ByteBuffer, return null if fetch
+   * fails
+   */
   public ByteBuffer[] getWalDirectByteBuffer() {
     ByteBuffer[] res = new ByteBuffer[2];
     synchronized (walByteBufferPool) {
@@ -276,7 +279,7 @@ public class StorageGroupProcessor {
       int MAX_WAL_BYTEBUFFER_NUM =
           config.getConcurrentWritingTimePartition()
               * config.getMaxWalBytebufferNumForEachPartition();
-      while (walByteBufferPool.isEmpty() && currentWalPoolSize.get() + 2 > MAX_WAL_BYTEBUFFER_NUM) {
+      while (walByteBufferPool.isEmpty() && currentWalPoolSize + 2 > MAX_WAL_BYTEBUFFER_NUM) {
         try {
           walByteBufferPool.wait();
         } catch (InterruptedException e) {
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -252,7 +252,7 @@
 
   private final Deque<ByteBuffer> walByteBufferPool = new LinkedList<>();
 
-  private int currentWalPoolSize = 0;
+  private AtomicInteger currentWalPoolSize = new AtomicInteger(0);
```
JackieTien97, line 255: Why change the type from `int` to `AtomicInteger`? `synchronized` key word has already ensure the concurrent safety

server/src/main/java/org/apache/iotdb/db/writelog/manager/MultiFileLogNodeManager.java
```diff
@@ -46,6 +46,12 @@
 public class MultiFileLogNodeManager implements WriteLogNodeManager, IService {
 
   private static final Logger logger = LoggerFactory.getLogger(MultiFileLogNodeManager.class);
+  // if OOM occurs when registering bytebuffer, getNode method will sleep awhile and then try again
+  private static final int REGISTER_BUFFER_SLEEP_INTERVAL = 20;
+  // if total sleep time exceeds this, getNode method will reject this write and throw original
+  // exception.
+  private static final int REGISTER_BUFFER_REJECT_THRESHOLD = 1_000;
```
JackieTien97, line 53: Make these two threshold configurable, and indicate theirs unit in the comment, like `ms` or `s`.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
```diff
@@ -298,9 +298,21 @@
...
+        } catch (OutOfMemoryError e) {
+          if (res[0] != null) {
+            MmapUtil.clean((MappedByteBuffer) res[0]);
+            currentWalPoolSize.getAndDecrement();
+          }
+          if (res[1] != null) {
+            MmapUtil.clean((MappedByteBuffer) res[1]);
+            currentWalPoolSize.getAndDecrement();
+          }
+          throw e;
```
JackieTien97, line 314: return null instead of throw this exception again. Exception handling is much more expensive. Since you have already catch exception here, you can return null here.

server/src/main/java/org/apache/iotdb/db/writelog/manager/MultiFileLogNodeManager.java
```diff
@@ -90,9 +96,39 @@ public WriteLogNode getNode(String identifier, Supplier<ByteBuffer[]> supplier)
...
+            } catch (InterruptedException interruptedException) {
+              e.addSuppressed(interruptedException);
+              nodeMap.remove(identifier);
+              throw e;
+            }
+            // sleep too long, throw exception
+            if (sleepTimeInMs >= REGISTER_BUFFER_REJECT_THRESHOLD) {
+              nodeMap.remove(identifier);
+              throw e;
+            }
```
JackieTien97, line 127: just judge whether the buffers is null, and no need to catch OOM here.

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
HeimingZ: I didn't see the synchronized key word. Already fixed.

server/src/main/java/org/apache/iotdb/db/writelog/manager/MultiFileLogNodeManager.java
HeimingZ: I add ms unit for them. But, I think it's bad to make these threshold params configurable, because they may make users confused.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessor.java
HeimingZ: fixed

server/src/main/java/org/apache/iotdb/db/writelog/manager/MultiFileLogNodeManager.java
HeimingZ: fixed

JackieTien97 (approved):

PR #4662, opened 2021-12-29
[IOTDB-2233] Grafana plugin: add `control` field for the `expression` panel

CloudWise-Lukemiao:
grafana-plugin add control and rest service add IT


SteveYurongSu (approved):
LGTM

PR #4663, opened 2021-12-29
[ISSUE-4651][grafana-plugin] Failed to display data when executing aggregation queries on non-existed timeseries

CloudWise-Lukemiao:
fixed issue-4651，Adding supports querying an un created time series


SteveYurongSu (approved):
LGTM

PR #4664, opened 2021-12-30
[IOTDB-2195] Control the concurrent query execution thread - Part 2

Alima777:
In this pr:
1. Complete QueryTaskThreadPool to control the concurrent query execution thread. Even if hundreds of clients execute query together, the thread number that execute query actully is limited by the pool size.
2. Introduce the `concurrentQueryThread` parameter to adjust the thread number that execute query statement, and rename `concurrent_sub_rawQuery_thread` to represent the read data task for raw data query.

On my machine（6 cores and 12 threads CPU), the threads of query are like:
Query threads are limited to 12, and read data threads of raw query are limited to 8. 

image


server/src/main/java/org/apache/iotdb/db/service/thrift/impl/TSServiceImpl.java
```diff
@@ -148,6 +151,81 @@
 /** Thrift RPC implementation at server side. */
 public class TSServiceImpl extends BasicServiceProvider implements TSIService.Iface {
 
+  protected class QueryTask implements Callable<TSExecuteStatementResp> {
```
JackieTien97, line 154: put it into an individual class file

JackieTien97 (changes requested):

PR #4673, opened 2021-12-30
[IOTDB-2234] fix NPE when last value cache is absent

cornmonster:
See: https://issues.apache.org/jira/browse/IOTDB-2234

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/query/last/ClusterLastQueryExecutor.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/query/last/ClusterLastQueryExecutor.java
@@ -33,6 +33,7 @@ import org.apache.iotdb.cluster.server.member.DataGroupMember;
 import org.apache.iotdb.cluster.server.member.MetaGroupMember;
 import org.apache.iotdb.db.concurrent.IoTDBThreadPoolFactory;
 import org.apache.iotdb.db.exception.StorageEngineException;
+import org.apache.iotdb.db.exception.metadata.MetadataException;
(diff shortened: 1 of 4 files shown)


cluster/src/main/java/org/apache/iotdb/cluster/query/last/ClusterLastQueryExecutor.java
```diff
@@ -105,29 +109,39 @@ public ClusterLastQueryExecutor(LastQueryPlan lastQueryPlan, MetaGroupMember met
...
-    for (TSDataType dataType : dataTypes) {
-      dataTypeOrdinals.add(dataType.ordinal());
+    Map<PartitionGroup, List<PartialPath>> groupToPaths = new HashMap<>();
+    Map<PartitionGroup, List<TSDataType>> groupToDataTypes = new HashMap<>();
+    try {
+      for (int i = 0; i < seriesPaths.size(); i++) {
+        PartialPath path = seriesPaths.get(i);
+        TSDataType dataType = dataTypes.get(i);
+        // TODO: doesn't work if time series is partitioned by time
+        PartitionGroup group = metaGroupMember.getPartitionTable().partitionByPathTime(path, 0);
```
ericpai, line 119: I think we should add a `if-else` to check whether timePartition is enabled:
- If it is enabled, it should pass the query to all data groups, as we don't know where the last value is in.
- If it is disabled, we can do an optimization that passing the query to the data group owing the timeseries.

ericpai (changes requested):

PR #4676, opened 2021-12-31
[IOTDB-2205] Optimize unaligned int/long read/write functions in BytesUtils

543202718:
注意到在现有的代码中，BytesUtils中非对齐的int/long的读写函数一次循环只能处理一个比特，效率较低，可以对其进行优化，使得一次循环可以处理一个字节。

该优化改动了下面的函数：
+ void longToBytes(long srcNum, byte[] result, int pos, int width)
+ void intToBytes(int srcNum, byte[] result, int pos, int width)
+ long bytesToLong(byte[] result, int pos, int width)
+ int bytesToInt(byte[] result, int pos, int width)

通过对用例的查找，发现该优化将会对使用TS_2DIFF编码的数据读写产生影响。

单元性能测试表明，该优化可以提高函数接近5倍的性能。总体性能测试表明，对于采用TS_2DIFF编码且不压缩的INT32/INT64数据，该优化可以减少约40%的读取时间开销。

--------------------------------------------------------------------------------------------------------------------------

In the current codes, the read and write functions of unaligned int/long in BytesUtils can only process one bit per loop, which is inefficient. It can be optimized to one byte per loop.

The following functions are optimized:
+ void longToBytes(long srcNum, byte[] result, int pos, int width)
+ void intToBytes(int srcNum, byte[] result, int pos, int width)
+ long bytesToLong(byte[] result, int pos, int width)
+ int bytesToInt(byte[] result, int pos, int width)

Through the use case search, the optimization influences the reading and writing of data encoded by TS_2DIFF.

In the unit performance test, the optimized functions are about 5x faster the current ones. In the overall test, about 40% time cost is reduced in the query of the uncompressed TS_2DIFF-coded INT32/INT64 data due to the optimization .

### DIFF
--- a/tsfile/src/main/java/org/apache/iotdb/tsfile/utils/BytesUtils.java
+++ b/tsfile/src/main/java/org/apache/iotdb/tsfile/utils/BytesUtils.java
@@ -79,7 +79,7 @@ public class BytesUtils {
    * convert an integer to a byte array which length is width, then copy this array to the parameter
    * result from pos.
    *
-   * @param srcNum input integer variable. All but the lowest {@code width} bits are 0. 
+   * @param srcNum input integer variable. All but the lowest {@code width} bits are 0.
    * @param result byte array to convert
    * @param pos start position
    * @param width bit-width
@@ -509,7 +509,7 @@ public class BytesUtils {
    * convert a long to a byte array which length is width, then copy this array to the parameter
    * result from pos.
    *
-   * @param srcNum input long variable. All but the lowest {@code width} bits are 0. 
+   * @param srcNum input long variable. All but the lowest {@code width} bits are 0.
    * @param result byte array to convert
    * @param pos start position
    * @param width bit-width
--- a/tsfile/src/main/java/org/apache/iotdb/tsfile/utils/BytesUtils.java
+++ b/tsfile/src/main/java/org/apache/iotdb/tsfile/utils/BytesUtils.java
@@ -79,15 +79,12 @@ public class BytesUtils {
    * convert an integer to a byte array which length is width, then copy this array to the parameter
    * result from pos.
    *
-   * @param srcNum input integer variable
+   * @param srcNum input integer variable. All but the lowest {@code width} bits are 0. 
    * @param result byte array to convert
    * @param pos start position
    * @param width bit-width
    */
   public static void intToBytes(int srcNum, byte[] result, int pos, int width) {
-    if (width < 32) {
-      srcNum = srcNum & ~(-1 << width);
-    }
     int cnt = pos & 0x07;
     int index = pos >> 3;
     try {
@@ -509,18 +506,15 @@ public class BytesUtils {
   }
 
   /**
-   * convert an long to a byte array which length is width, then copy this array to the parameter
+   * convert a long to a byte array which length is width, then copy this array to the parameter
    * result from pos.
    *
-   * @param srcNum input long variable
+   * @param srcNum input long variable. All but the lowest {@code width} bits are 0. 
    * @param result byte array to convert
    * @param pos start position
    * @param width bit-width
    */
   public static void longToBytes(long srcNum, byte[] result, int pos, int width) {
-    if (width < 64) {
-      srcNum = srcNum & ~(-1L << width);
-    }
     int cnt = pos & 0x07;
     int index = pos >> 3;
     try {
--- a/tsfile/src/main/java/org/apache/iotdb/tsfile/utils/BytesUtils.java
+++ b/tsfile/src/main/java/org/apache/iotdb/tsfile/utils/BytesUtils.java
@@ -85,20 +85,35 @@ public class BytesUtils {
    * @param width bit-width
    */
   public static void intToBytes(int srcNum, byte[] result, int pos, int width) {
-    int temp = 0;
-    for (int i = 0; i < width; i++) {
-      temp = (pos + width - 1 - i) / 8;
-      try {
-        result[temp] = setByteN(result[temp], pos + width - 1 - i, getIntN(srcNum, i));
-      } catch (Exception e) {
-        LOG.error(
-            "tsfile-common BytesUtils: cannot convert an integer {} to a byte array, "
-                + "pos {}, width {}",
-            srcNum,
-            pos,
-            width,
-            e);
+    if (width < 32) {
+      srcNum = srcNum & ~(-1 << width);
+    }
+    int cnt = pos % 8;
+    int index = pos / 8;
+    try {
+      while (width > 0) {
+        int m = width + cnt >= 8 ? 8 - cnt : width;
+        width -= m;
+        int mask = 1 << (8 - cnt);
+        cnt += m;
+        byte y = (byte) (srcNum >> width);
+        y = (byte) (y << (8 - cnt));
+        mask = ~(mask - (1 << (8 - cnt)));
+        result[index] = (byte) (result[index] & mask | y);
+        srcNum = srcNum & ~(-1 << width);
+        if (cnt == 8) {
+          index++;
+          cnt = 0;
+        }
       }
+    } catch (Exception e) {
+      LOG.error(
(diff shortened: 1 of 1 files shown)


tsfile/src/main/java/org/apache/iotdb/tsfile/utils/BytesUtils.java
```diff
@@ -495,19 +518,35 @@ public static boolean bytesToBool(byte[] b, int offset) {
...
-        result[temp] = setByteN(result[temp], pos + width - 1 - i, getLongN(srcNum, i));
-      } catch (Exception e) {
-        LOG.error(
-            "tsfile-common BytesUtils: cannot convert a long {} to a byte array, pos {}, width {}",
-            srcNum,
-            pos,
-            width,
-            e);
+    if (width < 64) {
+      srcNum = srcNum & ~(-1L << width);
```
xingtanzjr, line 522: We can consider to remove this `pre-processing` for `srcNum` because the `scrNum` will always be represented by `width` bits if this method is invoked

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/BytesUtils.java
```diff
@@ -495,19 +518,35 @@ public static boolean bytesToBool(byte[] b, int offset) {
...
+    try {
+      while (width > 0) {
+        int m = width + cnt >= 8 ? 8 - cnt : width;
+        width -= m;
+        int mask = 1 << (8 - cnt);
+        cnt += m;
+        byte y = (byte) (srcNum >> width);
+        y = (byte) (y << (8 - cnt));
+        mask = ~(mask - (1 << (8 - cnt)));
+        result[index] = (byte) (result[index] & mask | y);
```
xingtanzjr, line 535: Maybe `& mask` is unnecessary here because the bits between index `pos%8` - `pos%8+m` should always be `0`

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/BytesUtils.java
```diff
@@ -85,20 +85,35 @@ private BytesUtils() {}
...
-                + "pos {}, width {}",
-            srcNum,
-            pos,
-            width,
-            e);
+    if (width < 32) {
+      srcNum = srcNum & ~(-1 << width);
+    }
+    int cnt = pos % 8;
+    int index = pos / 8;
```
JackieTien97, line 92:
```suggestion
    int cnt = pos & 0x07;
    int index = pos >> 3;
```

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/BytesUtils.java
```diff
@@ -187,15 +202,23 @@ public static int bytesToInt(byte[] bytes, int offset) {
...
-
-    int value = 0;
-    int temp = 0;
-
-    for (int i = 0; i < width; i++) {
-      temp = (pos + width - 1 - i) / 8;
-      value = setIntN(value, i, getByteN(result[temp], pos + width - 1 - i));
+    int ret = 0;
+    int cnt = pos % 8;
+    int index = pos / 8;
```
JackieTien97, line 207:
```suggestion
    int cnt = pos & 0x07;
    int index = pos >> 3;
```

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/BytesUtils.java
```diff
@@ -495,19 +518,35 @@ public static boolean bytesToBool(byte[] b, int offset) {
...
-            "tsfile-common BytesUtils: cannot convert a long {} to a byte array, pos {}, width {}",
-            srcNum,
-            pos,
-            width,
-            e);
+    if (width < 64) {
+      srcNum = srcNum & ~(-1L << width);
+    }
+    int cnt = pos % 8;
+    int index = pos / 8;
```
JackieTien97, line 525:
```suggestion
    int cnt = pos & 0x07;
    int index = pos >> 3;
```

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/BytesUtils.java
```diff
@@ -550,13 +589,23 @@ public static long bytesToLong(byte[] byteNum, int len) {
...
    */
   public static long bytesToLong(byte[] result, int pos, int width) {
-    long value = 0;
-    int temp = 0;
-    for (int i = 0; i < width; i++) {
-      temp = (pos + width - 1 - i) / 8;
-      value = setLongN(value, i, getByteN(result[temp], pos + width - 1 - i));
+    long ret = 0;
+    int cnt = pos % 8;
+    int index = pos / 8;
```
JackieTien97, line 594:
```suggestion
    int cnt = pos & 0x07;
    int index = pos >> 3;
```

JackieTien97 (changes requested):
Great job! I wonder that how much improvement does this change bring to write performance？ Since you only mentioned the query performance improvements in your pr description.

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/BytesUtils.java
543202718: Thanks. But the original bits in `result` are not always 0 because the buffer is reused without clearing in DeltaEncoder.

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/BytesUtils.java
543202718: Thanks for your comment. The pre-processing is removed in the new commit.

543202718:
With the help of IoTDB-benchmark, I measured the performance improvement. Due to the difference in test data, the results are slightly different from those previously reported.

- The write throughput before improvement is 232,268 pt/s and 237,046 pt/s after improvement. The gap is within the fluctuation range because the write bottleneck is not the optimized function.
- The query throughput before improvement is 5,648,345 pt/s and 6,767,949 pt/s after improvement. Thers is a 20% improvement.

The modified config of IoTDB-benchmark is listed as follows while others are still the default.
```
LOOP=1000
SENSOR_NUMBER=4
CLIENT_NUMBER=1
POINT_STEP=1000
INSERT_DATATYPE_PROPORTION=0:1:1:1:1:0
COMPRESSOR=UNCOMPRESSED
ENCODING_INT32=TS_2DIFF
ENCODING_INT64=TS_2DIFF
ENCODING_FLOAT=TS_2DIFF
ENCODING_DOUBLE=TS_2DIFF
LINE_RATIO=0
SIN_RATIO=0
SQUARE_RATIO=0
RANDOM_RATIO=1
CONSTANT_RATIO=0
BATCH_SIZE_PER_WRITE=10000
QUERY_INTERVAL=100000000000

OPERATION_PROPORTION=1:0:0:0:0:0:0:0:0:0:0 (for write)
OPERATION_PROPORTION=0:0:1:0:0:0:0:0:0:0:0 (for query)
```

HTHou:
Hi, plz run `mvn spotless:apply` to fix the format issue.

543202718:
> Hi, plz run `mvn spotless:apply` to fix the format issue.

Thanks. The format issue is fixed in the new commit.

jihaitangguo:
Note: There was a bug in the code changes submitted here, [IOTDB-4633](https://issues.apache.org/jira/browse/IOTDB-4633), which has been fixed in #7669.

```
byte y = (byte) (srcNum >> width);
```

has been corrected to

```
byte y = (byte) (srcNum >>> width);
```

jihaitangguo:
Note: There was a bug in the code submitted here, [IOTDB-4633](https://issues.apache.org/jira/browse/IOTDB-4633), which has been fixed in #7669, where

```
byte y = (byte) (srcNum >> width);
```

has been corrected to

```
byte y = (byte) (srcNum >>> width);
```

PR #4690, opened 2022-01-01
[IOTDB-2237] coverage check encludes influxdb target

jun0315:
see https://issues.apache.org/jira/browse/IOTDB-2237


SteveYurongSu (approved):
LGTM

PR #4720, opened 2022-01-06
Add compatible interface with 0.12.x for creating template of flat measurements

bigreybear:
To be compatible with 0.12x, now you can create a schema template with flat measurements by specifying all measurements at once from a session.


session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -1918,6 +1919,34 @@ public void createSchemaTemplate(Template template)
     defaultSessionConnection.createSchemaTemplate(req);
   }
 
+  /**
+   * Create a template with flat measurements, not tree structured. Need to specify datatype,
+   * encoding and compressor of each measurement, and alignment of these measurements at once.
+   */
+  public void createSchemaTemplate(
```
SilverNarcissus, line 1926: We should explain the parameter of this method in java doc version.

SilverNarcissus (approved):
Brilliant work! Looking forward to use

session/src/main/java/org/apache/iotdb/session/Session.java
bigreybear: thanks! now there is a more detailed doc for the interface

session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
```diff
@@ -1380,6 +1380,43 @@ public void createSchemaTemplate(Template template)
...
+   * @param compressors compression type of each measurement in the template
+   * @param isAligned specify whether these flat measurements are aligned
+   */
+  public void createSchemaTemplate(
+      String templateName,
+      List<String> measurements,
+      List<TSDataType> dataTypes,
+      List<TSEncoding> encodings,
+      List<CompressionType> compressors,
+      boolean isAligned)
```
zyk990424, line 1400: Is there "isAligned" parameter in the interface of v0.12? If not, should we add a same interface as that of 0.12 and use default isAligned = false ?

bigreybear: Thanks! Now there is a completely compatible interface to creaete schema template with 0.12 in session

zyk990424 (approved):
LGTM~

qiaojialin (approved):

PR #4734, opened 2022-01-08
[IOTDB-2256][IOTDB-2318] Fix permission check bug while query with non timeseries & Fix doc

zyk990424:
### Problem
When authority module checks a query plan with empty path sets, which is caused by path pattern matching non timeseries, the system will return "user no permission", which should not happen.

### Solution
Add check on path nums before authority check. If empty path sets, authority check is unnecessary.

### Others
Fix some mistakes in UserGuide doc

### DIFF
--- a/integration/src/test/java/org/apache/iotdb/db/integration/auth/IoTDBAuthorizationIT.java
+++ b/integration/src/test/java/org/apache/iotdb/db/integration/auth/IoTDBAuthorizationIT.java
@@ -100,13 +100,8 @@ public class IoTDBAuthorizationIT {
         }
         assertTrue(caught);
 
-        caught = false;
-        try {
-          userStmt.execute("SELECT * from root.a");
-        } catch (SQLException e) {
-          caught = true;
-        }
-        assertTrue(caught);
+        // empty result timeseries set doesn't need authority check
+        userStmt.execute("SELECT * from root.a");
 
         caught = false;
         try {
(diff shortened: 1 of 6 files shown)


integration/src/test/java/org/apache/iotdb/db/integration/auth/IoTDBAuthorizationIT.java
```diff
@@ -1178,4 +1178,26 @@ public void testSelectUDTF() throws ClassNotFoundException, SQLException {
...
+      adminStatement.execute("GRANT USER a_application PRIVILEGES READ_TIMESERIES on root.ln;");
+
+      adminStatement.execute("INSERT INTO root.ln.wt01.wf01(time, s1) VALUES(1, 2)");
+    }
+
+    try (Connection userConnection =
+            DriverManager.getConnection(
+                Config.IOTDB_URL_PREFIX + "127.0.0.1:6667/", "a_application", "a_application");
+        Statement userStatement = userConnection.createStatement();
+        ResultSet resultSet = userStatement.executeQuery("SELECT * FROM root.ln.*")) {
```
jt2594838, line 1199: Does this correspond to "empty set"?

jt2594838 (changes requested):

jt2594838 (approved):

qiaojialin (approved):

PR #4762, opened 2022-01-11
[IOTDB-2142][IOTDB-2281][IOTDB-2286][IOTDB-2258] fix csv bugs and add a parm to use aligned interface

xuanronaldo:


cli/src/main/java/org/apache/iotdb/tool/ImportCsv.java
```diff
@@ -70,14 +71,17 @@
   private static final String FAILED_FILE_ARGS = "fd";
   private static final String FAILED_FILE_NAME = "failed file directory";
 
+  private static final String ALIGNED_ARGS = "aligned";
+  private static final String ALIGNED_Name = "use the aligned interface";
```
HTHou, line 75:
```suggestion
  private static final String ALIGNED_NAME = "use the aligned interface";
```

xuanronaldo: got it

HTHou (approved):

PR #4764, opened 2022-01-11
[Discussions-4763]Restructure export/import csv tool;

limeng32:
Hello, everyone. I recently used iotdb in the production environment and found that the import tool is almost unusable in the real scene. I studied this piece of code and found that the stream was not used correctly, so I refactored it, and I also added some other functions. What I do is as follows:

1. Optimize the implementation of import, so that the records in the backup file can be read into the memory, written to the database, and moved out of the memory line by line. The new tool no longer has any requirements on the size of the backup file;
2. Added compression options. It is a common behavior to compress while exporting. At the same time, importing from compressed files should also be supported. At present, I have implemented it in two ways: gzip and snappy;
3. There is corresponding progress feedback when importing/exporting. At present, I use feedback every time I export 10 million points or import 1 million points, so that no matter what type of data, I can have a suitable feedback rhythm;

Because the original code is almost completely unused, the new tools are named export-csv-new and import-csv-new, which coexist with existing tools;

### DIFF
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -80,6 +80,15 @@
             <artifactId>commons-csv</artifactId>
             <version>1.9.0</version>
         </dependency>
+        <dependency>
+            <groupId>joda-time</groupId>
+            <artifactId>joda-time</artifactId>
+            <version>2.6</version>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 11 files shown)


limeng32:
see https://github.com/apache/iotdb/pull/4774

PR #4768, opened 2022-01-11
[IOTDB-2276] Optimize aligned query with value filter to accelerate it and reduce its memory footprint

JackieTien97:
Previously, raw query with value filter won't use cache in where clause, in this pr, we use it.
Aggregate query with value filter has something wrong with the `MetaUtils.groupAlignedSeriesWithAggregations`, it will change the `selectedSeries` which will casue it won't use cache either


server/src/main/java/org/apache/iotdb/db/metadata/path/AlignedPath.java
```diff
@@ -201,6 +218,7 @@ public PartialPath copy() {
     result.fullPath = fullPath;
     result.device = device;
     result.measurementList = new ArrayList<>(measurementList);
```
Alima777, line 220: Add ` result.isUnderAlignedEntity = isUnderAlignedEntity;`

Alima777: MeasurementPath's copy() method need to be modified too.

JackieTien97: `isUnderAlignedEntity` is a filed of `MeasurementPath`, not `AlignedPath`.

JackieTien97: > MeasurementPath's copy() method need to be modified too.

done.

Alima777 (approved):
LGTM

PR #4778, opened 2022-01-11
[IOTDB-2313] GC overhead limit when use "select * from root.**"

CloudWise-Lukemiao:
fixed GC overhead limit when use "select * from root.**"


server/src/main/java/org/apache/iotdb/db/conf/rest/IoTDBRestServiceConfig.java
```diff
@@ -140,4 +142,12 @@ public int getCacheInitNum() {
   public void setCacheInitNum(int cacheInitNum) {
     this.cacheInitNum = cacheInitNum;
   }
+
+  public int getRestQueryFetchSize() {
+    return restQueryFetchSize;
```
SteveYurongSu, line 147: Does it seem that we need to invoke the method in Descriptor?

server/src/main/java/org/apache/iotdb/db/protocol/rest/impl/GrafanaApiServiceImpl.java
```diff
@@ -141,6 +144,15 @@ public Response expression(ExpressionRequest expressionRequest, SecurityContext
 
       PhysicalPlan physicalPlan = basicServiceProvider.getPlanner().parseSQLToPhysicalPlan(sql);
 
+      if (physicalPlan instanceof QueryPlan) {
```
SteveYurongSu, line 147: We'd better not add the limit here. This may cause unexpected behavior when using the grafana plugin.

SteveYurongSu (approved):

PR #4782, opened 2022-01-12
[IOTDB-2315] Modify some problems of metric

stormbroken:
1. Fix some error in metric doc model.
2. Fix some description of iotdb-metric.yml
3. Modify config of metric model into hot load way.
4. add `preDefinedMetrics` in iotdb-metric.yml


metrics/dropwizard-metrics/src/test/java/org/apache/iotdb/metrics/dropwizard/DropwizardMetricTestPlan.java
```diff
@@ -52,7 +53,6 @@ private static void test(Integer metric, Integer tag) {
   }
 
   public static void main(String[] args) {
```
jixuan1989, line 55: where a main() method here?

stormbroken: This main method is only used for performance and memory test between dropwizard and micrometer, so whether should I need to remove it?

stormbroken: This main method is only used for performance and memory test between dropwizard and micrometer, so whether should I need to remove it?

xinzhongtianxia (approved):
LGTM

PR #4783, opened 2022-01-12
[IOTDB-2283][IOTDB-2310][IOTDB-2311][IOTDB-2372][IOTDB-2313][IOTDB-2362] REST: metadata query interface

CloudWise-Lukemiao:
IOTDB-2283 REST: metadata query interface

It also fixes:
IOTDB-2310 [grafana-plugin] NPE, Using aggregate queries with "group by level=*"
IOTDB-2311 [grafana-plugin] NPE, Using queries with "align by device"
IOTDB-2362 [rest api] Use noQuery to execute "count", return "SHOW is not supported"
IOTDB-2372 [rest api] disable align not return value, no error

And it is another patch of IOTDB-2313.


server/src/main/java/org/apache/iotdb/db/protocol/rest/handler/RequestValidationHandler.java
```diff
@@ -29,6 +29,9 @@ private RequestValidationHandler() {}
 
   public static void validateSQL(SQL sql) {
     Objects.requireNonNull(sql.getSql(), "sql should not be null");
+    if (sql.getRowLimit() != null && sql.getRowLimit() <= 0) {
+      Objects.requireNonNull(null, "rowLimit should be greater than or equal to 0");
```
SteveYurongSu, line 33: There are many ways to throw a validation error as a RuntimeException.

PR #4790, opened 2022-01-12
[IOTDB-2354] row number is not as expected when using nested expressions containing count() function in 'group by time' clause

Plutooooooo:
## Description

see [IOTDB-2354](https://issues.apache.org/jira/browse/IOTDB-2354) row number is not as expected when using nested expressions containing count() function in 'group by time' clause


ericpai:
LGTM, but the CI of some basic unit tests are failed due to other reasons.

ericpai (approved):

PR #4792, opened 2022-01-12
[IOTDB-2329] Deploy IoTDB on Kubernetes with Helm

xinzhongtianxia:
I submit a PR to support deploying iotdb both standalone and cluster on kubernetes with helm.
1. It is not a k8s operator, and does not support scale out. In the future, an operator is needed, probably after the completion of the new cluster architecture. For now, helm is enough.
2. In k8s, a pod's ip will change after restarting, so we should use pod's FQDN as seed_nodes. So I removed the logical of replacing hostname with ip when starting iotdb-cluster.
3. Added EN/ZH docs for iotdb-on-kubernetes

jira: https://issues.apache.org/jira/browse/IOTDB-2329


jixuan1989:
Hi, it is better to copy  "iotdb-engine.properties" and "iotdb-cluster.properties" files to `kubernetes/helm/cluster/conf` in maven's `generate-resources` phase, rather than upload duplicated files directly.

Example codes: see `client-py/pom.xml`

jixuan1989 (changes requested):
Hi, it is better to copy "iotdb-engine.properties" and "iotdb-cluster.properties" files to kubernetes/helm/cluster/conf in maven's generate-resources phase, rather than upload duplicated files directly.

Example codes: see client-py/pom.xml

xinzhongtianxia:
> Hi, it is better to copy "iotdb-engine.properties" and "iotdb-cluster.properties" files to kubernetes/helm/cluster/conf in maven's generate-resources phase, rather than upload duplicated files directly.
> 
> Example codes: see client-py/pom.xml

Done.

PR #4793, opened 2022-01-12
[IOTDB-2187] Update code of pulsar example with session API

Alibaba-HZY:
Update code of pulsar example with session api


SteveYurongSu (approved):
LGTM. Thanks for your contribution!

PR #4809, opened 2022-01-14
[IOTDB-2367] Linear fill should log warning for unsupported data type

CRZbulabula:


server/src/main/java/org/apache/iotdb/db/query/dataset/groupby/GroupByFillDataSet.java
```diff
@@ -174,13 +177,18 @@ public RowRecord nextWithoutConstraint() throws IOException {
...
     } catch (QueryProcessException e) {
-      logger.warn("group by fill has an exception while sliding: ", e);
+      logger.error("group by fill has an exception while sliding: ", e);
     }
 
     return record;
   }
 
   private void fillRecord(int index, RowRecord record) throws IOException {
+    if (unsupportedFillMethod[index]) {
```
ericpai, line 187: It's better to add test cases to cover unsatisfied data types with `GROUP BY FILL`.

server/src/main/java/org/apache/iotdb/db/query/executor/FillQueryExecutor.java
```diff
@@ -108,6 +113,15 @@ public QueryDataSet execute(QueryContext context)
 
         IFill fill = fillExecutors[i];
 
+        if (fill instanceof LinearFill
```
ericpai, line 116: It's better to add test cases to cover unsatisfied data types with `FILL`.

ericpai (changes requested):

server/src/main/java/org/apache/iotdb/db/query/dataset/groupby/GroupByFillDataSet.java
CRZbulabula: I've added tests in last commit

server/src/main/java/org/apache/iotdb/db/query/executor/FillQueryExecutor.java
CRZbulabula: I've added tests in last commit

ericpai (approved):

PR #4815, opened 2022-01-14
[IOTDB-2394] Fixed temporary network unavailability caused by frequent closed connections

LebronAl:
see [jira](https://issues.apache.org/jira/browse/IOTDB-2394)

image

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/client/ClientPoolFactory.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/client/ClientPoolFactory.java
@@ -40,6 +40,7 @@ public class ClientPoolFactory {
 
   protected long waitClientTimeoutMS;
   protected int maxConnectionForEachNode;
+  protected int maxIdleConnectionForEachNode;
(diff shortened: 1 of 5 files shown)


cluster/src/main/java/org/apache/iotdb/cluster/client/ClientPoolFactory.java
```diff
@@ -54,6 +54,7 @@ public ClientPoolFactory() {
             : new TBinaryProtocol.Factory();
     poolConfig = new GenericKeyedObjectPoolConfig();
     poolConfig.setMaxTotalPerKey(maxConnectionForEachNode);
+    poolConfig.setMaxIdlePerKey(maxConnectionForEachNode / 2);
```
wangchao316, line 57: Good.
Can we consider setting this parameter in the configuration file?

LebronAl: done.

wangchao316 (approved):

PR #4825, opened 2022-01-14
[IOTDB-2392] Memory control of raft log in cluster 

MrQuansy:
See jira: https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-2392.


jt2594838 (approved):
Good job. But the rejection is based on the number of entries instead of their memory footprint. Could you please add some todos that we use precise memory calculation for rejection?

cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
```diff
@@ -2034,6 +2073,10 @@ private long appendEntries(
     }
 
     synchronized (logManager) {
+      if ((logManager.getCommitLogIndex() - logManager.getMaxHaveAppliedCommitIndex())
+          > config.getUnAppliedRaftLogNumForRejectThreshold()) {
+        return Response.RESPONSE_TOO_BUSY;
+      }
```
jt2594838, line 2079: Adding some wait time here may improve robustness.

MrQuansy: Ok, I add some statements to wait some time before rejection.

jt2594838 (approved):

LebronAl (approved):
LGTM

PR #4829, opened 2022-01-14
[IOTDB-2355] Wildcard adaption

zyk990424:
### Problem

** and * in SQL are incompatible with v0.12

### Solution

Add isPrefixMatch functions to process requests from client of v0.12
1. add isPrefixMatch param in Operator and PhysicalPlan which is set based on the clientVersion from client. 
2. when parsing logical plan, remove wildcard based on the isPrefixMatch param in Operator and pass this param to the generated PhysicalPlan
3. modify related functions in PlanExecutor and MManager via adding isPrefixMatch param in interfaces

notice that
1. In standalone mode, the adaption is implemented via adding interfaces with isPrefixMatch param in MManager, which has no impact on operation performance.
2. To adapt deletion function, since the path will be stored in Mods file and be processed by Compaction module, the prefixMatch will be parsed as two deleteion, for example, "delete from root.sg.d" will be operated as "delete from root.sg.d" and "delete from root.sg.d.**"
3. In cluster mode, to avoid modify communication interfaces in cluster module, the adaption is implemented same as that on deletion, which means double process on one prefix path.


SilverNarcissus (approved):
Crystal adaption. Clear and easy to understand. Look good to me~

qiaojialin (approved):

qiaojialin (approved):

PR #4830, opened 2022-01-14
[IOTDB-2403] The parameter "ascending" is not set correctly when getting QueryDataSource

liuminghui233:
The parameter "ascending" is not set correctly when getting QueryDataSource.


server/src/main/java/org/apache/iotdb/db/metadata/lastCache/LastCacheManager.java
```diff
@@ -222,7 +222,7 @@ public static long getLastTimeStamp(IMeasurementMNode node, QueryContext queryCo
       try {
         QueryDataSource dataSource =
             QueryResourceManager.getInstance()
-                .getQueryDataSource(node.getPartialPath(), queryContext, null);
+                .getQueryDataSource(node.getPartialPath(), queryContext, null, false);
```
JackieTien97, line 225: Add a comment, `true` or `false` both ok here, because LastPointerReader will do itself sort logic instead of depending on `fillOrderIndex`.

server/src/main/java/org/apache/iotdb/db/query/executor/fill/LinearFill.java
```diff
@@ -153,7 +153,8 @@ public TimeValuePair getFillResult()
   protected TimeValuePair calculatePrecedingPoint()
       throws QueryProcessException, StorageEngineException, IOException {
     QueryDataSource dataSource =
-        QueryResourceManager.getInstance().getQueryDataSource(seriesPath, context, beforeFilter);
+        QueryResourceManager.getInstance()
+            .getQueryDataSource(seriesPath, context, beforeFilter, false);
```
JackieTien97, line 157: The same as above

server/src/main/java/org/apache/iotdb/db/query/executor/fill/PreviousFill.java
```diff
@@ -115,7 +115,8 @@ public void configureFill(
   public TimeValuePair getFillResult()
       throws IOException, QueryProcessException, StorageEngineException {
     QueryDataSource dataSource =
-        QueryResourceManager.getInstance().getQueryDataSource(seriesPath, context, timeFilter);
+        QueryResourceManager.getInstance()
+            .getQueryDataSource(seriesPath, context, timeFilter, false);
```
JackieTien97, line 119: Same as above

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/query/executor/fill/LinearFill.java
liuminghui233: fixed

server/src/main/java/org/apache/iotdb/db/query/executor/fill/PreviousFill.java
liuminghui233: fixed

PR #4847, opened 2022-01-15
[IOTDB-2406] Incorrect docs; Incorrect error-log

Cpaulyz:
1. Fix incorrect docs

   * " -> '
   * use absolute path as example 

2. Use more accurate error messages

   * if target file doesn't exist

     ```
     IoTDB> load '/root.sg1/1642237791044-1-0-0.tsfile'
     Msg: 411: File path '/root.sg1/1642237791044-1-0-0.tsfile' doesn't exists.
     
     IoTDB> unload '/Users/chenyanze/projects/JavaProjects/iotdb/iotdb/data/data/root.sg1/0/0' '/'
     Msg: 411: File '/Users/chenyanze/projects/JavaProjects/iotdb/iotdb/data/data/root.sg1/0/0' doesn't exist.
     ```

   * if remove/unload file exists but not an internal tsfile

     ```
     IoTDB> unload '/Users/chenyanze/projects/JavaProjects/iotdb/iotdb/data/data/sequence/root.sg1/0/0' '/'
     Msg: 411: Cannot unload file '/Users/chenyanze/projects/JavaProjects/iotdb/iotdb/data/data/sequence/root.sg1/0/0' to target directory '/' because /Users/chenyanze/projects/JavaProjects/iotdb/iotdb/data/data/sequence/root.sg1/0/0 is not a legal path, because it's not an internal tsfile.
     ```

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/StorageEngine.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/StorageEngine.java
@@ -950,8 +950,19 @@ public class StorageEngine implements IService {
    * @param file internal file
    * @return sg name
    */
-  public String getSgByEngineFile(File file) {
-    return file.getParentFile().getParentFile().getParentFile().getName();
+  public String getSgByEngineFile(File file) throws IllegalPathException {
+    File dataDir =
+        file.getParentFile().getParentFile().getParentFile().getParentFile().getParentFile();
+    if (dataDir.exists()) {
+      String[] dataDirs = IoTDBDescriptor.getInstance().getConfig().getDataDirs();
+      for (String dir : dataDirs) {
+        File f = new File(dir);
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/apache/iotdb/db/engine/StorageEngine.java
```diff
@@ -950,8 +950,23 @@ public boolean unloadTsfile(File tsfileToBeUnloaded, File targetDir)
    * @param file internal file
    * @return sg name
    */
-  public String getSgByEngineFile(File file) {
-    return file.getParentFile().getParentFile().getParentFile().getName();
+  public String getSgByEngineFile(File file, boolean needCheck) throws IllegalPathException {
```
qiaojialin, line 953: add javadoc for needCheck

like: check if the tsfile is an internal TsFile, if you make sure it is inside, no need to check

qiaojialin (approved):

PR #4850, opened 2022-01-17
[IOTDB-2391] Use blocking queue and semaphore to replace timed compaction submission thread

THUMarkLau:


server/src/main/java/org/apache/iotdb/db/engine/compaction/task/AbstractCompactionTask.java
```diff
@@ -71,6 +71,7 @@ public Void call() throws Exception {
         CompactionTaskManager.getInstance().removeRunningTaskFromList(this);
       }
       this.currentTaskNum.decrementAndGet();
+      CompactionTaskManager.semaphore.release();
```
xingtanzjr, line 74: Consider to use `callback` function here to decrease circle dependencies.

server/src/main/java/org/apache/iotdb/db/engine/compaction/CompactionTaskManager.java
```diff
@@ -283,7 +272,7 @@ public synchronized void removeRunningTaskFromList(AbstractCompactionTask task)
    *
    * @throws RejectedExecutionException
    */
-  public synchronized void submitTask(
+  public void submitTask(
       String fullStorageGroupName, long timePartition, Callable<Void> compactionMergeTask)
       throws RejectedExecutionException {
     if (taskExecutionPool != null && !taskExecutionPool.isTerminated()) {
```
xingtanzjr, line 278: If this condition is `false`, the sub-thread won't be started, which will lead to the `semaphore` cannot be `released`. We need to handle the scenario here

server/src/main/java/org/apache/iotdb/db/engine/compaction/CompactionTaskManager.java
```diff
@@ -343,8 +332,6 @@ public void restart() {
               IoTDBThreadPoolFactory.newScheduledThreadPool(
                   IoTDBDescriptor.getInstance().getConfig().getConcurrentCompactionThread(),
                   ThreadName.COMPACTION_SERVICE.getName());
-      this.compactionTaskSubmissionThreadPool =
-          IoTDBThreadPoolFactory.newScheduledThreadPool(1, ThreadName.COMPACTION_SERVICE.getName());
```
xingtanzjr, line 347: Even if this method is only used by Tests, I am not sure whether we need to handle the `semaphore` here because the `semaphore` is also initialized in `start()`.

server/src/main/java/org/apache/iotdb/db/engine/compaction/CompactionTaskManager.java
```diff
@@ -224,26 +207,32 @@ public synchronized boolean addTaskToWaitingQueue(AbstractCompactionTask compact
    * This method will submit task cached in queue with most priority to execution thread pool if
    * there is available thread.
    */
-  public synchronized void submitTaskFromTaskQueue() {
-    while (currentTaskNum.get()
-            < IoTDBDescriptor.getInstance().getConfig().getConcurrentCompactionThread()
-        && candidateCompactionTaskQueue.size() > 0) {
-      AbstractCompactionTask task = candidateCompactionTaskQueue.poll();
+  public void submitTaskFromTaskQueue() {
```
xingtanzjr, line 210: It seems that this method is only invoked in Tests

server/src/main/java/org/apache/iotdb/db/utils/datastructure/FixedPriorityBlockingQueue.java
```diff
@@ -0,0 +1,134 @@
...
+  private Condition notEmpty = lock.newCondition();
+
+  public FixedPriorityBlockingQueue(int maxSize, Comparator<T> comparator) {
+    this.maxSize = maxSize;
+    this.comparator = comparator;
+    this.queue = MinMaxPriorityQueue.orderedBy(comparator).maximumSize(maxSize).create();
+  }
+
+  public void put(T element) throws InterruptedException {
+    final ReentrantLock lock = this.lock;
```
xingtanzjr, line 47: Why do we need a local final var here

THUMarkLau: The BlockingQueue source code uses locks like this, so I think we should stick with it

server/src/main/java/org/apache/iotdb/db/engine/compaction/CompactionTaskManager.java
THUMarkLau: A thread is used in the Start function of CompactionTaskManager to execute this function

xingtanzjr (approved):

PR #4854, opened 2022-01-17
[IOTDB-2412] Fetch Result Task is not taken account into query execution control

liuminghui233:
Now we control the query execution thread, but we forgot the fecthResult interface, which means we cannot take control of that. When the client fetches the second batch, the query threads will increase to the connection numbers.

Related PR: #4660 & #4664


server/src/main/java/org/apache/iotdb/db/service/thrift/impl/TSServiceImpl.java
```diff
@@ -946,47 +997,10 @@ public TSFetchResultsResp fetchResults(TSFetchResultsReq req) {
             RpcUtils.getStatus(TSStatusCode.EXECUTE_STATEMENT_ERROR, "Has not executed query"));
       }
 
-      serviceProvider.genQueryContext(
-          req.queryId, false, System.currentTimeMillis(), req.statement, req.timeout);
```
liuminghui233, line 950: I see the QueryContext generated here. So I removed it. Please confirm it~

PR #4863, opened 2022-01-17
[IOTDB-2414][IOTDB-2419]Fix partial insert failure exception message && Fix three level aligned path query bug

zyk990424:
### Fix partial insert failure exception message
see IOTDB-2414

solution: make each insertRowPlan in insertRowsPlan not share measurements array

### Fix three level aligned path query bug
see IOTDB-2419

solution: override the getDevicePath in AlignedPath

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/qp/logical/crud/InsertOperator.java
+++ b/server/src/main/java/org/apache/iotdb/db/qp/logical/crud/InsertOperator.java
@@ -110,7 +110,7 @@ public class InsertOperator extends Operator {
                 measurementsNum, valueLists.get(i).length));
       }
       InsertRowPlan insertRowPlan =
-          new InsertRowPlan(device, times[i], measurementList, valueLists.get(i));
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/apache/iotdb/db/qp/logical/crud/InsertOperator.java
```diff
@@ -110,7 +110,7 @@ public PhysicalPlan generatePhysicalPlan(PhysicalGenerator generator)
                 measurementsNum, valueLists.get(i).length));
       }
       InsertRowPlan insertRowPlan =
-          new InsertRowPlan(device, times[i], measurementList, valueLists.get(i));
+          new InsertRowPlan(device, times[i], measurementList.clone(), valueLists.get(i));
```
wangchao316, line 113: Does clone affect write performance?

qiaojialin: This is a quick-fix version... This will be fully refactored in the next version.

PR #4878, opened 2022-01-18
[IOTDB-2428][monitor] remove monitorStat

stormbroken:
remove monitorStat


server/src/main/java/org/apache/iotdb/db/engine/StorageEngine.java
```diff
@@ -33,14 +33,7 @@
...
 import org.apache.iotdb.db.engine.storagegroup.virtualSg.StorageGroupManager;
-import org.apache.iotdb.db.exception.BatchProcessException;
-import org.apache.iotdb.db.exception.LoadFileException;
-import org.apache.iotdb.db.exception.ShutdownException;
-import org.apache.iotdb.db.exception.StorageEngineException;
-import org.apache.iotdb.db.exception.StorageGroupProcessorException;
-import org.apache.iotdb.db.exception.TsFileProcessorException;
-import org.apache.iotdb.db.exception.WriteProcessException;
-import org.apache.iotdb.db.exception.WriteProcessRejectException;
+import org.apache.iotdb.db.exception.*;
```
SteveYurongSu, line 36: Actually we don't use * :)

server/src/main/java/org/apache/iotdb/db/engine/StorageEngine.java
```diff
@@ -601,16 +579,6 @@ public void insert(InsertRowPlan insertRowPlan) throws StorageEngineException, M
 
     try {
       virtualStorageGroupProcessor.insert(insertRowPlan);
-      if (config.isEnableStatMonitor()) {
```
SteveYurongSu, line 604: It seems that we have to remove the parameter in the config file.

stormbroken: ok I remove it.

SteveYurongSu (approved):

PR #4908, opened 2022-01-19
[IOTDB-2448] unload doesn't work correctly: it's not an internal tsfile

Cpaulyz:
* It caused by the wrong way of files' comparison in `getSgByEngineFile`.

* Use `Files.isSameFile()` instead of comparing strings of absolute paths.

### DIFF
--- a/integration/src/test/java/org/apache/iotdb/db/integration/IoTDBLoadExternalTsfileIT.java
+++ b/integration/src/test/java/org/apache/iotdb/db/integration/IoTDBLoadExternalTsfileIT.java
@@ -199,7 +199,9 @@ public class IoTDBLoadExternalTsfileIT {
         tmpDir.mkdirs();
       }
       for (TsFileResource resource : resources) {
-        statement.execute(String.format("unload '%s' '%s'", resource.getTsFilePath(), tmpDir));
+        // test unload using relative path
+        statement.execute(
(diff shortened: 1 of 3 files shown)


Cpaulyz:
The original implementation will be incorrect if file path or dataDirs' path contains `.` or `..`
The original implementation will be incorrect if file path or dataDirs' path contains `.` or `..`

For example, if IotDB is started by `sh start-server.sh`，it will receive parameter `IOTDB_HOME` and add  `IOTDB_HOME` in front ofthe relative path of the configuration file. e.g. `data/data `->`sbin/../data/data `

Then, if user execute `unload '/Users/chenyanze/projects/JavaProjects/iotdb/iotdb/server/target/iotdb-server-0.13.0-SNAPSHOT/data/data/sequence/root.sg1/0/0/1642605548103-1-0-1.tsfile' '/'`

* when `getSgByEngineFile`, we getDataDirs from config and comprare to user params' parent dir of remove/unload. 

* user params' dir : `/Users/chenyanze/projects/JavaProjects/iotdb/iotdb/server/target/iotdb-server-0.13.0-SNAPSHOT/data/data`
* IoTDB engine data dir: `/Users/chenyanze/projects/JavaProjects/iotdb/iotdb/server/target/iotdb-server-0.13.0-SNAPSHOT/sbin/../data/data`

Therefore, we should use `Files.isSameFile()` instead of comparing strings of absolute paths.

qiaojialin (approved):

PR #4923, opened 2022-01-20
rest and grafana plugin add login method

CloudWise-Lukemiao:
rest and grafana plugin add login method


server/src/main/java/org/apache/iotdb/db/protocol/rest/impl/GrafanaApiServiceImpl.java
```diff
@@ -168,4 +168,14 @@ public Response expression(ExpressionRequest expressionRequest, SecurityContext
       return Response.ok().entity(ExceptionHandler.tryCatchException(e)).build();
     }
   }
+
+  @Override
+  public Response login(SecurityContext securityContext) throws NotFoundException {
+    return Response.ok()
+        .entity(
```
SteveYurongSu, line 175: No auth operation?

SteveYurongSu (approved):

PR #4928, opened 2022-01-20
[IOTDB-2457] Write is blocked after set time_index_level=FILE_TIME_INDEX

chengjianyun:
JIRA link: https://issues.apache.org/jira/browse/IOTDB-2457

Data insert is blocked if set time_index_level=FILE_TIME_INDEX


JackieTien97:
It's a mistake, we should not let users to set `FILE_TIME_INDEX`, we only support automatically degraded from `DEVICE_TIME_INDEX` to `FILE_TIME_INDEX`.

chengjianyun:
> It's a mistake, we should not let users to set `FILE_TIME_INDEX`, we only support automatically degraded from `DEVICE_TIME_INDEX` to `FILE_TIME_INDEX`.

How could the one have set it in 0.12?  And current implementation doesn't backward compatiable so the upgrade will cause reading resoure fail.

jixuan1989:
I see @HTHou closed this issue,  do we have any solution about this?

PR #4938, opened 2022-01-21
[IOTDB-2453] Set the default compaction priority to `BALANCE`

THUMarkLau:


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
```diff
@@ -101,7 +101,8 @@
 
   protected volatile boolean closed = false;
   private volatile boolean deleted = false;
-  volatile boolean isMerging = false;
+  volatile boolean isCompacting = false;
+  volatile boolean compactionCandidate = false;
```
qiaojialin, line 105: merge to one status

server/src/test/java/org/apache/iotdb/db/engine/compaction/CompactionTaskComparatorTest.java
```diff
@@ -1,7 +1,7 @@
 /*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
- * distributed with this work for additional information
+ * distributed with this work for putitional information
```
qiaojialin, line 4:
```suggestion
 * distributed with this work for additional information
```

PR #4946, opened 2022-01-21
[IOTDB-2465][imple] fix implement of snapshot of dropwizard to UniformReservoir.

stormbroken:
When run CI, there are following problems:

image

It's because the implementation of snapshot for dropwizard use wighted one as default, so I change to uniform one now.


SteveYurongSu (approved):
LGTM

PR #4949, opened 2022-01-21
[IOTDB-2425][env] rename IOTDB_CLI_HOME to IOTDB_HOME

stormbroken:
See: https://issues.apache.org/jira/browse/IOTDB-2425


HTHou (approved):
Already tested on Windows and mac!

PR #4960, opened 2022-01-24
Make FileTimeIndex directly return true while doing containsDevice() 

JackieTien97:
* While one TsFileResouece's ITimeIndex degrade from DeviceTimeIndex to FileTimeIndex, each time a query arrived, it will read from tsfile to judge whether this tsfile contains current device. However, this is totally unnecessary. We can simply return true here, and then do this judgement in the following query process, while getting the TimeseriesMetadata, it can also use the bloomfilter in tsfile footer to accelerate this process.

* Encapsulate the process of opening and releasing file handles into `FileReaderManager`.


chengjianyun (approved):
LGTM

qiaojialin (approved):

PR #4963, opened 2022-01-24
Avoid flush stream in getPosition() to improve flush performance

chengjianyun:
current getPosition() will trigger flush which cause the bufferStream doesn't take effect as getPosition is triggered in too many times.

Fix the issue to help improve flush performance.

### DIFF
--- a/tsfile/src/main/java/org/apache/iotdb/tsfile/write/writer/LocalTsFileOutput.java
+++ b/tsfile/src/main/java/org/apache/iotdb/tsfile/write/writer/LocalTsFileOutput.java
@@ -19,8 +19,6 @@
 package org.apache.iotdb.tsfile.write.writer;
 
 import java.io.BufferedOutputStream;
-import java.io.File;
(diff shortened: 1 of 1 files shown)


qiaojialin (approved):

JackieTien97 (changes requested):
It seems that `synchronized` is unnecessary. Will the `LocalTsFileOutput` be used in multi-thread case?

chengjianyun:
> It seems that `synchronized` is unnecessary. Will the `LocalTsFileOutput` be used in multi-thread case?

I think the class is a stream wrapper and a low-level class. It's better to make it multi-thread safe just as the other stream implementations do.

qiaojialin:
> > It seems that `synchronized` is unnecessary. Will the `LocalTsFileOutput` be used in multi-thread case?
> 
> The class is a low-level stream wrapper. It's better to make it multi-thread safe just as the other stream implementations do. How to you think?

+1

JackieTien97 (approved):

PR #4964, opened 2022-01-24
Apply string intern for device String

chengjianyun:
The implementation of cacheString pool has the memory leak risk. We can use string intern instead of do it by ourself.


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
```diff
@@ -54,13 +54,7 @@
...
 import java.nio.file.Files;
 import java.nio.file.Paths;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Random;
-import java.util.Set;
+import java.util.*;
```
wangchao316, line 57: Good
does not import *

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/VirtualStorageGroupProcessor.java
```diff
@@ -45,14 +45,7 @@
...
 import org.apache.iotdb.db.engine.version.VersionController;
-import org.apache.iotdb.db.exception.BatchProcessException;
-import org.apache.iotdb.db.exception.DiskSpaceInsufficientException;
-import org.apache.iotdb.db.exception.LoadFileException;
-import org.apache.iotdb.db.exception.StorageGroupProcessorException;
-import org.apache.iotdb.db.exception.TriggerExecutionException;
-import org.apache.iotdb.db.exception.TsFileProcessorException;
-import org.apache.iotdb.db.exception.WriteProcessException;
-import org.apache.iotdb.db.exception.WriteProcessRejectException;
+import org.apache.iotdb.db.exception.*;
```
wangchao316, line 48: the same as above...

wangchao316 (changes requested):

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
chengjianyun: Done.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/VirtualStorageGroupProcessor.java
chengjianyun: Done.

wangchao316 (approved):
LGTM

HTHou:
Hi, is there any detailed reason about the memory leak risk?

chengjianyun:
> Hi, is there any detailed reason about the memory leak risk?

`cachedDevicePool` never remove its key which is not reasonable. For example, if we delete some storage groups or timeseries, some string would be never used but they are always kept in `cachedDevicePool` which cause memory leak.

For string intern, JVM takes care of remove the useless string objects.

Alima777:
> For example, if we delete some storage groups or timeseries, some string would be never used but they are always kept in `cachedDevicePool` which cause memory leak.

Hi, it's true that we can't handle the useless string if we delete some devices. But the intern() method is not that fast[1][2] since the purpose of it is to save memory. By comparison, `concurrentHashMap` has a good performance. 
I believe that the situation about deleting timeseries or storage group is relatively few. And we use `deviceIndexMap` when we initialize TsFileResource, when restarting or ingesting data, the performance seems more important. Maybe we need to make a balance here. Or we can delete useless string from `cachedStringPool` manually when time series is deleted.
What do you think?


[1] https://stackoverflow.com/questions/56203500/java-string-intern-use-hashtable-instead-of-concurrenthashmap
[2] https://www.fatalerrors.org/a/jvm-series-string.intern-performance-of.html

chengjianyun:
> > For example, if we delete some storage groups or timeseries, some string would be never used but they are always kept in `cachedDevicePool` which cause memory leak.
> 
> Hi, it's true that we can't handle the useless string now if we delete some devices. But the `intern()` method is not that fast[1][2] since the purpose of it is to save memory. By comparison, `concurrentHashMap` has a good performance. I believe that the situation about deleting timeseries or storage group is relatively few. And we use `deviceIndexMap` when we initialize TsFileResource, when restarting or ingesting data, the performance seems more important. Maybe we need to make a balance here. Or we can delete useless string from `cachedStringPool` manually when time series is deleted instead of just using `intern()`. What do you think?
> 
> [1] https://stackoverflow.com/questions/56203500/java-string-intern-use-hashtable-instead-of-concurrenthashmap [2] https://www.fatalerrors.org/a/jvm-series-string.intern-performance-of.html

@Alima777, about performance of  `String.intern()`, you can also refer  [this blog](http://java-performance.info/string-intern-in-java-6-7-8/). I have verified the conclusion and got similar result on JDK8. So, it's better to set the `-XX:StringTableSize` when go bad perfromance on `intern()` method. And the performance improved much since JDK11 event without  `-XX:StringTableSize` parameter.

BTW, memory leak is intolerable in any cases especially in a database server system.

Alima777:
> BTW, memory leak is intolerable in any cases especially in a database server system in my opinion.

Hi, I understand your concerns. So what I said is `we can delete useless string from cachedStringPool manually when time series is deleted` to avoid memory leak. I didn't check the conclusion about memory footprint using `WeakHashMap`, maybe it stores more things than normal HashMap?

chengjianyun:
> we can delete useless string from cachedStringPool manually when time series is deleted

As far as I know, we can do this by count references to each object like older verion JVM GC does. If we do like this, why don't let JVM help recycle the memory usage by use `WeakHashMap`? Or do you have other ideas?

> `WeakHashMap` is 4~5 times than`intern()`, is `concurrentHashMap` still the same?

The HashMap or ConcurrentHashMap implemented CachePool is wrong without explicit clear, so I didn't verify with them.

Alima777:
> As far as I know, we can do this by count references to each object like older verion JVM GC does.

Hi, you should know we only use this `cachedStringPool` in limited situation. It's only used in TsFileResource. So count references maybe not needed, we just add it when initialize, and remove it when delete timeseries.

chengjianyun:
> So count references maybe not needed, we just add it when initialize, and remove it when delete timeseries.

I don't think this way would work. As the strings from `CacheStringPool` are passed to different instances, you must know all of the references of a key are released then the `CacheStringPool` could remove the key. Otherwise, you would hold different instances for same string which will lose the meaning of `CacheStringPool`. 

And for the class, it designs to provide ability like string intern, but has implementation defects. You can't restrict it only used to hold device or measurement path unless you name it something like `DevicePathPool` or `MeasurementPathPool` instead of `CacheStringPool`.

The two ways I can figure out are either use `WeakHashMap<String, WeakRefernce<String>>` instead of `ConcurrentHashMap` or  `String.intern()`. I prefer the later one in the PR.

qiaojialin:
> memory leak is intolerable in any cases especially in a database server system in my opinion.

+1

JackieTien97 (approved):

qiaojialin (approved):

PR #4968, opened 2022-01-25
[IOTDB-2433] Fix insert aligned timeseries performance reduction

HTHou:
## Description

There is code logic optimization. When insert aligned time series data and check the measurements order in insertPlan, we can use the HashMap instead of the ArrayList. 

The indexOf method in ArrayList seems taking too much time.... :(

Screen Shot 2022-01-25 at 9 43 22 AM


qiaojialin (approved):

wangchao316 (approved):
LGTM

PR #4969, opened 2022-01-25
fix storage group name

cigarl:


cigarl:
already fixed in the master branch

PR #4977, opened 2022-01-26
[IOTDB-2476] Fix concurrent bug during entityMNode replacement

zyk990424:
### Problem
During concurrent timeseries auto creation, MTree will throw NPE.
See  IOTDB-2476

### Cause
The replacement operation of EntityMNode is not a synchronized operation, thus a thread will read a null between the deletion and addition, which is the existing implementation of repalcement.

### Solution
Use replace method of ConcurrentHashMap to make the node replacement a synchronized operation and make the replace method of MNode a synchronized operation.


chengjianyun (approved):
LGTM

qiaojialin (approved):

PR #4978, opened 2022-01-26
Removing some useless codes in Statistics

yuyuankang:
This pull request removes some methods that are never used.


HTHou (approved):

tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/statistics/DoubleStatistics.java
```diff
@@ -38,12 +37,17 @@
   private double sumValue;
 
   static final int DOUBLE_STATISTICS_FIXED_RAM_SIZE = 80;
+  private final String DOUBLE = "Double";
```
qiaojialin, line 40:
```suggestion
  private static final String DOUBLE = "Double";
```

PR #4985, opened 2022-01-26
[IOTDB-2487] Time series containing special characters can be created but connot be used normally

liuminghui233:
In IoTDB, `splitPathToDetachedPath()` in `MetaUtils` is often called to convert `String` to `PartialPath`. However, some time series cannot be used normally due to ambiguity, even if they are created. 

Eg.
```
[root, sg, "s1"."s2"] vs [root, sg, "s1", "s2"]
[root, sg, "a."] vs [root, sg, "a, "]
[root, sg, "ab, cd"] vs [root, sg, "ab.cd"]
```

After this PR,
- Quotes can only be escaped with a backslash(`\`)
- Backslashe will be stored as part of the node name

```
[root, sg, "s1\".\"s2"] vs [root, sg, "s1", "s2"]
[root, sg, "a."] vs [root, sg, \"a, \"]
[root, sg, \"ab, cd\"] vs [root, sg, "ab.cd"]
```


SteveYurongSu (dismissed):
LGTM

SteveYurongSu (commented):
The validation process seems to be different for show timeseries and create timeseries statements?

SteveYurongSu (changes requested):
We should fundamentally solve the problem of serialization and deserialization of ts

SteveYurongSu (approved):
LGTM

PR #4988, opened 2022-01-27
[IOTDB-2501]Fix timestamp not updated bug in FirstValueAggrResult.updateResultUsingValues()

czllgzmzl:
## Description
Fix a bug in FirstValueAggrResult.updateResultUsingValues()
The behaviour of FirstValueAggrResult.updateResultUsingValues() and LastValueDescAggrResult.updateResultUsingValues() should be the same.

### Content
Update the timestamp of FirstValueAggrResult when the result is updated from ValueIterator.

### Problem caused by the bug

The result of first_value aggregation may be wrong when query with "value filter + group by level", since the merge() of FirstValueAggrResults is no-op when timestamps of two FirstValueAggrResults are both Long.MAX_VALUE.

### Test
Add a test case in GroupByLevelDataSetTest.java. The query statementis like "select first_value(s0) from root.** where s0!=0 or time<10 group by level=3". The result before the fix is wrong.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/query/aggregation/impl/FirstValueAggrResult.java
+++ b/server/src/main/java/org/apache/iotdb/db/query/aggregation/impl/FirstValueAggrResult.java
@@ -113,7 +113,7 @@ public class FirstValueAggrResult extends AggregateResult {
       return;
     }
     if (valueIterator.hasNext()) {
-	  timestamp = timestamps[valueIterator.getCurPos()];
+      timestamp = timestamps[valueIterator.getCurPos()];
       setValue(valueIterator.next());
     }
   }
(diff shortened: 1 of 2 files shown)


qiaojialin:
Hi, please create an issue in Jira: https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-1753?filter=allopenissues
Then this will be added to the release note.

Besides, it's better to add an IT for this case~

ericpai:
LGTM, could you add a new IT or modify an existed one in `IoTDBAggregation*IT` of this special case?

Alima777 (approved):
LGTM

ericpai (approved):

PR #5026, opened 2022-02-09
[IOTDB-2523] Fix exiting with 0 when executing error in cli

ericpai:
See JIRA: https://issues.apache.org/jira/browse/IOTDB-2523

After this fix being applied, we got this result in my desktop with Windows 10 installed, which is as our expected.


ericpai:
@HTHou @cornmonster PTAL~

PR #5046, opened 2022-02-11
[To rel/0.12] [IOTDB-2532] Query with align by device can't get value after clear cache

Alima777:
Path in where clause is not included in measurements of specific device.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/qp/physical/crud/RawDataQueryPlan.java
+++ b/server/src/main/java/org/apache/iotdb/db/qp/physical/crud/RawDataQueryPlan.java
@@ -23,7 +23,9 @@ import org.apache.iotdb.db.metadata.PartialPath;
 import org.apache.iotdb.db.qp.logical.Operator;
 import org.apache.iotdb.tsfile.file.metadata.enums.TSDataType;
 import org.apache.iotdb.tsfile.read.common.Path;
+import org.apache.iotdb.tsfile.read.expression.IBinaryExpression;
 import org.apache.iotdb.tsfile.read.expression.IExpression;
+import org.apache.iotdb.tsfile.read.expression.impl.SingleSeriesExpression;
 
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/apache/iotdb/db/qp/physical/crud/RawDataQueryPlan.java
```diff
@@ -54,6 +56,19 @@ public IExpression getExpression() {
 
   public void setExpression(IExpression expression) throws QueryProcessException {
     this.expression = expression;
+    updateDeviceMeasurementsUsingExpression(expression);
+  }
```
JackieTien97, line 60: * `setExpression` will be called twice, once in `transformFilterOperatorToExpression` of `QueryOperator` which generating logical plan, once in `QueryRouter` after optimizing this expression. Although it won't make it it doesn't affect its correctness, the performance may be worse than before.
* In align by device query, the `deviceToMeasurements` should only contains current device, in your current implementation, all the devices will be saved in `deviceToMeasurements`, even if the devices that has been traversed.

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/qp/physical/crud/RawDataQueryPlan.java
Alima777: Oh, Why don't we move `optimizing this expression` from `QueryRouter` to `QueryOperator` when generating this expression, then setExpression will be invoked only once.

Alima777: > `all the devices will be saved in deviceToMeasurements`
That's not right. In align by device query, there is a map `<device -> expression>`, so one expression will only include current device.

JackieTien97: > Oh, Why don't we move `optimizing this expression` from `QueryRouter` to `QueryOperator` when generating this expression, then setExpression will be invoked only once.

It sounds ok, you can move the optimization logic from `QueryRouter` to `transformQuery` function of  `PhysicalGenerator`. I found that only `rawQuery`, `aggregate`, `groupBy` and `udtfQuery` call the `optimize` function, you can figure it out whether other queries can apply that `optimize`.

JackieTien97: > > `all the devices will be saved in deviceToMeasurements`
> 
> That's not right. In align by device query, there is a map `<device -> expression>`, so one expression will only include current device.

I see, although different devices share the same `RawDataQueryPlan`, `setDeduplicatedPaths` of `RawDataQueryPlan` will call `clear()` function of `deviceToMeasurements`.

JackieTien97 (changes requested):

PR #5001, opened 2022-01-30
[IOTDB-2510] Reconstructing the password encryption mode

wangchao316:
1. We can adapt to different password encryption modes based on our requirements. 
2. MD5 is not a secure encryption mode and can be decrypted.

PR #5052, opened 2022-02-14
[IOTDB-2540] Hide password encrypt configuration until more encryption added

HTHou:
## Description

#5001 added the configuration of changing password encrypt algorithm. However we have still only one encrypt algorithm MD5 in IoTDB currently. These configs are meaningless for normal users. Therefore, I think we should hide them until more encryption methods added.


wangchao316 (approved):
Good.
sorry, I forgetting this...

qiaojialin (approved):

PR #5053, opened 2022-02-14
[To rel/0.12][IOTDB-2281][IOTDB-2286] restructure the csv import tool 

xuanronaldo:


cli/src/main/java/org/apache/iotdb/tool/ImportCsv.java
```diff
@@ -65,14 +72,22 @@
   private static final String FAILED_FILE_ARGS = "fd";
   private static final String FAILED_FILE_NAME = "failed file directory";
 
+  private static final String ALIGNED_ARGS = "aligned";
+  private static final String ALIGNED_NAME = "use the aligned interface";
+
```
HTHou, line 77: This is for aligned timeseries. We shouldn't introduce it to 0.12.

cli/src/main/java/org/apache/iotdb/tool/AbstractCsvTool.java
```diff
@@ -99,6 +99,9 @@
         "yyyy/MM/dd'T'HH:mm:ss",
         "yyyy.MM.dd'T'HH:mm:ss"
       };
+  protected static final int CODE_OK = 0;
+  protected static final int CODE_ERROR = 1;
+
```
HTHou, line 104: These lines from come PR #5026. Consider updating export-csv.bat and import-csv.bat ?

cli/src/main/java/org/apache/iotdb/tool/ImportCsv.java
xuanronaldo: got it!

PR #5061, opened 2022-02-15
[IOTDB-2538] The UDTF Transform method supports getting the start and end times of rowWindow

SteveYurongSu:
See IOTDB-2538 for more detail.


iskytek:
LGTM

PR #5063, opened 2022-02-16
[IOTDB-2543] Support all time filters in last query

cornmonster:
Please see [IOTDB-2543](https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-2543?filter=myopenissues) for detailed information.


server/src/main/java/org/apache/iotdb/db/query/executor/LastQueryExecutor.java
```diff
@@ -215,26 +220,35 @@ public QueryDataSet execute(QueryContext context, LastQueryPlan lastQueryPlan)
...
-                  resultContainer.get(i).right.getTimestamp());
-            }
+    for (int i = 0; i < cachedLastPairs.size(); i++) {
+      if (cachedLastPairs.get(i).right == null) {
+        cachedLastPairs.get(i).right = readerList.get(index++).readLastPoint();
+        // Update the cache only when,
+        // 1. the last value cache doesn't exist
+        // 2. the value is not null
+        // 3. last value cache is enabled
+        // 3. the filter is gt (greater than) or ge (greater than or equal to)
```
ericpai, line 230: `3.` => `4.`

ericpai (changes requested):

server/src/main/java/org/apache/iotdb/db/query/executor/LastQueryExecutor.java
cornmonster: Fixed.

ericpai:
Need to update the docs.  As we have two scenarios of `LAST` query now:
1. Query with only time filters `>` AND `>=` or without any filters: IoTDB will try to get from the last cache first and then tsfile, and then update the cache.
2. Query with time filters '<' or any other composed ones: It will query the tsfile directly without updating the cache.

We should let users know the difference of the performace impact of them. And leave the choice to users. The `LAST` keyword in 2 is just a syntax sugar.

server/src/main/java/org/apache/iotdb/db/qp/physical/crud/LastQueryPlan.java
```diff
@@ -80,16 +77,12 @@ public void setExpression(IExpression expression) throws QueryProcessException {
...
   }
 
   // Only > and >= are supported in time filter
   private boolean isValidExpression(IExpression expression) {
-    if (expression instanceof GlobalTimeExpression) {
-      Filter filter = ((GlobalTimeExpression) expression).getFilter();
-      return filter instanceof TimeGtEq || filter instanceof TimeGt;
-    }
-    return false;
+    return expression instanceof GlobalTimeExpression;
```
HTHou, line 86: Only time filters are supported in LAST query.
What about other time filter like `TimeEq`, `TimeNotEq` or `TimeIn` ?

server/src/main/java/org/apache/iotdb/db/qp/physical/crud/LastQueryPlan.java
```diff
@@ -80,16 +77,12 @@ public void setExpression(IExpression expression) throws QueryProcessException {
     if (isValidExpression(expression)) {
       super.setExpression(expression);
     } else {
-      throw new QueryProcessException("Only '>' and '>=' are supported in LAST query");
+      throw new QueryProcessException("Only time filters are supported in LAST query");
     }
   }
 
   // Only > and >= are supported in time filter
```
HTHou, line 84: Change this?

cornmonster:
> Need to update the docs. As we have two scenarios of `LAST` query now:
> 
> 1. Query with only time filters `>` AND `>=` or without any filters: IoTDB will try to get from the last cache first and then tsfile, and then update the cache.
> 2. Query with time filters '<' or any other composed ones: It will query the tsfile directly without updating the cache.
> 
> We should let users know the difference of the performace impact of them. And leave the choice to users. The `LAST` keyword in 2 is just a syntax sugar.

Agree.

server/src/main/java/org/apache/iotdb/db/qp/physical/crud/LastQueryPlan.java
cornmonster: Good question.

Time filter is used in 2 scenarios,
1. judge if the last value cache satisfies the filter
2. being passed to the LastPointReader

Any time filter works with the two scenarios above.

server/src/main/java/org/apache/iotdb/db/qp/physical/crud/LastQueryPlan.java
cornmonster: Good catch! I have fixed the comment and updated the documents as @ericpai  suggested.

ericpai (approved):

ericpai:
LGTM

docs/UserGuide/Query-Data/Last-Query.md
```diff
@@ -21,27 +21,27 @@
 
 # Last Query
 
-The last query is a special query provided in the time series database Apache IoTDB. The last query returns the data point with the largest timestamp in the time series, that is, the latest state of a sequence. Users can specify the last query through `select last`. It is especially important in IoT data analysis scenarios as the latest point data characterizes the current state. In order to provide a millisecond-level return speed, Apache IoTDB optimizes the cache for the last query to meet users' performance requirements for real-time monitoring of devices.
+The last query is a special type of query in Apache IoTDB. It returns the data point with the largest timestamp of the specified time series. In other word, it returns the latest state of a time series. This feature is especially important in IoT data analysis scenarios. To meets the performance requirement of real-time device monitoring systems, Apache IoTDB caches the latest values of all time series to achieve microsecond read latency.
```
JackieTien97, line 24:
```suggestion
The last query is a special type of query in Apache IoTDB. It returns the data point with the largest timestamp of the specified time series. In other word, it returns the latest state of a time series. This feature is especially important in IoT data analysis scenarios. To meet the performance requirement of real-time device monitoring systems, Apache IoTDB caches the latest values of all time series to achieve microsecond read latency.
```

server/src/main/java/org/apache/iotdb/db/query/executor/LastQueryExecutor.java
```diff
@@ -171,8 +175,9 @@ public QueryDataSet execute(QueryContext context, LastQueryPlan lastQueryPlan)
             nonCachedPaths,
             nonCachedDataTypes,
             context.isDebug());
+
     if (nonCachedPaths.isEmpty()) {
-      return resultContainer;
+      return cachedLastPairs.stream().map(pair -> pair.right).collect(Collectors.toList());
```
JackieTien97, line 180: Actually, the reason that we directly return the resultContainer is that we want to reuse it instead of creating another ArrayList, even if the pair.left will never be used. It is for performance. If you really want to not return extra result, you can make readLastPairsFromCache to return Pair<boolean[], List<TimeValuePair>> instead and return res.right.

server/src/main/java/org/apache/iotdb/db/query/executor/LastQueryExecutor.java
```diff
@@ -215,26 +220,35 @@ public QueryDataSet execute(QueryContext context, LastQueryPlan lastQueryPlan)
...
-          if (CACHE_ENABLED) {
-            cacheAccessors.get(i).write(resultContainer.get(i).right);
-            if (context.isDebug()) {
-              DEBUG_LOGGER.info(
-                  "[LastQueryExecutor] Update last cache for path: {} with timestamp: {}",
-                  seriesPaths,
-                  resultContainer.get(i).right.getTimestamp());
-            }
+    for (int i = 0; i < cachedLastPairs.size(); i++) {
+      if (cachedLastPairs.get(i).right == null) {
```
JackieTien97, line 224: Here, if the filter is gt (greater than) or ge (greater than or equal to) and cachedLastPairs.get(i).left is Boolean.TRUE, we don't need to read from disk again.

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/qp/physical/crud/LastQueryPlan.java
JackieTien97: Maybe you could add `TimeEq`, `TimeNotEq` and `TimeIn` case in IT.

server/src/main/java/org/apache/iotdb/db/query/executor/LastQueryExecutor.java
cornmonster: Good catch! I am busy working on a test right now, will update the PR in 2 days.

server/src/main/java/org/apache/iotdb/db/query/executor/LastQueryExecutor.java
cornmonster: Good point. I refactored the code to solve this problem and make the code more readable. Please take a look.

cornmonster:
@JackieTien97

HTHou (approved):
LGTM
@JackieTien97

server/src/main/java/org/apache/iotdb/db/query/executor/LastQueryExecutor.java
```diff
@@ -150,36 +153,115 @@ public QueryDataSet execute(QueryContext context, LastQueryPlan lastQueryPlan)
...
+      List<TimeValuePair> nonCachedLastPairs =
+          readLastPairsFromStorage(
+              nonCachedPaths, nonCachedDataTypes, filter, context, deviceMeasurementsMap);
+      for (int i = 0; i < nonCachedLastPairs.size(); i++) {
+        // Update the cache only when,
+        // 1. the last value cache doesn't exist
+        // 2. the actual last value is not null
+        // 3. last value cache is enabled
+        // 4. the filter is gt (greater than) or ge (greater than or equal to)
+        if (lastPairs.get(i) == null
```
JackieTien97, line 207:
```suggestion
        if (lastPairs.get(nonCachedIndices.get(i)) == null
```

JackieTien97 (changes requested):

JackieTien97 (approved):

PR #5066, opened 2022-02-17
[IOTDB-2522] Check WAL files before upgrading 0.12 to 0.13

HeimingZ:
The wal of 0.12 is not compatible with 0.13. Therefore, we need to check whether the WAL file existed before upgrading to 0.13.


HTHou (approved):
LGTM

PR #5070, opened 2022-02-17
[IOTDB-2511]Change import-csv-tool  feedback from synchronous to asynchronous

limeng32:
When importing massive data, synchronous feedback will have a great impact on processing time, so I changed the feedback that impacts on performance to asynchronous way. Based on the principle of minimum modification, there is no change of feedback that has little impact on performance.

### DIFF
--- a/cli/src/main/java/org/apache/iotdb/tool/ImportCsv.java
+++ b/cli/src/main/java/org/apache/iotdb/tool/ImportCsv.java
@@ -37,6 +37,8 @@ import org.apache.commons.csv.CSVParser;
 import org.apache.commons.csv.CSVRecord;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.thrift.annotation.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.FileInputStream;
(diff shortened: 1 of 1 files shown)


cli/src/main/java/org/apache/iotdb/tool/ImportCsv.java
```diff
@@ -360,9 +364,15 @@ private static void writeDataAlignedByTime(
...
-                        record.getRecordNumber(), header, value);
+                    logger.error(
+                        new StringBuilder("Line '")
+                            .append(record.getRecordNumber())
+                            .append("', column '")
+                            .append(header)
+                            .append("': '")
+                            .append(value)
+                            .append("' unknown type")
+                            .toString());
```
HTHou, line 375: Hi, I think we can use
```java
logger.error("Line {}, column {}: {} unknown type", record.getRecordNumber(), header, value);
```

limeng32: I will do it

limeng32:
Please review again.

HTHou (approved):

HTHou (changes requested):
Hi, I see there are still many `System.out.println();` in this class, like L539. 
Should we change all of them to use `logger`?

limeng32:
I only modified System.out in the stream, and did not change the one that has little effect on performance. It would be a big project if we changed all.

PR #5072, opened 2022-02-17
[IOTDB-2557] Fix non-data-read QueryContext in getAllMeasurementSchema

ericpai:
See JIRA https://issues.apache.org/jira/browse/IOTDB-2557

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
@@ -29,8 +29,10 @@ import org.apache.iotdb.cluster.coordinator.Coordinator;
 import org.apache.iotdb.cluster.exception.CheckConsistencyException;
 import org.apache.iotdb.cluster.exception.UnsupportedPlanException;
 import org.apache.iotdb.cluster.partition.PartitionGroup;
+import org.apache.iotdb.cluster.query.RemoteQueryContext;
 import org.apache.iotdb.cluster.query.manage.QueryCoordinator;
 import org.apache.iotdb.cluster.rpc.thrift.GetAllPathsResult;
+import org.apache.iotdb.cluster.rpc.thrift.MeasurementSchemaRequest;
 import org.apache.iotdb.cluster.rpc.thrift.Node;
 import org.apache.iotdb.cluster.rpc.thrift.RaftNode;
 import org.apache.iotdb.cluster.server.handlers.caller.ShowTimeSeriesHandler;
@@ -1543,7 +1545,7 @@ public class CMManager extends MManager {
     if (group.contains(metaGroupMember.getThisNode())) {
       showLocalTimeseries(group, plan, context, handler);
     } else {
-      showRemoteTimeseries(group, plan, handler);
+      showRemoteTimeseries(group, plan, context, handler);
     }
   }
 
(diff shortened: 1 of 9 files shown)


ericpai:
Can someone help to review these codes?

OneSizeFitsQuorum:
@Alima777  CC

cluster/src/main/java/org/apache/iotdb/cluster/query/LocalQueryExecutor.java
```diff
@@ -602,16 +602,15 @@ public long querySingleSeriesByTimestamp(SingleSeriesQueryRequest request)
...
-    ShowTimeSeriesPlan plan = (ShowTimeSeriesPlan) PhysicalPlan.Factory.create(planBuffer);
+    ShowTimeSeriesPlan plan = (ShowTimeSeriesPlan) PhysicalPlan.Factory.create(request.planBinary);
     List<ShowTimeSeriesResult> allTimeseriesSchema;
-    allTimeseriesSchema =
-        getCMManager()
-            .showLocalTimeseries(
-                plan, new QueryContext(SessionManager.getInstance().requestQueryId(false)));
+    RemoteQueryContext queryContext =
+        queryManager.getQueryContext(request.getRequester(), request.getQueryId());
+    allTimeseriesSchema = getCMManager().showLocalTimeseries(plan, queryContext);
```
OneSizeFitsQuorum, line 613: Looks reasonable, but if we pass the queryId this way, we need to send an endQuery request from the coordinator node after the query completes to clean up the resources in the queryManager, otherwise there will be a memory leak. So far I haven't found any code related to resource cleaning, what do you think?

OneSizeFitsQuorum (commented):
PTAL

cluster/src/main/java/org/apache/iotdb/cluster/query/LocalQueryExecutor.java
ericpai: @OneSizeFitsQuorum I debug this case and see that the queryId is assigned at QueryTask.call(). (Here the queryId is 31)
image

And the query resource is released in `FetchResultTask` and `SessionManager.closeDataSet()` (The query id is 31 as well)

image
image

For any query, it will execute the query firstly and get a query id. Then it will fetch results, if necessary. Fetching results request will be sent with this query id. When fetching results ends, it will release all local and remote query resources.

So I think there's no resource leak. And you can reference `ClusterSessionManager.releaseQueryResource()` to get the logics of remote query resource release.

OneSizeFitsQuorum: `ClusterSessionManager.releaseQueryResource()` will release resource in the remote node only when the remote node is registered by `RemoteQueryContext.registerRemoteNode(Node node, RaftNode header)`. 

Now it's possible that showTimeseriesPlan holds the state of the remote node, but it doesn't seem to register the remote node during execution?

image

ericpai: @OneSizeFitsQuorum Nice catch! It's my mistake to miss the `registerRemoteNode()`. I add this in a finally block in `private void showRemoteTimeseries(PartitionGroup group, ShowTimeSeriesPlan plan, QueryContext context, ShowTimeSeriesHandler handler)` in `CMManager`.

cluster/src/main/java/org/apache/iotdb/cluster/metadata/CMManager.java
```diff
@@ -1603,6 +1608,9 @@ private void showRemoteTimeseries(
       } catch (InterruptedException e) {
         logger.error("Interrupted when getting timeseries schemas in node {}.", node, e);
         Thread.currentThread().interrupt();
+      } finally {
+        // record the queried node to release resources later
+        ((RemoteQueryContext) context).registerRemoteNode(node, group.getHeader());
```
ericpai, line 1613: @OneSizeFitsQuorum Here I reference the `ClusterReaderFactory.getRemoteGroupByExecutor()` to handle the `registerRemoteNode` to make sure the remote resource can be actually released after query.

OneSizeFitsQuorum (approved):
LGTM

PR #5073, opened 2022-02-17
[IOTDB-2562] Change default value of sync mlog period parameter

HeimingZ:
Change default value of sync mlog period parameter from 0 to 100.

### DIFF
--- a/docs/UserGuide/Reference/Config-Manual.md
+++ b/docs/UserGuide/Reference/Config-Manual.md
@@ -490,13 +490,13 @@ The permission definitions are in ${IOTDB\_CONF}/conf/jmx.access.
 |Default| 1048576 |
 |Effective|After restart system|
 
-* force\_mlog\_period\_in\_ms
+* sync\_mlog\_period\_in\_ms
 
-|Name| force\_mlog\_period\_in\_ms |
+|Name| sync\_mlog\_period\_in\_ms |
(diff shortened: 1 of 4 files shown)


qiaojialin (approved):

chengjianyun:
Is there any background of the update?

qiaojialin:
0 will sharply decrease the performance of creating timeseries. So change it to 100, the same as force_wal_period_in_ms

HTHou (approved):

chengjianyun:
> 0 will sharply decrease the performance of creating timeseries. So change it to 100, the same as force_wal_period_in_ms

I see. How sharply is that? Do you have any test result?

mychaow:
> > 0 will sharply decrease the performance of creating timeseries. So change it to 100, the same as force_wal_period_in_ms
> 
> I see. How sharply is that? Do you have any test result?

I tested it with benchmark and found that the registration timing was so slow that I killed the test. This can estimate the speed drop. When it is 0, it takes about 10ms to access the disk each time a sequence is created, and it is estimated that about 100 sequences are created per second. When it is 100, a time series is 300 bytes, 1MB cache can carry about 3000 times, and the cache is flushed 10 times per second, and about 30,000 time series can be created. There are two orders of magnitude difference in speed.

PR #5076, opened 2022-02-17
[IOTDB-2564]Not allowed to activate template on a path not set template

bigreybear:


SilverNarcissus (approved):
Nice fix~  Look good to me

PR #5090, opened 2022-02-18
[To rel/0.12][IOTDB-2567]  Fix thread and ByteBuffer leak after service stopped

ericpai:
See JIRA: https://issues.apache.org/jira/browse/IOTDB-2567.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/virtualSg/VirtualStorageGroupManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/virtualSg/VirtualStorageGroupManager.java
@@ -18,6 +18,7 @@
  */
 package org.apache.iotdb.db.engine.storagegroup.virtualSg;
 
+import org.apache.iotdb.db.concurrent.ThreadName;
 import org.apache.iotdb.db.engine.StorageEngine;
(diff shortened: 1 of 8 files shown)


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/virtualSg/VirtualStorageGroupManager.java
```diff
@@ -432,4 +434,12 @@ public void releaseWalDirectByteBufferPool() {
   public void reset() {
     Arrays.fill(virtualStorageGroupProcessor, null);
   }
+
+  public void stopSchedulerPool() {
+    for (StorageGroupProcessor vsg : this.virtualStorageGroupProcessor) {
+      if (vsg != null) {
+        ThreadUtils.stopThreadPool(vsg.getWALTrimScheduleTask(), ThreadName.WAL_TRIM);
```
HeimingZ, line 441: Types are incompatible, please call getName() method like this:
```
ThreadUtils.stopThreadPool(vsg.getWALTrimScheduleTask(), ThreadName.WAL_TRIM.getName());
```

HeimingZ (changes requested):

PR #5110, opened 2022-02-23
[IOTDB-2598]Enable append operation on template after setting and using

bigreybear:


server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTree.java
```diff
@@ -1799,21 +1838,19 @@ protected boolean processFullMatchedMNode(IMNode node, int idx, int level)
   public boolean isTemplateSetOnMTree(String templateName) {
     // check whether template has been set
     Deque<IMNode> nodeStack = new ArrayDeque<>();
```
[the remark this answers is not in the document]
bigreybear: Thanks! But this method had been removed since every Template object will record storage group it has been set now.

MarcosZyk (commented):
LGTM~

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTree.java
```diff
@@ -1718,105 +1758,96 @@ public int getMountedNodeIndexOnMeasurementPath(PartialPath measurementPath)
     return fullPathNodes.length - 1;
   }
 
-  public List<String> getPathsSetOnTemplate(String templateName) throws MetadataException {
+  public List<String> getPathsSetOnTemplate(Template template) throws MetadataException {
+    String templateName = template == null ? "" : template.getName();
```
MarcosZyk, line 1762: Should this template name "" replaced by * ? I see * means all template.

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTree.java
```diff
@@ -1718,105 +1758,96 @@ public int getMountedNodeIndexOnMeasurementPath(PartialPath measurementPath)
...
-    setTemplatePaths.traverse();
+          };
+      setTemplatePaths.traverse();
+    }
     return resSet;
   }
 
-  public List<String> getPathsUsingTemplate(String templateName) throws MetadataException {
+  public List<String> getPathsUsingTemplate(Template template) throws MetadataException {
+    String templateName = template == null ? "" : template.getName();
```
MarcosZyk, line 1806: Should we use * ?

bigreybear: thanks! fixed now

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTree.java
bigreybear: fixed now

MarcosZyk (approved):

bigreybear:
NOTICE:
Given that we strongly recommend not setting templates on the nodes above the storage group, and that we want to disturb the mlog upgrade as little as possible, there was a very rare case that could cause an error after this pull request:
Suppose there is a user had set a template to a node above the storage group, and this MTree was recorded in the mlog. After this pull request, any new storage group under the set node will not be able to visit the template by querying paths set or using the template or be designated a new template.

qiaojialin (approved):

PR #5113, opened 2022-02-24
[To rel/0.12][IOTDB-2603]Fix compaction recover

choubenson:
Fix 2 problems:
1. There may be more data after iotdb restart. The reason is that cross space compaction recovery did not consisder one case, which is when some source files lost.
2. If there is deletion during cross space compaction process, if the iOTDB is stopped then and all the source files are not deleted, then the deleted data will be queried during the next restart.

### DIFF
--- a/server/src/assembly/resources/conf/iotdb-engine.properties
+++ b/server/src/assembly/resources/conf/iotdb-engine.properties
@@ -451,11 +451,6 @@ timestamp_precision=ms
 # If you are feeling the rebooting is too slow, set this to false, false by default
 # continue_merge_after_reboot=false
 
-# When set to true, all unseq merges becomes full merge (the whole SeqFiles are re-written despite how
-# much they are overflowed). This may increase merge overhead depending on how much the SeqFiles
-# are overflowed.
-# force_full_merge=true
-
 # How many threads will be set up to perform compaction, 10 by default.
 # Set to 1 when less than or equal to 0.
 # compaction_thread_num=10
--- a/server/src/main/java/org/apache/iotdb/db/engine/merge/task/RecoverMergeTask.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/merge/task/RecoverMergeTask.java
@@ -24,7 +24,6 @@ import org.apache.iotdb.db.engine.merge.recover.MergeLogAnalyzer;
 import org.apache.iotdb.db.engine.merge.recover.MergeLogAnalyzer.Status;
 import org.apache.iotdb.db.engine.merge.recover.MergeLogger;
 import org.apache.iotdb.db.engine.modification.Modification;
-import org.apache.iotdb.db.engine.modification.ModificationFile;
 import org.apache.iotdb.db.engine.storagegroup.TsFileResource;
 import org.apache.iotdb.db.exception.metadata.MetadataException;
 import org.apache.iotdb.tsfile.fileSystem.FSFactoryProducer;
@@ -36,7 +35,6 @@ import java.io.File;
 import java.io.IOException;
 import java.util.List;
 
-import static org.apache.iotdb.db.engine.storagegroup.StorageGroupProcessor.MERGING_MODIFICATION_FILE_NAME;
 import static org.apache.iotdb.db.engine.storagegroup.TsFileResource.modifyTsFileNameUnseqMergCnt;
 
 /**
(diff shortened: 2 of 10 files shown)


server/src/assembly/resources/conf/iotdb-engine.properties
```diff
@@ -451,11 +451,6 @@ timestamp_precision=ms
 # If you are feeling the rebooting is too slow, set this to false, false by default
 # continue_merge_after_reboot=false
 
-# When set to true, all unseq merges becomes full merge (the whole SeqFiles are re-written despite how
-# much they are overflowed). This may increase merge overhead depending on how much the SeqFiles
-# are overflowed.
-# force_full_merge=true
-
```
THUMarkLau, line 458: remove the argument in `IoTDBConfig` and `IoTDBDescriptor`

server/src/main/java/org/apache/iotdb/db/engine/merge/task/RecoverMergeTask.java
```diff
@@ -80,7 +95,57 @@ public void recoverMerge() throws IOException, MetadataException {
...
+      File targetFileResource = new File(targetFile.getPath() + TsFileResource.RESOURCE_SUFFIX);
+      File tmpTargetFile = new File(sourceSeqResource.getTsFilePath() + MERGE_SUFFIX);
+      // move to target file and target resource file
+      if (!targetFileResource.exists()) {
+        if (!targetFile.exists()) {
+          // move target file
+          FSFactoryProducer.getFSFactory().moveFile(tmpTargetFile, targetFile);
+        }
+        // move target resource file
+        FSFactoryProducer.getFSFactory()
```
THUMarkLau, line 117: If the target file already exists, its tsfileResource must exist, because SGP recovery generates a tsfileResource for a tsfile that does not have a tsfileResource (see `TsFileRecoverPerformer::recover`). If the target file does not exist and the corresponding temporary file does, the tsfileResource should be regenerated by reading from the temporary file rather than moving the tsfileResource of source file directly.

THUMarkLau (changes requested):
The process of `handleWhenSomeSourceFilesLost` need to be modifed.

server/src/main/java/org/apache/iotdb/db/engine/merge/task/RecoverMergeTask.java
choubenson: resolved

server/src/assembly/resources/conf/iotdb-engine.properties
choubenson: resolved

server/src/main/java/org/apache/iotdb/db/engine/merge/task/RecoverMergeTask.java
```diff
@@ -80,7 +95,39 @@ public void recoverMerge() throws IOException, MetadataException {
...
+      // delete source seq file
+      sourceSeqResource.remove();
+
+      // delete merge file
+      File mergeFile = new File(sourceSeqResource.getTsFilePath() + MERGE_SUFFIX);
+      if (mergeFile.exists()) {
+        mergeFile.delete();
+      }
+
+      sourceSeqResource.setFile(targetFile);
```
jt2594838, line 129: What is the point of `sourceSeqResource.setFile(targetFile);`?

server/src/main/java/org/apache/iotdb/db/engine/merge/task/RecoverMergeTask.java
```diff
@@ -80,7 +95,39 @@ public void recoverMerge() throws IOException, MetadataException {
...
+      if (!targetFile.exists()) {
+        // move target file
+        FSFactoryProducer.getFSFactory().moveFile(tmpTargetFile, targetFile);
+      }
+      // serialize target resource file
+      TsFileResource targetTsFileResouce = new TsFileResource(targetFile);
+      try (TsFileSequenceReader reader = new TsFileSequenceReader(targetFile.getAbsolutePath())) {
+        FileLoaderUtils.updateTsFileResource(reader, targetTsFileResouce);
+      }
+      targetTsFileResouce.serialize();
```
jt2594838, line 118: Is this newly generated resource managed by TsFileManagement?

jt2594838 (approved):

PR #5115, opened 2022-02-24
[IOTDB-2604] batch size is invalid in import-csv tool

xuanronaldo:


cli/src/main/java/org/apache/iotdb/tool/ImportCsv.java
```diff
@@ -455,7 +455,7 @@ private static void writeDataAlignedByDevice(
             writeAndEmptyDataSet(
                 deviceName.get(), times, typesList, valuesList, measurementsList, 3);
             deviceName.set(record.get(1));
-          } else if (record.getRecordNumber() - 1 % BATCH_SIZE == 0 && times.size() != 0) {
+          } else if (times.size() >= BATCH_SIZE && times.size() != 0) {
```
qiaojialin, line 458:
```suggestion
          } else if (times.size() >= BATCH_SIZE) {
```

qiaojialin (approved):

qiaojialin (approved):

PR #5124, opened 2022-02-26
[IOTDB-2595] add buildin UDTF ON_OFF

flashzxi:
## Description
[IOTDB-2595] add a buildin UDTF on_off

### Function
input: a time series (type: INT32, INT64, FLOAT or DOUBLE)
output: a time series (type: BOOLEAN)

parameters：
threshold (double)： if value >= threshold, output true, otherwise false

usage:
select s1, f(s1, 'threshold'='1.5') from root.sg.d


SteveYurongSu (changes requested):
Hi, thanks for your contribution!

Please add tests and user docs for the new function.

SteveYurongSu (approved):
LGTM

PR #5125, opened 2022-02-26
[IOTDB-2611] Fix incorrect result of COUNT NODES in cluster

ericpai:
See JIRA: https://issues.apache.org/jira/browse/IOTDB-2611


cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
```diff
@@ -322,59 +324,82 @@ protected int getNodesNumInGivenLevel(PartialPath path, int level) throws Metada
...
+      sgPathMap = new HashMap<>(extendedSgPathMap.size());
+      // We should remove the last ** to restore the pattern to make it equivalent with the query.
+      for (Entry<String, List<PartialPath>> extendSgPaths : extendedSgPathMap.entrySet()) {
+        List<PartialPath> validPaths =
+            extendSgPaths.getValue().stream()
+                .map(PartialPath::getDevicePath)
+                .collect(Collectors.toList());
+        sgPathMap.put(extendSgPaths.getKey(), validPaths);
+      }
+
```
jt2594838, line 353: Why is this necessary? If the query does not have a wildcard, we should match it strictly according to it like other queries.

cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
```diff
@@ -322,59 +324,82 @@ protected int getNodesNumInGivenLevel(PartialPath path, int level) throws Metada
...
+          boolean hasWildcard = false;
+          int targetLevel = Math.min(pattern.getNodeLength(), level);
+          for (int i = 0; i < targetLevel; i++) {
+            if (pattern.getNodes()[i].equals(IoTDBConstant.MULTI_LEVEL_PATH_WILDCARD)
+                || pattern.getNodes()[i].equals(IoTDBConstant.ONE_LEVEL_PATH_WILDCARD)) {
+              hasWildcard = true;
+              break;
             }
-            break;
           }
```
jt2594838, line 385: I think this may be unnecessarily complicated. Since we know the storage groups and what nodes are managing them, we can simply send the original query to them, only with specific storage groups.

jt2594838 (changes requested):

cluster/src/main/java/org/apache/iotdb/cluster/query/ClusterPlanExecutor.java
ericpai: Thanks for your comments. As we discussed online, I will refine those codes in a reasonable way soon.

ericpai:
As old cluster is deprecated, this PR is useless in the future.

PR #5127, opened 2022-02-27
[IOTDB-2612] Fix cannot select compaction candidate after exception occurrence in previous compaction

THUMarkLau:
see [IOTDB-2612](https://issues.apache.org/jira/browse/IOTDB-2612)

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/compaction/cross/AbstractCrossSpaceCompactionTask.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/compaction/cross/AbstractCrossSpaceCompactionTask.java
@@ -71,10 +71,12 @@ public abstract class AbstractCrossSpaceCompactionTask extends AbstractCompactio
 
     for (TsFileResource resource : selectedSequenceFiles) {
       resource.setCompacting(true);
+      resource.setCompactionCandidate(false);
     }
 
     for (TsFileResource resource : selectedUnsequenceFiles) {
       resource.setCompacting(true);
+      resource.setCompactionCandidate(false);
     }
 
     return true;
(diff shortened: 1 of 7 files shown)


server/src/main/java/org/apache/iotdb/db/engine/compaction/cross/AbstractCrossSpaceCompactionTask.java
```diff
@@ -95,7 +87,7 @@ public String toString() {
 
   @Override
   public void resetCompactionCandidateStatusForAllSourceFiles() {
```
qiaojialin, line 89:
```suggestion
  public void resetCompactionStatusForAllSourceFiles() {
```

server/src/main/java/org/apache/iotdb/db/engine/compaction/inner/AbstractInnerSpaceCompactionTask.java
```diff
@@ -131,6 +127,6 @@ public String toString() {
 
   @Override
   public void resetCompactionCandidateStatusForAllSourceFiles() {
```
qiaojialin, line 129:
```suggestion
  public void resetCompactionStatusForAllSourceFiles() {
```

server/src/main/java/org/apache/iotdb/db/engine/compaction/cross/AbstractCrossSpaceCompactionTask.java
THUMarkLau: resolve

server/src/main/java/org/apache/iotdb/db/engine/compaction/inner/AbstractInnerSpaceCompactionTask.java
THUMarkLau: resolve

qiaojialin (approved):

PR #5129, opened 2022-02-28
[IOTDB-2620] Unrecognizable operator type (SHOW) for AuthorityChecker

SteveYurongSu:
See IOTDB-2620 for more detail.


Alima777 (approved):

server/src/main/java/org/apache/iotdb/db/auth/AuthorityChecker.java
```diff
@@ -125,6 +125,7 @@ private static int translateToPermissionId(Operator.OperatorType type) {
       case DELETE:
       case DROP_INDEX:
         return PrivilegeType.DELETE_TIMESERIES.ordinal();
+      case SHOW:
```
HTHou, line 128: Does 0.12 have this issue?

SteveYurongSu: Yes, it has...

PR #5130, opened 2022-02-28
[To rel/0.12][IOTDB-2624] "overlapped data should be consumed first" occurs when executing query

THUMarkLau:
see [IOTDB-2624](https://issues.apache.org/jira/browse/IOTDB-2624)

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
+++ b/server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
@@ -153,7 +153,7 @@ public class IoTDBConfig {
   private int flushWalThreshold = 10000;
 
   /** this variable set timestamp precision as millisecond, microsecond or nanosecond */
-  private String timestampPrecision = "ms";
+  private String timestampPrecision = "ns";
 
   /**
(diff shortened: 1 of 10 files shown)


server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
```diff
@@ -153,7 +153,7 @@
   private int flushWalThreshold = 10000;
 
   /** this variable set timestamp precision as millisecond, microsecond or nanosecond */
-  private String timestampPrecision = "ms";
+  private String timestampPrecision = "ns";
```
HTHou, line 156: Don't change this.

server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
```diff
@@ -715,7 +715,7 @@
   private int ioTaskQueueSizeForFlushing = 10;
 
   /** the number of virtual storage groups per user-defined storage group */
-  private int virtualStorageGroupNum = 1;
+  private int virtualStorageGroupNum = 8;
```
HTHou, line 718: Do not change it.

PR #5131, opened 2022-02-28
[IOTDB-2625] Fix election deadlock when updating partition table

ericpai:
See JIRA: https://issues.apache.org/jira/browse/IOTDB-2625

### DIFF
--- a/cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
+++ b/cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
@@ -260,6 +260,7 @@ public abstract class RaftMember implements RaftMemberMBean {
     }
 
     startBackGroundThreads();
+    setSkipElection(false);
     logger.info("{} started", name);
   }
 


cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
```diff
@@ -260,6 +260,7 @@ public void start() {
     }
 
     startBackGroundThreads();
+    setSkipElection(false);
```
jamber001, line 263: setSkipElection(false);  放到 startBackGroundThreads(); 前面， 是不是 更好？

ericpai: If so, the status will be 'ready to launch election' before all the necessary threads starting (You can reference `RaftMember.startBackGroundThreads`). There are some corner cases that the heartbeat thread starts and launchs the election, but other threads, like `commitLog` and `appendLog` is not ready yet. Some unknown status and error may occur.
The safe way is to let `start()` be just a reverse-mirror of `stop()`.

OneSizeFitsQuorum (approved):
LGTM

cluster/src/main/java/org/apache/iotdb/cluster/server/member/RaftMember.java
jamber001: If because of this concern, what will happen when first time running this startBackGroundThreads() after IoTDB process starting up?
At that time, the skipElection is just false before running startBackGroundThreads().

ericpai: You are right! The default value should be `true`, not `false`. I will optimize this in a new PR.

PR #5136, opened 2022-02-28
[IOTDB-2597] Add four functions to find suitable intervals

flashzxi:
### Overview
Add four function to calculate intervals in which data points always satisfy some specific conditions.
- zero_duration 
- non_zero_duration
- zero_count
- non_zero_count

### Details 
`zero_duration` will take two parameters `min`(default value **0**) and `max`(default value Long.MAX_VALUE),
It will return intervals' start times and duration times (`end_time - start_time`) in which the value is always 0(false), and the duration time `t` satisfy `t >= min && t <= max`. The unit of `t` is ms.

`non_zeroDuration` is similar to `zero_duration`, only with points always not 0;

`zero_count` will take two parameters `min`(default value **1**) and `max`(default value Long.MAX_VALUE);
It will return intervals' start times and the number of data points in the interval in which the value is always 0(false), and the number n satisfy `n >= min && n <= max`;

`non_zero_count` is similar

### example
```
IoTDB> select s1, zero_count(s1) from root.sg.d2;
+-----------------------------+-------------+-------------------------+
|                         Time|root.sg.d2.s1|zero_count(root.sg.d2.s1)|
+-----------------------------+-------------+-------------------------+
|1970-01-01T08:00:00.000+08:00|            0|                        1|
|1970-01-01T08:00:00.001+08:00|            1|                     null|
|1970-01-01T08:00:00.002+08:00|            1|                     null|
|1970-01-01T08:00:00.003+08:00|            0|                        1|
|1970-01-01T08:00:00.004+08:00|            1|                     null|
|1970-01-01T08:00:00.005+08:00|            0|                        2|
|1970-01-01T08:00:00.006+08:00|            0|                     null|
|1970-01-01T08:00:00.007+08:00|            1|                     null|
+-----------------------------+-------------+-------------------------+

```


SteveYurongSu:
Please resolve the conflict :)

server/src/main/java/org/apache/iotdb/db/query/udf/builtin/UDTFContinuouslySatisfy.java
```diff
@@ -0,0 +1,240 @@
...
+import org.apache.iotdb.tsfile.utils.Pair;
+
+import java.io.IOException;
+import java.util.ArrayList;
+
+public abstract class UDTFContinuouslySatisfy implements UDTF {
+  protected Long min;
+  protected Long max;
+  protected Long defaultMax = Long.MAX_VALUE;
+  protected Long defaultMin = 0L;
```
SteveYurongSu, line 42: These 2 fields can be static final. And... Does it seem that we can merge min(max) and defaultMin(defaultMax) into 1 field?

server/src/main/java/org/apache/iotdb/db/query/udf/builtin/UDTFContinuouslySatisfy.java
```diff
@@ -0,0 +1,240 @@
...
+    if (satisfyDouble(value) && satisfyValueCount == 0L) {
+      satisfyValueCount++;
+      satisfyValueStartTime = time;
+      satisfyValueLastTime = time;
+    } else if (satisfyDouble(value)) {
+      satisfyValueCount++;
+      satisfyValueLastTime = time;
+    } else {
+      if (getRecord() >= min && getRecord() <= max && satisfyValueCount > 0) {
+        intervals.add(new Pair<>(satisfyValueStartTime, getRecord()));
```
SteveYurongSu, line 133: Try to output the result to PointCollector here. It's not recommended to keep all results in memory, because there is a risk to cause OOM.

SteveYurongSu (changes requested):
Thanks for your contribution :D

server/src/main/java/org/apache/iotdb/db/query/udf/builtin/UDTFContinuouslySatisfy.java
SteveYurongSu: Yes, we do memory control in PointCollector.

server/src/main/java/org/apache/iotdb/db/query/udf/builtin/UDTFContinuouslySatisfy.java
```diff
@@ -0,0 +1,243 @@
...
+        break;
+      case BOOLEAN:
+        needAddNewRecord = transformBoolean(row.getTime(), row.getBoolean(0));
+        break;
+      default:
+        // This will not happen
+        throw new UDFInputSeriesDataTypeNotValidException(
+            0, dataType, TSDataType.INT32, TSDataType.INT64, TSDataType.FLOAT, TSDataType.DOUBLE);
+    }
+    if (needAddNewRecord) collector.putLong(interval.left, interval.right);
```
SteveYurongSu, line 112: Please add the {} here.

server/src/main/java/org/apache/iotdb/db/query/udf/builtin/UDTFContinuouslySatisfy.java
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  protected boolean transformDouble(long time, double value) {
+    if (satisfyDouble(value) && satisfyValueCount == 0L) {
```
SteveYurongSu, line 116: Try to do the judge satisfyDouble only once.

server/src/main/java/org/apache/iotdb/db/query/udf/builtin/UDTFContinuouslySatisfy.java
```diff
@@ -0,0 +1,243 @@
...
+        satisfyValueCount = 0L;
+        return true;
+      }
+      satisfyValueCount = 0L;
+    }
+    return false;
+  }
+
+  protected abstract Long getRecord();
+
```
SteveYurongSu, line 211: Please explain the meaning of record by add some comments here.

server/src/main/java/org/apache/iotdb/db/query/udf/builtin/UDTFContinuouslySatisfy.java
```diff
@@ -109,104 +109,115 @@ public void transform(Row row, PointCollector collector)
...
-        satisfyValueCount = 0L;
-        return true;
+    if (satisfyDouble(value)) {
+      if (satisfyValueCount == 0L) {
+        satisfyValueCount++;
+        satisfyValueStartTime = time;
+        satisfyValueLastTime = time;
+      } else {
+        satisfyValueCount++;
+        satisfyValueLastTime = time;
```
SteveYurongSu, line 125: These 2 lines can be moved out of the if else

SteveYurongSu (approved):

SteveYurongSu (approved):
LGTM

PR #5146, opened 2022-03-01
[IOTDB-2596] Add buildin UDTF in_range

flashzxi:
## Description
[issue2596](https://issues.apache.org/jira/browse/IOTDB-2596)
Added a buildin UDTF in_range

## Details
in_range takes two double type parameters `lower` and `upper`
It will return a timesery of Boolean,  return values are  `value >= lower && value <= upper`

## example

IoTDB> select ts, in_range(ts,'lower'='2', 'upper'='3.1') from root.test;
+-----------------------------+------------+--------------------------------------------------+
|                         Time|root.test.ts|in_range(root.test.ts, "lower"="2", "upper"="3.1")|
+-----------------------------+------------+--------------------------------------------------+
|1970-01-01T08:00:00.001+08:00|           1|                                             false|
|1970-01-01T08:00:00.002+08:00|           2|                                              true|
|1970-01-01T08:00:00.003+08:00|           3|                                              true|
|1970-01-01T08:00:00.004+08:00|           4|                                             false|
|1970-01-01T08:00:00.005+08:00|           5|                                             false|
+-----------------------------+------------+--------------------------------------------------+


server/src/main/java/org/apache/iotdb/db/query/udf/builtin/UDTFInRange.java
```diff
@@ -0,0 +1,90 @@
...
+import org.apache.iotdb.db.query.udf.api.customizer.parameter.UDFParameters;
+import org.apache.iotdb.db.query.udf.api.customizer.strategy.RowByRowAccessStrategy;
+import org.apache.iotdb.db.query.udf.api.exception.UDFException;
+import org.apache.iotdb.db.query.udf.api.exception.UDFInputSeriesDataTypeNotValidException;
+import org.apache.iotdb.tsfile.file.metadata.enums.TSDataType;
+
+import java.io.IOException;
+
+public class UDTFInRange implements UDTF {
+  TSDataType dataType;
```
SteveYurongSu, line 37:
```suggestion
  protected TSDataType dataType;
```

SteveYurongSu (approved):

PR #5152, opened 2022-03-02
RocksDB based metadata management

chengjianyun:


wangchao316:
Hi，jianyun，  we should add doc for this function.

PR #5158, opened 2022-03-02
[IOTDB-2634] Fix batch script exit directly from cmd.exe

ericpai:
See JIRA: https://issues.apache.org/jira/browse/IOTDB-2634

### DIFF
--- a/cli/src/assembly/resources/sbin/start-cli.bat
+++ b/cli/src/assembly/resources/sbin/start-cli.bat
@@ -74,4 +74,4 @@ pause
 
 ENDLOCAL
 
-EXIT %ret_code%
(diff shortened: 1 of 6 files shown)


ericpai:
@HTHou PTAL

ericpai:
It seems that the exit code is not as expected, I will check this later.

ericpai:
@HTHou I have tested both in cmd.exe and Powershell, they all only exit the .bat, not the terminal.

HTHou:
> @HTHou I have tested both in cmd.exe and Powershell, they all only exit the .bat, not the terminal.



HTHou (approved):
Tested.

PR #5159, opened 2022-03-03
[To new_sync][IOTDB-1907] implement customized sync process: sender 3

yschengzi:
Fix BufferedPipeDataQueue, add BufferedPipeDataQueue for sender, add recMsg in SenderService, add IT.


server/src/main/java/org/apache/iotdb/db/newsync/sender/service/SenderService.java
```diff
@@ -175,35 +181,45 @@ public Pipe parseCreatePipePlan(CreatePipePlan plan, PipeSink pipeSink, long pip
...
+    try {
+      if (!(pipeSink instanceof IoTDBPipeSink)) {
+        throw new PipeException(
+            String.format(
+                "Wrong pipeSink type %s for create pipe %s", pipeSink.getType(), pipe.getName()));
+      }
+      ITransportClient transportClient =
+          new TransportClient(
+              pipe, ((IoTDBPipeSink) pipeSink).getIp(), ((IoTDBPipeSink) pipeSink).getPort());
+      pipe.setTransportHandler(
```
Cpaulyz, line 196: It's neccessary to send heartbeat CREATE to receiver to regisiter this pipe.

Why not add a new function `create()` for pipe, which creates  TransportClient and send CREATE heartbeat?

server/src/main/java/org/apache/iotdb/db/newsync/transport/conf/TransportSenderConfig.java
```diff
@@ -0,0 +1,129 @@
...
+
+  private String lastFileInfoPath;
+
+  private String snapshotPath;
+
+  /** The maximum number of retry when syncing a file to receiver fails. */
+  private int maxNumOfSyncFileRetry = 5;
+
+  /** Storage groups which participate in sync process */
+  private List<String> storageGroupList = new ArrayList<>();
```
Cpaulyz, line 45: seems that it's need to use this parameter in new sync?

PR #5163, opened 2022-03-03
[IOTDB-2654] Alias doesn't show when using group by level

Alima777:
insert into root.sg.d1(time,s1,s2) values(1,1,1);
insert into root.sg.d2(time,s1,s2) values(1,1,1);

image
image
image


server/src/main/java/org/apache/iotdb/db/qp/utils/GroupByLevelController.java
```diff
@@ -136,6 +139,44 @@ public void control(ResultColumn rawColumn, List<ResultColumn> resultColumns)
...
+
+  private void checkAliasAndUpdateAliasMap(ResultColumn rawColumn, String originName)
+      throws LogicalOptimizeException {
+    if (!rawColumn.hasAlias()) {
+      return;
+    } else if (aliasMap == null) {
+      aliasMap = new HashMap<>();
+    }
+    // If an alias is corresponding to more than one result column, throw an exception
+    if (aliasMap.get(originName) == null && aliasMap.containsValue(rawColumn.getAlias())) {
```
JackieTien97, line 167: Better to add a `Map<String, String> aliasToColumnName`, time complexity of `containsValue` in `HashMap` is O(N).

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/qp/utils/GroupByLevelController.java
Alima777: I know that...But I did this based on that there are usually not so many aliases in one query, and there has been so many data structures that I don't want to add more.

JackieTien97: > there are usually not so many aliases in one query

Some users may not write the sql manual, they may concat the sql in their program, they may write hundreds or thousands of column in select clause instead of using `*`, like: `sum(s1) as s1_sum, sum(s2) as s2_sum, .... sum(s1000) as s1000_sum`. We won't know how users will write the sql. 

> there has been so many data structures

One more field won't be detrimental to the readability of the program if you add enough comments to explain the usage of that field you add.

server/src/main/java/org/apache/iotdb/db/query/expression/ResultColumn.java
```diff
@@ -88,12 +88,15 @@ public ResultColumn(Expression expression) {
   /**
    * @param prefixPaths prefix paths in the from clause
    * @param resultColumns used to collect the result columns
+   * @param isGroupByLevel used to skip illegal alias judgement here. Because count(*) may be *
+   *     unfolded to more than one expression, but it still can be aggregated together later
    */
-  public void concat(List<PartialPath> prefixPaths, List<ResultColumn> resultColumns)
+  public void concat(
+      List<PartialPath> prefixPaths, List<ResultColumn> resultColumns, boolean isGroupByLevel)
```
JackieTien97, line 95: After discussing with Steve Yurong SU @SteveYurongSu , it is not very elegant to add `isGroupByLevel` here. It' better to add an interface, like `Validator` which has a `validate` method in it. And AliasValidator implements that interface, and GroupByLevelAliasValidator `override` that `validate` method.

server/src/main/java/org/apache/iotdb/db/query/expression/ResultColumn.java
```diff
@@ -88,12 +88,15 @@ public ResultColumn(Expression expression) {
...
       throws LogicalOptimizeException {
     List<Expression> resultExpressions = new ArrayList<>();
     expression.concat(prefixPaths, resultExpressions);
-    if (hasAlias() && 1 < resultExpressions.size()) {
+    if (!isGroupByLevel && hasAlias() && 1 < resultExpressions.size()) {
```
JackieTien97, line 99: same as above

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/query/expression/ResultColumn.java
Alima777: It's not necessary I think....
Let's think the impelement way. Do I need to implement a validator in every ResultColumn? It's not expected obviously.
Or pass the validator as a parameter of this method? Then what's the different from now?

JackieTien97: We can pass the validator as a parameter of this method. It's better to extend. If we have another rules about the alias or something else, we can implement another `Validator` and passing it instead of adding another `boolean` parameter, like `isGroupByLevel`.

PR #5173, opened 2022-03-07
[IOTDB-2692] Fix compaction exception caused by deleted timeseries

ericpai:
See JIRA: https://issues.apache.org/jira/browse/IOTDB-2692

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/compaction/CompactionUtils.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/compaction/CompactionUtils.java
@@ -31,6 +31,7 @@ import org.apache.iotdb.db.engine.storagegroup.TsFileResource;
 import org.apache.iotdb.db.exception.StorageEngineException;
 import org.apache.iotdb.db.exception.metadata.IllegalPathException;
 import org.apache.iotdb.db.exception.metadata.MetadataException;
+import org.apache.iotdb.db.exception.metadata.PathNotExistException;
 import org.apache.iotdb.db.metadata.path.AlignedPath;
 import org.apache.iotdb.db.metadata.path.MeasurementPath;
 import org.apache.iotdb.db.metadata.path.PartialPath;
@@ -60,6 +61,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 /**
  * This tool can be used to perform inner space or cross space compaction of aligned and non aligned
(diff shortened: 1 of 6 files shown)


ericpai:
@THUMarkLau @choubenson PTAL

server/src/main/java/org/apache/iotdb/db/engine/compaction/CompactionUtils.java
```diff
@@ -125,8 +125,12 @@ private static void compactAlignedSeries(
     List<IMeasurementSchema> measurementSchemas = new ArrayList<>();
     for (String measurement : allMeasurments) {
       // TODO: use IDTable
-      measurementSchemas.add(
-          IoTDB.metaManager.getSeriesSchema(new PartialPath(device, measurement)));
+      PartialPath p = new PartialPath(device, measurement);
+      if (!IoTDB.metaManager.isPathExist(p)) {
+        logger.info("As path {} is dropped we skip it", p.getFullPath());
+        continue;
+      }
```
qiaojialin, line 132: Hi, what if the deletion just takes effect between the check and the getSeriesSchema?

ericpai: That's a good catch, maybe we can just catch the exception without pre existence check?

THUMarkLau (commented):
Please try to add an UT

server/src/main/java/org/apache/iotdb/db/engine/compaction/CompactionUtils.java
qiaojialin: Yes, I prefer to just catch the exception and ignore it

ericpai: `try-catch` is added instead of pre-check.

ericpai:
> Please try to add an UT

Thanks for your comments! I have updated the existence test cases to cover the situation that timeseries are deleted before compaction.

qiaojialin (commented):
@THUMarkLau  please check the MeasurementIterator

server/src/main/java/org/apache/iotdb/db/engine/compaction/inner/utils/InnerSpaceCompactionUtils.java
```diff
@@ -108,12 +112,22 @@ private static void compactNotAlignedSeries(
     while (seriesIterator.hasNextSeries()) {
       checkThreadInterrupted(targetResource);
       // TODO: we can provide a configuration item to enable concurrent between each series
-      String currentSeries = seriesIterator.nextSeries();
+      PartialPath p = new PartialPath(device, seriesIterator.nextSeries());
+      IMeasurementSchema ms;
```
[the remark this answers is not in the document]
ericpai: Fixed.

server/src/main/java/org/apache/iotdb/db/engine/compaction/inner/utils/InnerSpaceCompactionUtils.java
```diff
@@ -108,12 +112,22 @@ private static void compactNotAlignedSeries(
     while (seriesIterator.hasNextSeries()) {
       checkThreadInterrupted(targetResource);
       // TODO: we can provide a configuration item to enable concurrent between each series
-      String currentSeries = seriesIterator.nextSeries();
+      PartialPath p = new PartialPath(device, seriesIterator.nextSeries());
```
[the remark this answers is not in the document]
ericpai: The logic of constructing the full timeseries path is not changed. 
Besides, the data of `seriesIterator` comes from https://github.com/apache/iotdb/blob/master/tsfile/src/main/java/org/apache/iotdb/tsfile/read/TsFileSequenceReader.java#L1896-L1931 , which returns a map with keys of type `TimeseriesMetaData.getMeasurementId()`.

qiaojialin (approved):
LGTM

server/src/main/java/org/apache/iotdb/db/engine/compaction/CompactionUtils.java
```diff
@@ -121,20 +123,30 @@ private static void compactAlignedSeries(
...
-    List<String> allMeasurments = alignedMeasurmentIterator.getAllMeasurements();
+    List<String> allMeasurements = alignedMeasurmentIterator.getAllMeasurements();
     List<IMeasurementSchema> measurementSchemas = new ArrayList<>();
-    for (String measurement : allMeasurments) {
+    for (String measurement : allMeasurements) {
       // TODO: use IDTable
-      measurementSchemas.add(
-          IoTDB.metaManager.getSeriesSchema(new PartialPath(device, measurement)));
+      try {
+        measurementSchemas.add(
```
THUMarkLau, line 131: What if the series get deleted when compacting aligned series in inner space?

THUMarkLau (commented):
Consider a aligned measurement is deleted when executing inner space compaction.

choubenson (approved):

server/src/main/java/org/apache/iotdb/db/engine/compaction/CompactionUtils.java
ericpai: IMO, we can category the timing of deleting timeseries and compaction:
1. If the timeseries deletion happens before compaction, all the closed tsfiles will skip those deleted timeseries data and the merged one will not contain them.
2. If the timeseries deletion happens during compaction, as it has already checked the measurements validation before compaction, the deleted timeseries data is still merged in this compaction work. Although it's a waste of the resource, but no program or data error happens. And in next compaction task, the former deleted data will be processed as condition 1.

PR #5197, opened 2022-03-10
Add methods to SessionUtils！

Nice9166:
1.Generate by class to Template
2.Add a record entry parameter based on entity generation
3.Any level of splicing entity simple name generation prefixPath
4.Any level of splicing entity attribute name generation prefixPaths
5.Query arbitrary level data based on time and return Java array


MarcosZyk (commented):
Please add some comments and test cases to help user and programmer understand the scenario.

PR #5198, opened 2022-03-11
[IOTDB-2503] Import a Status for TsFileResource

THUMarkLau:
see [IOTDB-2503](https://issues.apache.org/jira/browse/IOTDB-2503)

Currently, there are four boolean flag in TsFileResource representing the status of the object, they are `closed`, `compaction candidate`, `compacting`, `deleted`. This pr aims to use one status variable to replace the four boolean flag mentioned before. The status variable is enum type, with five avaiable value: `UNCLOSED`, `CLOSED`, `COMPACTION_CANDIDATE`, `COMPACTING`, `DELETED`.


server/src/main/java/org/apache/iotdb/db/engine/compaction/inner/sizetiered/SizeTieredCompactionTask.java
```diff
@@ -253,17 +253,15 @@ public boolean checkValidAndSetMerging() {
...
       if (isHoldingReadLock[i]) {
         selectedTsFileResourceList.get(i).readUnlock();
       }
       if (isHoldingWriteLock[i]) {
         selectedTsFileResourceList.get(i).writeUnlock();
       }
-      if (resetCompactingStatus) {
-        selectedTsFileResourceList.get(i).setCompacting(false);
-      }
+      selectedTsFileResourceList.get(i).setStatus(TsFileResourceStatus.CLOSED);
```
jt2594838, line 264: Is it possible that a file is marked with DELETED during compaction by TTL and set back to CLOSED here?

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
```diff
@@ -577,32 +570,62 @@ public int hashCode() {
...
+      case CLOSED:
+        {
+          if (this.status != TsFileResourceStatus.DELETED) {
+            this.status = TsFileResourceStatus.CLOSED;
+            success = true;
+          }
+          break;
+        }
+      case UNCLOSED:
+        // Can we set a tsfile resource to unclosed status?
```
jt2594838, line 597: Check code usages to confirm this. If no place uses this, print a warn (or error) log with stack trace here. Other cases should be processed the same (print a log when unexpected cases happen).

jt2594838 (changes requested):

server/src/main/java/org/apache/iotdb/db/engine/compaction/inner/sizetiered/SizeTieredCompactionTask.java
THUMarkLau: If the status is `DELETED`, the status will not change to `CLOSED`.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
tsunghanjacktsai: We'd set it to an unclosed state in a test case in RewriteCompactionFileSelectorTest.  So I think we could just keep it and print a warning to users.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
```diff
@@ -577,32 +570,62 @@ public int hashCode() {
...
+    return this.status == TsFileResourceStatus.COMPACTION_CANDIDATE;
   }
 
-  public void setCompactionCandidate(boolean compactionCandidate) {
-    this.compactionCandidate = compactionCandidate;
+  public boolean setStatus(TsFileResourceStatus status) {
+    boolean success = false;
+    switch (status) {
+      case CLOSED:
+        {
```
tsunghanjacktsai, line 589: LGTM and thanks for getting my assigned job done. Just one small style-related thing here. I'm thinking that the curly braces-style switch statement seems a bit unnecessary here and I haven't seen any curly braces-style switch statement in other places of IoTDB yet.

tsunghanjacktsai (changes requested):

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
THUMarkLau: resolve

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
THUMarkLau: resolve

jt2594838 (approved):

choubenson (approved):

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
```diff
@@ -577,32 +570,56 @@ public int hashCode() {
...
+        LOGGER.warn("The status of {} is setting to unclosed", file, e);
+        break;
+      case DELETED:
+        if (this.status == TsFileResourceStatus.CLOSED
+            || this.status == TsFileResourceStatus.COMPACTING) {
+          this.status = TsFileResourceStatus.DELETED;
+          success = true;
+        }
+        break;
+      case COMPACTING:
```
qiaojialin, line 607: !(compaction candidate) -> compacting, throw exception
!unclosed -> unclosed, throw exception
unlcosed -> deleted, throw exception

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
```diff
@@ -577,32 +570,56 @@ public int hashCode() {
...
   }
 
   public boolean isCompactionCandidate() {
-    return compactionCandidate;
+    return this.status == TsFileResourceStatus.COMPACTION_CANDIDATE;
   }
 
-  public void setCompactionCandidate(boolean compactionCandidate) {
-    this.compactionCandidate = compactionCandidate;
+  public boolean setStatus(TsFileResourceStatus status) {
```
qiaojialin, line 585:
```suggestion
  public void setStatus(TsFileResourceStatus status) throw StatusChangeException{
```

THUMarkLau: resolve

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
THUMarkLau: resolve

qiaojialin (approved):

PR #5200, opened 2022-03-11
[To new_sync][IOTDB-2700] Sync transfer adaption to receiver

Cpaulyz:
Sync transfer adaption to receiver


server/src/main/java/org/apache/iotdb/db/newsync/transport/server/TransportServiceImpl.java
```diff
@@ -170,31 +182,32 @@ public TransportStatus transportData(
...
+        try {
+          PipeData pipeData = PipeData.deserialize(byteArray);
+          if (type == Type.TSFILE) {
+            // Do with file
+            handleTsFilePipeData((TsFilePipeData) pipeData, fileDir);
+          }
+          PipeDataQueueFactory.getBufferedPipeDataQueue(getPipeLogDirPath(identityInfo))
+              .offer(pipeData);
         } catch (IOException | IllegalPathException e) {
           e.printStackTrace();
```
irvine0109, line 198: It's my fault. May use the statement like `logger.error("Error XXX", e);` is appropriate. The statement like `e.printStackTrace();` may not print to logger.

server/src/test/java/org/apache/iotdb/db/newsync/transport/TransportServiceTest.java
```diff
@@ -0,0 +1,193 @@
...
+    for (int i = 0; i < resPipeData.size(); i++) {
+      Assert.assertEquals(pipeDataList.get(i), resPipeData.get(i));
+    }
+  }
+
+  private void compareFile(File firFile, File secFile) {
+    try {
+      BufferedInputStream fir = new BufferedInputStream(new FileInputStream(firFile));
+      BufferedInputStream sec = new BufferedInputStream(new FileInputStream(secFile));
+      // 比较文件的长度是否一样
```
irvine0109, line 176: // To compare the length of the files.
BTW, Why don't compare the hash of the files.

server/src/main/java/org/apache/iotdb/db/newsync/transport/server/TransportServiceImpl.java
Cpaulyz: fixed

server/src/test/java/org/apache/iotdb/db/newsync/transport/TransportServiceTest.java
Cpaulyz: Yeah, I compare the length and hash of the files now.

server/src/main/java/org/apache/iotdb/db/newsync/receiver/ReceiverService.java
```diff
@@ -218,7 +220,6 @@ public void start() throws StartupException {
 
   @Override
   public void stop() {
-    stopPipeServer();
     try {
       receiverManager.close();
```
yschengzi, line 224: why delete stop pipe server

Cpaulyz: `stop()` is called when deregister service such as exiting IoTDB. 

`stopPipeServer()` stops pipeserver and marks pipeserver unenabled to disk. Receiver service will read disk file and recover status, the correct status of pipeserver should not be unenabled.

PR #5203, opened 2022-03-11
Update dependabot to check rel/0.13

HTHou:
## Description

Update dependabot to check rel/0.13 branch.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -35,12 +35,12 @@ updates:
     directory: "/"
     schedule:
       interval: monthly
-    target-branch: "rel/0.12"
+    target-branch: "rel/0.13"
     ignore:
       - dependency-name: "*"
         update-types: [ "version-update:semver-major", "version-update:semver-minor" ]
     labels:
-      - "0.12"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -61,11 +61,11 @@ updates:
     directory: "/"
     schedule:
       interval: monthly
-    target-branch: "rel/0.12"
+    target-branch: "rel/0.13"
```
SteveYurongSu, line 64: What about the branch rel/0.12?

HTHou: We usually maintain two major branches. For now, they should be 0.13 and 0.14. I think 0.12 can be a big-fix branch only and it's not necessary to check its dependencies. 

How do you think?

SteveYurongSu: IMO, if we are still maintaining the rel/0.12 branch, we'd better check the dependencies for security and safety.

HTHou: Fixed

SteveYurongSu (approved):

PR #5205, opened 2022-03-11
[IOTDB-2675][IOTDB-2524] Manage Metadata by Storage Group && Support alias, tag and attributes on aligned timeseries

MarcosZyk:
### Motivation
1. Prepare for metadata management of new cluster.
2. Prepare to enhance the performance of the coming feature -- persistence schema 

### Main changes
1. Set up SGMManager to only manage the metadata of one storage group, including mlog and tagFile. Each storage group will be represented as a folder on disk named after its fullPath, and the SetStorageGroupPlan and DeleteStorageGroupPlan is implemented as creating or deleting the target folder
2. The implementation of MManager is refactored based on SGMManager. MManager map the task to SGMManagers and reduce all the returned results to generate the final resultSet
3. Split MTree to MTreeAboveSG, which is used to manage storage groups, and MTreeBelowSg, which is used to manage devices and measurements of one certain storage group
4. Enable Traverser to start traverse from an InternalMNode, thus MTreeBelowSG could concentrate on the process of the sg subtree
5. The template now is persisted by templateLogFile individually
6. Templates are forbidden to be set above storage group
7. Extract TimeseriesStatistics to help control the memory resource, which is refactored from the existing code
8. Eliminate ``` create schema snapshot``` and it won't be supported anymore. 
9. Upgrade programs has been developed to help user recover their metadata from files in the schema dir of old versions; metadata in snapshot will be recovered and the mlog and tagFile will be split by storage group

None behavior or definition of the interfaces of MManager is changed. 
The CMManager of existing cluster is not affected since it dependents on the interfaces of MManager rather than the inner concrete implementation.

### Future work
1. Rename MManager to SchemaEngine and rename SGMManager to MManager

## Support alias, tag and attributes on aligned timeseries

1. Support alias, tag and attributes on aligned timeseries
2. Support upgrade on CreateAlignedTimeseriesPlan since the serialize method and file format have been changed.


server/src/main/java/org/apache/iotdb/db/metadata/storagegroup/SGMManager.java
```diff
@@ -0,0 +1,1765 @@
...
+import java.util.stream.Stream;
+
+import static java.util.stream.Collectors.toList;
+import static org.apache.iotdb.db.utils.EncodingInferenceUtils.getDefaultEncoding;
+import static org.apache.iotdb.tsfile.common.constant.TsFileConstant.PATH_SEPARATOR;
+
+/**
+ * This class takes the responsibility of serialization of all the metadata info and persistent it
+ * into files. This class contains all the interfaces to modify the metadata for delta system. All
+ * the operations will be insert into the logs temporary in case the downtime of the delta system.
```
xingtanzjr, line 117: `will be inserted`

server/src/main/java/org/apache/iotdb/db/metadata/storagegroup/SGMManager.java
```diff
@@ -0,0 +1,1765 @@
...
+        logger.info(
+            "create storage group schema folder {} failed.", sgSchemaFolder.getAbsolutePath());
+      }
+    }
+    logFilePath = sgSchemaDirPath + File.separator + MetadataConstant.METADATA_LOG;
+
+    logFile = SystemFileFactory.INSTANCE.getFile(logFilePath);
+
+    try {
+      isRecovering = true;
```
xingtanzjr, line 249: Maybe we need to consider combine the `init()` with constructor. `isRecovering = true;` appears both in constructor and `init()`, which is confusing. As general thinking, this line should appear only once in the whole procedure of initialization of SGMManager

server/src/main/java/org/apache/iotdb/db/metadata/storagegroup/SGMManager.java
MarcosZyk: Fixed.

server/src/main/java/org/apache/iotdb/db/metadata/storagegroup/SGMManager.java
MarcosZyk: Thanks for suggestion. Already fixed.

server/src/main/java/org/apache/iotdb/db/metadata/storagegroup/SGMManager.java
```diff
@@ -0,0 +1,1762 @@
...
+    storageGroupFullPath = storageGroupMNode.getFullPath();
+
+    sgSchemaDirPath = config.getSchemaDir() + File.separator + storageGroupFullPath;
+    File sgSchemaFolder = SystemFileFactory.INSTANCE.getFile(sgSchemaDirPath);
+    if (!sgSchemaFolder.exists()) {
+      if (sgSchemaFolder.mkdirs()) {
+        logger.info("create storage group schema folder {}", sgSchemaFolder.getAbsolutePath());
+      } else {
+        logger.info(
+            "create storage group schema folder {} failed.", sgSchemaFolder.getAbsolutePath());
```
xingtanzjr, line 237: Shall we interrupt the process if the folder cannot be created ?

server/src/main/java/org/apache/iotdb/db/metadata/storagegroup/StorageGroupManager.java
```diff
@@ -0,0 +1,241 @@
...
+    if (mtree != null) {
+      mtree.clear();
+    }
+  }
+
+  @Override
+  public void setStorageGroup(PartialPath path) throws MetadataException {
+    IStorageGroupMNode storageGroupMNode = mtree.setStorageGroup(path);
+    SGMManager sgmManager = new SGMManager();
+    storageGroupMNode.setSGMManager(sgmManager);
```
xingtanzjr, line 84: Inside the implementation of `setSGMManager`, the storageGroupMNode has been set as a variable of sgmManager. So why do we need to pass it again in the following `init()` ?

Suggest to decouple the relation about storageGroupMNode and sgmManager. The circle dependency of these two class may make the codes to be complicated to read.

server/src/main/java/org/apache/iotdb/db/metadata/mnode/InternalMNode.java
```diff
@@ -169,6 +169,12 @@ public synchronized void replaceChild(String oldChildName, IMNode newChildNode)
       newChildNode.getAsEntityMNode().setUseTemplate(oldChildNode.isUseTemplate());
     }
 
+    if (newChildNode.isStorageGroup() && oldChildNode.isStorageGroup()) {
```
xingtanzjr, line 172: It is better to make these logic as part of the implementation of subclass.

Otherwise, we need to add more `if-else` here if we add new type of IMNode, which is easy to be missed...

server/src/main/java/org/apache/iotdb/db/metadata/mnode/InternalMNode.java
```diff
@@ -169,6 +169,12 @@ public synchronized void replaceChild(String oldChildName, IMNode newChildNode)
       newChildNode.getAsEntityMNode().setUseTemplate(oldChildNode.isUseTemplate());
     }
```
xingtanzjr, line 171: Just be curious, why do we need to clone the children from `oldChildNode` to `newChildNode`...

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeBelowSG.java
```diff
@@ -177,14 +169,28 @@ public void init() throws IOException {
     File mtreeSnapshot = SystemFileFactory.INSTANCE.getFile(mtreeSnapshotPath);
     long time = System.currentTimeMillis();
     if (mtreeSnapshot.exists()) {
-      this.root = deserializeFrom(mtreeSnapshot).root;
-      logger.debug(
-          "spend {} ms to deserialize mtree from snapshot", System.currentTimeMillis() - time);
+      IStorageGroupMNode recoveredTree =
+          deserializeFrom(mtreeSnapshot, storageGroupMNode.getFullPath());
+      if (recoveredTree != null) {
+        this.storageGroupMNode.getParent().replaceChild(storageGroupMNode.getName(), recoveredTree);
```
xingtanzjr, line 175: From this line, I cannot infer that some attributes inside recoveredTree has been modified. But it actually did. 

Suggest to use more clear way to implement the changes here...

server/src/main/java/org/apache/iotdb/db/metadata/storagegroup/SGMManager.java
```diff
@@ -0,0 +1,1762 @@
...
+    logFile = SystemFileFactory.INSTANCE.getFile(logFilePath);
+
+    try {
+      // do not write log when recover
+      isRecovering = true;
+
+      tagManager = new TagManager(sgSchemaDirPath);
+      tagManager.init();
+      mtree = new MTreeBelowSG(storageGroupMNode);
+      mtree.init();
```
xingtanzjr, line 251: Maybe we can put the `init()` inside the constructor of MTreeBelowSG ?

xingtanzjr:
From my understanding, the `MTreeAboveSG` is used to store the tree whose leaf node is storage-group; and the `MTreeBelowSG` is used to represented a MTree whose root is stroage-group. So... we may consider a more suitable name for these two classes because the `Above` and `Below` are not very accurate to describe the usage...

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -230,67 +178,20 @@ protected MManager() {
         logger.info("create system folder {} failed.", schemaFolder.getAbsolutePath());
```
xingtanzjr, line 178: What will happen if this condition is triggered? 
It is a legacy issue but it is better to handle the scenario in this re-factor

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -230,67 +178,20 @@ protected MManager() {
...
       return;
     }
-    logFile = SystemFileFactory.INSTANCE.getFile(logFilePath);
 
     try {
-      isRecovering = true;
+      timeseriesStatistics.init();
+      templateManager.init();
+      storageGroupManager = new StorageGroupManager();
+      storageGroupManager.init();
```
xingtanzjr, line 193: Emm...总感觉这个地方让强迫症看着不太舒服...或许`new StroageGroupManager()`可以放在构造函数里？

xingtanzjr:
A quick question... Does this version is compatible with metadata files from old version ?

xingtanzjr:
After this change, It seems that every storage group will run two threads pool to execute some fixed rate tasks such as doing snapshot. If the count of thread pool is large, how do we control the number of these threads ?

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeBelowSG.java
```diff
@@ -177,14 +169,28 @@ public void init() throws IOException {
...
+
+  public void setStorageGroupMNode(IStorageGroupMNode storageGroupMNode) {
+    this.storageGroupMNode = storageGroupMNode;
+    levelOfSG = storageGroupMNode.getPartialPath().getNodeLength() - 1;
+  }
+
   public void clear() {
-    root = new InternalMNode(null, IoTDBConstant.PATH_ROOT);
+    storageGroupMNode = null;
   }
```
xingtanzjr, line 195: Maybe we can add the SGM info in the log of createSnapshot()...

server/src/main/java/org/apache/iotdb/db/metadata/storagegroup/StorageGroupManager.java
```diff
@@ -0,0 +1,241 @@
...
+    for (SGMManager sgmManager : getAllSGMManagers()) {
+      sgmManager.clear();
+    }
+    if (mtree != null) {
+      mtree.clear();
+    }
+  }
+
+  @Override
+  public void setStorageGroup(PartialPath path) throws MetadataException {
```
xingtanzjr, line 81: This method is not thread-safe. The lines 82-84 may lead to unexpected result if this method is invoked in parallel.

server/src/main/java/org/apache/iotdb/db/metadata/storagegroup/StorageGroupManager.java
```diff
@@ -0,0 +1,241 @@
...
+    IStorageGroupMNode storageGroupMNode = mtree.setStorageGroup(path);
+    SGMManager sgmManager = new SGMManager();
+    storageGroupMNode.setSGMManager(sgmManager);
+    sgmManager.init(storageGroupMNode);
+  }
+
+  @Override
+  public SGMManager getBelongedSGMManager(PartialPath path) throws MetadataException {
```
xingtanzjr, line 89: Not sure whether this method is ok with lock-free. We need to ensure this method could be invoked in parallel.

The root node of SGMManager will be changed when this method is invoked, we need to consider whether it is ok here without concurrent control.

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -18,52 +18,37 @@
  */
```
xingtanzjr, line 18: Just a question here, `ensureStorageGroup` use the default level as the storage group level. If the user already set some other storage group whose level is not equal to the default level, what will happen ?

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeAboveSG.java
```diff
@@ -0,0 +1,533 @@
...
+    while (i < nodeNames.length - 1) {
+      IMNode temp = cur.getChild(nodeNames[i]);
+      if (temp == null) {
+        if (cur.isUseTemplate() && upperTemplate.hasSchema(nodeNames[i])) {
+          throw new PathAlreadyExistException(
+              cur.getPartialPath().concatNode(nodeNames[i]).getFullPath());
+        }
+        cur.addChild(nodeNames[i], new InternalMNode(cur, nodeNames[i]));
+      } else if (temp.isStorageGroup()) {
+        // before set storage group, check whether the exists or not
```
xingtanzjr, line 105: grammar issue

server/src/main/java/org/apache/iotdb/db/metadata/storagegroup/SGMManager.java
MarcosZyk: Fixed. Since SGMManager may be initialized when setStorageGroup, thus the process should be interrupted and now an exception will be thrown.

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeAboveSG.java
```diff
@@ -0,0 +1,533 @@
...
+    int i = 1;
+    // e.g., path = root.a.b.sg, create internal nodes for a, b
+    while (i < nodeNames.length - 1) {
+      IMNode temp = cur.getChild(nodeNames[i]);
+      if (temp == null) {
+        if (cur.isUseTemplate() && upperTemplate.hasSchema(nodeNames[i])) {
+          throw new PathAlreadyExistException(
+              cur.getPartialPath().concatNode(nodeNames[i]).getFullPath());
+        }
+        cur.addChild(nodeNames[i], new InternalMNode(cur, nodeNames[i]));
```
xingtanzjr, line 103: Concurrent issue.

If several threads execute this line in parallel with same noneName, there will be some concurrent issue here

server/src/main/java/org/apache/iotdb/db/metadata/storagegroup/StorageGroupManager.java
MarcosZyk: Fixed. Now sgmmanager won't take a storageGroupMNode as an attribute and the setStorageGroupMNode interface is deleted. The init() method should take storageGroupMNode as input to initialize MTreeBelowSG.

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeAboveSG.java
```diff
@@ -0,0 +1,533 @@
...
+        // before set storage group, check whether the exists or not
+        throw new StorageGroupAlreadySetException(temp.getFullPath());
+      }
+      cur = cur.getChild(nodeNames[i]);
+      upperTemplate = cur.getSchemaTemplate() == null ? upperTemplate : cur.getSchemaTemplate();
+      i++;
+    }
+
+    // synchronize check and add, we need addChild become atomic operation
+    // only write on mtree will be synchronized
```
xingtanzjr, line 114: Grammar issue.

Let's check all the comments and ensure the grammar is OK.

server/src/main/java/org/apache/iotdb/db/metadata/mnode/InternalMNode.java
MarcosZyk: Fixed. I added a new interface, moveDataToNewNode(IMNode newMNode), to IMNode and the replace method will invoke the oldChildNode.moveDataToNewNode(newChildNode). The moveDataToNewNode will be overrided by different types of mnode.

server/src/main/java/org/apache/iotdb/db/metadata/mnode/InternalMNode.java
MarcosZyk: We define the parent node of measurementMNode as Device/Entity, thus there will be a MNode type change when creating timeseries under an existing InternalMNode.

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeBelowSG.java
MarcosZyk: Some comments have been added. Hope it will help understand.

server/src/main/java/org/apache/iotdb/db/metadata/storagegroup/SGMManager.java
MarcosZyk: Fixed. Now the init() is inside the constructor.

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeAboveSG.java
```diff
@@ -0,0 +1,533 @@
...
+    synchronized (this) {
+      if (cur.hasChild(nodeNames[i])) {
+        // node b has child sg
+        if (cur.getChild(nodeNames[i]).isStorageGroup()) {
+          throw new StorageGroupAlreadySetException(path.getFullPath());
+        } else {
+          throw new StorageGroupAlreadySetException(path.getFullPath(), true);
```
xingtanzjr, line 121: What does `hasChild` mean ?

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
MarcosZyk: When trigger this condition, IOException will be thrown when init mmanager, since the mlog or tagFile won't be found.

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeAboveSG.java
```diff
@@ -0,0 +1,533 @@
...
+        }
+      } else {
+        if (cur.isUseTemplate() && upperTemplate.hasSchema(nodeNames[i])) {
+          throw new PathAlreadyExistException(
+              cur.getPartialPath().concatNode(nodeNames[i]).getFullPath());
+        }
+        IStorageGroupMNode storageGroupMNode =
```
xingtanzjr, line 128: Where is the `SGMManager` set to the newly created `StorageGroupMNode` ?

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
MarcosZyk: Fixed. Now singletone is implemented on StorageGroupManager.

server/src/main/java/org/apache/iotdb/db/metadata/storagegroup/StorageGroupManager.java
MarcosZyk: Comments have been added. The synchronized block in MTreeBelowSG and concurrentHashMap in MNode ensure that this is thread-safe.

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -547,70 +339,13 @@ public void createTimeseries(CreateTimeSeriesPlan plan) throws MetadataException
...
-      throw new MetadataException(e);
-    }
-
-    // update id table if not in recovering or disable id table log file
-    if (config.isEnableIDTable() && (!isRecovering || !config.isEnableIDTableLogFile())) {
-      IDTable idTable = IDTableManager.getInstance().getIDTable(plan.getPath().getDevicePath());
-      idTable.createTimeseries(plan);
-    }
+    ensureStorageGroup(plan.getPath());
+    storageGroupManager.getBelongedSGMManager(plan.getPath()).createTimeseries(plan, offset);
```
xingtanzjr, line 347: If the SGMManager is not set in `ensureStorageGroup(plan.getPath());`, this line will trigger NPE

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeBelowSG.java
MarcosZyk: The info of SGM and MTreeAboveSG is all recorded in the sg folders' names.

server/src/main/java/org/apache/iotdb/db/metadata/storagegroup/StorageGroupManager.java
MarcosZyk: The node change is just the nodeType change, from storageGroupMNode to storageGroupEntityMNode, the reference of SGMManager will be taken by both the old node and new node, thus there's no concurrent problem. Since all the references points to the old node will re-point to the new node, thus the old node will be collected as a garbage.

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
MarcosZyk: StorageGroupAlreadSetException will be thrown.

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeAboveSG.java
MarcosZyk: ConcurrentHashMap helps avoid concurrent issue.

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeAboveSG.java
MarcosZyk: If there's already a storage group, named root.a.sg, and users try to set storage group to root.a, then a StorageGroupAlreadySetException with hasChild=true will be thrown. This will forbid MManager from retrying setStorageGroup.

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
MarcosZyk: Exception will be thrown in ensureStorageGroup, and the following code won't be executed.

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -738,23 +439,31 @@ private void ensureStorageGroup(PartialPath path) throws MetadataException {
...
-        // PathNotExistException.
-        throw new PathNotExistException(pathPattern.getFullPath());
-      }
+    List<SGMManager> sgmManagers =
+        storageGroupManager.getInvolvedSGMManagers(pathPattern, isPrefixMatch);
+    if (sgmManagers.isEmpty()) {
+      // In the cluster mode, the deletion of a timeseries will be forwarded to all the nodes. For
+      // nodes that do not have the metadata of the timeseries, the coordinator expects a
+      // PathNotExistException.
+      throw new PathNotExistException(pathPattern.getFullPath());
```
xingtanzjr, line 447: If a exception is threw here, these two scenarios are not distinguished:
 * This node actually doesn't have this series and it is expected.
 * The parameter is not correct for this method and it should return an exception

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeAboveSG.java
MarcosZyk: I think the MTree should not mind the creation or initialization of SGMManager, thus I put the related process in ```StorageGroupManager.setStorageGroup(PartialPath path)```

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeBelowSG.java
```diff
@@ -536,7 +506,7 @@ public void createAlignedTimeseries(
   public Pair<PartialPath, IMeasurementMNode> deleteTimeseriesAndReturnEmptyStorageGroup(
```
xingtanzjr, line 507: Does this method need concurrent control ?

server/src/main/java/org/apache/iotdb/db/metadata/storagegroup/StorageGroupManager.java
```diff
@@ -0,0 +1,259 @@
...
+    return new Pair<>(resultAboveSG.left, sgmManagers);
+  }
+
+  @Override
+  public Pair<List<PartialPath>, List<SGMManager>> getNodesListInGivenLevel(
+      PartialPath pathPattern, int nodeLevel, MManager.StorageGroupFilter filter)
+      throws MetadataException {
+    Pair<List<PartialPath>, Set<IStorageGroupMNode>> resultAboveSG =
+        mtree.getNodesListInGivenLevel(pathPattern, nodeLevel, filter);
+    List<SGMManager> sgmManagers = new LinkedList<>();
```
xingtanzjr, line 219: Why does the `resultAboveSG.left` is needed ？
From my understanding, the MTreeAboveSG does not contain the detailed series.

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeAboveSG.java
```diff
@@ -0,0 +1,533 @@
...
+    }
+    return ret;
+  }
+
+  /**
+   * Check whether the given path exists.
+   *
+   * @param path a full path or a prefix path
+   */
+  public boolean isPathExist(PartialPath path) {
```
xingtanzjr, line 387: The name of this method is confusing.

Actually, this method cannot judge whether a path exists or not. It can only tell the invoker whether the SG of the path exists or not.

server/src/main/java/org/apache/iotdb/db/metadata/storagegroup/IStorageGroupManager.java
```diff
@@ -0,0 +1,231 @@
...
+  List<IStorageGroupMNode> getAllStorageGroupNodes();
+
+  /**
+   * Check whether the path exists in MTreeAboveSG. The path may beyond the MTreeAboveSG scope, then
+   * return true if the covered part exists. The rest part will be checked by certain storage group
+   * subTree.
+   *
+   * @param path a full path or a prefix path
+   */
+  boolean isPathExist(PartialPath path);
```
xingtanzjr, line 177: Rename the method

server/src/main/java/org/apache/iotdb/db/metadata/storagegroup/StorageGroupManager.java
```diff
@@ -0,0 +1,255 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import static org.apache.iotdb.db.conf.IoTDBConstant.PATH_ROOT;
+import static org.apache.iotdb.db.conf.IoTDBConstant.PATH_SEPARATOR;
+
+// This class implements all the interfaces for storage group management. The MTreeAboveSg is used
+// to manage all the storage groups and MNodes above storage group.
+public class StorageGroupManager implements IStorageGroupManager {
```
qiaojialin, line 47: This is duplicated with StorageGroupManager in StorageEngine, rename to StorageGroupSchemaManager

server/src/main/java/org/apache/iotdb/db/metadata/storagegroup/IStorageGroupManager.java
```diff
@@ -0,0 +1,231 @@
...
+import org.apache.iotdb.db.metadata.path.PartialPath;
+import org.apache.iotdb.tsfile.utils.Pair;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+// This class declares all the interfaces for storage group management.
+public interface IStorageGroupManager {
```
qiaojialin, line 34:
```suggestion
public interface IStorageGroupSchemaManager {
```

server/src/main/java/org/apache/iotdb/db/metadata/storagegroup/StorageGroupManager.java
MarcosZyk: This feature manages the mnode in MTree, including measurementMNodes, thus there may be matched results in MTreeAboveSG.

server/src/main/java/org/apache/iotdb/db/metadata/storagegroup/StorageGroupManager.java
MarcosZyk: Fixed.

server/src/main/java/org/apache/iotdb/db/metadata/storagegroup/IStorageGroupManager.java
MarcosZyk: Fixed.

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeAboveSG.java
MarcosZyk: Fixed. Already rename this method to ```isStorageGroupAlreadySet``` and comments have been updated.

server/src/main/java/org/apache/iotdb/db/metadata/storagegroup/IStorageGroupManager.java
MarcosZyk: Fixed.

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
xingtanzjr: Got it

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeAboveSG.java
xingtanzjr: OK, I miss the following codes...

But on the other hand, maybe the concurrent issue should be handled here ? Because the node may be updated by several threads in parallel.

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeAboveSG.java
xingtanzjr: Got it

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeAboveSG.java
xingtanzjr: My point is not regarding to the map's write operation.

For example, thread A is setting the SGMManager to the StorageGroupMNode with below codes. If another thread B executes line 103 at the same time, the StorageGroupMNode's parent Node will be replaced in the whole tree

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeBelowSG.java
xingtanzjr: LOL

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeAboveSG.java
MarcosZyk: ```addChild``` invoke concurrentHashMap's ```putIfAbsent```, which ensures that there won't be parent replacement. After line 103, the added child will be re-get. The replacement of existing node only happens when a measurmentMNode is going to be added. If there's no nested device, which means all children of a device is measurement, there won't be subtree loss.

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
MarcosZyk: This is the implementation of old cluster. It's just kept to help pass IT/UT of old cluster. This will be removed in future development of new cluster.

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeAboveSG.java
MarcosZyk: I've changed the following code. Now there's a result check of addChild and the SGMManager will be initialized before the storageGroupMNode added  to MTree, which makes sure that any thread reach the storageGroup node won't get a null nor a uninitialized SGMManager.

On the other hand, we've analyzed the replacement of parent node and it is ensured that the replacement won't happen on nodes above sg, thus the data of setStorageGroup won't lose.

MarcosZyk:
> A quick question... Does this version is compatible with metadata files from old version ?

Upgrade program has been developed and pushed to this pr. Please take a look at MetadataUpgrader and MetadataUpgradeTest

server/src/main/java/org/apache/iotdb/db/metadata/storagegroup/StorageGroupManager.java
MarcosZyk: This method is used by the old cluster to finish task of  ```getMeasurementGroupByLevel```. This may be removed in future development of new cluster.

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeAboveSG.java
MarcosZyk: Fixed.

MarcosZyk:
Let's conclude the concurrent issues and related solutions.

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeBelowSG.java
MarcosZyk: Fixed. I've synchronized all deletion methods. The ```deleteStorageGroup``` is synchronized in ```StorageGroupSchemaManager``` and ```deleteTimeseries``` is synchronized in belonged SGMManager.

MarcosZyk:
Let's conclude the concurrent issues and solutions.
1. Concurrent deletion: all deletion is synchronized.
2. Node replacement, which may cause subTree loss. The solution is the reference to  ```children``` attribute in old node will be directly copied to new node, which avoids the sutree loss.

MarcosZyk:
Hi, I have an idea about the concurrent issues which is possibly occurred during changing the structure of MTree. 

The safest solution is to synchronize all methods that may invoke ```mnode.addChild``` and ```mnode.replaceChild```, including ```setStorageGroup``` in MTreeAboveSG and ```createTimeserie```, ```createAlignedTimeseries``` and ```getDeviceNodeWithAutoCreate``` in MTreeBelowSG. 

This will make all ```setStorageGroup``` in parallel and make all timeseries and device creation in parallel in one storage group. This definitely affects the performance. 

The creation of storage group is not that frequent, while the creation of timeseries is frequent. If the performance of timeseries creation decrease to a degree that we cannot accept, may be we could add a interface for atomic creation of batched timeseries under one certain device. I think this is worthy, since we can group batched creation of timeseries into one record in MLog, which increases the disk I/O.

xingtanzjr (approved):

qiaojialin (approved):

PR #5208, opened 2022-03-13
[To new_cluster] Resolve conversation for new cluster

CRZbulabula:


confignode/src/main/java/org/apache/iotdb/confignode/manager/ConfigManager.java
```diff
@@ -47,6 +47,7 @@
 
   private LoadBalancer loadBalancer;
 
+  /** Test only constructor */
   public ConfigManager(String hashExecutorClass, int deviceGroupCount) {
```
wangchao316, line 51: if a method only for test,  need add @TestOnly

wangchao316 (changes requested):

qiaojialin (approved):

confignode/src/main/java/org/apache/iotdb/confignode/manager/ConfigManager.java
CRZbulabula: I found that @TestOnly is in org.apache.iotdb.db.utils.TestOnly package. But I couldn't dependence iotdb-server.

CRZbulabula: Fix it

wangchao316 (approved):
LGTM

PR #5217, opened 2022-03-14
[To rel/0.13][IOTDB-2732] Reject inserting an invalid infinity float value 

HTHou:
## Description

See IOTDB-2732
Cherry pick #5210


wangchao316 (approved):
LGTM

PR #5235, opened 2022-03-15
[IOTDB-2743] ttl cannot be set if storage group is not exist

cigarl:
We should tell the user that they are using the wrong storage group,then let them troubleshoot the problem.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
+++ b/server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
@@ -1630,15 +1630,21 @@ public class PlanExecutor implements IPlanExecutor {
     }
   }
 
-  private void operateTTL(SetTTLPlan plan) throws QueryProcessException {
+  private void operateTTL(SetTTLPlan plan)
+      throws QueryProcessException, StorageGroupNotSetException {
     try {
       List<PartialPath> storageGroupPaths =
           IoTDB.schemaProcessor.getMatchedStorageGroups(
               plan.getStorageGroup(), plan.isPrefixMatch());
+      if (storageGroupPaths.isEmpty()) {
+        throw new StorageGroupNotSetException(plan.getStorageGroup().getFullPath());
+      }
       for (PartialPath storagePath : storageGroupPaths) {
         IoTDB.schemaProcessor.setTTL(storagePath, plan.getDataTTL());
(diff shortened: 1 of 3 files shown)


HTHou:
Hi, create a jira issue and add some test about it?

cigarl:
> Hi, create a jira issue and add some test about it?

done. 
PTAL~

server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
```diff
@@ -1496,14 +1496,20 @@ private void operateUnloadFile(OperateFilePlan plan) throws QueryProcessExceptio
...
           IoTDB.metaManager.getMatchedStorageGroups(plan.getStorageGroup(), plan.isPrefixMatch());
+      if (storageGroupPaths.isEmpty()) {
+        throw new StorageGroupNotSetException(plan.getStorageGroup().getFullPath());
+      }
       for (PartialPath storagePath : storageGroupPaths) {
         IoTDB.metaManager.setTTL(storagePath, plan.getDataTTL());
         StorageEngine.getInstance().setTTL(storagePath, plan.getDataTTL());
       }
+    } catch (StorageGroupNotSetException e) {
+      throw e;
```
wangchao316, line 1512: This don't need to catch here, just throw it.

cigarl: `StorageGroupNotSetException `extends from `MetadataException`, and `MManager `throws `MetadataException`.
The `StorageGroupNotSetException `here will be caught by `MetadataException`. So individually catch and throw it.

wangchao316: ok

wangchao316 (approved):
LGTM

wangchao316 (changes requested):
test failed, you could see..

cigarl:
@HTHou @wangchao316 
 Sorry, i just got back from other threads.I reviewed the issue and there are two major changes:
 a. The related unit tests bypass the `planExecutor` and validate the `LocalSchemaProcessor` directly.
 b. The test for TTL in `IoTDBTtlIT` catches the exception as a `TIMESERIES_NOT_EXIST`, which is obviously not correct.
PTAL~

PR #5261, opened 2022-03-16
[IOTDB-2752] reconstructing the start and stop command

wangchao316:
1. Extract TestOnly to the common module.
2. The startup and scaling commands are reconstructed. Separate the parsing of commands from the service.


confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeConfCheck.java
```diff
@@ -70,10 +70,10 @@ public void checkConfig() throws RepeatConfigurationException, IOException {
       }
     }
 
-    FileInputStream inputStream = new FileInputStream(specialPropertiesFile);
-    specialProperties.load(inputStream);
-    checkSpecialProperties();
-    inputStream.close();
+    try (FileInputStream inputStream = new FileInputStream(specialPropertiesFile)) {
```
CRZbulabula, line 73: I think maybe the `try` statement is useless. Because if there occurs an IOException, it will be throwed outside, and the ConfigNode won't be started

qiaojialin (approved):

confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeConfCheck.java
wangchao316: closed where it is used. Because we can't control where this method will be called later, if the caller gets an exception and does not kill the process, a link leak will occur.

CRZbulabula (commented):
LGTM

CRZbulabula (approved):

PR #5267, opened 2022-03-17
[IOTDB-2754] SessionPool auto redirect IoTDB instance

CRZbulabula:


example/session/src/main/java/org/apache/iotdb/SessionPoolExample.java
```diff
@@ -32,31 +32,54 @@
...
 
-  private static SessionPool pool;
+  private static SessionPool customSessionPool;
+  private static SessionPool redirectSessionPool;
   private static ExecutorService service;
 
   public static void main(String[] args)
       throws StatementExecutionException, IoTDBConnectionException, InterruptedException {
-    pool =
+    // Custom SessionPool
```
SilverNarcissus, line 41: We should avoid commented code in example

SilverNarcissus (commented):
Nice work!  We also need some test for this new session pool, could you please add some IT?

example/session/src/main/java/org/apache/iotdb/SessionPoolExample.java
OneSizeFitsQuorum, line 41: You'd better clean up the example code. Don't comment out the code...

OneSizeFitsQuorum (approved):
LGTM. Only minor issue. PTAL

session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
```diff
@@ -108,6 +111,20 @@ public SessionPool(String host, int port, String user, String password, int maxS
         Config.DEFAULT_CONNECTION_TIMEOUT_MS);
   }
 
+  public SessionPool(List<String> nodeUrls, String user, String password, int maxSize) {
```
OneSizeFitsQuorum, line 114: Please update the sessionPool documentation to show greater fault tolerance when passing multiple endpoints like a session

OneSizeFitsQuorum (commented):
PTAL

CRZbulabula:
> Nice work! We also need some test for this new session pool, could you please add some IT?

Sorry, I can't add IT for this new feature. Because our IT environment only supports to run a single IoTDB Instance. In fact, this example code acts as a test. You can start IoTDB-server either on port 6667 or on port 6668 and this example code still work.

session/src/main/java/org/apache/iotdb/session/pool/SessionPool.java
CRZbulabula: Done

SilverNarcissus (approved):
Look good to me~

PR #5268, opened 2022-03-17
[IOTDB-2755] Fix tests of flink-tsfile-connector not executed in CI

HTHou:
## Description

The following failed tests not be executed in CI.

image


wangchao316 (approved):
LGTM

PR #5271, opened 2022-03-17
[To rel/0.13][IOTDB-2759]Complete result of show template using or set on template

bigreybear:
This issue is fixed on 0.13 by a extra check on the related storage group of the target template.

On 0.14 or further version, this issue will be avoided by a refactored Traverser.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTree.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTree.java
@@ -1767,6 +1767,13 @@ public class MTree implements Serializable {
             : template.getRelatedStorageGroup();
     List<String> resSet = new ArrayList<>();
     for (PartialPath sgPath : initPath) {
+      // before Traverser refactored as 0.14, need check storage group alone
+      if (getNodeByPath(sgPath).getSchemaTemplate() != null) {
+        if (templateName.equals(ONE_LEVEL_PATH_WILDCARD)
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTree.java
```diff
@@ -1767,6 +1767,13 @@ public int getMountedNodeIndexOnMeasurementPath(PartialPath measurementPath)
             : template.getRelatedStorageGroup();
     List<String> resSet = new ArrayList<>();
     for (PartialPath sgPath : initPath) {
+      // before Traverser refactored as 0.14, need check storage group alone
+      if (getNodeByPath(sgPath).getSchemaTemplate() != null) {
+        if (templateName.equals(ONE_LEVEL_PATH_WILDCARD)
+            || templateName.equals(getNodeByPath(sgPath).getSchemaTemplate().getName())) {
+          resSet.add(sgPath.getFullPath());
+        }
+      }
```
MarcosZyk, line 1776: According to this scenario, if you want to include the sg node, it is better to use ```traverser.setPrefixMatch(true)``` .

bigreybear: Thanks a lot ! And I have fixed as you suggested

MarcosZyk (approved):
LGTM~

PR #5281, opened 2022-03-18
Modify the jdbc query time column to be empty

RYH61:
When the query statement is executed with jdbc, when the last column of the previous row is empty, the next row last Read Was Null will still be true, resulting in the Timestamp column not displaying time information


qiaojialin (commented):
Hi, please try to add an IT~

mychaow (approved):
LGTM

PR #5294, opened 2022-03-21
[IOTDB-2726][IOTDB-2727] Basic interfaces definition of mpp scheduling and memory control

ericpai:


ericpai:
@JackieTien97 @xingtanzjr @cornmonster @qiaojialin PTAL

server/src/main/java/org/apache/iotdb/db/mpp/schedule/task/FragmentInstanceID.java
```diff
@@ -0,0 +1,67 @@
...
+ * under the License.
+ */
+package org.apache.iotdb.db.mpp.schedule.task;
+
+import org.apache.iotdb.db.mpp.schedule.queue.ID;
+
+import org.jetbrains.annotations.NotNull;
+
+/** the class of id of the fragment instance */
+public class FragmentInstanceID implements ID, Comparable<FragmentInstanceTask> {
```
xingtanzjr, line 26: should be `Comparable<FragmentInstanceID>`

server/src/main/java/org/apache/iotdb/db/mpp/buffer/IDataBlockManager.java
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.iotdb.db.mpp.schedule.task.FragmentInstanceID;
+import org.apache.iotdb.db.mpp.schedule.task.FragmentInstanceTask;
+
+public interface IDataBlockManager {
+
+  /**
+   * Register a new fragment instance. The block manager will start looking for upstream data blocks
+   * and flushing data blocks generated to downstream fragment instances.
+   */
+  void registerFragmentInstance(FragmentInstanceTask task);
```
xingtanzjr, line 32: A simple question here.

Do we need FragmentInstanceTask to have some interfaces to tell the BlockManager its source and sink info ? Or we only do the registration here and the corresponding operator will trigger it later

ericpai: @xingtanzjr Sorry, I missed a commit. Please recheck it again.

server/src/main/java/org/apache/iotdb/db/mpp/schedule/task/FragmentInstanceID.java
ericpai: Fixed.

server/src/main/java/org/apache/iotdb/db/mpp/memory/MemoryPool.java
```diff
@@ -0,0 +1,90 @@
...
+
+  public long getMaxBytes() {
+    return maxBytes;
+  }
+
+  public boolean tryReserve(String queryId, long bytes) {
+    Validate.notNull(queryId);
+    Validate.isTrue(bytes > 0L);
+
+    synchronized (this) {
```
xingtanzjr, line 54: Maybe we can put this `synchronized (this)` before line 58. So that the method can return immediately if the free memory is not enough

server/src/main/java/org/apache/iotdb/db/mpp/schedule/task/FragmentInstanceTask.java
```diff
@@ -0,0 +1,196 @@
...
+      if (o1.getId().equals(o2.getId())) {
+        return 0;
+      }
+      if (o1.getDDL() < o2.getDDL()) {
+        return -1;
+      }
+      if (o1.getDDL() > o2.getDDL()) {
+        return 1;
+      }
+      return o1.getId().compareTo(o2);
```
xingtanzjr: `return o1.getId().compareTo(o2.getId());`

server/src/main/java/org/apache/iotdb/db/mpp/schedule/task/FragmentInstanceTask.java
```diff
@@ -0,0 +1,196 @@
...
+      if (o1.getId().equals(o2.getId())) {
+        return 0;
+      }
+      if (o1.getSchedulePriority() > o2.getSchedulePriority()) {
+        return -1;
+      }
+      if (o1.getSchedulePriority() < o2.getSchedulePriority()) {
+        return 1;
+      }
+      return o1.getId().compareTo(o2);
```
xingtanzjr: `return o1.getId().compareTo(o2.getId());`

ericpai: Fixed.

server/src/main/java/org/apache/iotdb/db/mpp/schedule/queue/IndexedBlockingQueue.java
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  /**
+   * Remove and return the element by id. It returns null if it doesn't exist.
+   *
+   * @param id the id of the element to be removed.
+   * @return the removed element.
+   */
+  public synchronized E remove(ID id) {
+    queryHolder.setId(id);
```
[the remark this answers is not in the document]
ericpai: That's a good point. I will add some hints in the java doc of `IDIndexedAccessible` interface.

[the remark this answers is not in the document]
ericpai: The comments has been added.

server/src/main/java/org/apache/iotdb/db/mpp/schedule/FragmentInstanceManager.java
```diff
@@ -0,0 +1,286 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(FragmentInstanceManager.class);
+
+  public static IFragmentInstanceManager getInstance() {
+    return InstanceHolder.instance;
+  }
+
+  private final IndexedBlockingQueue<FragmentInstanceTask> readyQueue;
+  private final IndexedBlockingQueue<FragmentInstanceTask> timeoutQueue;
+  private final Set<FragmentInstanceTask> blockedTasks;
+  private final Map<String, Set<FragmentInstanceTask>> queryMap;
```
xingtanzjr, line 56: We defined a class `QueryId` so maybe we can use it as the Key here

server/src/main/java/org/apache/iotdb/db/mpp/schedule/queue/L2PriorityQueue.java
```diff
@@ -0,0 +1,104 @@
...
+      workingElements = idleElements;
+      idleElements = tmp;
+    }
+    return workingElements.remove(workingElements.firstKey());
+  }
+
+  @Override
+  protected void pushToQueue(E element) {
+    workingElements.remove(element);
+    idleElements.put(element, element);
```
JackieTien97, line 79: Why we need to remove it from `workingElements` first?

server/src/main/java/org/apache/iotdb/db/mpp/schedule/FragmentInstanceManager.java
ericpai: Fixed.

server/src/main/java/org/apache/iotdb/db/mpp/schedule/FragmentInstanceManager.java
```diff
@@ -0,0 +1,286 @@
...
+      } finally {
+        task.unlock();
+      }
+    }
+  }
+
+  @Override
+  public void abortQuery(String queryId) {
+    Set<FragmentInstanceTask> queryRelatedTasks = queryMap.remove(queryId);
+    if (queryRelatedTasks != null) {
```
xingtanzjr, line 140: If the tasks for a query is being added to the `readyQueue` on line 129 and a abortQuery command for this query is triggered at the same time, the abort command may only clear part of the tasks.

I am not sure whether this scenario will lead to some unexpected results or not...

ericpai: Good catch! I have fixed by re-check the status after getting the lock in `submitFragmentInstance`.

server/src/main/java/org/apache/iotdb/db/mpp/schedule/AbstractExecutor.java
```diff
@@ -0,0 +1,76 @@
...
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.concurrent.ExecutionException;
+
+/** an abstract executor for {@link FragmentInstanceTask} */
+public abstract class AbstractExecutor extends Thread implements Closeable {
+
+  private static final Logger logger = LoggerFactory.getLogger(AbstractExecutor.class);
+  private final IndexedBlockingQueue<FragmentInstanceTask> queue;
+  private final ITaskScheduler scheduler;
```
JackieTien97, line 36: Why not change the `private` to `protected`, so that `FragmentInstanceTaskExecutor` will avoid calling `getScheduler()` method all the time.

server/src/main/java/org/apache/iotdb/db/mpp/schedule/FragmentInstanceManager.java
```diff
@@ -0,0 +1,287 @@
...
+
+  @Override
+  public void submitFragmentInstances(QueryId queryId, List<ExecFragmentInstance> instances) {
+    Set<FragmentInstanceTask> tasks =
+        instances.stream()
+            .map(
+                v ->
+                    new FragmentInstanceTask(v, QUERY_TIMEOUT_MS, FragmentInstanceTaskStatus.READY))
+            .collect(Collectors.toSet());
+    queryMap.put(queryId, Collections.synchronizedSet(tasks));
```
JackieTien97, line 125: One query may call `submitFragmentInstances` many times, so `put` here may not be right. `computeIfAbsent` plus `add` will be better.

server/src/main/java/org/apache/iotdb/db/mpp/schedule/FragmentInstanceManager.java
```diff
@@ -0,0 +1,286 @@
...
+  public void submitFragmentInstances(String queryId, List<ExecFragmentInstance> instances) {
+    Set<FragmentInstanceTask> tasks =
+        instances.stream()
+            .map(
+                v ->
+                    new FragmentInstanceTask(v, QUERY_TIMEOUT_MS, FragmentInstanceTaskStatus.READY))
+            .collect(Collectors.toSet());
+    queryMap.put(queryId, Collections.synchronizedSet(tasks));
+    for (FragmentInstanceTask task : tasks) {
+      task.lock();
```
xingtanzjr, line 127: Just a question here...

From my understanding, that put the `task.lock()` here is definitely ok. And I found that every time the task is transfer from one queue to another, we always put a `task.lock()`. But I don't come up the scenarios what will happen if we don't add the lock...

server/src/main/java/org/apache/iotdb/db/mpp/schedule/FragmentInstanceManager.java
```diff
@@ -0,0 +1,287 @@
...
+  public void submitFragmentInstances(QueryId queryId, List<ExecFragmentInstance> instances) {
+    Set<FragmentInstanceTask> tasks =
+        instances.stream()
+            .map(
+                v ->
+                    new FragmentInstanceTask(v, QUERY_TIMEOUT_MS, FragmentInstanceTaskStatus.READY))
+            .collect(Collectors.toSet());
+    queryMap.put(queryId, Collections.synchronizedSet(tasks));
+    for (FragmentInstanceTask task : tasks) {
+      task.lock();
```
JackieTien97, line 127: Why we need task.lock() here, it seems that `queue.push()` won't cause some race condition.

server/src/main/java/org/apache/iotdb/db/mpp/schedule/FragmentInstanceManager.java
ericpai: Considering a task is from running to blocked and has been timeout at the same time:
1. the working thread acquired the lock, and changed the status to `BLOCKED`, and then free the lock.
2. the working thread attempts to move it to blocking queue, but the thread is scheduled out.
3. the timeout sentinel acquired the lock, changed it to `ABORTED`, and then free the lock.
4. the timeout sentinel clear the task from any other inner queues or sets.
5. the working thread is returned to running, and add it to the blocking queue, which will be never cleaned.

ericpai: In a word, as one task should satisfy the pre-status condition before any status convertion happens, we should let the status change and queue movement as an atomic step.

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/mpp/schedule/FragmentInstanceManager.java
ericpai: Here we should re-check whether the status is `READY`, and then put them to the queues. They should act as an atomic one.

server/src/main/java/org/apache/iotdb/db/mpp/schedule/FragmentInstanceManager.java
ericpai: You mean that the coordinator will send instances of one query via more than 1 RPC requests?

server/src/main/java/org/apache/iotdb/db/mpp/schedule/FragmentInstanceManager.java
xingtanzjr: Got it. Thanks!

server/src/main/java/org/apache/iotdb/db/mpp/schedule/queue/L2PriorityQueue.java
ericpai: The design of the queue is to be randomly accessed via id, like a set. So we should ensure only one element with same ids can be in this queue.

server/src/main/java/org/apache/iotdb/db/mpp/schedule/AbstractExecutor.java
ericpai: Fixed.

server/src/main/java/org/apache/iotdb/db/mpp/schedule/FragmentInstanceManager.java
ericpai: Fixed.

server/src/main/java/org/apache/iotdb/db/mpp/memory/MemoryPool.java
cornmonster: Good catch! I guess it can be optimized like

```
if (maxBytes - reservedBytes < bytes) {
        return false;
}
synchronized (this) {
    if (maxBytes - reservedBytes < bytes) {
        return false;
    }
    reservedBytes += bytes;
    queryMemoryReservations.merge(queryId, bytes, Long::sum);
}
```

ericpai: Optimized.

server/src/main/java/org/apache/iotdb/db/mpp/schedule/FragmentInstanceManager.java
JackieTien97: It seems that `queue.push` methods don't do this check.

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/mpp/schedule/queue/L2PriorityQueue.java
ericpai: Fixed with `assert`.

server/src/main/java/org/apache/iotdb/db/mpp/schedule/FragmentInstanceManager.java
ericpai: The check has already been added.

JackieTien97 (approved):

PR #5317, opened 2022-03-22
[IOTDB-2727] async memory reservation

cornmonster:
Async memory reservation for input/output buffer management.


server/src/main/java/org/apache/iotdb/db/mpp/buffer/DataBlockServiceImpl.java
```diff
@@ -1,50 +0,0 @@
...
-
-import org.apache.iotdb.mpp.rpc.thrift.DataBlockService;
-import org.apache.iotdb.mpp.rpc.thrift.EndOfDataBlockEvent;
-import org.apache.iotdb.mpp.rpc.thrift.GetDataBlockReqest;
-import org.apache.iotdb.mpp.rpc.thrift.GetDataBlockResponse;
-import org.apache.iotdb.mpp.rpc.thrift.NewDataBlockEvent;
-
-import org.apache.thrift.TException;
-
-public class DataBlockServiceImpl implements DataBlockService.Iface {
```
JackieTien97, line 30: Why we delete this class?

server/src/main/java/org/apache/iotdb/db/mpp/memory/MemoryPool.java
```diff
@@ -79,8 +142,37 @@ public synchronized void free(String queryId, long bytes) {
     } else {
       queryMemoryReservations.put(queryId, queryReservedBytes);
     }
-
     reservedBytes -= bytes;
+
+    List<MemoryReservationFuture<Void>> futures =
```
JackieTien97, line 147: In this way, if one query free its memory, it will only notify blocking Futures belonging to this query. Other queries won't be notified and still waiting. 
And it seems that one free operation will only notify one blocking Future, but what if this freeing operation frees a large memory block which is large enough for multiple blocking Futures.

server/src/main/java/org/apache/iotdb/db/mpp/memory/MemoryPool.java
```diff
@@ -19,24 +19,60 @@
...
+    }
+  }
+
   private final String id;
   private final long maxBytes;
+  private final long maxBytesPerQuery;
 
   private long reservedBytes = 0L;
   private final Map<String, Long> queryMemoryReservations = new HashMap<>();
+  private final Map<String, List<MemoryReservationFuture<Void>>> queryIdToFutures = new HashMap<>();
```
JackieTien97, line 66: We have `QueryId` class now, you can use that class to replace `String`.

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/mpp/memory/MemoryPool.java
cornmonster: Make sense. Will fix it.

PR #5324, opened 2022-03-23
[To 0.13][IOTDB-2787] Fix aligned mem chunk concurrent problem

SilverNarcissus:
Aligned mem chunk is not synchronized properly. This pr is to fix this problem.


HTHou (approved):
LGTM

PR #5347, opened 2022-03-25
[IOTDB-2780] Config node ratis consensus protocol implementation

CRZbulabula:


confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeDescriptor.java
```diff
@@ -175,6 +175,24 @@ private void loadProps() {
               properties.getProperty(
                   "data_region_count", String.valueOf(conf.getDataRegionCount()))));
 
+      String addresses = properties.getProperty("config_node_group_address_list", null);
```
OneSizeFitsQuorum, line 178: Maybe we can abstract these method to iotdb-common so that datanode can use this logic too?

confignode/src/test/java/org/apache/iotdb/confignode/manager/ConfigManagerTest.java
```diff
@@ -0,0 +1,129 @@
...
+import org.apache.thrift.TException;
+import org.apache.thrift.protocol.TBinaryProtocol;
+import org.apache.thrift.transport.TTransport;
+import org.apache.thrift.transport.TTransportException;
+import org.junit.Assert;
+
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+public class ConfigManagerTest {
```
OneSizeFitsQuorum, line 38: Since the class is not in the Test package, might it be better to have a name that doesn't end with Test? Or just put it in the test package,  the test will not run it without the corresponding overloaded function

OneSizeFitsQuorum (commented):
PTAL

confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeDescriptor.java
CRZbulabula: Of course, I think we need a common .properties file reading process in iotdb-common. But I think this job isn't urgent currently, we could optimize both IoTDBDescriptor and ConfigNodeDescriptor later.

confignode/src/test/java/org/apache/iotdb/confignode/manager/ConfigManagerTest.java
CRZbulabula: This test is currently not run unless the corresponding test function is preceded by @Test. And I renamed it with "manual" now.

confignode/src/assembly/resources/conf/iotdb-confignode.properties
```diff
@@ -57,6 +53,23 @@ config_node_rpc_port=22277
...
 # thrift_init_buffer_size=1024
 
+####################
+### consensus protocol configuration
+####################
+
+# ConfigNodeGroup consensus protocol type
+# These consensus protocol are currently supported:
+# 1. standalone(No protocol, only supports stand-alone machine)
+# 2. ratis(Ratis protocol)
```
qiaojialin, line 63:
```suggestion
# 2. ratis(Raft protocol)
```

Users only know Raft, they do not know Ratis

confignode/src/assembly/resources/conf/iotdb-confignode.properties
```diff
@@ -57,6 +53,23 @@ config_node_rpc_port=22277
...
+# Datatype: String
+# consensus_type=standalone
```
qiaojialin, line 65: Default is Ratis?

confignode/src/main/java/org/apache/iotdb/confignode/consensus/ConsensusType.java
```diff
@@ -0,0 +1,54 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iotdb.confignode.consensus;
+
+import java.util.Arrays;
+
+public enum ConsensusType {
```
qiaojialin, line 23: put this to consensus module?

confignode/src/main/java/org/apache/iotdb/confignode/manager/ConfigManager.java
```diff
@@ -87,33 +89,80 @@ public int getDeviceGroupID(String device) {
...
-    this.consensusGroupId = new ConsensusGroupId(GroupType.PartitionRegion, 0);
-    this.consensusImpl.addConsensusGroup(
-        this.consensusGroupId,
+    // Implement specific consensus
+    switch (conf.getConsensusType()) {
+      case STANDALONE:
+        constructStandAloneConsensus();
+        break;
+      case RATIS:
+        constructRatisConsensus();
```
qiaojialin, line 115: looks like the constructors of these two consensus are not consistent

confignode/src/main/java/org/apache/iotdb/confignode/manager/ConfigManager.java
```diff
@@ -48,36 +52,34 @@
...
+  private static final ConfigNodeConf conf = ConfigNodeDescriptor.getInstance().getConf();
 
   private IConsensus consensusImpl;
   private ConsensusGroupId consensusGroupId;
 
   private DeviceGroupHashExecutor hashExecutor;
 
-  @TestOnly
-  public ConfigManager(String hashExecutorClass, int deviceGroupCount) {
-    setHashExecutor(hashExecutorClass, deviceGroupCount);
```
wangchao316, line 59: why delete?  we need only  test configManager.

confignode/src/main/java/org/apache/iotdb/confignode/consensus/ConsensusType.java
OneSizeFitsQuorum: The consensus layer provides multiple implementations, but which consensus algorithm is needed should be decided when other modules instantiate consensus classes, so it seems odd to put this on consensus?

confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeDescriptor.java
OneSizeFitsQuorum: Agree with that

confignode/src/main/java/org/apache/iotdb/confignode/manager/ConfigManager.java
CRZbulabula: Because this constructor is useless...

confignode/src/assembly/resources/conf/iotdb-confignode.properties
CRZbulabula: Set default to standalone will be easier for develop. I suppose that change this default value after our development is almost completed will be better.

confignode/src/main/java/org/apache/iotdb/confignode/consensus/ConsensusType.java
CRZbulabula: I've disscued this with @OneSizeFitsQuorum. Maintain the ConsensusType in confignode currently is ok. We will optimize that later.

confignode/src/main/java/org/apache/iotdb/confignode/manager/ConfigManager.java
CRZbulabula: Yes, because they construct different consensus protocol.

confignode/src/assembly/resources/conf/iotdb-confignode.properties
CRZbulabula: Fixed.

wangchao316 (approved):

OneSizeFitsQuorum (approved):
LGTM

qiaojialin (approved):

PR #5348, opened 2022-03-26
Ratis Consensus Impl Bug Fix

SzyWilliam:
## Description
In this PR, I fix two bugs:
1. Add volatile & synchronize to Response and Request to prevent parallel serialization race condition.
2. Use abbreviations to name Ratis Group, in case of the group-name length overflows


OneSizeFitsQuorum (approved):
LGTM

PR #5361, opened 2022-03-28
Demo for integration of ConfigNode and ratis-consensus

CRZbulabula:
Demo code for showing integration of ConfigNode and ratis-consensus.

- ratisConsensusDemo: Shows that ConfigNode can register Datanodes and set StorageGroups using the ratis consensus protocol.


OneSizeFitsQuorum (approved):
LGTM

PR #5363, opened 2022-03-28
[IOTDB-2811] Fix compaction exception handle failure cause by deletion of storage group

THUMarkLau:
The detail report see [IOTDB-2811](https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-2811?filter=allopenissues)

If user deletes the storage group when it is executing compaction, the `CompactionExceptionHandler` may fail to handle exception because it find that both source files and target files are lost. Consequently, the system will be set to **read-only** because the `CompactionExceptionHandler` takes it as a very serious situation.
The solution is to collect the `Future` object when the compaction task is submitted to thread pool, and cancel the executing of compaction task before deleting a storage group.


server/src/main/java/org/apache/iotdb/db/engine/compaction/CompactionTaskManager.java
```diff
@@ -290,35 +292,34 @@ public synchronized void removeRunningTaskFromList(AbstractCompactionTask task)
...
-        next.cancel(true);
+  public synchronized void abortCompaction(String storageGroupName) {
+    for (Pair<AbstractCompactionTask, Future<Void>> futurePair : futureList) {
+      if (futurePair.left.getFullStorageGroupName().contains(storageGroupName)) {
+        futurePair.right.cancel(true);
       }
-      subIterator.remove();
     }
+
+    candidateCompactionTaskQueue.clear();
```
choubenson, line 322: Is it necessary to remove the task from` futureList` and `runningCompactionTaskList` here?

server/src/test/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessorTest.java
```diff
@@ -705,6 +711,53 @@ public void testMerge()
...
+              true,
+              new AtomicInteger(0));
+      Future<Void> future = CompactionTaskManager.getInstance().submitTask(task);
+      Thread.sleep(20);
+      StorageEngine.getInstance().deleteStorageGroup(new PartialPath(storageGroup));
+      Thread.sleep(500);
+
+      for (TsFileResource resource : processor.getSequenceFileList()) {
+        Assert.assertTrue(resource.getTsFile().exists());
+      }
```
choubenson, line 741: Source files should not exist after deleting sg.

server/src/main/java/org/apache/iotdb/db/engine/compaction/CompactionTaskManager.java
THUMarkLau: When the compaction task finish, it will remove from the `futureList` and `runningCompactionTaskList` itself

server/src/test/java/org/apache/iotdb/db/engine/storagegroup/StorageGroupProcessorTest.java
THUMarkLau: resolve

server/src/main/java/org/apache/iotdb/db/engine/compaction/CompactionTaskManager.java
choubenson: Got it.

choubenson (approved):

tsunghanjacktsai (approved):

server/src/main/java/org/apache/iotdb/db/engine/compaction/CompactionTaskManager.java
```diff
@@ -64,8 +60,8 @@
       new FixedPriorityBlockingQueue<>(1024, new CompactionTaskComparator());
   // <logicalStorageGroupName,futureSet>, it is used to terminate all compaction tasks under the
```
qiaojialin: logicalstoragegroup-virtualstoragegroup=fullsgname

qiaojialin (approved):

PR #5364, opened 2022-03-28
Connect the Query Plan stage from SQL to DistributionPlan with UT and mock interfaces

xingtanzjr:
## Description

* Polish the PlanNode and supply unimplemented method for some PlanNode.
* Add a method to print the PlanNode tree in a more friendly way
* Prepare a UT to represent the stage from SQL to DistributedPlan. See the following snapshot.
image
image


server/src/main/java/org/apache/iotdb/db/mpp/common/schematree/SchemaTree.java
```diff
@@ -36,12 +39,47 @@
...
+          lastCharIsStar = c == '*';
+          continue;
+        }
+        lastCharIsStar = false;
+        noStar.append(String.valueOf(c));
+      }
+
+      for (String path : paths) {
+        if (path.contains(noStar)) {
+          result.add(new MeasurementPath(path));
```
liuminghui233, line 75:
```suggestion
          result.add(new MeasurementPath(path, new MeasurementSchema(path.getMeasurement(), TSDataType.DOUBLE)));
```
It is better to mock the data type as well.

liuminghui233 (changes requested):

PR #5367, opened 2022-03-29
[IOTDB-2727] data block manager impl

cornmonster:
Design: https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=199538158

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
+++ b/server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
@@ -826,6 +826,18 @@ public class IoTDBConfig {
   /** The max time of data node waiting to join into the cluster */
   private long joinClusterTimeOutMs = TimeUnit.SECONDS.toMillis(60);
 
+  /** Port that data block manager thrift service listen to. */
+  private int dataBlockManagerPort = 7777;
+
+  /** Core pool size of data block manager. */
+  private int dataBlockManagerCorePoolSize = 1;
+
+  /** Max pool size of data block manager. */
+  private int dataBlockManagerMaxPoolSize = 5;
+
+  /** Thread keep alive time in ms of data block manager. */
+  private int dataBlockManagerKeepAliveTimeInMs = 1000;
+
   public IoTDBConfig() {
     try {
       internalIp = InetAddress.getLocalHost().getHostAddress();
@@ -2585,5 +2597,37 @@ public class IoTDBConfig {
 
   public void setJoinClusterTimeOutMs(long joinClusterTimeOutMs) {
     this.joinClusterTimeOutMs = joinClusterTimeOutMs;
+  }
+
+  public int getDataBlockManagerPort() {
+    return dataBlockManagerPort;
+  }
+
+  public void setDataBlockManagerPort(int dataBlockManagerPort) {
+    this.dataBlockManagerPort = dataBlockManagerPort;
+  }
+
+  public int getDataBlockManagerCorePoolSize() {
+    return dataBlockManagerCorePoolSize;
+  }
+
+  public void setDataBlockManagerCorePoolSize(int dataBlockManagerCorePoolSize) {
+    this.dataBlockManagerCorePoolSize = dataBlockManagerCorePoolSize;
+  }
+
+  public int getDataBlockManagerMaxPoolSize() {
+    return dataBlockManagerMaxPoolSize;
+  }
+
+  public void setDataBlockManagerMaxPoolSize(int dataBlockManagerMaxPoolSize) {
+    this.dataBlockManagerMaxPoolSize = dataBlockManagerMaxPoolSize;
+  }
+
+  public int getDataBlockManagerKeepAliveTimeInMs() {
+    return dataBlockManagerKeepAliveTimeInMs;
+  }
+
+  public void setDataBlockManagerKeepAliveTimeInMs(int dataBlockManagerKeepAliveTimeInMs) {
+    this.dataBlockManagerKeepAliveTimeInMs = dataBlockManagerKeepAliveTimeInMs;
   }
 }
--- a/server/src/main/java/org/apache/iotdb/db/mpp/buffer/DataBlockManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/mpp/buffer/DataBlockManager.java
@@ -131,7 +131,7 @@ public class DataBlockManager implements IDataBlockManager {
           sourceHandles
               .getOrDefault(e.getTargetFragmentInstanceId(), Collections.emptyMap())
               .get(e.getTargetOperatorId());
-      sourceHandle.setNoMoreTsBlocks();
+      sourceHandle.setNoMoreTsBlocks(e.getLastSequenceId());
     }
   }
 
--- a/server/src/main/java/org/apache/iotdb/db/mpp/buffer/DataBlockManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/mpp/buffer/DataBlockManager.java
@@ -62,6 +62,7 @@ public class DataBlockManager implements IDataBlockManager {
   /** Handle thrift communications. */
   class DataBlockServiceImpl implements DataBlockService.Iface {
 
+    @Override
     public GetDataBlockResponse getDataBlock(GetDataBlockRequest req) throws TException {
       logger.debug(
           "Get data block request received. Asking for data block [{}, {}) from {}.",
@@ -75,11 +76,9 @@ public class DataBlockManager implements IDataBlockManager {
                 + ".");
       }
       GetDataBlockResponse resp = new GetDataBlockResponse();
+      SinkHandle sinkHandle = sinkHandles.get(req.getSourceFragnemtInstanceId());
       for (int i = req.getStartSequenceId(); i < req.getEndSequenceId(); i++) {
-        ByteBuffer serializedTsBlock =
-            sinkHandles
-                .get(req.getSourceFragnemtInstanceId())
-                .getSerializedTsBlock(i);
+        ByteBuffer serializedTsBlock = sinkHandle.getSerializedTsBlock(i);
         resp.addToTsBlocks(serializedTsBlock);
       }
       return resp;
@@ -100,8 +99,10 @@ public class DataBlockManager implements IDataBlockManager {
               .get(e.getTargetFragmentInstanceId())
               .get(e.getTargetOperatorId())
               .isClosed()) {
-        logger.info("{} is ignored because the source handle does not exist or has been closed", e);
-        return;
+        throw new TException(
+            "Target fragment instance not found. Fragment instance ID: "
+                + e.getTargetFragmentInstanceId()
+                + ".");
       }
 
       SourceHandle sourceHandle =
@@ -124,8 +125,10 @@ public class DataBlockManager implements IDataBlockManager {
               .get(e.getTargetFragmentInstanceId())
               .get(e.getTargetOperatorId())
               .isClosed()) {
-        logger.info("{} is ignored because the source handle does not exist or has been closed", e);
-        return;
+        throw new TException(
+            "Target fragment instance not found. Fragment instance ID: "
+                + e.getTargetFragmentInstanceId()
+                + ".");
       }
       SourceHandle sourceHandle =
           sourceHandles
--- a/server/src/main/java/org/apache/iotdb/db/mpp/buffer/DataBlockManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/mpp/buffer/DataBlockManager.java
@@ -20,6 +20,7 @@
 package org.apache.iotdb.db.mpp.buffer;
 
 import org.apache.iotdb.db.mpp.memory.LocalMemoryManager;
+import org.apache.iotdb.mpp.rpc.thrift.AcknowledgeDataBlockEvent;
 import org.apache.iotdb.mpp.rpc.thrift.DataBlockService;
 import org.apache.iotdb.mpp.rpc.thrift.EndOfDataBlockEvent;
 import org.apache.iotdb.mpp.rpc.thrift.GetDataBlockRequest;
@@ -65,18 +66,18 @@ public class DataBlockManager implements IDataBlockManager {
     @Override
     public GetDataBlockResponse getDataBlock(GetDataBlockRequest req) throws TException {
       logger.debug(
-          "Get data block request received. Asking for data block [{}, {}) from {}.",
+          "Get data block request received, for data blocks whose sequence ID in [{}, {}) from {}.",
           req.getStartSequenceId(),
           req.getEndSequenceId(),
-          req.getSourceFragnemtInstanceId());
-      if (!sinkHandles.containsKey(req.getSourceFragnemtInstanceId())) {
+          req.getSourceFragmentInstanceId());
+      if (!sinkHandles.containsKey(req.getSourceFragmentInstanceId())) {
         throw new TException(
             "Source fragment instance not found. Fragment instance ID: "
-                + req.getSourceFragnemtInstanceId()
+                + req.getSourceFragmentInstanceId()
                 + ".");
       }
       GetDataBlockResponse resp = new GetDataBlockResponse();
-      SinkHandle sinkHandle = sinkHandles.get(req.getSourceFragnemtInstanceId());
+      SinkHandle sinkHandle = sinkHandles.get(req.getSourceFragmentInstanceId());
       for (int i = req.getStartSequenceId(); i < req.getEndSequenceId(); i++) {
         ByteBuffer serializedTsBlock = sinkHandle.getSerializedTsBlock(i);
         resp.addToTsBlocks(serializedTsBlock);
@@ -85,12 +86,30 @@ public class DataBlockManager implements IDataBlockManager {
     }
 
     @Override
+    public void onAcknowledgeDataBlockEvent(AcknowledgeDataBlockEvent e) throws TException {
+      logger.debug(
+          "Acknowledge data block event received, for data blocks whose sequence ID in [{}, {}) from {}.",
+          e.getStartSequenceId(),
+          e.getEndSequenceId(),
+          e.getSourceFragmentInstanceId());
+      if (!sinkHandles.containsKey(e.getSourceFragmentInstanceId())) {
+        throw new TException(
+            "Source fragment instance not found. Fragment instance ID: "
+                + e.getSourceFragmentInstanceId()
+                + ".");
+      }
+      sinkHandles
+          .get(e.getSourceFragmentInstanceId())
+          .acknowledgeTsBlock(e.getStartSequenceId(), e.getEndSequenceId());
+    }
+
+    @Override
     public void onNewDataBlockEvent(NewDataBlockEvent e) throws TException {
       logger.debug(
           "New data block event received, for operator {} of {} from {}.",
           e.getTargetOperatorId(),
           e.getTargetFragmentInstanceId(),
-          e.getSourceFragnemtInstanceId());
+          e.getSourceFragmentInstanceId());
       if (!sourceHandles.containsKey(e.getTargetFragmentInstanceId())
           || !sourceHandles
               .get(e.getTargetFragmentInstanceId())
@@ -116,7 +135,7 @@ public class DataBlockManager implements IDataBlockManager {
           "End of data block event received, for operator {} of {} from {}.",
           e.getTargetOperatorId(),
           e.getTargetFragmentInstanceId(),
-          e.getSourceFragnemtInstanceId());
+          e.getSourceFragmentInstanceId());
       if (!sourceHandles.containsKey(e.getTargetFragmentInstanceId())
           || !sourceHandles
               .get(e.getTargetFragmentInstanceId())
--- a/server/src/main/java/org/apache/iotdb/db/mpp/buffer/DataBlockManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/mpp/buffer/DataBlockManager.java
@@ -79,7 +79,7 @@ public class DataBlockManager implements IDataBlockManager {
         ByteBuffer serializedTsBlock =
             sinkHandles
                 .get(req.getSourceFragnemtInstanceId())
-                .getSerializedTsBlock(req.getStartSequenceId());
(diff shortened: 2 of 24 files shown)


server/src/main/java/org/apache/iotdb/db/mpp/memory/MemoryPool.java
```diff
@@ -79,11 +141,35 @@ public synchronized void free(String queryId, long bytes) {
...
+      MemoryReservationFuture<Void> future = iterator.next();
+
+      if (future.isCancelled()) {
+        iterator.remove();
+        continue;
+      }
+
+      long bytesToReserve = future.getBytes();
+      if (maxBytes - reservedBytes < bytesToReserve
+          || maxBytesPerQuery - queryMemoryReservations.getOrDefault(queryId, 0L)
```
xingtanzjr, line 160: It seems that `MemoryReservationFuture` doesn't know which query it belongs to. In other words, the `future` and the `queryId` may represent different query. So we cannot ensure the queryMemoryReservations of the `future` does not exceed the `maxBytesPerQuery `

server/src/main/java/org/apache/iotdb/db/mpp/buffer/DataBlockManager.java
```diff
@@ -19,98 +19,296 @@
...
+
+    void onClosed(SinkHandle sinkHandle);
+
+    void onAborted(SinkHandle sinkHandle);
+  }
+
+  /** Handle thrift communications. */
+  class DataBlockServiceImpl implements DataBlockService.Iface {
+
+    public GetDataBlockResponse getDataBlock(GetDataBlockRequest req) throws TException {
```
JackieTien97, line 65:
```suggestion
    @Override
    public GetDataBlockResponse getDataBlock(GetDataBlockRequest req) throws TException {
```

server/src/main/java/org/apache/iotdb/db/mpp/buffer/SinkHandle.java
```diff
@@ -0,0 +1,346 @@
...
+      retainedSizeInBytes += tsBlock.getRetainedSizeInBytes();
+    }
+    int currentEndSequenceId;
+    List<Long> tsBlockSizes = new ArrayList<>();
+    synchronized (this) {
+      currentEndSequenceId = bufferedTsBlocks.size();
+      blocked =
+          localMemoryManager
+              .getQueryPool()
+              .reserve(localFragmentInstanceId.getQueryId(), retainedSizeInBytes);
```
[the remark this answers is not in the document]
cornmonster: Yes. Since SinkHandle provides `isFull` for callers to decide if they are allowed to call `send`, it is safe to assume that the listenable future `blocked` is completed.

server/src/main/java/org/apache/iotdb/db/mpp/memory/MemoryPool.java
cornmonster: Make sense. Will fix it.

server/src/main/java/org/apache/iotdb/db/mpp/memory/LocalMemoryManager.java
```diff
@@ -32,8 +32,8 @@ public LocalMemoryManager() {
     long maxMemory = Runtime.getRuntime().maxMemory();
     // Save 20% memory for untracked allocations.
     maxBytes = (long) (maxMemory * 0.8);
-    // Allocate 50% memory for query execution.
-    queryPool = new MemoryPool("query", (long) (maxBytes * 0.5));
+    // Allocate 50% memory for query execution. Each query can allocate up to 30% memory.
+    queryPool = new MemoryPool("query", (long) (maxBytes * 0.5), (long) (maxMemory * 0.3));
```
JackieTien97, line 36: You can get the query memory limit from `IoTDBConfig.allocateMemoryForRead`

server/src/main/java/org/apache/iotdb/db/mpp/buffer/SinkHandle.java
```diff
@@ -0,0 +1,346 @@
...
+  private final String remoteOperatorId;
+  private final TFragmentInstanceId localFragmentInstanceId;
+  private final LocalMemoryManager localMemoryManager;
+  private final ExecutorService executorService;
+  private final DataBlockService.Client client;
+  private final TsBlockSerde serde;
+  private final SinkHandleListener sinkHandleListener;
+
+  // TODO: a better data structure to hold tsblocks
+  private final List<TsBlock> bufferedTsBlocks = new LinkedList<>();
```
JackieTien97, line 64: I think it will be better use `TsBlockWrapper` which contains `TsBlock` and `long` indicating its sequence and then each time the sourceHandle call `getDataBlock`, we can iterate from the beginning of this linkedList and remove the `TsBlockWrapper` whose `sequenceId` is before `startSequenceId` of this request.

JackieTien97: In this way, you can avoid random index access of `LinkedList` which is not very efficient. And also, avoid just setting null for consumed TsBlock.

server/src/main/java/org/apache/iotdb/db/mpp/buffer/SinkHandle.java
```diff
@@ -0,0 +1,346 @@
...
+  }
+
+  ByteBuffer getSerializedTsBlock(int sequenceId) {
+    TsBlock tsBlock;
+    synchronized (this) {
+      tsBlock = bufferedTsBlocks.get(sequenceId);
+      if (tsBlock == null) {
+        throw new IllegalStateException("The data block doesn't exist. Sequence ID: " + sequenceId);
+      }
+      bufferedTsBlocks.set(sequenceId, null);
```
JackieTien97, line 256: Setting null here is not safe, what if this response failed, so you can try the above way which can avoid this.

server/src/main/java/org/apache/iotdb/db/mpp/buffer/DataBlockService.java
```diff
@@ -38,8 +50,25 @@ public ThriftService getImplementation() {
   @Override
   public void initTProcessor()
       throws ClassNotFoundException, IllegalAccessException, InstantiationException {
-    impl = new DataBlockServiceImpl();
-    processor = new Processor<>(impl);
+    executorService =
+        IoTDBThreadPoolFactory.newCachedThreadPoolWithDaemon("data-block-manager-task-executors");
```
JackieTien97, line 54: `newCachedThreadPoolWithDaemon`'s `corePoolSize` is `0` and `maximumPoolSize` is `Integer.MAX_VALUE`. It's better to use FixedCountThreadPool or set the `corePoolSize` and `maximumPoolSize` in `CachedThreadPool` to a reasonable value.

server/src/main/java/org/apache/iotdb/db/mpp/buffer/SinkHandle.java
```diff
@@ -0,0 +1,346 @@
...
+              .reserve(localFragmentInstanceId.getQueryId(), retainedSizeInBytes);
+      bufferRetainedSizeInBytes += retainedSizeInBytes;
+      bufferedTsBlocks.addAll(tsBlocks);
+      numOfBufferedTsBlocks += tsBlocks.size();
+      for (int i = currentEndSequenceId; i < currentEndSequenceId + tsBlocks.size(); i++) {
+        tsBlockSizes.add(bufferedTsBlocks.get(i).getRetainedSizeInBytes());
+      }
+    }
+
+    submitSendNewDataBlockEventTask(currentEndSequenceId, tsBlockSizes);
```
JackieTien97, line 143: Currently, each time we will only send one TsBlock, if in this way, each calling `send` will submit one `SendNewDataBlockEvent` causing a network communication which is not very efficient.

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/mpp/buffer/SourceHandle.java
```diff
@@ -19,72 +19,286 @@
...
     TsBlock tsBlock = bufferedTsBlocks.poll();
     if (tsBlock != null) {
-      bufferRetainedSizeInBytes -= getRetainedSizeInBytes(tsBlock);
+      bufferRetainedSizeInBytes -= tsBlock.getRetainedSizeInBytes();
+      localMemoryManager
+          .getQueryPool()
+          .free(localFragmentInstanceId.getQueryId(), tsBlock.getRetainedSizeInBytes());
     }
-    if (bufferedTsBlocks.isEmpty() && !finished && blocked.isDone()) {
+    if (bufferedTsBlocks.isEmpty() && !isFinished() && blocked.isDone()) {
```
xingtanzjr, line 117: It seems that the `blocked.isDone()` is always `true` here

server/src/main/java/org/apache/iotdb/db/mpp/buffer/DataBlockManager.java
```diff
@@ -19,98 +19,296 @@
...
+      if (!sourceHandles.containsKey(e.getTargetFragmentInstanceId())
+          || !sourceHandles
+              .get(e.getTargetFragmentInstanceId())
+              .containsKey(e.getTargetOperatorId())
+          || sourceHandles
+              .get(e.getTargetFragmentInstanceId())
+              .get(e.getTargetOperatorId())
+              .isClosed()) {
+        logger.info("{} is ignored because the source handle does not exist or has been closed", e);
+        return;
```
xingtanzjr, line 104: If the NewDataBlockEvent is omitted because the corresponding SourceHandle is not created yet, the SourceHandle will keep the `blocked` status after being creating in the future until next event comes.

And if the SinkHandle cannot send the next event due to memory limit or there is no more TsBlocks to be sent, the SourceHandle may be blocked permanently...

Maybe we can create the SourceHandle and put it in the map if it hasn't be created when receiving the NewDataBlockEvent

server/src/main/java/org/apache/iotdb/db/mpp/memory/MemoryPool.java
cornmonster: Fixed.

server/src/main/java/org/apache/iotdb/db/mpp/buffer/SourceHandle.java
```diff
@@ -19,72 +19,286 @@
...
+      if (reserved) {
+        currSequenceId += 1;
+        reservedBytes += bytesToReserve;
+        bufferRetainedSizeInBytes += bytesToReserve;
+      } else {
+        break;
+      }
+    }
+
+    if (currSequenceId > startSequenceId) {
```
xingtanzjr, line 152: If we cannot get enough memory in one try, it seems that there is no more event to notify the SourceHandle to try again if the memory is released

server/src/main/java/org/apache/iotdb/db/mpp/buffer/SinkHandle.java
xingtanzjr: Furthermore, the `bufferedTsBlocks`'s size will continue to grow with the sequenceId...

server/src/main/java/org/apache/iotdb/db/mpp/buffer/DataBlockManager.java
cornmonster: Good catch. I forget to change this after I changed the transmission protocol.

> If the NewDataBlockEvent is omitted because the corresponding SourceHandle is not created yet, the SourceHandle will keep the blocked status after being creating in the future until next event comes.
True. Actually, it is not likely to happen since the series scan operator takes some time to produce data for downstream instances, I believe instances should be available then.

> And if the SinkHandle cannot send the next event due to memory limit or there is no more TsBlocks to be sent, the SourceHandle may be blocked permanently...
The SinkHandle will attempt to send the events at most 3 times. If it doesn't work, the SinkHandle throws an 
I/O exception on next invocation of any method, telling the caller that there is something wrong.

server/src/main/java/org/apache/iotdb/db/mpp/buffer/DataBlockService.java
```diff
@@ -38,8 +50,25 @@ public ThriftService getImplementation() {
...
+    clientFactory = new DataBlockServiceClientFactory();
+    this.dataBlockManager =
+        new DataBlockManager(
+            localMemoryManager, tsBlockSerdeFactory, executorService, clientFactory);
+    processor = new Processor<>(dataBlockManager.getOrCreateDataBlockServiceImpl());
```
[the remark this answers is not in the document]
cornmonster: getOrCreateDataBlockServiceImpl indicates the DataBlockServiceImpl will be reused internally, while createDataBlockServiceImpl indicates there will be a new DataBlockServiceImpl created on each invocation.

I believe getOrCreateDataBlockServiceImpl describes the behavior of DataBlockManager, right?

server/src/main/java/org/apache/iotdb/db/mpp/buffer/SourceHandle.java
cornmonster: trySubmitGetDataBlocksTask will be called in 2 cases:

1. new data blocks generated by upstream instance -> see SourceHandle#updatePendingDataBlockInfo
2. memory released -> see SourceHandle#receive

server/src/main/java/org/apache/iotdb/db/mpp/buffer/DataBlockService.java
cornmonster: Agree. But the `reasonable` value needs some further testing. What about setting the `corePoolSize` to 0 and `maximumPoolSize` to 10 for now?

server/src/main/java/org/apache/iotdb/db/mpp/buffer/SinkHandle.java
cornmonster: It is a dilemma of network efficiency or query latency. The current implementation trades the network for better latency.

server/src/main/java/org/apache/iotdb/db/mpp/buffer/DataBlockService.java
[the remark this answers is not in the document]
JackieTien97: Fine

server/src/main/java/org/apache/iotdb/db/mpp/buffer/DataBlockService.java
JackieTien97: you can set both `corePoolSize` and `maximumPoolSize` to 4 and to make it configurable, you can define it in the `iotdb-engine.properties`.

server/src/main/java/org/apache/iotdb/db/mpp/buffer/SinkHandle.java
cornmonster: Agree. LinkedList was for the POC. I will replace it with some other data structure.

server/src/main/java/org/apache/iotdb/db/mpp/buffer/SinkHandle.java
cornmonster: Also, the size of a TsBlock should be limited by an upper bound, like 1MB. But I believe the sink operator may produce more than that at each time. What do you think?

server/src/main/java/org/apache/iotdb/db/mpp/buffer/SinkHandle.java
cornmonster: Fixed.

server/src/main/java/org/apache/iotdb/db/mpp/buffer/SinkHandle.java
cornmonster: Fixed.

server/src/main/java/org/apache/iotdb/db/mpp/buffer/SinkHandle.java
cornmonster: TODO added.

server/src/main/java/org/apache/iotdb/db/mpp/buffer/DataBlockService.java
cornmonster: Fixed.

server/src/main/java/org/apache/iotdb/db/mpp/memory/LocalMemoryManager.java
cornmonster: Fixed.

server/src/main/java/org/apache/iotdb/db/mpp/buffer/SourceHandle.java
cornmonster: Fixed.

server/src/main/java/org/apache/iotdb/db/mpp/memory/MemoryPool.java
```diff
@@ -79,11 +147,35 @@ public synchronized void free(String queryId, long bytes) {
...
+      if (future.isCancelled()) {
+        iterator.remove();
+        continue;
+      }
+
+      long bytesToReserve = future.getBytes();
+      if (maxBytes - reservedBytes < bytesToReserve
+          || maxBytesPerQuery - queryMemoryReservations.getOrDefault(future.getQueryId(), 0L)
+              < bytesToReserve) {
+        return;
```
xingtanzjr, line 168: We need to separate these two conditions here. If the maxBytes is not hit but the maxBytesPerQuery is hit, we need to continue to judge whether next future is satisfied or not because next future may belong to another query.

```
if (maxBytes - reservedBytes < bytesToReserve) {
   return;
}
if (maxBytesPerQuery - queryMemoryReservations.getOrDefault(future.getQueryId(), 0L)
              < bytesToReserve) {
   continue;
}
```

server/src/main/java/org/apache/iotdb/db/mpp/buffer/SourceHandle.java
```diff
@@ -19,72 +19,389 @@
...
+                    sequenceIdOfUnReservedDataBlock,
+                    sequenceIdOfUnReservedDataBlock + 1,
+                    sizeOfUnReservedDataBlock));
+            numActiveGetDataBlocksTask += 1;
+            bufferRetainedSizeInBytes += sizeOfUnReservedDataBlock;
+          },
+          executorService);
+
+      // Schedule another call of trySubmitGetDataBlocksTask for the rest of blocks.
+      future.addListener(SourceHandle.this::trySubmitGetDataBlocksTask, executorService);
```
xingtanzjr, line 196: If we schedule another `trySubmitGetDataBlocksTask` task once the future is done. Is `line 181-193` not necessary ?

And on the other hand, if `trySubmitGetDataBlocksTask` runs before the 186, the total memory reserved may exceed the threshold.

server/src/main/java/org/apache/iotdb/db/mpp/buffer/SourceHandle.java
```diff
@@ -19,72 +19,389 @@
...
+      final int sequenceIdOfUnReservedDataBlock = endSequenceId;
+      final long sizeOfUnReservedDataBlock = sequenceIdToDataBlockSize.get(endSequenceId);
+      future.addListener(
+          () -> {
+            executorService.submit(
+                new GetDataBlocksTask(
+                    sequenceIdOfUnReservedDataBlock,
+                    sequenceIdOfUnReservedDataBlock + 1,
+                    sizeOfUnReservedDataBlock));
+            numActiveGetDataBlocksTask += 1;
```
xingtanzjr, line 190: need `reservedBytes += bytesToReserve;` here

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/mpp/buffer/SourceHandle.java
cornmonster: MemoryPool will increase the reserved amount of memory before marking the future as completed. Thus, the next call of `trySubmitGetDataBlocksTask ` is safe.

server/src/main/java/org/apache/iotdb/db/mpp/memory/MemoryPool.java
cornmonster: Fixed.

server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
```diff
@@ -826,6 +826,18 @@
...
+  private int dataBlockManagerPort = 7777;
+
+  /** Core pool size of data block manager. */
+  private int dataBlockManagerCorePoolSize = 1;
+
+  /** Max pool size of data block manager. */
+  private int dataBlockManagerMaxPoolSize = 5;
+
+  /** Thread keep alive time in ms of data block manager. */
+  private int dataBlockManagerKeepAliveTimeInMs = 1000;
```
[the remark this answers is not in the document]
cornmonster: Done.

thrift/src/main/thrift/mpp.thrift
```diff
@@ -27,23 +27,34 @@ struct TFragmentInstanceId {
...
+  3: required TFragmentInstanceId sourceFragmentInstanceId
+  4: required i32 startSequenceId
+  5: required list<i64> blockSizes
 }
 
 struct EndOfDataBlockEvent {
-  1: required TFragmentInstanceId fragmentInstanceId
-  2: required string operatorId
+  1: required TFragmentInstanceId targetFragmentInstanceId
+  2: required string targetOperatorId
```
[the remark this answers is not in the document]
cornmonster: Done.

thrift/src/main/thrift/mpp.thrift
```diff
@@ -27,23 +27,34 @@ struct TFragmentInstanceId {
...
+  2: required i32 startSequenceId
+  3: required i32 endSequenceId
+}
+
 struct NewDataBlockEvent {
-  1: required TFragmentInstanceId fragmentInstanceId
-  2: required string operatorId
-  3: required i64 blockId
+  1: required TFragmentInstanceId targetFragmentInstanceId
+  2: required string targetOperatorId
```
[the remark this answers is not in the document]
cornmonster: Done.

xingtanzjr (approved):

PR #5370, opened 2022-03-29
 Fix NoClassDefFound CI error

JackieTien97:
`jacoco` in `maven` which is used to calculate code coverage will add some synthetic fields in class, you can see in the following picture:
image
the last two fields: boolean[] and `$jacocoData` is not defined in my class, so when I iterate the fields in this class using `Class<?>.getDeclaredFields()` method, we will get this unexpected one.


ericpai:
LGTM

ericpai (approved):

PR #5380, opened 2022-03-30
[IOTDB-2826]Unmark storage group among templates when deleted

bigreybear:
In Template, we mark which storage group this template is set, this is act as a cache to accelerate related traverse.
When a storage group is deleted, we need to discard this cache in related templates.


MarcosZyk (approved):
LGTM~

PR #5382, opened 2022-03-30
[IOTDB-2650] Tablet supports adding String value

jiazhiren:
This PR creates syntactic sugar by supporting String value insertion to Tablet. #[IOTDB-2650]

### DIFF
--- a/tsfile/src/main/java/org/apache/iotdb/tsfile/write/record/Tablet.java
+++ b/tsfile/src/main/java/org/apache/iotdb/tsfile/write/record/Tablet.java
@@ -140,8 +140,7 @@ public class Tablet {
           Binary[] sensor = (Binary[]) values[indexOfSchema];
           if (value instanceof Binary)
             sensor[rowIndex] = value != null ? (Binary) value : Binary.EMPTY_VALUE;
-          else
-            sensor[rowIndex] = value != null ? new Binary((String) value) : Binary.EMPTY_VALUE;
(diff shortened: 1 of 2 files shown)


tsfile/src/main/java/org/apache/iotdb/tsfile/write/record/Tablet.java
```diff
@@ -138,7 +138,7 @@ private void addValueOfDataType(
       case TEXT:
         {
           Binary[] sensor = (Binary[]) values[indexOfSchema];
-          sensor[rowIndex] = value != null ? (Binary) value : Binary.EMPTY_VALUE;
+          sensor[rowIndex] = value != null ? new Binary((String) value) : Binary.EMPTY_VALUE;
```
HTHou, line 141: Thanks for your contribution!

IMO, we should also support adding a Binary value here to keep the compatibility. :)

HTHou: Besides, consider adding some tests in `IoTDBSessionSimpleIT`?

jiazhiren: Thanks for your valuable advice! I will work on them immediately.

HTHou:
Hi, you can execute `mvn spotless:apply` to fix the test issue.

jiazhiren:
> 

Thanks for your tip. The code format has been corrected.

HTHou (approved):

PR #5383, opened 2022-03-31
Implement SchemaTree Search

MarcosZyk:
## Description

Implement the local search interface of SchemaTree via SchemaTreeVisitor. Now it support * and **, slimit and offset, and alias.

## Future work
Implement SchemaTree serialize and deserialize, and ISchemaFetcher.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/mpp/common/schematree/SchemaNode.java
+++ b/server/src/main/java/org/apache/iotdb/db/mpp/common/schematree/SchemaNode.java
@@ -58,10 +58,10 @@ public abstract class SchemaNode {
   }
 
   public SchemaEntityNode getAsEntityNode() {
-    throw new ClassCastException("This not isn't instance of SchemaEntityNode.");
+    throw new UnsupportedOperationException("This not isn't instance of SchemaEntityNode.");
   }
 
   public SchemaMeasurementNode getAsMeasurementNode() {
-    throw new ClassCastException("This not isn't instance of SchemaMeasurementNode.");
+    throw new UnsupportedOperationException("This not isn't instance of SchemaMeasurementNode.");
   }
 }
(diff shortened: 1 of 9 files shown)


server/src/main/java/org/apache/iotdb/db/mpp/common/schematree/SchemaNode.java
```diff
@@ -19,7 +19,49 @@
...
+  public boolean isEntity() {
+    return false;
+  }
+
+  public boolean isMeasurement() {
+    return false;
+  }
+
+  public SchemaEntityNode getAsEntityNode() {
+    throw new ClassCastException("This not isn't instance of SchemaEntityNode.");
```
SilverNarcissus, line 61: Usually it's should be a UnsupportedOperationException.

server/src/main/java/org/apache/iotdb/db/mpp/common/schematree/SchemaTreeVisitor.java
```diff
@@ -0,0 +1,223 @@
...
+    if (!hasNext()) {
+      throw new NoSuchElementException();
+    }
+    MeasurementPath result = generateMeasurementPath();
+    nextMatchedNode = null;
+    return result;
+  }
+
+  public List<MeasurementPath> getAllResult() {
+    List<MeasurementPath> result = new LinkedList<>();
```
SilverNarcissus, line 88: Why we use a LinkedList rather than ArrayList here?

SilverNarcissus (approved):
Nice work~  Only minor issues, please have a look.

server/src/main/java/org/apache/iotdb/db/mpp/common/schematree/SchemaNode.java
MarcosZyk: Fixed.

liuminghui233 (approved):
LGTM 👍

server/src/main/java/org/apache/iotdb/db/mpp/common/schematree/SchemaTreeVisitor.java
MarcosZyk: Fixed. I thought that since there won't be random access on this list, it will be better to use LinkedList, which will avoid Array copy during add.

bigreybear (approved):
LGTM~

PR #5385, opened 2022-03-31
[IOTDB-2828] Update iotdb_version after upgrading finished

ericpai:
Currently after upgrading from 0.12 or 0.13 to newest version, the system.properties file is updated, but the `iotdb_version` still remains the same. It may cause the future upgrading incorrectly.


ericpai:
@HTHou @qiaojialin PTAL

HTHou (approved):
LGTM

HTHou:
@HeimingZ PTAL

HeimingZ:
LGTM

PR #5386, opened 2022-03-31
Add interface in ConfigIService

liuminghui233:
## Description

Add some interfaces in ConfigIService, which are used in IPartitionFetcher.

### DIFF
--- a/confignode/src/main/java/org/apache/iotdb/confignode/service/thrift/server/ConfigNodeRPCServerProcessor.java
+++ b/confignode/src/main/java/org/apache/iotdb/confignode/service/thrift/server/ConfigNodeRPCServerProcessor.java
@@ -32,6 +32,7 @@ import org.apache.iotdb.confignode.rpc.thrift.DataNodeMessage;
 import org.apache.iotdb.confignode.rpc.thrift.DataNodeRegisterReq;
 import org.apache.iotdb.confignode.rpc.thrift.DataNodeRegisterResp;
 import org.apache.iotdb.confignode.rpc.thrift.DataPartitionInfo;
+import org.apache.iotdb.confignode.rpc.thrift.DataPartitionInfoResp;
 import org.apache.iotdb.confignode.rpc.thrift.DeleteStorageGroupReq;
 import org.apache.iotdb.confignode.rpc.thrift.DeviceGroupHashInfo;
 import org.apache.iotdb.confignode.rpc.thrift.GetDataPartitionReq;
@@ -39,6 +40,7 @@ import org.apache.iotdb.confignode.rpc.thrift.GetSchemaPartitionReq;
 import org.apache.iotdb.confignode.rpc.thrift.SchemaPartitionInfo;
 import org.apache.iotdb.confignode.rpc.thrift.SetStorageGroupReq;
 import org.apache.iotdb.confignode.rpc.thrift.StorageGroupMessage;
+import org.apache.iotdb.confignode.rpc.thrift.fetchDataPartitionReq;
 import org.apache.iotdb.consensus.common.Endpoint;
 import org.apache.iotdb.consensus.common.response.ConsensusReadResponse;
 import org.apache.iotdb.consensus.common.response.ConsensusWriteResponse;
@@ -157,6 +159,11 @@ public class ConfigNodeRPCServerProcessor implements ConfigIService.Iface {
   }
 
   public void handleClientExit() {}
+
+  @Override
+  public DataPartitionInfoResp getDataPartitionInfo(fetchDataPartitionReq req) throws TException {
+    return null;
+  }
 
   // TODO: Interfaces for data operations
 }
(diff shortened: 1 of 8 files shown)


thrift-confignode/src/main/thrift/confignode.thrift
```diff
@@ -21,6 +21,11 @@ include "rpc.thrift"
 namespace java org.apache.iotdb.confignode.rpc.thrift
 namespace py iotdb.thrift.confignode
 
+struct EndPoint {
```
CRZbulabula, line 24: The EndPoint is already defined in rpc.thrift. And I've included it before. Why not just use it?

CRZbulabula:
I noticed that there are more interfaces defined in IPartitionFetcher. Why not add them to the confignode.thrift in this PR?

By the way, you should also implement the empty interfaces in ConfigNodeRPCServerProcessor. Otherwise, the code couldn't compile successfully. Finally, I will fill these interfaces later.

thrift-confignode/src/main/thrift/confignode.thrift
liuminghui233: fixed ~

liuminghui233:
@CRZbulabula @MarcosZyk @MrQuansy @JackieTien97 PTAL :)

thrift-confignode/src/main/thrift/confignode.thrift
```diff
@@ -76,6 +76,40 @@ struct DeviceGroupHashInfo {
...
+
+struct DataPartitionInfoResp {
+    // Map<StorageGroup, Map<DeviceGroupID, Map<TimePartitionId, List<DataRegionReplicaInfo>>>>
+    1: required map<string, map<i32, map<i64, list<RegionInfo>>>> dataPartitionMap
+}
+
+struct SchemaPartitionInfoResp {
+    // Map<StorageGroup, Map<DeviceGroupID, SchemaRegionPlaceInfo>>
+    1: required map<string, map<i32, RegionInfo>> schemaPartitionInfo
+}
```
MarcosZyk, line 104: The deviceId of i32 can't support wildcard match of ParthPattern. Maybe should replace it with the raw devicePath

MarcosZyk, line 104: The deviceId of i32 can't support wildcard match of ParthPattern. Maybe should replace it with the raw devicePath

JackieTien97: If you have `*` in devicePath, you will need to send your request to all the RegionInfo.

liuminghui233: same as above

MrQuansy:
I noticed that the new interface in ConfigIService, the parameter of fetch data partition is device paths instead of device group ids. At first, we hope data node can execute the hash calculation to decrease calculation workload on config nodes.
So I think we can replace the device path with device group ids, or offering another interface.

JackieTien97 (approved):

JackieTien97:
> I noticed the new interface in ConfigIService. The parameter of fetching data partition and fetching schema partition are the device paths instead of device group ids. At first, we hope data node can execute the hash calculation to decrease calculation workload on config nodes.
> 
> So I think we can replace the device path with device group ids, or offering another interface. We can discuss afterwards.

When we have `*` in our path, we can't calculate the DeviceGroupId.

MarcosZyk:
> > I noticed the new interface in ConfigIService. The parameter of fetching data partition and fetching schema partition are the device paths instead of device group ids. At first, we hope data node can execute the hash calculation to decrease calculation workload on config nodes.
> > So I think we can replace the device path with device group ids, or offering another interface. We can discuss afterwards.
> 
> When we have `*` in our path, we can't calculate the DeviceGroupId.

Got it. i32 will be ok.

confignode/src/main/java/org/apache/iotdb/confignode/service/thrift/server/ConfigNodeRPCServerProcessor.java
```diff
@@ -156,6 +162,22 @@ public DataPartitionInfo getDataPartition(GetDataPartitionReq req) throws TExcep
...
 
+  @Override
+  public DataPartitionInfoResp fetchDataPartitionInfo(FetchDataPartitionReq req) throws TException {
+    return null;
+  }
+
+  @Override
+  public SchemaPartitionInfoResp fetchSchemaPartitionInfo(FetchSchemaPartitionReq req)
+      throws TException {
+    return null;
```
CRZbulabula, line 173: Will return null cause trouble? Might it be better to return a resp with empty maps?

confignode/src/main/java/org/apache/iotdb/confignode/service/thrift/server/ConfigNodeRPCServerProcessor.java
```diff
@@ -156,6 +162,22 @@ public DataPartitionInfo getDataPartition(GetDataPartitionReq req) throws TExcep
...
+  }
+
+  @Override
+  public PartitionInfoResp fetchPartitionInfo(FetchPartitionReq req) throws TException {
+    return null;
```
CRZbulabula, line 178: Will return null cause trouble? Might it be better to return a resp with empty maps?

CRZbulabula (approved):
LGTM.

confignode/src/main/java/org/apache/iotdb/confignode/service/thrift/server/ConfigNodeRPCServerProcessor.java
liuminghui233: These methods have not been used yet. It's nothing serious ~

PR #5398, opened 2022-04-01
[IOTDB-2825] Fix count node by level

MarcosZyk:
### Cause
When counting nodes above sg, the node cannot be counted correctly because the existing method in MTreeBelowSG won't check the node above sg and also mistaken the level comparision.

### Solution
Enable MTreeBelowSG to check the node above sg, remove duplicated nodes in SchemaProcessor and finally return the num of matched nodes.


SilverNarcissus (approved):
Look good to me~

PR #4316, opened 2021-11-04
[IOTDB-1923] Separate the request unpacking and execution processing logic of TSServiceImpl

jun0315:
see https://issues.apache.org/jira/browse/IOTDB-1923

PR #5399, opened 2022-04-01
[IOTDB-2836] Fix ErrorHandlingUtils exception stacktrace logging

ericpai:
See JIRA: https://issues.apache.org/jira/browse/IOTDB-2836

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/utils/ErrorHandlingUtils.java
+++ b/server/src/main/java/org/apache/iotdb/db/utils/ErrorHandlingUtils.java
@@ -56,9 +56,9 @@ public class ErrorHandlingUtils {
       Exception e, String operation, TSStatusCode statusCode) {
     String message = String.format("[%s] Exception occurred: %s failed. ", statusCode, operation);
     if (e instanceof IOException || e instanceof NullPointerException) {
-      LOGGER.error("Status code: {}, operation: {} failed", statusCode, operation, e);
+      LOGGER.error("Status code: " + statusCode + ", operation: " + operation + " failed", e);
     } else {
-      LOGGER.warn("Status code: {}, operation: {} failed", statusCode, operation, e);
+      LOGGER.warn("Status code: " + statusCode + ", operation: " + operation + " failed", e);
     }
     return RpcUtils.getStatus(statusCode, message + e.getMessage());
   }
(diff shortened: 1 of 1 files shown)


ericpai:
@HTHou PTAL

server/src/main/java/org/apache/iotdb/db/utils/ErrorHandlingUtils.java
```diff
@@ -150,7 +151,7 @@ public static TSStatus tryCatchNonQueryException(Exception e) {
       // ignore logging sg not ready exception
       if (!(rootCause instanceof StorageGroupNotReadyException)) {
         if (((IoTDBException) e).isUserException()) {
-          LOGGER.warn(message + e.getMessage());
+          LOGGER.warn(message, e);
```
HTHou, line 154: Why do you change this? I think there is a reason for printing `e.getMessage()` only... @HeimingZ

ericpai: I'm not sure whether we don't need the stackstace indeed.

HeimingZ: @jun0315 PTAL

jun0315: In the original version:
https://github.com/apache/iotdb/pull/4316/files#diff-94a8b388744fb4d4422cb90e8fd8650803610de8ea6e734b3924f2d3a7cb86f4L2318 (In this page, we need click the `TSServiceImpl.java` `Load diff` button)
it was printed `e.getMessage()`, which I transplanted directly at that time. 
But now I think maybe need the stackstace is a better way?
LGTM

jun0315 (approved):

server/src/main/java/org/apache/iotdb/db/utils/ErrorHandlingUtils.java
ericpai: Thanks @jun0315

server/src/main/java/org/apache/iotdb/db/utils/ErrorHandlingUtils.java
```diff
@@ -150,7 +151,7 @@ public static TSStatus tryCatchNonQueryException(Exception e) {
...
         } else {
           LOGGER.warn(message, e);
         }
```
HTHou, line 157: Ok, and then we don't need the if else here?

ericpai: Good catch! I removed the `if-else` statements.

HTHou (approved):
LGTM

PR #5402, opened 2022-04-02
[IOTDB-2846] Add Driver which implements ExecFragmentInstance interface

JackieTien97:
This pr mainly does five things:

1. Add `SchemaDriver` and `DataDriver` which implements `ExecFragmentInstance` interface, `SchemaDriver` for schema related query and `DataDriver` for data related query. `DataDriver` has to do initializaton things like preparing related TsFileResouce List for data querying
2. Add UT for `DataDriver`
3. Rename previous `FragmentInstanceManager` to `FragmentInstanceScheduler`
4. Add a new `FragmentInstanceManager` to manage all the query related things, it's more like a query entrance
5. Implement contructing process for `TimeJoinOperator`, `SeriesScanOperator` and `LimitOperator` in `LocalExecutionPlanner`
6. add abort method for `IFragmentInstanceScheduler` to abort one fragment instance


server/src/main/java/org/apache/iotdb/db/mpp/execution/DataDriver.java
```diff
@@ -0,0 +1,324 @@
...
+    return driverContext.getId();
+  }
+
+  @Override
+  public void close() {
+    Operator root;
+    ISinkHandle sinkHandle;
+    synchronized (this) {
+      closed = true;
+      root = this.root;
```
cornmonster, line 156: root 和 sinkHandle 是 final 的，似乎没有必要再次赋值

server/src/main/java/org/apache/iotdb/db/mpp/execution/DataDriver.java
```diff
@@ -0,0 +1,324 @@
...
+    SettableFuture<Void> future = SettableFuture.create();
+    future.set(null);
+    driverBlockedFuture.set(future);
+  }
+
+  @Override
+  public boolean isFinished() {
+    try {
+      boolean isFinished =
+          closed || (driverBlockedFuture.get().isDone() && root != null && root.isFinished());
```
cornmonster, line 89: 1. 从实现上看，finished 语义表示的是是否需要继续被调度执行吧？ExecFragmentInstance#isFinished 的 javadoc可能有点不准确？
2. driverBlockedFuture 是否 complete 和 finished 似乎没有关系

server/src/main/java/org/apache/iotdb/db/mpp/execution/DataDriver.java
```diff
@@ -0,0 +1,324 @@
...
+      logger.error(
+          "Failed to query whether the data driver {} is finished", driverContext.getId(), t);
+      driverContext.failed(t);
+      close();
+      return true;
+    }
+  }
+
+  @Override
+  public ListenableFuture<Void> processFor(Duration duration) {
```
ericpai, line 104: In the `processFor` implementation, it should return a future with  `isCancelled() == true` if any error happens. So a single `NOT_BLOCKED` `ListenableFuture` may be insufficient.

server/src/main/java/org/apache/iotdb/db/mpp/execution/FragmentInstanceExecution.java
```diff
@@ -0,0 +1,87 @@
...
+import static java.util.Objects.requireNonNull;
+
+public class FragmentInstanceExecution {
+
+  private final IFragmentInstanceScheduler scheduler;
+
+  private final FragmentInstanceId instanceId;
+  private final FragmentInstanceContext context;
+
+  private final ExecFragmentInstance driver;
```
cornmonster, line 37: FragmentInstanceExecution 和 ExecFragmentInstance 命名有点冲突，把 ExecFragmentInstance 改成 Driver 如何？

server/src/main/java/org/apache/iotdb/db/mpp/execution/FragmentInstanceContext.java
```diff
@@ -69,4 +77,35 @@ public OperatorContext addOperatorContext(
...
+  public void cancel() {
+    state.set(FragmentInstanceState.CANCELED);
+  }
+
+  public void abort() {
+    state.set(FragmentInstanceState.ABORTED);
+  }
+
+  public void finish() {
+    if (state.get().isDone()) {
```
ericpai, line 106: Is there any other state execpt `FINISHED` when `state.get().isDone() == true`?

server/src/main/java/org/apache/iotdb/db/mpp/execution/FragmentInstanceManager.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicReference;
+
+import static java.util.Objects.requireNonNull;
+
+public class FragmentInstanceManager {
+
+  private final Map<FragmentInstanceId, FragmentInstanceContext> instanceContext;
+  private final Map<FragmentInstanceId, FragmentInstanceExecution> instanceExecution;
```
ericpai, line 38: When are the entries of the two maps removed?

server/src/main/java/org/apache/iotdb/db/mpp/execution/SchemaDriver.java
```diff
@@ -0,0 +1,146 @@
...
+        if (!future.isDone()) {
+          return updateDriverBlockedFuture(future);
+        }
+      } while (System.nanoTime() - start < maxRuntime && !root.isFinished());
+    } catch (Throwable t) {
+      logger.error("Failed to execute fragment instance {}", driverContext.getId(), t);
+      driverContext.failed(t);
+      close();
+    }
+    return NOT_BLOCKED;
```
ericpai, line 101: The same as `DataDriver`. It should return an `isCancelled() == true` `ListenableFuture` to tell the scheduler that it's stopped abnormally.

ericpai (changes requested):

server/src/main/java/org/apache/iotdb/db/mpp/execution/DataDriver.java
JackieTien97: Yes, here just want to make sure that if driverBlockedFuture is not done, isFinished method should never return true, otherwise it will have some problems.

server/src/main/java/org/apache/iotdb/db/mpp/execution/DataDriver.java
JackieTien97: Yes, that's right and `synchronized` here is needless because `DataDriver` is not thread safe class.

server/src/main/java/org/apache/iotdb/db/mpp/execution/DataDriver.java
JackieTien97: > 1. 从实现上看，finished 语义表示的是是否需要继续被调度执行吧？ExecFragmentInstance#isFinished 的 javadoc可能有点不准确？

Actually, I think there is a causal relationship between these two. This instance don't need to be scheduled because of this instance has no more data to process.

cornmonster: > Actually, I think there is a causal relationship between these two. This instance don't need to be scheduled because of this instance has no more data to process.

An instance terminated due to failure also shouldn't be scheduled for execution anymore.

server/src/main/java/org/apache/iotdb/db/mpp/execution/FragmentInstanceContext.java
JackieTien97: image
These four states will all return true.

server/src/main/java/org/apache/iotdb/db/mpp/execution/SchemaDriver.java
JackieTien97: fixed

server/src/main/java/org/apache/iotdb/db/mpp/execution/DataDriver.java
JackieTien97: fixed

server/src/main/java/org/apache/iotdb/db/mpp/execution/FragmentInstanceManager.java
JackieTien97: good catch, I will add a background timer thread to clear

ericpai (approved):

server/src/main/java/org/apache/iotdb/db/mpp/execution/DataDriver.java
JackieTien97: That's true, i will fix that

cornmonster:
LGTM

PR #5404, opened 2022-04-02
fix bug in ratis consensus test

SzyWilliam:
## Description
Fix bugs in current RatisConsensusTest:
1. Use CountDownLatch to wait for all writes complete instead of shutting down executor-service after timeout. 
2. Read from leader instead of a random picked peer.


OneSizeFitsQuorum (approved):
LGTM

wangchao316 (approved):
LGTM

PR #5410, opened 2022-04-03
[IOTDB-2843] remove unused jna dependency in cli

qiaojialin:
I found that this dependency is not used in code.


qiaojialin:
Do Not Merge, This dependency is used in windows, we need to add javadoc later

HTHou (approved):

PR #5416, opened 2022-04-05
[To new_sync][IOTDB-2772] Add testcontainer e2e test for newsync

Cpaulyz:
* Add testcontainer test for newsync
  * normal case
  * weak network case
* Add sync test to git workflow as E2E Test


Cpaulyz:
@irvine0109 have a look plz :D

.github/workflows/e2e.yml
```diff
@@ -5,10 +5,12 @@ name: E2E Tests
...
   push:
-    branches: 
-      - test/e2e
+    branches:
+      - master
+      - new_sync
     paths-ignore:
-      - 'docs/**'
+      - master
+      - new_sync
```
HTHou, line 13: No necessary to change this?

Cpaulyz: has reverted this change

testcontainer/src/test/java/org/apache/iotdb/db/newsync/SyncWeakNetworkIT.java
```diff
@@ -0,0 +1,39 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iotdb.db.newsync;
+
+import org.junit.Assert;
+import org.testcontainers.containers.Container;
+import org.testcontainers.containers.ContainerState;
+
+/** Simulate network delay and loos. */
```
irvine0109, line 25: loos -> loss

testcontainer/src/test/java/org/apache/iotdb/db/newsync/SyncWeakNetworkIT.java
```diff
@@ -0,0 +1,39 @@
...
+public class SyncWeakNetworkIT extends SyncIT {
+  @Override
+  public void init() throws Exception {
+    super.init();
+    // set delay is 200±50ms that conform to a normal distribution;  loss rate is 5%
```
irvine0109, line 30: Maybe net packet corrupt, reorder and duplicate is also needed to test.

testcontainer/src/test/java/org/apache/iotdb/db/newsync/SyncIT.java
```diff
@@ -0,0 +1,322 @@
...
+      stopPipe();
+      prepareIns3();
+      startPipe();
+      checkResult();
+    } catch (Exception e) {
+      e.printStackTrace();
+      Assert.fail();
+    }
+  }
+
```
irvine0109, line 255: Should we test the file transfer?

testcontainer/src/test/java/org/apache/iotdb/db/newsync/SyncWeakNetworkIT.java
Cpaulyz: fixed

irvine0109 (approved):

irvine0109 (commented):
Does testcontainer module run?

.github/workflows/e2e.yml
```diff
@@ -58,4 +59,5 @@ jobs:
 
       - name: TestContainer
         run: |
-          mvn -B -T 4 integration-test -pl testcontainer
+#          TODO: add cluster test
+          mvn -B -T 4 integration-test -pl testcontainer -P sync
```
HTHou, line 63: How about using test-sync? And consider changing the first line `name: E2E Tests` to `name: sync Tests`?

```suggestion
          mvn -B -T 4 integration-test -pl testcontainer -P test-sync
```

Cpaulyz: 1. rename sync -> test-sync
2. I don't think it's necessary to create a separate test category for sync... What's more, there are still some other tests here(although they are simple).  BTW, are we going to add cluster test in future？

HTHou:
name: Sync Tests

on:
  push:
    branches:
      - master
    paths-ignore:
      - 'docs/**'
  pull_request:
    branches:
      - master
      - cluster_new
    paths-ignore:
      - 'docs/**'
  # allow manually run the action:
  workflow_dispatch:

concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true

env:
  MAVEN_OPTS: -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.http.retryHandler.class=standard -Dmaven.wagon.http.retryHandler.count=3

jobs:
  E2E:
    runs-on: ubuntu-latest
    strategy:
      fail-fast: true
      matrix:
        case:
          - cli
    steps:
      - uses: actions/checkout@v2

      - uses: actions/cache@v2
        with:
          path: ~/.m2
          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
          restore-keys: ${{ runner.os }}-m2

      - name: Cache Maven packages
        uses: actions/cache@v2
        with:
          path: ~/.m2
          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
          restore-keys: ${{ runner.os }}-m2-

      - name: Build IoTDB server distribution zip
        run:  mvn -B clean install -pl distribution -am -DskipTests

      - name: Build Docker Image
        run: |
          docker build . -f docker/src/main/Dockerfile-single -t "iotdb:$GITHUB_SHA"
          docker images

      - name: TestContainer
        # TODO: add cluster test
        run: |
          mvn -B -T 4 integration-test -pl testcontainer -P test-sync

.github/workflows/e2e.yml
HTHou: IMO, one workflow is for one kind of test. We don’t need to test other modules in sync test at all…

We actually going to abandon the e2e test for cluster since we are refactoring the IT framework for new cluster.

HTHou (approved):

PR #5419, opened 2022-04-06
IOTDB Operator

SzyWilliam:
Kubernetes Operator


pom.xml
```diff
@@ -114,6 +114,7 @@
         <module>metrics</module>
         <module>integration</module>
         <module>consensus</module>
+        <module>operator</module>
```
wangchao316, line 117: what does this module do?

wangchao316 (commented):
bind jira to this pr.

jufickel-b:
What is the general state of a Kubernetes Operator for IotDB? Not many cloud providers offer a managed instance of IoTDB (do any?). But to deploy it by one self, an operator is crucial. Or differently stated: not having a mature IoTDB operator rules this database out, despite it would technically be the best option for an IoT solution.

PR #5451, opened 2022-04-08
[IOTDB-2815] Add a method to insert records into an existed timeseries without TSDataType

cchen1115:
## Description
add a method to insert records into an existed timeseries without TSDataType.

#5352
https://github.com/apache/iotdb/discussions/5352

### DIFF
--- a/session/src/main/java/org/apache/iotdb/session/Session.java
+++ b/session/src/main/java/org/apache/iotdb/session/Session.java
@@ -805,14 +805,10 @@ public class Session {
    * @see Session#insertRecords(List, List, List, List, List)
    * @see Session#insertTablet(Tablet)
    */
-  public void insertRecord2(
-          String deviceId,
-          long time,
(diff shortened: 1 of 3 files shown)


session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -1188,6 +1307,42 @@ public void insertRecords(
...
+  /**
+   * Insert multiple rows, which can reduce the overhead of network. This method is just like jdbc
+   * executeBatch, we pack some insert request in batch and send them to server. If you want improve
+   * your performance, please see insertTablet method
+   *
+   * <p>Each row is independent, which could have different deviceId, time, number of measurements
+   *
+   * @see Session#insertTablet(Tablet)
+   */
+  public void insertRecords2(
```
qiaojialin, line 1319: do not use 2

PR #5460, opened 2022-04-08
[IOTDB-2864] Fix Read-only occurred when insert Text values to aligned timeseries

HTHou:
## Description

image

When extending column in AlignedTVList, we need to supply bitmaps for the extending column. There is a bug causes the last supplying bitmap doesn't be marked, which will cause NPE when the extending column is TEXT dataType and execute flushing. Finally the server will turn to read-only because of it.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/utils/datastructure/AlignedTVList.java
+++ b/server/src/main/java/org/apache/iotdb/db/utils/datastructure/AlignedTVList.java
@@ -298,8 +298,9 @@ public class AlignedTVList extends TVList {
           break;
       }
       BitMap bitMap = new BitMap(ARRAY_SIZE);
-      // last bitmap should be marked to the tslist size's position
-      if (i == timestamps.size() - 1) {
+      // last bitmap should be marked to the tvlist size's position
+      // if rowCount % ARRAY_SIZE != 0, last bitmap should mark all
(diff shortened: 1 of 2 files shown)


qiaojialin (approved):

server/src/main/java/org/apache/iotdb/db/utils/datastructure/AlignedTVList.java
```diff
@@ -298,8 +298,9 @@ public void extendColumn(TSDataType dataType) {
           break;
       }
       BitMap bitMap = new BitMap(ARRAY_SIZE);
-      // last bitmap should be marked to the tslist size's position
-      if (i == timestamps.size() - 1) {
+      // last bitmap should be marked to the tvlist size's position
+      // if rowCount % ARRAY_SIZE != 0, last bitmap should mark all
```
HTHou, line 302:
```suggestion
      // The following code is for these 2 kinds of scenarios.

      // Eg1: If rowCount=5 and ARRAY_SIZE=2, we need to supply 3 bitmaps for the extending column.
      // The first 2 bitmaps should mark all bits to represent 4 nulls and the 3rd bitmap should
      // mark
      // the 1st bit to represent 1 null value.

      // Eg2: If rowCount=4 and ARRAY_SIZE=2, we need to supply 2 bitmaps for the extending column.
      // These 2 bitmaps should mark all bits to represent 4 nulls.
```

PR #5465, opened 2022-04-09
Refactor schema region management and Improve DeleteStorageGroup procedure

MarcosZyk:
### Refactor schema region management
SchemaEngine  takes on all the responsibility of schema region management. Now the localConfigManager only manages the schemaRegionId and doesn't provide interfaces for schema region management.

### Improve DeleteStorageGroup procedure
Now only when all related schemaRegions are deleted, will the storageGroup be deleted.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/LocalConfigManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/LocalConfigManager.java
@@ -39,6 +39,7 @@ import org.apache.iotdb.db.metadata.storagegroup.IStorageGroupSchemaManager;
 import org.apache.iotdb.db.metadata.storagegroup.StorageGroupSchemaManager;
 import org.apache.iotdb.db.metadata.template.Template;
 import org.apache.iotdb.db.metadata.template.TemplateManager;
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/apache/iotdb/db/metadata/LocalConfigManager.java
```diff
@@ -289,6 +261,49 @@ private void ensureStorageGroup(PartialPath path, boolean shouldAllocateSchemaRe
...
+   * @param storageGroups list of paths to be deleted.
+   */
+  public void deleteStorageGroups(List<PartialPath> storageGroups) throws MetadataException {
+    for (PartialPath storageGroup : storageGroups) {
+      deleteStorageGroup(storageGroup);
+    }
+  }
+
+  private void deleteStorageGroup(PartialPath storageGroup) throws MetadataException {
+    deleteSchemaRegionsInStorageGroup(
```
cigarl, line 276: If a `storageGroup` has five `schemaRegions` and failed to delete the third one（In distributed scenarios, the health status of different nodes may be inconsistent, such as disks）, does the user still think the `storage group` is available（because the deletion failed）?  

Maybe we need to fix consistency instead of just fixing a corner case.

PR #5466, opened 2022-04-09
[IOTDB-2868] Construct ExchangeOperator in LocalExecutionPlanner

JackieTien97:
1. Construct `ExchangeOperator` in `LocalExecutionPlanner` 
2. Add a `DataSourceOperator` extends `SourceOperator` and move `initQueryDataSource` method from `SourceOperator` to `DataSourceOperator` because schema query related source operators don't need to `initQueryDataSource`
3. add `sourceId` field in `SeriesScanOperator` to make its `getSourceId()` not returning null.
4. call `execDataQueryFragmentInstance` of `FragmentInstanceManager` in `DataRegionStateMachine`


server/src/main/java/org/apache/iotdb/db/mpp/operator/source/ExchangeOperator.java
```diff
@@ -0,0 +1,93 @@
...
+      return sourceHandle.receive();
+    } catch (IOException e) {
+      throw new RuntimeException(
+          "Error happened while reading from source handle " + sourceHandle, e);
+    }
+  }
+
+  @Override
+  public boolean hasNext() {
+    return sourceHandle.isFinished();
```
xingtanzjr, line 64: should be `!sourceHandle.isFinished();`

JackieTien97: fix

xingtanzjr (approved):

PR #5485, opened 2022-04-12
[IOTDB-2882] Fixed display of unary expression 

flashzxi:
## Description
Display of the unary expression doesn't show brackets.

Fixed this bug


SteveYurongSu (approved):
LGTM

PR #5511, opened 2022-04-13
[IOTDB-2910] Fix Count aggregate is not right after delete sg bug

JackieTien97:
In 0.13 and master, we add BloomFilterCache, but we forgot to add clear cache for it in `operateClearCache`.


HTHou (approved):
LGTM

PR #5521, opened 2022-04-14
[IOTDB-2888] Unable to select an unary expression followed by a constant expression

flashzxi:
When you select unary expression with a constant expression, it will throw 
```java.lang.UnsupportedOperationException```

for example
```sql
select - 1 + s1 from root.test;
```

now fixed this bug


SteveYurongSu (changes requested):

SteveYurongSu:
Please notice the spotless issue :D

PR #5536, opened 2022-04-14
[IOTDB-2906] Commit pipe data serial number error, Can not find tsfile

Cpaulyz:
The reason for this error message is that the system deletes the file when it is loaded successfully. So `commit()` function in PipeDataQueue can not find tsfile that has been loaded to system.

I add a parameters to config whether to delete file after loading.
* If true, use move
* If false, create hardlink


yschengzi (approved):

yschengzi (approved):
LGTM

HTHou (approved):

PR #5553, opened 2022-04-15
[IOTDB-2922] Fix NPE when compacting with files that contains zero device

THUMarkLau:
See [IOTDB-2922](https://issues.apache.org/jira/browse/IOTDB-2922).
The reason for this bug is that, when compacting with files that contains zero device, the variable `currentDevice` in `TsFileDeviceIterator` is initialized as null, and if we visit it, NPE will occurs.


MarcosZyk (approved):
The modification in ```IoTDBShutdownHook``` help prevent compaction from visiting cleared schema module. 

LGTM~

choubenson (approved):

PR #5557, opened 2022-04-15
[IOTDB-2933] detect sender exit and set pipestatus=STOP

Cpaulyz:
* Use thrift to detect sender exit
* Add concurrent control in receiverService


yschengzi (commented):
some advices, please check

server/src/main/java/org/apache/iotdb/db/sync/transport/client/TransportClient.java
```diff
@@ -451,6 +442,14 @@ public void run() {
         throw new SyncConnectionException(
             String.format("Handshake with receiver %s:%d error.", ipAddress, port));
       }
+      SenderService.getInstance()
+          .receiveMsg(
+              heartbeat(
+                  new SyncRequest(
+                      RequestType.CREATE,
```
Cpaulyz, line 449: start
```suggestion
                      RequestType.START,
```

yschengzi (approved):
LGTM

PR #5564, opened 2022-04-15
Change assertion to prevent SchemaBuilderTest from failing due to indeterminate ordering

faaple:
## Description
All tests in the class `org.apache.iotdb.tsfile.write.schema.converter.SchemaBuilderTest` can fail based on the order of iteration in `HashMap`.

I found the issue using [NonDex](https://github.com/TestingResearchIllinois/NonDex):
```
mvn test-compile -pl tsfile -am
mvn -pl tsfile edu.illinois:nondex-maven-plugin:1.1.2:nondex -Dtest=org.apache.iotdb.tsfile.write.schema.converter.SchemaBuilderTest
```
The test can fail because it assumes that the order in `timeseries` is determinate, and can be fixed it by ignoring the order.

### DIFF
--- a/tsfile/src/test/java/org/apache/iotdb/tsfile/write/schema/converter/SchemaBuilderTest.java
+++ b/tsfile/src/test/java/org/apache/iotdb/tsfile/write/schema/converter/SchemaBuilderTest.java
@@ -30,8 +30,8 @@ import org.apache.iotdb.tsfile.write.schema.Schema;
 
 import org.junit.Test;
 
-import java.util.Arrays;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


tsfile/src/test/java/org/apache/iotdb/tsfile/write/schema/converter/SchemaBuilderTest.java
```diff
@@ -65,10 +68,13 @@ public void testJsonConverter1() {
...
     };
-    int i = 0;
+    Set<String> expected = new HashSet<String>(Arrays.asList(tsDesStrings));
+    List<String> actual = new ArrayList<String>();
     for (IMeasurementSchema desc : timeseries) {
-      assertEquals(tsDesStrings[i++], desc.toString());
+      actual.add(desc.toString());
     }
+    assertEquals(expected.size(), actual.size());
+    assertEquals(expected, new HashSet<String>(actual));
```
jixuan1989, line 77: `nondex` seems very interesting. 
However, I have a question: 
nondex is for checking `iterator of HashMap`, should we also consider the useage of `List`?
For example, in this case, `timeseres` is an `ArrayList`.
 
The second, `assertEquals(expected, new HashSet<String>(actual));` will compare two `HashSet` ( and use while-loop for checking all elements), is it ok (and is it more stable than ArrayList?)

MarcyGO: Thanks for your question. 
First, `nondex` doesn't shuffle lists because they have a specific iteration order. Second, the assertion compares two sets rather than two lists precisely to avoid the dependency on the order.

PR #5582, opened 2022-04-18
Optimize dependencies management to reduce the lib size

HTHou:
## Description

In this PR, I change the Junit maven scope to `test` to reduce the distribution package size.

After these change, the distribution package size can be 1MB smaller than before.


chengjianyun (approved):
LGTM

PR #5590, opened 2022-04-19
Rename thrift endpoint

cornmonster:
It is quite confusing when there are 2 classes named "EndPoint". They are org.apache.iotdb.commons.cluster.EndPoint and org.apache.iotdb.common.rpc.thrift.EndPoint.

By convention, the name of thrift generated classes should start with "T", so I renamed the org.apache.iotdb.common.rpc.thrift.EndPoint to org.apache.iotdb.common.rpc.thrift.TEndPoint.


JackieTien97 (approved):

OneSizeFitsQuorum (approved):
We didn't do this in last refactoring because of compatibility issues. Recently after our discussion, we found that it is really hard for the old version of client to be compatible with the new version of cluster because we will change a lot of APIs, so now I have no objection to this rename work.

qiaojialin (approved):

PR #5602, opened 2022-04-19
[IOTDB-2958]create a simple schemaCache for datanode metadata

yifuzhou:
## Description
1. use caffeine to implement a schemaEntityCache (partialPath -> SchemaEntity) for the management of datanode metadata.
2. SchemaEntity currently just includes timeseries metadata schema (TSDataType, TSEncoding, CompressionType, alias, isAligned)
3. provide two function interface to get SchemaEntity from measurementPath and AlignedPath, also can identfy whether need auto create schema.

Furture work:
1.TODO work in the code

### Content3 ...

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


server/src/main/java/org/apache/iotdb/db/mpp/sql/analyze/DataNodeSchemaCache.java
```diff
@@ -0,0 +1,136 @@
...
+ * StorageEngine
+ */
+public class DataNodeSchemaCache {
+  private static final Logger logger = LoggerFactory.getLogger(DataNodeSchemaCache.class);
+
+  private static IoTDBConfig config = IoTDBDescriptor.getInstance().getConfig();
+  private SchemaTree root;
+
+  private LoadingCache<MeasurementPath, SchemaNode> schemaNodeCache;
+
```
MarcosZyk, line 49: It seems all the key of this cache already maintains all the schema info we need. We implement this schemaTree to compress the prefix of series and this feature is broken by the schemaNodeCache.

yifuzhou: use partialPath ->SchemaEntity after out talk.

SpriCoder:
Good job! Please check the format of code.!

yifuzhou:
> Good job! Please check the format of code.



> Good job! Please check the format of code.

fix the checkstyle :)

SpriCoder (approved):
LGTM

MarcosZyk (approved):

server/src/main/java/org/apache/iotdb/db/metadata/cache/DataNodeSchemaCache.java
```diff
@@ -0,0 +1,223 @@
...
+/**
+ * This class takes the responsibility of metadata cache management of all DataRegions under
+ * StorageEngine
+ */
+public class DataNodeSchemaCache {
+  private static final Logger logger = LoggerFactory.getLogger(DataNodeSchemaCache.class);
+
+  private static IoTDBConfig config = IoTDBDescriptor.getInstance().getConfig();
+
+  private Cache<PartialPath, SchemaCacheEntity> schemaEntityCache;
```
qiaojialin, line 55: Could we use MTree directly?

server/src/main/java/org/apache/iotdb/db/metadata/cache/DataNodeSchemaCache.java
```diff
@@ -0,0 +1,223 @@
...
+  public SchemaCacheEntity getSingleSchemaEntity(
+      PartialPath partialPath, boolean isAutoCreateSchema) {
+    SchemaCacheEntity schemaCacheEntity;
+    if (partialPath instanceof AlignedPath) {
+      return null;
+    }
+    schemaCacheEntity = schemaEntityCache.getIfPresent(partialPath);
+    if (schemaCacheEntity == null) {
+      PartialPath cacheKey;
+      if (partialPath instanceof MeasurementPath) {
```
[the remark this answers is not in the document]
yifuzhou: Yes, if we need this partialPath to auto create schema, so the partialPath should contains MeasurmentSchema, and that will be a MeasurementPath.

server/src/main/java/org/apache/iotdb/db/metadata/cache/DataNodeSchemaCache.java
MarcosZyk: There's no cache eviction adapt to MTree, that is why we use existing map cache implement. We can implement a TreeCache in future work, which could be used in many scenarios related to tree structure and prefix compress.

yifuzhou: The design intention for this cache is, for insert operation, they just need to provide me a partialPath key and then I will tell then a related schemaCacheEntity. If they want to auto create schema, the partialPath should be MeasurementPath or AlignedPath which should contain some schema info. So it seems that current MTree structure is kind of redundancy because here we do not need such a tree structure.

server/src/main/java/org/apache/iotdb/db/metadata/cache/DataNodeSchemaCache.java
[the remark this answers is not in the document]
MarcosZyk: Currently, the ```MeasurementPath``` and ```AlignedPath``` only act as a schema query response for data query, rather than a request. They will be renamed to ```QueryUnit``` or someting like this. There's no need to wrap the params to a measurementPath and then unwrap them.

qiaojialin (approved):

PR #5608, opened 2022-04-20
[IOTDB-2965] Skip failed mlog and snapshot items when upgrading happens

ericpai:
After this fix, the `MetadataUpgrader` will not throw any exceptions if recovering one plan from snapshot or mlog.bin failed.

I think it's better to tell DBAs to backup the `system` directory before upgrading to new version.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/upgrade/MetadataUpgrader.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/upgrade/MetadataUpgrader.java
@@ -109,12 +109,12 @@ public class MetadataUpgrader {
    *   <li>Try set storage group based on the recovered StorageGroupMNodes and create timeseries
    *       based on the recovered MeasurementMNode
    *   <li>Redo the mlog
-   *   <li>delete the files of version, the order is:
+   *   <li>rename and backup the files in the same directory, the order is:
    *       <ol>
-   *         <li>mlog
-   *         <li>snapshot
-   *         <li>tag file
-   *         <li>tmp snapshot
+   *         <li>mlog.bin to mlog.bin.bak
+   *         <li>mtree-1.snapshot.bin to mtree-1.snapshot.bin.bak
+   *         <li>tlog.txt to tlog.txt.bak
+   *         <li>mtree-1.snapshot.bin.tmp to mtree-1.snapshot.bin.tmp.bak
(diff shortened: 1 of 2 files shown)


ericpai:
@qiaojialin @MarcosZyk PTAL

xingtanzjr (approved):

MarcosZyk:
If we don't interrupt the upgrade process, the mlog and snapshot of last version will be removed after upgrading. Maybe it's not safe.

ericpai:
> If we don't interrupt the upgrade process, the mlog and snapshot of last version will be removed after upgrading. Maybe it's not safe.

Yes, this is also my concern. But if it stops the whole process, how can we do the upgrading successfully finally? We haven't had a metadata fix tool yet.
Maybe we can backup these files previously before recovering from snapshot?

qiaojialin:
> > If we don't interrupt the upgrade process, the mlog and snapshot of last version will be removed after upgrading. Maybe it's not safe.
> 
> Yes, this is also my concern. But if it stops the whole process, how can we do the upgrading successfully finally? We haven't had a metadata fix tool yet. Maybe we can backup these files previously before recovering from snapshot?

backup +1, we could let users to handle this manually

ericpai:
> 

Maybe there's another discussion: does the MetadataUpgrader have the responsibility to backup old files in `system`?
For any IoTDB instance with millions of timeseries, the `system` dictionary may be very large, and it will cost more time when upgrading.

MarcosZyk:
Since we will produce new file to store all metadata, maybe we can simply rename the old files rather than delete them directly? And we can remind users that the can handle the backup file, renamed old file, based on the recover situation.

ericpai:
> Since we will produce new file to store all metadata, maybe we can simply rename the old files rather than delete them directly? And we can remind users that the can handle the backup file, renamed old file, based on the recover situation.

OK, move/rename file is a low cost operation. If any move/rename operation fails, the upgrading will be aborted. As it's only caused by OS storage.

ericpai:
I will add another commit in this PR to rename old files.

ericpai:
@MarcosZyk I have updated the rename logic and enhance the test case. Please take a look~

MarcosZyk (approved):
LGTM~

qiaojialin (approved):

PR #5613, opened 2022-04-20
[IOTDB-2845] Implementation of DeviceViewOperator Part1

Alima777:
What DeviceViewOperator need to do is traversing the device child operators, get all tsBlocks of one device and transform it to the form we need, then get the next device operator until no next device.

If some columns are not existing in one device, those columns will be null. e.g.  [s1,s2,s3] is query, but only [s1, s3] exists in device1, then the column of s2 will be filled with NullColumn.


tsfile/src/main/java/org/apache/iotdb/tsfile/read/common/block/column/ColumnBuilderStatus.java
```diff
@@ -41,6 +41,10 @@ public ColumnBuilderStatus(TsBlockBuilderStatus tsBlockBuilderStatus) {
         requireNonNull(tsBlockBuilderStatus, "tsBlockBuilderStatus must not be null");
   }
 
+  public ColumnBuilderStatus() {
+    this.tsBlockBuilderStatus = new TsBlockBuilderStatus();
+  }
+
```
JackieTien97, line 47:
```suggestion
```

tsfile/src/main/java/org/apache/iotdb/tsfile/read/common/block/TsBlock.java
```diff
@@ -154,6 +154,18 @@ public TsBlock appendValueColumn(Column column) {
...
+    if (positionCount != column.getPositionCount()) {
+      throw new IllegalArgumentException("Block does not have same position count");
+    }
+
+    Column[] newBlocks = Arrays.copyOf(valueColumns, valueColumns.length + 1);
+    System.arraycopy(newBlocks, index, newBlocks, index + 1, valueColumns.length - index);
+    newBlocks[index] = column;
+    return wrapBlocksWithoutCopy(positionCount, timeColumn, newBlocks);
+  }
+
```
JackieTien97, line 168:
```suggestion
```

JackieTien97 (changes requested):

JackieTien97 (approved):

PR #5630, opened 2022-04-21
Add ColumnMerger to merge multipul input columns of same sensor into one column

JackieTien97:
1. Add `ColumnMerger` to merge multipul input columns of same sensor into one column. Currently, I only implement `SingleColumnMerger`.
2. Fix `TimeJoinOperator` bug that if we query one non-exist sensor, we will get NPE. Add UT for this corner case.


cornmonster:
LGTM

PR #5633, opened 2022-04-22
[IOTDB-2964] Fix exception occurs when selecting cross space compaction task

THUMarkLau:
See [IOTDB-2964](https://issues.apache.org/jira/browse/IOTDB-2964).
The reason for this exception is that, the validation check for selected sequence files does not check the existence of the files. So in the end of IT, though the file is clear, the deleted files will be chosen.


tsunghanjacktsai (approved):

qiaojialin:
add test

server/src/main/java/org/apache/iotdb/db/engine/compaction/cross/rewrite/selector/RewriteCompactionFileSelector.java
```diff
@@ -167,6 +167,12 @@ void select(boolean useTightBound) throws IOException {
     while (unseqIndex < resource.getUnseqFiles().size() && timeConsumption < timeLimit) {
       // select next unseq files
       TsFileResource unseqFile = resource.getUnseqFiles().get(unseqIndex);
+      if (!unseqFile.getTsFile().exists()
+          || unseqFile.isCompacting()
+          || !unseqFile.isClosed()
+          || unseqFile.isCompactionCandidate()) {
+        break;
+      }
```
choubenson, line 175: Only need to judge whether the file exists, because the others are filtered out when the `CompactResource` is constructed earlier.

choubenson (approved):

server/src/main/java/org/apache/iotdb/db/engine/compaction/cross/rewrite/selector/RewriteCompactionFileSelector.java
THUMarkLau: reoslve

PR #5636, opened 2022-04-22
Fix the issue that EndTime in FragmentInstanceContext is not set

xingtanzjr:
## Description
1. When the FragmentInstance is not complete, its `endTime` is not set. It will lead to NPE when parse it to `long` when fetching FragmentInstance status.
2. Fix an issue that the coordinator cannot cleanup QueryExecution if the QueryExecution throws exception when constructing, which will lead to client stuck.


JackieTien97 (changes requested):
There is `executionEndTime.compareAndSet(null, now);` in `updateStatsIfDone` of `FragmentInstanceContext`, you can change it to `executionEndTime.compareAndSet(-1, now);`

xingtanzjr:
> There is `executionEndTime.compareAndSet(null, now);` in `updateStatsIfDone` of `FragmentInstanceContext`, you can change it to `executionEndTime.compareAndSet(-1, now);`

fixed

PR #5640, opened 2022-04-22
Filter non schemaRegionDir

MarcosZyk:
### Cause
In Mac os, there are hidden dirs or files, named with ```.DS_Store``` suffix. When recover schemaEngine and such dirs or files is scanned, NumberFormatException will be thrown.

### Solution
When encountering NumberFormatException, ignore this and skip this dir, since a schemaRegion dir is named after a numeric number.


HTHou (approved):
lgtm

PR #5643, opened 2022-04-22
[IOTDB-2982] Recover tsfile after datanode restart

HeimingZ:
Support recovering TsFile from InsertNode.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/wal/recover/file/TsFilePlanRedoer.java
+++ b/server/src/main/java/org/apache/iotdb/db/wal/recover/file/TsFilePlanRedoer.java
@@ -154,18 +154,13 @@ public class TsFilePlanRedoer {
         return;
       }
     }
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/apache/iotdb/db/wal/recover/file/TsFilePlanRedoer.java
```diff
@@ -182,6 +177,7 @@ void redoInsert(InsertNode node) throws WriteProcessException, QueryProcessExcep
             (InsertTabletNode) node, 0, ((InsertTabletNode) node).getRowCount());
       }
     }
+    logger.info("");
```
HTHou, line 180: empty log?

HeimingZ: fixed

qiaojialin (approved):

PR #5646, opened 2022-04-23
[IOTDB-2960]Add partition cache

SpriCoder:
Add partition cache in clusterParitionFetcher.
SchemaPartitionCache: device -> storagegroup, TRegionReplicaSet.
DataPartitionCache: storageGroup, TSeriesPartitionSlot, TTimePartitionSlot -> List<TRegionReplicaSet>
Notice: if there are no  TTimePartitionSlot in DataPartitionQueryParam, it's means fetch all data so cache will miss all the time.

Besides, I have do test in 3 confifnodes and 3 datanodes cluster and run some scripts as follows:
image


SpriCoder:
@MarcosZyk PTAL！

PR #5647, opened 2022-04-23
[IOTDB-2930]Fix concurrent UnPin bug & Improve template implementation

MarcosZyk:
### Fix concurrent UnPin bug
After a thread unPin a node, it should check the cache status of the node. If the node is now not Pinned, then its parent should be unPin once, which means one of the children is not pinned any more. However, the operation and check should be synchronized, otherwise the change may result in multi threads get wrong cache status and do extra unpin, which may result in negative semaphore value.

### Improve template implementation
The old implementation of template check during insert takes low performance and is hard to understand. Thus the implementation is rewritten.


MarcosZyk:
@bigreybear @SilverNarcissus PZTL

bigreybear (approved):
LGTM~

qiaojialin (approved):

PR #5651, opened 2022-04-24
Fix schema file with cache of size 1

bigreybear:
## Description
Before this pull request, schema file stores all pages in a map, and only flushes dirty pages if the cache had no more space for new pages. 
With the cache map of size 1, a dirty page may be evicted from the cache during a method which modifies multi pages (like creating a new multi-page segment) and this makes a exception unexpected.

### What's new in this PR
Now, schema file uses an extra map called dirtyPages to collect all pages modified during write or delete method within the file instance, and flushes will occur before the method ends.


bigreybear:
@MarcosZyk PTAL

SilverNarcissus (approved):
Look good to me~

MarcosZyk (approved):

PR #5622, opened 2022-04-21
[IOTDB-2970] Code style: Avoid wildcard imports

ericpai:
This PR contains:
1. Enable wildcard import check in checkstyle.xml.
2. Fix all wildcard imports codes.
3. Add document about code style things.

PR #5652, opened 2022-04-24
[IOTDB-2996] Fix wildcard import in test folders

ericpai:
Folloup #5622 , now the codes in test folders are also enabled wildcard import check.


ericpai:
@HTHou PTAL~

HTHou (approved):

PR #5655, opened 2022-04-25
Support set storage group in standalone mpp mode

HTHou:
## Description

1. move LocalConfigNode to localconfignode package
2. implement some interfaces of Standalone schema fetcher and Standalone schema fetcher 
3. support set storage group in standalone mpp mode.


server/src/main/java/org/apache/iotdb/db/service/DataNode.java
```diff
@@ -170,6 +170,7 @@ public void joinCluster() throws StartupException {
   public void active() throws StartupException {
     // set the mpp mode to true
     IoTDBDescriptor.getInstance().getConfig().setMppMode(true);
+    IoTDB.getInstance().setClusterMode();
```
HTHou, line 173:
```suggestion
```

server/src/main/java/org/apache/iotdb/db/mpp/execution/config/SetStorageGroupTask.java
```diff
@@ -45,31 +48,39 @@ public SetStorageGroupTask(SetStorageGroupStatement setStorageGroupStatement) {
...
-      if (TSStatusCode.SUCCESS_STATUS.getStatusCode() != tsStatus.getCode()) {
-        LOGGER.error(
-            "Failed to execute set storage group {} in config node, status is {}.",
-            setStorageGroupStatement.getStorageGroupPath(),
-            tsStatus);
-        future.setException(new StatementExecutionException(tsStatus));
-      } else {
-        future.set(new ConfigTaskResult(TSStatusCode.SUCCESS_STATUS));
+    if (IoTDB.getInstance().isMppClusterMode()) {
+      // Construct request using statement
```
MarcosZyk, line 52: Is this ```if-else``` a temporary implementation? I think this judgement should be integrated in a high level framework?

MarcosZyk (commented):
Excellent work！Only one question. Please take a look.

server/src/main/java/org/apache/iotdb/db/mpp/execution/config/SetStorageGroupTask.java
HTHou: Yes, it is a temporary implementation. We need to fix it in the future. Let me add a TODO mark.

qiaojialin (approved):

PR #5659, opened 2022-04-25
Optimize source handle error handling

cornmonster:
Add `SourceHandleListener#onFailure`. The method will be called when fatal exceptions happen during data transmission.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/mpp/buffer/DataBlockManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/mpp/buffer/DataBlockManager.java
@@ -50,19 +50,21 @@ public class DataBlockManager implements IDataBlockManager {
   private static final Logger logger = LoggerFactory.getLogger(DataBlockManager.class);
 
   public interface SourceHandleListener {
-    void onFinished(SourceHandle sourceHandle);
+    void onFinished(ISourceHandle sourceHandle);
 
-    void onClosed(SourceHandle sourceHandle);
+    void onClosed(ISourceHandle sourceHandle);
+
+    void onFailure(ISourceHandle sourceHandle, Throwable t);
   }
 
   public interface SinkHandleListener {
-    void onFinish(SinkHandle sinkHandle);
+    void onFinish(ISinkHandle sinkHandle);
 
-    void onClosed(SinkHandle sinkHandle);
+    void onClosed(ISinkHandle sinkHandle);
 
-    void onAborted(SinkHandle sinkHandle);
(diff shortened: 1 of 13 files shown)


server/src/main/java/org/apache/iotdb/db/mpp/buffer/DataBlockManager.java
```diff
@@ -290,7 +312,7 @@ public ISinkHandle createSinkHandle(
             executorService,
             clientFactory.getDataBlockServiceClient(remoteEndpoint),
             tsBlockSerdeFactory.get(),
-            new SinkHandleListenerImpl(instanceContext));
+            new SinkHandleListenerImpl(instanceContext, instanceContext::failed));
```
xingtanzjr, line 315: Now that we use `failureCallback`, we can decouple `FragmentInstanceContext` with this method. I mean, we can change the parameter from `FragmentInstanceContext` to `FailureCallback`

server/src/main/java/org/apache/iotdb/db/mpp/execution/QueryExecution.java
```diff
@@ -210,8 +208,7 @@ public TsBlock getBatchResult() {
         return null;
       }
       return resultHandle.receive();
-
-    } catch (ExecutionException | IOException | CancellationException e) {
+    } catch (ExecutionException e) {
```
xingtanzjr, line 211: We can remove IOException but we need keep the `CancellationException` here. Because the blocked.get() may throw the CancellationException which is necessary to be caught here.

server/src/main/java/org/apache/iotdb/db/mpp/buffer/DataBlockManager.java
cornmonster: You are right.

But to achieve that, we have to introduce a few more callbacks like onFinishedCallback and onClosedCallback. Do you think it is necessary to do it or leaving a TODO is good for now?

server/src/main/java/org/apache/iotdb/db/mpp/execution/QueryExecution.java
cornmonster: OK, will fix it.

server/src/main/java/org/apache/iotdb/db/mpp/buffer/DataBlockManager.java
xingtanzjr: Sure we can leave a TODO now

xingtanzjr (approved):

PR #5666, opened 2022-04-25
Change assertion to prevent IoTDBInterpreterTest from failing due to indeterminate ordering

MarcyGO:
Some tests in the class `org.apache.zeppelin.iotdb.IoTDBInterpreterTest` can fail based on the order of iteration in `ConcurrentHashMap`. These tests are `testShowTimeseries`, `testShowDevices`, `testShowDevicesWithSg`, `testShowAllTTL`, and `testShowStorageGroup`.

I found the issue using [NonDex](https://github.com/TestingResearchIllinois/NonDex):
```
mvn test-compile -pl tsfile -am
mvn -pl Zeppelin-interpreter edu.illinois:nondex-maven-plugin:1.1.2:nondex -Dtest=org.apache.zeppelin.iotdb.IoTDBInterpreterTest
```
The test can fail because it assumes that the order of the data is determinate, and can be fixed it by ignoring the order.

### DIFF
--- a/zeppelin-interpreter/src/test/java/org/apache/zeppelin/iotdb/IoTDBInterpreterTest.java
+++ b/zeppelin-interpreter/src/test/java/org/apache/zeppelin/iotdb/IoTDBInterpreterTest.java
@@ -328,11 +328,7 @@ public class IoTDBInterpreterTest {
             + "root.test.wf01.wt01.hardware\tnull\troot.test.wf01\tFLOAT\tGORILLA\tSNAPPY\tnull\tnull";
     Assert.assertNotNull(actual);
     Assert.assertEquals(Code.SUCCESS, actual.code());
-    List<String> actual_list =
-        new ArrayList<String>(Arrays.asList(actual.message().get(0).getData().split("\n")));
-    List<String> expected_list = new ArrayList<String>(Arrays.asList(gt.split("\n")));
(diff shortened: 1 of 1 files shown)


zeppelin-interpreter/src/test/java/org/apache/zeppelin/iotdb/IoTDBInterpreterTest.java
```diff
@@ -389,4 +393,11 @@ public void testListUser() {
     Assert.assertEquals(Code.SUCCESS, actual.code());
     Assert.assertEquals(gt, actual.message().get(0).getData());
   }
+
+  private void assertEqualsNonStrict(String gt, String actual) {
+    List<String> actualList = new ArrayList<String>(Arrays.asList(actual.split("\n")));
+    List<String> expectedList = new ArrayList<String>(Arrays.asList(gt.split("\n")));
+    Assert.assertEquals(expectedList.get(0), actualList.get(0));
+    Assert.assertEquals(new HashSet<String>(expectedList), new HashSet<String>(actualList));
```
jixuan1989, line 401: This modification has a benifit that we can know which line of results are incorrect.
However, IMO, it is sensitive about the order of the resultset table. 
So, using Hashset to compare the result is not very good. 
We may need to use ArrayList to split the result and compare them line by line.

PR #5672, opened 2022-04-26
[IOTDB-3012][De-Singleton-1] remove singleton pattern for some classes

chengjianyun:
As have discussed, singleton pattern is making code inflexible, hard to make UI and IT tests. We shouldn't use singleton pattern too much. 

I'd like to propose several PRs to remove the singleton pattern for most of singleton pattern implementations before the `getInstance()` are widely used. This one is the first.


wangchao316 (approved):
LGTM

JackieTien97 (approved):

PR #5673, opened 2022-04-26
[To rel/0.13][IOTDB-2983] Serialization error in Partial insert

HeimingZ:
Serialization of failed data types and values causes error when redoing wal, so remove serialization of failed data types and values.


HTHou:
Add some tests?

PR #5679, opened 2022-04-26
Set correct InternalEndpoint and DataBlockEndpoint when planning

xingtanzjr:
## Description
Before this change, we did not distinguish the port of DataBlockManager and InternalService. And we have to reference it from IoTDBConfig in many places.

Now, we set the Endpoint to correct value when doing the distribution planning so that we can reference it by a more correct way.


server/src/main/java/org/apache/iotdb/db/mpp/execution/Coordinator.java
```diff
@@ -53,7 +53,12 @@ public class Coordinator {
   private static final String COORDINATOR_SCHEDULED_EXECUTOR_NAME = "MPPCoordinatorScheduled";
   private static final int COORDINATOR_SCHEDULED_EXECUTOR_SIZE = 1;
 
-  private static final TEndPoint LOCAL_HOST =
+  private static final TEndPoint LOCAL_HOST_DATA_BLOCK_ENDPOINT =
+      new TEndPoint(
+          IoTDBDescriptor.getInstance().getConfig().getRpcAddress(),
```
JackieTien97, line 58:
```suggestion
          IoTDBDescriptor.getInstance().getConfig().getInternalIp(),
```

server/src/main/java/org/apache/iotdb/db/mpp/execution/Coordinator.java
```diff
@@ -53,7 +53,12 @@ public class Coordinator {
...
+          IoTDBDescriptor.getInstance().getConfig().getDataBlockManagerPort());
+
+  private static final TEndPoint LOCAL_HOST_INTERNAL_ENDPOINT =
       new TEndPoint(
           IoTDBDescriptor.getInstance().getConfig().getRpcAddress(),
```
JackieTien97, line 63:
```suggestion
          IoTDBDescriptor.getInstance().getConfig().getInternalIp(),
```

server/src/main/java/org/apache/iotdb/db/mpp/execution/QueryExecution.java
```diff
@@ -300,9 +300,7 @@ private void initResultHandle() {
               .createSourceHandle(
                   context.getResultNodeContext().getVirtualFragmentInstanceId().toThrift(),
                   context.getResultNodeContext().getVirtualResultNodeId().getId(),
-                  new TEndPoint(
-                      context.getResultNodeContext().getUpStreamEndpoint().getIp(),
-                      IoTDBDescriptor.getInstance().getConfig().getDataBlockManagerPort()),
+                  context.getLocalDataBlockEndpoint(),
```
JackieTien97, line 303: The result handle in coordinator is sourceHanlde, endPoint of it should be upStreamEndPoint, not its own endPoint.
It shoud be `context.getResultNodeContext().getUpStreamEndpoint()`

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/mpp/execution/QueryExecution.java
xingtanzjr: Fixed

server/src/main/java/org/apache/iotdb/db/mpp/execution/Coordinator.java
xingtanzjr: FIxed

PR #5692, opened 2022-04-27
[IOTDB-3021] Fix sink/source handle memory leak

cornmonster:
1. Fix sink/source handle memory leak
2. Fix SourceHandle#setNoMoreTsBlocks doesn't trigger listener
3. rename SinkHandle#close to setNoMoreTsBlocks
4. rename SourceHandle#close to abort for consistency

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/mpp/memory/MemoryPool.java
+++ b/server/src/main/java/org/apache/iotdb/db/mpp/memory/MemoryPool.java
@@ -22,8 +22,9 @@ package org.apache.iotdb.db.mpp.memory;
 import com.google.common.util.concurrent.AbstractFuture;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
-import javax.annotation.Nullable;
 import org.apache.commons.lang3.Validate;
+
+import javax.annotation.Nullable;
 
 import java.util.HashMap;
 import java.util.Iterator;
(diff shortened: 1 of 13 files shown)


server/src/main/java/org/apache/iotdb/db/mpp/memory/MemoryPool.java
```diff
@@ -133,6 +133,31 @@ public boolean tryReserve(String queryId, long bytes) {
...
+      return 0L;
+    }
+    Iterator<MemoryReservationFuture<Void>> iterator = memoryReservationFutures.iterator();
+    while (iterator.hasNext()) {
+      MemoryReservationFuture<Void> f = iterator.next();
+      if (future.equals(f)) {
+        f.cancel(true);
+        iterator.remove();
+        return f.getBytes();
+      }
```
JackieTien97, line 155: Here, time complexity of removing is O(N), but we may not need to do real removing here. We can just cancel it. Removing job can be left to free method

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/mpp/memory/MemoryPool.java
cornmonster: Make sense. Fixed.

JackieTien97 (approved):

PR #5695, opened 2022-04-27
[IOTDB-3020][IOTDB-3022]Multi-thread MLog recovery & Fix SchemaFile dirty page loses

bigreybear:
### About Multi-Thread MLog Recovery
MLog.bin used to be recovered one by one, which may undermine the performance from the very begining.
This PR accelerate recovery by Multi-Thread style. 
### About SchemaFile Dirty Pages
When a single schema file contains many series like more than 20 millions, the page preallocated for child internal/entity might be lost during a write method.
This PR collect all the dirty pages during a write method and this exception shall not appear again.


MarcosZyk (approved):

server/src/main/java/org/apache/iotdb/db/metadata/schemaregion/SchemaEngine.java
```diff
@@ -143,10 +160,37 @@ public synchronized void createSchemaRegion(
...
+      } catch (MetadataException e) {
+        throw new RuntimeException(e);
+      }
+    };
+  }
+
+  public void createSchemaRegionWithoutExistenceCheck(
+      PartialPath storageGroup, SchemaRegionId schemaRegionId) throws MetadataException {
+    ISchemaRegion schemaRegion = null;
+    this.localStorageGroupSchemaManager.ensureStorageGroup(storageGroup);
```
MarcosZyk, line 190: Why we provide this as a public interface? It seems a creation with existence check is better to be provided.

bigreybear: fixed now

server/src/main/java/org/apache/iotdb/db/metadata/schemaregion/SchemaRegionSchemaFileImpl.java
```diff
@@ -430,10 +430,20 @@ public void createTimeseries(CreateTimeSeriesPlan plan) throws MetadataException
...
+          throw new MetadataException(
+              "IoTDB system load is too large to create timeseries, "
+                  + "please increase MAX_HEAP_SIZE in iotdb-env.sh/bat and restart");
+        }
+
+        logger.warn("IoTDB loads too many timeseries, waiting for IO swapping...");
+        Thread.sleep(10000L);
+      }
+    } catch (InterruptedException e) {
+      e.printStackTrace();
```
qiaojialin, line 445: use logger.log

bigreybear: fixed by an alternative impl

qiaojialin (approved):

PR #5702, opened 2022-04-27
[IOTDB-3026] fix cannot load tsfile correctly after restart iotdb

Cpaulyz:
The version number should start from 1, not from 0. Otherwise, partitionMaxFileVersions cannot be computed correctly after iotdb restart(in `updatePartitionFileVersion`). If load tsfile, the version number will be 0, which is duplicated.


yschengzi (approved):
LGTM

qiaojialin (approved):

PR #5707, opened 2022-04-28
[IOTDB-3030] delete storage group with ** error

yschengzi:
problem:
delete storage group with ** in path error, the storage group is created by select into statements.
error happens in parse delete storage group to device paths in sync.

solution:
block parse process when sync is not enable.
block storage groups with ** in path.


HTHou (commented):
Add some tests?

HTHou:
https://github.com/apache/iotdb/pull/5718

PR #5709, opened 2022-04-28
[IOTDB-3028] Fix PartitionCache when there are ** in path

SpriCoder:
See more details: https://issues.apache.org/jira/browse/IOTDB-3028


SpriCoder:
I have tested locally(3-confignode, 3-datanode), result is as follow:

image

xingtanzjr (approved):

PR #5736, opened 2022-04-29
[IOTDB-3046] Fix port binding error in ClientManagerTest

ericpai:
See JIRA: https://issues.apache.org/jira/browse/IOTDB-3046

### DIFF
--- a/node-commons/src/test/java/org/apache/iotdb/commons/ClientManagerTest.java
+++ b/node-commons/src/test/java/org/apache/iotdb/commons/ClientManagerTest.java
@@ -37,6 +37,7 @@ import org.junit.Before;
 import org.junit.Test;
 
 import java.io.IOException;
+import java.net.InetSocketAddress;
 import java.net.ServerSocket;
 
 public class ClientManagerTest {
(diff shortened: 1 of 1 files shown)


OneSizeFitsQuorum (approved):
LGTM

node-commons/src/test/java/org/apache/iotdb/commons/ClientManagerTest.java
```diff
@@ -408,7 +409,9 @@ public void InvalidAsyncClientReturnClientManagersTest() throws Exception {
   }
 
   public void startServer() throws IOException {
-    metaServer = new ServerSocket(9003);
+    metaServer = new ServerSocket();
+    metaServer.setReuseAddress(true);
+    metaServer.bind(new InetSocketAddress(9003));
```
JackieTien97, line 414:
```suggestion
    // reuse the port to avoid `Bind Address already in use` which is caused by TIME_WAIT state because port won't be usable immediately after we close it.
    metaServer.setReuseAddress(true);
    metaServer.bind(new InetSocketAddress(9003));
```

JackieTien97 (changes requested):

node-commons/src/test/java/org/apache/iotdb/commons/ClientManagerTest.java
ericpai: Fixed.

PR #5737, opened 2022-04-29
[IOTDB-2994] Separate RSchemaEngine into a separate module

THUMarkLau:
See [IOTDB-2994](https://issues.apache.org/jira/browse/IOTDB-2994).
The `RSchemaRegion` is a SchemaEngine based on RocksDB, which introduces Rocksdb as a denpendency of IoTDB. The size of jar file of rocksdb is about 45 MB, resulting in a considerable increment in size of distribution package for IoTDB.
This PR seperate RSchemaEngine as a seperate module. We don't have to package it in server package or distribution package. It can be pacakged separately if we need to use it.


MarcosZyk (approved):
LGTM~

cigarl (approved):
LGTM~
By the way, it may be better to put the jar packages related to rocksdb in the specified directory (eg., lib/rocksdb/*.jar) and then reload them all than to use jar name.

THUMarkLau:
> 
Get it

THUMarkLau:
> LGTM~ By the way, it may be better to put the jar packages related to rocksdb in the specified directory (eg., lib/rocksdb/*.jar) and then reload them all than to use jar name.

Get it.

qiaojialin (approved):

PR #5741, opened 2022-04-30
[IOTDB-3043] AuthorInfo serialization and deserialization interfaces

RYH61:
AuthorInfo serialization and deserialization interfaces


node-commons/src/main/java/org/apache/iotdb/commons/utils/BasicStructureSerDeUtil.java
```diff
@@ -239,4 +241,78 @@ public static Map<Integer, List<Integer>> readIntMapLists(ByteBuffer buffer) {
...
+    int length = readInt(buffer);
+    if (length == -1) {
+      return null;
+    }
+    List<String> strings = new ArrayList<>();
+    for (int i = 0; i < length; i++) {
+      strings.add(readString(buffer));
+    }
+    return strings;
+  }
```
CRZbulabula, line 317: Please add tests in BasicStructureSerDeUtilTest.

server/src/main/java/org/apache/iotdb/db/auth/authorizer/BasicAuthorizer.java
```diff
@@ -404,4 +410,105 @@ public void replaceAllUsers(Map<String, User> users) throws AuthException {
...
+          path = BasicStructureSerDeUtil.readString(buffer);
+        }
+        privilegeList.add(new PathPrivilege(privilegesId, path));
+      }
+      roleMap.put(roleName, new Role(roleName, privilegeList));
+    }
+
+    ((BasicUserManager) userManager).setUserMap(userMap);
+    ((BasicRoleManager) roleManager).setRoleMap(roleMap);
+  }
```
CRZbulabula, line 513: Please add serialize and deserialize tests for BasicAuthorizer.

CRZbulabula (changes requested):
PTAL~

node-commons/src/main/java/org/apache/iotdb/commons/utils/BasicStructureSerDeUtil.java
RYH61: fix

server/src/main/java/org/apache/iotdb/db/auth/authorizer/BasicAuthorizer.java
RYH61: fix

CRZbulabula (approved):
LGTM~

qiaojialin (commented):
Could we directly use the file of .user and .role for the auth snapshot?

RYH61:
> Could we directly use the file of .user and .role for the auth snapshot?
At present, when the system is started, the userMap has root information by default, but no other user information. When other users are used, they will be added to the user map. If you use these files directly, may not know which user information to restore.

PR #5744, opened 2022-04-30
[IOTDB-3048] Refactor binary transformer

flashzxi:
A lot of switch case statements are used in BinaryTransformer, and the enum 

TransformerType seems to be redundant. We can rewrite it wilh sbstract methods


flashzxi (commented):
rename

server/src/main/java/org/apache/iotdb/db/query/udf/core/transformer/binary/CompareNonEqualTransformer.java
```diff
@@ -29,12 +33,24 @@ public CompareNonEqualTransformer(
...
+    } else {
+      cachedBoolean =
+          evaluate(
+              castCurrentValueToDoubleOperand(leftPointReader),
+              castCurrentValueToDoubleOperand(rightPointReader));
+    }
+  }
+
+  protected boolean evaluate(boolean leftOperand, boolean rightOperand) {
+    return leftOperand == rightOperand;
```
SteveYurongSu, line 49:
```suggestion
    return leftOperand != rightOperand;
```

SteveYurongSu (approved):

PR #5696, opened 2022-04-27
[IOTDB-2981] Modify antlr, IoTDBSqlVisitor and ASTVisitor for new syntax convention

Plutooooooo:
### Description
In this PR, I mainly modified antlr, IoTDBSqlVisitor and ASTVisitor to adapt to new syntax convention. The logic of split path str to nodes in MetaUtils.spiltPathToDetachedPath is also changed. IoTDBSyntaxConventionStringLiteralIT and IoTDBSyntaxConventionIdentifierIT covers basic rules mentioned in syntax convention doc.
see [IOTDB-2391](https://issues.apache.org/jira/browse/IOTDB-2981) for detailed information.

PR #5756, opened 2022-05-01
[IOTDB-2797] Fix [privilege] root.** doesn't work

LIU-WEI-git:
Fix [IOTDB-2797](https://issues.apache.org/jira/browse/IOTDB-2797) and add related tests.


server/src/main/java/org/apache/iotdb/db/utils/AuthUtils.java
```diff
@@ -188,10 +188,14 @@ public static boolean validatePassword(String originPassword, String encryptPass
...
    * @param pathA sub-path
    * @param pathB path
-   * @return True if pathA == pathB, or pathA is an extension of pathB, e.g. pathA = "root.a.b.c"
-   *     and pathB = "root.a"
+   * @return True if pathA == pathB, or pathA is an extension of pathB, or pathA is contained in the
+   *     semantics of "**" in pathB, e.g. pathA = "root.a.b.c" and pathB = "root.a" or pathB =
+   *     "root.a.**"
    */
   public static boolean pathBelongsTo(String pathA, String pathB) {
+    if (pathB.endsWith("**")) {
```
SteveYurongSu, line 196: Can it handle the paths like `root.sg.d.*`?

LIU-WEI-git: OK, I'll add this feature.

server/src/main/java/org/apache/iotdb/db/utils/AuthUtils.java
```diff
@@ -188,10 +188,22 @@ public static boolean validatePassword(String originPassword, String encryptPass
...
-   * @return True if pathA == pathB, or pathA is an extension of pathB, e.g. pathA = "root.a.b.c"
-   *     and pathB = "root.a"
+   * @return True if pathA == pathB, or pathA is an extension of pathB, or pathA is contained in the
+   *     semantics of "*" and "**" in pathB, e.g. pathA = "root.a.b.c" and pathB = "root.a",
+   *     "root.a.b.*" or "root.a.**"
    */
   public static boolean pathBelongsTo(String pathA, String pathB) {
+    if (pathB.endsWith(".*")) {
+      if (pathA.endsWith("`")) {
+        pathA = pathA.substring(0, pathA.lastIndexOf(IoTDBConstant.PATH_SEPARATOR + "`"));
```
MarcosZyk, line 198: Why should we process back quote ``? It seems back quote will be removed after sql parse.

LIU-WEI-git: If the path node contains `.`, the lastIndexOf will get a value we don't want such as root.b.\`c.d\`

MarcosZyk: We use single quote '' or double quote "" to help use * in node name, rather than back quote ``.

server/src/main/java/org/apache/iotdb/db/utils/AuthUtils.java
```diff
@@ -188,10 +188,22 @@ public static boolean validatePassword(String originPassword, String encryptPass
...
+      } else {
+        pathA = pathA.substring(0, pathA.lastIndexOf(IoTDBConstant.PATH_SEPARATOR));
+      }
```
MarcosZyk, line 201: OK, I got it. It seems you should process single quote '' and double quote "" as well.

LIU-WEI-git: The PR #5696 make it only support back quote as the wrapper of the layer name in path, so we don't need consider single quote and double quote.

MarcosZyk:
@RYH61 PTAL

MarcosZyk:
The * and ** in privilege path seems useless, since we just need the prefix before them. What about we forbid user grant privilege on a path pattern? It is more clearly to allow user only grant privilege on a explicit prefix path.

RYH61:
> The * and ** in privilege path seems useless, since we just need the prefix before them. What about we forbid user grant privilege on a path pattern? It is more clearly to allow user only grant privilege on a explicit prefix path.

Ok, for root.* and root.** are the same permissions as root, so no need to add * and **，Using the prefix path shown is sufficient

SteveYurongSu (commented):
Please add some tests :D

SteveYurongSu (approved):
LGTM

LIU-WEI-git:
This PR will be reverted, please see the mail [Refactor the rule of auth check](https://lists.apache.org/thread/y9god4drws6x527qfcyonyyrro9j6qkn).

PR #5757, opened 2022-05-01
[IOTDB-2844] Implementation of Aggregator and Accumulator

Alima777:


server/src/main/java/org/apache/iotdb/db/mpp/aggregation/LastValueAccumulator.java
```diff
@@ -0,0 +1,305 @@
...
+        throw new UnSupportedDataTypeException(
+            String.format("Unsupported data type in LastValue: %s", seriesDataType));
+    }
+  }
+
+  @Override
+  public void addStatistics(Statistics statistics) {
+    switch (seriesDataType) {
+      case INT32:
+        updateIntLastValue((int) statistics.getLastValue(), statistics.getStartTime());
```
JackieTien97, line 104:
```suggestion
        updateIntLastValue((int) statistics.getLastValue(), statistics.getEndTime());
```

server/src/main/java/org/apache/iotdb/db/mpp/aggregation/LastValueDescAccumulator.java
```diff
@@ -0,0 +1,73 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iotdb.db.mpp.aggregation;
+
+import org.apache.iotdb.tsfile.file.metadata.enums.TSDataType;
+import org.apache.iotdb.tsfile.utils.Binary;
+
+public class LastValueDescAccumulator extends LastValueAccumulator {
```
JackieTien97, line 25: We should also override `addXXXInput` method, because we can break the loop in desc order.

server/src/main/java/org/apache/iotdb/db/mpp/aggregation/MaxTimeDescAccumulator.java
```diff
@@ -0,0 +1,41 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iotdb.db.mpp.aggregation;
+
+public class MaxTimeDescAccumulator extends MaxTimeAccumulator {
```
JackieTien97, line 22: We should override `addInput`, because we can break loop in desc order.

server/src/main/java/org/apache/iotdb/db/mpp/aggregation/MaxValueAccumulator.java
```diff
@@ -0,0 +1,270 @@
...
+      if (curTime >= timeRange.getMax() || curTime < timeRange.getMin()) {
+        break;
+      }
+      if (!column[1].isNull(i)) {
+        updateIntResult(column[1].getInt(i));
+      }
+    }
+  }
+
+  private void updateIntResult(int minVal) {
```
JackieTien97, line 204:
```suggestion
  private void updateIntResult(int val) {
```

JackieTien97: same as below

JackieTien97 (changes requested):

PR #5760, opened 2022-05-01
Demo: Inconvenient JDBC usage with keywords escaped

ericpai:


integration/src/test/java/org/apache/iotdb/db/integration/IoTDBKeywordQueryIT.java
```diff
@@ -0,0 +1,68 @@
...
+      }
+    }
+  }
+
+  /**
+   * columnWrapper is a special helper function to make business domain words compatible with IoTDB
+   * keywords
+   */
+  private String columnWrapper(String column) {
+    return String.format("`%s`", column);
```
Plutooooooo, line 66:
```suggestion
    return String.format("`%s`", column);
```
```suggestion
    if(column.contains("`")){
        column = column.replace("`", "``");
    }
    return String.format("`%s`", column);
```

Plutooooooo, line 66:
```suggestion
    return String.format("`%s`", column);
```
```suggestion
    if(column.contains("`")){
        column = column.replace("`", "``");
    }
    return String.format("`%s`", column);
```

Plutooooooo: Thanks for the example~ Syntax convention related PR(https://github.com/apache/iotdb/pull/5696) has just been merged, now this demo could run as expected, I think. I will look into this issue futher once I finished tsfile module related work. Thanks again, enjoy your holiday~

ericpai: @Plutooooooo Thanks, but it seems that the error is still existed in the query related codes (`rs.getInt(String columnName)`):
1. If the column is wrapped by ``, then the query of 
```
rs.getInt("root.sg.d1.`select`")
```
throws NPE.
2. If not, then the query of 
```
rs.getInt("root.sg.d1.www.baidu.com")
```
throws NPE.

Now as a user I don't know which column I should use the wrapper.

Plutooooooo: > @Plutooooooo Thanks, but it seems that the error is still existed in the query related codes (`rs.getInt(String columnName)`):
> 
> 1. If the column is wrapped by ``, then the query of
> 
> ```
> rs.getInt("root.sg.d1.`select`")
> ```
> 
> throws NPE. 2. If not, then the query of
> 
> ```
> rs.getInt("root.sg.d1.www.baidu.com")
> ```
> 
> throws NPE.
> 
> Now as a user I don't know which column I should use the wrapper.

Sorry I missed this part. It is confusing for the users indeed...... In SQL, when to use `` is specified, but in the header of result dataset, we only quote node that contains '`' or '.' which means node names that are keywords like "select" are not quoted. How to handle the header of result more properly is also an issue that needs further discussing.

PR #5763, opened 2022-05-02
[IOTDB-3034] Partial insert in new cluster

HeimingZ:
Log warn info in the data node and return success to the client when partial insert exists.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/mpp/sql/planner/plan/node/write/InsertRowNode.java
+++ b/server/src/main/java/org/apache/iotdb/db/mpp/sql/planner/plan/node/write/InsertRowNode.java
@@ -53,6 +53,7 @@ import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Objects;
 
@@ -159,7 +160,7 @@ public class InsertRowNode extends InsertNode implements WALEntryValue {
   }
 
   @Override
-  public boolean validateSchema(SchemaTree schemaTree) {
+  public boolean validateAndSetSchema(SchemaTree schemaTree) {
     DeviceSchemaInfo deviceSchemaInfo =
         schemaTree.searchDeviceSchemaInfo(devicePath, Arrays.asList(measurements));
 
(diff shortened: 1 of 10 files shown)


HTHou (approved):
LGTM

server/src/main/java/org/apache/iotdb/db/mpp/sql/planner/plan/node/write/InsertRowNode.java
```diff
@@ -223,7 +209,7 @@ public void transferType(List<MeasurementSchema> measurementSchemas)
       for (int i = 0; i < measurementSchemas.size(); i++) {
         if (measurementSchemas.get(i) == null) {
```
qiaojialin, line 210: what is equal to null mean?  add javadoc

server/src/main/java/org/apache/iotdb/db/mpp/sql/planner/plan/node/write/InsertRowNode.java
```diff
@@ -172,17 +173,16 @@ public boolean validateSchema(SchemaTree schemaTree) {
...
     } else {
-      // todo partial insert
       if (deviceSchemaInfo.isAligned() != isAligned) {
         return false;
       }
 
       for (int i = 0; i < measurementSchemas.size(); i++) {
         if (dataTypes[i] != measurementSchemas.get(i).getType()) {
-          if (IoTDBDescriptor.getInstance().getConfig().isEnablePartialInsert()) {
+          if (!IoTDBDescriptor.getInstance().getConfig().isEnablePartialInsert()) {
```
qiaojialin, line 182: add a test case

HeimingZ: UT or IT?

server/src/main/java/org/apache/iotdb/db/mpp/sql/planner/plan/node/write/InsertRowNode.java
HeimingZ: I'm also confused about this. @MrQuansy

server/src/main/java/org/apache/iotdb/db/mpp/sql/planner/plan/node/write/InsertRowNode.java
HTHou: We don't have IT for new cluster yet. For now, UT is enough.

PR #5765, opened 2022-05-02
[IOTDB-3075] RatisConsensus UUID bugfix

SzyWilliam:
## Description
Currently in Ratis, it requires 16-bytes-long ByteString as the unique RaftGroupID of a consensus raft group. However, internally Ratis extracts and only uses 6 bytes of the given RaftGroupID. In current group id encoding schema, we may encounter the situation where two 16bytes-id differ while there underlying 6bytes-id equals. Thus we need to reverse engineer and encode the total group information into 6bytes, uniquely.


OneSizeFitsQuorum (approved):
LGTM

PR #5767, opened 2022-05-02
Fix concurrent failure of testInsertMultiTabletPlanParallel

MarcosZyk:
### Cause
When isAutoCreateSchemaEnabled, the testInsertMultiTabletPlanParallel may throw npe, due to one of the resources, storage group, schemaRegionId table and schemaRegion, is not prepared ready by another thread.

### Solution
Ensure the resources during insert.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/localconfignode/LocalConfigNode.java
+++ b/server/src/main/java/org/apache/iotdb/db/localconfignode/LocalConfigNode.java
@@ -546,7 +546,9 @@ public class LocalConfigNode {
     // Thread B: access sg, access partitionTable to get schemaRegionId, access schemaEngine to get
     // schemaRegion
     // When A and B are running concurrently, B may get null while getting schemaRegionId or
-    // schemaRegion.
+    // schemaRegion. This means B must run after A ends.
+    // To avoid this exception, please invoke getBelongedSchemaRegionIdWithAutoCreate according to
+    // the scenario.
     if (schemaRegionId == null) {
       throw new MetadataException(
           String.format(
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/apache/iotdb/db/localconfignode/LocalConfigNode.java
```diff
@@ -539,7 +539,17 @@ public List<IStorageGroupMNode> getAllStorageGroupNodes() {
    */
   public SchemaRegionId getBelongedSchemaRegionId(PartialPath path) throws MetadataException {
     PartialPath storageGroup = storageGroupSchemaManager.getBelongedStorageGroup(path);
-    return schemaPartitionTable.getSchemaRegionId(storageGroup, path);
+    SchemaRegionId schemaRegionId = schemaPartitionTable.getSchemaRegionId(storageGroup, path);
+    if (schemaRegionId == null) {
+      throw new MetadataException(
+          String.format("Storage group %s has not been prepared well.", storageGroup));
```
qiaojialin, line 545: Print the schema region id.
SchemaRegion-x of Storage group %s is not registered？

server/src/main/java/org/apache/iotdb/db/localconfignode/LocalConfigNode.java
```diff
@@ -539,7 +539,17 @@ public List<IStorageGroupMNode> getAllStorageGroupNodes() {
...
+    }
+    ISchemaRegion schemaRegion = schemaEngine.getSchemaRegion(schemaRegionId);
+    if (schemaRegion == null) {
+      throw new MetadataException(
+          String.format("Storage group %s has not been prepared well.", storageGroup));
```
qiaojialin, line 550: SchemaRegion-x of Storage Group %s is not initiated.

server/src/main/java/org/apache/iotdb/db/localconfignode/LocalConfigNode.java
MarcosZyk: Already add comments in code to explain this check.

qiaojialin (approved):

PR #5775, opened 2022-05-03
[IOTDB-2919] Delete data/schema region through data/schema state machine

cmlmakahts:
Interface called by confignode in order to delete data/schema regions  at datanodes.  

Delete operation works through consensus layer.


CRZbulabula (commented):
PTAL:

- Don't forget to update this doc [DataNode for ConfigNode RPC interfaces](https://y8dp9fjm8f.feishu.cn/docs/doccnFe11jMpLgkzyZU3ZnzKqRb)
- Is it possible to add test for deleteRegion interface?

server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/plan/node/DeleteRegionNode.java
```diff
@@ -0,0 +1,109 @@
...
+import org.apache.iotdb.commons.consensus.ConsensusGroupId;
+import org.apache.iotdb.consensus.common.request.IConsensusRequest;
+import org.apache.iotdb.db.mpp.plan.analyze.Analysis;
+import org.apache.iotdb.tsfile.utils.ReadWriteIOUtils;
+
+import java.nio.ByteBuffer;
+import java.util.List;
+
+public class DeleteRegionNode extends WritePlanNode implements IConsensusRequest {
+
```
MarcosZyk, line 32: Shall we separate DeleteDataRegionNode and DeleteSchemaRegionNode from this?

MarcosZyk (commented):
PTAL

cmlmakahts:
See pr https://github.com/apache/iotdb/pull/5811

PR #5778, opened 2022-05-03
[IOTDB-2957] Fix a bug in authentication module

flashzxi:
# description
Login with  wrong password the server will also add this session into sessionId-username map. And all login check are depend on whether sessionId is in the map. Remove requestSessionId() when password is wrong


SteveYurongSu (approved):
LGTM

PR #5782, opened 2022-05-03
Handle the scenarios that there is no data for storage group

xingtanzjr:
## Description
Before this change, there are 2 scenarios which are not handled by query.
1. Query schema if the storage group exists but there is no timeseries inside it.
2. Query the data if the storage group exists but there is no timeseries and data inside it.

After this change, the query in these two scenarios are like:

1. query schema
image

2. query data
image


MarcosZyk (approved):
LGTM~

PR #5784, opened 2022-05-03
[IOTDB-3090] Fix the exception of the newly initialized consensus group read function && fix the recycle logic when RaftClient fails && Added exception handling for RaftClientReply

OneSizeFitsQuorum:
see [jira3090](https://issues.apache.org/jira/browse/IOTDB-3090) for detail


SzyWilliam (approved):
LGTM

PR #5791, opened 2022-05-04
[IOTDB-2845] Implementation of DeviceMergeOperator

Alima777:
The design doc is in https://y8dp9fjm8f.feishu.cn/docs/doccnYSHUUEi4xWkebcT6ITGpiP#


server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/DeviceMergeOperator.java
```diff
@@ -0,0 +1,306 @@
...
+  }
+
+  private void initTsBlockBuilderFromTsBlock(TsBlock tsBlock) {
+    dataTypes = tsBlock.getValueDataTypes();
+    tsBlockBuilder = new TsBlockBuilder(Arrays.asList(dataTypes));
+  }
+
+  /** DeviceColumn must be the first value column of tsBlock transferred by DeviceViewOperator. */
+  private String getDeviceNameFromTsBlock(TsBlock tsBlock) {
+    if (tsBlock.getColumn(0).isNull(0)) {
```
JackieTien97, line 268:
```suggestion
    if (tsBlock == null || tsBlock.getPositionCount() == 0 || tsBlock.getColumn(0).isNull(0)) {
```

tsfile/src/main/java/org/apache/iotdb/tsfile/read/common/block/TsBlock.java
```diff
@@ -213,6 +214,14 @@ public Column[] getColumns(int[] columnIndexes) {
     return columns;
   }
 
+  public TSDataType[] getValueDataTypes() {
+    TSDataType[] dataTypes = new TSDataType[valueColumns.length];
+    for (int i = 0; i < valueColumns.length; i++) {
+      dataTypes[i] = valueColumns[i].getDataType();
+    }
+    return dataTypes;
+  }
```
JackieTien97, line 223: Better not to try to get TsDataType info from TsBlock, because we may make TsBlock typeless which means Float and Int share the same Int32Column.

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/DeviceMergeOperator.java
```diff
@@ -0,0 +1,306 @@
...
+        break;
+      }
+    }
+    return finished;
+  }
+
+  private void initTsBlockBuilderFromTsBlock(TsBlock tsBlock) {
+    dataTypes = tsBlock.getValueDataTypes();
+    tsBlockBuilder = new TsBlockBuilder(Arrays.asList(dataTypes));
+  }
```
JackieTien97, line 264: You don't have to get dataType from child's TsBlock in runtime. Instead, you can get the dataType in `LocalExecutionPlanner` before you run the operator.

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/DeviceMergeOperator.java
```diff
@@ -0,0 +1,306 @@
...
+                  break;
+                case DOUBLE:
+                  valueColumnBuilders[j].writeDouble(
+                      deviceTsBlocks[i].getColumn(j).getDouble(rowIndex));
+                  break;
+                case TEXT:
+                  valueColumnBuilders[j].writeBinary(
+                      deviceTsBlocks[i].getColumn(j).getBinary(rowIndex));
+                  break;
+              }
```
JackieTien97, line 192: use `write(Column column, int index);` function in `ColumnBuilder` to remove this ugly switch-case.

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/DeviceMergeOperator.java
```diff
@@ -0,0 +1,306 @@
...
+ * in order of device. If the same device exists in different nodes, the tsBlocks need to be merged
+ * by time within the device.
+ *
+ * <p>The form of tsBlocks from input operators should be the same strictly, which is transferred by
+ * DeviceViewOperator.
+ */
+public class DeviceMergeOperator implements ProcessOperator {
+
+  private final OperatorContext operatorContext;
+  // The size devices and deviceOperators should be the same.
```
JackieTien97, line 52: Size of `devices` is same to size of all devices, but size of `deviceOperators` is equal to size of children operator, they are not same.

JackieTien97: BTW, I think we can totally get devices from children operator, we don't need to save this field in `DeviceMergeOperator`. It can be optimized in future.

JackieTien97 (changes requested):

PR #5792, opened 2022-05-04
[IOTDB-3080] Implementation of AlignedSeriesScanOperator

JackieTien97:


tsfile/src/main/java/org/apache/iotdb/tsfile/read/common/block/TsBlock.java
```diff
@@ -357,52 +357,94 @@ public Object[] next() {
...
+      return rowIndex < positionCount;
     }
 
     @Override
     public boolean hasNext(long minBound, long maxBound) {
-      while (super.hasNext() && currentValue() == null) {
+      while (hasNext()) {
         if (currentTime() < minBound || currentTime() >= maxBound) {
           break;
         }
```
Alima777, line 378: wrong?

JackieTien97: Maybe, @liuminghui233 is trying to fix that, it's not my fault, the previous code was like that.

Alima777 (approved):

PR #5800, opened 2022-05-05
[IOTDB-3016] do flush when graceful stop IoTDB

jamber001:
https://issues.apache.org/jira/browse/IOTDB-3016

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/service/IoTDBShutdownHook.java
+++ b/server/src/main/java/org/apache/iotdb/db/service/IoTDBShutdownHook.java
@@ -18,8 +18,10 @@
  */
 package org.apache.iotdb.db.service;
 
+import org.apache.iotdb.db.engine.StorageEngine;
 import org.apache.iotdb.db.engine.compaction.CompactionTaskManager;
 import org.apache.iotdb.db.utils.MemUtils;
+import org.apache.iotdb.db.wal.WALManager;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/iotdb/db/service/IoTDBShutdownHook.java
```diff
@@ -33,6 +35,11 @@ public void run() {
     CompactionTaskManager.getInstance().stop();
     // close rocksdb if possible to avoid lose data
     IoTDB.configManager.clear();
+
+    // == flush data to Tsfile and remove WAL log files
+    StorageEngine.getInstance().syncCloseAllProcessor();
+    WALManager.getInstance().deleteOutdatedWALFiles();
```
wangchao316, line 41: whether could call IoTDB#stop() ?  other some service need clear together.

jamber001: Current codes changes have been tested. It can work well.

Now, I am not sure whether calling IoTDB#stop() will bring other issue. I will test it at once.

jamber001: > whether could call IoTDB#stop() ? other some service need clear together.

I just tested this. 
If we call IoTDB#stop() in IoTDBShutdownHook.java, it will be hanged for very long time. Because some IService can not exit soon.
So, if no special need, I prefer to keep current code change, which can get good closing speed and keep data safe.

wangchao316: ok

jamber001:
@HTHou  Could you help merge this PR?    :-)

ericpai (approved):

PR #5818, opened 2022-05-06
[IOTDB-3108]Optimize register process of datanode

MrQuansy:
Register datanode with single config node in configuration files

For instance, we start a 3 config nodes cluster firstly.

Set one config node in iotdb-engine.properties and start datanode:

image

After starting three datanodes:

image


server/src/main/java/org/apache/iotdb/db/service/DataNode.java
```diff
@@ -158,10 +155,21 @@ public void joinCluster() throws StartupException {
...
+        }
+        config.setConfigNodeList(configNodeList);
+        IoTDBConfigCheck.getInstance().serializeConfigNodeList(configNodeList);
+
         if (dataNodeRegisterResp.getStatus().getCode()
                 == TSStatusCode.SUCCESS_STATUS.getStatusCode()
             || dataNodeRegisterResp.getStatus().getCode()
                 == TSStatusCode.DATANODE_ALREADY_REGISTERED.getStatusCode()) {
+          logger.info(
+              "Register current node using request {} with response {}", req, dataNodeRegisterResp);
```
CRZbulabula, line 172: Please print dataNodeRegisterResp.getStatus().getMessage() individually~

CRZbulabula (changes requested):
PTAL

server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfigCheck.java
```diff
@@ -108,6 +112,8 @@ public class IoTDBConfigCheck {
 
   private static final String DATA_NODE_ID = "data_node_id";
 
+  private static final String CONFIG_NODE_LIST = "config_node_list";
+
```
HTHou, line 116: Change the class name to `IoTDBStartCheck`?

server/src/main/java/org/apache/iotdb/db/service/DataNode.java
MrQuansy: Add it~

server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfigCheck.java
MrQuansy: Ok~

CRZbulabula (approved):
LGTM.

server/src/main/java/org/apache/iotdb/db/client/ConfigNodeClient.java
```diff
@@ -169,10 +168,17 @@ public TDataNodeRegisterResp registerDataNode(TDataNodeRegisterReq req)
     for (int i = 0; i < RETRY_NUM; i++) {
       try {
         TDataNodeRegisterResp resp = client.registerDataNode(req);
-        if (!updateConfigNodeLeader(resp.status)) {
+
+        // check if config nodes are inconsistent
```
qiaojialin, line 172: describe more detail, confignode_list fetched from ConfigNode are inconsistent with which?

MrQuansy: I change the process here... When it needs to redirect to config node leader, set latest config node list in client to reconnect.

server/src/main/java/org/apache/iotdb/db/service/DataNode.java
```diff
@@ -158,21 +155,36 @@ public void joinCluster() throws StartupException {
...
+        IoTDBStartCheck.getInstance().serializeConfigNodeList(configNodeList);
+
         if (dataNodeRegisterResp.getStatus().getCode()
                 == TSStatusCode.SUCCESS_STATUS.getStatusCode()
             || dataNodeRegisterResp.getStatus().getCode()
                 == TSStatusCode.DATANODE_ALREADY_REGISTERED.getStatusCode()) {
+          logger.info("Register current node successfully using request {}", req);
+          logger.info(dataNodeRegisterResp.getStatus().getMessage());
           int dataNodeID = dataNodeRegisterResp.getDataNodeId();
           if (dataNodeID != config.getDataNodeId()) {
```
qiaojialin, line 174: is this needed?

server/src/main/java/org/apache/iotdb/db/service/DataNode.java
```diff
@@ -158,21 +155,36 @@ public void joinCluster() throws StartupException {
         req.setDataNodeLocation(location);
 
         TDataNodeRegisterResp dataNodeRegisterResp = configNodeClient.registerDataNode(req);
+
+        // store config node lists from resp
+        List<TEndPoint> configNodeList = new ArrayList<>();
+        for (TConfigNodeLocation configNodeLocation : dataNodeRegisterResp.getConfigNodeList()) {
+          configNodeList.add(configNodeLocation.getInternalEndPoint());
+        }
```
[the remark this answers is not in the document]
MrQuansy: I think we can not get information from a local client. DataNode should get latest config nodes from 'DataNodeRegisterResp' sent by config node. And the config nodes in configNodeClient may be not latest ones.

server/src/main/java/org/apache/iotdb/db/service/DataNode.java
MrQuansy: I remove the first one. But the message in resp gives important notice, such as telling users to add 2 more datanodes to enable service because of replication of datanode.

HTHou (approved):

qiaojialin (approved):

PR #5821, opened 2022-05-07
Fix order by when aligned by device

luciferz2021:
## Description

In version 0.13, when aligned by device, the records are sorted within device so the order is wrong.

image

It is fixed by update the implementation of AlignByDeviceDataSet.java

image

<hr>

This PR has:
- [x] been self-reviewed.
    - [x] concurrent read
    - [x] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

- AlignByDeviceDataSet.java
- RawDataQueryPlan.java
- ServerTimeGenerator.java

### DIFF
--- a/server/src/assembly/resources/conf/iotdb-engine.properties
+++ b/server/src/assembly/resources/conf/iotdb-engine.properties
@@ -932,3 +932,9 @@ timestamp_precision=ms
 ####################
 # Datatype: float
 # group_by_fill_cache_size_in_mb=1.0
+
+####################
+### Master-Slave Sync Configuration
+####################
+# Datatype: boolean
+enable_sync=true
--- a/server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
+++ b/server/src/main/java/org/apache/iotdb/db/qp/executor/PlanExecutor.java
@@ -277,6 +277,7 @@ public class PlanExecutor implements IPlanExecutor {
         return true;
       case INSERT:
         insert((InsertRowPlan) plan);
+        PlanSyncer.SyncInsertRowPlan((InsertRowPlan) plan);
         return true;
       case BATCH_INSERT_ONE_DEVICE:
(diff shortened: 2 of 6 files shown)


Alima777:
Hi, in align by device query, we see each device as one relational table with (device + timestamp) as the primary key. When querying multiple devices, the result is sorted by device order first, and then by timestamp order. Therefore, **it's not a wrong order, but the current standard we defined.**

However, this definition does not satisfy some users' cases indeed. But if modified simply as you did, maybe other users still want the result order before. So we will redefine the semantics in the next version and support these two order at the same time.

luciferz2021:
> Hi, in align by device query, we see each device as one relational table with (device + timestamp) as the primary key. When querying multiple devices, the result is **sorted by device order first, and then by timestamp order.** Therefore, **it's not a wrong order, but the current standard we defined.**
> 
> However, this standard does not satisfy some users' cases indeed. But **if modified simply as you did, maybe other users still want the result order before.** So we have been tried to redefine the semantics in the next version and support these two order at the same time.
> 
> Thank you for the contribution anyway :D

Is there any document or web page that describes the new semantics? We want to use this feature in our system. If the semantics have been redefined, we can adopt to it first in our inner version to minimize the effort to upgrade later.

JackieTien97:
> > Hi, in align by device query, we see each device as one relational table with (device + timestamp) as the primary key. When querying multiple devices, the result is **sorted by device order first, and then by timestamp order.** Therefore, **it's not a wrong order, but the current standard we defined.**
> > However, this standard does not satisfy some users' cases indeed. But **if modified simply as you did, maybe other users still want the result order before.** So we have been tried to redefine the semantics in the next version and support these two order at the same time.
> > Thank you for the contribution anyway :D
> 
> Is there any document or web page that describes the new semantics? We want to use this feature in our system. If the semantics have been redefined, we can adopt to it first in our inner version to minimize the effort to upgrade later.

It seems that docs doesn't explain this semantics, we will add it soon.

PR #5837, opened 2022-05-09
[IOTDB-3038] Implementation of fill operator

JackieTien97:


server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/fill/previous/BinaryPreviousFill.java
```diff
@@ -0,0 +1,83 @@
...
+        return new RunLengthEncodedColumn(BinaryColumnBuilder.NULL_VALUE_BLOCK, size);
+      } else {
+        return new RunLengthEncodedColumn(
+            new BinaryColumn(1, Optional.empty(), new Binary[] {value}), size);
+      }
+    } else {
+      Binary[] array = new Binary[size];
+      boolean[] isNull = new boolean[size];
+      // have no null value
+      boolean nonNullValue = true;
```
Alima777, line 61: It cost my lots of time to understand this parameter... Maybe keep it consistent with isNull[] array is a better choice. Which means, just changing the parameter to `hasNullValue`, whose default value is false.

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/fill/previous/BinaryPreviousFill.java
```diff
@@ -0,0 +1,83 @@
...
+    int size = valueColumn.getPositionCount();
+    // if this valueColumn doesn't have any null value, or it's empty, just return itself;
+    if (!valueColumn.mayHaveNull() || size == 0) {
+      if (size != 0) {
+        previousIsNull = false;
+        // update the value using last non-null value
+        value = valueColumn.getBinary(size - 1);
+      }
+      return valueColumn;
+    }
```
Alima777, line 48: I  recommend coding this way...
```suggestion
    if (size == 0) {
       return valueColumn;
    }
    if (!valueColumn.mayHaveNull()) {
        previousIsNull = false;
        // update the value using last non-null value
        value = valueColumn.getBinary(size - 1);
        return valueColumn;
      }
```

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/fill/previous/BinaryPreviousFill.java
```diff
@@ -0,0 +1,83 @@
...
+    // if its values are all null
+    if (valueColumn instanceof RunLengthEncodedColumn) {
+      if (previousIsNull) {
+        return new RunLengthEncodedColumn(BinaryColumnBuilder.NULL_VALUE_BLOCK, size);
```
Alima777, line 52: Use `NullColumn.create()`  instead ~

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/fill/linear/LinearFill.java
```diff
@@ -0,0 +1,177 @@
...
+ */
+package org.apache.iotdb.db.mpp.execution.operator.process.fill.linear;
+
+import org.apache.iotdb.tsfile.read.common.block.column.Column;
+import org.apache.iotdb.tsfile.read.common.block.column.RunLengthEncodedColumn;
+import org.apache.iotdb.tsfile.read.common.block.column.TimeColumn;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+public abstract class LinearFill {
```
Alima777, line 27:
```suggestion
/**
*The result of Linear Fill functions at timestamp "T" is calculated by performing a linear fitting method on two timeseries values, one is at the closest timestamp before T, and the other is at the closest timestamp after T. Linear Fill function calculation only supports numeric types including int, double and float.
*/
public abstract class LinearFill {
```

Alima777: Add comment please

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/LinearFillOperator.java
```diff
@@ -0,0 +1,189 @@
...
+            return nextBlock;
+          } else { // otherwise, we cache it
+            cachedTsBlock.add(nextBlock);
+          }
+        }
+
+        // next TsBlock has already been in the cachedTsBlock
+        while (nextTsBlockIndex[currentFilledColumnIndex] < cachedTsBlock.size()) {
+          TsBlock nextTsBlock = cachedTsBlock.get(nextTsBlockIndex[currentFilledColumnIndex]);
+          nextTsBlockIndex[currentFilledColumnIndex]++;
```
Alima777, line 123: I think you abused the word `next`. So much `next` in your code, `nextTime`, `nextTimeInCurrentColumn`, `nextBlock`, `nextTsBlockIndex`. The key is they DON'T mean exactly the same thing. Look at the code, `nextBlock` just means next Block, but `nextTsBlock` is the same thing? You have to rename some parameters.

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/LinearFillOperator.java
```diff
@@ -0,0 +1,189 @@
...
+        // current column's filling is not finished using current owning TsBlock
+        if (cachedFilledValueColumns[currentFilledColumnIndex] == null) {
+          if (noMoreTsBlock) { // there is no more TsBlock, just use current owing TsBlock to fill
+            cachedFilledValueColumns[currentFilledColumnIndex] =
+                fillArray[currentFilledColumnIndex].fill(
+                    block.getTimeColumn(), block.getColumn(currentFilledColumnIndex));
+          } else { // next TsBlock is not ready, more TsBlocks is needed to do current column's
+            // fill, so current calculation is not finished, and we just return null
+            return null;
+          }
```
Alima777, line 144: From my perspective, it's easy to control time slice that you return null when you can't get enough information. But I still don't think It's a good way. The code is a bit confusing. I will support just getting next tsBlock until you can fill value using current information.

Alima777, line 144: From my perspective, it's easy to control time slice that you return null when you can't get enough information. But I still don't think It's a good way. The code is a bit confusing. I will support just getting next tsBlock until you can fill value using current information. Then the code will be very clear. You fill the value, if you can't, you get the following tsBlock.

Alima777 (changes requested):

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/fill/previous/BinaryPreviousFill.java
JackieTien97: NullColumn still need to switch-case to decide dataType, here we can directly create the corresponding NULL_VALUE_BLOCK.

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/LinearFillOperator.java
JackieTien97: If so, it won't be consistent with our design philosophy. We should not let Operator to keep calling child.next, it should only call once because before you call next, you should also call isBlocked and hasNext which will also make the Operator code confusing, you should also dealing with blocked situation in each Operator.
Confusing code is not bad if it's correct and well-annotated.

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/LinearFillOperator.java
JackieTien97: `nextBlock` is same as `nextTsBlock`, I will rename `nextBlock` to `nextTsBlock`.

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/fill/previous/BinaryPreviousFill.java
JackieTien97: fix

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/fill/previous/BinaryPreviousFill.java
JackieTien97: fix

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/fill/linear/LinearFill.java
JackieTien97: fix

Alima777 (approved):
LGTM

PR #5839, opened 2022-05-09
[IOTDB-3122] win: stop-server.bat can not close server

jamber001:
https://issues.apache.org/jira/browse/IOTDB-3122
**1 bug + 1 enhance**

1) **Bug**: In windows, stop-server.bat can not close server. this bug should have existed since 2021.12.

2) **Enhance**: let stop-server.bat support flushing data before closeing IoTDB Server. (Please refet to [IOTDB-3016](https://issues.apache.org/jira/browse/IOTDB-3016))


jamber001:
@HTHou   Who can help review this PR?   :-)

HTHou:
I don't think put the binary `exe` files in our source code is a good idea...
Can you please provide more information about these two files?

jamber001:
> I don't think put the binary `exe` files in our source code is a good idea... Can you please provide more information about these two files?

In fact, I don't want to add the exe files.  :-)       The reason is:
1) Current IoTDB shutdown mechanism depends on signal  processing mechanism, which can work well in linux.
2) Though Win OS also have signal  processing mechanism, but WinOS defautly have no cmd or tool (similiar to Linux kill cmd)  to send different signal to iotdb process.

So, for letting Win OS has same shutdown flow to linux, I need to find 1 Win tool that has similiar function to linux kill cmd.
The tool is only used to send sigal to IoTDB process in win OS, and has no other special funciton for us.
But 
1) It is not easy to dev such tool in Win OS becasue of Windows' characters. 
2) I found 1 open source tool windows-kill.exe thant is similiar to linux Kill. But it is not easy to compile in our dev env.

So, I include the windows-kill (2 exe files) to IoTDB project.

HTHou:
@qiaojialin How do you think?

qiaojialin:
Hi, where are these tools come from, and what the license is?

jamber001:
> license

jamber001:
> > license
@qiaojialin 
The Tool is from :   https://github.com/ElyDotDev/windows-kill/
The license file :  https://github.com/ElyDotDev/windows-kill/blob/master/LICENSE

HTHou:
Hi, is it possible to get the binary file from https://github.com/ElyDotDev/windows-kill/releases? If so, we can import them by maven instead of uploading to code repo.

qiaojialin:
Hi, MIT license is ok for Apache, but we need to add this into LICENSE-binary if decided to import this in our binary distribution.  I suggest sending a mail to dev@ list to discuss this imporing:)

jixuan1989:
The source code repo and Apache source release can not has any binary files (so, it is not accepted to upload  the exe files in the repo) 
as for the binary release, yes we need to discuss and announce it..

jamber001:
**This JIRA IOTDB-3122 has 2 parts:** 
1) **Bug**: In windows, stop-server.bat can not close server. this bug should have existed since 2021.12.

2) **Enhance**: let stop-server.bat support flushing data before closeing IoTDB Server. (Please refet to [IOTDB-3016]

The bug part need to be fixed.
About the Enhance part "Windows support  flushing data before closeing IoTDB Server. " . Is it necessary?  If it is not mandatory, we may discard it. 

How do you think?

PR #5842, opened 2022-05-09
[IOTDB-3127] Fix tree traverse algorithm

MarcosZyk:
### 2. Add status record to avoid repeating check
In ```findLastMatch``` procedure, the node matching will be repeated, implemented as nested loop. Add status record to reduce repeating check.

#### 3.1 Cause
Before consuming the device node, the algorithm pushed the children into visitStack and the device node to ancestorStack. This will bring repeating device node name when constructing the result.

#### 3.2. Solution
Add ```shouldVisitSubtree``` variable to help control the process. When meeting a matched device node, save its matching info and context before return. After consuming the device node, process its subtree based on the saved info and context.


xingtanzjr:
LGTM

PR #5843, opened 2022-05-09
[IOTDB-3117][snapshot] add authInfo snapshot.

SpriCoder:
implement snapshot of authInfo.
When take snapshot, authInfo will copy both user folder and role folder into snapshot.
When load snapshot, authInfo will copy user folder and role folder from snapshot into system folder.


confignode/src/main/java/org/apache/iotdb/confignode/persistence/AuthorInfo.java
```diff
@@ -331,6 +348,121 @@ public PermissionInfoResp executeListUserPrivileges(AuthorReq plan) throws AuthE
...
+    File userTmpSnapshotDir =
+        systemFileFactory.getFile(userSnapshotDir.getAbsolutePath() + "-" + UUID.randomUUID());
+    File roleFolder = systemFileFactory.getFile(commonConfig.getRoleFolder());
+    File roleSnapshotDir = systemFileFactory.getFile(snapshotDir, roleSnapshotFileName);
+    File roleTmpSnapshotDir =
+        systemFileFactory.getFile(roleSnapshotDir.getAbsolutePath() + "-" + UUID.randomUUID());
+
+    boolean result = true;
+    synchronized (IAuthorizer.class) {
+      try {
```
MarcosZyk, line 365: If some other authorInfo write operation doesn't synchronize its process, will this synchronize mechanism work?

MarcosZyk (commented):
PTAL

confignode/src/main/java/org/apache/iotdb/confignode/persistence/AuthorInfo.java
MarcosZyk: It seems this synchronize is only used for avoid concurrent snapshot creation. The snapshot invoking program should take the responsibility to acquire the global write lock of configNode.

SpriCoder: I use ReentrantReadWriteLock in AuthorizerManager to fix this problem.

SpriCoder:
Addition: in this PR, I move FileUtils from server to node-commons.

qiaojialin (approved):

PR #5847, opened 2022-05-09
[IOTDB-3092] Pooling config node clients using ClientManager

MrQuansy:
Start data node successfully!
image

Check authority successfully!
image

Execute tasks with config nodes:
image


server/src/main/java/org/apache/iotdb/db/client/ConfigNodeClient.java
```diff
@@ -36,35 +44,41 @@
...
   private static final Logger logger = LoggerFactory.getLogger(ConfigNodeClient.class);
 
-  private static final int TIMEOUT_MS = 10000;
-
   private static final int RETRY_NUM = 5;
 
   public static final String MSG_RECONNECTION_FAIL =
       "Fail to connect to any config node. Please check server it";
 
+  private int connectionTimeout = 10000;
```
OneSizeFitsQuorum, line 80: Use `ClientPoolProperty.DefaultProperty.WAIT_CLIENT_TIMEOUT_MS`

OneSizeFitsQuorum (approved):
We are adding more and more singleton and clientManager...Maybe we need to talk about how to control their ceiling... @JackieTien97

qiaojialin (approved):

PR #5858, opened 2022-05-10
[IOTDB-3079] Implememtation of complete LogicalPlanner

liuminghui233:
In this PR, I implememtation of complete LogicalPlanner to support most query statements.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/LocalExecutionPlanner.java
+++ b/server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/LocalExecutionPlanner.java
@@ -194,7 +194,7 @@ public class LocalExecutionPlanner {
           new SeriesScanOperator(
               node.getPlanNodeId(),
               seriesPath,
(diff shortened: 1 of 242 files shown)


server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/LocalExecutionPlanner.java
```diff
@@ -138,10 +138,12 @@ public static LocalExecutionPlanner getInstance() {
   public DataDriver plan(
       PlanNode plan,
       TypeProvider types,
+      Map<String, List<Integer>> deviceToMeasurementIndexesMap,
```
JackieTien97, line 141: You could put this field into DeviceMergeNode.

JackieTien97 (approved):

server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/LocalExecutionPlanner.java
liuminghui233: fixed

PR #5859, opened 2022-05-10
[IOTDB-3144][IOTDB-3130][IOTDB-3133] Set the default pipe status of receiver after reboot to STOP

Cpaulyz:
[IOTDB-3144][IOTDB-3130][IOTDB-3133]
状态

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/sync/receiver/ReceiverService.java
+++ b/server/src/main/java/org/apache/iotdb/db/sync/receiver/ReceiverService.java
@@ -103,7 +103,7 @@ public class ReceiverService implements IService {
 
   private void checkPipe(String pipeName, String remoteIp, long createTime) throws IOException {
     PipeInfo pipeInfo = receiverManager.getPipeInfo(pipeName, remoteIp, createTime);
-    if (pipeName != null && pipeInfo.getStatus().equals(PipeStatus.STOP)) {
+    if (pipeInfo != null && pipeInfo.getStatus().equals(PipeStatus.STOP)) {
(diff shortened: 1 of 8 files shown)


server/src/main/java/org/apache/iotdb/db/sync/sender/service/TransportHandler.java
```diff
@@ -88,6 +88,7 @@ public void start() {
             SyncConstant.HEARTBEAT_DELAY_SECONDS,
             SyncConstant.HEARTBEAT_DELAY_SECONDS,
             TimeUnit.SECONDS);
+    sendHeartbeat(); // send a heartbeat immediately
```
yschengzi, line 91: It's bad for restart of IoTDB.

server/src/main/java/org/apache/iotdb/db/sync/receiver/ReceiverService.java
```diff
@@ -109,12 +101,20 @@ public synchronized void stopPipeServer() throws PipeServerException {
...
+    if (pipeInfo != null && pipeInfo.getStatus().equals(PipeStatus.STOP)) {
+      startPipe(pipeName, remoteIp, createTime);
+    }
+  }
+
   /** heartbeat RPC handle */
   public synchronized SyncResponse receiveMsg(SyncRequest request) {
     SyncResponse response = new SyncResponse(ResponseType.INFO, "");
     try {
       switch (request.getType()) {
```
yschengzi, line 115: I think it's ok for cancel heartbeat.

yschengzi (approved):
LGTM

PR #5861, opened 2022-05-11
[IOTDB-3146] add confignode to distribution package

HTHou:
## Description

In this PR
 * Add confignode in distribution package, we can now execute `mvn clean package -pl distribution -am -DskipTests` to get the compiled confignode package.
 * fix CI still runs old cluster tests
 * Add confignode in all-in-one package

Now the all-in-one structure is

image

### DIFF
--- a/.github/workflows/main-unix.yml
+++ b/.github/workflows/main-unix.yml
@@ -56,4 +56,4 @@ jobs:
         shell: bash
         # we do not compile client-cpp for saving time, it is tested in client.yml
         # old cluster is deprecated, skip the test
-        run: mvn -B clean verify -Dtest.port.closed=true -Dcluster.skip.test=true -P '!testcontainer'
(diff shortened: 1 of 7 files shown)


irvine0109:
Q1. Why set the separate folder for the config node and the data node?

Q2. Are the log folder and the data folder in both the `confignode` folder and the `datanode` folder?

I guess you design this folder structure to start the config node and data node in one distribution folder of one node.

HTHou:
> Q1. Why set the separate folder for the config node and the data node?

config node and data node have some sub directories or files with some names. Eg, data, logs, config/logback.xml... Separated folders are benefit to distinguish them.

> Q2. Are the log folder and the data folder in both the `confignode` folder and the `datanode` folder?

Yes, the log and the data folder can be separated by different node types.

PR #5892, opened 2022-05-13
[IOTDB-3160] TsFile will be corrupted when flushing memtable appears OOM

HeimingZ:
Corrupted TsFile doesn't recover when exceptions occurs in the flushing memtable procedure, because we only interrupt and do nothing else. To fix this problem, I drop the memTable and skip ending tsfile when meeting flush error.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileProcessor.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileProcessor.java
@@ -1191,6 +1191,22 @@ public class TsFileProcessor {
     }
   }
 
+  /** This method will synchronize the memTable and release its flushing resources */
+  private void syncReleaseFlushedMemTable(IMemTable memTable) {
+    synchronized (memTable) {
+      releaseFlushedMemTable(memTable);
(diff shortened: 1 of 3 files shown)


HTHou (approved):

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileProcessor.java
```diff
@@ -1236,7 +1252,24 @@ public void flushOneMemTable() {
                 tsFileResource.getTsFile().getName(),
                 e1);
           }
-          Thread.currentThread().interrupt();
+          // release resource
+          try {
+            syncReleaseFlushedMemTable(memTableToFlush);
+            for (CloseFileListener closeFileListener : closeFileListeners) {
+              closeFileListener.onClosed(this);
```
qiaojialin, line 1259: This will register this file as SealedTsFile into TsFileManager, which may cause bugs in query

HeimingZ: Fixed, I will set time index to its initial status.

PR #5894, opened 2022-05-13
[IOTDB-3091] Scripts of stop datanodes

MrQuansy:


qiaojialin (approved):
Only Stop DataNodet is ok, Remove is an empty implementation.

PR #5902, opened 2022-05-13
[IOTDB-3179] Printing logs when get/getOrCreate Partition in ConfigNode

CRZbulabula:
Currently, we don't know what PartitionTable that ConfigNode returns to DataNode, which is not conducive to development debug. So I have temporarily added logs for getSchemaPartition getOrCreateSchemaPartition getDataPartition, getOrCreateDataPartition  interfaces.


MarcosZyk (approved):

chinausers:
Why not check out a new branch.

PR #5921, opened 2022-05-16
[IOTDB-3200] [PartitionInfo] replace bytebuffer with stream

cigarl:
This PR has:
- [x] Replace bytebuffer with stream in `PartitionInfo`.
- [x] Fixed incorrect serialization of multiple sub-maps in `dataPartitionMap`
- [x] Resolve snapshot temporary file deletion failure.

### DIFF
--- a/node-commons/src/main/java/org/apache/iotdb/commons/partition/DataPartition.java
+++ b/node-commons/src/main/java/org/apache/iotdb/commons/partition/DataPartition.java
@@ -261,16 +261,9 @@ public class DataPartition extends Partition {
             seriesPartitionSlotMapEntry.getValue().entrySet()) {
           timePartitionSlotListEntry.getKey().write(protocol);
           ReadWriteIOUtils.write(timePartitionSlotListEntry.getValue().size(), dataOutputStream);
-          timePartitionSlotListEntry
-              .getValue()
-              .forEach(
-                  x -> {
-                    try {
-                      x.write(protocol);
-                    } catch (TException e) {
-                      throw new RuntimeException(e);
-                    }
-                  });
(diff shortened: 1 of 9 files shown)


node-commons/src/main/java/org/apache/iotdb/commons/partition/DataPartition.java
```diff
@@ -248,75 +245,76 @@ public void createDataPartition(
...
+          timePartitionSlotListEntry.getKey().write(protocol);
+          ReadWriteIOUtils.write(timePartitionSlotListEntry.getValue().size(), dataOutputStream);
+          timePartitionSlotListEntry
+              .getValue()
+              .forEach(
+                  x -> {
+                    try {
+                      x.write(protocol);
+                    } catch (TException e) {
+                      throw new RuntimeException(e);
```
CRZbulabula, line 271: You can delete this.

CRZbulabula (commented):
PTAL.

node-commons/src/main/java/org/apache/iotdb/commons/partition/DataPartition.java
cigarl: fixed

CRZbulabula (commented):
Don't forget to clean up redundant snapshot files when takeSnapshot returns false.

cigarl:
> Don't forget to clean up redundant snapshot files when takeSnapshot returns false.

Each module that operates on snapshots, such as `partitionInfo` and `clusterSchema`, already deleted temporary files with the finally clause.  There is no need to clean up again in `ConfigRequestExecutor`.

CRZbulabula (approved):
LGTM.

qiaojialin (approved):

PR #5925, opened 2022-05-17
[IOTDB-3024] Implement SchemaRegion Memory mode snapshot

MarcosZyk:
## Description

Implement snapshot feature for SchemaRegion Memory mode to speed up schemaRegion consensus group recovery

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/schemaregion/SchemaRegionMemoryImpl.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/schemaregion/SchemaRegionMemoryImpl.java
@@ -422,7 +422,14 @@ public class SchemaRegionMemoryImpl implements ISchemaRegion {
 
     try {
       logWriter.copyTo(mLogSnapshotTmp);
-      mLogSnapshot.delete();
+      if (mLogSnapshot.exists() && !mLogSnapshot.delete()) {
+        logger.error(
+            "Failed to delete old snapshot file {} while creating snapshot for schemaRegion {}.",
+            mLogSnapshot.getName(),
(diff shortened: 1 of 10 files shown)


server/src/main/java/org/apache/iotdb/db/metadata/schemaregion/SchemaRegionMemoryImpl.java
```diff
@@ -409,6 +413,72 @@ public synchronized void deleteSchemaRegion() throws MetadataException {
...
+    try {
+      logWriter.copyTo(mLogSnapshotTmp);
+      mLogSnapshot.delete();
+      if (!mLogSnapshotTmp.renameTo(mLogSnapshot)) {
+        logger.error(
+            "Failed to rename {} to {} while creating snapshot for schemaRegion {}.",
+            mLogSnapshotTmp.getName(),
+            mLogSnapshot.getName(),
+            schemaRegionId);
+        mLogSnapshot.delete();
```
xingtanzjr, line 432: Why is `delete()` invoked again here ?

SpriCoder (approved):
LGTM！

server/src/main/java/org/apache/iotdb/db/metadata/schemaregion/SchemaRegionMemoryImpl.java
MarcosZyk: I'm not sure whether the ```mLogSnapshot``` will be produced if the rename operation is failed. If not, the invoking is unnecessary.

xingtanzjr (approved):

PR #5951, opened 2022-05-19
[IOTDB-3072] add measurement syntax check in TSServiceImpl and DataNodeTSIServiceImpl

Plutooooooo:
## Description
This PR aims to ensure that restrictions on path in  RPC is the same as SQL statement.


SteveYurongSu (approved):
LGTM

PR #5956, opened 2022-05-19
[IOTDB-3238] Fix cannot insert data into data node

THUMarkLau:
See [IOTDB-3238](https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-3238).
The reason for this bug is that we did not implement `accept` method for plan node executed in DataRegionStateMachine.


MarcosZyk (approved):

liuminghui233 (approved):
The INSERT statement can now be executed successfully :)

PR #5960, opened 2022-05-19
[IOTDB-3239] fix missing aligned data

Plutooooooo:
## Description
For aligned path, sensor is "". see more details at: [IOTDB-3239](https://issues.apache.org/jira/browse/IOTDB-3239)


tsfile/src/main/java/org/apache/iotdb/tsfile/read/common/Path.java
```diff
@@ -106,7 +106,8 @@ public Path(String device, String measurement) {
       String[] deviceNodes = PathNodesGenerator.splitPathToNodes(device);
       this.device = transformNodesToString(deviceNodes, deviceNodes.length);
       this.measurement = measurement;
-      this.fullPath = device;
+      // for aligned path, sensor is ""
```
JackieTien97, line 109:
```suggestion
      // for aligned path, sensor name for time column is ""
```

JackieTien97 (changes requested):

PR #5968, opened 2022-05-20
[IOTDB-2880] Fix procedure worker threads do not await correctly and …

cmlmakahts:
1. Fixed procedure worker threads do not await correctly
2. Remove some useless import and classes

66e7c115dce5e5a1806e7dfd9520c77


OneSizeFitsQuorum (approved):
LGTM

PR #5992, opened 2022-05-23
Fix time iterator bug

Alima777:
hasNext() method in iterator should not move iterator to next postition.


liuminghui233 (approved):
LGTM :-)

PR #5995, opened 2022-05-23
[IOTDB-3268] Refactor measurement check logic in session api for better performance

Plutooooooo:


server/src/main/java/org/apache/iotdb/db/service/thrift/impl/DataNodeTSIServiceImpl.java
```diff
@@ -1196,28 +1222,55 @@ private QueryId genQueryId(long id) {
...
-  protected void isLegalMeasurementLists(List<List<String>> measurementLists) throws TException {
+  protected void isLegalMeasurementLists(List<List<String>> measurementLists)
+      throws MetadataException {
     if (measurementLists == null) {
       return;
     }
+    StringBuilder path = new StringBuilder(IoTDBConstant.PATH_ROOT);
     for (List<String> measurementList : measurementLists) {
-      isLegalMeasurements(measurementList);
+      for (String measurement : measurementList) {
```
mychaow, line 1232: why not just call isLegalMeasurement?

server/src/main/java/org/apache/iotdb/db/service/thrift/impl/TSServiceImpl.java
```diff
@@ -2131,28 +2138,55 @@ private void addOperationLatency(Operation operation, long startTime) {
...
-  protected void isLegalMeasurementLists(List<List<String>> measurementLists) throws TException {
+  protected void isLegalMeasurementLists(List<List<String>> measurementLists)
+      throws MetadataException {
     if (measurementLists == null) {
       return;
     }
+    StringBuilder path = new StringBuilder(IoTDBConstant.PATH_ROOT);
     for (List<String> measurementList : measurementLists) {
-      isLegalMeasurements(measurementList);
+      for (String measurement : measurementList) {
```
mychaow, line 2148: same as above, the logic seems duplicated

mychaow (commented):
PTAL

server/src/main/java/org/apache/iotdb/db/service/thrift/impl/DataNodeTSIServiceImpl.java
Plutooooooo: because call isLegalMeasurement needs to do PathUtil.isLegalPath() size() times. By concating string, we only need to do that once.

mychaow: the path is concat all measurement of the  two level list?

Plutooooooo: It's only for check. Measurement is supposed to be a single node name(exclude root). For example, measurementLists is {{"a","b"},{"c","d"}} and we need to check each node name. Check each node individually is the same as check "a.b.c.d". Hope  it is clearer now.

mychaow: ok, got it.

mychaow (approved):
LGTM

Plutooooooo:
> LGTM

thanks for your review~

SteveYurongSu (approved):

PR #5998, opened 2022-05-24
[IOTDB-3184] Implement Timeseries version and blacklist

MarcosZyk:
### Implement Timeseries version
1. When generating createTimeseriesNode, a version will be generated for timeseries possibly created by this request
2. Store version in measurementMNode and DataNodeSchemaCache, and transport it via MeasurementPath and SchemaTree
3. There's no need to persist version in mlog, since the black list will only be constructed during system running and it's just a runtime check

### Add blacklist to SchemaValidator
The blacklist is mainly used to filtering insertion. Thus it is set in SchemaValidator. All the fetched schema should go through the blacklist check.

### Remove alias from DataNodeSchemaCache
The alias feature is developed to help data query rather than insertion. Remove it from DataNodeSchemaCache to simplify the process of alias updating operation.

### Future work
1. Implement blacklist operation
2. Implement DeteleTimeseries based on version and blacklist


yifuzhou (approved):
LGTM~ just kindly reminder that storageGroup should also has the version in confignode cache.

xingtanzjr (approved):

PR #6002, opened 2022-05-24
[IOTDB-3258] implement count nodes

SpriCoder:
See details: https://issues.apache.org/jira/browse/IOTDB-3258.


server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/LogicalPlanner.java
```diff
@@ -423,6 +424,15 @@ public PlanNode visitCountLevelTimeSeries(
           .getRoot();
     }
 
+    @Override
+    public PlanNode visitCountNodes(CountNodesStatement countStatement, MPPQueryContext context) {
+      LogicalPlanBuilder planBuilder = new LogicalPlanBuilder(context);
+      return planBuilder
+          .planChildPathsSchemaSource(countStatement.getPartialPath(), countStatement.getLevel())
+          .planCountMerge()
+          .getRoot();
```
yifuzhou, line 433: How about both sg above and sg below need to count together? like there has sg: root.ln, root.abc.a1.a2.a3 and ts:root.ln.a1.a2.a3. We want count nodes root.** level=3? I think planNodeManagementMemoryMerge() should be used.

confignode/src/test/java/org/apache/iotdb/confignode/service/thrift/ConfigNodeRPCServiceProcessorTest.java
```diff
@@ -1000,4 +1003,61 @@ private void cleanUserAndRole() throws TException {
       Assert.assertEquals(TSStatusCode.SUCCESS_STATUS.getStatusCode(), status.getCode());
     }
   }
+
+  @Test
+  public void testDeleteStorageGroup() throws TException {
```
yifuzhou, line 1008: It seems that previous pr changes has been removed... This function is the same as deleteStorageGroupTest(). I just change the name. Test class uses ....Test format and test function uses test... format should be more normalized.

yifuzhou (commented):
PTAL~

confignode/src/test/java/org/apache/iotdb/confignode/service/thrift/ConfigNodeRPCServiceProcessorTest.java
SpriCoder: Ok, get it! I will have a check.

server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/LogicalPlanner.java
SpriCoder: From my perspective, there is not allowed to have both root.ln and root.ln.a1.a2.a3?

SpriCoder: Besides, I use get child path in count node, so if there is possible having same path?

SpriCoder:
Already fixed! PTAL @yifuzhou

yifuzhou (approved):
LGTM!

MarcosZyk:
If the conduction cost of the above suggestion is heave, the suggested simplified modification is that we should not extend the interface of child management with a parameter level, which is confusing, instead we should implement independent node management configRequest and NodeScanOperator.

server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/LogicalPlanBuilder.java
```diff
@@ -607,20 +607,31 @@ public LogicalPlanBuilder planLevelTimeSeriesCountSource(
     return this;
   }
 
-  public LogicalPlanBuilder planChildPathsSchemaSource(PartialPath partialPath) {
-    this.root = new ChildPathsSchemaScanNode(context.getQueryId().genPlanNodeId(), partialPath);
+  public LogicalPlanBuilder planChildPathsSchemaSource(PartialPath partialPath, Integer level) {
```
MarcosZyk, line 610: Rename this to planNodePathsSchemaSource, or anything without child but related to node management

server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/LocalExecutionPlanner.java
```diff
@@ -467,40 +467,51 @@ public Operator visitLevelTimeSeriesCount(
 
     @Override
     public Operator visitChildPathsSchemaScan(
```
MarcosZyk, line 469: Rename to ```visitNodePathsSchemaScan```

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/schema/NodePathsConvertOperator.java
```diff
@@ -0,0 +1,100 @@
...
+
+    for (int i = 0; i < block.getPositionCount(); i++) {
+      String path = block.getColumn(0).getBinary(i).toString();
+      PartialPath partialPath;
+      try {
+        partialPath = new PartialPath(path);
+      } catch (IllegalPathException e) {
+        LOGGER.warn("Failed to convert node path to PartialPath {}", path);
+        continue;
+      }
```
MarcosZyk, line 77: Hope for an improvement of generic support on TsBlock. Thus the try-catch will be saved.

MarcosZyk (commented):
Excellent work! Only a few problems. PTAL.

server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/LocalExecutionPlanner.java
SpriCoder: fixed.

server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/LogicalPlanBuilder.java
SpriCoder: fixed.

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/schema/NodePathsConvertOperator.java
SpriCoder: I have the same idea.

MarcosZyk (approved):

PR #6013, opened 2022-05-25
[IOTDB-3290] modify check logic of CreateTimeSeries

Plutooooooo:


server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/Analyzer.java
```diff
@@ -887,16 +887,18 @@ public Analysis visitCreateTimeseries(
...
-      if (!props.containsKey(IoTDBConstant.COLUMN_TIMESERIES_DATATYPE.toLowerCase())) {
-        throw new SemanticException("datatype must be declared");
+      if (props.containsKey(IoTDBConstant.COLUMN_TIMESERIES_DATATYPE.toLowerCase())) {
+        String datatypeString =
+            props.get(IoTDBConstant.COLUMN_TIMESERIES_DATATYPE.toLowerCase()).toUpperCase();
+        try {
+          createTimeSeriesStatement.setDataType(TSDataType.valueOf(datatypeString));
+          props.remove(IoTDBConstant.COLUMN_TIMESERIES_DATATYPE.toLowerCase());
+        } catch (Exception e) {
+          throw new SemanticException(String.format("Unsupported datatype: %s", datatypeString));
```
MarcosZyk, line 898: Does this means the message of props is put in to the dataType?

Plutooooooo: datatype, encoding, compressor are no longer key words. It implicitly forbids users to set a property named datatype.

PR #6051, opened 2022-05-27
[IOTDB-3168][To rel/0.13] Fix the path with * could be executed successfully when inserting

23931017wu:
### Create a new function in bb
  /**  check whether the node name uses "*" correctly */
  private static void checkNameFormatIfHasstar(String name) throws MetadataException {
    if (!((name.startsWith("'") && name.endsWith("'"))
            || (name.startsWith("\"") && name.endsWith("\"")))
            &&  name.contains("*")

    ) {
      throw new MetadataException(String.format("%s is an illegal name.", name));
    }
  }

### Content2 ...
  /** check whether the given path is well formatted */
  public static void checkTimeseries(PartialPath timeseries) throws IllegalPathException {
    try {
      checkCharacters(timeseries.getFullPath());
    } catch (MetadataException e) {
      throw new IllegalPathException(timeseries.getFullPath(), e.getMessage());
    }
    for (String name : timeseries.getNodes()) {
      try {
        checkReservedNames(name);
        checkNameFormatIfHasstar(name);
        checkNameFormat(name);
      } catch (MetadataException e) {
        throw new IllegalPathException(timeseries.getFullPath(), e.getMessage());
      }
    }
  }

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/utils/MetaFormatUtils.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/utils/MetaFormatUtils.java
@@ -82,7 +82,7 @@ public class MetaFormatUtils {
   private static void checkNameFormat(String name) throws MetadataException {
     if (!((name.startsWith("'") && name.endsWith("'"))
             || (name.startsWith("\"") && name.endsWith("\"")))
-        && (name.contains(".") ||  name.contains("*"))) {
+        && (name.contains(".") || name.contains("*"))) {
       throw new MetadataException(String.format("%s is an illegal name.", name));
     }
   }
--- a/server/src/main/java/org/apache/iotdb/db/metadata/utils/MetaFormatUtils.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/utils/MetaFormatUtils.java
@@ -53,7 +53,6 @@ public class MetaFormatUtils {
     for (String name : timeseries.getNodes()) {
       try {
         checkReservedNames(name);
-        checkNameFormatIfHasstar(name);
         checkNameFormat(name);
       } catch (MetadataException e) {
         throw new IllegalPathException(timeseries.getFullPath(), e.getMessage());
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/apache/iotdb/db/service/IoTDB.java
```diff
@@ -109,7 +109,7 @@ public void active() {
 
     // set recover config, avoid creating deleted time series when recovering wal
     boolean prevIsAutoCreateSchemaEnabled = config.isAutoCreateSchemaEnabled();
-    config.setAutoCreateSchemaEnabled(false);
+    config.setAutoCreateSchemaEnabled(true);
```
mychaow, line 112: why change the mode? We need to disable auto create when recover.

server/src/main/java/org/apache/iotdb/db/metadata/utils/MetaFormatUtils.java
```diff
@@ -87,6 +88,16 @@ private static void checkNameFormat(String name) throws MetadataException {
     }
   }
 
+  /**  check whether the node name uses "*" correctly */
+  private static void checkNameFormatIfHasstar(String name) throws MetadataException {
+    if (!((name.startsWith("'") && name.endsWith("'"))
```
mychaow, line 93: use backquote, not the quotation

mychaow (changes requested):

server/src/main/java/org/apache/iotdb/db/service/IoTDB.java
23931017wu: Okay, got it.The state was changed to find the error.

HTHou (commented):
lgtm

Consider adding some IT cases?

23931017wu:
Modified running result：
image


IoTDB>  create timeseries root.test.*.a.b
 create timeseries root.test.*.a.b
Msg: 401: Error occurred while parsing SQL to physical plan: line 1:28 mismatched input '*' expecting {STRING_LITERAL, INTEGER_LITERAL, ID, QUTOED_ID_IN_NODE_NAME}
IoTDB> insert into root.test.*.a(timestamp, b) values(1509465600000, true)
insert into root.test.*.a(timestamp, b) values(1509465600000, true)
Msg: 313: failed to insert measurements [b] caused by root.test.*.a.b is not a legal path, because * is an illegal name.
IoTDB>  show timeseries root.test.**
 show timeseries root.test.**
+----------+-----+-------------+--------+--------+-----------+----+----------+
|timeseries|alias|storage group|dataType|encoding|compression|tags|attributes|
+----------+-----+-------------+--------+--------+-----------+----+----------+
+----------+-----+-------------+--------+--------+-----------+----+----------+
Empty set.
It costs 0.025s
IoTDB>

mychaow:
LGTM

mychaow (approved):

HTHou (approved):

PR #6075, opened 2022-05-30
[To rel/0.12][IOTDB-3306] Use rpc port check to avoid starting same IoTDB twice

HTHou:


server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
```diff
@@ -713,7 +713,7 @@ public class IoTDBConfig {
 
   /** the size of ioTaskQueue */
   private int ioTaskQueueSizeForFlushing = 10;
-
+le
```
HTHou, line 716:
```suggestion
l
```

server/src/main/java/org/apache/iotdb/db/service/IoTDB.java
```diff
@@ -106,6 +106,10 @@ private void setUp() throws StartupException {
 
     Runtime.getRuntime().addShutdownHook(new IoTDBShutdownHook());
     setUncaughtExceptionHandler();
```
mychaow, line 108: maybe rpc is enable, but mtree is not ready, so user send a request will call npe exception?

mychaow, line 108: maybe rpc is enable, but mtree is not ready, so user send a request will call npe exception?

qiaojialin (approved):

PR #6076, opened 2022-05-30
[IOTDB-3335] Redundant data recovered by wal in multi-leader consensus

HeimingZ:
Some data will be recovered wrongly because wal is kept, but memtable id still starts from 0 when using multi-leader consensus. So, I recover memtable id when using multi-leader consensus.


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/DataRegion.java
```diff
@@ -3541,7 +3542,9 @@ public IDTable getIdTable() {
   }
 
   public IWALNode getWALNode() {
-    if (!config.isClusterMode()) {
+    if (!config
+        .getDataRegionConsensusProtocolClass()
+        .equals(ConsensusFactory.MultiLeaderConsensus)) {
       throw new UnsupportedOperationException();
     }
```
HTHou, line 3549: Is there any problem if started as a standalone IoTDB?

HeimingZ: No, only multi-leader consensus will invoke this method.

PR #6083, opened 2022-05-31
[IOTDB-2988] Cache leader in session

MrQuansy:


server/src/main/java/org/apache/iotdb/db/mpp/plan/execution/QueryExecution.java
```diff
@@ -360,6 +359,41 @@ private void initResultHandle() {
...
+          tsstatus.setCode(TSStatusCode.NEED_REDIRECTION.getStatusCode());
+          for (TEndPoint endPoint : redirectNodeList) {
+            subStatus.add(
+                StatusUtils.getStatus(TSStatusCode.NEED_REDIRECTION).setRedirectNode(endPoint));
+          }
+          tsstatus.setSubStatus(subStatus);
+        }
+      } else {
+        // single device
+        tsstatus.setRedirectNode(redirectNodeList.get(0));
```
OneSizeFitsQuorum, line 390: Set `NEED_REDIRECTION` ?

OneSizeFitsQuorum (commented):
Can we confirm that the first one should be the leader?

xingtanzjr:
LGTM

server/src/main/java/org/apache/iotdb/db/mpp/plan/execution/QueryExecution.java
MrQuansy: > 
We get data partition info from config nodes or local partition cache. Config nodes will set the leader at the first one in `TRegionReplicaSet`. When the leader has been transferred, we will update local partition cache afterwards.

MrQuansy: Session can use TSStatus.isSetRedirectNode to check if the redirect TEndPoint is set to cache locally.

OneSizeFitsQuorum: You could add some comments to these two questions both.

OneSizeFitsQuorum (approved):
LGTM

PR #6096, opened 2022-05-31
[IOTDB-3355] Adapt metric framework for config node and check metrics

SpriCoder:
See more details: https://issues.apache.org/jira/browse/IOTDB-3355.

In 3C 3D with metric module enable:

Config node's dashboard:
image

Data node's dashboard:
image
image
image
image


SteveYurongSu (approved):
LGTM

PR #6087, opened 2022-05-31
[IOTDB-3289][IOTDB-3291] Validate maxRegionCount

CRZbulabula:

PR #6127, opened 2022-06-01
[IOTDB-3261][IOTDB-3332] Make the RegionBalancer thread-safe

CRZbulabula:
The reason why cause the [IOTDB-3261](https://issues.apache.org/jira/browse/IOTDB-3261) and [IOTDB-3332](https://issues.apache.org/jira/browse/IOTDB-3332) is that, the RegionAllocator is now not thread-safe. Therefore, when the Regions are allocated concurrently will cause the NPE.

### DIFF
--- a/confignode/src/main/java/org/apache/iotdb/confignode/manager/load/balancer/RegionBalancer.java
+++ b/confignode/src/main/java/org/apache/iotdb/confignode/manager/load/balancer/RegionBalancer.java
@@ -41,12 +41,9 @@ import java.util.List;
 public class RegionBalancer {
 
   private final Manager configManager;
-  private final IRegionAllocator regionAllocator;
 
   public RegionBalancer(Manager configManager) {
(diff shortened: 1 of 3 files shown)


CRZbulabula:
The current PR should cherry-pick [THIS](https://github.com/apache/iotdb/pull/6087) before test

confignode/src/main/java/org/apache/iotdb/confignode/manager/load/balancer/RegionBalancer.java
```diff
@@ -41,12 +41,9 @@
 public class RegionBalancer {
 
   private final Manager configManager;
-  private final IRegionAllocator regionAllocator;
 
   public RegionBalancer(Manager configManager) {
```
wangchao316, line 45: why?    RegionBalancer have more instance?

PR #6135, opened 2022-06-01
[ISSUE-5881] Fix `FileOutputStream` not close automatically

### LINKED ISSUE #5881: Use try-with-resources or close this "FileOutputStream" in a "finally" clause.
Code Path:confignode/src/main/java/org/apache/iotdb/confignode/persistence/NodeInfo.java
image
Connections, streams, files, and other classes that implement the Closeable interface or its super-interface, AutoCloseable, needs to be closed after use. Further, that close call must be made in a finally block otherwise an exception could keep the call from being made. Preferably, when class implements AutoCloseable, resource should be created using "try-with-resources" pattern and will be closed automatically.

Failure to properly close resources will result in a resource leak which could bring first the application and then perhaps the box the application is on to their knees.

wallezhang:
<hr>

This PR has:
- [x] been self-reviewed.
    - [x] concurrent read
    - [x] concurrent write

Close #5881 

<hr>


qiaojialin (approved):
LGTM

PR #6144, opened 2022-06-02
[IOTDB-3374] NPE when AggregationOperator receive the null tsblock from the child

Alima777:


server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/AggregationOperator.java
```diff
@@ -98,14 +98,23 @@ public TsBlock next() {
     // update input tsBlock
     curTimeRange = timeRangeIterator.nextTimeRange();
```
liuminghui233, line 99: May cause current time window be skipped before it has been calculated

liuminghui233 (changes requested):

wangchao316 (approved):
LGTM

PR #6160, opened 2022-06-05
[To rel/0.12] add log in query when a file does not contain a device that should in

qiaojialin:
When a device should be in a TsFile, we throw an exception when query. 
I change this to a warn log with file name.


wangchao316 (approved):
LGTM

Issue #6171, opened 2022-06-06
In java client, what function is create timeseries after set schema template?


Coding-Deng:
For example, I create a schema template:
`       
        Template template = new Template("treeTemplate_1", true);
        TemplateNode iNodeGPSX = new MeasurementNode("x", TSDataType.FLOAT, TSEncoding.RLE, CompressionType.SNAPPY);
        template.addToTemplate(iNodeGPSX);
        session.createSchemaTemplate(template);
`
then set it in a prefix:
`
session.setSchemaTemplate("treeTemplate_java", "root.java.unaligned");
`
the next step in JDBC, the next code is 
`
create timeseries of schema template on root.java.unaligned
`
but in java client, I don't know what function like this?

qiaojialin:
(1) Yes, there is not a NoSQL interface for this SQL, we will add one later.

Except for Writing data, other NoSQL APIs are just the simple form of SQL(maybe can not cover all function of SQL). 

So, if you find there is a SQL do not has a NoSQL API, you could execute it by one of the two interfaces:
executeQueryStatement(), executeNonQueryStatement()

In this case, the create timeseires of schema template is a nonQuery, so you could use the executeNonQueryStatement interface.

(2) For schema template, "create timeseries of schema template" is not needed, you could directly write data under the root.java.unaligned, we will create timeseries and use the template automatically.

bigreybear:
Thanks for your advice! and PR #6180 and #6181 (for rel.13) may solve the problme. PTAL~

PR #6180, opened 2022-06-07
[ISSUE-6171]Support createTimeseriesOfTemplate in Session 

bigreybear:
As issue-#6171 suggested, this PR adds an interface in `Session` and `SessionPool` to support create timeseries of schema template on a designated path if had been set a template, which acts like the sql-statement sharing the same syntax.


MarcosZyk (commented):
If we keep the old version standalone mode, this pr is OK to be merged. Otherwise, we can pending this to when we start developing Template for new cluster and new standalone.

MarcosZyk:
Already implemented in new cluster.

PR #6191, opened 2022-06-08
[IOTDB-3422]Fix metric tag in compaction task

xinzhongtianxia:
see https://issues.apache.org/jira/browse/IOTDB-3422


SpriCoder (approved):
LGTM!

PR #6196, opened 2022-06-08
[IOTDB-3425] [IOTDB-3415] add permission operate to LocalConfigNode && Abstract an interface for Standalone Authority Check

RYH61:
In this PR:
1、Add an interface IAuthorityFetcher，class StandaloneAuthorityFetcher and class ClusterAuthorityFetcher implement this interface。
2、Add standalone permission operations in the class StandaloneAuthorityFetcher and the class LocalConfigNode
     Authority Check、login、permission write、permission read
3、move permission cache to ClusterAuthorityFetcher


HTHou (approved):
LGTM

server/src/main/java/org/apache/iotdb/db/auth/StandaloneAuthorityFetcher.java
```diff
@@ -0,0 +1,128 @@
...
+
+  @Override
+  public TSStatus checkUserPrivileges(String username, List<String> allPath, int permission) {
+    boolean checkStatus = true;
+    String checkMessage = null;
+    for (String path : allPath) {
+      try {
+        if (!checkOnePath(username, path, permission)) {
+          checkStatus = false;
+        }
```
neuyilan, line 78: if return false when checking one path, we should just break the loop.

for example, if we have 3 paths,the user have no permission  in the second path, in this current implemention, the final result of `checkStatus` is true.

server/src/main/java/org/apache/iotdb/db/auth/StandaloneAuthorityFetcher.java
```diff
@@ -0,0 +1,128 @@
...
+      TAuthorizerReq authorizerReq, ConfigNodeClient configNodeClient) {
+    SettableFuture<ConfigTaskResult> future = SettableFuture.create();
+    try {
+      LocalConfigNode.getInstance().operatorPermission(authorizerReq);
+    } catch (AuthException e) {
+      future.setException(e);
+    }
+    future.set(new ConfigTaskResult(TSStatusCode.SUCCESS_STATUS));
+    return future;
+  }
```
neuyilan, line 114: in the final, the future will be set `TSStatusCode.SUCCESS_STATUS` regardless of whether exception happens or not?

The same as others,

server/src/main/java/org/apache/iotdb/db/auth/ClusterAuthorityFetcher.java
```diff
@@ -0,0 +1,337 @@
...
+    } catch (TException | IOException e) {
+      logger.error("Failed to connect to config node.");
+      status = new TPermissionInfoResp();
+      status.setStatus(
+          RpcUtils.getStatus(
+              TSStatusCode.EXECUTE_STATEMENT_ERROR, "Failed to connect to config node."));
+    } finally {
+      if (status == null) {
+        status = new TPermissionInfoResp();
+      }
```
neuyilan, line 241: What is the meaning of this code？
And I think if this code is triggered, this method will return empty status contains nothing.

server/src/main/java/org/apache/iotdb/db/auth/ClusterAuthorityFetcher.java
```diff
@@ -0,0 +1,337 @@
...
+  /**
+   * Initialize user and role cache information.
+   *
+   * <p>If the permission information of the role changes, only the role cache information is
+   * cleared. During permission checking, if the role belongs to a user, the user will be
+   * initialized.
+   */
+  public boolean invalidateCache(String username, String roleName) {
+    if (userCache.getIfPresent(username) != null) {
+      List<String> roleList = userCache.getIfPresent(username).getRoleList();
```
neuyilan, line 260: this may produce 'NullPointerException'

server/src/main/java/org/apache/iotdb/db/auth/ClusterAuthorityFetcher.java
```diff
@@ -0,0 +1,337 @@
...
+  /**
+   * Convert user privilege information obtained from confignode to PathPrivilege
+   *
+   * @param path permission path
+   * @param privilege privilegeIds
+   * @return
+   */
+  private PathPrivilege toPathPrivilege(String path, String privilege) {
+    PathPrivilege pathPrivilege = new PathPrivilege();
+    String[] privileges = privilege.replace(" ", "").split(",");
```
neuyilan, line 320: can use `privilege.trim()` to replase `privilege.replace(" ", "")`

neuyilan (changes requested):

server/src/main/java/org/apache/iotdb/db/auth/ClusterAuthorityFetcher.java
RYH61: trim() can only delete the leading and trailing spaces. Here,  need to remove the middle spaces as well

server/src/main/java/org/apache/iotdb/db/auth/StandaloneAuthorityFetcher.java
RYH61: fixed~

server/src/main/java/org/apache/iotdb/db/auth/StandaloneAuthorityFetcher.java
RYH61: fixed~

server/src/main/java/org/apache/iotdb/db/auth/ClusterAuthorityFetcher.java
RYH61: fixed~

server/src/main/java/org/apache/iotdb/db/auth/ClusterAuthorityFetcher.java
RYH61: Deleted~

neuyilan (approved):
LGTM

PR #6197, opened 2022-06-08
[ISSUE-5885] Synchronize this method to match the synchronization on "setModFile".

### LINKED ISSUE #5885: Synchronize this method to match the synchronization on "setModFile".
Code Path:server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
image

When one part of a getter/setter pair is synchronized the other part should be too. Failure to synchronize both sides of a pair may result in inconsistent behavior at runtime as callers access an inconsistent method state.

This rule raises an issue when either the method or the contents of one method in a getter/setter pair are synchrnoized but the other is not.

dafei1288:
### fix #5885

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
@@ -848,8 +848,10 @@ public class TsFileResource {
     return newResource;
   }
 
-  public synchronized void setModFile(ModificationFile modFile) {
-    this.modFile = modFile;
+  public void setModFile(ModificationFile modFile) {
+    synchronized (this) {
+      this.modFile = modFile;
+    }
   }
 
   /** @return resource map size */


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
```diff
@@ -848,8 +848,10 @@ public TsFileResource createHardlink() {
     return newResource;
   }
 
-  public synchronized void setModFile(ModificationFile modFile) {
-    this.modFile = modFile;
+  public void setModFile(ModificationFile modFile) {
```
ericpai, line 851: Why we need to change this. IMO they have the same meaning, just locking the object itself.

ericpai (changes requested):

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
dafei1288: I also think they have the same meaning.  So why it (#5885) is a "easy-fix" tag . I think it's probably designed to meet some kind of programming specification.  On the other hand, is it a worse choice to add a modifier ， ‘synchronized ’ ， to the setModFile method ?  Or I misunderstood the issure.

ericpai: It's wired indeed. Is it the matter of sonar-lint report? @MinaQin

xingtanzjr: Yes, it is a sonar-lint issue. See https://github.com/apache/iotdb/issues/5885. 

It says that the `synchronized` keywords must appear as a pair for `getter` and `setter`. Here, we tag the `synchronized` for `setModFile()` but don't do that for `getModFile()`. The easiest way to fix it is to add `synchronized` to `getModFile()`. But it is not very efficient because `getModFile()` did a concurrent control already. So... to resolve this issue and keep the original logic, we can put the `synchronized` into the method rather than tag it on the method.

ericpai: > Yes, it is a sonar-lint issue. See #5885.
> 
> It says that the `synchronized` keywords must appear as a pair for `getter` and `setter`. Here, we tag the `synchronized` for `setModFile()` but don't do that for `getModFile()`. The easiest way to fix it is to add `synchronized` to `getModFile()`. But it is not very efficient because `getModFile()` did a concurrent control already. So... to resolve this issue and keep the original logic, we can put the `synchronized` into the method rather than tag it on the method.

Got it.
Thanks for your contribution! @dafei1288

As our pull requests will be merged in a squash-and-rebase way. The PR title will be the final commit message in the target branch. So please do some additional work to refine this PR title. The formats are:
1. For an JIRA issue, the format is: [IOTDB-_jira_id_] _the intent of this PR_
2. For an Github issue, the format is: [ISSUE-_issue_id_] _the intent of this PR_

dafei1288: Sure , be happy to .

ericpai (approved):

PR #6200, opened 2022-06-08
[IOTDB-3424] avoid running duplicated IotDB servers

jamber001:
https://issues.apache.org/jira/browse/IOTDB-3424

### DIFF
--- a/server/src/assembly/resources/sbin/start-datanode.sh
+++ b/server/src/assembly/resources/sbin/start-datanode.sh
@@ -97,6 +97,69 @@ launch_service()
 	return $?
 }
 
+
+# check whether tool 'lsof' exists
+check_tool_env() {
+  if  ! type lsof > /dev/null 2>&1 ; then
+    echo ""
+    echo " Warning: No tool 'lsof', Please install it."
+    echo " Note: Some checking function need 'lsof'."
+    echo ""
+    return 1
+  else
+    return 0
+  fi
+}
+
+# convert path to real full-path.
+# If path has been deleted, return ""
+get_real_path() {
+  local path=$1
+  local real_path=""
+  cd $path > /dev/null 2>&1
+  if [ $? -eq 0 ] ; then
+    real_path=$(pwd -P)
+    cd -  > /dev/null 2>&1
+  fi
+  echo "${real_path}"
+}
+
+# check whether same directory's IotDB node process has been running
+check_running_process() {
+  check_tool_env
+
+  PIDS=$(ps ax | grep "$classname" | grep java | grep DIOTDB_HOME | grep -v grep | awk '{print $1}')
+  for pid in ${PIDS}
+  do
+    run_conf_path=""
+    run_cwd=$(lsof -p $pid 2>/dev/null | awk '$4~/cwd/ {print $NF}')
+    run_home_path=$(ps -fp $pid | sed "s/ /\n/g" | sed -n "s/-DIOTDB_HOME=//p")
+    run_home_path=$(get_real_path "${run_cwd}/${run_home_path}")
+
+    #if dir ${run_home_path} has been deleted
+    if [ "${run_home_path}" == "" ]; then
+      continue
+    fi
+
+    current_home_path=$(get_real_path ${IOTDB_HOME})
+    if [ "${run_home_path}" == "${current_home_path}" ]; then
+      echo ""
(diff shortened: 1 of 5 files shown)


jamber001:
@HTHou  Could you help review this PR?

HTHou:
Hi, would you like to change the DataNode class and the start-Datanode script as well?

Any idea about the Windows script?

HTHou:
Also, we have a new standalone IoTDB instance called NewIoTDB now. I think we should change it too.

jamber001:
> Hi, would you like to change the DataNode class and the start-Datanode script as well?
> 
> Any idea about the Windows script?
No problem, I will change  DataNode class and the start-Datanode script .

For Windows script, I am not sure. I will do some test at first.

jamber001:
> Also, we have a new standalone IoTDB instance called NewIoTDB now. I think we should change it too.
I will changed it oo.

jamber001:
@HTHou 
I have changed start-server.sh、 start-node.sh、start-new-server.sh and related .java files.

For Windows, I do some testing and think that windows can not use same solution.
1 possible solution is to generate 1 file to save server process pid.

server/src/assembly/resources/sbin/start-datanode.sh
```diff
@@ -97,6 +97,49 @@ launch_service()
...
+
+# check whether tool 'lsof' & 'realpath' exists
+check_tool_env() {
+  if  ! type lsof > /dev/null 2>&1 ; then
+    echo ""
+    echo " Error: No tool 'lsof', Please install it."
+    exit 1
+  fi
+
+  if  ! type realpath > /dev/null 2>&1 ; then
```
HTHou, line 109: If don't have this `realpath` command on computer, then the IoTDB cannot start? It seems not friendly to users  who is tried to install IoTDB on edge devices...  As I know, even macos doesn't contain this command....

jamber001: Strange. 
1) In past start stop scripts, they also depend on some third-party tool  such netstat, ps etc.  If they are absent, user need also to insteall .
2) When I dev this PR, I specially test realpath in Linux and Mac. Normally, Linux and Mac has tool realpath. Does your Mac have no this tool ?
3) For edge device, if it is linux OS. Normally, it contain realpath. Because realpath belong to coreutils .

jamber001: Another possible solution is to gen 1 file in data directroy to save pid info.
Next time start-server.sh will use this file to check whether last process is still alive.

HTHou: Ok, it seems that the problem only appeared on Arm-base macos. What you mentioned does make sense.

One more suggestion, if users don't have these command on computer, I think we should skip these check and allow IoTDB to be started.

jamber001: I have just submit new codes. It does not use tool realpath now..

HTHou (approved):
LGTM

server/src/main/java/org/apache/iotdb/db/service/DataNode.java
```diff
@@ -286,6 +286,11 @@ private void setUp() throws StartupException, QueryProcessException {
 
     registerManager.register(ReceiverService.getInstance());
 
+    // in cluster mode, RPC service is not enabled.
+    if (IoTDBDescriptor.getInstance().getConfig().isEnableRpcService()) {
+      registerManager.register(RPCService.getInstance());
+    }
+
```
HTHou, line 293: `RPCService` has already registered in `setUpRPCService()` method.

HTHou (changes requested):
Please merge the master branch again, the CI are able to pass now.

server/src/main/java/org/apache/iotdb/db/service/DataNode.java
jamber001: OK. I check this.

jamber001:
> Please merge the master branch again, the CI are able to pass now.

good news.  :-)

server/src/main/java/org/apache/iotdb/db/service/DataNode.java
jamber001: I have  just removed this part of codes.

HTHou (approved):
LGTM now

PR #6215, opened 2022-06-10
[IOTDB-3444] Make periodic serivce be scheduled more flexible

ericpai:
Here we add two another unsafe wrapper methods to let those tasks which wants to quit schedule once an unexcepted exception happens. Next step is to ensure all the periodic services use the correct error handling strategy.

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -56,7 +56,7 @@
             <property name="format" value="schedule(AtFixedRate|WithFixedDelay)\("/>
             <property name="severity" value="error"/>
             <property name="ignoreComments" value="true"/>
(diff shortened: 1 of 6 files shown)


ericpai:
Due to the Friday afternoon, the statstics in the end of this doc https://apache-iotdb.feishu.cn/docx/doxcnT0NAd2Ug80JXQ68HaTpo3g indicates that only the following four periodic services need to abort scheduling when unexpected RuntimeException is thrown:
- timedForceMLogThread
- WAL-Delete
- TTL-Check (Both in StorageEngine and StorageEngineV2)
@MarcosZyk @HeimingZ PTAL

MarcosZyk (approved):
LGTM~

PR #6232, opened 2022-06-10
[IOTDB-3439] Move Setting up RPC Service to the last step of starting DataNode 2

Cpaulyz:
`initProtocals()` after datanode is set up and registered with confignode succussfully.


neuyilan (approved):
LGTM

HTHou (approved):

PR #6235, opened 2022-06-11
[IoTDB-3124][IoTDB-3051]Initialize ConfigNode with multi-seed mode and ConfigNode shrinking process

leety1228:
1.Initialize ConfigNode with multi-seed mode;
   start-confignode.sh -s
2.Adding New ConfigNodes.
  start-confignode.sh -s
3.Remove ConfigNode out of Cluster.
  remove-confignode.sh -r ip:port


confignode/src/main/java/org/apache/iotdb/confignode/client/SyncConfigNodeClientPool.java
```diff
@@ -42,43 +44,143 @@ public class SyncConfigNodeClientPool {
...
+      List<TEndPoint> endPointList, TConfigNodeRegisterReq req) {
+    // TODO: Unified retry logic
+    for (TEndPoint endPoint : endPointList) {
+      for (int retry = 0; retry < retryNum; retry++) {
+        try (SyncConfigNodeIServiceClient client = clientManager.borrowClient(endPoint)) {
+          TConfigNodeRegisterResp resp = client.registerConfigNode(req);
+          if (updateConfigNodeLeader(resp.status)) {
+            client.close();
+            return clientManager.borrowClient(configNodeLeader).registerConfigNode(req);
+          } else {
```
neuyilan, line 99: can we make sure the `configNodeLeader` is not null?

confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeRemoveCheck.java
```diff
@@ -0,0 +1,89 @@
...
+      if (getConfigNdoeList().contains(nodeLocation)) {
+        return nodeLocation;
+      }
+    } catch (IOException | BadNodeUrlException e) {
+      LOGGER.error("Load system properties file failed.", e);
+    }
+
+    return nodeLocation;
+  }
+
```
neuyilan, line 68: The method may return an `EMPTY` TConfigNodeLocation,
so the method caller should check this or return null if the endPoint is not in the configNodeList. 
what do you think?

PR #6240, opened 2022-06-12
[IOTDB-3458] Fix incorrect error handling strategy of some periodic services

ericpai:
The following tasks has been fixed according to the doc https://apache-iotdb.feishu.cn/docx/doxcnT0NAd2Ug80JXQ68HaTpo3g
- timedForceMLogThread: From continuously to abort.
- Timed-Flush-Seq-Memtable(Both in StorageEngine and StorageEngineV2):  From continuously to abort, and remove outer try-catch.
- Timed-Flush-UnSeq-Memtable(Both in StorageEngine and StorageEngineV2):  From continuously to abort, and remove outer try-catch.
- TTL-check(Both in StorageEngine and StorageEngineV2): From abort to continuously.
- WAL-delete: From abort to continuously.


HeimingZ (approved):
LGTM :)

PR #6247, opened 2022-06-13
[IOTDB-3331] Fix create three level timeseries.

SpriCoder:
See more details: https://issues.apache.org/jira/browse/IOTDB-3331

Test on 3C3D:
image
image


node-commons/src/main/java/org/apache/iotdb/commons/utils/PathUtils.java
```diff
@@ -126,6 +126,10 @@ public static void isLegalMeasurements(List<String> measurements) throws Illegal
     }
   }
 
+  public static boolean match(String deviceName, String storageGroup) {
+    return deviceName.equals(storageGroup) || deviceName.startsWith(storageGroup + ".");
+  }
```
MarcosZyk, line 131: Maybe we can rename this method to ```isStartWith```.

qiaojialin (approved):

node-commons/src/main/java/org/apache/iotdb/commons/utils/PathUtils.java
SpriCoder: fixed~

qiaojialin (approved):

PR #6268, opened 2022-06-14
[IOTDB-3299] Migrate udf api to a seperate module

Plutooooooo:
## Description

Following changes are made:
1. UDF API is a independent module now.
2. Use Type as a substituion class for TSDataType in APIs. UDFDataTypeTransformer is responsible for managing transformation between Type and TSDataType.
3. Use udf.api.Binary as a substituion class for tsfile.util.Binary in APUs. UDFBinaryTransformer is responsible for managing transformation between udf.api.Binary and tsfile.util.Binary.


udf-api/src/main/java/org/apache/iotdb/udf/api/type/Binary.java
```diff
@@ -0,0 +1,133 @@
...
+    if (getClass() != other.getClass()) {
+      return false;
+    }
+
+    return compareTo((Binary) other) == 0;
+  }
+
+  @Override
+  public int hashCode() {
+    return Arrays.hashCode(values);
```
ericpai, line 88: If the `values` is not changed, the `hashCode()` may compute only once. So is there any places to use `setBytes()`? If not, we can let `values` be final and compute the hashCode in a lazy mode.

udf-api/src/main/java/org/apache/iotdb/udf/api/type/Binary.java
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  /**
+   * get length.
+   *
+   * @return length
+   */
+  public int getLength() {
+    if (this.values == null) {
+      return -1;
```
ericpai, line 98: IMO no matter whether `values` is null or an empty array, the length should be 0. Or it will be a little strage for caller.

udf-api/src/main/java/org/apache/iotdb/udf/api/type/Binary.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+    return this.values.length;
+  }
+
+  public String getStringValue() {
+    return new String(this.values, STRING_CHARSET);
```
ericpai, line 104: Like `hashCode()`, if the `values` is immutable, we can `getStringValue()` in a lazy mode.

node-commons/src/main/java/org/apache/iotdb/commons/udf/utils/UDFBinaryTransformer.java
```diff
@@ -0,0 +1,34 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iotdb.commons.udf.utils;
+
+/** Transform Binary to Binary / Binary to Binary */
```
ericpai, line 21: A strange comment :D . May refine like this `Transform between {@link org.apache.iotdb.tsfile.utils.Binary} and {@link org.apache.iotdb.tsfile.utils.Binary}`

udf-api/pom.xml
```diff
@@ -0,0 +1,31 @@
...
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>iotdb-parent</artifactId>
+        <groupId>org.apache.iotdb</groupId>
+        <version>0.14.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>udf-api</artifactId>
```
ericpai, line 30: I think this module should be added to `iotdb-distribution` and give out as a maven lib when new version is releasing. So developers can add it as their project's dependency to implement their UDFs.

ericpai (changes requested):

node-commons/src/main/java/org/apache/iotdb/commons/udf/utils/UDFDataTypeTransformer.java
```diff
@@ -0,0 +1,74 @@
...
+package org.apache.iotdb.commons.udf.utils;
+
+import org.apache.iotdb.tsfile.file.metadata.enums.TSDataType;
+import org.apache.iotdb.udf.api.type.Type;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+/** Transform Type to TsDataType / TsDataType to Type */
+public class UDFDataTypeTransformer {
```
ericpai, line 28: A strange comment :D . May refine like this Transform between {@link org.apache.iotdb.tsfile.file.metadata.enums.TSDataType} and {@link org.apache.iotdb.udf.api.type.Type}

node-commons/src/main/java/org/apache/iotdb/commons/udf/utils/UDFDataTypeTransformer.java
```diff
@@ -0,0 +1,74 @@
...
+ */
+package org.apache.iotdb.commons.udf.utils;
+
+import org.apache.iotdb.tsfile.file.metadata.enums.TSDataType;
+import org.apache.iotdb.udf.api.type.Type;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+/** Transform Type to TsDataType / TsDataType to Type */
```
ericpai, line 27: A strange comment :D . May refine like this Transform between {@link org.apache.iotdb.tsfile.file.metadata.enums.TSDataType} and {@link org.apache.iotdb.udf.api.type.Type}

udf-api/src/main/java/org/apache/iotdb/udf/api/type/Binary.java
Plutooooooo: In tsfile.util.Binary, this method returns -1 when values is null. But I think it's better to return 0, too.

ericpai: So we should make sure is it necessary to distinguish `empty` and `null` using `getLength()`? Or just adding another `isNull()` may be more clear.

udf-api/src/main/java/org/apache/iotdb/udf/api/type/Binary.java
Plutooooooo: Fixed.

udf-api/src/main/java/org/apache/iotdb/udf/api/type/Binary.java
Plutooooooo: Fixed.

node-commons/src/main/java/org/apache/iotdb/commons/udf/utils/UDFBinaryTransformer.java
Plutooooooo: Fixed.

node-commons/src/main/java/org/apache/iotdb/commons/udf/utils/UDFDataTypeTransformer.java
Plutooooooo: Fixed.

udf-api/src/main/java/org/apache/iotdb/udf/api/type/Binary.java
```diff
@@ -95,13 +102,16 @@ public int hashCode() {
...
-      return -1;
+      return 0;
     }
     return this.values.length;
   }
 
   public String getStringValue() {
-    return new String(this.values, STRING_CHARSET);
+    if (stringCache == null) {
+      stringCache = new String(this.values, STRING_CHARSET);
```
ericpai, line 116: Maybe we should check whether `this.values  == null`.

udf-api/src/main/java/org/apache/iotdb/udf/api/type/Binary.java
Plutooooooo: Perhaps we can add a isNull() method and let getLength() remain the same?

udf-api/src/main/java/org/apache/iotdb/udf/api/type/Binary.java
```diff
@@ -85,7 +89,10 @@ public boolean equals(Object other) {
 
   @Override
   public int hashCode() {
-    return Arrays.hashCode(values);
+    if (hash == 0) {
+      hash = Arrays.hashCode(values);
```
ericpai, line 93: there's a corner case... If the result just equals 0, then it will be computed every time... Maybe we should add another boolean value to indicate whether the hashCode has been calculated...

udf-api/src/main/java/org/apache/iotdb/udf/api/type/Binary.java
ericpai: Then we need to explain this in the javadoc comment.

udf-api/src/main/java/org/apache/iotdb/udf/api/type/Binary.java
Plutooooooo: Right... Using Integer needs Boxing and Unboxing, I added a boolean value.

udf-api/src/main/java/org/apache/iotdb/udf/api/type/Binary.java
```diff
@@ -90,13 +93,16 @@ public boolean equals(Object other) {
   @Override
   public int hashCode() {
     if (hash == 0) {
-      hash = Arrays.hashCode(values);
+      if (!hasCalculatedHash) {
```
ericpai, line 96: Just check `hasCalculatedHash` is enough. There's no need to check `hash == 0`.

ericpai (approved):

SteveYurongSu (approved):

PR #6279, opened 2022-06-14
[IOTDB-3489] Add test name to log folder in cluster test

ericpai:
The output will be like this
image

### DIFF
--- a/integration-test/src/main/java/org/apache/iotdb/it/env/ClusterEnvBase.java
+++ b/integration-test/src/main/java/org/apache/iotdb/it/env/ClusterEnvBase.java
@@ -47,12 +47,16 @@ public abstract class ClusterEnvBase implements BaseEnv {
   private List<ConfigNode> configNodes;
   private List<DataNode> dataNodes;
   private final Random rand = new Random();
+  private String nextTestCase = null;
 
   protected void initEnvironment(int configNodesNum, int dataNodesNum) throws InterruptedException {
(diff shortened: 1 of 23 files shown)


JackieTien97 (approved):

integration-test/src/main/java/org/apache/iotdb/it/env/ClusterEnvBase.java
```diff
@@ -47,12 +47,16 @@ public abstract class ClusterEnvBase implements BaseEnv {
...
   private final Random rand = new Random();
+  private String nextTestCase = null;
 
   protected void initEnvironment(int configNodesNum, int dataNodesNum) throws InterruptedException {
     this.configNodes = new ArrayList<>();
     this.dataNodes = new ArrayList<>();
 
     String testName = getTestClassName();
+    if (nextTestCase != null) {
+      testName = testName + "_" + nextTestCase;
```
irvine0109, line 58: Should we use `.` instead of `_`? 
In the name of a class or a method, `_` is legal, and `.` is illegal.

ericpai: Indeed `_` is a legal character in Java identifier and there's some chance to meet case conflict in theory. However in Java the identifier always follows camel case style and it's rare to use `_` in an identifier. What's more, `.` may have special meaning in some file system as it will be a part of the dictionary. So I would like to use `_` not `.`.

PR #6285, opened 2022-06-14
[IOTDB-3182] Judge DataNode status through heartbeat(Simple version)

CRZbulabula:
Temporary, we judge the DataNode's running status and load score by the following simple logic:

1. If the ConfigNode-leader doesn't receive any heartbeat from a DataNode within **20 seconds**, that DataNode is considered faulty and its status is set to Unknown.
2. Cluster read/write requests are always sent to the DataNode with the lowest load score. We currently believe that the newer the received heartbeat sample is,  the lower the load score is.


confignode/src/main/java/org/apache/iotdb/confignode/manager/load/heartbeat/HeartbeatCache.java
```diff
@@ -18,34 +18,58 @@
...
 public class HeartbeatCache implements IHeartbeatStatistic {
 
-  private boolean containsCache = false;
+  // Cached heartbeat samples
+  private final HeartbeatWindow window;
 
-  // Map<DataNodeId, HeartbeatWindow>
-  private final Map<Integer, HeartbeatWindow> windowMap;
+  // For guiding queries, the higher the score the higher the load
+  private volatile float loadScore;
```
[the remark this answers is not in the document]
CRZbulabula: At present, we have not enough time to design and implement more complex load balancing algorithms, but monitoring of CPU load, memory load and disk capacity will be added later. The higher these hardware indicators are, the higher the node load will be. This is why the current loadScore is positively correlated with the actual load rating.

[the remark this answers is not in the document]
neuyilan: > so I think if the follower's load is not high, then the lastSendTime of the follower will be higher, so the loadScore will be smaller. so I think in the implementation, the lower the score the higher the load.


Yeah, I know, I thought there was something wrong with your calculation of `loadScore,`  so I reconsidered my explanation and found that I misunderstood it. The above explanation, the higher the `lastSendTime`, the lower the load,  and the smaller the `loadScore`.  so the comment is correct, sorry for that.

neuyilan (approved):
LGTM

confignode/src/main/java/org/apache/iotdb/confignode/manager/load/LoadManager.java
```diff
@@ -179,6 +158,51 @@ public void run() {
...
+    // Discard heartbeat cache of the offline DataNodes
+    Set<Integer> onlineIds = new HashSet<>();
+    for (TDataNodeInfo dataNodeInfo : onlineDataNodes) {
+      onlineIds.add(dataNodeInfo.getLocation().getDataNodeId());
+    }
+    for (Integer dataNodeId : heartbeatCacheMap.keySet()) {
+      if (!onlineIds.contains(dataNodeId)) {
+        heartbeatCacheMap.remove(dataNodeId);
+      }
+    }
```
qiaojialin, line 191: The cache could be remove when removing datanode. We do not discard cache here.

confignode/src/main/java/org/apache/iotdb/confignode/manager/load/heartbeat/HeartbeatWindow.java
```diff
@@ -48,4 +48,10 @@ public void addHeartbeat(HeartbeatPackage newHeartbeat) {
       }
     }
   }
+
```
qiaojialin, line 51: reduce the window size to 100

confignode/src/main/java/org/apache/iotdb/confignode/manager/load/heartbeat/HeartbeatWindow.java
```diff
@@ -48,4 +48,10 @@ public void addHeartbeat(HeartbeatPackage newHeartbeat) {
...
+  public long getLastSendTime() {
```
qiaojialin, line 52: remove the HeartBeatWindow class, move List<HeartbeatPackage> to HeartbeatCache class

confignode/src/main/java/org/apache/iotdb/confignode/manager/load/LoadManager.java
CRZbulabula: Fixed.

confignode/src/main/java/org/apache/iotdb/confignode/manager/load/heartbeat/HeartbeatWindow.java
CRZbulabula: Fixed.

qiaojialin (approved):

PR #6287, opened 2022-06-14
[IOTDB-3491] Assign an unique id to each ConfigNode

CRZbulabula:
ConfigNode does not have an index, the output ConfigNode id of the show cluster command is arbitrary:
fcdfa6edbee95dc7c58ceb131b29818

Therefore I assign an unique id to each ConfigNode after the new ConfigNode is applied.

Start three ConfigNodes and then three Datanodes. The result of show cluster is as follows:
3efec31abc0e9f8b1ad05a52a9b7a75
  
Start ConfigNode and DataNode alternately, and the result of show cluster is as follows:
2bb34c9d9b743196e5a6387d14d328d


confignode/src/main/java/org/apache/iotdb/confignode/persistence/NodeInfo.java
```diff
@@ -242,6 +242,11 @@ public TSStatus updateConfigNodeList(ApplyConfigNodeReq applyConfigNodeReq) {
     TSStatus status = new TSStatus();
     configNodeInfoReadWriteLock.writeLock().lock();
     try {
+      if (nextNodeId.get() < applyConfigNodeReq.getConfigNodeLocation().getConfigNodeId()) {
+        // In this case, at least one ConfigNode is registered with the leader node,
+        // so the nextNodeID of the followers needs to be added
+        nextNodeId.getAndIncrement();
```
wangchao316, line 248: nextDataNodeId.set(applyConfigNodeReq.getConfigNodeLocation().getConfigNodeId());    ?

CRZbulabula: When there are multiple ConfigNode or DataNode registered concurrently, the applyReq with less id might arrive early.

CRZbulabula: update

qiaojialin (approved):

PR #6290, opened 2022-06-15
[IOTDB-3482] Add IT for UDF management in new Cluster

Plutooooooo:


integration-test/src/test/java/org/apache/iotdb/db/it/udf/IoTDBUDFManagementIT.java
```diff
@@ -16,77 +16,69 @@
...
-        CompressionType.UNCOMPRESSED,
-        null);
-    Class.forName(Config.JDBC_DRIVER_NAME);
+    enableSeqSpaceCompaction = ConfigFactory.getConfig().isEnableSeqSpaceCompaction();
+    enableUnseqSpaceCompaction = ConfigFactory.getConfig().isEnableUnseqSpaceCompaction();
+    enableCrossSpaceCompaction = ConfigFactory.getConfig().isEnableCrossSpaceCompaction();
+    ConfigFactory.getConfig().setEnableSeqSpaceCompaction(false);
+    ConfigFactory.getConfig().setEnableUnseqSpaceCompaction(false);
+    ConfigFactory.getConfig().setEnableCrossSpaceCompaction(false);
+    EnvFactory.getEnv().initBeforeClass();
```
JackieTien97, line 68:
```suggestion
    EnvFactory.getEnv().initBeforeTest();
```
If you mean that you need to do this setUp() method every time test case is running, you should call `initBeforeTest`.

integration-test/src/test/java/org/apache/iotdb/db/it/udf/IoTDBUDFManagementIT.java
```diff
@@ -16,77 +16,69 @@
...
   }
 
   @After
-  public void tearDown() throws Exception {
-    EnvironmentUtils.cleanEnv();
+  public void tearDown() {
+    EnvFactory.getEnv().cleanAfterClass();
```
JackieTien97, line 73:
```suggestion
    EnvFactory.getEnv().cleanAfterTest();
```

integration-test/src/test/java/org/apache/iotdb/db/it/udf/IoTDBUDFManagementIT.java
```diff
@@ -279,86 +253,58 @@ public void testDropFunction2() { // drop
...
-        assertTrue(throwable.getMessage().contains("Built-in function"));
+        assertTrue(
+            throwable.getMessage().contains("Built-in function ABS can not be deregistered"));
       }
-      statement.execute("INSERT INTO root.vehicle.d1(time, s1) VALUES(1, -10.0)");
-      ResultSet rs = statement.executeQuery("SELECT ABS(s1) FROM root.vehicle.d1");
-      Assert.assertTrue(rs.next());
-      Assert.assertEquals(1, rs.getLong(1));
-      Assert.assertEquals(10.0F, rs.getFloat(2), 0.00001);
-      Assert.assertFalse(rs.next());
```
JackieTien97, line 297: Why delete these assert

integration-test/src/test/java/org/apache/iotdb/db/it/udf/IoTDBUDFManagementIT.java
```diff
@@ -16,77 +16,69 @@
...
 
   @Test
   public void testCreateReflectShowDrop() {
-    try (Connection connection =
-            DriverManager.getConnection(
-                Config.IOTDB_URL_PREFIX + "127.0.0.1:6667/", "root", "root");
+    try (Connection connection = EnvFactory.getEnv().getConnection();
         Statement statement = connection.createStatement()) {
       statement.execute("create function udf as 'org.apache.iotdb.db.query.udf.example.Adder'");
       statement.execute("select udf(*, *) from root.vehicle");
```
JackieTien97, line 84: Why executing query sql here, not verifying the correctness?

JackieTien97 (changes requested):

integration-test/src/test/java/org/apache/iotdb/db/it/udf/IoTDBUDFManagementIT.java
Plutooooooo: CI was not stable when calling insert yesterday, these are added back now.

integration-test/src/test/java/org/apache/iotdb/db/it/udf/IoTDBUDFManagementIT.java
Plutooooooo: I think query here is to make sure that udf function itself is created correctly, So in this IT we don't need to check the result of this sql.

integration-test/src/test/java/org/apache/iotdb/db/it/udf/IoTDBUDFManagementIT.java
Plutooooooo: Fixed.

integration-test/src/test/java/org/apache/iotdb/db/it/udf/IoTDBUDFManagementIT.java
Plutooooooo: Fixed.

PR #6292, opened 2022-06-15
[IOTDB-3494] Fix TypeError in python session

THUMarkLau:
See [IOTDB-3494](https://issues.apache.org/jira/browse/IOTDB-3494).
The reason the this bug is that session does not format the string properly when printing logs using the python-logging library.


HTHou (approved):
LGTM

PR #6301, opened 2022-06-16
[IOTDB-3504] Fix inferTypes, return value and column label of IN/LIKE…

Wei-hao-Li:
1. Fix inferTypes of children expressions for nest expression in SELECT component;
2. Make IN/LIKE/REGEXP expression returns false instead of null  in SELECT component;
3. Alter column label for nest query combined with IN, like (s1 + s2) IN (1,2,3).


Plutooooooo (commented):
LGTM~

PR #6308, opened 2022-06-16
[IOTDB-3510] Read/Write Routing policy (Routing to DataNode with the lowest-loaded)

CRZbulabula:
**New feature:** Add RouteBalancer into LoadManager. The RouteBalancer constructs the IRouter to generate the current optimal real-time routing policy for read/write requests. IRouter based on loadScore has been implemented. See **LoadManager, RouteBalancer, IRoute and LoadScoreGreedyRouter** for more details.

**Code refactoring:** For PartitionTable queries in ConfigNode, the last level of mapping result is only to TConsensusGroupId. In order to correspond to the current DataNode's interfaces, ConfigManager will map the TConsensusGroupId to the TRegionReplicaSet. Since this


CRZbulabula:
I start a testing environment of 1C3D, and set replication_factor=3, then do the following test:

Insert some data:

image


Shutdown 6667 and query through 6668, success:

image


Restart 6667, shutdown 6668, and query through 6669, success but need retry for several times:

image


Restart 6668, shutdown 6669, and query through 6667, success but also need retry:

image

confignode/src/main/java/org/apache/iotdb/confignode/consensus/response/DataPartitionResp.java
```diff
@@ -49,28 +61,51 @@ public boolean isAllPartitionsExist() {
...
-    return dataPartition;
-  }
-
   /**
    * Convert DataPartitionDataSet to TDataPartitionResp
    *
    * @param resp TDataPartitionResp
    */
-  public void convertToRpcDataPartitionResp(TDataPartitionResp resp) {
+  public void convertToRpcDataPartitionResp(
```
xingtanzjr, line 69: Suggest make the `resp` as return value rather than a parameter, which makes this method not pure.

confignode/src/main/java/org/apache/iotdb/confignode/consensus/response/SchemaNodeManagementResp.java
```diff
@@ -19,19 +19,26 @@
...
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 public class SchemaNodeManagementResp implements DataSet {
 
   private TSStatus status;
 
-  private SchemaPartition schemaPartition;
+  // Map<StorageGroup, SchemaPartitionTable>
+  // TODO: Replace this map whit new SchemaPartition
```
xingtanzjr, line 40: typo `with`

confignode/src/main/java/org/apache/iotdb/confignode/manager/load/balancer/router/LoadScoreGreedyRouter.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Vector;
+import java.util.concurrent.ConcurrentHashMap;
+
+/** The LoadScoreGreedyRouter always pick the Replica with the lowest loadScore */
+public class LoadScoreGreedyRouter implements IRouter {
+
+  // Map<DataNodeId, loadScore>
+  private final Map<Integer, Float> loadScoreMap;
```
xingtanzjr, line 36: Suggest define a class such as `LoadScore` to replace the raw `Float` here. 

Even if we only consider the Float here, we may have more scores from different dimension to calculate the raw score of DataNode

confignode/src/main/java/org/apache/iotdb/confignode/consensus/response/DataPartitionResp.java
CRZbulabula: Fixed.

confignode/src/main/java/org/apache/iotdb/confignode/consensus/response/SchemaNodeManagementResp.java
CRZbulabula: Fixed.

confignode/src/main/java/org/apache/iotdb/confignode/manager/load/balancer/router/LoadScoreGreedyRouter.java
CRZbulabula: Good idea! But the LoadScoreGreedyRouter is only sorted based on the loadScore, then greedily select DataNode with the lowest loadScore, where using a Float should be sufficient. I think it would be more appropriate to create the LoadScore class later when we extending other function classes that implement the IRouter interface.

qiaojialin (approved):

PR #6314, opened 2022-06-16
Change the meaning of `Flush` on cluster mode

HTHou:
`Flush` command in cluster should flush data on all data nodes.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/mpp/plan/execution/config/FlushTask.java
+++ b/server/src/main/java/org/apache/iotdb/db/mpp/plan/execution/config/FlushTask.java
@@ -54,10 +54,10 @@ public class FlushTask implements IConfigTask {
       tFlushReq.setIsSeq(flushStatement.isSeq().toString());
     }
     IoTDBConfig config = IoTDBDescriptor.getInstance().getConfig();
-    if (flushStatement.isLocal()) {
-      tFlushReq.setDataNodeId(config.getDataNodeId());
-    } else {
+    if (flushStatement.isCluster()) {
       tFlushReq.setDataNodeId(-1);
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/apache/iotdb/db/mpp/plan/execution/config/FlushTask.java
```diff
@@ -54,10 +54,10 @@ public ListenableFuture<ConfigTaskResult> execute(IConfigTaskExecutor configTask
       tFlushReq.setIsSeq(flushStatement.isSeq().toString());
     }
     IoTDBConfig config = IoTDBDescriptor.getInstance().getConfig();
-    if (flushStatement.isLocal()) {
-      tFlushReq.setDataNodeId(config.getDataNodeId());
-    } else {
+    if (flushStatement.isCluster()) {
       tFlushReq.setDataNodeId(-1);
+    } else {
+      tFlushReq.setDataNodeId(config.getDataNodeId());
```
ericpai, line 60: If SQL is `FLUSH` in standalone mode, then `flushStatement.isCluster() == true` in your ASTVisitor modificaton. So what will happen here?

HTHou: Standalone mode won’t check it, only flush the data in StorageEngine.

ericpai: Got it.

ericpai (approved):

ericpai:
LGTM

qiaojialin (approved):

PR #6317, opened 2022-06-16
uncomment the data_block_manager_port and change default data region to StandAlone Consensus

qiaojialin:
uncomment the data_block_manager_port=8777
add final to some field
change default data region consensus to standalone
remove some unused field in ClusterScheduler
rename ConfigNodeConf to ConfigNodeConfig
rename DataNode startup process of prepareJoin() and join() to register() and active()


CRZbulabula (approved):
LGTM~

PR #6321, opened 2022-06-17
[To rel/0.13] make operation sync cover all protocols

qiaojialin:
Move operation sync related from TSServiceImpl to ServiceProvider to cover all protocols, such as RPC, MQTT, and Influxdb protocol.

Test sync MQTT insertions.

image


HTHou (approved):

CRZbulabula (approved):
LGTM.

PR #6326, opened 2022-06-17
Accelerate New Cluster IT

JackieTien97:


confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeDescriptor.java
```diff
@@ -129,6 +129,10 @@ private void loadProps() {
           properties.getProperty(
               "series_partition_executor_class", conf.getSeriesPartitionExecutorClass()));
 
+      conf.setConfigNodeConsensusProtocolClass(
+          properties.getProperty(
+              "config_node_consensus_protocol_class", conf.getConfigNodeConsensusProtocolClass()));
```
ericpai, line 134: It seems that this property is not in `iotdb-confignode.properties`, please add this property with comments and default values to it.

ericpai (changes requested):

ericpai (approved):

PR #6337, opened 2022-06-20
[IOTDB-3255] Turn off MLog and Implement snapshot while using Ratis

MarcosZyk:
## Description

Raft takes responsibility of wal and system recover logic, which overlaps some feature in schemaRegion. To avoid performance loss, this pr adapts schemaRegion impl to raft while using ratis.


OneSizeFitsQuorum (approved):
LGTM

PR #6338, opened 2022-06-20
[IOTDB-3436] Add aggregation IT without value filter

Alima777:


integration-test/src/test/java/org/apache/iotdb/db/it/aggregation/IoTDBAggregationDeleteIT.java
```diff
@@ -71,10 +69,8 @@ public void tearDown() throws Exception {
...
     try (Connection connection = EnvFactory.getEnv().getConnection();
         Statement statement = connection.createStatement()) {
-      boolean hasResultSet = statement.execute("select count(*) from root");
 
-      assertTrue(hasResultSet);
-      try (ResultSet resultSet = statement.getResultSet()) {
+      try (ResultSet resultSet = statement.executeQuery("select count(*) from root")) {
         while (resultSet.next()) {
           assertEquals("3", resultSet.getString(count("root.turbine.d1.s1")));
         }
```
irvine0109, line 76: If resultSet.next() returns false at the first time of this while loop, should this test fail?

Alima777: Nope...Ususally, it will test together with a `count` of resultset.

liuminghui233 (approved):

PR #6339, opened 2022-06-20
[IOTDB-3544] Packaging udf-api as dependency jar in build

ericpai:
See JIRA: https://issues.apache.org/jira/browse/IOTDB-3544

As we want to release udf-api as a dependency jar for users, it should be packaged during the build.


ericpai:
@SteveYurongSu @Plutooooooo PTAL

Plutooooooo (approved):
LGTM~

PR #6342, opened 2022-06-20
[IOTDB-3534] Small tsfile flushed by wal effective information ratio mechanism

HeimingZ:
The flush of small tsfile is caused by concurrency error between normal flush and flush caused by wal effective information ratio mechanism. To fix this, I will compare memTable after getting the writelock of DataRegion.


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/DataRegion.java
```diff
@@ -1372,19 +1374,25 @@ private void tryToUpdateInsertLastCache(InsertRowNode node, long latestFlushedTi
...
       TsFileProcessor tsFileProcessor;
       if (sequence) {
         tsFileProcessor = workSequenceTsFileProcessors.get(timeRangeId);
       } else {
         tsFileProcessor = workUnsequenceTsFileProcessors.get(timeRangeId);
       }
-      if (tsFileProcessor != null) {
+      boolean shouldSubmit =
+          tsFileProcessor != null && tsFileProcessor.getWorkMemTable() == memTable;
+      if (shouldSubmit) {
```
HTHou, line 1392: Add more commit to explain this part of code?

PR #6358, opened 2022-06-21
Optimize write and MPP framework

xingtanzjr:
## Description
1. Remove some duplicated log
2. Fix the issue that `getStatus()` will be queued for a long while (50-70ms)
3. Fix the issue that the thread is occupied when sleeping in QueryTerminator, which lead to the task in same thread-pool will be queued for a long while to waiting the thread being released.
4. The write operation is changed to Sync using client thread. (The testing is in progress and the async write won't be discard)

After this change, the write throughput in 1C1D is similar with new-standalone.


server/src/assembly/resources/conf/iotdb-engine.properties
```diff
@@ -1020,6 +1020,14 @@ timestamp_precision=ms
 # Datatype: int
 # datanode_schema_cache_size=10000
 
+# thread pool size for read operation in DataNode's coordinator.
+# Datatype: int
+# coordinator_read_executor_size=30
+
+# thread pool size for write operation in DataNode's coordinator.
+# Datatype: int
+# coordinator_read_executor_size=30
```
HTHou, line 1029:
```suggestion
# coordinator_write_executor_size=30
```

server/src/assembly/resources/conf/iotdb-engine.properties
```diff
@@ -1020,6 +1020,14 @@ timestamp_precision=ms
 # Datatype: int
 # datanode_schema_cache_size=10000
```
HTHou, line 1021:
```suggestion
# datanode_schema_cache_size=1000000
```

server/src/assembly/resources/conf/iotdb-engine.properties
xingtanzjr: Fixed

server/src/assembly/resources/conf/iotdb-engine.properties
xingtanzjr: Fixed

HTHou (approved):
LGTM

PR #6359, opened 2022-06-21
[IOTDB-3549] Fix delete data in new standalone

MarcosZyk:
### Cause

The implementation of delete data for cluster mode use the slotId to get dataPartition. Although this saves the duplicated device slotId calculation, the standalonePartitionFetcher doesn't support this.

### Solution

Use device path to get data partition in new standalone.

### Future work

Make the behavior of partition fetcher and configNode of standalone consistent with those of cluster


HTHou (approved):
LGTM

PR #6360, opened 2022-06-21
[IOTDB-3582] Fix client connections leak caused by delete storage group

cmlmakahts:
Use try-with-resources to return client

During executing test.sh, client threads did not increase rapidly.
image

image

_test.sh_ and _suite.sql_ are attached to https://issues.apache.org/jira/browse/IOTDB-3582

### DIFF
--- a/confignode/src/main/java/org/apache/iotdb/confignode/client/SyncDataNodeClientPool.java
+++ b/confignode/src/main/java/org/apache/iotdb/confignode/client/SyncDataNodeClientPool.java
@@ -41,7 +41,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
-/** Asynchronously send RPC requests to DataNodes. See mpp.thrift for more details. */
+/** Synchronously send RPC requests to DataNodes. See mpp.thrift for more details. */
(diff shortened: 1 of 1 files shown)


confignode/src/main/java/org/apache/iotdb/confignode/client/SyncDataNodeClientPool.java
```diff
@@ -57,10 +57,8 @@ private SyncDataNodeClientPool() {
 
   public TSStatus invalidatePartitionCache(
       TEndPoint endPoint, TInvalidateCacheReq invalidateCacheReq) {
-    SyncDataNodeInternalServiceClient client;
     TSStatus status;
-    try {
-      client = clientManager.borrowClient(endPoint);
+    try (SyncDataNodeInternalServiceClient client = clientManager.borrowClient(endPoint)){
```
wangchao316, line 61: ){ 
space.

cmlmakahts: Fixed

wangchao316 (approved):
LGTM

PR #6370, opened 2022-06-21
[IOTDB-3584] Snapshot unstable due to segment size

SzyWilliam:
When 400000 raft logs append, Ratis will ask statemachine to take a snapshot and purge the logs.
Currently, log segment size is 8MB, which means 400000 logs will add to 10000+ logs files, and purge them will take up to 10 seconds, which is longer enough to cause the current leader to step down.
Therefore, in this PR I change to segment size to 128 MB to reduce file numbers and corresponding purge time.


OneSizeFitsQuorum (approved):
LGTM

PR #6376, opened 2022-06-22
[IOTDB-3589] Fix dataRegion cannot recover from snapshot

THUMarkLau:
See [IOTDB-3589](https://issues.apache.org/jira/browse/IOTDB-3589).
Before this PR, data region takes snapshot without a directory structure, the info is stored in the filename, which is not elegant. This PR changes the way DataRegion creates snapshots which has a directory structure, and changes how DataRegion loads snapshot.


OneSizeFitsQuorum (approved):
LGTM

PR #6377, opened 2022-06-22
[IOTDB-3509] Read/Write Routing Policy (Routing to leader)

CRZbulabula:
This routing policy work as the following process:

1. ConfigNode-leader asks all DataNodes to judge its RegionGroups' leader in every 5s
2. ConfigNode-leader pick RegionGroup's leader as the first choice, then sort other Replicas by their loadScores for read/write requests.


confignode/src/main/java/org/apache/iotdb/confignode/manager/load/balancer/RouteBalancer.java
```diff
@@ -46,8 +48,13 @@ public Map<TConsensusGroupId, TRegionReplicaSet> genRealTimeRoutingPolicy(
   }
 
   private IRouter genRouter() {
-    // TODO: The Router should be configurable
-    return new LoadScoreGreedyRouter(getLoadManager().getAllLoadScores());
+    String policy = ConfigNodeDescriptor.getInstance().getConf().getRoutingPolicy();
+    if (policy.equals("leader")) {
```
wangchao316, line 52: extract “leader” to final static ....    other place could use together.

confignode/src/main/java/org/apache/iotdb/confignode/manager/load/heartbeat/RegionGroupCache.java
```diff
@@ -0,0 +1,47 @@
...
+  private long timestamp;
+
+  private final AtomicInteger leaderId;
+
+  public RegionGroupCache() {
+    this.leaderId = new AtomicInteger(-1);
+  }
+
+  @Override
+  public synchronized void updateLeader(long timestamp, int dataNodeId) {
```
wangchao316, line 36: Does leader of SchemaRegion, and data region persist?

CRZbulabula: I think there is no need to persist since they are dynamic status.

confignode/src/main/java/org/apache/iotdb/confignode/manager/load/balancer/RouteBalancer.java
CRZbulabula: Fixed~

confignode/src/main/java/org/apache/iotdb/confignode/manager/load/heartbeat/RegionGroupCache.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class RegionGroupCache implements IRegionGroupCache {
+
+  // TODO: This class might be split into SchemaRegionGroupCache and DataRegionGroupCache
+
+  private long timestamp;
+
+  private final AtomicInteger leaderId;
```
qiaojialin, line 29:
```suggestion
  private final AtomicInteger leaderDataNodeId;
```

CRZbulabula: fix

qiaojialin (approved):

PR #6382, opened 2022-06-22
[IOTDB-3558] Dump node stacks if IT test request timeout

ericpai:
JIRA: https://issues.apache.org/jira/browse/IOTDB-3558
For any test case failed by request timeout, a thread dump file will be output to `cluster-logs` dir of every node.
image

### DIFF
--- a/integration-test/src/main/java/org/apache/iotdb/it/env/AbstractEnv.java
+++ b/integration-test/src/main/java/org/apache/iotdb/it/env/AbstractEnv.java
@@ -28,19 +28,23 @@ import org.apache.iotdb.jdbc.Config;
 import org.apache.iotdb.jdbc.Constant;
 import org.apache.iotdb.jdbc.IoTDBConnection;
 
+import org.apache.commons.lang3.SystemUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.File;
+import java.io.IOException;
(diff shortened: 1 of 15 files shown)


integration-test/src/main/java/org/apache/iotdb/it/env/AbstractEnv.java
```diff
@@ -200,14 +193,53 @@ public void cleanAfterTest() {
...
+              .collect(Collectors.toList());
+      String cmd = getSearchAvailablePortCmd(requiredPorts);
+
+      try {
+        Process proc = Runtime.getRuntime().exec(cmd);
+        if (proc.waitFor() == 1 && lockFile.createNewFile()) {
+          return requiredPorts.stream().mapToInt(Integer::intValue).toArray();
+        }
+      } catch (IOException | InterruptedException ex) {
+        // ignore
```
wangchao316, line 217: Is  add "Thread.currentThread().interrupt();"   ?

wangchao316 (commented):
Good

integration-test/src/main/java/org/apache/iotdb/it/env/AbstractEnv.java
ericpai: Good catch! It has been fixed.

irvine0109 (approved):
LGTM

PR #6384, opened 2022-06-22
[IOTDB-3604] Refactor the time zone setter by using the setClientInfo method of JDBC

irvine0109:


jdbc/src/main/java/org/apache/iotdb/jdbc/Utils.java
```diff
@@ -18,13 +18,20 @@
...
+import java.time.ZoneId;
+import java.time.zone.ZoneRulesException;
 import java.util.Properties;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 /** Utils to convert between thrift format and TsFile format. */
 public class Utils {
 
+  private static final Logger logger = LoggerFactory.getLogger(Utils.class);
```
ericpai, line 34: Don't add logger in SDK, as the user may be confused by the output, and there's no way to silent the log. I think it should throw a SQLException directly.

ericpai (changes requested):

jdbc/src/main/java/org/apache/iotdb/jdbc/Utils.java
irvine0109: done

ericpai (approved):

HTHou (approved):
LGTM
Spotless issue...

PR #6388, opened 2022-06-23
[IOTDB-3513] Delete tsfiles when recovering data region

HeimingZ:
Delete tsfiles when recovering data region.


OneSizeFitsQuorum (approved):
LGTM

PR #6390, opened 2022-06-23
delete file before recovery in ratis mode

MarcosZyk:
## Description

When IoTDB been shutdown, the file in schemaRegion dir should be deleted during recovery, otherwise it will affect the raftlog redo.


OneSizeFitsQuorum (approved):
LGTM

PR #6391, opened 2022-06-23
[IOTDB-3622] Use hardlink to decrease the copied file size in cluster test

ericpai:
See JIRA: https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-3622?filter=myopenissues


integration-test/src/main/java/org/apache/iotdb/it/env/AbstractNodeWrapper.java
```diff
@@ -79,14 +83,32 @@ public AbstractNodeWrapper(String testClassName, String testMethodName, int[] po
   @Override
   public void createDir() {
     // Copy templateNodePath to nodePath
+    String destPath = getNodePath();
     try {
-      File nodeDir = new File(getNodePath());
+      File nodeDir = new File(destPath);
       try {
         FileUtils.forceDelete(nodeDir);
```
irvine0109, line 90: Use PathUtils.deleteDirectory() instead of FileUtils.forceDelete()？

irvine0109 (approved):
LGTM

HTHou:
Hi, does the symlink work on Windows?

ericpai:
> Hi, does the symlink work on Windows?

Although I haven't tested in my own computer yet, the github action has already runned without any filesytem errors. If the symlink doesn't work in Windows, all the tasks will fail. Besides, the symlink is created through maven plugin, not native scripts, which should be platform independent.

ericpai:
maven-antrun-plugin v3.1.0 embeded apache ant v1.10.12. It implements making symlink via java nio API, which is definitely platform independent. https://github.com/apache/ant/blob/rel/1.10.12/src/main/org/apache/tools/ant/taskdefs/optional/unix/Symlink.java#L451

ericpai:
As creating symbolic link in Windows needs privileges, in order to make it easy to use, I use hard link instead.

HTHou (approved):

PR #6408, opened 2022-06-23
Fix IoTDBCreateTimeseriesIT

HTHou:


integration-test/src/test/java/org/apache/iotdb/db/it/IoTDBCreateTimeseriesIT.java
```diff
@@ -73,36 +63,36 @@ public static void tearDown() throws Exception {
...
-    try {
+    try (Connection connection = EnvFactory.getEnv().getConnection();
+        Statement statement = connection.createStatement()) {
+      statement.execute(String.format("SET storage group TO %s", storageGroup));
       statement.execute(
           String.format(
               "create timeseries %s with datatype=INT64, encoding=PLAIN, compression=SNAPPY",
               storageGroup));
+
     } catch (Exception ignored) {
```
ericpai, line 74: It seems that this exception is ignored, is it as our expected?

HTHou: The statement is trying to create timeseries with the same name of the storage group, the exception is expected.

ericpai (approved):

PR #6410, opened 2022-06-24
[IOTDB-3640] onlinecoonfignode does not contains location of leader a…

wangchao316:
OnlineConfigNode does not contains Location of leader after AddConfigNode


qiaojialin:
Fixed in https://github.com/apache/iotdb/pull/6426

PR #6412, opened 2022-06-24
[IOTDB-3635][IOTDB-3583] Increase the stability of the getLeader interface

CRZbulabula:
ConsensusManager's getLeader interface returns the result for the consensusLayer. However, the consensusLayer's getLeader interface may return null when the consensusLayer is on leader election or when ConfigNode isn't added to any consensusGroup. Therefore, this PR optimizes the stability logic of the ConsensusManager's getLeader interface.

### DIFF
--- a/confignode/src/main/java/org/apache/iotdb/confignode/manager/ConsensusManager.java
+++ b/confignode/src/main/java/org/apache/iotdb/confignode/manager/ConsensusManager.java
@@ -165,7 +165,7 @@ public class ConsensusManager {
 
   /** @return ConfigNode-leader's location if leader exists, null otherwise. */
   public TConfigNodeLocation getLeader() {
-    for (int retry = 0; retry < 5; retry++) {
+    for (int retry = 0; retry < 50; retry++) {
       Peer leaderPeer = consensusImpl.getLeader(consensusGroupId);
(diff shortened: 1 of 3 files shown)


confignode/src/main/java/org/apache/iotdb/confignode/manager/ConsensusManager.java
```diff
@@ -158,15 +163,21 @@ public boolean isLeader() {
...
+  public TConfigNodeLocation getLeader() {
+    for (int retry = 0; retry < 5; retry++) {
+      Peer leaderPeer = consensusImpl.getLeader(consensusGroupId);
+      List<TConfigNodeLocation> onlineConfigNodes = getNodeManager().getOnlineConfigNodes();
+      TConfigNodeLocation leaderLocation =
+          onlineConfigNodes.stream()
+              .filter(leader -> leader.getConsensusEndPoint().equals(leaderPeer.getEndpoint()))
+              .findFirst()
+              .orElse(null);
+      if (leaderLocation != null) {
```
wangchao316, line 176: add sleep

CRZbulabula: Fixed~

qiaojialin (approved):

PR #6417, opened 2022-06-24
[To rel/0.13]Deactivation and Privilege for Schema Template

bigreybear:
## Backgroud
Since feature of schema template is released with Apache IoTDB v0.13, several issues about it have been put forward. Before this PR, if a user needs to unset a template which has been activated from one node (means data has been inserted into the measuremnt inside the template),   the user has to delete the whole storage group first. This is a pretty weird practice as it is supposed to ban the unset of activated template in the very beginning. What's more, only `ROOT` user can utilize template as there is no privilege entry for template. This PR tackles these issues at once.

## Deactivation of Schema Template
When a node or its ancestor has been set a template on, it can be set using template by `create timeseries of schema template` or data insertion. Once a node is using template, it or its ancestor on which has been set template cannot unset the template until a deactivation issued. And the deactivation mentioned above, can be issued by statement `deactivate schema tempalte <template> from <path> ` or interface `deactivateTemplate` in `Session`.

## Privilege of Schema Template
Two more privilege entry has been added to handle authority of schema template. `UPDATE_TEMPLATE` is an entry independent of path and stands for create, drop, append and prune of any schema template. `APPLY_TEMPLATE` is a dependent one which stands for set, unset and activate of schema template on the authorized paths. As long as the deactivation may delete the corresponding time series data points, it is authorized by the entry `DELETE_TIMESERIES`.


RYH61 (approved):
LGTM~

SpriCoder (approved):
LGTM~

liuminghui233 (approved):
LGTM!

MarcosZyk (approved):

PR #6428, opened 2022-06-24
[IOTDB-3260] Fix npe while concurrent delete storage group

MarcosZyk:
### NPE Cause

When executing predelete and delete sg on partitionInfo, the target storageGroupPartitionTable will be taken from a concurrentHashMap. If the deletion is executed concurrently, the taken table may be null since there may be some thread else delete the table, which may results in NPE if there's no check.

### Solution
Before invoke any interface of storageGroupPartitionTable taken from the map, check whether the table is null.


SpriCoder (approved):
LGTM~

PR #6441, opened 2022-06-25
upgrade spotless to fix jvmargs issue

tisonkun:
Since https://github.com/diffplug/spotless/issues/834 resolved, upgrade spotless version for suppressing jvmargs issue on JDK17+.


tisonkun:
cc @jixuan1989

HTHou (approved):
Yes, it works. Thanks for your contribution!

PR #6478, opened 2022-06-28
[IOTDB-3636] Fix delete non existing sg error msg

MarcosZyk:
## Description

To stick with 0.13. While trying to delete non existing storage group, in other words the input path patterns doesn't match any existing storage group, message like "path not exists" should be returned.

Result 
image


HTHou:
Hi, does new standalone have the same problem?

MarcosZyk:
> Hi, does new standalone have the same problem?

I've fixed the same problem in new standalone version. PTAL.

HTHou (approved):

PR #6483, opened 2022-06-28
[IOTDB-3173] Support Between expression

Wei-hao-Li:


server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/ExpressionAnalyzer.java
```diff
@@ -590,6 +723,10 @@ public static Pair<Filter, Boolean> transformToGlobalTimeFilter(Expression predi
         return new Pair<>(timeInRightFilter, false);
       }
       return new Pair<>(null, true);
+    } else if (predicate instanceof LikeExpression || predicate instanceof RegularExpression) {
+      return new Pair<>(null, true);
+    } else if (predicate instanceof BetweenExpression) {
+      return new Pair<>(null, true);
```
JackieTien97, line 729: Do we now support `where time between 1 and 10` syntax? If so, here we should return the corresponding time filter.

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/ExpressionAnalyzer.java
Wei-hao-Li: Yes, we support. Is corresponding time filter used to filter quickly? We can get the same result even if treat it as value filter.

tsfile/src/main/java/org/apache/iotdb/tsfile/read/filter/operator/Between.java
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  @Override
+  public boolean satisfy(Statistics statistics) {
+    if (filterType == FilterType.TIME_FILTER) {
+      long time1 = (Long) value1, time2 = (Long) value2;
+      if (not) {
+        return statistics.getStartTime() < time1 || statistics.getEndTime() > time2;
+      } else {
+        return statistics.getEndTime() >= time1 || statistics.getStartTime() <= time2;
```
JackieTien97, line 87:
```suggestion
        return statistics.getEndTime() >= time1 && statistics.getStartTime() <= time2;
```

tsfile/src/main/java/org/apache/iotdb/tsfile/read/filter/operator/Between.java
```diff
@@ -0,0 +1,141 @@
...
+    } else {
+      if (statistics.getType() == TSDataType.TEXT || statistics.getType() == TSDataType.BOOLEAN) {
+        return true;
+      }
+      if (not) {
+        return ((T) statistics.getMinValue()).compareTo(value1) < 0
+            || ((T) statistics.getMaxValue()).compareTo(value2) > 0;
+      } else {
+        return ((T) statistics.getMaxValue()).compareTo(value1) >= 0
+            || ((T) statistics.getMinValue()).compareTo(value2) <= 0;
```
JackieTien97, line 98:
```suggestion
        return ((T) statistics.getMaxValue()).compareTo(value1) >= 0
            && ((T) statistics.getMinValue()).compareTo(value2) <= 0;
```

JackieTien97 (changes requested):
CI is failed, please correct it.

JackieTien97 (approved):

PR #6484, opened 2022-06-28
Fix the issues in MultiLeader log dispatcher

xingtanzjr:
## Description
1. Fix the bug in WALNode that iterator won't be set to null when invoking skip.
2. Use iterator to replace fetch WALEntry one by one


qiaojialin (commented):
Approved

PR #6492, opened 2022-06-28
[IOTDB-3672] Fix reloading online config nodes from configuration

MrQuansy:
This pr fix a bug of reloading online config node list from the configuration, and rename the configuration of configNodeList to targetConfigNodeList which fits for its usage.
System.properties can persist the full online config node list.
image


HTHou (approved):

CRZbulabula (approved):
LGTM

PR #6530, opened 2022-06-30
[IOTDB-3696] Transport Schema Related IT

MarcosZyk:
## Description

This pr mainly transport schema related IT to new integration-test. Meanwhile, to ensure the IT won't fail, some bugs have been fixed in this PR as well.

### Bugs and Modification

1. Fix show sg and count sg bug in configNode
2. Fix delete data bug while split plan node by partition
3. Fix show timeseries result format bug
4. Fix show latest timeseries limit offset bug
5. Fix count devices and count timeseries memory source visitor to return 0 rather than empty
6. Fix queue destroyed bug while executing count nodes and show latest timeseries. The ```isBlocked``` method in related operator has been rewritten
7. Fix show child node result set bug
8. Support schema node management in new standalone mode
9. Fix new standalone exception bubbling bug to return correct error msg
10. Support execution TSStatus bubbling for MPP


jun0315 (approved):
LGTM in influxdb-protocol module

SpriCoder (approved):
LGTM for sg related module.

server/src/main/java/org/apache/iotdb/db/mpp/plan/scheduler/FragInstanceDispatchResult.java
```diff
@@ -19,14 +19,27 @@
 
 package org.apache.iotdb.db.mpp.plan.scheduler;
 
+import org.apache.iotdb.common.rpc.thrift.TSStatus;
+
 public class FragInstanceDispatchResult {
   private boolean successful;
 
+  private TSStatus failureStatus;
```
xingtanzjr, line 27: Maybe we can add one more parameter such as `message` and construct it when dispatch failed.

The error message can be fetched from remote response or local dispatch result.

And furthermore, we can change the `failureStatus` to `status` and remove the original param `successful` because it seems be duplicated.

MarcosZyk: The message can be obtained from the failureStatus.

xingtanzjr (approved):

PR #6566, opened 2022-07-02
[IOTDB-3697]  [IOTDB-3141] Add show version function

RYH61:
图片

PR #6571, opened 2022-07-03
[IOTDB-3141] Support show version

SpriCoder:
See more details: https://issues.apache.org/jira/browse/IOTDB-3141

Test on 1C1D:
image


SpriCoder:
Duplicate with #6566

Issue #6551, opened 2022-07-01
内存表刷新报数组下标越界错，iotdbserver进入只读模式


zhouzhiy:
报错日志截图如下
![1656642207424_1656641353522_CA91DBB2-D662-40c5-BCAA-55FBBE27FEE1](https://user-images.githubusercontent.com/52025735/176814065-98854be7-5db8-4737-8b1e-98f24d08052a.png)

jixuan1989:
issue #6551

jixuan1989:
@Cpaulyz  @qiaojialin , has is been fixed in both v0.13 and v0.14?

Cpaulyz:
> [To rel/0.13][IOTDB-3730][ISSUE-6551] ArrayIndexOutOfBounds when flushing a memtable #6576

Yes. It has been fixed in 0.12 #6572, 0.13 #6576 and master #6577(0.14).

PR #6576, opened 2022-07-04
[To rel/0.13][IOTDB-3730][ISSUE-6551] ArrayIndexOutOfBounds when flushing a memtable

Cpaulyz:
The problem is caused by a loss of precision in converting integers to floating point numbers


jixuan1989:
is this related to #6551 ?
if so, pls review #6558 .

qiaojialin (approved):

PR #6581, opened 2022-07-04
[IOTDB-3717] Add metrics for thrift connections and worker threads

xinzhongtianxia:
Add metrics for current number of thrift connections and worker threads, which will provide us with an insight into the state of system entrance.
They're important both for problem detection and thrift configurations optimization.


SpriCoder:
LGTM~

SpriCoder (approved):

PR #6592, opened 2022-07-05
[IOTDB-3750] Fix NPE caused by SchemaQueryMergeOperator#next returning null

xinzhongtianxia:
see https://issues.apache.org/jira/browse/IOTDB-3750

PR #6598, opened 2022-07-06
[IOTDB-3550]Fix show child nodes

MarcosZyk:
## Description

See IOTDB-3550

The duplicate result has been fixed in #6530.

Besides, there is still NPE problem, which is fixed in this pr.


yifuzhou (commented):
Hi, I met this issue as well, I think 'show child paths' still have problem. The root cause for NPE is in this PR #6592

MarcosZyk:
> Hi, I met this issue as well, I think 'show child paths' still have problem. The root cause for NPE is in this PR #6592

OK, I will have a check.

PR #6604, opened 2022-07-06
Add test stats in cluster IT for future use

ericpai:


integration-test/src/main/java/org/apache/iotdb/it/env/IoTDBTestRunner.java
```diff
@@ -26,24 +27,60 @@
...
 
   public IoTDBTestRunner(Class<?> testClass) throws InitializationError {
     super(testClass);
   }
 
+  @Override
+  public void run(RunNotifier notifier) {
+    super.run(notifier);
+    Collections.sort(testStats);
+    System.out.println("==== Top 30 slowest cases ====");
```
irvine0109, line 47: Use logger.info()？

irvine0109 (approved):

integration-test/src/main/java/org/apache/iotdb/it/env/IoTDBTestRunner.java
ericpai: Fixed.

PR #6610, opened 2022-07-07
[IOTDB-3767] Add getPartitionTable and getOrCreatePartitionTable interfaces

CRZbulabula:
We've created SchemaPartitionTable and DataPartitionTable whose last level points to the TConsensusGroupId rather than a specific TRegionReplicaSet, in order to reduce the memory overhead for PartitionTable in both ConfigNode and DataNode. Therefore we should add getPartitionTable and getOrCreatePartitionTable interfaces to correspond these updates. And these interfaces will eventually take the place of getPartition and getOrCreatePartition.


SpriCoder (approved):
LGTM~

xingtanzjr (approved):

PR #6616, opened 2022-07-07
[IOTDB-3770]Fix rename bug in TsFileResource

choubenson:
**Description**
Fail to rename the mods file at the second time, and compaction mods file does not rename.

**Solution**
Regardless of whether the `mods` attribute in `TsFileResource` is null or not, rename the mods file. Also rename the compaction mods file generated by deletion during compaction.


HeimingZ:
LGTM

PR #6617, opened 2022-07-07
[IOTDB-3734] Set safely deleted search index directly in multi-leader

HeimingZ:
Set safely deleted search index directly in multi-leader to delete outdated wal files.


SpriCoder (approved):
LGTM~

xingtanzjr (approved):

PR #6619, opened 2022-07-07
for the Chinese version of User Guide，delete an extra symbol

wanghui42:
I have deleted a redundant symbol in the Chinese version. Copying will report an error if not removed, and the English version is correct without this symbol.


HTHou (approved):
LGTM

PR #6624, opened 2022-07-08
[IOTDB-3352][IOTDB-3782] Fix the cache problem in PartitionCache and Optimize.

SpriCoder:
In this PR, fix partitionCache and optimize the performance.
1. use consensusGroupId -> RegionReplicaSet cache to reduce the communication overhead between datanode and confignode, and reduce the memory size of partitionCache.
2. update schemaPartitionCache: storageGroup -> SchemaPartitionTable 
3. update dataPartitionCache: storageGroup -> DataPartitionTable
4. finish replicaSet cache.


SpriCoder:
Test on 3C3D: 
image

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
```diff
@@ -0,0 +1,710 @@
...
+
+  // region storage group cache
+
+  /** get storage group of device, return null if not cache */
+  public String getStorageGroup(String devicePath) {
+    synchronized (storageGroupCache) {
+      for (String storageGroup : storageGroupCache) {
+        if (PathUtils.isStartWith(devicePath, storageGroup)) {
+          return storageGroup;
+        }
```
CRZbulabula, line 128: In the case of we have StorageGroup "root.sg" and "root.sg1", could the device "root.sg.d1" will still get the correct StorageGroup by PathUtils.isStartWith()?

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
```diff
@@ -0,0 +1,710 @@
...
+      }
+    }
+    return null;
+  }
+
+  /** get device to storage group map */
+  public boolean getDeviceToStorageGroupMap(
```
CRZbulabula, line 135: We'd better add function annotation to explain what these input parameters for and what this interface do since this is a **public** interface~

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
```diff
@@ -0,0 +1,710 @@
...
+      storageGroupCacheLock.readLock().unlock();
+    }
+  }
+
+  /** get storage group to device map with failed devices */
+  public boolean getStorageGroupToDeviceMap(
+      List<String> devicePaths,
+      Map<String, List<String>> storageGroupToDeviceMap,
+      List<String> failedDevices,
+      boolean failFast) {
```
CRZbulabula, line 174: We'd better add function annotation to explain what these input parameters for and what this interface do since this is a **public** interface~

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
```diff
@@ -0,0 +1,710 @@
...
+      CacheMetricsRecorder.record(result, STORAGE_GROUP_CACHE_NAME);
+      return result;
+    } finally {
+      storageGroupCacheLock.readLock().unlock();
+    }
+  }
+
+  /** get storage group to device map in three trys */
+  public Map<String, List<String>> getStorageGroupToDevice(
+      List<String> devicePaths, boolean isAutoCreate) {
```
CRZbulabula, line 208: I suppose that this interface provides the service of _get StorageGroup to Device_. We need to extract the _getStorageGroupToDeviceMap_ thus we can facilitate the retry logic inside this interface. Therefore, I wonder that would it be better if we make the _getStorageGroupToDeviceMap_ **private**?

By the way, we need a more detailed function annotation, which including the definition of every corresponding parameters, if my hypothesize is correct~

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
```diff
@@ -0,0 +1,710 @@
...
+        throw new StatementAnalyzeException(
+            "An error occurred when executing getStorageGroupToDevice():" + e.getMessage());
+      }
+    }
+    return storageGroupToDeviceMap;
+  }
+
+  /** get storage group to device map in three trys */
+  public Map<String, String> getDeviceToStorageGroup(
+      List<String> devicePaths, boolean isAutoCreate) {
```
CRZbulabula, line 255: The same as above

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
```diff
@@ -0,0 +1,710 @@
...
+        TSetStorageGroupReq req = new TSetStorageGroupReq(storageGroupSchema);
+        TSStatus tsStatus = client.setStorageGroup(req);
+        if (TSStatusCode.SUCCESS_STATUS.getStatusCode() == tsStatus.getCode()) {
+          successFullyCreatedStorageGroup.add(storageGroupName);
+        } else {
+          updateStorageCache(successFullyCreatedStorageGroup);
+          throw new RuntimeException(new IoTDBException(tsStatus.message, tsStatus.code));
+        }
+      }
+      updateStorageCache(storageGroupNamesNeedCreated);
```
CRZbulabula, line 333: Would it be better if we log out all the StorageGroups that were created successfully or failed?

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
```diff
@@ -0,0 +1,710 @@
...
+        throw new StatementAnalyzeException(
+            "An error occurred when executing getRegionReplicaSet():" + e.getMessage());
+      }
+    }
+    regionReplicaSet = groupIdToReplicaSetMap.get(consensusGroupId);
+    if (null == regionReplicaSet) {
+      throw new RuntimeException(
+          "ReplicaSet of consensus group[id= " + consensusGroupId + "] is null.");
+    }
+    return regionReplicaSet;
```
CRZbulabula, line 396: The RuntimeException might cause a NPE in upper layer. We should notify this in the function annotation in order to reduce the communication costs that might be inccured later.

CRZbulabula (commented):
PTAL~

confignode/src/test/resources/confignode1conf/iotdb-confignode.properties
```diff
@@ -20,7 +20,7 @@
 internal_address=0.0.0.0
 internal_port=22277
 consensus_port=22278
-target_config_nodes=0.0.0.0:22277
+target_config_nodes=127.0.0.1:22277
```
xingtanzjr, line 23: Why do we need this change ?

SpriCoder: fixed~

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
SpriCoder: Yes, it will

image

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
SpriCoder: Get it ~

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
SpriCoder: Get it ~

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
SpriCoder: Get it ~

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
SpriCoder: Get it ~

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
SpriCoder: Get it ~

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
xingtanzjr: Yes. And on the other hand, I think we need to avoid change the parameter of the function, even though it is by design to make the parameter as a returned result.

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
```diff
@@ -0,0 +1,710 @@
...
+    this.partitionExecutor =
+        SeriesPartitionExecutor.getSeriesPartitionExecutor(
+            this.seriesSlotExecutorName, this.seriesPartitionSlotNum);
+  }
+
+  // region storage group cache
+
+  /** get storage group of device, return null if not cache */
+  public String getStorageGroup(String devicePath) {
+    synchronized (storageGroupCache) {
```
xingtanzjr, line 124: Shall we add the read-write lock here ?

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
```diff
@@ -0,0 +1,710 @@
...
+      }
+      logger.debug("[{} Cache] hit when search device {}", STORAGE_GROUP_CACHE_NAME, devicePaths);
+      CacheMetricsRecorder.record(result, STORAGE_GROUP_CACHE_NAME);
+      return result;
+    } finally {
+      storageGroupCacheLock.readLock().unlock();
+    }
+  }
+
+  /** get storage group to device map in three trys */
```
xingtanzjr: Wrong comments

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
```diff
@@ -0,0 +1,710 @@
...
+        return storageGroupToDeviceMap;
+      }
+    }
+    // first try to hit storage group in fast-fail way
+    boolean firstTryResult =
+        getStorageGroupToDeviceMap(devicePaths, storageGroupToDeviceMap, new ArrayList<>(), true);
+    if (!firstTryResult) {
+      try {
+        // when local cache not have, then try to fetch all storage group from config node
+        getAllStorageGroupFromConfigNode();
```
xingtanzjr, line 161: Suggest to rename this method to some name such as `fetchStorageGroupAndUpdateCache()`

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
```diff
@@ -0,0 +1,710 @@
...
+        if (!secondTryResult && isAutoCreate) {
+          // try to auto create storage group of failed device
+          Set<String> storageGroupNamesNeedCreated = new HashSet<>();
+          for (String devicePath : failedDevices) {
+            PartialPath storageGroupNameNeedCreated =
+                MetaUtils.getStorageGroupPathByLevel(
+                    new PartialPath(devicePath), config.getDefaultStorageGroupLevel());
+            storageGroupNamesNeedCreated.add(storageGroupNameNeedCreated.getFullPath());
+          }
+          createNotExistedStorageGroup(storageGroupNamesNeedCreated);
```
xingtanzjr, line 175: Shall we consider the concurrent creation issue here ?

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
```diff
@@ -0,0 +1,710 @@
...
+      logger.debug("[{} Cache] hit when search device {}", STORAGE_GROUP_CACHE_NAME, devicePaths);
+      CacheMetricsRecorder.record(result, STORAGE_GROUP_CACHE_NAME);
+      return result;
+    } finally {
+      storageGroupCacheLock.readLock().unlock();
+    }
+  }
+
+  /** get storage group to device map in three trys */
+  public Map<String, List<String>> getStorageGroupToDevice(
```
xingtanzjr, line 146: It seems that the operation for this method is similar with `getToDeviceStorageGroup` and the only difference is the data structure of result. Maybe we can consider to remove the duplicated codes and generate the result from `getToDeviceStorageGroup` ?

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
```diff
@@ -0,0 +1,710 @@
...
+          == TSStatusCode.SUCCESS_STATUS.getStatusCode()) {
+        Set<String> storageGroupNames = storageGroupSchemaResp.getStorageGroupSchemaMap().keySet();
+        // update all storage group into cache
+        updateStorageCache(storageGroupNames);
+      }
+    }
+  }
+
+  /** create not existed storage group and update storage group cache */
+  private void createNotExistedStorageGroup(Set<String> storageGroupNamesNeedCreated)
```
xingtanzjr, line 261: Suggest to rename this method to `createStorageGroupAndUpdateCache()`

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
```diff
@@ -0,0 +1,757 @@
...
+      for (String storageGroupName : storageGroupNames) {
+        storageGroupCache.remove(storageGroupName);
+      }
+    } finally {
+      storageGroupCacheLock.writeLock().unlock();
+    }
+  }
+
+  /** invalid all storage group cache */
+  public void invalidAllStorageGroupCache() {
```
xingtanzjr, line 407: Suggest to rename to `invalidStorageGroupCache()`

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
```diff
@@ -0,0 +1,757 @@
...
+      storageGroupCacheLock.writeLock().unlock();
+    }
+  }
+
+  /**
+   * invalid storage group cache
+   *
+   * @param storageGroupNames the storage groups that need to invalid
+   */
+  public void invalidStorageGroupCache(List<String> storageGroupNames) {
```
xingtanzjr, line 395: suggest to rename to `removeFromStorageGroupCache()`

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
```diff
@@ -0,0 +1,757 @@
...
+    }
+  }
+
+  // endregion
+
+  // region replicaSet cache
+  /** get regionReplicaSet from consensusGroupId */
+  public TRegionReplicaSet getRegionReplicaSet(TConsensusGroupId consensusGroupId) {
+    TRegionReplicaSet regionReplicaSet;
+    if (!groupIdToReplicaSetMap.containsKey(consensusGroupId)) {
```
xingtanzjr, line 422: Need to handle concurrent issue here. The sample code should be

```
if (! xxx) {
   lock(cache) {
     if (! xxx) {
        fetchGroupIdToReplicaSet
     }
   }
}
```

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
```diff
@@ -0,0 +1,757 @@
...
+    // if timestamp is greater than latestUpdateTime, then update
+    if (result) {
+      groupIdToReplicaSetMap.clear();
+      groupIdToReplicaSetMap.putAll(map);
+    }
+    return result;
+  }
+
+  /** invalid all replica Cache */
+  public void invalidReplicaCache() {
```
xingtanzjr, line 459: Suggest to rename to `invalidReplicaSetCache()`

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
SpriCoder: this function will change to **private** because there are no need to get just one devicePath outside PartitionCache. Besides, getStorageGroupToDevice and getDeviceToStorageGroup are the only two **public** functions

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
SpriCoder: Get it.

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
SpriCoder: Get it.

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
SpriCoder: Get it

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
SpriCoder: get it

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
SpriCoder: Get it

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
SpriCoder: Get it.

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
SpriCoder: Get it

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
```diff
@@ -0,0 +1,757 @@
...
+    try {
+      for (Map.Entry<String, Map<TSeriesPartitionSlot, TConsensusGroupId>> entry1 :
+          schemaPartitionTable.entrySet()) {
+        String storageGroup = entry1.getKey();
+        SchemaPartitionTable result = schemaPartitionCache.getIfPresent(storageGroup);
+        if (null == result) {
+          result = new SchemaPartitionTable();
+          schemaPartitionCache.put(storageGroup, result);
+        }
+        Map<TSeriesPartitionSlot, TConsensusGroupId> result2 = result.getSchemaPartitionMap();
```
xingtanzjr, line 560: parameter‘s name should be more specific

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
```diff
@@ -0,0 +1,757 @@
...
+      if (devicePath.contains("*")) {
+        return deviceToStorageGroupMap;
+      }
+    }
+    // first try to hit storage group in fast-fail way
+    boolean firstTryResult =
+        getDeviceToStorageGroupMap(devicePaths, deviceToStorageGroupMap, new ArrayList<>(), true);
+    if (!firstTryResult) {
+      try {
+        // when local cache not have, then try to fetch all storage group from config node
```
xingtanzjr, line 196: have should be hit... and the grammar.....

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
```diff
@@ -0,0 +1,757 @@
...
+        getAllStorageGroupFromConfigNode();
```
xingtanzjr: Let's consider an issue here. That is, how to avoid multi-concurrent request here ?

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
SpriCoder: Get, I will refact this part

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
SpriCoder: Done~

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
SpriCoder: Fixed~

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
SpriCoder: Get it~

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/PartitionCache.java
SpriCoder: Fixed~

xingtanzjr (approved):

PR #6644, opened 2022-07-12
[IOTDB-3796] Throw exception while using without null clause in MPP

liuminghui233:
see [IOTDB-3796](https://issues.apache.org/jira/browse/IOTDB-3796).


liuminghui233:
Before:
2996e137697ab1a43cdfedd63249ef5

After:
31db4cb034d8f9f124f309f9810c7d6

JackieTien97 (approved):

PR #6647, opened 2022-07-12
Add more log to the cancellation operation of FragmentInstanceStateTracker

xingtanzjr:
## Description
During our test, we found that sometimes the FragmentInstanceStateTracker may won't be cancelled successfully after the QueryExecution finishing. And it would print lots of logs such as `State is NO_SUCH_INSTANCE`

We investigated this issue and found that the `cancel()` of FixedRateScheduleTask may failed due to some unknown reason. We have some replacement implementation here but we want to find the root cause of the cancel failure.

So we add more logs here currently and will investigate more once it appears next time.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/mpp/plan/scheduler/FixedRateFragInsStateTracker.java
+++ b/server/src/main/java/org/apache/iotdb/db/mpp/plan/scheduler/FixedRateFragInsStateTracker.java
@@ -43,9 +43,13 @@ public class FixedRateFragInsStateTracker extends AbstractFragInsStateTracker {
 
   private static final Logger logger = LoggerFactory.getLogger(FixedRateFragInsStateTracker.class);
 
+  private static final long SAME_STATE_PRINT_RATE_IN_MS = 10 * 60 * 1000;
+
   // TODO: (xingtanzjr) consider how much Interval is OK for state tracker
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/apache/iotdb/db/mpp/plan/scheduler/FixedRateFragInsStateTracker.java
```diff
@@ -43,9 +43,13 @@ public class FixedRateFragInsStateTracker extends AbstractFragInsStateTracker {
 
   private static final Logger logger = LoggerFactory.getLogger(FixedRateFragInsStateTracker.class);
 
+  private static final long SAME_STATE_PRINT_RATE_IN_MS = 10 * 60 * 1000;
+
   // TODO: (xingtanzjr) consider how much Interval is OK for state tracker
   private static final long STATE_FETCH_INTERVAL_IN_MS = 500;
   private ScheduledFuture<?> trackTask;
+  private FragmentInstanceState lastState;
+  private long durationToLastPrintInMS;
```
JackieTien97, line 52:
```suggestion
  private volatile FragmentInstanceState lastState;
  private volatile long durationToLastPrintInMS;
```

JackieTien97 (changes requested):

PR #6648, opened 2022-07-12
[IOTDB-3475] Fix ResultSet format for empty query

liuminghui233:
see IOTDB-3475.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/utils/QueryDataSetUtils.java
+++ b/server/src/main/java/org/apache/iotdb/db/utils/QueryDataSetUtils.java
@@ -203,7 +203,7 @@ public class QueryDataSetUtils {
       }
       TsBlock tsBlock = optionalTsBlock.get();
       if (tsBlock.isEmpty()) {
-        break;
+        continue;
       }
 
       int currentCount = tsBlock.getPositionCount();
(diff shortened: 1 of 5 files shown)


liuminghui233:
Before:
c623609241dcc11557f3543d2cd4297

After:
a6b7a67aa929ef0f3c5c7e581570aa5

server/src/main/java/org/apache/iotdb/db/utils/QueryDataSetUtils.java
```diff
@@ -202,6 +202,10 @@ public static TSQueryDataSet convertTsBlockByFetchSize(
         break;
       }
       TsBlock tsBlock = optionalTsBlock.get();
+      if (tsBlock.isEmpty()) {
+        break;
+      }
```
JackieTien97, line 207: We cannot simply break here, there is no guarantee that tsBlock won't be empty. If it's really finished, optionalTsBlock.isPresent() will return false.

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/utils/QueryDataSetUtils.java
liuminghui233: fixed

PR #6669, opened 2022-07-13
[IOTDB-3823] Optimize leader routing policy

CRZbulabula:
Now the leader policy will broadcast RegionRouteMap only if the leadership has changed. However, if we use MultiLeader protocol, the ConfigNode-leader still broadcast RegionRouteMap continually since each Region always takes of itself as the leader.


xingtanzjr:
Great

SpriCoder (approved):
LGTM~

PR #6702, opened 2022-07-18
[To rel/0.13][IOTDB-3826]Fix duplicate success when concurrent creating same timeseries

MarcosZyk:
### Problem

see IOTDB-3826

### Cause

The check and creation operation on device node is not locked. Although the operation of upgrading an internal node to device node is locked, different threads may take the internal node before upgrading when entering the synchronized code block, and then all the synchronized check is on the invalidate internal node. The actual device node has already been set to MTree. 

### Solution

Lock the check and creation operation on device node.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTree.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTree.java
@@ -502,12 +502,13 @@ public class MTree implements Serializable {
 
   private Pair<IMNode, Template> checkAndAutoCreateDeviceNode(
       String deviceName, IMNode deviceParent, Template upperTemplate)
-      throws PathAlreadyExistException {
+      throws PathAlreadyExistException, TemplateImcompatibeException {
     if (!deviceParent.hasChild(deviceName)) {
       if (upperTemplate != null && upperTemplate.getDirectNode(deviceName) != null) {
-        throw new PathAlreadyExistException(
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTree.java
```diff
@@ -340,55 +340,23 @@ public IMeasurementMNode createTimeseries(
...
-
-    if (cur.isMeasurement()) {
-      throw new PathAlreadyExistException(cur.getFullPath());
-    }
-
-    MetaFormatUtils.checkTimeseriesProps(path.getFullPath(), props);
-
-    String leafName = path.getMeasurement();
+    PartialPath devicePath = path.getDevicePath();
+    Pair<IMNode, Template> pair = checkAndAutoCreateInternalPath(devicePath);
```
bigreybear, line 345: I see `pair`  is reassigned inside synchronized block, is it necessary to assign here?

bigreybear (commented):
Significant fix ! only one line about assignment to assure

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTree.java
MarcosZyk: The operation in synchronized block is the core modification to ensure the operation on device node is synchronized.

bigreybear (approved):
LGTM~

PR #6704, opened 2022-07-19
[IOTDB-3837] Exclude md5 file when get latest snapshot

SzyWilliam:
Ratis requires to send snapshot file together with its md5 digest. We decide to calculate it once and store md5 in file under the snapshot dir. However, when send snapshot second time, ratis will consider these md5 file also one part of the snapshot. So we have to exclude the md5 file when fetching the latest snapshot.


wangchao316:
I have a question. Will md5 be sent ?

SzyWilliam:
@wangchao316 The md5 digest will be sent if leader installSnapshot to a follower.

JackieTien97 (approved):

PR #6705, opened 2022-07-19
Place the addMetrics for configManager after setUpInternalServices

23931017wu:
### Due to the initialization MetricService operation is after initConfigManager, so when adding configManager indicators metricmanager is empty, cause 'partitionManager. AddMetrics ()' and 'nodeManager.addmetrics ()' execution failed.
image

### Before
image

### After
image
image


neuyilan (approved):
LGTM

PR #6715, opened 2022-07-19
[To rel/0.13][IOTDB-3742]Fix count nodes

MarcosZyk:
## Description

see IOTDB-3742

The node management feature is to help manage the instantiated nodes on MTree, excluding the nodes inside template. 
Since there's memory protection on schema storage, the node info won't be a heavy load to system.

Therefore, remove the node traversal logic inside template.


bigreybear (approved):
LGTM ~

PR #6717, opened 2022-07-19
Add query retry logic to make the query won't fail when some DataNode shutdown

xingtanzjr:
Add retry logic for QueryExecution.

In order to implement the availability for Query that it can switch to alive DataNode's replica when some DataNode is down, we add the retry logic for QueryExecution.

Before every retry, the PartitionCache will be invalid so that the query can fetch latest PartitionCache from ConfigNode to reach the alive DataNode.

See the test here:
image


Beyyes (approved):

server/src/main/java/org/apache/iotdb/db/mpp/execution/QueryStateMachine.java
```diff
@@ -125,6 +133,13 @@ public void transitionToAborted() {
     queryState.set(QueryState.ABORTED);
   }
 
+  public void transitionToFailed() {
+    if (queryState.get().isDone()) {
+      return;
+    }
+    queryState.set(QueryState.FAILED);
+  }
+
```
JackieTien97, line 142: Better to reuse the `public void transitionToFailed(Throwable throwable)`, you can pass a IoTDBException("retry times reach max...") or something like that.

server/src/main/java/org/apache/iotdb/db/mpp/plan/scheduler/ClusterScheduler.java
```diff
@@ -109,11 +109,11 @@ public void start() {
         return;
       }
     } catch (InterruptedException | ExecutionException e) {
-      // If the dispatch failed, we make the QueryState as failed, and return.
+      // If the dispatch request cannot be sent or TException is caught, we will retry this query.
       if (e instanceof InterruptedException) {
         Thread.currentThread().interrupt();
       }
-      stateMachine.transitionToFailed(e);
+      stateMachine.transitionToRetrying(e);
```
JackieTien97, line 116: In most cases, we won't reach there, if the node is killed, we will get a TException in `dispatchRemote` and then throw a FragmentInstanceDispatchException. Finally,  in `dispatchRead` we will catch that Exception and just return a FragInstanceDispatchResult whose successful is false and failureStatus is the root cause.

PR #6724, opened 2022-07-20
[IOTDB-3871] purge log instantly when take a snapshot

SzyWilliam:
# What changes proposed in this PR?
Currently, when leader takes a snapshot, it can purge the logs included in the snapshot. However, by default, if group contains a slow follower, leader will preserve all the raft logs this slow follower doesn't have. This will cause the log in leader accumulate and take a lot of disk space.


OneSizeFitsQuorum (approved):
LGTM

Issue #5880, opened 2022-05-12
Save and re-use this "Random".


MinaQin:
Code Path: confignode/src/main/java/org/apache/iotdb/confignode/manager/PartitionManager.java
<img width="860" alt="image" src="https://user-images.githubusercontent.com/55030443/168055853-74a33d4b-e7a3-4bf0-a0d3-7ca161adb3b3.png">
Creating a new Random object each time a random value is needed is inefficient and may produce numbers which are not random depending on the JDK. For better efficiency and randomness, create a single Random, then store, and reuse it.

The Random() constructor tries to set the seed with a distinct value every time. However there is no guarantee that the seed will be random or even uniformly distributed. Some JDK will use the current time as seed, which makes the generated numbers not random at all.

This rule finds cases where a new Random is created each time a method is invoked.

azhsmesos:
My idea is to add a cache to store random. If this idea is possible, please assigned it to me

azhsmesos:
thank you

PR #6730, opened 2022-07-20
feat(ThreadLocalRandom): replace new Random() with ThradLocalRandom.c…

azhsmesos:


example/mqtt/src/main/java/org/apache/iotdb/mqtt/MQTTClient.java
```diff
@@ -35,7 +35,7 @@ public static void main(String[] args) throws Exception {
     BlockingConnection connection = mqtt.blockingConnection();
     connection.connect();
 
-    Random random = new Random();
+    ThreadLocalRandom random = ThreadLocalRandom.current();
```
wangchao316, line 38: why?   Will performance be affected?

azhsmesos: https://github.com/apache/iotdb/issues/5880 
Threadlocalrandom's method nextseed() that generates random seeds in a multithreaded environment has higher performance, while random is generated through CAS' method

azhsmesos:
想知道这个pr有人跟一下吗？目前不知道ci是哪儿错了

PR #6744, opened 2022-07-21
[IOTDB-3593] Broadcast latest ConfigNodeGroup

imquanke:
### test results on condition: three configNodes, three dataNodes
* one configNode and three dataNodes

* add another one configNode
logs on the first configNode

logs on the new configNode

logs on dataNode

query result on cli

* add another one configNode
* kill the first configNode
logs on the newly elected leader configNode

query result on cli


CRZbulabula (approved):
LGTM~

confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/RemoveConfigNodeProcedure.java
```diff
@@ -76,6 +76,7 @@ protected Flow executeFromState(ConfigNodeProcedureEnv env, RemoveConfigNodeStat
           break;
         case STOP_CONFIG_NODE:
           env.stopConfigNode(tConfigNodeLocation);
+          env.broadCastTheLatestConfigNodeGroup();
```
wangchao316, line 79: I feel broadCastTheLatestConfigNodeGroup should move REMOVE_CONFIG_NODE_PREPARE

PR #6747, opened 2022-07-21
[IOTDB-3923] Add Disordered Cluster Test ResultSet for the Function of last

irvine0109:


integration-test/src/main/java/org/apache/iotdb/itbase/runtime/ClusterTestConnection.java
```diff
@@ -46,6 +46,7 @@ public class ClusterTestConnection implements Connection {
   private final NodeConnection writeConnection;
   private final List<NodeConnection> readConnections;
   private boolean isClosed;
+  private boolean isResultSetDisorder;
```
ericpai, line 49: Use `resultSetDisordered` instead.

integration-test/src/main/java/org/apache/iotdb/itbase/runtime/SortedClusterTestResultSet.java
```diff
@@ -0,0 +1,1218 @@
...
+      if (!result.equals(anotherResult)) {
+        throw new InconsistentDataException(cachedResult, endpoints);
+      }
+    }
+  }
+
+  private <T> T cachedResultAndReturnOne(List<T> data) {
+    for (int i = 0; i < data.size(); i++) {
+      List<StringBuilder> list = cachedResult.get(i);
+      list.get(list.size() - 1).append(data.get(i).toString()).append(",");
```
ericpai, line 120: If `data.get(i)` returns `null`, there will be an NullPointerException thrown. I think it's better to use `List<Set<List<Object>>>` to cache the result data, which is designed as the follow:
1. The first dimension is the node, which is the same as the current design.
2. The second dimension is the row results of one node, as the row sequence is unpredictable, we use a set to do later   comparsion.
3. The thrid dimension is the column result of one row, as we may have different data types in one row, we use a generic list to save them.

Then we can simply use `Objects.equal(cachedResult.get(0), cachedResult.get(i))` to check whether they are consistent.

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBConnection.java
```diff
@@ -421,6 +421,8 @@ public void setClientInfo(String name, String value) throws SQLClientInfoExcepti
       } catch (TException | IoTDBSQLException e) {
         throw new SQLClientInfoException("Set time_zone error: ", null, e);
       }
+    } else if (name.equalsIgnoreCase("is_result_set_disorder")) {
```
ericpai, line 424: Use `result_set_disordered` instead.

integration-test/src/main/java/org/apache/iotdb/itbase/runtime/SortedClusterTestResultSet.java
```diff
@@ -0,0 +1,1218 @@
...
+import java.sql.Statement;
+import java.sql.Time;
+import java.sql.Timestamp;
+import java.util.ArrayList;
+import java.util.Calendar;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+public class SortedClusterTestResultSet implements ResultSet {
```
ericpai, line 48: The name of this class is not consistent with the true meaning of its implementation. How about `DisorderedClusterTestResultSet` ?

integration-test/src/main/java/org/apache/iotdb/itbase/runtime/SortedClusterTestResultSet.java
```diff
@@ -0,0 +1,1218 @@
...
+    for (int i = 1; i < cachedResult.size(); i++) {
+      List<String> anotherResult = new ArrayList<>();
+
+      for (StringBuilder item : cachedResult.get(i)) {
+        anotherResult.add(item.toString());
+      }
+
+      Collections.sort(anotherResult);
+
+      if (!result.equals(anotherResult)) {
```
ericpai, line 111: Use `Objects.equals` instead.

integration-test/src/main/java/org/apache/iotdb/itbase/runtime/SortedClusterTestResultSet.java
```diff
@@ -0,0 +1,1218 @@
...
+    if (result) {
+      for (List<StringBuilder> list : cachedResult) {
+        list.add(new StringBuilder());
+      }
+    }
+    return result;
+  }
+
+  @Override
+  public void close() throws SQLException {
```
ericpai, line 85: As we have enriched the logic in `close()` when using disordered result set, it implies that the developers should remember to close the dataset after usage. Is there any way to force to check the `Connection`, `Statement` and `ResultSet` closed? Here I have two ideas:
1. Use `checkstyle` to prevent unclosed resources, which will be a prefect and simple solution but I'm not sure whether `checkstyle` can really do this.
2. When closing an upper layer resource, It should check whether the direct lower layer resources have been closed. Otherwise it should throw an Exception. So we need to record the allocations of Connection in an Env, allocations of Statement in an Connection, and the allocations of ResultSet in an Statement. After an Env is destoryed, it will check whether there are any unclosed connections. If so, thrown an Exception to tell the developers. So do the Connection and Statement.

ericpai (changes requested):

integration-test/src/main/java/org/apache/iotdb/itbase/runtime/SortedClusterTestResultSet.java
ericpai: If you want the error message to be more precisemore, You can use `List<Set<Map<String, Object>>>` to save the cache.
The third dimension is the map from 'columnName' to 'columnValue'.

irvine0109: Yeah, using `Set` to do the consistency check is simple, but using `Set` can't find the last record item, which used by getXXX() method like `getString(String columnLabel)` or `getString(int columnIndex)` in this ResultSet class. 

The detail is that the record item is one line of data that includes all information about one timeseries, we use the `next()` method in this ResultSet class to change to a new record item to append the information of the next timeseries, if we don't invoke the `next()` method, all getXXX() method will append to the last record item in `cachedResult`, i.e. the sequence is a big key point.

integration-test/src/main/java/org/apache/iotdb/itbase/runtime/SortedClusterTestResultSet.java
irvine0109: Good idea!

integration-test/src/main/java/org/apache/iotdb/itbase/runtime/ClusterTestConnection.java
irvine0109: OK

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBConnection.java
irvine0109: OK

integration-test/src/main/java/org/apache/iotdb/itbase/runtime/SortedClusterTestResultSet.java
ericpai: OK, maybe we can use `List<List<Map<String, Object>>>` in the future, store columnName to columnValue in each map,  and convert to a list in the final comparsion.

PR #6754, opened 2022-07-22
Update metric of confignode's number

23931017wu:
#### Before：  Start 3 ConfigNode
image
image

#### Before
image


SpriCoder (approved):
LGTM~

neuyilan (approved):
I think we merge this pr first, in the later, it would be better to distinguish the online and offline nodes.

23931017wu:
> I think we merge this pr first, in the later, it would be better to distinguish the online and offline nodes.

OK,I plan to use three tags to distinguish them(registered、running、unknown)

PR #6758, opened 2022-07-23
[IOTDB-3935] Refactor TriggerRegistrationService

LIU-WEI-git:
See [IOTDB-3935](https://issues.apache.org/jira/browse/IOTDB-3935).


wangchao316:
hi,  need add some IT for Trigger...

LIU-WEI-git:
> hi, need add some IT for Trigger...

TriggerService MPP development is still working in progress. New IT framework is not suitable for now. The test will be fail since the RPC communicate interface is not defined. Existed local standalone trigger tests all pass after refactor, which indicates the refactor is correct in function and won't influence using origin trigger. I will add new IT after developing all MPP trigger function.

LIU-WEI-git:
Trigger will be redesigned, and this pr should be closed.

PR #6763, opened 2022-07-25
[IOTDB-3942] Support count timeseries where tag1 = v1

RYH61:
add:
example:
 count timeseries where tag1 = v1
 count timeseries root.** where tag1 = v1
 count timeseries where tag1 = v1 group by level = 1
 count timeseries root.sg.** where tag1 = v1 group by level = 1

图片


server/src/main/java/org/apache/iotdb/db/metadata/schemaregion/ISchemaRegion.java
```diff
@@ -135,12 +135,13 @@ Pair<Integer, Set<String>> deleteTimeseries(PartialPath pathPattern, boolean isP
...
    */
-  int getAllTimeseriesCount(PartialPath pathPattern, boolean isPrefixMatch)
+  int getAllTimeseriesCount(PartialPath pathPattern, boolean isPrefixMatch, boolean hasTag)
       throws MetadataException;
 
   // The measurements will be grouped by the node in given level and then counted for each group.
   Map<PartialPath, Integer> getMeasurementCountGroupByLevel(
-      PartialPath pathPattern, int level, boolean isPrefixMatch) throws MetadataException;
+      PartialPath pathPattern, int level, boolean isPrefixMatch, boolean hasTag)
+      throws MetadataException;
```
MarcosZyk, line 144: What about extend a interface with same name but override it? It means keep the old method since it has been used widely, and use the new method to support new feature? I think this makes the modification more clear.

server/src/main/java/org/apache/iotdb/db/metadata/mtree/traverser/counter/MeasurementCounter.java
```diff
@@ -43,6 +62,11 @@ protected boolean processFullMatchedMNode(IMNode node, int idx, int level) {
...
       return false;
     }
+    if (hasTag) {
+      if (!timeseries.contains(node.getFullPath())) {
+        return true;
+      }
+    }
     count++;
     return true;
   }
```
neuyilan, line 72: if haveTag=true, here return, the count will not be calculated?
the same as others

neuyilan (approved):
LGTM

MarcosZyk (approved):
Excellent work! LGTM~

PR #6770, opened 2022-07-26
[IOTDB-3961] fix bug of Tsfile compact

jamber001:
## Description
https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-3961

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/compaction/CompactionWorker.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/compaction/CompactionWorker.java
@@ -47,23 +47,27 @@ public class CompactionWorker implements Runnable {
   @Override
   public void run() {
     while (!Thread.currentThread().isInterrupted()) {
-      AbstractCompactionTask task = null;
       try {
-        task = compactionTaskQueue.take();
-      } catch (InterruptedException e) {
-        log.warn("CompactionThread-{} terminates because interruption", threadId);
-        return;
-      }
-      if (task != null) {
-        // add metrics
-        CompactionMetricsRecorder.recordTaskInfo(
-            task, CompactionTaskStatus.POLL_FROM_QUEUE, compactionTaskQueue.size());
-        if (task.checkValidAndSetMerging()) {
-          CompactionTaskSummary summary = task.getSummary();
-          CompactionTaskFuture future = new CompactionTaskFuture(summary);
-          CompactionTaskManager.getInstance().recordTask(task, future);
-          task.start();
+        AbstractCompactionTask task = null;
+        try {
+          task = compactionTaskQueue.take();
+        } catch (InterruptedException e) {
+          log.warn("CompactionThread-{} terminates because interruption", threadId);
+          return;
         }
+        if (task != null) {
+          // add metrics
+          CompactionMetricsRecorder.recordTaskInfo(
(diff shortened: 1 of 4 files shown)


jamber001:
@THUMarkLau   This is 1 critical bug fix.
Could you help review it?

server/src/main/java/org/apache/iotdb/db/engine/compaction/CompactionWorker.java
```diff
@@ -46,25 +46,29 @@ public CompactionWorker(
 
   @Override
   public void run() {
-    while (!Thread.currentThread().isInterrupted()) {
```
ericpai, line 49: Two more easy solutions:
1. Let CompactionWorker extend WrappedRunnable which will automatically log the errors.
2. Use `taskExecutionPool.execute()` instead of `taskExecutionPool.submit()`, which will make use of `IoTDBDefaultThreadExceptionHandler` if any exception is thrown.

server/src/main/java/org/apache/iotdb/db/engine/compaction/CompactionWorker.java
```diff
@@ -46,25 +46,29 @@ public CompactionWorker(
...
+            CompactionTaskSummary summary = task.getSummary();
+            CompactionTaskFuture future = new CompactionTaskFuture(summary);
+            CompactionTaskManager.getInstance().recordTask(task, future);
+            task.start();
+          }
         }
       }
+    } catch (Throwable t) {
+      log.error("CompactionWorker.run(), Exception.", t);
     }
```
THUMarkLau, line 72: If exception is throw inside the `task.start()`, the compaction thread will still exits and cannot recover. Move the `try-catch` to the inner of the loop.

THUMarkLau (changes requested):
Plz move the `try-catch` to the inner of the loop

server/src/main/java/org/apache/iotdb/db/engine/compaction/CompactionWorker.java
jamber001: > If exception is throw inside the `task.start()`, the compaction thread will still exits and cannot recover. Move the `try-catch` to the inner of the loop.

We have ever discussed this point internally. 
If we let compact continuing after exception happens, we worry that it will bring risk to damage tsfile files.
It is safe way to print erorr log and interrupt consequent compact.

So, do you still want to continue compacting after exception?

choubenson: Exceptions in the compaction process will be captured and processed internally in the `innerSpaceCompactionTask`and `CrossSpaceCompactionTask`, and there will be no file corruption. Even if an exception is thrown to the `compactionWorker`, it will be an exception elsewhere. So you can place your `try catch` inside.

THUMarkLau: Benson is right, all throwable should be handled correctly by CompactionExceptionHandler.

jamber001: Thank you, 
I chang the try--catch to the inner of the loop.

ericpai (approved):

choubenson (approved):
LGTM

PR #6779, opened 2022-07-26
[To rel/0.13][IOTDB-3831]Fix ttl on last cache

MarcosZyk:
## Description

1. Add lastCache check in TTLTask in StorageEngine
2. Add lastValue filter to lastPointReader


HTHou (approved):
LGTM

PR #6783, opened 2022-07-26
[To rel/0.13] [IOTDB-3943] Optimize session timeout feature

suchenglong:
## Description
Users may forget to close sessions. Such idle connections lead to a waste of server resource and in the worst case, take up all the memory of server. Session timeout is trying to solve such problem.

https://github.com/apache/iotdb/pull/3483

Currently this feature has other problems

Once the session is cleared by session timeout, the client will not get the status, when client tries to use the session again, server will throw NPE. We should throw connect exception here to client, so that session pool can catch it and uses another session to connect.
Metric module will insert data to IoTDB, we need to consider use session pool in this module.

<hr>

This PR has:
  for an unfamiliar reader.
  for code coverage.

<hr>


suchenglong:
fix the issue of session timeout

HTHou (approved):

server/src/main/java/org/apache/iotdb/db/query/control/SessionTimeoutManager.java
```diff
@@ -106,6 +106,13 @@ public static SessionTimeoutManager getInstance() {
     return SessionTimeoutManagerHelper.INSTANCE;
   }
 
+  public boolean isSessionExits(long sessionId) {
```
HTHou, line 109:
```suggestion
  public boolean isSessionAlive(long sessionId) {
```

service-rpc/src/main/java/org/apache/iotdb/rpc/SessionTimeoutException.java
```diff
@@ -0,0 +1,43 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iotdb.rpc;
+
+import org.apache.iotdb.service.rpc.thrift.TSStatus;
+
+public class SessionTimeoutException extends StatementExecutionException {
```
qiaojialin, line 23:
```suggestion
public class SessionTimeoutException extends IoTDBConnectionException {
```

suchenglong: 在现有的方法中只返回了StatementExecutionException，如果修改为IoTDBConnectionException，需要修改所有引用该方法的异常处理部分。
image

HTHou (approved):

PR #6804, opened 2022-07-27
[IOTDB-3781] Reinforce the regionCleaner task's startup logic

CRZbulabula:
The regionCleaner task in ConfigNode's PartitionManager will start as soon as the PartitionManager is initialized, which is unsafe however. Ever interior ScheduledExecutor should not be started until the current ConfigNode confirms it's leadership. And these inner ScheduleExecutor should be shutdown immediatly when the current ConfigNode loses it's leadership.

Test result, all ScheduledExecutor started after the current ConfigNode becomes leader:

image


wangchao316 (approved):
LGTM

RYH61 (approved):
LGTM~

PR #6808, opened 2022-07-27
[To rel/0.13][IOTDB-3969]Fix timeseries statistic bug

MarcosZyk:
### Problem
When delete storage group, the timeseries represent by template will be counted and the result will be used to update the timeseries statistic number. The number doesn't update when activate template. Thus the number will be negative.

### Solution
Use two number to statistic template series and normal series separately.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
@@ -324,7 +324,9 @@ public class MManager {
             totalNormalSeriesNumber,
             AtomicLong::get,
             Tag.NAME.toString(),
-            "normal-timeSeries");
+            "timeSeries",
+            Tag.TYPE.toString(),
+            "normal");
 
     MetricsService.getInstance()
         .getMetricManager()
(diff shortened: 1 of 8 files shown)


bigreybear (approved):
LGTM~

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
```diff
@@ -319,10 +321,20 @@ private void startStatisticCounts() {
...
+
+    MetricsService.getInstance()
+        .getMetricManager()
+        .getOrCreateAutoGauge(
+            Metric.QUANTITY.toString(),
+            MetricLevel.IMPORTANT,
+            totalTemplateSeriesNumber,
+            AtomicLong::get,
+            Tag.NAME.toString(),
+            "template-timeSeries");
```
SpriCoder, line 337: From my perspective, I think these two metrics can use `type` tag to distinguish.
For example,  TAG.NAME.toString(), "timeSeries", TAG.TYPE.toString(), "normal/template"
Besides, please update Metric doc at the same time: 
1. docs/UserGuide/Maintenance-Tools/Metric-Tool.md
2. docs/zh/UserGuide/Maintenance-Tools/Metric-Tool.md

SpriCoder (changes requested):
PTAL~

server/src/main/java/org/apache/iotdb/db/metadata/MManager.java
MarcosZyk: Fixed.

SpriCoder (approved):
LGTM~

PR #6809, opened 2022-07-27
[To rel/0.13][IOTDB-3906][IOTDB-3921]Support template DDL under DoubleLive mode

bigreybear:
## Description
Add support for create/drop, set/unset, activate/deactivate, append/prune template operation when considering whether to invoke a double-live session during process of a DDL statement.


CRZbulabula (approved):
LGTM~

PR #6812, opened 2022-07-27
[IOTDB-2669] C++ client optimize the values type in tablet class

liruizhi19:
C++ client optimize the values type in tablet class.

The old values type is:
std::vector<std::vector<std::string>> values;

The new values type is:
std::vector<void*> values;

For eliminating two bytes conversion, so the performance is improved.
[performance test.xlsx](https://github.com/apache/iotdb/files/9198932/performance.test.xlsx)
image


liuminghui233 (approved):
LGTM :-)

PR #6818, opened 2022-07-28
[IOTDB-3986] Unified ConfigNode heartbeat request

CRZbulabula:
The THeartbeatReq generated in each heartbeat loop is independent for each node after several code iteration, which is against the original design. Therefore, we should unify the THeartbeatReq used in each heartbeat cycle.


Beyyes (approved):
This pr has no problem.
I just need some time to test in old branch~

PR #6757, opened 2022-07-22
[IOTDB-3934] Compatibility of Apache IoTDB with InfluxDB——example client

jun0315:
see https://issues.apache.org/jira/browse/IOTDB-3934

PR #6828, opened 2022-07-29
Influxdb service adapts to distributed MPP framework and fixes some bugs in InfluxFunction and TagInfoRecord

KeeProMise:
## Background
The existing InfluxDB service does not adapt to the new distributed MPP framework. I developed a new InfluxDB service that adapts to the new distributed MPP framework without affecting the original influxDB service.

### added some new classes
 - **AbstractInfluxDBMetaManager**: manages the mapping of tags->path, provides template methods, and subclasses only need to implement abstract methods
   - **InfluxDBMetaManager**: An existing implementation class that uses ServiceProvider to perform physical planning.
   - **NewInfluxDBMetaManager**: An implementation class adapted to the MPP framework, which uses ClientRPCServiceImpl to process read and write requests.
- **AbstractQueryHandler**: handles query requests, provides template methods, and subclasses only need to implement abstract methods.
   - **QueryHandler**: Existing implementation class, which uses ServiceProvider to process the read physical plan
   - **NewQueryHandler**: An implementation class adapted to the MPP framework, which uses ClientRPCServiceImpl to process read requests.
- **InfluxDBSeviceImpl**: The existing influxDBRPCSevice instance.
- **NewInfluxDBSeviceImpl**: an instance of influxDBRPCSevice adapted to the MPP framework.
Detailed class diagrams and designs：https://apache-iotdb.feishu.cn/docx/doxcnZzEZ1BwXS6EQfrUEuGOX4d#doxcnO2Kue0WCMog6En9LtAJEJh
### fixed some bugs
the classes involved：
- **TagInfoRecords**：This class is used to write the sequence of each tag into IoTDB. At present, there may be multiple adjacent tags with the same timestamp, which causes the tags written later in IoTDB to overwrite the tags written before.
- **InfluxMeanFunction**：The type of the incoming parameter is double, but the type of the member variable sum is long, resulting in a type mismatch exception
- **InfluxFirstFunction and InfluxLastFunction**：There is a problem with the logic. The existing implementation reverses the logic of last and first, and the result of the calculation is null.


server/src/main/java/org/apache/iotdb/db/protocol/influxdb/function/selector/InfluxFirstFunction.java
```diff
@@ -45,10 +45,7 @@ public InfluxFunctionValue calculateByIoTDBFunc() {
 
   @Override
   public void updateValueIoTDBFunc(InfluxFunctionValue... functionValues) {
-    if (value == null && getTimestamp() == null) {
-      value = functionValues[0].getValue();
-      setTimestamp(functionValues[0].getTimestamp());
-    } else if (getTimestamp() < functionValues[0].getTimestamp()) {
+    if (functionValues[0].getTimestamp() < getTimestamp()) {
```
jun0315, line 48: If it is the first time to enter the updateValueIoTDBFunc, the result of getTimestamp should be NULL, Will this cause NullPointerException?

jun0315, line 48: If it is the first time to enter the updateValueIoTDBFunc, the result of getTimestamp should be NULL, Will this cause NullPointerException?

jun0315:
@LIU-WEI-git If you have free time, please review it together :D

server/src/main/java/org/apache/iotdb/db/protocol/influxdb/function/selector/InfluxFirstFunction.java
KeeProMise:
```
public InfluxLastFunction(List<Expression> expressionList) {
    super(expressionList);
    this.setTimestamp(Long.MIN_VALUE);
  }
public InfluxFirstFunction(List<Expression> expressionList) {
    super(expressionList);
    this.setTimestamp(Long.MAX_VALUE);
 }
```
The value of the timestamp will be assigned in the constructor

jun0315: OK, I remember!

server/src/main/java/org/apache/iotdb/db/protocol/influxdb/meta/TagInfoRecords.java
```diff
@@ -65,11 +66,8 @@ public TagInfoRecords() {
 
   public void add(String database, String measurement, String tag, int order) {
     deviceIds.add(TAG_INFO_DEVICE_ID);
-    if (times.size() == 0) {
-      times.add(System.currentTimeMillis());
-    } else {
-      times.add(times.get(times.size() - 1) + 1);
-    }
+    // Multiple adjacent records, possibly with the same timestamp
+    times.add(System.currentTimeMillis());
```
LIU-WEI-git, line 70: Hi, nice work! I have a little confuse about this bug fix. Since the value of this timestamp just be used to record the order of inserting, why not use AtomicLong and its getAndIncrement method to generate the timestamp?

KeeProMise: Your idea is indeed feasible, you can first take a look at https://github.com/apache/iotdb/pull/6757, this is @jun0315 recently merged code, he uses the system time here.

KeeProMise: When I submitted the code and merged the conflict, I forgot to resolve the conflict and used my code directly. How to solve the bug here, what are @jun0315  thoughts?

jun0315: @LIU-WEI-git Thanks for your review, I think your idea is very good. If we consider the distributed situation, using AtomicLong is a good solution.
@KeeProMise I also support use AtomicLong and its getAndIncrement method.

KeeProMise: ok , it has been modified to use the AtomicLong to generate the tag timestamps

jun0315: @KeeProMise I saw your latest commit. Nice work!

MarcosZyk (approved):

jun0315:
Thanks for your contribution, LGTM!

PR #6829, opened 2022-07-29
[IOTDB-3985] Retry removePeer for region bug

BanDuWuZi:
the retry  logic in removePeer() method exist bug, if retry happen the method always return failed status, need to fix.

class: RegionMigrateService.java


wangchao316 (approved):
LGTM

CRZbulabula (approved):
LGTM～

PR #6830, opened 2022-07-29
Extract ISchemaTree interface

MarcosZyk:
### Motivation
The original schemaTree is designed for cluster mode operations, including schema info transport and merge.

For standalone mode, we don't need to construct and initialize schemaTree objects for query and insertion. The best way is to use LocalConfigNode and SchemaEngine to implement a StandaloneSchemaTree, which requires few extra memory usage.
 
### Modification

Thus we extract an interface ```ISchemaTree```, which defines the necessary schema info query methods. 
The original schemaTree is renamed to ```ClusterSchemaTree```.

### Future work
For standalone mode, we will implement ```StandaloneSchemaTree```.


liuminghui233 (approved):
LGTM :-)

PR #6831, opened 2022-07-29
[IOTDB-3844] Clear up Non-Seed-ConfigNode register process

CRZbulabula:
Updates:

1. It's enough to use TSStatus as registerConfigNode interface's response
2. Rename the input parameter of addConsensusGroup interface to TAddConsensusGroupReq

And these modifications are safe:

image


RYH61 (approved):
LGTM~

confignode/src/main/java/org/apache/iotdb/confignode/manager/ConfigManager.java
```diff
@@ -698,103 +697,65 @@ public TPermissionInfoResp checkUserPrivileges(
...
     if (status.getCode() == TSStatusCode.SUCCESS_STATUS.getStatusCode()) {
-      TConfigNodeRegisterResp errorResp1 = checkConfigNodeRegisterResp(req);
-      if (errorResp1 != null) return errorResp1;
+      TSStatus errorStatus = checkConfigNodeGlobalConfig(req);
+      if (errorStatus != null) {
+        return errorStatus;
+      }
 
       procedureManager.addConfigNode(req);
-      return nodeManager.registerConfigNode(req);
```
wangchao316, line 710: nodemanager does not registerConfigNode?

CRZbulabula: The nodeManager.registerConfigNode was used to get the current PartitionRegionId and ConfigNodeList. However, the purpose of registerConfigNode is only to activate the addConfigNodeProcedure, thus we don't need to get these information anymore. Therefore, I just delete this redundant interface.

wangchao316 (approved):
LGTM

PR #6836, opened 2022-07-29
[IOTDB-4000] Add the control of the size of wal folder in multiLeader consensus.

SpriCoder:
1. Add `multiLeader_max_wal_buffer_size_in_byte` param in datanode config.
2. Add the control of the size of wal folder in multiLeader consensus.

See: https://issues.apache.org/jira/browse/IOTDB-4000


server/src/assembly/resources/conf/iotdb-datanode.properties
```diff
@@ -178,6 +178,11 @@ target_config_nodes=127.0.0.1:22277
 # Datatype: long
 # delete_wal_files_period_in_ms=20000
 
+# Maximum size of wal buffer used in MultiLeader Consensus
+# If it's a value smaller than 0, use the default value 10 * 1024 * 1024 * 1024 bytes (10GB).
+# Datatype: long
+# multiLeader_max_wal_buffer_size_in_byte=10737418240
```
neuyilan, line 184: The description is inconsistent with the code. In fact, when the wal size exceeds this value, the writing will slow down. And the final stored wal size may exceed this value. I think we can revise the description to avoid misunderstanding.

neuyilan: And i think we can change the `multiLeader_max_wal_buffer_size_in_byte` to `multiLeader_max_wal_size_in_byte` becase the `multiLeader_max_wal_buffer_size_in_byte` may confuse people with the `wal_buffer_size_in_byte`

consensus/src/main/java/org/apache/iotdb/consensus/multileader/MultiLeaderServerImpl.java
```diff
@@ -110,6 +109,17 @@ public void stop() {
...
     synchronized (stateMachine) {
+      if (needToThrottleDown()) {
+        logger.info(
+            "[Throttle Down] index:{}, safeIndex:{}",
+            getIndex(),
+            getCurrentSafelyDeletedSearchIndex());
+        try {
+          stateMachine.wait(config.getReplication().getThrottleTimeOutMs());
+        } catch (InterruptedException e) {
+          logger.error("Failed to throttle down because ", e);
```
xingtanzjr, line 120: Add `Thread.getCurrentThread.interrupt()`

server/src/assembly/resources/conf/iotdb-datanode.properties
xingtanzjr: Another suggestion is that we can use a more concrete name for this parameter according to our scenario such as `multi_leader_throttle_down_threshold_by_wal`

server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
```diff
@@ -954,6 +954,9 @@ public class IoTDBConfig {
   /** Maximum execution time of a DriverTask */
   private int driverTaskExecutionTimeSliceInMs = 100;
 
+  /** Maximum size of wal buffer used in MultiLeader consensus. Unit: byte */
+  private long multiLeaderMaxWalBufferSize = 10 * 1024 * 1024 * 1024L;
```
xingtanzjr, line 958: Let's change the default value to 50GB

consensus/src/main/java/org/apache/iotdb/consensus/config/MultiLeaderConfig.java
```diff
@@ -253,6 +274,9 @@ public static class Builder {
       private int maxWaitingTimeForAccumulatingBatchInMs = 500;
       private long basicRetryWaitTimeMs = TimeUnit.MILLISECONDS.toMillis(100);
       private long maxRetryWaitTimeMs = TimeUnit.SECONDS.toMillis(20);
+      private long maxWalBufferSize = 10 * 1024 * 1024 * 1024L;
```
xingtanzjr, line 277: we can use more simple names here such as `throttleDownThreshold` and `throttleUpThreshold`

server/src/assembly/resources/conf/iotdb-datanode.properties
SpriCoder: Get it, I will fixed it by using `multi_leader_throttle_down_threshold_in_byte`

server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
SpriCoder: Get it

consensus/src/main/java/org/apache/iotdb/consensus/config/MultiLeaderConfig.java
SpriCoder: Get it~

consensus/src/main/java/org/apache/iotdb/consensus/multileader/MultiLeaderServerImpl.java
SpriCoder: okay~

xingtanzjr (approved):

PR #6855, opened 2022-08-01
[To rel/0.13] Improve the process of secondary

ljn55966005:


server/src/main/java/org/apache/iotdb/db/engine/StorageEngine.java
```diff
@@ -393,6 +404,14 @@ public void start() {
     startTimedService();
   }
 
+  private void checkSecondaryIsLife() {
```
Beyyes, line 407:
```suggestion
  private void checkSecondaryAlive() {
```

Beyyes: `life` is a noun

server/src/main/java/org/apache/iotdb/db/engine/StorageEngine.java
```diff
@@ -114,6 +97,7 @@ public class StorageEngine implements IService {
 
   private static final IoTDBConfig config = IoTDBDescriptor.getInstance().getConfig();
   private static final long TTL_CHECK_INTERVAL = 60 * 1000L;
+  private static final long HEARTBEAT_CHECK_INTERVAL = 30L;
```
Beyyes, line 100: If it's too long to check the alive info of Secondary in every 30 seconds

server/src/main/java/org/apache/iotdb/db/doublelive/OperationSyncProducer.java
```diff
@@ -33,22 +35,47 @@
...
+      Pair<ByteBuffer, OperationSyncPlanTypeUtils.OperationSyncPlanType> planPair,
+      String deviceName) {
+    ByteBuffer headBuffer;
+    headBuffer = planPair.left;
+    headBuffer.position(0);
+    int index;
+    if (deviceName == null) {
+      index = operationSyncQueues.size();
+    } else {
+      index = deviceName.hashCode() % (operationSyncQueues.size() - 1);
```
Beyyes, line 62: The result of deviceName.hashCode() may be negative

server/src/main/java/org/apache/iotdb/db/doublelive/OperationSyncConsumer.java
```diff
@@ -39,10 +40,10 @@ public class OperationSyncConsumer implements Runnable {
 
   public OperationSyncConsumer(
       BlockingQueue<Pair<ByteBuffer, OperationSyncPlanTypeUtils.OperationSyncPlanType>>
-          OperationSyncQueue,
+          operationSyncQueue,
       SessionPool operationSyncSessionPool,
       OperationSyncLogService dmlLogService) {
-    this.OperationSyncQueue = OperationSyncQueue;
+    this.OperationSyncQueue = operationSyncQueue;
```
Beyyes, line 46: this.operationSyncQueue
The first letter should be lower case.

server/src/main/java/org/apache/iotdb/db/doublelive/OperationSyncProducer.java
```diff
@@ -33,22 +35,52 @@
...
+  public void put(
+      Pair<ByteBuffer, OperationSyncPlanTypeUtils.OperationSyncPlanType> planPair,
+      String deviceName) {
+
+    ByteBuffer headBuffer;
+    headBuffer = planPair.left;
+    headBuffer.position(0);
+    int index;
+    if (deviceName == null) {
+      index = operationSyncQueues.size() - 1;
```
CRZbulabula, line 61: Would it be better if we use random selection here?

server/src/main/java/org/apache/iotdb/db/doublelive/OperationSyncProducer.java
```diff
@@ -33,22 +35,52 @@
...
+    } else {
+      try {
+        index = Math.abs(deviceName.hashCode()) % (operationSyncQueues.size() - 1);
+      } catch (Exception e) {
+        index = 1;
+      }
```
CRZbulabula, line 67: I wonder in which case will this block of code throw an Exception? And the same as before, would it be better if we use random selection when some Exceptions are thrown?

server/src/main/java/org/apache/iotdb/db/doublelive/OperationSyncProducer.java
```diff
@@ -33,22 +35,52 @@
...
+    }
+    for (int i = 0; i < RETRY; i++) {
+      // retry 3 times
+      if (operationSyncQueues.get(index).offer(planPair)) {
+        return;
+      }
+    }
```
CRZbulabula, line 74: It's better to sleep 1s or less between each loop of retry.

server/src/main/java/org/apache/iotdb/db/engine/StorageEngine.java
```diff
@@ -87,33 +86,17 @@
...
-import java.util.Set;
-import java.util.concurrent.ArrayBlockingQueue;
-import java.util.concurrent.BlockingQueue;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Future;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
```
CRZbulabula, line 91: We should always avoid to `import *`

server/src/main/java/org/apache/iotdb/db/engine/StorageEngine.java
```diff
@@ -122,6 +105,7 @@ public class StorageEngine implements IService {
   private static OperationSyncProducer operationSyncProducer;
   private static OperationSyncDDLProtector operationSyncDDLProtector;
   private static OperationSyncLogService operationSyncDDLLogService;
+  private static boolean isSecondaryLife;
```
CRZbulabula, line 108: This field should better to be set to AtomicBoolean or volatile for thread-safety

CRZbulabula (changes requested):
PTAL

server/src/main/java/org/apache/iotdb/db/doublelive/OperationSyncProducer.java
ljn55966005: we must prove the order of the same device

server/src/main/java/org/apache/iotdb/db/doublelive/OperationSyncProducer.java
ljn55966005: when the queue size is 1

Beyyes (approved):
LGTM

qiaojialin (approved):

PR #6857, opened 2022-08-01
[To rel/0.13][IOTDB-3861]Adds null value handling logic to the Session API

suchenglong:
## Description
Adds null value handling logic to the Session API

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -1038,6 +1138,210 @@ public void insertRecords(
...
+        nullMap.put(measurementsList.get(i), valuesList.get(i));
+        valuesList.remove(i);
+        measurementsList.remove(i);
+      }
+    }
+    if (valuesList.size() == 0) {
+      logger.warn("All values of the {} are null,null values are {}", deviceId, nullMap.toString());
+      return true;
+    } else {
+      logger.warn("Some values of {} are null,null values are {}", deviceId, nullMap.toString());
```
qiaojialin, line 1320: As we allow users to pass null values in this interface, we could log info.

suchenglong: now,when code like that,I use log info.

PR #6861, opened 2022-08-02
[TO rel/0.13] [IOTDB-4018] The path in the permissions contains wildcards * and ** path adaptation does not work

RYH61:
create user, grant insert_timeseries
图片

login new user
图片

login root,grant read_timeseries
图片

login newuser
图片


qiaojialin:
This verification is not needed. We could allow user to insert data on root.**, then users insert into root.sg.d1.s1

RYH61:
> This verification is not needed. We could allow user to insert data on root.**, then users insert into root.sg.d1.s1

ok

SpriCoder (commented):
Good job! Please check CI.

Issue #6710, opened 2022-07-19
C++接口执行查询SQL，返回的结果集记录数错误


xuzhi01:
执行的查询SQL:
select netvalue from  root.wd.mnl2.wdw.0 where time >=1656777600000 and time < 1656864000000
实际应该返回的记录数应该是86400条，但是通过C++接口查询返回的结果集只能提取到10000+记录（不固定，比如10153条），且只有前面10000条记录是正确的，第10001条以及之后的记录都是错误的。

1656787596000; 51355
1656787597000; 50528
1656787598000; 4407
1656787599000; 46189        -- 此处是第10000条记录
0; 0
0; 0
0; 0
0; 0
0; 0
-1601321572565188592; 0
8315171469725540462; 0
7310598255932695840; 0
7382085253694780015; 0
8031095125825187182; 0
7796344995253268016; 0
2339453316362477684; 0
7596839335601451318; 0
3834312843041583152; 0
...

执行查询SQL的函数如下：
void query()
{
	RowRecord *x;
	int i = 0;

	ofstream  outFile;
	outFile.open("f:\\out_iotdb.txt");

	unique_ptr<SessionDataSet> dataSet = session->executeQueryStatement(
		"select netvalue from  root.wd.mnl2.wdw.0 where time >=1656777600000 and time < 1656864000000");

	for (const string &name : dataSet->getColumnNames()) {
		cout << name << endl;
	}

	dataSet->setBatchSize(1000);
	bool n;
	while (1) {
		if (!dataSet->hasNext()) break;

		x = dataSet->next();
		outFile << x->timestamp << "; " << x->fields[0].intV << endl;
		i++;
	}

	dataSet->closeOperationHandle();
	outFile.close();
}

jixuan1989:
Hi, @liruizhi19  can you have a check?

liruizhi19:
Problem has been resolved.
Later I will submit PR.

PR #6870, opened 2022-08-02
[IOTDB-4023]C++ interface execute SQL query statement and the returned result set records error

liruizhi19:
To solve the problem “C++ interface execute SQL query statement and the returned result set records error”


jixuan1989:
you may need to link Github issue in the PR.  #6710

qiaojialin (approved):

PR #6873, opened 2022-08-03
[IOTDB-4026] unknown PhysicalPlan type: 51

BanDuWuZi:
when remove datanode, confignode log has IOException:

java.io.IOException: unknown PyhsicalPlan type: 51


wangchao316 (approved):
LGTM

Issue #6751, opened 2022-07-22
How to implement this query method


skyplume:
For the data of the same indicator over a period of time, I want to clean out the data with the same value as the previous one, and only keep the changed data. I want to achieve the same effect as “SELECT a.id ,a.value FROM test a left join test b on a.id+1=b.id where  a.value!=b.value” does, but the DML and UDF provided by querying the official website documents seem to have no way to achieve similar effects. 
How to achieve this effect? thank you

qiaojialin:
Hi, this may be implemented by a UDF. You could buffer the current data, and output the current data once you meet a different value.

Wei-hao-Li:
> For the data of the same indicator over a period of time, I want to clean out the data with the same value as the previous one, and only keep the changed data. I want to achieve the same effect as “SELECT a.id ,a.value FROM test a left join test b on a.id+1=b.id where a.value!=b.value” does, but the DML and UDF provided by querying the official website documents seem to have no way to achieve similar effects. How to achieve this effect? thank you

Hi, we have implement a UDF as Built-in Function to do this job, you can see [PR](https://github.com/apache/iotdb/pull/6874).

skyplume:
Sorry, I just finished two weeks of sick leave. I will check the new UDF. Thank you for your help!

PR #6874, opened 2022-08-03
[IOTDB-3973] Implement a new UDF named 'deDup' to only keep distinct …

Wei-hao-Li:
related issue: https://github.com/apache/iotdb/issues/6751


node-commons/src/main/java/org/apache/iotdb/commons/udf/builtin/UDTFDeDuplication.java
```diff
@@ -0,0 +1,139 @@
...
+ */
+public class UDTFDeDuplication implements UDTF {
+  boolean isFirst = true;
+  Type dataType;
+  boolean cacheBoolean;
+  int cacheInt;
+  long cacheLong;
+  float cacheFloat;
+  double cacheDouble;
+  String cacheString;
```
JackieTien97, line 43: use private

JackieTien97 (changes requested):

JackieTien97 (approved):

PR #6886, opened 2022-08-03
[IOTDB-4028] Npe in updateNodeLoadStatistic of LoadManager when cluster first starts.

23931017wu:

### DIFF
--- a/confignode/src/main/java/org/apache/iotdb/confignode/persistence/NodeInfo.java
+++ b/confignode/src/main/java/org/apache/iotdb/confignode/persistence/NodeInfo.java
@@ -269,7 +269,7 @@ public class NodeInfo implements SnapshotProcessor {
     return result;
   }
 
-  /** Return the number of registered ConfigNodes */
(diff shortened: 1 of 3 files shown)


confignode/src/main/java/org/apache/iotdb/confignode/persistence/NodeInfo.java
```diff
@@ -268,6 +268,21 @@ public int getRegisteredConfigNodeCount() {
     }
     return result;
   }
+
+  /** Return the number of registered ConfigNodes */
+  public int getRegisteredNodeCount() {
+    int result;
```
neuyilan, line 274: Change the comment, please. actually, this method calculated the total number of nodes for confignodes and datanodes.

23931017wu: Okay, I get it ~

neuyilan (approved):

PR #6888, opened 2022-08-03
[IOTDB-4037] reduce thread count for new standalone server

xingtanzjr:
* Clean up useless code and parameters for new standalone server
* Reduce the thread number for new standalone server


HTHou (approved):
LGTM

PR #6889, opened 2022-08-03
[IOTDB-4037] Change default query execute threads to cpu cores

JackieTien97:
Change default query execute threads to cpu cores and don't set up DataExchangeService in standalone mode.


HTHou (approved):
LGTM

PR #6893, opened 2022-08-04
modify deserialize function of FunctionExpression and fix tests of String UDF

Wei-hao-Li:


tsfile/src/main/java/org/apache/iotdb/tsfile/utils/ReadWriteIOUtils.java
```diff
@@ -691,6 +692,22 @@ public static Map<String, String> readMap(ByteBuffer buffer) {
     return map;
   }
 
+  public static LinkedHashMap<String, String> readLinkedHashMap(ByteBuffer buffer) {
+    int length = readInt(buffer);
+    if (length == NO_BYTE_TO_READ) {
+      return null;
```
SteveYurongSu, line 698: It's not a good practice to return null here. It may cause NPE.

I knew that there are a lot of `return null` in this file. I think it's time to do refactor.

SteveYurongSu (changes requested):

tsfile/src/main/java/org/apache/iotdb/tsfile/utils/ReadWriteIOUtils.java
SteveYurongSu: image

This is not good :(

JackieTien97 (approved):

PR #6899, opened 2022-08-05
[IOTDB-3845] Add annotations in confignode.thrift

CRZbulabula:
Ad interface annotation in confignode.thrift file

### DIFF
--- a/confignode/src/main/java/org/apache/iotdb/confignode/manager/ClusterSchemaManager.java
+++ b/confignode/src/main/java/org/apache/iotdb/confignode/manager/ClusterSchemaManager.java
@@ -185,9 +185,9 @@ public class ClusterSchemaManager {
           storageGroups.add(storageGroup);
           dnlToSgMap.put(dataNodeLocation.getDataNodeId(), storageGroups);
         } else {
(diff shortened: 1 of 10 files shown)


confignode/src/main/java/org/apache/iotdb/confignode/manager/ClusterSchemaManager.java
```diff
@@ -354,25 +354,15 @@ public TGetAllTemplatesResp getAllTemplates() {
...
     return resp;
   }
 
-  /**
-   * show nodes in schema template
-   *
-   * @param req
-   * @return
-   */
+  /** show nodes in schema template */
```
wangchao316, line 365: why ?   Generally, comments must contain parameter and return value descriptions.

CRZbulabula: But this annotation doesn't interprete the meaning of parameter and return value. I've left TODO for the relevant schoolmates and hope they'll add more detailed instructions later.

Beyyes (approved):

RYH61 (approved):
LGTM~ I will add annotations in the next PR

PR #6911, opened 2022-08-08
[IOTDB-3980][Metric] Use asynchronous way to collect some predefined metrics.

SpriCoder:
See more details: https://issues.apache.org/jira/browse/IOTDB-3980.

Use asynchronous way to collect some predefined metrics. they are `file_count`, `file_size`, `sys_disk_total_space`, `sys_disk_free_space`.
Also, I add `asyncCollectPeriodInSecond` param for use to control the task that updated the metrics above.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/service/metrics/predefined/FileMetrics.java
+++ b/server/src/main/java/org/apache/iotdb/db/service/metrics/predefined/FileMetrics.java
@@ -25,119 +25,70 @@ import org.apache.iotdb.db.conf.IoTDBDescriptor;
 import org.apache.iotdb.db.service.metrics.enums.Metric;
 import org.apache.iotdb.db.service.metrics.enums.Tag;
 import org.apache.iotdb.metrics.MetricManager;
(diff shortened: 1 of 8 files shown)


server/src/main/java/org/apache/iotdb/db/service/metrics/predefined/FileMetrics.java
```diff
@@ -146,4 +98,112 @@ public void bindTo(MetricManager metricManager) {
...
+            .sum();
+    unsequenceFileTotalSize =
+        Stream.of(dataDirs)
+            .mapToLong(
+                dir -> {
+                  dir += File.separator + IoTDBConstant.UNSEQUENCE_FLODER_NAME;
+                  return FileUtils.getDirSize(dir);
+                })
+            .sum();
+    walFileTotalCount =
```
xingtanzjr, line 135: Let's use the new interface from WALNode to reduce calculation

xingtanzjr (approved):

PR #6916, opened 2022-08-08
[IOTDB-4056] Fix the problem that could not find or load main class in shell `remove-confignode.sh`

Beyyes:
## Description

see details in https://issues.apache.org/jira/browse/IOTDB-4056

The problem is that the conf and lib directory is different in `confignode` and `distribution`.

image

image

change codes are shown as below
image
---

- 1.1 start-confignode in confignode directory
image

- 1.2 remove-confignode in confignode directory
image

- 1.3 stop-confignode in confignode directory
image

- 2.1 start-confignode in distribution directory
image

- 2.2 remove-confignode in distribution directory
image

- 2.3 stop-confignode in distribution directory
image


qiaojialin (approved):

CRZbulabula (approved):
LGTM

PR #6923, opened 2022-08-09
[IOTDB-4038] Add the leader metrics to the cluster

23931017wu:
### After
image
image

PR #6924, opened 2022-08-09
[IOTDB-4072] Updated the ConfigNode Dashboard to add a display of node online and offline information and the distribution of leaders

23931017wu:
### After
image

For the presentation of these indicators, if the leader of the confignode appears in the timeframe displayed by the graffana, the indicators may appear duplicated on the panel. At present, there is no good solution, which can only allow users to avoid the time of leader transformation when using it. Since the utility of this data is still greater than the cost of adjusting the grafana time frame operation that the user may have, I personally feel that it can be shown in this way for the time being.

### DIFF
--- a/grafana-metrics-example/cluster/Apache IoTDB ConfigNode Dashboard v0.14.0.json
+++ b/grafana-metrics-example/cluster/Apache IoTDB ConfigNode Dashboard v0.14.0.json
@@ -418,7 +418,7 @@
           },
           "editorMode": "code",
           "exemplar": true,
-          "expr": "cluster_node_status{name =~ \"EndPoint.*\"}",
+          "expr": "cluster_node_status{}",
           "legendFormat": "{{instance}}  {{type}} {{name}}",
           "range": true,
           "refId": "A"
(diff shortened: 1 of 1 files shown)


Beyyes (commented):
@SpriCoder need your review~

SpriCoder:
Good Job, but I have one problem that is what is the meaning of "Select the node that you want to see"?

23931017wu:
> Good Job, but I have one problem that is what is the meaning of "Select the node that you want to see"?

The reason for this title is that the metric of the survival status of nodes is to describe whether they are online in terms of 0 and 1, so if there are multiple nodes on the graph, their polylines will coincide. So if you want to observe its survival status, you need to click on the properties of the corresponding node below, so that you can display only the polyline of the selected node.

SpriCoder:
Get it, but I think the title of panel should describe the meaning of the data in this panel instead of a operation description. So may be the title can be "The status of cluster node" and set a value mapping(0 -> "unknown", 1 -> "survive"), see: https://grafana.com/docs/grafana/latest/panels/format-data/

23931017wu:
> 



> Get it, but I think the title of panel should describe the meaning of the data in this panel instead of a operation description. So may be the title can be "The status of cluster node" and set a value mapping(0 -> "unknown", 1 -> "survive"), see: https://grafana.com/docs/grafana/latest/panels/format-data/

Okay, good idea ~

SpriCoder (commented):
LGTM after #6923 merged.

SpriCoder (approved):
LGTM~

Beyyes (approved):
LGTM

PR #6933, opened 2022-08-09
[IOTDB-4077] DataNode report region migrate result to FOLLOWER ConfigNode sometimes, the LEADER Confignode is still waiting

BanDuWuZi:
https://issues.apache.org/jira/browse/IOTDB-4077
1.when remove DataNode, the leader ConfigNode will send region migrate task to DataNode
2. if DN finished the region migrate task, it will report result to ConfigNode, and ConfigNode will contined.
but some times,  the DataNode report region migrate result to FOLLOWER ConfigNode, so the leader ConfigNode is still waiting


Beyyes (approved):
LGTM

wangchao316 (approved):
LGTM

PR #6936, opened 2022-08-09
Optimize MultileaderConsensus SyncLog

xingtanzjr:
> will add more description later
## Description
1. Do the serialization before putting the IndexedRequest to dispatching cache Queue. So that the Dispatcher won't consume lots of time when preparing the request from queue.
2. Fix a bug that sometimes the getBatch() will find request from WAL rather than Queue even if the Queue is not full
3. Implement the concurrent deserialization in follower side and use queue to keep write order.

Test:
image

We tested several scenarios for MultiLeader:
1. Shutdown follower and restart follower. No data lost
2. Shutdown leader and restart leader again. The `leader` will be changed to another peer. Previous leader will sync last batch to another 2 peers; the new leader will sync all new writes to another 2 peers (including previous leader)
3. After several shutdown and restart, we continue to run several BM. And finally, the data between these three peers are consistent


SpriCoder (approved):
LGTM~

SpriCoder (approved):
LGTM~

HTHou (approved):

PR #6949, opened 2022-08-10
[IOTDB-3938] Optimize Standalone schema fetch performance

MarcosZyk:
### Motivation
The mechanism of schema fetch in cluster mode is not that efficient in standalone mode, comparing with that of old standalone. The main cause is the redundant process on pattern tree and schema tree.

### Modification
1. Set up a data insertion specified ```DeviceGroupSchemaTree```, which only serves standalone data insertion.
2. Add an interface in schemaRegion to implement deviceInfo fetch and auto create.

The final cpu sample result is illustrated in following picture. The time consumption of schema validation decreased from 20% to only 2%.

image

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/mpp/common/schematree/MeasurementSchemaInfo.java
+++ b/server/src/main/java/org/apache/iotdb/db/mpp/common/schematree/MeasurementSchemaInfo.java
@@ -1,0 +1,47 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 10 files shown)


server/src/main/java/org/apache/iotdb/db/mpp/common/schematree/MeasurementSchemaInfo.java
```diff
@@ -0,0 +1,47 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iotdb.db.mpp.common.schematree;
+
+import org.apache.iotdb.tsfile.write.schema.MeasurementSchema;
+
+public class MeasurementSchemaInfo {
```
HTHou, line 24: Add some javadoc?

MarcosZyk: Fixed.

HTHou (approved):

Issue #6937, opened 2022-08-09
After restart, the aligned series turns to non-aligned in 0.13.0


DanielWang2035:
iotdb 0.13.0用Java原生接口创建对齐序列后，重启后对齐序列变成非对齐。

复现代码：

    SessionPool pool = null;
    try {
      pool = new SessionPool("127.0.0.1", 6667, "root", "root", 16);
      List<String> str_list = new ArrayList<>();
      str_list.add("at1");
      List<TSDataType> type_list = new ArrayList<>();
      type_list.add(TSDataType.FLOAT);
      List<TSEncoding> encoding_list = new ArrayList<>();
      encoding_list.add(TSEncoding.GORILLA);
      List<CompressionType> compression_type_list = new ArrayList<>();
      compression_type_list.add(CompressionType.SNAPPY);
      pool.createSchemaTemplate("t1", str_list, type_list, encoding_list, compression_type_list, true);
      pool.addAlignedMeasurementInTemplate("t1", "at2", TSDataType.FLOAT, TSEncoding.GORILLA, CompressionType.SNAPPY);
      pool.addAlignedMeasurementInTemplate("t1", "at3", TSDataType.FLOAT, TSEncoding.GORILLA, CompressionType.SNAPPY);
  
      pool.executeNonQueryStatement("set template t1 to root.SG1");
      pool.executeNonQueryStatement("create timeseries of schema template on root.SG1.a");
    } catch (IoTDBConnectionException | StatementExecutionException | IOException e) {
      e.printStackTrace();
    } finally {
      if (pool != null)
        pool.close();
    }

注意到代码中创建template时只有一个物理量，创建完后又增加了两个。此时`show devices`可见`isAligned=true`，但重启iotdb后再`show devices`，发现`isAligned=false`。
推测：iotdb读日志重建schema时，读到创建template的语句，发现只有一个物理量，会自动设置为非对齐。但实际上应该有三个物理量，改成非对齐会导致未知错误。

iotdb版本：0.13.0
操作系统：Windows10和Ubuntu均有此问题。

bigreybear:
Hi, thanks for your issue! I have put forward a PR trying to fix this, PLTA https://github.com/apache/iotdb/pull/6950

bigreybear:
Hi, the issue should have been fixed by the latest commit on rel/0.13, you can try it later.

PR #6950, opened 2022-08-10
[To rel/0.13][ISSUE-6937]Add alignment flag for single measurement template

bigreybear:
The root cause of #6937 is that the information of whether the template which has only one measurement inside is directly aligned is indeed regrettably lost within the former version of MLog. At the very begining, this information is supposed to be deduced by the structure of nested lists indicating measurement hierarchy, and the template with only one measurement was considered as a trivial case which  however occurs the issue today.

This PR added an extra constant flag `DIR_ALI_FLG` to indicate to indicate whether the template is directly aligned. The flag reuses bits which are used to indicate `schemaNames` are deprecated.  Considering the reason above and that the template with only one measurement is not that prevalent, the fix affects only few classes, while leaving other files untouched.


MarcosZyk (approved):
LGTM

PR #6955, opened 2022-08-11
[IOTDB-4093] Use online DataNodes to allocate new Regions

CRZbulabula:
We currently use registered DataNodes to allocate new Regions due to the influence of ConfigNodeRPCProcessorTest framework. Actually, we should use online DataNodes to do the Region allocation.


MiniSho:
LGTM

Beyyes (approved):

PR #6960, opened 2022-08-11
[IOTDB-3825] Reinforce Cluster IT framework's startup check

MiniSho:
Check Nodes' states by using showCluster, and the test fails when states are not "Running".

### DIFF
--- a/integration-test/src/main/java/org/apache/iotdb/it/env/AbstractEnv.java
+++ b/integration-test/src/main/java/org/apache/iotdb/it/env/AbstractEnv.java
@@ -211,7 +211,8 @@ public abstract class AbstractEnv implements BaseEnv {
             break;
           }
         }
-        if (flag) {
+        int nodeNum = configNodeWrapperList.size() + dataNodeWrapperList.size();
+        if (flag && nodeStatus.size() == nodeNum) {
           return;
         }
         Thread.sleep(1000);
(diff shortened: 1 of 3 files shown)


integration-test/src/main/java/org/apache/iotdb/it/env/AbstractEnv.java
```diff
@@ -185,12 +187,40 @@ public void testWorking() {
...
 
+  private void checkNodeHeartbeat() {
+    TShowClusterResp showClusterResp;
+    try (SyncConfigNodeIServiceClient client =
+        (SyncConfigNodeIServiceClient) getConfigNodeConnection()) {
+      boolean flag;
+      for (int i = 0; i < 30; i++) {
+        flag = true;
+        showClusterResp = client.showCluster();
+        Map<Integer, String> nodeStatus = showClusterResp.getNodeStatus();
```
CRZbulabula, line 207: We should also check the size of nodeStatus map

CRZbulabula (changes requested):
PTAL~

integration-test/src/main/java/org/apache/iotdb/it/env/AbstractEnv.java
```diff
@@ -185,12 +187,40 @@ public void testWorking() {
...
+            break;
+          }
+        }
+        if (flag) {
+          return;
+        }
+        Thread.sleep(1000);
+      }
+    } catch (Exception e) {
+      fail("Not all nodes are in the Running states");
```
Beyyes, line 220: If we need sleeping some time to wait that all the status of nodes have been running?

integration-test/src/main/java/org/apache/iotdb/it/env/AbstractEnv.java
MiniSho: done

Beyyes (approved):

CRZbulabula (approved):
LGTM~

PR #6962, opened 2022-08-11
[IOTDB-4111][IOTDB-3461] Improve data type infer and auto create schema logic in schema fetcher

MarcosZyk:
### Motivation
When the timeseries does exist in system, there's no need to infer data types of an InsertNode. The redundant type infer brings performance loss.

### Modification

Change the interface of schemaFetcher for schema validation. InsertNode provides a function object for schema fetcher, and it will only be invoked when auto creating schema.


HTHou (approved):
LGTM

PR #6985, opened 2022-08-14
Add Authorization parameters to confignode.properties

yifuzhou:
Authorization configuration parameters should also use in confignode.properties.


qiaojialin (commented):
Hi, as the Standalone version and cluster version both need this parameter, we could move such parameters into another properties file: iotdb-common.properties

qiaojialin (approved):

PR #6994, opened 2022-08-15
Optimize ml wal limitation.

SpriCoder:


consensus/src/main/java/org/apache/iotdb/consensus/multileader/MultiLeaderServerImpl.java
```diff
@@ -108,49 +115,59 @@ public void stop() {
...
-            getIndex(),
-            getCurrentSafelyDeletedSearchIndex());
-        try {
-          stateMachine.wait(config.getReplication().getThrottleTimeOutMs());
-        } catch (InterruptedException e) {
-          logger.error("Failed to throttle down because ", e);
-          Thread.currentThread().interrupt();
+    stateMachineLock.lock();
+    try {
+      synchronized (stateMachine) {
```
xingtanzjr, line 120: We can remove this `synchronized` here because we already lock the `stateMachineLock`. On the other hand, when current thread trigger the `await` of condition, it still hold the `stateMachine`'s object lock, which may lead unexpected behaviors.

PR #6998, opened 2022-08-15
[IOTDB-4121] logical optimizations for ImportCsv

2b3c511:


cli/src/main/java/org/apache/iotdb/tool/ImportCsv.java
```diff
@@ -829,7 +833,7 @@ private static Object typeTrans(String value, TSDataType type) {
           if (value.startsWith("\"") && value.endsWith("\"")) {
             return value.substring(1, value.length() - 1);
           }
-          return null;
+          return value;
```
HTHou, line 836: Update the user guides in https://github.com/apache/iotdb/blob/master/docs/UserGuide/Write-And-Delete-Data/CSV-Tool.md and https://github.com/apache/iotdb/blob/master/docs/zh/UserGuide/Write-And-Delete-Data/CSV-Tool.md .

PR #7002, opened 2022-08-15
[IOTDB-4138] Refactor consensus api and add javadoc

SzyWilliam:
Currently in IConsensus, addConsensusGroup and removeConsensusGroup is  ambiguous and often cause misunderstandings. We should change its name and add a javadoc to describe its usage and implementations. 
see https://issues.apache.org/jira/browse/IOTDB-4138

### DIFF
--- a/consensus/src/main/java/org/apache/iotdb/consensus/IConsensus.java
+++ b/consensus/src/main/java/org/apache/iotdb/consensus/IConsensus.java
@@ -50,7 +50,8 @@ public interface IConsensus {
    * group. This node will prepare and initialize local statemachine {@link IStateMachine} and other
    * data structures. After this method returns, we can call {@link #addPeer(ConsensusGroupId,
    * Peer)} to notify original group that this new Peer is prepared to be added into the latest
-   * configuration.
+   * configuration. createPeer should be called on a node that does not contain any peer of the
+   * consensus group, to avoid one node having more than one replica.
    *
    * @param groupId the consensus group this Peer belongs
    * @param peers other known peers in this group
@@ -61,7 +62,7 @@ public interface IConsensus {
    * When the <em>local node</em> is no longer a member of the given consensus group, call this
    * method to do cleanup works. This method will close local statemachine {@link IStateMachine},
    * delete local data and do other cleanup works. Be sure this method is called after successfully
-   * removing local peer from current consensus group configuration (by calling {@link
+   * removing this peer from current consensus group configuration (by calling {@link
    * #removePeer(ConsensusGroupId, Peer)} or {@link #changePeer(ConsensusGroupId, List)}).
    *
    * @param groupId the consensus group this Peer used to belong
@@ -75,6 +76,8 @@ public interface IConsensus {
    * #createPeer(ConsensusGroupId, List)} on the new Peer before calling this method. When this
    * method returns, the group data should be already transmitted to the new Peer. That is, the new
    * peer is available to answer client requests by the time this method successfully returns.
+   * addPeer should be called on a living peer of the consensus group. For example: We'd like to add
+   * a peer D to (A, B, C) group. We need to execute addPeer in A, B or C.
    *
    * @param groupId the consensus group this peer belongs
    * @param peer the newly added peer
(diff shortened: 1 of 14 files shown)


consensus/src/main/java/org/apache/iotdb/consensus/IConsensus.java
```diff
@@ -44,15 +44,61 @@ public interface IConsensus {
...
 
+  /**
+   * Tell the group to remove an active Peer. The removed peer can no longer answer group requests
+   * when this method successfully returns. Call {@link #deletePeer(ConsensusGroupId)} on the
+   * removed Peer to do cleanup jobs after this method successfully returns.
+   *
+   * @param groupId the consensus group this peer belongs
+   * @param peer the peer to be removed
+   */
   ConsensusGenericResponse removePeer(ConsensusGroupId groupId, Peer peer);
```
qiaojialin, line 92: removePeer should be called on a living peer of its consensus group.
For example: a group has A, B, C. We'd like to remove C, in case C is dead, the removePeer should be sent to A or B.

consensus/src/main/java/org/apache/iotdb/consensus/IConsensus.java
```diff
@@ -44,15 +44,61 @@ public interface IConsensus {
...
+  /**
+   * Tell the group that a new Peer is prepared to be added into this group. Call {@link
+   * #createPeer(ConsensusGroupId, List)} on the new Peer before calling this method. When this
+   * method returns, the group data should be already transmitted to the new Peer. That is, the new
+   * peer is available to answer client requests by the time this method successfully returns.
+   *
+   * @param groupId the consensus group this peer belongs
+   * @param peer the newly added peer
+   */
   ConsensusGenericResponse addPeer(ConsensusGroupId groupId, Peer peer);
```
qiaojialin, line 82: addPeer is called on a living peer of the consensus group.
For example: We'd like to add a peer D to (A, B, C) group. We need to execute addPeer in A, B or C.

consensus/src/main/java/org/apache/iotdb/consensus/IConsensus.java
```diff
@@ -44,15 +44,61 @@ public interface IConsensus {
...
+   * @param groupId the consensus group this Peer belongs
+   * @param peers other known peers in this group
+   */
+  ConsensusGenericResponse createPeer(ConsensusGroupId groupId, List<Peer> peers);
+
+  /**
+   * When the <em>local node</em> is no longer a member of the given consensus group, call this
+   * method to do cleanup works. This method will close local statemachine {@link IStateMachine},
+   * delete local data and do other cleanup works. Be sure this method is called after successfully
+   * removing local peer from current consensus group configuration (by calling {@link
```
qiaojialin, line 64:
```suggestion
   * removing this peer from current consensus group configuration (by calling {@link
```

consensus/src/main/java/org/apache/iotdb/consensus/IConsensus.java
```diff
@@ -44,15 +44,61 @@ public interface IConsensus {
...
+   * Require the <em>local node</em> to create a Peer and become a member of the given consensus
+   * group. This node will prepare and initialize local statemachine {@link IStateMachine} and other
+   * data structures. After this method returns, we can call {@link #addPeer(ConsensusGroupId,
+   * Peer)} to notify original group that this new Peer is prepared to be added into the latest
+   * configuration.
+   *
+   * @param groupId the consensus group this Peer belongs
+   * @param peers other known peers in this group
+   */
+  ConsensusGenericResponse createPeer(ConsensusGroupId groupId, List<Peer> peers);
```
qiaojialin, line 58: createPeer should be called on a node that does not contain any peer of the consensus group, to avoid one node having more than one replica.

consensus/src/main/java/org/apache/iotdb/consensus/IConsensus.java
SzyWilliam: done

consensus/src/main/java/org/apache/iotdb/consensus/IConsensus.java
SzyWilliam: done

consensus/src/main/java/org/apache/iotdb/consensus/IConsensus.java
SzyWilliam: done

consensus/src/main/java/org/apache/iotdb/consensus/IConsensus.java
SzyWilliam: done

qiaojialin (approved):

PR #7006, opened 2022-08-16
[IOTDB-4092] Protecting Region creation process by CreateRegionGroupsProcedure

CRZbulabula:
There is a potential bug within the creation process of RegionGroups. That is, the ConfigNode-leader always chooses some online DataNodes as the holder of new RegionReplicas. However, if a DataNode was shutdown suddenly and there occurs a requirement for creating new RegionGroups in the same time, the DataNode that just been shutdown recently will still have the probability to be chosen as the holder of the new RegionReplica. Because it takes 20s for the ConfigNode-leader to mark that DataNode as "UNKNOWN" status.

Therefore, I use the Procedure framework to create the CreateRegionGroupsProcedure to prevent this potential bug:

1. A batch of RegionGroups are considered as created successfully only if they are all physically created on those chosen DataNodes.
2. As long as there exists one RegionReplica failed to create, this batch of RegionGroups will be considered as fail to create. In the same time, the corresponding getOrCreatePartition request will be rejected and receive error code "CREATE_REGION_FAIL". Finally, these redundant ReionReplicas that fail to construct new RegionGroups will be physically removed asynchronously.

You can see the [Partition doc](https://apache-iotdb.feishu.cn/docs/doccnzlKpkxhuRw52P9pp6SfELf#kJavEE) and the [RegionGroup management doc](https://apache-iotdb.feishu.cn/docx/doxcnMQppTJCURUuWZ5LfWA5dK6) for more details.

========== Local Test ==========

By the way, I'm starting a 1C4D local environment and doing the following tests:

image

1. Shutdown the DataNode on port 9006 and insert some new data immediately(in less than 20s):

image

The error message shown in Cli is because the ConfigNode-leader pick the DataNode on port 9006 as the holder of one RegionReplica, and the creation process of that RegionReplica will fail obviously.

3. Insert some new data after the DataNode on port 9006 was shutdown more than 20s:

image

image

The index of the RegionGroups is started from 2 because the former RegionGroups were failed to create.

4. Check the RegionCleaner related logs:

image

The RegionCleaner deletes the redundant RegionReplicas successfully.


confignode/src/main/java/org/apache/iotdb/confignode/client/sync/datanode/SyncDataNodeClientPool.java
```diff
@@ -134,14 +134,21 @@ private void deleteRegions(
...
       final TSStatus status =
           sendSyncRequestToDataNodeWithRetry(
               endPoint, regionId, DataNodeRequestType.DELETE_REGIONS);
+
       if (status.getCode() == TSStatusCode.SUCCESS_STATUS.getStatusCode()) {
-        LOGGER.info("DELETE Region {} successfully", regionId);
-        deletedRegionSet.removeIf(k -> k.getRegionId().equals(regionId));
+        LOGGER.info("Delete RegionReplica: {} on DataNode: {} successfully", regionId, endPoint);
+      } else {
+        LOGGER.info(
```
wangchao316, line 145: warn level

confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/CreateRegionGroupsProcedure.java
```diff
@@ -0,0 +1,133 @@
...
+import org.apache.iotdb.common.rpc.thrift.TRegionReplicaSet;
+import org.apache.iotdb.confignode.consensus.request.write.CreateRegionGroupsPlan;
+import org.apache.iotdb.confignode.consensus.request.write.DeleteRegionGroupsPlan;
+import org.apache.iotdb.confignode.procedure.env.ConfigNodeProcedureEnv;
+import org.apache.iotdb.confignode.procedure.exception.ProcedureException;
+import org.apache.iotdb.confignode.procedure.state.CreateRegionGroupsState;
+
+import java.util.Map;
+
+public class CreateRegionGroupsProcedure extends AbstractNodeProcedure<CreateRegionGroupsState> {
```
wangchao316, line 31: should extends StateMachineProcedure. see :“StorageGroupSnapshotProcedure”
because AbstractNodeProcedure has nodeLock. all node operate one by one.

wangchao316 (changes requested):

confignode/src/main/java/org/apache/iotdb/confignode/client/sync/datanode/SyncDataNodeClientPool.java
CRZbulabula: Fixed~

confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/CreateRegionGroupsProcedure.java
CRZbulabula: Fixed~

wangchao316 (approved):
LGTM

Beyyes (approved):

PR #7007, opened 2022-08-16
[IOTDB-3938] Fix SeriesSlotExecutor initialization logic for insertion performance

MarcosZyk:
### Motivation

The initialization of SeriesSlotExecutor in Partition seriously affects the insertion performance, causing 6% performance loss in new standalone comparing with that in old standalone.

### Modification

Since the param of SeriesSlotExecutor is immutable during system running and it is stateless, the SeriesSlotExecutor is implemented as a singleton.

Remove the ```isClusterMode``` in ```IoTDB.class``` and use IoTDBConfig to store the state instead. Decouple DataNode.class from IoTDB.class


CRZbulabula (approved):
LGTM～

HTHou (approved):

PR #7008, opened 2022-08-16
[IOTDB-4140] Mark the datanode as removing status when execute remove-datanode.sh

Beyyes:
## Description

When executed remove-datanode.sh TARGET_NODE, the node status of TARGET_NODE should be marked as `REMOVING`,  and the LatestRegionRouteMap should be broadcasted to assure that no more read/write requests are forwarded to this datanode.

Running screenshots

1). 1 confignode and 4 datanodes.
image
image

2). after execute remove-datanode.sh 127.0.0.1:6669

In `REMOVE_DATA_NODE_PREPARE` state of `RemoveDataNodeProcedure`,
the region map order is [3,1,2] rather than [2,3,1]
image

When remove-datanode.sh is executed completely.
image
image

3) The removing status of show cluster.
image

The mainly changed are this method.
image

TODO：
change the datanode status when RemoveDataNodeProcedure is completed


wangchao316 (approved):
LGTM

confignode/src/main/java/org/apache/iotdb/confignode/manager/NodeManager.java
```diff
@@ -569,6 +571,23 @@ public Map<Integer, Long> getAllLoadScores() {
...
+  public boolean isNodeRemoving(int dataNodeId) {
+    DataNodeHeartbeatCache cache =
+        (DataNodeHeartbeatCache) configManager.getNodeManager().getNodeCacheMap().get(dataNodeId);
+    if (cache != null) {
+      return cache.isRemoving();
+    }
+    return false;
+  }
+
+  public void setNodeRemoving(int dataNodeId) {
```
wangchao316, line 583: TODO: add a interface,  set node removing is false.   
if  procedure failed and rollback.

confignode/src/main/java/org/apache/iotdb/confignode/manager/PartitionManager.java
```diff
@@ -544,8 +548,12 @@ public Map<TConsensusGroupId, Integer> getAllLeadership() {
                       regionReplicaSet.getDataNodeLocations().get(0).getDataNodeId()));
     } else {
       regionGroupCacheMap.forEach(
-          (consensusGroupId, regionGroupCache) ->
-              result.put(consensusGroupId, regionGroupCache.getLeaderDataNodeId()));
+          (consensusGroupId, regionGroupCache) -> {
+            int leaderDataNodeId = regionGroupCache.getLeaderDataNodeId();
+            if (!configManager.getNodeManager().isNodeRemoving(leaderDataNodeId)) {
+              result.put(consensusGroupId, leaderDataNodeId);
+            }
```
CRZbulabula, line 555: We'd better put -1 instead of nothing when the leaderNode happens to be in the status of Removing. This might avoid NPE.

CRZbulabula (changes requested):
PTAL. Besides the comments above, we should better special treat the updateLoadStatistic() operation on the Removing DataNode. Because they might trigger newer broadcast of  RegionRouteMap, which will replace previous broadcast map.

confignode/src/main/java/org/apache/iotdb/confignode/manager/load/heartbeat/DataNodeHeartbeatCache.java
```diff
@@ -22,27 +22,44 @@
...
+ * DataNodeHeartbeatCache caches and maintains all the heartbeat data
+ *
+ * <p>TODO: This class might be split into DataNodeCache and ConfigNodeCache
+ */
 public class DataNodeHeartbeatCache implements INodeCache {
 
-  // TODO: This class might be split into DataNodeCache and ConfigNodeCache
+  // when the response time of heartbeat is more than 20s,
+  // the datanode is considered as down
+  private static final int HEARTBEAT_TIMEOUT_TIME = 20_000;
```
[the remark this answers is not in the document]
Beyyes: fixed. and I think changing INodeCache from `Interface` to `Abstract class` is better. Because some variables are common used.

confignode/src/main/java/org/apache/iotdb/confignode/manager/PartitionManager.java
```diff
@@ -531,7 +532,10 @@ public Map<TConsensusGroupId, Integer> getAllLeadership() {
       regionGroupCacheMap.forEach(
           (consensusGroupId, regionGroupCache) -> {
             if (consensusGroupId.getType().equals(TConsensusGroupType.SchemaRegion)) {
-              result.put(consensusGroupId, regionGroupCache.getLeaderDataNodeId());
+              int leaderDataNodeId = regionGroupCache.getLeaderDataNodeId();
+              if (!configManager.getNodeManager().isNodeRemoving(leaderDataNodeId)) {
+                result.put(consensusGroupId, leaderDataNodeId);
+              }
```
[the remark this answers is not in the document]
Beyyes: fixed

confignode/src/main/java/org/apache/iotdb/confignode/manager/PartitionManager.java
Beyyes: fixed

CRZbulabula (approved):
LGTM~

PR #7011, opened 2022-08-16
[IOTDB-4132] Fix concurrent bug caused by sharing same DatasetHeader

liuminghui233:
see [IOTDB-4132](https://issues.apache.org/jira/browse/IOTDB-4132).


server/src/main/java/org/apache/iotdb/db/mpp/common/header/ColumnHeaderConstant.java
```diff
@@ -0,0 +1,218 @@
...
+
+  // column names for show datanodes
+  public static final String COLUMN_DATA_REGION_NUM = "DataRegionNum";
+  public static final String COLUMN_SCHEMA_REGION_NUM = "SchemaRegionNum";
+
+  // column names for show schema template statement
+  public static final String COLUMN_TEMPLATE_NAME = "template name";
+
+  public static final List<ColumnHeader> lastQueryColumnHeaders =
+      Arrays.asList(
```
JackieTien97, line 92: Change it to ImmutableList

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/mpp/common/header/ColumnHeaderConstant.java
liuminghui233: fixed

JackieTien97 (approved):

PR #7012, opened 2022-08-16
Optimize ml WAL limitation for MultiLeader

xingtanzjr:
## Description
Use a simple parameter to throttle the write when WAL size reaches the threshold when using MultiLeaderConsensus


consensus/src/main/java/org/apache/iotdb/consensus/multileader/MultiLeaderServerImpl.java
```diff
@@ -238,13 +253,20 @@ public MultiLeaderConfig getConfig() {
   }
 
   public boolean needToThrottleDown() {
```
qiaojialin, line 255: needBlock

consensus/src/main/java/org/apache/iotdb/consensus/multileader/MultiLeaderServerImpl.java
```diff
@@ -238,13 +253,20 @@ public MultiLeaderConfig getConfig() {
...
-    return reader.getTotalSize() > config.getReplication().getThrottleDownThreshold();
+    return reader.getTotalSize() > config.getReplication().getWalThrottleThreshold();
   }
 
   public boolean needToThrottleUp() {
```
qiaojialin, line 259: unblockWrite

qiaojialin (approved):

PR #7017, opened 2022-08-16
[IOTDB-4133] Throw illegal path exception contains number which has been surrounded with quote

lancelly:
More details: https://issues.apache.org/jira/browse/IOTDB-4133


tsfile/src/main/java/org/apache/iotdb/tsfile/read/common/parser/PathVisitor.java
```diff
@@ -64,12 +64,28 @@ private String parseNodeName(PathParser.NodeNameContext ctx) {
...
+    int index = 0;
+    // remove zeros
+    for (int i = 0, n = str.length(); i < n; i++) {
+      if (str.charAt(i) != '0') {
+        index = i;
+        break;
+      }
+    }
+    return NumberUtils.isCreatable(str.substring(index));
+  }
```
JackieTien97, line 90: duplicated with previous, and it seems that `+0001` and `-00001` won't be considered

JackieTien97 (changes requested):

tsfile/src/main/java/org/apache/iotdb/tsfile/read/common/parser/PathVisitor.java
lancelly: Fixed.

PR #7018, opened 2022-08-16
[IOTDB-4150] Client output has double error code

yifuzhou:
[IOTDB-4150](https://issues.apache.org/jira/browse/IOTDB-4150)
If use StatementExecutionException in future.setException(), it will combine message and code as a new message, when client end outputs the exception, the client will use StatementExecutionException again so there are two error codes have been combined.
Before fix:
image

After fix:
image


yifuzhou:
@qiaojialin @MarcosZyk @SpriCoder Can you review it? Thanks!

PR #7021, opened 2022-08-16
[IOTDB-4061] Support hot configuration of data_dirs

HeimingZ:
To support hot configuration of data_dirs, I support load configuration command in new cluster.


SpriCoder:
Please check wether file metrics is valid after hot load.

HeimingZ:
> Please check wether file metrics is valid after hot load.

I move dataDirs and walDirs into the collect method, so both params will be updated at each invocation.

qiaojialin (approved):

PR #7022, opened 2022-08-16
[IOTDB-3938]Fix measurements check of batch insertion

MarcosZyk:
### Motivation

The wrong use of measurements check results serious performance loss in new standalone, only 50% of that of old standalone.
The wrong method causes redundant and unnecessary path split. 

### Modification

Fix the measurements check methods.

#### Before fix
image

#### After fix
image


lancelly (approved):
LGTM

HTHou (approved):

PR #7031, opened 2022-08-17
[IOTDB-4084] remove DataNode with exit code

BanDuWuZi:
https://issues.apache.org/jira/browse/IOTDB-4084
remove datanode by shell : remove-datanode.sh, it implements in DataNodeServerCommandLine, the class has no exit code ,so I can not use '$?' to show if the remove datanode request is accept or not

'$?' is linux shell command, it can get the last command result, if last command run succeed, '$?' is 0, if  failed ,'$?' is not 0
ex:

A. list a exit path
[root@host-XXX iotdbserver]# ls /tmp/11
/tmp/11
[root@host-XXX iotdbserver]# echo $?
0

B. list a nonexit path
[root@host-XXX iotdbserver]# ls /tmp/tmp/tmp
ls: cannot access /tmp/tmp/tmp: No such file or directory
[root@host-XXX iotdbserver]# echo $?
2

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/service/DataNodeServerCommandLine.java
+++ b/server/src/main/java/org/apache/iotdb/db/service/DataNodeServerCommandLine.java
@@ -24,6 +24,7 @@ import org.apache.iotdb.common.rpc.thrift.TEndPoint;
 import org.apache.iotdb.commons.ServerCommandLine;
 import org.apache.iotdb.commons.exception.BadNodeUrlException;
 import org.apache.iotdb.commons.exception.ConfigurationException;
+import org.apache.iotdb.commons.exception.IoTDBException;
 import org.apache.iotdb.commons.utils.NodeUrlUtils;
 import org.apache.iotdb.confignode.rpc.thrift.TDataNodeConfigurationResp;
 import org.apache.iotdb.confignode.rpc.thrift.TDataNodeRemoveReq;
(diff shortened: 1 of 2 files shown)


wangchao316 (approved):
LGTM

CRZbulabula (approved):
LGTM～

server/src/main/java/org/apache/iotdb/db/service/DataNodeServerCommandLine.java
```diff
@@ -158,8 +154,10 @@ private void removeNodesFromCluster(String[] args) throws BadNodeUrlException {
     try (ConfigNodeClient configNodeClient = new ConfigNodeClient()) {
       TDataNodeRemoveResp removeResp = configNodeClient.removeDataNode(removeReq);
       logger.info("Remove result {} ", removeResp.toString());
-    } catch (TException e) {
-      logger.error("send remove Data Node request failed!", e);
+      if (removeResp.getStatus().getCode() != TSStatusCode.SUCCESS_STATUS.getStatusCode()) {
+        throw new IoTDBException(
+            removeResp.getStatus().toString(), removeResp.getStatus().getCode());
```
Beyyes, line 159: If it's better to set the error message `send remove Data Node request failed!` in IoTDBException

Beyyes (approved):

server/src/main/java/org/apache/iotdb/db/service/DataNodeServerCommandLine.java
BanDuWuZi: ConfigNode returned status has more detail message,  In my test case, it is :

org.apache.iotdb.commons.exception.IoTDBException: TSStatus(code:916, message:lack replication, allow most removed Data Node size : 0)
	at org.apache.iotdb.db.service.DataNodeServerCommandLine.removeNodesFromCluster(DataNodeServerCommandLine.java:224)

PR #7044, opened 2022-08-18
[IOTDB-3855] Roll wal file when necessary

HeimingZ:
Roll wal file when current file or last file is not empty.


xingtanzjr:
I will talk with Haiming for this change later to check whether this change is necessary.

Today we are focusing on fixing the bug in WAL, which makes data lost for follower when restarting

xingtanzjr:
This PR should not be merged because this fix is not used to fix the root cause of the issue, even if the investigation of root cause is still in progress

PR #7046, opened 2022-08-18
update user-guide and grant ttl operation to SET_STORAGE_GROUP privilege

yifuzhou:
Hi, privileges of schema template and count have already implemented, so just remove it from user-guide, ttl will bind to set_storage_group because it always use with setting storage group.


RYH61 (approved):
LGTM~

MarcosZyk (approved):
LGTM

PR #7049, opened 2022-08-18
[IOTDB-4020] Use canonical path when recovering wal

HeimingZ:
This error is caused by using absolute path because it contains "." and "..", making it not unique.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/wal/recover/WALNodeRecoverTask.java
+++ b/server/src/main/java/org/apache/iotdb/db/wal/recover/WALNodeRecoverTask.java
@@ -199,9 +199,8 @@ public class WALNodeRecoverTask implements Runnable {
     for (MemTableInfo memTableInfo : memTableId2Info.values()) {
       firstValidVersionId = Math.min(firstValidVersionId, memTableInfo.getFirstFileVersionId());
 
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/iotdb/db/wal/recover/WALNodeRecoverTask.java
```diff
@@ -199,9 +199,8 @@ private void recoverInfoFromCheckpoints() {
     for (MemTableInfo memTableInfo : memTableId2Info.values()) {
       firstValidVersionId = Math.min(firstValidVersionId, memTableInfo.getFirstFileVersionId());
 
-      File tsFile = new File(memTableInfo.getTsFilePath());
       UnsealedTsFileRecoverPerformer recoverPerformer =
-          walRecoverManger.removeRecoverPerformer(tsFile.getAbsolutePath());
+          walRecoverManger.removeRecoverPerformer(new File(memTableInfo.getTsFilePath()));
```
JackieTien97, line 203: You should change the filePath in MemtableInfo to canonical path which is in `TsFileResource.getTsFilePath()`.

JackieTien97 (changes requested):
Don't forget to check whether 0.13 has this problem or not.

JackieTien97 (approved):

PR #7057, opened 2022-08-19
[IOTDB-4125] Clear NodeCache and RegionGroupCache in time

CRZbulabula:


confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/RemoveConfigNodeProcedure.java
```diff
@@ -72,6 +68,7 @@ protected Flow executeFromState(ConfigNodeProcedureEnv env, RemoveConfigNodeStat
           LOG.info("Remove Consensus Group {}", tConfigNodeLocation);
           break;
         case STOP_CONFIG_NODE:
+          env.broadCastTheLatestConfigNodeGroup();
```
wangchao316, line 71: why ?  We should broadcast and quarantine before we delete it.

CRZbulabula: The old framework broadcasts as soon as the RemoveConfigNodeProcedure begins, which means the broadcasted ConfigNodeList still contain the ConfigNode to be removed. Therefore, we should broadcast the ConfigNodeList after we actually remove this ConfigNode.

wangchao316 (approved):
LGTM

Beyyes (approved):

PR #7081, opened 2022-08-22
[IOTDB-4202]fix delete from sql IT and user-guide

yifuzhou:
After discussion, we think we should return some warn messages and still let it return success.
 
If this way works, I think we should consider some pervious error codes and let these return success but with warn message because of this rule:

"It should be noted that when the deleted path does not exist, IoTDB will not prompt that the path does not exist, but that the execution is successful, because SQL is a declarative programming method. Unless it is a syntax error, insufficient permissions and so on, it is not considered an error".

Thanks!

previous:
IoTDB> delete from root.sgcc.wf01.wt01.ts0
IoTDB> delete from root.sgcc.wf01.wt01.ts0abcefg

after fix:
image

### DIFF
--- a/docs/UserGuide/Write-And-Delete-Data/Delete-Data.md
+++ b/docs/UserGuide/Write-And-Delete-Data/Delete-Data.md
@@ -62,7 +62,7 @@ expressions like : time > XXX, time <= XXX, or two atomic expressions connected 
 If no "where" clause specified in a delete statement, all the data in a timeseries will be deleted.
 
 ```sql
-delete from root.ln.wf02.status
+delete from root.ln.wf02.wt02.status
 ```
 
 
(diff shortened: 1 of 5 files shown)


wangchao316 (approved):
LGTM

MarcosZyk (approved):

wangchao316 (changes requested):
需要注意的是，当删除的路径不存在时，IoTDB 不会提示路径不存在，而是显示执行成功，因为 SQL 是一种声明式的编程方式，除非是语法错误、权限不足等，否则都不认为是错误，如下所示。

Need to discuss

yifuzhou:
@qiaojialin can you review it? Should we let all delete sqls return successfully or give an error code if paths do not exist?

qiaojialin:
Hi, as no user raise this issue, we'd better keep the notation consistent with 0.13 version, just return success is fine.

qiaojialin (changes requested):
Hi, there is no detailed requirements, won't do now.

wangchao316 (approved):
LGTM

qiaojialin (approved):

PR #7087, opened 2022-08-23
[IOTDB-4210] Fix deadlock during WALNode recover

ericpai:
See JIRA: https://issues.apache.org/jira/browse/IOTDB-4210

If there's no suitable walFiles to recover, the recover task will end without notifying the listeners.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/wal/recover/WALNodeRecoverTask.java
+++ b/server/src/main/java/org/apache/iotdb/db/wal/recover/WALNodeRecoverTask.java
@@ -226,6 +226,7 @@ public class WALNodeRecoverTask implements Runnable {
                 WALFileUtils.walFilenameFilter(dir, name)
                     && WALFileUtils.parseVersionId(name) >= firstValidVersionId);
     if (walFiles == null) {
+      endRecovery();
       return;
     }
     // asc sort by version id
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/apache/iotdb/db/wal/recover/WALNodeRecoverTask.java
```diff
@@ -226,6 +226,9 @@ private void recoverTsFiles() {
                 WALFileUtils.walFilenameFilter(dir, name)
                     && WALFileUtils.parseVersionId(name) >= firstValidVersionId);
     if (walFiles == null) {
+      for (UnsealedTsFileRecoverPerformer recoverPerformer : memTableId2RecoverPerformer.values()) {
+        recoverPerformer.getRecoverListener().succeed();
+      }
```
HeimingZ, line 231: Thank you for the bug fix, but UnsealedTsFileRecoverPerformer.endRecovery should be called before return success.

```suggestion
      for (UnsealedTsFileRecoverPerformer recoverPerformer : memTableId2RecoverPerformer.values()) {
      try {
        recoverPerformer.endRecovery();
        recoverPerformer.getRecoverListener().succeed();
      } catch (Exception e) {
        recoverPerformer.getRecoverListener().fail(e);
      }
    }
```

ericpai: Fixed, with a little of refactors.

HeimingZ: LGTM 👍

HeimingZ (approved):

PR #7091, opened 2022-08-23
[IOTDB-4210] Fix deadlock during WALNode recover if there's another one already running

ericpai:
Here's the error log
image


HeimingZ (approved):
LGTM

PR #7112, opened 2022-08-24
[IOTDB-4222] Create data region repeatedly when concurrently deleting and creating storage groups

HeimingZ:
Synchronize ready condition rather than MNode. Besides, StorageEngineV2 doesn‘t have this problem.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/dataregion/StorageGroupManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/dataregion/StorageGroupManager.java
@@ -158,14 +158,15 @@ public class StorageGroupManager {
   }
 
   @SuppressWarnings("java:S2445")
-  // actually storageGroupMNode is a unique object on the mtree, synchronize it is reasonable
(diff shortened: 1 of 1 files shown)


HTHou (approved):

ericpai:
@HeimingZ Are there any other code blocks that should be changed as well? If one code uses synchronized keyword, then there must be at least another one.

HeimingZ:
> @HeimingZ Are there any other code blocks that should be changed as well? If one code uses synchronized keyword, then there must be at least another one.

@ericpai Just block here is ok, this method is the only interface to get data region. For example, when two threads hold different MNode to get the same data region, they will block at the same ready condition.

ericpai:
> > @HeimingZ Are there any other code blocks that should be changed as well? If one code uses synchronized keyword, then there must be at least another one.
> 
> @ericpai Just block here is ok, this method is the only interface to get data region. For example, when two threads hold different MNode to get the same data region, they will block at the same ready condition.

Thanks. LGTM!

ericpai (approved):

PR #7115, opened 2022-08-24
[IOTDB-4185] Prevent RatisConsensus Start Race

SzyWilliam:
Race condition detected as follows:
1. PartitionRegionStateMachine is initialized.
2. ConsensusManager constructor is called, but not returned.
3. Inside ConsensusManager ctor, RaftServer is initialized in ConsensusManager and start election.
4. Inside ConsensusManager ctor, RaftServer is elected and notifies PartitionRegionStateMachine.
5. PartitionRegionStateMachine starts heartbeat service, which uses ConsensusManager.
6. However, in Step5, ConsensusManager Constructor is not returned, which means the ref is null. It can cause NPE

### DIFF
--- a/confignode/src/main/java/org/apache/iotdb/confignode/manager/NodeManager.java
+++ b/confignode/src/main/java/org/apache/iotdb/confignode/manager/NodeManager.java
@@ -68,6 +68,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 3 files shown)


confignode/src/main/java/org/apache/iotdb/confignode/manager/NodeManager.java
```diff
@@ -432,14 +433,18 @@ public void startHeartbeatService() {
...
-    if (getConsensusManager().isLeader()) {
-      // Generate HeartbeatReq
-      THeartbeatReq heartbeatReq = genHeartbeatReq();
-      // Send heartbeat requests to all the registered DataNodes
-      pingRegisteredDataNodes(heartbeatReq, getRegisteredDataNodes());
-      // Send heartbeat requests to all the registered ConfigNodes
-      pingRegisteredConfigNodes(heartbeatReq, getRegisteredConfigNodes());
-    }
+    Optional.ofNullable(getConsensusManager())
+        .ifPresent(
```
CRZbulabula, line 437: We'd better add annotation about why should we must use Optional.ofNullable.ifPresent here.

confignode/src/main/java/org/apache/iotdb/confignode/manager/PartitionManager.java
```diff
@@ -496,19 +497,23 @@ public String getRegionStorageGroup(TConsensusGroupId regionId) {
...
-            deletedRegionSet.stream()
-                .map(TRegionReplicaSet::getRegionId)
-                .collect(Collectors.toList()));
-        deletedRegionSet.forEach(
-            regionReplicaSet -> removeRegionGroupCache(regionReplicaSet.regionId));
-        SyncDataNodeClientPool.getInstance().deleteRegions(deletedRegionSet);
-      }
-    }
+    Optional.ofNullable(getConsensusManager())
+        .ifPresent(
```
CRZbulabula, line 501: We'd better add annotation about why should we must use Optional.ofNullable.ifPresent here.

CRZbulabula (commented):
PTAL~

confignode/src/main/java/org/apache/iotdb/confignode/manager/NodeManager.java
SzyWilliam: Done

confignode/src/main/java/org/apache/iotdb/confignode/manager/PartitionManager.java
SzyWilliam: Done

CRZbulabula (approved):
LGTM~

PR #7126, opened 2022-08-25
[IOTDB-4235] Fix deadlock in cancel query and timeout detect thread

JackieTien97:
details can be seen in this [feishu](https://apache-iotdb.feishu.cn/docx/doxcnbUyfLosIQhnjn6kQAOceCe) doc


xingtanzjr (approved):

xingtanzjr:
This fix also explains the issue that lots of write operations are suspended, which leads to increase of InternalServiceRPC threads. 

The direct cause is that all threads in pool `MPPCoordinator` are occupied when the stateMachine revoking `releaseResource()`. They are blocked when invoking SourceHandle.abort(), which is caused by the issue in this PR

ericpai:
Good catch!

qiaojialin (approved):

PR #7133, opened 2022-08-25
[IOTDB-4114] Add safety check for createPeer

SzyWilliam:
When ConfigNode or DataNode calls `Consensus.createPeer` twice with exact groupID, `RatisConsensus` will allow the second `createPeer` to be executed. This will cause unexpected errors, as overlapping files could interfere each other.
In this PR I propose to add a pre-condition check in `createPeer` so that the same Peer will not be created twice on one node.

### DIFF
--- a/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
+++ b/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
@@ -38,6 +38,7 @@ import org.apache.iotdb.consensus.common.response.ConsensusReadResponse;
 import org.apache.iotdb.consensus.common.response.ConsensusWriteResponse;
 import org.apache.iotdb.consensus.config.ConsensusConfig;
 import org.apache.iotdb.consensus.exception.ConsensusException;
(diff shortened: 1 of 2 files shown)


consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
```diff
@@ -256,6 +257,12 @@ public ConsensusGenericResponse createPeer(ConsensusGroupId groupId, List<Peer>
     if (!group.getPeers().contains(myself)) {
       return failed(new ConsensusGroupNotExistException(groupId));
     }
+    RaftGroupId raftGroupId = Utils.fromConsensusGroupIdToRaftGroupId(groupId);
+    RaftGroup existGroup = getGroupInfo(raftGroupId);
+    // pre-conditions: the new group not exists
+    if (existGroup != null) {
+      return failed(new ConsensusGroupAlreadyExistException(groupId));
+    }
```
JackieTien97, line 265: It seems that even if you do this check, it could not totally prevent this problem because this checking and creating are not atomically.

SzyWilliam:
This Pr should be close because the issue4114 is caused by starting confignode twice on one single machine.

PR #7134, opened 2022-08-25
[IOTDB-2760][IOTDB-2769][IOTDB-3302] Fix some auth problems and add Auth IT test.

SpriCoder:
1. fix when user try to grant role to user when have privileges.
2. fix the check of query in standalone
3. add auth IT test.


RYH61 (approved):
LGTM~

qiaojialin (approved):

PR #7136, opened 2022-08-25
[IOTDB-3455] Make data_region_num takes effect in new standalone

THUMarkLau:
See [IOTDB-3455](https://issues.apache.org/jira/browse/IOTDB-3455).


HTHou:
Spotless...

server/src/main/java/org/apache/iotdb/db/localconfignode/LocalConfigNode.java
```diff
@@ -141,7 +141,7 @@ public class LocalConfigNode {
 
   private final StorageEngineV2 storageEngine = StorageEngineV2.getInstance();
 
-  private final LocalDataPartitionTable dataPartitionTable = LocalDataPartitionTable.getInstance();
+  private final LocalDataPartitionInfo dataPartitionTable = LocalDataPartitionInfo.getInstance();
```
xingtanzjr, line 144: Align the variable table with class name

server/src/main/java/org/apache/iotdb/db/localconfignode/DataRegionIdGenerator.java
```diff
@@ -0,0 +1,46 @@
...
+ * under the License.
+ */
+
+package org.apache.iotdb.db.localconfignode;
+
+import org.apache.iotdb.commons.utils.TestOnly;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class DataRegionIdGenerator {
```
xingtanzjr, line 26: Maybe we can put this class as a private static class inner LocalDataPartitionInfo as it is only used there

server/src/main/java/org/apache/iotdb/db/localconfignode/LocalDataPartitionInfo.java
```diff
@@ -0,0 +1,110 @@
...
+import org.apache.iotdb.commons.path.PartialPath;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+// This class is used for data partition maintaining the map between storage group and
+// dataRegionIds.
+public class LocalDataPartitionInfo {
```
xingtanzjr, line 33: Why do we split the original LocalDataPartitionTable into two classes `Info` and `Table` ?

These two classes are a litter bit confusing because their names are similar

THUMarkLau: In IoTDB ConfigNode, the DataPartitionTable and PartitionInfo classes are used to record the data partition, so this design is borrowed in LocalConfigNode. Table is global and records the mapping between a logical storage group and PartitionInfo. Info is one for each logical storage group and records the mapping between each device in the logical storage group and the DataRegion.

xingtanzjr (approved):

PR #7137, opened 2022-08-26
idTable supports auto-increment id, in the stand-alone environment

KeeProMise:
## Description
provide a new deviceID, use a long variable composed of schemaRegion+int_32 instead of devicePath, which can easily build an index for the device and reduce the storage space occupied by storing long devicePath strings

## Different
-  **DeviceIDFactory** : add the getDeviceIDWithAutoCreateFunction method to separate the writing and reading of deviceID, because writing requires a new deviceID object to be generated when a device is added for the first time, but reading does not need it. Continuing to use the same method cannot distinguish between reading and writing.
-   **DeviceIDFactory**：add getDeviceIDWithRecoverFunction method for recovery when the system restarts
-  **StandAloneAutoIncDeviceID** :  using auto-incrementing id as device id,A complete auto-increment id consists of schemaRegionID and autoIncrementID, where the upper 32 bits are schemaRegionID and the lower 32 bits are
autoIncrementID
-  move the IDeviceID interface and its implementation class from the entry package to the new deviceID package, the structure is clearer

Detailed documentation:https://apache-iotdb.feishu.cn/docx/doxcn0K2B6LI1bGr9QpvkYfFUCe


server/src/main/java/org/apache/iotdb/db/metadata/idtable/IDTableHashmapImpl.java
```diff
@@ -485,6 +523,25 @@ private DeviceEntry getDeviceEntryWithAlignedCheck(String deviceName, boolean is
...
+  private DeviceEntry getDeviceEntryWithAutoCreate(IDeviceID deviceID, boolean isAligned) {
+    int slot = calculateSlot(deviceID);
+
+    DeviceEntry deviceEntry = idTables[slot].get(deviceID);
+    // new device
+    if (deviceEntry == null) {
+      deviceEntry = new DeviceEntry(deviceID);
+      deviceEntry.setAligned(isAligned);
+      idTables[slot].put(deviceID, deviceEntry);
+    }
```
MarcosZyk, line 542: There may exists some concurrent problem when concurrent creating deviceEntry.

server/src/main/java/org/apache/iotdb/db/metadata/idtable/deviceID/DeviceIDFactory.java
```diff
@@ -0,0 +1,178 @@
...
+  public static DeviceIDFactory getInstance() {
+    return DeviceIDFactoryHolder.INSTANCE;
+  }
+
+  private DeviceIDFactory() {
+    if (IoTDBDescriptor.getInstance().getConfig().isEnableIDTable()) {
+      if (IoTDBDescriptor.getInstance()
+          .getConfig()
+          .getDeviceIDTransformationMethod()
+          .equals("SHA256")) {
```
MarcosZyk, line 64: Make this transformationMethod an Enum.

server/src/main/java/org/apache/iotdb/db/metadata/idtable/deviceID/DeviceIDFactory.java
```diff
@@ -0,0 +1,178 @@
...
+public class DeviceIDFactory {
+
+  /** used to obtain a IDeviceID instance in the query operation */
+  Function<String, IDeviceID> getDeviceIDFunction;
+
+  /** used to obtain a IDeviceID instance in the insert operation */
+  Function<String, IDeviceID> getDeviceIDWithAutoCreateFunction;
+
+  /** used to obtain a IDeviceID instance in the system restart */
+  Function<String[], IDeviceID> getDeviceIDWithRecoverFunction;
```
MarcosZyk, line 38: Consider encapsulate this three function into one strategy class. Each strategy takes care of there own init, reset and clear etc. behaviors.

server/src/main/java/org/apache/iotdb/db/metadata/idtable/deviceID/StandAloneAutoIncDeviceID.java
```diff
@@ -0,0 +1,310 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Using auto-incrementing id as device id,A complete auto-increment id consists of schemaRegionID
+ * and autoIncrementID, where the upper 32 bits are schemaRegionID and the lower 32 bits are
+ * autoIncrementID
+ */
+public class StandAloneAutoIncDeviceID extends SHA256DeviceID {
```
MarcosZyk, line 45: Rename to ```AutoIncDeviceID```. It will be applied in cluster too.

server/src/main/java/org/apache/iotdb/db/metadata/idtable/deviceID/StandAloneAutoIncDeviceID.java
```diff
@@ -0,0 +1,310 @@
...
+
+  // if the device represented by devicePath is not written to the metadata module, use this
+  // constant instead of devicePath to generate a sha266 value of StandAloneAutoIncDeviceID instance
+  private static final String INVALID_DEVICE_PATH = "invalid.device.path";
+
+  // if the schemaRegionId==-1 of a StandAloneAutoIncDeviceID instance, it means that the device
+  // corresponding to the StandAloneAutoIncDeviceID instance does not exist
+  // return the deviceIdOfNonExistentDevice object for all devicePaths of unmanaged devices in the
+  // metadata module, which can avoid unnecessary object creation while ensuring correctness
+  private static StandAloneAutoIncDeviceID deviceIdOfNonExistentDevice;
```
MarcosZyk, line 65: All the static attributes and methods in this class should be implemented as an independent ```AutoIncDeviceIdManager```. This class should be a simple entity class.

server/src/main/java/org/apache/iotdb/db/metadata/idtable/deviceID/StandAloneAutoIncDeviceID.java
```diff
@@ -0,0 +1,310 @@
...
+  private static StandAloneAutoIncDeviceID buildDeviceID(String devicePath) {
+    try {
+      PartialPath path = new PartialPath(devicePath);
+      // use idTable to determine whether the device has been created
+      IDTable idTable = IDTableManager.getInstance().getIDTable(path);
+      StandAloneAutoIncDeviceID deviceID = new StandAloneAutoIncDeviceID(devicePath);
+      // this device is added for the first time
+      if (idTable.getDeviceEntry(deviceID) == null) {
+        SchemaRegionId schemaRegionId = configManager.getBelongedSchemaRegionId(path);
+        deviceID.schemaRegionId = schemaRegionId.getId();
```
MarcosZyk, line 218: When there's need to create the deviceId, we should already know which schemaRegion it belongs rather than do computation again.

server/src/test/java/org/apache/iotdb/db/metadata/idtable/IDTableRecoverTest.java
```diff
@@ -63,32 +63,32 @@ public void before() {
     isEnableIDTableLogFile = IoTDBDescriptor.getInstance().getConfig().isEnableIDTableLogFile();
 
     IoTDBDescriptor.getInstance().getConfig().setEnableIDTable(true);
-    IoTDBDescriptor.getInstance().getConfig().setDeviceIDTransformationMethod("SHA256");
+    IoTDBDescriptor.getInstance().getConfig().setDeviceIDTransformationMethod("AutoIncrement");
```
MarcosZyk, line 66: Why change this? If you want to test the ```AutoIncrement``` mode, it's better to add new tests. If you want to reuse the test case of ```SHA256```, maybe test class extension will work.

MarcosZyk (commented):
PTAL.

server/src/main/java/org/apache/iotdb/db/metadata/idtable/deviceID/DeviceIDFactory.java
```diff
@@ -49,23 +50,26 @@ public static DeviceIDFactory getInstance() {
...
-      getDeviceIDFunction = PlainDeviceID::new;
+    if (IoTDBDescriptor.getInstance().getConfig().isEnableIDTable()) {
+      switch (DeviceIDMode.valueOf(
+          IoTDBDescriptor.getInstance().getConfig().getDeviceIDTransformationMethod())) {
+        case SHA256:
+          getDeviceIDFunction = SHA256DeviceID::new;
+          return;
+        case AutoIncrement:
+          getDeviceIDFunction = IDTableAutoIncImpl::getDeviceID;
+          return;
```
MarcosZyk, line 61: Add default branch.

server/src/test/java/org/apache/iotdb/db/metadata/idtable/IDTableRecoverTest.java
```diff
@@ -63,32 +63,32 @@ public void before() {
...
   public void clean() throws IOException, StorageEngineException {
+    EnvironmentUtils.cleanEnv();
     IoTDBDescriptor.getInstance().getConfig().setEnableIDTable(isEnableIDTable);
     IoTDBDescriptor.getInstance()
         .getConfig()
         .setDeviceIDTransformationMethod(originalDeviceIDTransformationMethod);
     IoTDBDescriptor.getInstance().getConfig().setEnableIDTableLogFile(isEnableIDTableLogFile);
-
-    EnvironmentUtils.cleanEnv();
   }
```
MarcosZyk, line 79: The ```EnvironmentUtils.cleanEnv()``` is better to executed as the last step, otherwise something caused by config change won't be cleaned clear.

server/src/main/java/org/apache/iotdb/db/metadata/idtable/IDTableAutoIncImpl.java
```diff
@@ -0,0 +1,253 @@
...
+      return deviceIdOfNonExistentDevice;
+    }
+    List<DeviceEntry> deviceEntries = deviceEntrysMap.get(schemaRegionId);
+    synchronized (deviceEntries) {
+      DeviceEntry deviceEntry = deviceEntries.get(autoIncrementID);
+      return (AutoIncDeviceID) deviceEntry.getDeviceID();
+    }
+  }
+
+  public static AutoIncDeviceID fromIdTable(String devicePath) {
```
MarcosZyk, line 93: Rename this method to some name like ```getDeviceIdFromIdTable```. Add a verb.

server/src/main/java/org/apache/iotdb/db/metadata/idtable/deviceID/DeviceIDFactory.java
KeeProMise: added

server/src/main/java/org/apache/iotdb/db/metadata/idtable/IDTableAutoIncImpl.java
KeeProMise: renamed

PR #7156, opened 2022-08-29
[IOTDB-4253] Modify the jdbc query time column to be empty

RYH61:
When the query statement is executed with jdbc, when the last column of the previous row is empty, the next row last Read Was Null will still be true, resulting in the Timestamp column not displaying time information

### DIFF
--- a/integration-test/src/test/java/org/apache/iotdb/db/it/aggregation/IoTDBAggregationIT.java
+++ b/integration-test/src/test/java/org/apache/iotdb/db/it/aggregation/IoTDBAggregationIT.java
@@ -990,9 +990,24 @@ public class IoTDBAggregationIT {
     try (Connection connection = EnvFactory.getEnv().getConnection();
         Statement statement = connection.createStatement()) {
 
-      try (ResultSet resultSet =
-          statement.executeQuery(
-              "select max_value(temperature),count(status) from root.ln.wf01.wt01 group by ([0, 600), 100ms)")) {
+      // create timeseries
+      statement.execute(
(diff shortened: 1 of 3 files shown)


integration-test/src/test/java/org/apache/iotdb/db/it/aggregation/IoTDBAggregationIT.java
```diff
@@ -990,9 +990,24 @@ public void timeWasNullTest() throws Exception {
...
+      for (int i = 0; i < 10; i++) {
+        statement.addBatch(
+            "insert into root.sg1.d1(timestamp, s1, s2) values(" + i + "," + 1 + "," + 1 + ")");
+      }
+
+      statement.execute("insert into root.sg1.d1(timestamp, s3) values(103, 1)");
+      statement.execute("insert into root.sg1.d1(timestamp, s3) values(104, 1)");
+      statement.execute("insert into root.sg1.d1(timestamp, s3) values(105, 1)");
+      statement.executeBatch();
+      try (ResultSet resultSet = statement.executeQuery("select * from root.**")) {
```
JackieTien97, line 1010: Add a new one IT case instead of modifying the previous one

JackieTien97 (changes requested):

integration-test/src/test/java/org/apache/iotdb/db/it/aggregation/IoTDBAggregationIT.java
RYH61: please see https://github.com/apache/iotdb/pull/5281

It is not appropriate to put it here. I have moved it to IoTDBResultSetIT

JackieTien97 (approved):

qiaojialin (approved):

PR #7170, opened 2022-08-30
[IOTDB-4267]Add the logic to process other PlanNode besides InsertNode in MultiLeaderConsensus

xingtanzjr:
Fix the issue that `DeleteDataNode` is not processed when using MultiLeaderConsensus.

Besides, other PlanNode which does not belong to InsertNode won't be processed before this PR.

**Test：**
1. Execute delete operation using client when inserting data by BenchMark.
image

2. Finally, I verified the result for each replica
image


HTHou (approved):
LGTM

PR #7173, opened 2022-08-30
[IOTDB-2782] When users are not granted READ_TIMESERIES or INSERT_TIMESERIES, the verification prompt needs to be unified

RYH61:
1、
image

2、fix docs Fill-Null-Value.md


SpriCoder (approved):
LGTM~

PR #7181, opened 2022-08-31
Fix transport open bug

wangchao316:
We have written an SSL transport, which risks being opened repeatedly. The judgment is made in jdbconnection, and no problem occurs.


influxdb-protocol/src/main/java/org/apache/iotdb/influxdb/session/InfluxDBSession.java
```diff
@@ -128,7 +128,9 @@ public void init() throws IoTDBConnectionException {
           RpcTransportFactory.INSTANCE.getTransport(
               // as there is a try-catch already, we do not need to use TSocket.wrap
               defaultEndPoint.getIp(), defaultEndPoint.getPort(), connectionTimeoutInMs);
-      transport.open();
+      if (!transport.isOpen()) {
+        transport.open();
+      }
```
HTHou, line 133: Any description of this bug?

wangchao316: We have written an SSL transport, which risks being opened repeatedly. The judgment is made in jdbconnection, and no problem occurs.

HTHou (approved):

Beyyes (approved):

PR #7208, opened 2022-09-02
[IOTDB-4304]Add TriggerTable and TriggerInformation

Wei-hao-Li:
Signed-off-by: Weihao Li <18110526956@163.com>

### DIFF
--- a/node-commons/src/main/java/org/apache/iotdb/commons/trigger/TriggerTable.java
+++ b/node-commons/src/main/java/org/apache/iotdb/commons/trigger/TriggerTable.java
@@ -21,7 +21,7 @@ package org.apache.iotdb.commons.trigger;
 
 import org.apache.iotdb.confignode.rpc.thrift.TTriggerState;
 
-import net.jcip.annotations.NotThreadSafe;
+import javax.annotation.concurrent.NotThreadSafe;
 
 import java.util.HashMap;
 import java.util.Map;
--- a/node-commons/src/main/java/org/apache/iotdb/commons/trigger/TriggerTable.java
+++ b/node-commons/src/main/java/org/apache/iotdb/commons/trigger/TriggerTable.java
@@ -19,42 +19,46 @@
 
 package org.apache.iotdb.commons.trigger;
 
+import org.apache.iotdb.confignode.rpc.thrift.TTriggerState;
+
+import net.jcip.annotations.NotThreadSafe;
+
 import java.util.HashMap;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 
 /** This Class used to save the information of Triggers and implements methods of manipulate it. */
+@NotThreadSafe
 public class TriggerTable {
   private final Map<String, TriggerInformation> triggerTable;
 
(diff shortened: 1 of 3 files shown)


node-commons/src/main/java/org/apache/iotdb/commons/trigger/TriggerTable.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+package org.apache.iotdb.commons.trigger;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/** This Class used to save the information of Triggers and implements methods of manipulate it. */
+public class TriggerTable {
```
JackieTien97, line 27: Should this class be thread-safa? If so, you should use lock or synchronized keyword to protect the critical code path.

node-commons/src/main/java/org/apache/iotdb/commons/trigger/TriggerTable.java
```diff
@@ -0,0 +1,60 @@
...
+
+  public void addTriggerInformation(String triggerName, TriggerInformation triggerInformation) {
+    triggerTable.put(triggerName, triggerInformation);
+  }
+
+  public void deleteTriggerInformation(String triggerName) {
+    triggerTable.remove(triggerName);
+  }
+
+  public Map getAllTriggerStates() {
```
JackieTien97, line 46: avoid using raw type of `Map`, indicating directly its key type and value type

JackieTien97: And what's this method for? And in which case it will be called.

node-commons/src/main/java/org/apache/iotdb/commons/trigger/TriggerTable.java
```diff
@@ -0,0 +1,60 @@
...
+    // After we get the size of TriggerTable, maybe more values has put in it before we put all
+    // TriggerStates in map of allTriggerStates.
+    // Mostly we create only one Trigger at a time, so the initial size is 'triggerTable.size() +
+    // 1' here.
+    Map allTriggerStates = new HashMap(triggerTable.size() + 1);
```
JackieTien97, line 51: I still cannot under why we need to init the map with size as `triggerTable.size() + 1`.

node-commons/src/main/java/org/apache/iotdb/commons/trigger/TriggerTable.java
```diff
@@ -0,0 +1,60 @@
...
+    triggerTable.forEach((k, v) -> allTriggerStates.put(k, v.getTriggerState()));
+
+    return allTriggerStates;
+  }
+
+  public Map getTable() {
```
JackieTien97, line 57: avoid using raw type of Map and BTW, if you mean to design this class as thread safe, it's better not to return the underlying map outside.

JackieTien97 (changes requested):

JackieTien97 (approved):

PR #7212, opened 2022-09-02
[IOTDB-4300] writing data with benchmark then remove a datanode, there are some errors

BanDuWuZi:
remove datanode while writing data (with benchmark) will occur some errors(region not exist, ratis failed, or node maybe down)
details: https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-4300

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/mpp/plan/execution/QueryExecution.java
+++ b/server/src/main/java/org/apache/iotdb/db/mpp/plan/execution/QueryExecution.java
@@ -479,18 +479,21 @@ public class QueryExecution implements IQueryExecution {
     // collect redirect info to client for writing
     if (analysis.getStatement() instanceof InsertBaseStatement) {
       if (state != QueryState.FINISHED) {
-        //if insert not finished, then before redirect to client: clean partition cache and re analyse to get the current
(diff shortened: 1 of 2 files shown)


wangchao316 (approved):
LGTM

server/src/main/java/org/apache/iotdb/db/mpp/plan/execution/QueryExecution.java
```diff
@@ -478,10 +478,19 @@ private ExecutionResult getExecutionResult(QueryState state) {
 
     // collect redirect info to client for writing
     if (analysis.getStatement() instanceof InsertBaseStatement) {
+      if (state != QueryState.FINISHED) {
+        //if insert not finished, then before redirect to client: clean partition cache and re analyse to get the current
+        // region locations, so the redirect node is available. maybe region migrated when writing.
+        partitionFetcher.invalidAllCache();
```
MrQuansy, line 485: This will cause drastic decrease of writing and reading performance. You should invalid specific cache entries instead of all cache.

MrQuansy:
NEED_REDIRECRION status aims to tell client the location of a device, it does not tell the client "you send the request to an error datanode".

Issue #7211, opened 2022-09-02
Cannot read more PhysicalPlans from 0, successfully read index is /var/lib/iotdb/data/wal/root.ceshi.p1.s_1/0-1662101750410-1-0-0.tsfile/wal1. The reason is java.nio.BufferUnderflowException: null


SunnySaman:
**Describe the bug**

version : 0.13.1 release

iotdb 安装完成之后；

```
[root@iotdb-103 logs]# /home/iotdb/sbin/start-cli.sh -h localhost -p 6667 -u root -pw root
---------------------
Starting IoTDB Cli
---------------------
 _____       _________  ______   ______
|_   _|     |  _   _  ||_   _ `.|_   _ \
  | |   .--.|_/ | | \_|  | | `. \ | |_) |
  | | / .'`\ \  | |      | |  | | |  __'.
 _| |_| \__. | _| |_    _| |_.' /_| |__) |
|_____|'.__.' |_____|  |______.'|_______/  version 0.13.1


IoTDB> login successfully
IoTDB> create storage group root.ceshi.p1.s_1;
>
> create schema template t4ceshi aligned (isa boolean encoding=RLE,isb boolean encoding=RLE,pv text encoding=DICTIONARY,
> sid int32 encoding=RLE,eq text encoding=DICTIONARY,pty int32 encoding=RLE,ityp int32 encoding=RLE,sti int64 encoding=RLE,
> pti int64 encoding=RLE,ori text encoding=PLAIN,sv text encoding=DICTIONARY,iv int32 encoding=RLE,dov double encoding=GORILLA,
> bv boolean encoding=RLE,dv int64 encoding=RLE,lv int64 encoding=RLE,stid int32 encoding=RLE,fv float encoding=RLE,
> cv text encoding=PLAIN,wv int32 encoding=RLE);
>
> set schema template t4ceshi to root.ceshi.p1.s_1;
>
>
>
> insert into root.ceshi.p1.s_1.sys3.eq1.pv2
> (timestamp,isa,isb,pv,sid,eq,pty,ityp,sti,pti,ori,sv,iv,dov,bv,dv,lv,stid,fv,cv,wv)
> aligned VALUES
> (NOW(),False,False,'ceshi',24,'ceshi',9,1,1660891161764,1660891161764,'0ce3','',null,null,
> null,null,null,19,1.1,'ceshi',null);
Msg: The statement is executed successfully.
Msg: The statement is executed successfully.
Msg: The statement is executed successfully.
[...]

PR #7221, opened 2022-09-03
[To rel/0.13][IOTDB-4320][ISSUE-7211] Recover insert row with null value throw NPE

Cpaulyz:
## Bug Description
#7211 
## Solution
Serialize null type in `InsertRowPlan` if value is null.


HTHou (approved):

HeimingZ (approved):
LGTM 👍

PR #7240, opened 2022-09-06
[IOTDB-3642] Add retry mechanism when resource unavailable

SzyWilliam:


consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
```diff
@@ -148,6 +151,49 @@ public void stop() throws IOException {
...
 
+  /** launch a consensus write with retry mechanism */
+  private RaftClientReply writeWithRetry(CheckedSupplier<RaftClientReply, IOException> caller)
+      throws IOException {
+
+    Predicate<RaftClientReply> shouldRetry =
+        reply ->
+            !reply.isSuccess()
+                && (reply.getException() != null
+                    && reply.getException() instanceof ResourceUnavailableException);
```
JackieTien97, line 162: use a private method instead of lambda.

consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
```diff
@@ -148,6 +151,49 @@ public void stop() throws IOException {
...
+
+    final int maxRetryTimes = 3;
+    final long waitMillis = 500;
```
JackieTien97, line 165: put these parameters to config.

consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
JackieTien97, line 162: use a private method instead of lambda.

JackieTien97 (changes requested):

consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
SzyWilliam: done

consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
SzyWilliam: done

JackieTien97 (approved):

PR #7245, opened 2022-09-06
[IOTDB-4347]NPE error when session is expired

yifuzhou:
When session is expired or session is closed, it will occurs NPE when obtaining the username.
See IOTDB-4347

After fix:
image


MarcosZyk (approved):

HTHou (approved):

wangchao316 (approved):
LGTM

PR #7250, opened 2022-09-06
[IOTDB-4349] Remove error status in data node

HeimingZ:


server/src/assembly/resources/conf/iotdb-datanode.properties
```diff
@@ -319,9 +319,8 @@ timestamp_precision=ms
 # What will the system do when unrecoverable error occurs.
 # Datatype: String
 # Three strategies are as follows:
```
HTHou, line 321:
```suggestion
# Optional strategies are as follows:
```

HeimingZ: Fixed.

qiaojialin (approved):

PR #7263, opened 2022-09-07
[IOTDB-4358] set read consistency level to 'weak', quey SimpleFragmentParallelPlanner may exist out of bounds exception

BanDuWuZi:
set read consistency level to 'weak', quey SimpleFragmentParallelPlanner may exist out of bounds exception
see: https://issues.apache.org/jira/browse/IOTDB-4358

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/distribution/SimpleFragmentParallelPlanner.java
+++ b/server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/distribution/SimpleFragmentParallelPlanner.java
@@ -149,6 +149,7 @@ public class SimpleFragmentParallelPlanner implements IFragmentParallelPlaner {
       targetIndex = 0;
     } else {
       targetIndex = (int) (queryContext.getSession().getSessionId() % availableDataNodes.size());
+      targetIndex = Math.abs(targetIndex);
     }
     return availableDataNodes.get(targetIndex);
   }


xingtanzjr:
Can we see the log like `available replicas: ` before the exception occurs ?

BanDuWuZi:
> Can we see the log like `available replicas: ` before the exception occurs ?



> Can we see the log like `available replicas: ` before the exception occurs ?

there 3 available datanode,  see https://issues.apache.org/jira/browse/IOTDB-4358 exception info, I will find the detail log later
the reason of the issue is that the generated **sessionId is negative** , so  [queryContext.getSession().getSessionId() % availableDataNodes.size()] is negative.

xingtanzjr:
> 

Yes, your analysis is correct. But...it is wired that we got an index `-2` because the sessionId is always positive and the value of mod operation should be in [0, 2]. Why did we got a `-2` here ?

wangchao316:
> > > Can we see the log like `available replicas: ` before the exception occurs ?
> > 
> > 
> > there 3 available **datanode: java.lang.IndexOutOfBoundsException: Index: -2, Size: 3** (https://issues.apache.org/jira/browse/IOTDB-4358 exception info), I will find the detail log later. the reason of the issue is that the generated **sessionId is negative** , so [queryContext.getSession().getSessionId() % availableDataNodes.size()] is negative.
> 
> Yes, your analysis is correct. But...it is wired that we got an index `-2` because the sessionId is always positive and the value of mod operation should be in [0, 2]. Why did we got a `-2` here ? Thus I want to know the detailed log here

The session ID cannot be automatically incremented. It should be secure random number . In the case of a secure random number, a negative number appears. We're using a secure random number.

PR #7264, opened 2022-09-07
[IOTDB-4361] Add a precheck in removing datanode

MiniSho:
We should add a precheck when the confignode executes removing datanode :
1. Make sure the node which execute addRegionPeer is in running
2. Throw an error which says it's not supported to remove the datanode because the configuration of the cluster is only one replication.

In addition, new a package called confignode  in ConfigNodeIT.


SpriCoder (approved):

confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/DataNodeRemoveHandler.java
```diff
@@ -492,4 +524,21 @@ private Optional<TDataNodeLocation> filterDataNodeWithOtherRegionReplica(
...
+  /**
+   * Check the protocol of the cluster, standalone is not supported to remove data node currently
+   *
+   * @return SUCCEED_STATUS if the Cluster is not standalone protocol, REMOVE_DATANODE_FAILED
+   *     otherwise
+   */
+  private TSStatus checkClusterProtocol() {
+    TSStatus status = new TSStatus(TSStatusCode.SUCCESS_STATUS.getStatusCode());
+    if (CONF.getDataRegionConsensusProtocolClass().equals(ConsensusFactory.StandAloneConsensus)
+        && CONF.getSchemaRegionConsensusProtocolClass()
```
Beyyes, line 537:
```suggestion
        || CONF.getSchemaRegionConsensusProtocolClass()
```

Beyyes (commented):
add a running test screenshot

confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/DataNodeRemoveHandler.java
```diff
@@ -433,8 +445,28 @@ private TSStatus checkDataNodeExist(RemoveDataNodePlan removeDataNodePlan) {
    */
   private TSStatus checkRegionReplication(RemoveDataNodePlan removeDataNodePlan) {
     TSStatus status = new TSStatus(TSStatusCode.SUCCESS_STATUS.getStatusCode());
-    int removedDataNodeSize = removeDataNodePlan.getDataNodeLocations().size();
+    List<TDataNodeLocation> removedDataNodes = removeDataNodePlan.getDataNodeLocations();
     int allDataNodeSize = configManager.getNodeManager().getRegisteredDataNodeCount();
+    if (NodeInfo.getMinimumDataNode() == 1) {
```
CRZbulabula, line 450: We should step into this if statement as long as the schemaReplicationFactor or the dataReplicationFactor equals to 1

CRZbulabula, line 450: Add annotation in front of this block of code in order to explain the usage.

CRZbulabula (changes requested):
PTAL~

confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/DataNodeRemoveHandler.java
MiniSho: done

confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/DataNodeRemoveHandler.java
MiniSho: done

MiniSho: done

MiniSho:
when set the configuration as Standalone protocol, it throws an error
ba8f3fe60cbb242de3869c697f25121

MiniSho:
when remove a datanode whic is not in running state and the configuration of cluster is one replication, it throws an error
289a747043fd6a5108313d80ce46ee0

Beyyes (approved):
LGTM

CRZbulabula (approved):
LGTM~

PR #7269, opened 2022-09-08
[IOTDB-4366] Fix Ratis 1 replica migrate error; terminate the RegionMigrationProcedure process when datanode report failure result

Beyyes:
## Description

Two improvements:

1) executing remove-datanode.sh for Ratis 1 replica will be failed in the `deleteOldRegion` step, so we cancel the step of `deleteOldRegion` in RegionMigrateProcedure in the case of Ratis 1 replica. 
image

2) RegionMigrateProcedure step will be terminated when the datanode reported failed.
image

The test running results are as below.

SchemaRegion=Ratis, DataRegion=Ratis

Before remove datanode:
image

After remove the datanode 0.0.0.0:6669
image

jps result:
image
image


CRZbulabula (approved):
LGTM~

qiaojialin (approved):

PR #7283, opened 2022-09-09
[IOTDB-4208] Fix the path of wal folders in confignode.

SpriCoder:
See https://issues.apache.org/jira/browse/IOTDB-4208


CRZbulabula (approved):
LGTM~

xingtanzjr (approved):

PR #7286, opened 2022-09-09
[IOTDB-4110] customize trigger snapshot

sollhui:
### Content3 ...

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
```diff
@@ -575,6 +583,50 @@ public ConsensusGenericResponse triggerSnapshot(ConsensusGroupId groupId) {
     return ConsensusGenericResponse.newBuilder().setSuccess(reply.isSuccess()).build();
   }
 
+  private void triggerSnapshotByCustomize(ConsensusConfig config) {
```
SzyWilliam, line 586: maybe we should add synchronized to method signature. Consider this: this method is not finished when 60s past and another call started.

SzyWilliam (changes requested):
Thanks for working on this! There are some issues need to be fixed, PTAL. Also, use `mvn spotless` to format code before push.

SzyWilliam:
@HHoflittlefish777 Thanks for the update. Could you please see if you can add some unit tests to verify this?

SzyWilliam (approved):
LGTM

PR #7297, opened 2022-09-13
[IOTDB-4362] Abstract Session interface

MiniSho:
abstract all method in Session into ISession interface


session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -88,7 +88,7 @@
 import java.util.stream.Collectors;
 
 @SuppressWarnings({"java:S107", "java:S1135"}) // need enough parameters, ignore todos
-public class Session {
+public class Session implements ISession {
```
Beyyes, line 91: Methods need to be added `@Overried` annotation.

MiniSho: done

Beyyes (approved):

Issue #5884, opened 2022-05-12
"targetWriter" is a method parameter, and should not be used for synchronization.


MinaQin:
Code Path:server/src/main/java/org/apache/iotdb/db/engine/compaction/writer/AbstractCompactionWriter.java
<img width="831" alt="image" src="https://user-images.githubusercontent.com/55030443/168058181-3295b2fa-a3bb-4c63-8cbd-f4d1887cc8bd.png">
Synchronizing on a class field synchronizes not on the field itself, but on the object assigned to it. So synchronizing on a non-final field makes it possible for the field’s value to change while a thread is in a block synchronized on the old value. That would allow a second thread, synchronized on the new value, to enter the block at the same time.

The story is very similar for synchronizing on parameters; two different threads running the method in parallel could pass two different object instances in to the method as parameters, completely undermining the synchronization.

GuanchuShen:
[WeOpen Star]I would like to help

mroccyen:
i will try my best to do it，please assign to me

PR #7300, opened 2022-09-13
fix issue 5884

mroccyen:
## Description

this PR fix issue #5884


THUMarkLau:
For `CrossSpaceCompactionWriter`, multiple threads may write to multiple different `TsFileIOWriter`, if we synchronized them with a single object, it may lead to performance decrease. It's better to maintain a global array, which map sub-task-id to the writer they are using. When thread writes data, it only needs to synchronized the writer it is using.

THUMarkLau (commented):
For `CrossSpaceCompactionWriter`, multiple threads may write to multiple different `TsFileIOWriter`, if we synchronized them with a single object, it may lead to performance decrease. It's better to maintain a global array, which map sub-task-id to the writer they are using. When thread writes data, it only needs to synchronized the writer it is using.

PR #7311, opened 2022-09-14
add a section of delete data

wanghui42:


wanghui42:
Version:0.14
adding a section of "delete data" in user guide, changing the directory structure and folder file locations as a test. As follows: "Write and Delete Data" is divided into two chapters: "Write Data" and "Delete Data", and "TTL" is moved from "Operate-Metadata" to "Delete Data".

PR #7312, opened 2022-09-14
[IOTDB-4364]Reduce read amplication in compaction

choubenson:
**Description**
In massive timeseries scenarios, each device has 20 timeseries, and each file has tens of thousands of devices, that is, millions of timeseries. Read throughput in cross compaction is 100~1000 times greater than write throughput.

**Solution**
When compacting, it is necessary to obtain devices and timeseries from all source files, as well as the metadata of each timeseries. To avoid traversing the Zesong tree multiple times, we record the intermediate nodes of the Zesong tree. Please refer to [IOTDB-4364](https://issues.apache.org/jira/browse/IOTDB-4364) for the experimental results.


THUMarkLau:
CI is failed.

PR #7320, opened 2022-09-15
[IOTDB-4399] Control text chunk size in memtable

MrQuansy:


server/src/main/java/org/apache/iotdb/db/utils/datastructure/BinaryTVList.java
```diff
@@ -58,6 +62,12 @@ public void putBinary(long timestamp, Binary value) {
     if (sorted && rowCount > 1 && timestamp < getTime(rowCount - 2)) {
       sorted = false;
     }
+    rawSize += getBinarySize(value);
```
qiaojialin, line 65: This should already be estimated before write, store the memory in Binary and use here

server/src/main/java/org/apache/iotdb/db/utils/datastructure/AlignedTVList.java
```diff
@@ -582,6 +595,11 @@ protected void releaseLastValueArray() {
     }
   }
 
+  @Override
+  public boolean reachMaxChunkSizeThreshold() {
+    return false;
```
qiaojialin, line 600: What are you coding? You add a reachMaxChunkSizeFlag but never use it.
Add UTs

qiaojialin (changes requested):

server/src/assembly/resources/conf/iotdb-datanode.properties
```diff
@@ -385,6 +385,11 @@ timestamp_precision=ms
 # Datatype: int
 # avg_series_point_number_threshold=100000
 
+# When a chunk in memtable reaches this threshold, flush the memtable to disk.
+# The default threshold is 20 MB.
+# Datatype: int
```
HTHou, line 390: long?

server/src/main/java/org/apache/iotdb/db/utils/datastructure/AlignedTVList.java
```diff
@@ -54,6 +55,8 @@ public abstract class AlignedTVList extends TVList {
   // data types of this aligned tvList
   protected List<TSDataType> dataTypes;
 
+  protected long[] valueChunkRawSize;
```
HTHou, line 58: Add javadoc?

HTHou (approved):

PR #7321, opened 2022-09-15
[IOTDB-4245] Fix wrong integer upper bound about data type inferring of timeseries

lancelly:
Fix wrong integer upper bound about data type inferring of timeseries. According to userdoc, it should be 2^24, not 2^25.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/utils/TypeInferenceUtils.java
+++ b/server/src/main/java/org/apache/iotdb/db/utils/TypeInferenceUtils.java
@@ -64,7 +64,7 @@ public class TypeInferenceUtils {
   }
 
   private static boolean isConvertFloatPrecisionLack(String s) {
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/iotdb/db/utils/TypeInferenceUtils.java
```diff
@@ -64,7 +64,7 @@ private static boolean isBoolean(String s) {
   }
 
   private static boolean isConvertFloatPrecisionLack(String s) {
-    return Long.parseLong(s) > (2 << 24);
+    return Long.parseLong(s) > (1 << 24);
```
HTHou, line 67: Add UT?

lancelly: Done.

JackieTien97 (approved):

PR #7322, opened 2022-09-15
[IOTDB-4412][MultiLeader] add `cacheWindowTimeInMs` into config.

SpriCoder:
See: https://issues.apache.org/jira/browse/IOTDB-4412


qiaojialin (approved):

server/src/assembly/resources/conf/iotdb-datanode.properties
```diff
@@ -194,6 +194,11 @@ target_config_nodes=127.0.0.1:22277
 # Datatype: long
 # multi_leader_throttle_threshold_in_byte=53687091200
 
+# Maximum wait time of write cache in MultiLeader consensus
+# If it's a value smaller than 0, use the default value Long.MAX_VALUE.
```
qiaojialin, line 198:
```suggestion
# If this value is less than or equal to 0, use the default value Long.MAX_VALUE.
```

PR #7345, opened 2022-09-16
[IOTDB-4419] Maintain RegionStatus through cluster heartbeat

CRZbulabula:
I've made the following updates in this PR:

1. Refactor the RegionGroupCache, now ConfigNode will maintain separate RegionCaches for each Region.
2. Maintain RegionStatus through cluster heartbeat.
3. Move showRegion and showDataNode interface from ConfigManager to other sub-Managers. 

You can read the [Region State Machine](https://apache-iotdb.feishu.cn/docx/doxcnMQppTJCURUuWZ5LfWA5dK6) doc for more details.

I've started a 1C3D cluster and finished some tests on my local machine:

image

1. Create some RegionGroups, the show regions and show datanodes SQL are correct:

image

image

2. Shutdonw a DataNode, the corresponding RegionStatus and NodeStatus will become to UNKNOWN after 20s:
image


Beyyes (commented):
If RegionStatus also shown as `Unkonwn` when datanode is `ReadOnly` or `Removing`?
Do we need adding `ReadOnly` or `Removing` status for RegionStatus later?

wangchao316:
> If RegionStatus also shown as `Unkonwn` when datanode is `ReadOnly` or `Removing`? Do we need adding `ReadOnly` or `Removing` status for RegionStatus later?

support

CRZbulabula:
> If RegionStatus also shown as `Unkonwn` when datanode is `ReadOnly` or `Removing`? Do we need adding `ReadOnly` or `Removing` status for RegionStatus later?

I've thought these two cases:

1. For DataNode in 'ReadOnly' status: It doesn't mean that the Region is in 'ReadOnly' status if the DataNode where the Region situated is currently 'ReadOnly'. Actually, we havn't defined in which scenario a Region is 'ReadOnly'. On the other hand, the ConfigNode-leader will consider a RegionGroup is 'Disabled' if one of its Region's located DataNode is 'ReadOnly'.
2. For DataNode in 'Removing' status: It might be better to set the corresponding Regions into 'Removing' status either?

qiaojialin (approved):

PR #7354, opened 2022-09-16
[IOTDB-3656] mpp load supports modification

yschengzi:
- [x] support load .mods
- [ ] last cache
- [ ] clean up
- [ ] tsfile name restriction


integration-test/src/test/java/org/apache/iotdb/db/it/IOTDBLoadTsFileIT.java
```diff
@@ -0,0 +1,579 @@
...
+    private static final MeasurementSchema MEASUREMENT_30 =
+        new MeasurementSchema("sensor_30", TSDataType.INT32, TSEncoding.RLE);
+
+    // device 4, aligned, sg 1
+    private static final String DEVICE_4 = "root.sg.test_1.a_4";
+    private static final MeasurementSchema MEASUREMENT_40 =
+        new MeasurementSchema("sensor_40", TSDataType.INT32, TSEncoding.RLE);
+  }
+
+  public class TsFileGenerator implements AutoCloseable {
```
ericpai, line 410: It seems that `TsFileGenerator` is a common tool in integration test, despite to depend on those internal classes of tsFile directly, you can
1. Move it to src/main, as a separate class in a package.
2. Refine the methods of this class, let the parameters are raw Java types(int, string, not Path etc).

server/src/main/java/org/apache/iotdb/db/conf/IoTDBDescriptor.java
```diff
@@ -1006,6 +1006,10 @@ public void loadProperties(Properties properties) {
 
     conf.setTimePartitionIntervalForStorage(
         convertMilliWithPrecision(conf.getTimePartitionIntervalForStorage()));
+
+    if (!conf.isClusterMode()) {
+      conf.setTimePartitionIntervalForRouting(conf.getTimePartitionIntervalForStorage());
+    }
```
HTHou, line 1012: Why do u add this?

HTHou (commented):
please add the design doc

HTHou (approved):

PR #7359, opened 2022-09-19
[IOTDB-4391] Apply PatternTreeMap in parsing .mods file to optimize query

Cpaulyz:
## Description
* `QueryContext`: Use `PatternTreeMap` to parse .mods file. It constructs modification tree once when parsing and getOverlap method cost O(n) time complexity. It will save M*N traversal time if there is a query about M timeseries and length of .mods is N.
* `ChunkMetaData#insertIntoSortedDeletions`: Use TreeMap to store TimeRange. Reduce the time complexity from O(n^2) to O(nlogn).

BTW, I think there are still some potential optimization points. For example, we can sort and merge modification as possible in QueryContext. So if there is a query for a lot of timeseries and there are a lot of modifications about a path pattern, it only need to sort and merge once instead of do it many times in `ChunkMetaData#insertIntoSortedDeletions`.

Experiment：https://apache-iotdb.feishu.cn/docs/doccnMqwBpAYTqkUAj3xaehsXJf#sKUpPL


tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/ChunkMetadata.java
```diff
@@ -197,32 +198,30 @@ public void setVersion(long version) {
...
+        deleteIntervalSet.remove(interval);
+        interval = deleteIntervalSet.ceiling(timeRange);
+      }
+    } else {
+      deleteIntervalSet = new TreeSet<>();
     }
 
-    resultInterval.add(new TimeRange(startTime, endTime));
-    deleteIntervalList = resultInterval;
+    deleteIntervalSet.add(timeRange);
```
JackieTien97, line 224: The time complexity of this method seems not to be O(logN), becasue it's still possible that we need to merge more than one element in this TreeMap.
Actually, it's a classic algorithm problem called range merge, and in our case, we can also have all the time range firstly, and then do merge together(or if you like, you still can keep current style that update it every time this method is called.) You can refer to this [blog](https://juejin.cn/post/7083845140283916325 ) about the concrete algorithm solution

JackieTien97: BTW, I think we can move this merge phase into the generation of PatternTreeMap(make the `BiConsumer<V, Set<V>> appendFunction` in `getModsPatternTreeMap` more complicated) for Deletion and of course we need to take more thing into account like `version` and `offset`

Cpaulyz: > BTW, I think we can move this merge phase into the generation of PatternTreeMap(make the `BiConsumer<V, Set<V>> appendFunction` in `getModsPatternTreeMap` more complicated) for Deletion and of course we need to take more thing into account like `version` and `offset`

Yes.. In fact, my original implementation was to do the merge in appendFunction ([code](https://github.com/apache/iotdb/pull/7359/commits/ea497089cecc0902baaee0b154dd0c775bc97b07#diff-b24f923a2084b6c2e8abab3aa2be961842fef5b22d51466daddbdeb989a309cb)). But I found that it is not a good idea for some cases because it will merge all modification in modFile. For example, there are 10000 modifications about `root.sg1.d1.s1` and 1 modification about `root.sg1.d1.s2`. If query is about `root.sg1.d1.s2`, the merge operation about `root.sg1.d1.s1` is a waste of time. 

I think maybe I can apply range merge algorithm before `return allModifications.getOverlapped(path);`. It ensures the return list of `QueryContext#getPathModifications` is in order. And upper layer no need to sort time range again (seems save 1 time).

tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/ChunkMetadata.java
```diff
@@ -198,30 +197,35 @@ public void setVersion(long version) {
...
-        timeRange.merge(interval);
-        deleteIntervalSet.remove(interval);
-        interval = deleteIntervalSet.ceiling(timeRange);
+      for (TimeRange interval : deleteIntervalList) {
+        if (interval.getMax() < startTime) {
+          resultInterval.add(interval);
+        } else if (interval.getMin() > endTime) {
+          resultInterval.add(new TimeRange(startTime, endTime));
+          startTime = interval.getMin();
+          endTime = interval.getMax();
```
JackieTien97, line 219: It seems that you can add all the remaining `TimeRange` in `deleteIntervalList` and then break the loop early here, because the remaining intervals won't overlap.

JackieTien97 (changes requested):

tsfile/src/main/java/org/apache/iotdb/tsfile/file/metadata/ChunkMetadata.java
Cpaulyz: done

PR #7360, opened 2022-09-19
[To rel/0.13] [IOTDB-4451] Add log for cross space compaction selection when a file has been compacted too many times

THUMarkLau:
See [IOTDB-4451](https://issues.apache.org/jira/browse/IOTDB-4451).


choubenson (approved):
LGTM～

server/src/main/java/org/apache/iotdb/db/engine/compaction/cross/rewrite/selector/RewriteCompactionFileSelector.java
```diff
@@ -259,6 +260,7 @@ private boolean checkIsSeqFilesValid() {
    * @param unseqFile the tsFileResource of unseqFile to be compacted
    */
   private void selectOverlappedSeqFiles(TsFileResource unseqFile) {
+    final int SELECT_WARN_THRESHOLD = 10;
```
MrQuansy, line 263: Make it as a configuration or make it static

MrQuansy (changes requested):

PR #7373, opened 2022-09-21
[IOTDB-4479] Add pattern overlap check when deleting timeseries

MarcosZyk:
### Motivation

The input path patterns may cross different storage groups. SchemaRegion doesn't need to execute unrelated path pattern.

### Modification

Before executing in schemaRegion, filter out all path patterns that overlap with schemaRegion's storage group pattern.


wangchao316 (approved):
LGTM

PR #7378, opened 2022-09-21
[IOTDB-4477] Support datatype cast while inserting with schema mismatch

MrQuansy:
https://issues.apache.org/jira/browse/IOTDB-4477

Support
cast INT32 -> INT64, FLOAT, DOUBLE
cast INT64 -> DOUBLE
cast FLOAT -> DOUBLE

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/utils/CommonUtils.java
+++ b/server/src/main/java/org/apache/iotdb/db/utils/CommonUtils.java
@@ -127,6 +127,31 @@ public class CommonUtils {
     return false;
   }
 
+  public static Object castValue(TSDataType srcDataType, TSDataType destDataType, Object value) {
+    switch (srcDataType) {
+      case INT32:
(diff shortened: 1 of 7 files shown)


server/src/main/java/org/apache/iotdb/db/utils/CommonUtils.java
```diff
@@ -108,6 +109,65 @@ public static Object parseValue(TSDataType dataType, String value) throws QueryP
...
+        }
+        break;
+      case FLOAT:
+        float[] tmp = (float[]) value;
+        if (destDataType == TSDataType.INT64) {
+          long[] result = new long[tmp.length];
+          for (int i = 0; i < tmp.length; i++) {
+            result[i] = (long) tmp[i];
+          }
+          value = result;
```
HTHou, line 158: Remove FLOAT->INT64

HTHou (approved):

PR #7379, opened 2022-09-21
[IOTDB-4481] Enable time partition managed by config node

MrQuansy:
https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-4481


CRZbulabula (changes requested):
PTAL:

1. The enable_partition parameter should be removed from iotdb-datanode.properties
2. The ConfigNode should check the consistency of enable_partition when start
3. Better add ClusterIT to verify the partition function

PR #7380, opened 2022-09-21
[IOTDB-4470] Code generation to accelerate expression calculation

flashzxi:
At this stage, expression evaluation is based on the volcano model, and the evaluate function is called to the root node of the expression tree, and the tree will recursively call the evaluate function of the subtree to calculate the result of its own node. Such a procedure has a large number of virtual function calls, and each time needs to repeatedly call the function written by the corresponding type through the type of switch expression. The additional overhead of this process is enormous. It is hoped that the handwritten code of the calculation can be directly constructed according to the input expression to improve the execution efficiency


xingtanzjr (approved):

xingtanzjr:
We have tested the code with several experiments.

xingtanzjr:
@flashzxi  Please resolve the conflicts

PR #7381, opened 2022-09-21
[IOTDB-4478] [Ratis] Add RatisConsensus config parameters in confignode.properties

SzyWilliam:
# Description
Currently, there are many config parameters available for RatisConsensus. All params can refer to `RatisConfig`.
In this PR, I propose to expose **workload-related** params in confignode.properties so that the user can tune these params according to their workloads.
# Param List
1. ratis_snapshot_trigger_threshold, determine how many logs written to trigger a snapshot.
2. ratis_log_unsafe_flush_enable, determine whether we can flush raft log asynchronous without wait FileChannel.force to return.
3. ratis_log_segment_size_max, determine max capacity of a single Raft Log segment.
4. ratis_grpc_flow_control_window, determine gRPC log appender's flow control window size
5. ratis_rpc_leader_election_timeout_min_ms, determine min election timeout
6. ratis_rpc_leader_election_timeout_max_ms, determine max election timeout

### DIFF
--- a/confignode/src/assembly/resources/conf/iotdb-confignode.properties
+++ b/confignode/src/assembly/resources/conf/iotdb-confignode.properties
@@ -343,3 +343,21 @@ target_config_nodes=127.0.0.1:22277
 
 # max payload size for a single log-sync-RPC from leader to follower
 # ratis_log_appender_buffer_size_max = 4194304
+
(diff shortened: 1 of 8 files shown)


wangchao316 (approved):
LGTM

JackieTien97 (approved):

PR #7404, opened 2022-09-22
[IOTDB-3671] terminate workerMonitorExecutor in Join method

MiniSho:
see: https://issues.apache.org/jira/browse/IOTDB-3671
add a terminate method


CRZbulabula (approved):
LGTM

PR #7408, opened 2022-09-22
[IOTDB-4501]fix FileAlreadyExistsException when taking snapshot

yifuzhou:
When we use snapshot to backup data, when a snapshotTsFile already has a hard link and point to the same tsfile, it will occurs FileAlreadyExistsException. So each time when we do hard link function, we need to check and delete.


THUMarkLau (approved):

wangchao316 (approved):
LGTM

PR #7424, opened 2022-09-25
[IOTDB-4520] The tag schema region is implemented and supports the influxdb protocol

KeeProMise:
## Description
The current influxdb adapter uses iotdb to store the order of tags in order to generate a unique path for the combination of tags with the same semantics during implementation. In this way, in the implementation, it is necessary to first access iotdb to obtain the order of tags, and then convert the tags into paths and generate PhysicalPlan, which results in poor performance and is more complicated to implement a tag model adapter. This work enables the influxdb adapter to support the use of native tag model metadata management modules.
## New features
- lsm framework memory structure implementation：see https://apache-iotdb.feishu.cn/docx/doxcncdQ0sejMEEVqC1gIM9CKef#doxcni0oA80sqYOqeoEwp5kHun4
- tag inverted index：see https://apache-iotdb.feishu.cn/docx/doxcnlNVJnjaAMo0xVSpMsOAIVe
- tag schema region： see https://apache-iotdb.feishu.cn/docx/doxcn65XjYe64t8QyGiyLN2Ig0c
- influxdb adapter supports tag schema region：see https://apache-iotdb.feishu.cn/docx/doxcnZzEZ1BwXS6EQfrUEuGOX4d#doxcnKsqAa24qOwya4Oi1qdtFib


lsm/src/main/java/org/apache/iotdb/lsm/context/QueryRequestContext.java
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * represents the context of a query request, this class can be extended to implement a custom
+ * context
+ */
+public class QueryRequestContext extends RequestContext {
+
+  // save the key of each level
+  List<Object> keys;
+
```
MarcosZyk, line 35: We shall avoid direct use ```Object``` class. Is there any way to implement this feature with generics? 

Besides, the attribute ```keys``` seems to be part of the query request, rather than the query context. The ```context``` usually means to provide some states of the system or module when executing the request.

lsm/src/main/java/org/apache/iotdb/lsm/manager/LsmManager.java
```diff
@@ -0,0 +1,42 @@
...
+   */
+  void process(T memNode, C context) throws Exception;
+
+  /**
+   * add the LevelProcess of the next layer of memory nodes
+   *
+   * @param next LevelProcess of the next layer
+   * @return LevelProcess of the next layer
+   */
+  <O> LevelProcess<T, O, C> nextLevel(LevelProcess<T, O, C> next);
```
MarcosZyk, line 41: It seems this method is used to build the process method of ```LSMManager```. What about separating these building methods into an individual  ```LSMManagerBuilder```. The ```LSMManager``` will only keep the executing methods. PLZ refer to the design pattern, Builder Pattern.

lsm/src/main/java/org/apache/iotdb/lsm/wal/WALRecord.java
```diff
@@ -0,0 +1,51 @@
...
+ * under the License.
+ */
+package org.apache.iotdb.lsm.wal;
+
+import java.io.DataInputStream;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+
+/** represents a wal record, which can be extended to implement more complex wal records */
+public abstract class WALRecord implements Cloneable {
```
MarcosZyk, line 26: Make this an Interface. The abstract class is not that convenient in Java, since Java only supports single extension.

lsm/src/main/java/org/apache/iotdb/lsm/levelProcess/BasicLevelProcess.java
```diff
@@ -0,0 +1,79 @@
...
+  // the next level process
+  LevelProcess<O, ?, C> next;
+
+  /**
+   * process the current layer memory node
+   *
+   * @param memNode memory node
+   * @param context request context
+   */
+  public abstract void handle(I memNode, C context);
```
MarcosZyk, line 38: Maybe we should define an interface named ```IMemNode```, and state some methods in it.

lsm/src/main/java/org/apache/iotdb/lsm/wal/WALReader.java
```diff
@@ -0,0 +1,99 @@
...
+      // first clone the object through the prototype
+      nextRecord = prototype.clone();
+      // then perform deserialization and assign a value to the new object
+      nextRecord.deserialize(logStream);
+    } catch (EOFException e) {
+      logger.info(e.getMessage());
+      return false;
+    } catch (IOException e) {
+      logger.warn(e.getMessage());
+      fileCorrupted = true;
```
MarcosZyk, line 79: When file corrupted, we should recognize whether it is the middle of file or just the tail of the file. 

If is tail, it maybe the last log is not completed yet, and we can truncate the corrupted file tail directly. 

If is middle, it maybe complex and we should notice the user that the file is corrupted due to unknown cause.

schema-engine-tag/src/main/java/org/apache/iotdb/db/metadata/tagSchemaRegion/TagSchemaConfig.java
```diff
@@ -0,0 +1,55 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iotdb.db.metadata.tagSchemaRegion;
+
+/** tag schema region config */
+public class TagSchemaConfig {
```
MarcosZyk, line 22: Move this to package ```config```.

schema-engine-tag/src/main/java/org/apache/iotdb/db/metadata/tagSchemaRegion/TagSchemaDescriptor.java
```diff
@@ -0,0 +1,93 @@
...
+import java.io.BufferedInputStream;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+
+/** manager tag schema config */
+public class TagSchemaDescriptor {
```
MarcosZyk, line 35: Move this to package ```config```.

schema-engine-tag/src/main/java/org/apache/iotdb/db/metadata/tagSchemaRegion/deviceidlist/AppendOnlyDeviceIDListFileManager.java
```diff
@@ -0,0 +1,116 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+
+/** for append-only writing of device id list to disk */
+public class AppendOnlyDeviceIDListFileManager {
```
MarcosZyk, line 34: It seems we can recover the ```DeviceIDList``` by redo the file of IDTable.

MarcosZyk (commented):
PTAL.

schema-engine-tag/src/main/java/org/apache/iotdb/lsm/wal/WALRecord.java
```diff
@@ -0,0 +1,45 @@
...
+ * under the License.
+ */
+package org.apache.iotdb.lsm.wal;
+
+import java.io.DataInputStream;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+
+/** represents a wal record, which can be extended to implement more complex wal records */
+public interface WALRecord extends Cloneable {
```
MarcosZyk, line 26: Rename to ```IWALRecord```

server/src/assembly/resources/conf/iotdb-datanode.properties
```diff
@@ -1070,10 +1070,23 @@ timestamp_precision=ms
...
+
+# Choose the device id transformation method. The value could be Plain and SHA256. If the provided value doesn't match any pre-defined value, Plain will be used as default.
+# Datatype: string
+# device_id_transformation_method=Plain
+
 ####################
 ### Schema Engine Configuration
 ####################
-# Choose the mode of schema engine. The value could be Memory,Schema_File and Rocksdb_based. If the provided value doesn't match any pre-defined value, Memory mode will be used as default.
+# Choose the mode of schema engine. The value could be Memory,Schema_File,Rocksdb_based and Tag. If the provided value doesn't match any pre-defined value, Memory mode will be used as default.
```
MarcosZyk, line 1089: Hide this until ```IDTable``` and ```TagSchemaRegion``` features are ready for release.

schema-engine-tag/src/main/java/org/apache/iotdb/lsm/engine/LSMEngineBuilder.java
```diff
@@ -0,0 +1,108 @@
...
+    lsmEngine.setRecoverManager(recoverManager);
+    return this;
+  }
+
+  public LSMEngineBuilder<T> buildRootMemNode(T rootMemNode) {
+    lsmEngine.setRootMemNode(rootMemNode);
+    return this;
+  }
+
+  public LSMEngine<T> builder() {
```
MarcosZyk, line 105: Rename to ```build()```

schema-engine-tag/src/main/java/org/apache/iotdb/lsm/engine/LSMEngine.java
```diff
@@ -0,0 +1,170 @@
...
+    this.walManager = walManager;
+  }
+
+  public RecoverManager<LSMEngine<T>> getRecoverManager() {
+    return recoverManager;
+  }
+
+  public void setRecoverManager(RecoverManager<LSMEngine<T>> recoverManager) {
+    this.recoverManager = recoverManager;
+  }
```
MarcosZyk, line 155: Should these methods be package default only for LSMEngineBuilder usage?

schema-engine-tag/src/main/java/org/apache/iotdb/lsm/manager/ILSMManager.java
```diff
@@ -0,0 +1,51 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iotdb.lsm.manager;
+
+import org.apache.iotdb.lsm.context.RequestContext;
+
+// used to implement lsm manager
+public interface ILSMManager<T, R, C extends RequestContext> {
```
MarcosZyk, line 24: The comments on interface should define the behavior of the interface and describe the usage of this interface. The concrete implementation is not this interface's concern.

schema-engine-tag/src/main/java/org/apache/iotdb/lsm/request/Request.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+package org.apache.iotdb.lsm.request;
+
+import org.apache.iotdb.lsm.context.RequestContext;
+
+import java.util.List;
+
+public abstract class Request<K, V, R> {
+
+  RequestType requestType;
```
MarcosZyk, line 27: Since these bunch of abstract class ```Request``` and ```XXXRequest``` are stateless, make them as interface and it will be more flexible. The attribute ```requestType``` could be eliminated by setting default implementation of method ```getRequestType()```.

schema-engine-tag/src/main/java/org/apache/iotdb/lsm/property/Property.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.util.List;
+
+public class Property {
+
+  List<String> insertionLevelProcessClass;
+
+  List<String> deletionLevelProcessClass;
+
+  List<String> queryLevelProcessClass;
```
MarcosZyk, line 29: According to my understanding, this class is used for holding some strategy info of LSMEngine execution. Maybe we can integrate this into LSMEngineBuilde.

schema-engine-tag/src/main/java/org/apache/iotdb/lsm/engine/LSMEngineDirector.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.iotdb.lsm.request.QueryRequest;
+import org.apache.iotdb.lsm.request.Request;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.lang.reflect.InvocationTargetException;
+import java.util.List;
+
+public class LSMEngineDirector<T> {
```
MarcosZyk, line 40: Should this class be combined with LSMEngineBuilder?

schema-engine-tag/src/main/java/org/apache/iotdb/lsm/engine/LSMEngineDirector.java
```diff
@@ -0,0 +1,103 @@
...
+
+  private <O, R extends Request, C extends RequestContext>
+      ILevelProcess<T, O, R, C> generateLevelProcess(String className)
+          throws ClassNotFoundException, NoSuchMethodException, InvocationTargetException,
+              InstantiationException, IllegalAccessException {
+    Class c = Class.forName(className);
+    ILevelProcess<T, O, R, C> result =
+        (ILevelProcess<T, O, R, C>) c.getDeclaredConstructor().newInstance();
+    return result;
+  }
```
MarcosZyk, line 102: Using reflect to get specific class instance is not recommended.

MarcosZyk (commented):
PTAL

schema-engine-tag/src/main/java/org/apache/iotdb/lsm/engine/ILSMEngine.java
```diff
@@ -0,0 +1,52 @@
...
+import org.apache.iotdb.lsm.request.QueryRequest;
+
+import java.io.IOException;
+
+public interface ILSMEngine {
+
+  <K, V, R> void insert(InsertionRequest<K, V, R> insertionRequest) throws Exception;
+
+  <K, V, R> void insert(
+      InsertionRequest<K, V, R> insertionRequest, InsertRequestContext insertRequestContext)
```
MarcosZyk, line 35: Add java docs

schema-engine-tag/src/main/java/org/apache/iotdb/lsm/levelProcess/ILevelProcess.java
```diff
@@ -0,0 +1,41 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iotdb.lsm.levelProcess;
+
+import org.apache.iotdb.lsm.context.RequestContext;
+
+/** the processing method corresponding to each layer of memory nodes */
+public interface ILevelProcess<I, O, R, C extends RequestContext> {
```
MarcosZyk, line 24: rename to ```ILevelProcessor```

jun0315:
inflxudb-protocol module LGTM, Thanks for your contribution!

jun0315 (approved):

schema-engine-tag/src/main/java/org/apache/iotdb/db/metadata/tagSchemaRegion/tagIndex/Request/DeletionRequest.java
```diff
@@ -0,0 +1,69 @@
...
+public class DeletionRequest implements IDeletionRequest<String, Integer, RequestStatus> {
+
+  // tags
+  List<String> keys;
+
+  // int32 id
+  int value;
+
+  // deletion result
+  RequestStatus requestStatus;
```
MarcosZyk, line 36: The result or response should be hold and maintained in context rather request.

schema-engine-tag/src/main/java/org/apache/iotdb/lsm/annotation/DeletionProcessor.java
```diff
@@ -0,0 +1,36 @@
...
+ * Indicates that the class using this annotation is a DeleteLevelProcessor, and the attribute level
+ * of the annotation indicates which layer of memory node the deletion method of the
+ * DeleteLevelProcessor is.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface DeletionProcessor {
+  // level of the DeleteLevelProcessor
+  int level() default -1;
+}
```
MarcosZyk, line 36: Nice trial. The annotation does help simplify the code. However, is there any performance analysis to reveal that it is acceptable?

schema-engine-tag/src/main/java/org/apache/iotdb/db/metadata/tagSchemaRegion/tagIndex/Request/RequestStatus.java
```diff
@@ -0,0 +1,41 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iotdb.db.metadata.tagSchemaRegion.tagIndex.Request;
+
+/** Indicates the status of the request, whether it is successful after the request is processed */
+public class RequestStatus {
+
+  Boolean success;
+
```
MarcosZyk, line 25: Rename to Response or ExecutionResult or something like this.

schema-engine-tag/src/main/java/org/apache/iotdb/db/metadata/tagSchemaRegion/tagIndex/TagInvertedIndex.java
```diff
@@ -0,0 +1,193 @@
...
+  public synchronized void addTags(Map<String, String> tags, int id) {
+    try {
+      for (Map.Entry<String, String> tag : tags.entrySet()) {
+        InsertionRequest insertionRequest =
+            new InsertionRequest(generateKeys(tag.getKey(), tag.getValue()), id);
+        lsmEngine.insert(insertionRequest);
+      }
+    } catch (Exception e) {
+      logger.error(e.getMessage());
+    }
```
MarcosZyk, line 108: There's no exception thrown in the interface definition. Please polish the exception handler mechanism. The exception shall be thrown or passed in execution result.

MarcosZyk (commented):
Excellent work. Only few problems left. PTAL.

schema-engine-tag/src/main/java/org/apache/iotdb/db/metadata/tagSchemaRegion/tagIndex/Request/DeletionRequest.java
KeeProMise: We define the delete method of ILSMEngine as:
`<K, V, R> void delete(IDeletionRequest<K, V, R> deletionRequest);`
If you use context to save the operation result or response, is it necessary to rely on a DeletionRequestContext object inside the DeletionRequest, as well as InsertionRequest and QueryRequest？

Or the delete method can be defined as:
`<K, V,R> R delete(IDeletionRequest<K, V> deletionRequest);`
Which do you think is better, or do you have any other ideas?

schema-engine-tag/src/main/java/org/apache/iotdb/lsm/annotation/DeletionProcessor.java
MarcosZyk: Got it.

schema-engine-tag/src/main/java/org/apache/iotdb/db/metadata/tagSchemaRegion/tagIndex/Request/DeletionRequest.java
MarcosZyk: I think the second one is better. The Request shall not reply on the Response. Request and Response are both defined for and be relied on by the interface.

KeeProMise: ok, i think the same as you

KeeProMise: I understand that I save the response result in the context because the response needs to be constructed or updated in multi execution stages, and then modify the insert, delete, query interface methods in ILSMEngine to return the response instead of void

MarcosZyk (approved):
LGTM. Thanks for your all contribution.

PR #7428, opened 2022-09-26
[IOTDB-4519] Fix existing measurement check during auto creating schema

MarcosZyk:
### Cause

The consensus layer only supports passing message by TSStatus when executing writing operation. The existing measurement info can only be passed by string format in TSStatus. The transformation between string and bytes won't work if the encoding method use double-byte charsets. 

### Solution

When transform measurementPath to String, use single-byte charsets.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/ClusterSchemaFetcher.java
+++ b/server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/ClusterSchemaFetcher.java
@@ -27,7 +27,6 @@ import org.apache.iotdb.db.conf.IoTDBConfig;
 import org.apache.iotdb.db.conf.IoTDBDescriptor;
 import org.apache.iotdb.db.metadata.cache.DataNodeSchemaCache;
 import org.apache.iotdb.db.metadata.path.MeasurementPath;
-import org.apache.iotdb.db.metadata.path.PathDeserializeUtil;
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/ClusterSchemaFetcher.java
```diff
@@ -476,9 +474,7 @@ private List<MeasurementPath> executeInternalCreateTimeseriesStatement(
     for (TSStatus subStatus : executionResult.status.subStatus) {
       if (subStatus.code == TSStatusCode.MEASUREMENT_ALREADY_EXIST.getStatusCode()) {
         alreadyExistingMeasurements.add(
-            (MeasurementPath)
-                PathDeserializeUtil.deserialize(
-                    ByteBuffer.wrap(subStatus.getMessage().getBytes())));
+            MeasurementPath.parseDataFromString(subStatus.getMessage()));
```
wangchao316, line 477: subStatus.getMessage()  is measurementPathData ?

MarcosZyk: Yes. When trying auto creating schema, the timeseries may be concurrently created by other threads or requests. The subStatus takes the existing schema info back without further schema fetch query.

wangchao316 (approved):
LGTM

HTHou (approved):
LGTM

PR #7433, opened 2022-09-26
[IOTDB-4517] Fix cross_space and unseq_space compaction are slower at common timeseries and template timeseries

THUMarkLau:
See [IOTDB-4517](https://issues.apache.org/jira/browse/IOTDB-4517).

The fisrt reason is that the `currentChunkSize` does not reset when flushing chunk metadata to temp file in disk, leading to frequent flushing of chunk metadata.
The second reason is the concurrent issue caused by update to the start time and end time of tsfile resource in the compaction sub-task.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/compaction/performer/impl/ReadPointCompactionPerformer.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/compaction/performer/impl/ReadPointCompactionPerformer.java
@@ -199,11 +199,12 @@ public class ReadPointCompactionPerformer
     if (subTaskNums > 0) {
       // assign the measurements for each subtask
       List<String>[] measurementListArray = new List[subTaskNums];
-      for (int i = 0; i < allMeasurements.size(); ++i) {
-        if (measurementListArray[i % subTaskNums] == null) {
-          measurementListArray[i % subTaskNums] = new LinkedList<>();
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/apache/iotdb/db/engine/compaction/performer/impl/ReadPointCompactionPerformer.java
```diff
@@ -190,22 +191,30 @@ private void compactNonAlignedSeries(
...
-    compactionWriter.startChunkGroup(device, false);
-    for (int taskCount = 0; taskCount < allMeasurements.size(); ) {
+    if (subTaskNums > 0) {
+      // assign the measurements for each subtask
+      List<String>[] measurementListArray = new List[subTaskNums];
+      for (int i = 0; i < allMeasurements.size(); ++i) {
+        if (measurementListArray[i % subTaskNums] == null) {
+          measurementListArray[i % subTaskNums] = new LinkedList<>();
+        }
+        measurementListArray[i % subTaskNums].add(allMeasurements.get(i));
```
qiaojialin, line 206:
```suggestion
        int index = i % subTaskNums
        if (measurementListArray[index] == null) {
          measurementListArray[index] = new LinkedList<>();
        }
        measurementListArray[index].add(allMeasurements.get(i));
```

qiaojialin (approved):

PR #7472, opened 2022-09-28
[IOTDB-4062] Add schema write check above consensus layer

MarcosZyk:
### Motivation

Failing operation shall not persist in raft log.

### Modification

1. refactor region operation execution by setting up concentrate RegionReadExecutor and RegionWriteExecutor and both FragmentInstanceDispatcher and DataNodeInternalRpcService relying on it.
2. add measurement existence check before executing timeseries creation in consensus layer.


server/src/main/java/org/apache/iotdb/db/mpp/execution/executor/RegionWriteExecutor.java
```diff
@@ -0,0 +1,470 @@
...
+        if (hasFailedMeasurement) {
+          partialInsertMessage =
+              String.format(
+                  "Fail to insert measurements %s caused by %s",
+                  insertNode.getFailedMeasurements(), insertNode.getFailedMessages());
+          LOGGER.warn(partialInsertMessage);
+        }
+
+        ConsensusWriteResponse writeResponse =
+            executePlanNodeInConsensusLayer(context.getRegionId(), insertNode);
```
xingtanzjr, line 177: We can use `DataRegionConsensusImpl.getInstance().write(groupId, planNode)` to avoid ambiguity because current method is only used to process the PlanNode for DataRegion

MarcosZyk: Fixed.

xingtanzjr (approved):

PR #7480, opened 2022-09-29
Accelerate restart proces

JackieTien97:
We found that when memory is limited and there are too many TsFileResources, during restart prcessing, there are too much time spent on deserializing devices from TsFile, because `ITimeIndex` degrade to `FileTimeIndex`.
Dive into the restart process, we found that deserializing devices is for updating flush time of each device, so we can improve restarting by updating it before TsFileResource being added to `TsFileResourceManagement` when it can be degraded.
image


HeimingZ (approved):
LGTM

xingtanzjr (approved):

PR #7490, opened 2022-09-30
[IOTDB-4554] Start compaction only when IoTDB is fully recovered

THUMarkLau:
See [IOTDB-4554](https://issues.apache.org/jira/browse/IOTDB-4554).

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/service/DataNode.java
+++ b/server/src/main/java/org/apache/iotdb/db/service/DataNode.java
@@ -273,7 +273,6 @@ public class DataNode implements DataNodeMBean {
     registerManager.register(new JMXService());
     registerManager.register(FlushManager.getInstance());
     registerManager.register(CacheHitRatioMonitor.getInstance());
-    registerManager.register(CompactionTaskManager.getInstance());
     JMXService.registerMBean(getInstance(), mbeanName);
 
     // close wal when using ratis consensus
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/apache/iotdb/db/service/DataNode.java
```diff
@@ -250,6 +252,7 @@ private void active() throws StartupException {
       logger.error("Meet error while starting up.", e);
       throw new StartupException("Error in activating IoTDB DataNode.");
     }
+    activated = true;
```
yifuzhou, line 255: It seems that activated cannot turn false when deactivate the datanode.

yifuzhou (commented):
Hi, PTAL~

server/src/main/java/org/apache/iotdb/db/service/DataNode.java
THUMarkLau: Fixed, thank you.

server/src/main/java/org/apache/iotdb/db/service/DataNode.java
```diff
@@ -316,6 +315,7 @@ private void setUp() throws StartupException, QueryProcessException {
     registerManager.register(RegionMigrateService.getInstance());
 
     registerManager.register(MetricService.getInstance());
+    registerManager.register(CompactionTaskManager.getInstance());
```
xingtanzjr, line 318: As we change the sequence of invoking `CompactionTaskManager.getInstance()`, why do we need to add the variable `init` inside the class ?

xingtanzjr (approved):

PR #7513, opened 2022-10-04
Fix concurrent timeseries auto creation exception

MarcosZyk:
## Description

When concurrently creating timeseries, the existing timeseries schema should be taken for usage rather than throw MeasurementAlreadyExistsException.


Caideyipi:
Test successful. Able to merge.

PR #7521, opened 2022-10-05
Enhance DTW to support time series with different lengths

lizhizhou:
## Description
Current DTW UDF only supports time series with the same length. All row with null value is ignored.
This is not the common case that DTW will be used. DTW is usually used to check the distance between time series with different lengths.
This PR is to enhance DTW to support this case. 

### Content3 ...

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


HTHou:
@trin1t PTAL

PR #7527, opened 2022-10-06
[IOTDB-3656] mpp load supports tsfile with non-standard name

yschengzi:
- [x] support load .mods
- [ ] last cache
- [x] clean up
- [x] tsfile name restriction

solution:
assume two TsFiles with name tsfile1, and tsfile2
if tsfile1 and tsfile2 both have standard TsFile Name (i.e. createtime-version-mergecnt-mergecnt)
then compare two TsFiles with standard methods (i.e. compare createtime firstly, and then compre version.)
else compare two file name with String.compareTo


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
```diff
@@ -963,6 +963,20 @@ public static int compareFileName(TsFileResource o1, TsFileResource o2) {
...
+    if (timeDiff != 0) {
+      return timeDiff < 0 ? -1 : 1;
+    }
+    long versionDiff = tsFileName1.getVersion() - tsFileName2.getVersion();
+    if (versionDiff != 0) {
+      return versionDiff < 0 ? -1 : 1;
+    }
+    return 0;
+  }
+
```
HTHou, line 979: Add some annotations. 
It is hard to understand the differences of `compareFileName(TsFileResource o1, TsFileResource o2)` method.

HTHou (changes requested):

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
yschengzi: done.

HTHou (approved):

PR #7529, opened 2022-10-06
[IOTDB-3656] mpp load supports last cache

yschengzi:
- [x] support load .mods
- [x] last cache
- [x] clean up
- [x] tsfile name restriction

solution:
when loading tsfiles successfully, it will clean up all schema cache in datanode.


HTHou:
please fix the CI.

HTHou (approved):

PR #7537, opened 2022-10-07
[To rel/0.13] [IOTDB-4579] Remove the update of start time and end time in the target file after cross space compaction

THUMarkLau:
See [IOTDB-4579](https://issues.apache.org/jira/browse/IOTDB-4579).


MrQuansy (approved):
LGTM

PR #7548, opened 2022-10-09
[IOTDB-4427]Optimize the implementation of pending batch in MultiLeader

SpriCoder:
Optimize the implementation of pending batch: modify constructor method.


xingtanzjr:
It seems that this change won't slow down the write speed of MultiLeader.
But... I tested this code in 41-43 for serveral times, the results in each test are not stable.

Let's merge this PR until the test shows it is ok.

PR #7550, opened 2022-10-09
Make sure FI won't be aborted by timeout checker by mistake

JackieTien97:
I found a strange thing in CI failure.
image
The root cause is that the FI is aborted by `DriverTaskTimeoutSentinelThread`. I guess it's caused by Thread.sleep() may not  sleep that long, so I add a while loop and some debug logs if it happens again after I add this while loop.
image


ericpai (approved):

server/src/main/java/org/apache/iotdb/db/mpp/execution/fragment/FragmentInstanceExecution.java
```diff
@@ -21,17 +21,22 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import static java.util.Objects.requireNonNull;
 import static org.apache.iotdb.db.mpp.execution.fragment.FragmentInstanceState.FAILED;
 
 public class FragmentInstanceExecution {
 
+  private static final Logger LOGGER =
+      LoggerFactory.getLogger(DriverTaskTimeoutSentinelThread.class);
```
HTHou, line 39:
```suggestion
  private static final Logger LOGGER =
      LoggerFactory.getLogger(FragmentInstanceExecution.class);
```

PR #7553, opened 2022-10-10
[To rel/0.13] Fix PathAlreadyExistException during concurrent auto creating aligned timeseries

MarcosZyk:
### Problem 

see https://github.com/apache/iotdb/issues/7552

### Modification

The already existing timeseries shall directly be used in schema validation rather than causing a PathAlreadyExistException, thus ignore the exception generated by concurrent creation.


DanielWang2035 (approved):
LGTM

HTHou (approved):

PR #7575, opened 2022-10-12
[IOTDB-4589] Register/recover trigger when registering/restart a datanode

lancelly:
When registering/restart a datanode, we need to register/recover the trigger information that has been stored in the cluster. The process flow can be found at: https://apache-iotdb.feishu.cn/docx/doxcneSc8w4eRHntxGkH5jpMgfe
Briefly speaking, the main process is: 
1. Get trigger-table from confignode.
2. Get needed jar files from confignode.
3. Register triggers on the datanode and create trigger instances.
The test doc can be found at:  https://apache-iotdb.feishu.cn/docx/doxcn9uRj5bpzb58MqgeMokjMC0. Please refer to the DataNode part.

This pr is done with the help of @Wei-hao-Li ~


server/src/main/java/org/apache/iotdb/db/service/DataNode.java
```diff
@@ -397,6 +418,102 @@ private void initTriggerRelatedInstance() throws StartupException {
...
+      List<String> jarNameList =
+          triggerInformationList.stream()
+              .map(TriggerInformation::getJarName)
+              .collect(Collectors.toList());
+      TGetTriggerJarResp resp = configNodeClient.getTriggerJar(new TGetTriggerJarReq(jarNameList));
+      if (resp.getStatus().getCode() == TSStatusCode.EXECUTE_STATEMENT_ERROR.getStatusCode()) {
+        throw new StartupException("Failed to get trigger jar from config node.");
+      }
+      List<ByteBuffer> jarList = resp.getJarList();
+      for (int i = 0, n = triggerInformationList.size(); i < n; i++) {
```
Beyyes, line 473:
```suggestion
      for (int i = 0; i < triggerInformationList.size(); i++) {
``` is tidy

File: server/src/main/java/org/apache/iotdb/db/service/DataNode.java [hunk 2]
```diff
@@ -134,6 +147,8 @@ protected void doAddNode() {
       prepareDataNode();
       // register current DataNode to ConfigNode
       registerInConfigNode();
+      // get resources for trigger,udf...
+      prepareResources();
```
Beyyes, line 151: Do we have any tests on real cluster? I'm not sure if there exists problems if put `prepareResources()` ahead of `active()` and `setUpRPCService()`

server/src/main/java/org/apache/iotdb/db/service/DataNode.java
```diff
@@ -101,6 +109,11 @@ private DataNode() {
   private static final RegisterManager registerManager = new RegisterManager();
   public static ServiceProvider serviceProvider;
 
+  /** store the list when registering in config node for preparing trigger related resources */
+  private List<TriggerInformation> triggerInformationList;
```
Beyyes, line 113: Maybe it's better to maintain the `triggerInformationList` (informations about trigger) in an independent class? Because in the future CQ and other features are integrated, the variables amount of `DataNode` maybe large.

server/src/main/java/org/apache/iotdb/db/service/DataNode.java
```diff
@@ -397,6 +418,102 @@ private void initTriggerRelatedInstance() throws StartupException {
...
+      for (int i = 0, n = triggerInformationList.size(); i < n; i++) {
+        TriggerExecutableManager.getInstance()
+            .writeToLibDir(jarList.get(i), triggerInformationList.get(i).getJarName());
+      }
+    } catch (IOException | TException e) {
+      throw new StartupException(e);
+    }
+  }
+
+  private List<TriggerInformation> getTriggerNeedJarList() {
```
Beyyes, line 482:
```suggestion
  private List<TriggerInformation> getJarListForTrigger() {
```
is better?

server/src/main/java/org/apache/iotdb/db/service/DataNode.java
lancelly: I have tested it manually and I think it is ok? I moved this method into active() now.

server/src/main/java/org/apache/iotdb/db/service/DataNode.java
lancelly: Got it.

server/src/main/java/org/apache/iotdb/db/service/DataNode.java
lancelly: done.

server/src/main/java/org/apache/iotdb/db/service/DataNode.java
lancelly: done.

Beyyes (approved):
LGTM

PR #7585, opened 2022-10-13
[IOTDB-4643] Add Unknown_DataNode_Detector

Wei-hao-Li:
This PR add new Thread 'Unknown_DataNode_Detector' used to find the unknown DataNodes and activate Transfer of some Data, and some points need to pay attention:
1.If one DataNode is continuing Unknown, we will not activate Transfer of this Node continuously;
2.The selected DataNodes may not truly need to transfer, so we should ensure safety of the
 Data when implement transferMethod in Manager.


JackieTien97 (approved):

confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeDescriptor.java
```diff
@@ -225,6 +225,12 @@ private void loadProps() {
               properties.getProperty(
                   "heartbeat_interval", String.valueOf(conf.getHeartbeatInterval()))));
 
+      conf.setUnknownDataNodeDetectInterval(
+          Long.parseLong(
+              properties.getProperty(
+                  "unknown_datanode_detect_interval",
+                  String.valueOf(conf.getHeartbeatInterval()))));
```
CRZbulabula, line 232: Should we need to add another parameter? We should update iotdb-confignode.properties file if it's necessary to add this parameter.

confignode/src/main/java/org/apache/iotdb/confignode/manager/node/NodeManager.java
```diff
@@ -650,6 +662,69 @@ public Map<Integer, BaseNodeCache> getNodeCacheMap() {
...
+            DataNodeConfiguration -> {
+              TDataNodeLocation dataNodeLocation = DataNodeConfiguration.getLocation();
+              BaseNodeCache newestNodeInformation = nodeCacheMap.get(dataNodeLocation.dataNodeId);
+              if (newestNodeInformation != null) {
+                if (newestNodeInformation.getNodeStatus() == NodeStatus.Running) {
+                  oldUnknownNodes.remove(dataNodeLocation);
+                } else if (!oldUnknownNodes.contains(dataNodeLocation)
+                    && newestNodeInformation.getNodeStatus() == NodeStatus.Unknown) {
+                  newUnknownNodes.add(dataNodeLocation);
+                }
```
CRZbulabula, line 703: Is it better to use NodeStatus.isNormalStatus() instead of only consider Running and Unknown status?

CRZbulabula (commented):
PTAL～

CRZbulabula (approved):
LGTM~

PR #7591, opened 2022-10-13
[IOTDB-4623] Enhance the stability of remove-confignode.sh when executed in non-running node

Beyyes:
Description
see more details in https://issues.apache.org/jira/browse/IOTDB-4623

mainly do this improvement, remove-confignode.sh rpc transfer to leader confignode id firstly

image


confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeRemoveCheck.java
```diff
@@ -103,10 +110,15 @@ public void removeConfigNode(TConfigNodeLocation removedNode)
       if (status.getCode() == TSStatusCode.SUCCESS_STATUS.getStatusCode()) {
         break;
       }
+
+      if (status.getCode() == TSStatusCode.REMOVE_CONFIGNODE_FAILED.getStatusCode()) {
+        break;
```
wangchao316, line 115: print a log in failed situation

xingtanzjr (approved):

confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeRemoveCheck.java
Beyyes: i will fix it in next pr

PR #7594, opened 2022-10-13
Fix cycle initializing IoTDBDescriptor

ericpai:
See these error logs during starting.

image

This is a bug caused by #7577 

The root cause is that `IoTDBDescriptor` is a singleton and when the constructor is called, `loadProperties()` is called. However, there's a call of `IoTDBDescriptor.getInstance()` in `convertMilliTimeWithPrecision()`. As the initializing hasn't been finished, the `getInstance()` just returns `null`.

`DatetimeUtils` is a static class, so it's better not to depend on dynamic components, just decouples with `IoTDBDescriptor` through a method parameter.


ericpai:
@HTHou @HeimingZ PTAL

PR #7600, opened 2022-10-14
[IOTDB-4647] Fix CSV import error when import header name with type

RYH61:
see [https://issues.apache.org/jira/browse/IOTDB-4647](https://issues.apache.org/jira/browse/IOTDB-4647)

fix
图片


cross-tests/src/test/java/org/apache/iotdb/cross/tests/tools/importCsv/ImportCsvTestIT.java
```diff
@@ -256,7 +257,7 @@ public void testWithoutCreateSchema() throws IOException, ClassNotFoundException
    * @throws IOException
    * @throws ClassNotFoundException
    */
-  @Test
```
HTHou, line 259: Why ignore these tests?

RYH61: There are still problems here. I'm fixing them

HTHou (approved):

PR #7601, opened 2022-10-14
[IOTDB-4648] Fix bug ConsensusManager won't initialize when restart

CRZbulabula:


confignode/src/main/java/org/apache/iotdb/confignode/manager/ConsensusManager.java
```diff
@@ -57,12 +57,12 @@ public class ConsensusManager {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(ConsensusManager.class);
   private static final ConfigNodeConfig CONF = ConfigNodeDescriptor.getInstance().getConf();
+  private static final int SEED_CONFIG_NODE_Id = 0;
```
Beyyes, line 60:
```suggestion
  private static final int SEED_CONFIG_NODE_ID = 0;
```

PR #7610, opened 2022-10-14
[IOTDB-4641] New Standalone Sync Receiver TsFile Loader Implement

yschengzi:
Implement TsFile Loader for New Standalone Sync Receiver.

Using mpp load to handle TsFilePipeData.


Cpaulyz (approved):
LGTM

server/src/test/java/org/apache/iotdb/db/sync/transport/SyncTransportTest.java
```diff
@@ -70,6 +71,7 @@
 import java.util.Collections;
 import java.util.List;
 
+@Ignore
```
HTHou, line 74: Better to move it to IT package.

Cpaulyz: > Better to move it to IT package.

This is a white-box end-to-end test that cannot be implemented well in IT. We temporarily ignore it, keep the test cases and will re-implement it in the white-box UT test when the standalone architecture is removed.

HTHou (approved):

PR #7626, opened 2022-10-16
Refactor IT classes and packages

ericpai:
This PR fixes:
1. Bad regexp that causing all IT files are all ignored by the checkstyle tool, which is imported in the PR #7326 
2. Move session IT cases and confignode cases to a unify package to reduce the complexity of the regexp, and add particular import-control rules for each package.
3. Add a checker for checking whether a test class has been annotated with `@RunWith(IoTDBTestRuner.class)`.
4. Add all the missing  `@RunWith(IoTDBTestRuner.class)`.


ericpai:
@JackieTien97 @Beyyes PTAL

JackieTien97 (approved):

PR #7636, opened 2022-10-17
[IOTDB-4647] Fix CSV import error when import header name with type

RYH61:
see: [https://github.com/apache/iotdb/pull/7600](https://github.com/apache/iotdb/pull/7600)

### DIFF
--- a/cli/src/main/java/org/apache/iotdb/tool/ImportCsv.java
+++ b/cli/src/main/java/org/apache/iotdb/tool/ImportCsv.java
@@ -587,11 +587,11 @@ public class ImportCsv extends AbstractCsvTool {
           AtomicReference<Boolean> isFail = new AtomicReference<>(false);
 
           // read data from record
-          for (String measurement : headerNameMap.values()) {
-            String value = record.get(measurement);
+          for (Map.Entry<String, String> headerName : headerNameMap.entrySet()) {
+            String value = record.get(headerName.getValue());
             if (!"".equals(value)) {
               TSDataType type;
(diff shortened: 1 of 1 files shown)


HTHou (approved):

neuyilan (changes requested):
todo

cli/src/main/java/org/apache/iotdb/tool/ImportCsv.java
```diff
@@ -587,11 +587,11 @@ private static void writeDataAlignedByDevice(
           AtomicReference<Boolean> isFail = new AtomicReference<>(false);
 
           // read data from record
-          for (String measurement : headerNameMap.values()) {
-            String value = record.get(measurement);
+          for (Map.Entry<String, String> headerName : headerNameMap.entrySet()) {
+            String value = record.get(headerName.getValue());
```
neuyilan, line 591: It's better to use two variables to indict the headerName.getValue() and headerName.getKey(), this would be more readble.

eg：

`Map.Entry<String, String> headerNameEntry : headerNameMap.entrySet()

// headerNameWithoutType is equal to headerName if the CSV coloum do not have data type.
String  headerNameWithoutType =  headerNameEntry.getKey();
String  headerName =  headerNameEntry.getValue(); `

neuyilan (approved):

cli/src/main/java/org/apache/iotdb/tool/ImportCsv.java
RYH61: fix~

PR #7643, opened 2022-10-18
[IOTDB-4627]Trigger transfer

Wei-hao-Li:
1.Transfer stateful triggers when DataNode Status is Unknown;
2.Transfer stateful triggers when DataNode remove;
3.Has been tested, can see [Trigger测试方案](https://apache-iotdb.feishu.cn/docx/doxcn9uRj5bpzb58MqgeMokjMC0)
4.Cooperate with @lancelly.


confignode/src/main/java/org/apache/iotdb/confignode/manager/node/NodeManager.java
```diff
@@ -255,8 +255,17 @@ public DataSet removeDataNode(RemoveDataNodePlan removeDataNodePlan) {
...
+
     DataNodeToStatusResp dataSet = new DataNodeToStatusResp();
+    // do transfer of the DataNodes before remove
+    if (configManager.transfer(removeDataNodePlan.getDataNodeLocations()).getCode()
+        != TSStatusCode.SUCCESS_STATUS.getStatusCode()) {
+      dataSet.setStatus(
+          new TSStatus(TSStatusCode.NODE_DELETE_FAILED_ERROR.getStatusCode())
+              .setMessage("Fail to do transfer of the DataNodes"));
+      return dataSet;
+    }
```
CRZbulabula, line 267: Is the `transfer` process should be such important that the DataNode remove process couldn't continue as long as if it fails down?

CRZbulabula (commented):
PTAL～

confignode/src/main/java/org/apache/iotdb/confignode/manager/node/NodeManager.java
Wei-hao-Li: Yes, I think we should ensure the safety of data.

CRZbulabula (approved):
LGTM

JackieTien97 (approved):

confignode/src/main/java/org/apache/iotdb/confignode/manager/TriggerManager.java
```diff
@@ -132,8 +146,79 @@ public TGetTriggerJarResp getTriggerJar(TGetTriggerJarReq req) {
...
+    } finally {
+      triggerInfo.releaseTriggerTableLock();
+    }
+
     return new TSStatus(TSStatusCode.SUCCESS_STATUS.getStatusCode());
   }
+
+  public List<TSStatus> updateTriggerLocation(
+      String triggerName,
+      TDataNodeLocation dataNodeLocation,
```
Beyyes, line 213: parameter `dataNodeLocation` and `dataNodeLocationMap` are similar, maybe we can change `dataNodeLocation` to `targetDataNode`?

confignode/src/main/java/org/apache/iotdb/confignode/manager/node/NodeManager.java
Beyyes: Remove data process used RemoveDataProcedure and sub RegionMigrateProcedures, 
Will triggers implement it later or using other methods?
Or some errors may happen when ConfigNode down/restart.

image

Wei-hao-Li: We used to consider implement by Procedure, but later we found that we did not need to roll back when the transfer failed.

PR #7646, opened 2022-10-18
[IOTDB-4666]Get the devices on one leaf node at a time while compacting

choubenson:
At present, the system obtains all devices in the source file at one time when compacting, which may cause OOM problems in the scenario of a large number of devices.  
To solve this problem, we should read the devices on one device leaf node at a time instead of reading all devices from files at one time.


tsfile/src/main/java/org/apache/iotdb/tsfile/read/TsFileDeviceIterator.java
```diff
@@ -25,19 +25,28 @@
...
 public class TsFileDeviceIterator implements Iterator<Pair<String, Boolean>> {
   private final TsFileSequenceReader reader;
+
+  // device -> firstMeasurmentNode offset
   private final Queue<Pair<String, Pair<Long, Long>>> queue;
   private Pair<String, Boolean> currentDevice = null;
   private MetadataIndexNode measurementNode;
 
+  // <startOffset, endOffset>, device leaf node offset in this file
+  private final List<Pair<Long, Long>> leafDeviceNodeOffsetList;
```
Alima777, line 41: Try to replace Pair<Long, Long> with 'long[2]'

tsfile/src/main/java/org/apache/iotdb/tsfile/read/TsFileSequenceReader.java
```diff
@@ -633,37 +633,108 @@ private List<String> getAllDevices(MetadataIndexNode metadataIndexNode) throws I
...
+   * Get devices and first measurement node offset.
+   *
+   * @param startOffset start offset of device leaf node
+   * @param endOffset end offset of device leaf node
+   * @param measurementNodeOffsetQueue device -> first measurement node offset
+   */
+  public void getDevicesAndEntriesOfOneLeafNode(
+      Long startOffset,
+      Long endOffset,
+      Queue<Pair<String, Pair<Long, Long>>> measurementNodeOffsetQueue)
```
Alima777, line 665: Use primitive type long instead of `Long` as far as possible

Alima777 (commented):
Some details needs to be modified.

tsfile/src/main/java/org/apache/iotdb/tsfile/read/TsFileSequenceReader.java
choubenson: Resolved. Use long[2] instead of Pair<Long,Long>.

tsfile/src/main/java/org/apache/iotdb/tsfile/read/TsFileDeviceIterator.java
choubenson: Resolved.

Alima777 (approved):
LGTM

PR #7658, opened 2022-10-19
[IOTDB-4692] SchemaFile adaptation for cluster metadata validation

Cpaulyz:
## Description

1. Support these two interface in SchemaRegionSchemaFileImpl:

* checkMeasurementExistence()
* fetchSchema() without template

2. IoTDBMetadataFetchIT add schemaFile mode test
3. Add UT for ISchemaRegion. All test cases will be run in both Memory and Schema_File modes. In Schema_File mode, there are three kinds of test environment: full memory, partial memory and non memory.


server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeBelowSGCachedImpl.java
```diff
@@ -331,6 +335,62 @@ public List<IMeasurementMNode> createAlignedTimeseries(
...
+    }
+    try {
+      if (!device.isEntity()) {
+        return Collections.emptyMap();
+      }
+      Map<Integer, MetadataException> failingMeasurementMap = new HashMap<>();
+      for (int i = 0; i < measurementList.size(); i++) {
+        try {
+          if (store.hasChild(device, measurementList.get(i))) {
+            IMNode node = store.getChild(device, measurementList.get(i));
```
MarcosZyk, line 355: Use ```node = store.getChild``` and judge ```node == null``` to avoid reading disk twice.

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeBelowSGCachedImpl.java
```diff
@@ -331,6 +335,62 @@ public List<IMeasurementMNode> createAlignedTimeseries(
...
+                    i,
+                    new MeasurementInBlackListException(
+                        devicePath.concatNode(measurementList.get(i))));
+              } else {
+                failingMeasurementMap.put(
+                    i,
+                    new MeasurementAlreadyExistException(
+                        devicePath.getFullPath() + "." + measurementList.get(i),
+                        node.getAsMeasurementMNode().getMeasurementPath()));
+              }
```
MarcosZyk, line 368: UnPin the child node after usage.

integration-test/src/test/java/org/apache/iotdb/db/it/schema/IoTDBMetadataFetchSchemaFileIT.java
```diff
@@ -0,0 +1,54 @@
...
+@RunWith(IoTDBTestRunner.class)
+@Category({LocalStandaloneIT.class, ClusterIT.class})
+public class IoTDBMetadataFetchSchemaFileIT extends IoTDBMetadataFetchIT {
+
+  protected static String schemaEngineMode;
+
+  @Before
+  public void setUp() throws Exception {
+    schemaEngineMode = ConfigFactory.getConfig().getSchemaEngineMode();
+    ConfigFactory.getConfig().setSchemaEngineMode(SchemaEngineMode.Schema_File.toString());
```
MarcosZyk, line 45: What about directly use string to modify the config rather than import ```SchemaEngineMode```?

MarcosZyk (commented):
PTAL

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeBelowSGCachedImpl.java
Cpaulyz: done

server/src/main/java/org/apache/iotdb/db/metadata/mtree/MTreeBelowSGCachedImpl.java
Cpaulyz: done

integration-test/src/test/java/org/apache/iotdb/db/it/schema/IoTDBMetadataFetchSchemaFileIT.java
Cpaulyz: Seems that using String directly does not satisfy the coding standard...

MarcosZyk: IT is aimed for simulate user behavior, and the enum ```SchemaEngineMode``` is just exposed as string in properties file.

Cpaulyz: done

MarcosZyk (approved):

yifuzhou (approved):
LGTM~

PR #7660, opened 2022-10-19
[IOTDB-4669] Scheduled task for checking trigger-table of DataNode

lancelly:
The information of stateful trigger can be invalid when concurrency problems occur, so we need to update information of stateful trigger periodcally.


server/src/main/java/org/apache/iotdb/db/trigger/service/TriggerInformationUpdater.java
```diff
@@ -0,0 +1,104 @@
...
+      LOGGER.info("Stateful-Trigger-Information-Updater is successfully started.");
+    }
+  }
+
+  public void stopTriggerInformationUpdater() {
+    if (updateFuture != null) {
+      updateFuture.cancel(false);
+      updateFuture = null;
+      LOGGER.info("Stateful-Trigger-Information-Updater is successfully stopped.");
+    }
```
JackieTien97, line 78:
```suggestion
    }
    triggerInformationUpdateExecutor.shutdown();
```

JackieTien97 (approved):

PR #7669, opened 2022-10-20
[IOTDB-4633] Fix bugs of longToBytes in BytesUtils of tsfile

FrankHWD:
Since I have met a problem of "unequal value after bytes transformation by BytesUtils in tsfile" as mentioned in IOTDB-4633. After modifying the right arithmetic shift (算术右移) to right logical shift (逻辑右移) in longToBytes function of BytesUtils in tsfile, the problem has been solved successfully.

### DIFF
--- a/tsfile/src/main/java/org/apache/iotdb/tsfile/utils/BytesUtils.java
+++ b/tsfile/src/main/java/org/apache/iotdb/tsfile/utils/BytesUtils.java
@@ -93,7 +93,7 @@ public class BytesUtils {
         width -= m;
         int mask = 1 << (8 - cnt);
         cnt += m;
-        byte y = (byte) (srcNum >> width);
+        byte y = (byte) (srcNum >>> width);
         y = (byte) (y << (8 - cnt));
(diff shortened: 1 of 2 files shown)


tsfile/src/main/java/org/apache/iotdb/tsfile/utils/BytesUtils.java
```diff
@@ -523,7 +523,7 @@ public static void longToBytes(long srcNum, byte[] result, int pos, int width) {
         width -= m;
         int mask = 1 << (8 - cnt);
         cnt += m;
-        byte y = (byte) (srcNum >> width);
+        byte y = (byte) (srcNum >>> width);
```
HTHou, line 526: Add a UT for this bug?

Besides, `intToBytes` seems have a same bug...

FrankHWD: The UT tests for this bug have been added.

HTHou (approved):

qiaojialin (approved):

PR #7672, opened 2022-10-20
New distribution package structure

HTHou:
## Description

Design:

https://apache-iotdb.feishu.cn/docx/doxcnFpCzjZadvMPQHUROW4mrTe

image
image
image


SpriCoder (changes requested):
Both DataNode and ConfigNode use metric module. If they both use `iotdb-metric.yml` to config metric module, there will be some problems, like trying to bind the same port of one machine, PTAL~

qiaojialin:
move datanode/confignode to a data folder

HTHou:
> move datanode/confignode to a data folder

fixed

qiaojialin (approved):

PR #7683, opened 2022-10-21
[IOTDB-4711] Bind DataNodeInternalService to correct address

CRZbulabula:
The reason for this error is that the DataNodeInternalService binds to `rpc_address` instead of `internal_address`. Thus the ConfigNode can't send heartbeat to DataNode correctly. By the way, I also make the shown `host` information of cluster DataNodes consistent both in `show cluster` and `show datanodes`.

Local test screenshot: 
image


wangchao316 (approved):
LGTM

Beyyes (approved):
LGTM

PR #7704, opened 2022-10-24
[IOTDB-4729] Fix column display name dismatched with SQL input

ericpai:
See JIRA: https://issues.apache.org/jira/browse/IOTDB-4729

After this fix applied, the output will be the same as the SQL input

image


ericpai:
@JackieTien97 @liuminghui233 @Alima777 PTAL

liuminghui233:
Fixed by #8108

PR #7713, opened 2022-10-25
[IOTDB-4732] Optimize client borrow logic in ClusterIT environment

CRZbulabula:
The ClusterIT environment is now full of client related error logs due to unsafe client borrow logic: 

image

In this PR I optimizedd the client borrow logic and add function annotation to remind the caller how to use this interface safely.


Beyyes (approved):
LGTM

PR #7717, opened 2022-10-25
[To rel/0.13] [IOTDB-3662] add audit log.

caozj1011:
Add audit log for insert/delete/query.

For DDL: default is info level.

For DML: insert is debug level, delete and query are info level.

level: IoTDB/LOGGER/NONE


server/src/main/java/org/apache/iotdb/db/utils/AuditLogUtils.java
```diff
@@ -0,0 +1,79 @@
...
+  public static final String TYPE_DELETE = "DELETE";
+  public static final String LOG_LEVEL_IOTDB = "IOTDB";
+  public static final String LOG_LEVEL_LOGGER = "LOGGER";
+  public static final String LOG_LEVEL_NONE = "NONE";
+
+  public static void writeAuditLog(String type, String log) {
+    IoTDBConfig config = IoTDBDescriptor.getInstance().getConfig();
+    String auditLogStorage = config.getAuditLogStorage();
+    SessionManager instance = SessionManager.getInstance();
+    String username = instance.getUsername(instance.getCurrSessionId());
```
HTHou, line 56:
```suggestion
    String username = instance.getCurrSession().getUsername();
```

cluster/src/main/java/org/apache/iotdb/cluster/server/basic/ClusterServiceProvider.java
```diff
@@ -89,6 +90,8 @@ public QueryContext genQueryContext(
 
   @Override
   public boolean executeNonQuery(PhysicalPlan plan) {
+    AuditLogUtils.writeAuditLog(
+        plan.getOperatorName(), String.format("measurements size:%s", plan.getPaths().size()));
```
HTHou, line 94:
```suggestion
    AuditLogUtils.writeAuditLog(
        plan.getOperatorName(),
        String.format(
            "measurements size:%s", plan.getPaths() == null ? 0 : plan.getPaths().size()));
```

HTHou (approved):

PR #7719, opened 2022-10-25
[IOTDB-4734] Modify create function grammar and process

lancelly:
Now, we need to specify the path of jar file when creating UDF. DBA need not to put jar into specified dir manually after this pr. We will transfer the jar to nodes in the cluster.


LeiRui:
no grammar readme or test example?

lancelly:
> no grammar readme or test example?

Hi, we need to refactor process of create function on both datanode and confignode. This pr covers the changes on datanode. We did not chekcout a new branch, so the refactor is in an intermediate state now. Test and user doc will be updated before this Saturday.

PR #7725, opened 2022-10-25
[IOTDB-4681] speed up mpp load

yschengzi:
problem:
load is slower than write in 3C3D env, should be faster in theory.

solution:
- [x] only read TsFile once(speed up 60.86%)
In the origin implement, parse TsFile into many chunks, and when dispatching chunk, it will read from disk again. In new implement, it will only read TsFile once, when there is too many chunks in memory, it will dispatch chunks immediately.
- [x] only read CompressionType and TSEncoding while auto create schema in AnalyzeVisitor (optimize 8.09%)
Deserialize ChunkHeader cost too much time when auto create schema while only want the CompressionType and TSEncoding of a timeseries in parsing TsFile, but deserialize the measurement Id of this chunk is wasted. A new method is developed to only  
 read CompressionType nad TSEncoding of a timeseries.
- [x]  reduce the information of ChunkData to spend less time on serialize (optimize 6.90%)
There is some information can be analyzed from chunkHeader.
- [x] create a new method to generate TsFileResource faster (optimize 5.65%)
Origin TsFileResource generating method is using TimeseriesMetadata, and this is too slow for update time index of TsFileResource.
- [x] remove parse path in TsFileIOWriter.endFile()(optimize 49.06%)
see pr #7826


server/src/main/java/org/apache/iotdb/db/mpp/plan/scheduler/load/LoadTsFileDispatcherImpl.java
```diff
@@ -82,26 +82,46 @@ public void setUuid(String uuid) {
...
+  //            } catch (FragmentInstanceDispatchException e) {
+  //              return new FragInstanceDispatchResult(e.getFailureStatus());
+  //            } catch (Throwable t) {
+  //              logger.error("cannot dispatch FI for load operation", t);
+  //              return new FragInstanceDispatchResult(
+  //                  RpcUtils.getStatus(
+  //                      TSStatusCode.INTERNAL_SERVER_ERROR, "Unexpected errors: " +
+  // t.getMessage()));
+  //            }
+  //          }
```
HTHou, line 103: Remove this?

yschengzi: done.

HTHou (approved):

PR #5976, opened 2022-05-20
[IOTDB-3216]Refactor Segments in SchemaFile to B+ Tree structure from list structure

bigreybear:
## Background
Children of the same node on MTree will be stored into one or several linked segments in SchemaFile when Main Memory wears out after #4732 merged . These segments are linked like a list, which turned out poor performance when number of children surged.
## What's Different
This PR reshape the structure between segments which share parent node on MTree to a B+ Tree (both with name and alias), which should improve the performace even with tremendous number of children on one node.

Besides, this PR puts foward a primary framework  named ```PageManager``` which decouples ```SchemaPage``` management from ```SchemaFile``` and makes it more reasonable to implement different structure between ```SchemaPage```s.

And the structure this PR implemented is a bit different from a classical B+ Tree in 2 ways: 

1. A node on this structure may split into 2 part with different number of entries, which is up to the trend of the insertion;
2. The split may take a bulk way if the key of the newly insert is the biggest or smallest of the original node, which may help to improve a little.

SchemaFileClassDiagram drawio (1)

A detailed specification for diagram above: https://apache-iotdb.feishu.cn/docs/doccnJHlbakm1n3GN6uLU6rz8Zm

PR #4732, opened 2022-01-08
[IOTDB-2559][IOTDB-1840][IOTDB-2215][IOTDB-2216] Persistent Schema Management

MarcosZyk:
### Docs
All related docs are in the shared folder linked by the following url.
https://y8dp9fjm8f.feishu.cn/drive/folder/fldcnoxcC1ZU9Ux43DCfImzXYqg?from=from_copylink

### Implement MTree Cache Framework
1. Rewrite attribute ```children``` management in MNode and implement MNodeContainer
2.  Separate service and store of MTree 
    1.  implement MemMTreeStore and CachedMTreeStore for memory mode and persistent mode respectively
    2. rewrite service logic in MTreeBelowSG and Traverser based on the persistent mode
3. Implement CacheManager for CachedMTreeStore
    1. implement cache framework and two cache strategy, plain and LRU
    2. implement Pin/Unpin to manage mnode cache and buffer

### Implement SchemaFile 
1. Implement mnode operations 
2. Implement page space management

Currently, we have not ensured the consistency and reliance of schemafile when encountering unexpected abort. It could be understand as we extend the workspace of metadata from memory to disk. Now we still recover SchemaRegion based on redoing mlog.

### New config parameters

1. enable_persistent_schema: whether to use persistent schema mode
2. cached_metadata_size_in_persistent_mode: the memory used for metadata cache when using persistent schema, currently this represents the num of cached mnode
3. max_schema_flush_thread: the max num of thread used for flushing metadata to schema file
4. minimum_schema_file_segment_in_bytes: the minimum size (in bytes) for segment inside a schema file page
5. page_cache_in_schema_file: the cache size for schema page in one schema file

### Future work
implement SchemaRegion recover based on schemafile

PR #7726, opened 2022-10-25
[IOTDB-2884]Redo Log for SchemaFile

bigreybear:
## Description
After https://github.com/apache/iotdb/pull/4732 merged, nodes of mtree could be swapped to disk when necessary in a format named SchemaFile which are organized as hierarchical B+Trees as [#5976](https://github.com/apache/iotdb/pull/5976) described. Since initiation of mtree still needs to scan all entries of mlog and apply it with corresponding actions, it may not be acceptable with massive mtree nodes. To enable the initiation from SchemaFile, there is an expected mechanism to keep SchemaFile integrity while flushing to disk, e.g., a redo log like module. And this pull request has provided such mechanism.

As the logging procedure may decay the performance, this pr also makes the segment size design more reasonable to be fit in one segmented page, and estimation of segment size more efficient if a bunch nodes are to be flushed as appended children. Last but not least, fixed size of SchemaFileLog (which acts like a redo log) is bound to around 256 MB(default) by restricting recorded pages, which is configurable in `iotdb-datanode.properties` identified as `schema_file_log_size`. Related formula is provided as well.

As long as this feature may be released and updated further, a version flag is added to the header of schema file.


MarcosZyk (approved):
LGTM. Thanks for your all contribution.

PR #7736, opened 2022-10-26
[IOTDB-4695] GROUP LEVEL query de-duplicates result columns unexpected

RYH61:
see : [https://issues.apache.org/jira/browse/IOTDB-4695](https://issues.apache.org/jira/browse/IOTDB-4695)

图片

### DIFF
--- a/integration-test/src/test/java/org/apache/iotdb/db/it/aggregation/IoTDBAggregationByLevelIT.java
+++ b/integration-test/src/test/java/org/apache/iotdb/db/it/aggregation/IoTDBAggregationByLevelIT.java
@@ -33,6 +33,7 @@ import org.junit.runner.RunWith;
 
 import java.sql.Connection;
 import java.sql.ResultSet;
+import java.sql.SQLException;
 import java.sql.Statement;
 
 import static org.apache.iotdb.itbase.constant.TestConstant.TIMESTAMP_STR;
@@ -671,6 +672,25 @@ public class IoTDBAggregationByLevelIT {
           e.getMessage()
               .contains(
                   "Common queries and aggregated queries are not allowed to appear at the same time"));
+    }
+  }
+
+  @Test
+  public void groupByLevelWithSameColumn() throws SQLException {
+    String[] retArray = new String[] {"8", "8"};
+    try (Connection connection = EnvFactory.getEnv().getConnection();
+        Statement statement = connection.createStatement()) {
+
+      int cnt = 0;
+      try (ResultSet resultSet =
+          statement.executeQuery(
+              "select count(status),count(status) from root.*.* GROUP BY level=0")) {
+        while (resultSet.next()) {
+          String ans = resultSet.getString(count("root.*.*.status"));
+          Assert.assertEquals(retArray[cnt], ans);
+          cnt++;
+        }
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


ericpai (commented):
Better to add an IT case

RYH61:
> Better to add an IT case

ok~

integration-test/src/test/java/org/apache/iotdb/db/it/aggregation/IoTDBAggregationByLevelIT.java
```diff
@@ -674,6 +675,25 @@ public void TestGroupByLevelWithoutAggregationFunc() {
...
+    String[] retArray = new String[] {"8", "8"};
+    try (Connection connection = EnvFactory.getEnv().getConnection();
+        Statement statement = connection.createStatement()) {
+
+      int cnt = 0;
+      try (ResultSet resultSet =
+          statement.executeQuery(
+              "select count(status),count(status) from root.*.* GROUP BY level=0")) {
+        while (resultSet.next()) {
+          String ans = resultSet.getString(count("root.*.*.status"));
```
ericpai, line 689: It seems that this case can't verify whether the output has 2 or 1 columns. You should use `resultSet.getMetadata()` and check
1. It contains 2 columns.
2. The 1st and 2nd column name are both `count(root.*.*.status)`.

ericpai (changes requested):

integration-test/src/test/java/org/apache/iotdb/db/it/aggregation/IoTDBAggregationByLevelIT.java
RYH61: `String[] retArray = new String[] {"8", "8"};`

The results that already contain two columns should be changed here

ericpai (approved):

integration-test/src/test/java/org/apache/iotdb/db/it/aggregation/IoTDBAggregationByLevelIT.java
ericpai: Thanks. Maybe the while loop can be removed, and add another assert `assertEquals(count("root.*.*.status"), metadata.getColumn(1)); `

RYH61: ok~~

wangchao316 (approved):
LGTM

ericpai:
All ClusterIT tasks are failed. Maybe there're some potential bugs.

ericpai (changes requested):
All Cluster IT tasks are failed. Maybe there are some potential bugs.

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/AnalyzeVisitor.java
```diff
@@ -417,8 +424,9 @@ private List<Pair<Expression, String>> analyzeSelect(
           break;
         }
       }
+      outputExpressionMap.put(columnIndex, outputExpressions);
```
ericpai, line 427: I think the columnIndex is very useful in later use, so it's better to set from 0, not 1. If it begins at 0, we can easily get the corresponding ResultColumn by `queryStatement.getSelectComponent().getResultColumns().get(columnIndex)`

ericpai (approved):

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/AnalyzeVisitor.java
RYH61: ok~~

PR #7739, opened 2022-10-26
[IOTDB-4758]Delete snapshot after region migration

suchenglong:
### Content1 ...
Delete remote snapshot after region migration
### test
Test plan:
1. Enable the cluster 1C3D and two replicas
2. kiil drops the process where DataNode2 resides
3. Run the show regions command to view the replication information
4. Use the remove_node.sh script to remove DataNode2
5. Check whether snapshots are generated for the snapshot directories in DataNode1 and DataNode3
6. Wait for a period of time, and check whether the temporary file that completes the replica migration is deleted in DataNode1 and DataNode3. Run the show regions command to check whether the replica migration is correct
7. Connect to the datanode where the migrated replica resides to query data and verify that data is normal

Result:
image
Before volume reduction:
image
Stop node 2:
image
Removing node 2:
image
image
The data migration is complete and no temporary snapshot file is available：
image
Query migrated replica data on datanoe1：
image

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [x] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


consensus/src/main/java/org/apache/iotdb/consensus/multileader/MultiLeaderConsensus.java
```diff
@@ -278,6 +278,10 @@ public ConsensusGenericResponse addPeer(ConsensusGroupId groupId, Peer peer) {
       logger.info("[MultiLeaderConsensus] activate new peer...");
       impl.activePeer(peer);
 
+      // step 7: spot clean
+      logger.info("[MultiLeaderConsensus] do spot clean...");
```
wangchao316, line 282: why clean in addPeer()?

consensus/src/main/java/org/apache/iotdb/consensus/multileader/service/MultiLeaderRPCServiceProcessor.java
```diff
@@ -277,5 +279,33 @@ public void triggerSnapshotLoad(
...
+      status.setMessage(message);
+      resultHandler.onComplete(new TCleanupTransferredSnapshotRes(status));
+      return;
+    }
+    TSStatus responseStatus = null;
+    try {
+      impl.cleanupTransferredSnapshot(req.snapshotId);
+      responseStatus = new TSStatus(TSStatusCode.SUCCESS_STATUS.getStatusCode());
+    } catch (ConsensusGroupAddPeerException e) {
+      responseStatus = new TSStatus(TSStatusCode.INTERNAL_SERVER_ERROR.getStatusCode());
```
wangchao316, line 304: print  log exception.

consensus/src/main/java/org/apache/iotdb/consensus/multileader/MultiLeaderConsensus.java
suchenglong: After a peer is added, the snapshot replicas need to be transferred. After the snapshot replicas are transferred and loaded, you need to clear the temporary replica files on the remote server to prevent data space waste.

consensus/src/main/java/org/apache/iotdb/consensus/multileader/service/MultiLeaderRPCServiceProcessor.java
suchenglong: add error log

xingtanzjr (approved):

PR #7742, opened 2022-10-26
[To rel/0.13] [IOTDB-4767] Support fetching all connection info in Session & SessionPool

JackieTien97:
In this pr, I add a new interface to support fetching information of all the connections.
Each connection info will be like
```
struct TSConnectionInfo {
  1: required string userName
  2: required i64 logInTime
  3: required string connectionId // ip:port for thrift-based service and clientId for mqtt-based service
  4: required TSConnectionType type
}
```
If type is `THRIFT_BASED`, connectionId will be like `ip:port` style, and if type is `MQTT_BASED`, connectionId will just be `clientId` for MQTTClient.


thrift/src/main/thrift/rpc.thrift
```diff
@@ -439,6 +439,22 @@ struct TSBackupConfigurationResp {
...
 }
 
+enum TSConnectionType {
+  THRIFT_BASED
+  MQTT_BASED
+}
+
+struct TSConnectionInfo {
+  1: required string userName
+  2: required i64 logInTime
```
qiaojialin, line 449:
```suggestion
  2: required i64 loginTime
```

qiaojialin (approved):

PR #7712, opened 2022-10-24
[IOTDB-4742] [Consensus] add new api: addNewNodeToExistedGroup

SzyWilliam:
It is for RatisConsensus.

PR #7743, opened 2022-10-26
[IOTDB-4745] Add ADD_NEW_NODE state for ConfigNodeProcedure

Beyyes:
## Description

To fix the problem `Execute start-confignode.sh third times simultaneously meets error that one confignode cannot startup successflly`, using `addNewNodeToExistedGroup` to replace with `createPeer` and `addPeer`.

See the `addNewNodeToExistedGroup` impl: https://github.com/apache/iotdb/pull/7712


SzyWilliam (approved):
LGTM, Thanks for working on this!

PR #7745, opened 2022-10-26
[IOTDB-4769] When modifying an alias, you need to add back quotes when the alias is a special character.

23931017wu:
###1. When modifying a time series alias, if the alias is a real number, you need to add a back quote, otherwise an error should be reported. Now you can successfully modify the alias without adding a back quote, causing the alias to not be used normally.

#### Before
图片

When an alias is used and a piece of data is written, a new time series will be created incorrectly.

图片

#### After

Create Time Series and modify alias without back quotes.
图片

Change alias with back quotes, write a piece of data
图片

### 2. Fix that two back quotes in the node path do not resolve to a back quote
Before: 
图片
After:
图片

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/mpp/plan/parser/ASTVisitor.java
+++ b/server/src/main/java/org/apache/iotdb/db/mpp/plan/parser/ASTVisitor.java
@@ -1818,7 +1818,7 @@ public class ASTVisitor extends IoTDBSqlParserBaseVisitor<Statement> {
     if (ctx.constant() != null) {
       alias = parseStringLiteral(ctx.constant().getText());
       if (PathUtils.isRealNumber(alias)
-              || !TsFileConstant.IDENTIFIER_PATTERN.matcher(alias).matches()) {
+          || !TsFileConstant.IDENTIFIER_PATTERN.matcher(alias).matches()) {
         throw new SQLParserException("Not support for this alias, Please enclose in back quotes.");
       }
     } else {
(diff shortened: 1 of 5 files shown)


23931017wu:
@lancelly Hi, Please review!

server/src/main/java/org/apache/iotdb/db/mpp/plan/parser/ASTVisitor.java
```diff
@@ -1664,10 +1664,14 @@ private String parseNodeString(String nodeName) {
     }
     if (nodeName.startsWith(TsFileConstant.BACK_QUOTE_STRING)
         && nodeName.endsWith(TsFileConstant.BACK_QUOTE_STRING)) {
-      String unWrapped = nodeName.substring(1, nodeName.length() - 1);
+      String unWrapped =
+          nodeName
+              .substring(1, nodeName.length() - 1)
+              .replace(TsFileConstant.DOUBLE_BACK_QUOTE_STRING, TsFileConstant.BACK_QUOTE_STRING);
+      ;
```
lancelly, line 1671: This method should stay the same~

23931017wu: This is to solve the problem that the double back quotes in the node path do not resolve to single quotes.

图片

23931017wu: I think the document also needs to be modified. Let me restore it here first, and then mention a new pr~

lancelly: You are right. LGTM~

neuyilan (approved):

PR #7759, opened 2022-10-27
Add wait logic  to ensure no data lost when remove a Peer from MultiLeader consensus group

xingtanzjr:
## Description
Before this change, if the removed peer is Leader in the consensus group, the data may lost if the peer's log sync is not completed. 
In this PR, we add the logic to wait the SyncLog to be completed before removing it.


wangchao316 (approved):
LGTM

consensus/src/main/java/org/apache/iotdb/consensus/multileader/MultiLeaderServerImpl.java
```diff
@@ -479,6 +481,45 @@ public void notifyPeersToRemoveSyncLogChannel(Peer targetPeer)
     }
   }
 
+  public void waitTargetPeerUntilSyncLogCompleted(Peer targetPeer)
+      throws ConsensusGroupAddPeerException {
+    long checkIntervalInMs = 10_000L;
```
SpriCoder, line 486: Maybe this value can be a property of MultiLeaderConfig

consensus/src/main/java/org/apache/iotdb/consensus/multileader/MultiLeaderConsensus.java
```diff
@@ -310,13 +310,25 @@ public ConsensusGenericResponse removePeer(ConsensusGroupId groupId, Peer peer)
...
+
+    try {
+      // let target peer reject new write
+      impl.inactivePeer(peer);
+      // wait its SyncLog to complete
+      impl.waitTargetPeerUntilSyncLogCompleted(peer);
+    } catch (ConsensusGroupAddPeerException e) {
+      // we only log warning here because sometimes the target peer may already be down
+      logger.warn("cannot wait {} to complete SyncLog. error message: {}", peer, e.getMessage());
+    }
```
SpriCoder, line 330: Why we catch `ConsensusGroupAddPeerException` in removePeer method? I think we need to specific the exception

consensus/src/main/java/org/apache/iotdb/consensus/multileader/MultiLeaderServerImpl.java
```diff
@@ -479,6 +481,45 @@ public void notifyPeersToRemoveSyncLogChannel(Peer targetPeer)
...
+            targetPeer,
+            res.searchIndex,
+            res.safeIndex);
+        Thread.sleep(checkIntervalInMs);
+      }
+    } catch (IOException | TException e) {
+      throw new ConsensusGroupAddPeerException(
+          String.format(
+              "error when waiting %s to complete SyncLog. %s", targetPeer, e.getMessage()),
+          e);
```
SpriCoder, line 512: I think we shouldn't throw `ConsensusGroupAddPeerException` here which may cause confuse.

consensus/src/main/java/org/apache/iotdb/consensus/multileader/MultiLeaderServerImpl.java
```diff
@@ -479,6 +481,45 @@ public void notifyPeersToRemoveSyncLogChannel(Peer targetPeer)
...
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw new ConsensusGroupAddPeerException(
+          String.format(
+              "thread interrupted when waiting %s to complete SyncLog. %s",
+              targetPeer, e.getMessage()),
+          e);
```
SpriCoder, line 519: I think we shouldn't throw `ConsensusGroupAddPeerException` here which may cause confuse.

SpriCoder (changes requested):
PTAL~

consensus/src/main/java/org/apache/iotdb/consensus/multileader/MultiLeaderConsensus.java
xingtanzjr: Sure, I will define another similar exception

consensus/src/main/java/org/apache/iotdb/consensus/multileader/MultiLeaderServerImpl.java
xingtanzjr: Both is fine. This parameter is used to avoid frequent useless queries and a fixed suitable interval is enough. For example, `10s` and `20s` are both ok and it is unnecessary to change it. So I think we can use a fixed value here

consensus/src/main/java/org/apache/iotdb/consensus/multileader/MultiLeaderServerImpl.java
xingtanzjr: Sure, will define and use another exception

consensus/src/main/java/org/apache/iotdb/consensus/multileader/MultiLeaderServerImpl.java
xingtanzjr: Sure, will define and use another exception

JackieTien97 (approved):

PR #7768, opened 2022-10-28
[IOTDB-4782] Fix systemMetrics in confignode

SpriCoder:
See: https://issues.apache.org/jira/browse/IOTDB-4782

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/service/metrics/SystemMetrics.java
+++ b/server/src/main/java/org/apache/iotdb/db/service/metrics/SystemMetrics.java
@@ -63,17 +63,15 @@ public class SystemMetrics implements IMetricSet {
   @Override
   public void bindTo(AbstractMetricService metricService) {
     collectSystemCpuInfo(metricService);
-    if (isDataNode) {
-      collectSystemDiskInfo(metricService);
-    }
     collectSystemMemInfo(metricService);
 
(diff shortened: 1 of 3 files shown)


CRZbulabula (approved):
LGTM～

server/src/main/java/org/apache/iotdb/db/service/metrics/SystemMetrics.java
```diff
@@ -51,21 +51,25 @@ public class SystemMetrics implements IMetricSet {
...
     collectSystemCpuInfo(metricService);
-    collectSystemDiskInfo(metricService);
+    if (isDataNode) {
+      collectSystemDiskInfo(metricService);
+    }
     collectSystemMemInfo(metricService);
 
     // finally start to update the value of some metrics in async way
-    if (metricService.isEnable() && null == currentServiceFuture) {
+    if (metricService.isEnable() && null == currentServiceFuture && isDataNode) {
```
xingtanzjr, line 72: Let's combine the logic into collectSystemDiskInfo because it seems that this scheduled task is only serviced for `collectSystemDiskInfo `

SpriCoder: Already Fixed~

xingtanzjr (approved):

PR #7774, opened 2022-10-28
[IOTDB-4768] Balancing cluster RegionGroup leader distribution by MinimumCostFlow algorithm

CRZbulabula:
Balancing the cluster Region-leader distributiong by MinimumCostFlow algorithm.

The `RouteBalancer` starts a `Cluster-LeaderBalancing-Service`, and which will balance the leader distribution in every 5 loops of heartbeat service(5s per loop in default configuration). The leader distribution of SchemaRegion and DataRegion are calculated separately.

I've proved the optimality of MinimumCostFlow algorithm and the deficiency of greedy algorithm for the leader distribution problem at the [RegionGroup-leader balancing algorithm doc](https://apache-iotdb.feishu.cn/docx/doxcn6xa2p3bvqyF0cwBcit5fkh).


confignode/src/main/java/org/apache/iotdb/confignode/manager/load/balancer/RouteBalancer.java
```diff
@@ -241,7 +256,81 @@ public void stopRouteBalancingService() {
...
+              case DataRegion:
+              default:
+                consensusProtocolClass = DATA_REGION_CONSENSUS_PROTOCOL_CLASS;
+                break;
+            }
+            LOGGER.info(
+                "[LeaderBalancer] Try to change the leader of Region: {} to DataNode: {} ",
+                regionGroupId,
+                newLeaderId);
+            changeRegionLeader(
```
wangchao316, line 302: I have a question .   change region leader during removing datanode or adding datanode, Whether Scaling Is Affected ？

confignode/src/main/java/org/apache/iotdb/confignode/manager/load/balancer/RouteBalancer.java
```diff
@@ -219,8 +234,8 @@ public void startRouteBalancingService() {
                 leaderBalancingExecutor,
                 this::balancingRegionLeader,
```
wangchao316, line 235: Is there adding a config, can close balance?

CRZbulabula: Yes, I've added

confignode/src/main/java/org/apache/iotdb/confignode/manager/load/balancer/RouteBalancer.java
CRZbulabula: No, the ConfigNode-leader just try to optimize the RegionGroups' leader distribution, and this process doesn't effect the validity of cluster extension.

Beyyes (approved):

PR #7776, opened 2022-10-28
[IOTDB-4676][IOTDB-4677] Support sync DeleteTimeseries and sync Delete Data

yschengzi:
- Support collect DeleteTimeseries in sender
- Support load DeleteTimeseries in receiver


server/src/main/java/org/apache/iotdb/db/sync/sender/manager/LocalSyncManager.java
```diff
@@ -46,8 +53,30 @@ public LocalSyncManager(DataRegion dataRegion, Pipe pipe) {
...
+    Set<String> devices = new HashSet<>();
+    unsealedResources.forEach(o -> devices.addAll(o.getDevices()));
+    sealedResources.forEach(o -> devices.addAll(o.getDevices()));
+
+    for (String device : devices) {
+      try {
+        PartialPath path = new PartialPath(device);
+        if (deletion.getPath().overlapWith(path)) {
+          Deletion splitDeletion =
+              new Deletion(path, 0, deletion.getStartTime(), deletion.getEndTime());
```
Cpaulyz, line 69: When deletion.getPath() is a full path without wildcards, the splitting logic may be error.

e.g. "root.sg.d1.s1".overlapWith("root.sg.d1") is false.

Cpaulyz (changes requested):

server/src/main/java/org/apache/iotdb/db/sync/sender/manager/LocalSyncManager.java
yschengzi: done.

server/src/main/java/org/apache/iotdb/db/sync/sender/manager/LocalSyncManager.java
```diff
@@ -46,8 +54,31 @@ public LocalSyncManager(DataRegion dataRegion, Pipe pipe) {
...
+          Deletion splitDeletion =
+              new Deletion(deletion.getPath(), 0, deletion.getStartTime(), deletion.getEndTime());
+          syncPipe.collectRealTimeDeletion(
+              splitDeletion, dataRegion.getStorageGroupName(), dataRegionId);
+        }
+      } catch (IllegalPathException e) {
+        logger.error(
+            String.format(
+                "Parse device path %s to partial path error when splitting deletion.", device),
+            e);
```
neuyilan, line 79: These codes may not  really be understandable, 

If the codes in 69 is true for `device1`, then the deletion will be added to the sync pipe.

In the next loop, for `device2`, if the codes in 69 is still `TRUE`, then the deletion will be added to the sync pipe **again**. is this expected?

neuyilan (changes requested):

server/src/main/java/org/apache/iotdb/db/sync/sender/manager/LocalSyncManager.java
Cpaulyz: > These codes may not really be understandable,
> 
> If the codes in 69 is true for `device1`, then the deletion will be added to the sync pipe.
> 
> In the next loop, for `device2`, if the codes in 69 is still `TRUE`, then the deletion will be added to the sync pipe **again**. is this expected?

Yes.. The implementation is incorrect and still WIP. We should put the intersection of `device+".*"` and `deletion.getPath()` to splitDeletion.（#7823）

Cpaulyz (approved):
LGTM

server/src/main/java/org/apache/iotdb/db/sync/sender/manager/LocalSyncManager.java
```diff
@@ -46,8 +54,36 @@ public LocalSyncManager(DataRegion dataRegion, Pipe pipe) {
 
   /** tsfile */
   @Override
-  public void syncRealTimeDeletion(Deletion deletion) {
-    syncPipe.collectRealTimeDeletion(deletion, dataRegion.getStorageGroupName(), dataRegionId);
+  public void syncRealTimeDeletion(
+      Deletion deletion,
+      List<TsFileResource> unsealedResources,
+      List<TsFileResource> sealedResources) {
```
Cpaulyz, line 60: To avoid affecting the deletion performance, we should return directly if syncDelOp=false.

Maybe add a getter in TsFilePipe to get syncDelOp.

Cpaulyz (changes requested):

server/src/main/java/org/apache/iotdb/db/sync/sender/manager/LocalSyncManager.java
yschengzi: done.

neuyilan (approved):
LGTM

PR #7784, opened 2022-10-28
[IOTDB-4793] Prevent compatibility problems in ConfigNode

CRZbulabula:
Changes:

1. Use static field `planType` instead of .ordinal() in ConfigPhysicalPlanType
2. Use static field `typeCode` instead of .ordinal() in ProcedureType
3. Complement some Procedure serialize and deserialize UT
4. Register all snapshotProcessor in ConfigExecutor
5. The `testRatisSnapshotTriggerThreshold` parameter in IoTDBConfigNodeSnapshotIT is increased to 100 to avoid preparing test environment for too long.


Beyyes (approved):
LGTM

PR #7789, opened 2022-10-29
[IOTDB-3397] Support Template Session and Add Session IT

MarcosZyk:
### Modification

1. Support unset template and drop template in session
2. Add template session IT
3. Add template py session IT


HTHou (approved):
LGTM

PR #7793, opened 2022-10-29
Add UT for CreateCQProcedure

JackieTien97:
Add UT for CreateCQProcedure and fix a serde bug for procedure that missing a break in switch-case.


CRZbulabula (approved):
LGTM

PR #7808, opened 2022-10-31
[IOTDB-3149] There is a difference between adding ** path and without a path to list user privileges

RYH61:
This problem has been solved, but new problems have been found.

The user checks the permissions on the root. sg. ** path, and the root. ** is also matched

 
图片

after:

图片


SpriCoder (approved):
LGTM~

PR #7826, opened 2022-10-31
[IOTDB-4791] Optimize the endFile in `TsFileIOWriter`

THUMarkLau:
See [IOTDB-4791](https://issues.apache.org/jira/browse/IOTDB-4791).

The may reason for the performance bottleneck of `TsFileIOWriter::endFile` is the path validation for in the constructor function of Path. This pr optimizes it by skipping the validation of path when flushing metadata because the path is validated when data is inserting.

### DIFF
--- a/tsfile/src/main/java/org/apache/iotdb/tsfile/read/common/Path.java
+++ b/tsfile/src/main/java/org/apache/iotdb/tsfile/read/common/Path.java
@@ -93,10 +93,17 @@ public class Path implements Serializable, Comparable<Path> {
    *
    * @param device root.deviceType.d1
    * @param measurement s1 , does not contain TsFileConstant.PATH_SEPARATOR
+   * @param needCheck need to validate the correctness of the path
    */
-  public Path(String device, String measurement) {
+  public Path(String device, String measurement, boolean needCheck) {
     if (device == null || measurement == null) {
       throw new PathParseException(ILLEGAL_PATH_ARGUMENT);
+    }
+    if (!needCheck) {
+      this.measurement = measurement;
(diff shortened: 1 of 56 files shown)


tsfile/src/main/java/org/apache/iotdb/tsfile/read/common/Path.java
```diff
@@ -93,11 +93,18 @@ public Path(String pathSc, boolean needSplit) {
...
    */
-  public Path(String device, String measurement) {
+  public Path(String device, String measurement, boolean needCheck) {
     if (device == null || measurement == null) {
       throw new PathParseException(ILLEGAL_PATH_ARGUMENT);
     }
+    if (!needCheck) {
+      this.measurement = measurement;
+      this.device = device;
+      this.fullPath = device + "." + measurement;
```
xingtanzjr, line 105: Using `StringBuilder` or some other ways to avoid `+`

xingtanzjr:
> See [IOTDB-4791](https://issues.apache.org/jira/browse/IOTDB-4791).
> 
> The **may** reason for the performance bottleneck of `TsFileIOWriter::endFile` is the path validation for in the constructor function of Path. This pr optimizes it by skipping the validation of path when flushing metadata because the path is already validated when data is inserting.

possible reason....

xingtanzjr:
Can we keep the original constructor because there are two many places uses it. After this change, these invokers should always pass a parameter `true`. 

We can add a new constructor or a new static method to construct the `Path` without validation, which is more easily to be understood

xingtanzjr (approved):

PR #7859, opened 2022-11-01
Fix conflicts in path constructor

choubenson:
Run `TsFileWriteApiTest` and execute "mvn clean package -DskipTests -am -pl server" successfully.
image
image


xingtanzjr (approved):

xingtanzjr (approved):

xingtanzjr:
As this PR is used to fix the compile issue of current master branch, it should be merged as soon as possible. According to the change, @choubenson ran the UT locally and `mvn clean package` locally, which were both passed. Thus, I will merge this PR without waiting CI

PR #7862, opened 2022-11-01
Fix isBlocked bug of schema query

MarcosZyk:
## Description

The isBlocked ListenableFuture of child will change status after returned from current merge operator to query driver, which results in the driver may directly start invoking hasNext and next method of current merge operator before all data prepared well.


Cpaulyz (approved):
LGTM

JackieTien97 (approved):

PR #7865, opened 2022-11-02
[IOTDB-4230]fix start-cli.sh does not support -e

jixuan1989:
this is a bug introduced by jira iotdb-4230 and PR #7802


jixuan1989:
this PR has something error...

PR #7866, opened 2022-11-02
[IOTDB-4230]fix start-cli.sh does not support -e

jixuan1989:
this pr fix "-e can not be used in start-cli.sh" bug introduced in https://github.com/apache/iotdb/pull/7802


HTHou (approved):
LGTM
Tested on my local machine.
13301667352528_ pic

PR #7877, opened 2022-11-02
[IOTDB-4828] Implement TestOnly RPC interface setDataNodeStatus

CRZbulabula:
Implement a TestOnly RPC interface setDataNodeStatus as following:

```
/** Set the target DataNode to the specified status */
common.TSStatus setSystemStatus(TSetDataNodeStatusReq req)
```

Will be helpful for ClusterIT. i.e. we can change some DataNodes' status more easily so that the ClusterIT can cover more test cases.


Beyyes (approved):
LGTM

PR #7878, opened 2022-11-02
[IOTDB-4800] [Rewrite UT] Rewrite some UT in compaction module

THUMarkLau:
See [IOTDB-4800](https://issues.apache.org/jira/browse/IOTDB-4800).

Rewrite some UT to remove the use of old stand-alone.


JackieTien97 (changes requested):

server/src/test/java/org/apache/iotdb/db/engine/compaction/CompactionSchedulerTest.java
```diff
@@ -59,7 +52,6 @@
 import static org.junit.Assert.fail;
 
 // TODO don't use old standalone style IT
```
JackieTien97, line 54:
```suggestion
```

JackieTien97 (changes requested):

JackieTien97 (approved):

PR #7904, opened 2022-11-04
[IOTDB-4850] Disable first election

SzyWilliam:
First election aims to reduce the unavailable time during consensus group starts-up. However, in current implementation, it may cause new follower receive SHUTDOWN errors. 

I'll disable it for now.

### DIFF
--- a/confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeConfig.java
+++ b/confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeConfig.java
@@ -220,8 +220,8 @@ public class ConfigNodeConfig {
   private long dataRegionRatisPreserveLogsWhenPurge = 1000;
 
   /* first election timeout shares between 3 regions */
-  private long ratisFirstElectionTimeoutMinMs = 50;
-  private long ratisFirstElectionTimeoutMaxMs = 150;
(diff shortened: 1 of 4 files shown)


confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeConfig.java
```diff
@@ -220,8 +220,8 @@ public class ConfigNodeConfig {
   private long dataRegionRatisPreserveLogsWhenPurge = 1000;
 
   /* first election timeout shares between 3 regions */
-  private long ratisFirstElectionTimeoutMinMs = 50;
-  private long ratisFirstElectionTimeoutMaxMs = 150;
+  private long ratisFirstElectionTimeoutMinMs = 2000;
```
wangchao316, line 223: whether is  this time a better,  can our election-time greater than this time?

SzyWilliam: Yes. First election time is only a special case for election time. Currently there are bugs triggered by first election, so it's better to increase first election time to a matter of seconds.

SzyWilliam:
This PR is superseded by https://github.com/apache/iotdb/pull/8016. I'm closing this for now.

PR #7923, opened 2022-11-07
[IOTDB-4862] optimize transfer leader implementation

SzyWilliam:
Ratis supports to transfer leader quietly.
That is, the new leader can start the election immediately without waiting for an election timeout.

Add new unit tests to verify.


OneSizeFitsQuorum (approved):
LGTM

PR #7936, opened 2022-11-08
[IOTDB-4838] Adapt SchemaRegionSchemaFileImpl's recovery to Ratis.

Cpaulyz:
## Description

Currently, the SchemaRegionSchemaFileImpl relies on mlog and schema file to recover.

In cluster mode, schemaRegion relies on Ratis to guarantee consistency, while raftlog takes repeating info of mlog.

We want to use raftlog to recover info not in snapshot and use schema file to implement snapshot and online business.

This PR includes the implmentation of these two interface in `SchemaRegionSchemaFileImpl`:
1. createSnapshot
2. loadSnapshot


server/src/main/java/org/apache/iotdb/db/metadata/mtree/store/disk/schemafile/SchemaFile.java
```diff
@@ -405,4 +412,48 @@ public long getTargetSegmentOnTest(long srcSegAddr, String key)
...
+    File schemaFile =
+        SystemFileFactory.INSTANCE.getFile(
+            getDirPath(sgName, schemaRegionId), MetadataConstant.SCHEMA_FILE_NAME);
+    File schemaLogFile =
+        SystemFileFactory.INSTANCE.getFile(
+            getDirPath(sgName, schemaRegionId), MetadataConstant.SCHEMA_LOG_FILE_NAME);
+    Files.deleteIfExists(schemaFile.toPath());
+    Files.deleteIfExists(schemaLogFile.toPath());
+    Files.createLink(schemaFile.toPath(), snapshot.toPath());
+    return new SchemaFile(sgName, schemaRegionId, false, -1L, false);
```
bigreybear, line 454: I see a `-1L` for parameter of TTL, I'm not sure if it is ok for snapshot if the original SchemaFile had been set a ttl not equals -1L ?

server/src/test/java/org/apache/iotdb/db/metadata/schemaRegion/SchemaRegionBasicTest.java
```diff
@@ -86,6 +95,103 @@ public void tearDown() throws Exception {
...
+              tags,
+              null,
+              null),
+          -1);
+
+      File snapshotDir = new File(config.getSchemaDir() + File.separator + "snapshot");
+      snapshotDir.mkdir();
+      schemaRegion.createSnapshot(snapshotDir);
+
+      schemaRegion.loadSnapshot(snapshotDir);
```
bigreybear, line 134: I am not sure if it is better with another object `SchemaRegion` to load the snapshot and verify the test?

bigreybear (commented):
Greate job for me ! Only few comments as mentioned

bigreybear (approved):
LGTM~

server/src/main/java/org/apache/iotdb/db/metadata/mtree/store/disk/schemafile/SchemaFile.java
Cpaulyz: > I see a `-1L` for parameter of TTL, I'm not sure if it is ok for snapshot if the original SchemaFile had been set a ttl not equals -1L ?

Change to `CommonDescriptor.getInstance().getConfig().getDefaultTTLInMs()`. But it seems that whatever this field is, it will be overwritten in the `initFileHeader()`.

server/src/main/java/org/apache/iotdb/db/metadata/schemaregion/SchemaRegionSchemaFileImpl.java
```diff
@@ -164,13 +165,16 @@ public class SchemaRegionSchemaFileImpl implements ISchemaRegion {
...
+  private final String storageGroupFullPath;
+  private final SchemaRegionId schemaRegionId;
 
+  // the log file writer
+  private boolean usingMLog = true;
   // the log file seriesPath
-  private String logFilePath;
-  private File logFile;
+  //  private String logFilePath;
+  //  private File logFile;
```
MarcosZyk, line 177: remove this

MarcosZyk (approved):

PR #7939, opened 2022-11-08
[IOTDB-4879] Use CAS in reserve method of MultiLeaderMemoryManager

SpriCoder:
Try to use CAS in reserve method of MultiLeaderMemoryManager to avoid `synchronized` way.


OneSizeFitsQuorum (approved):
LGTM

PR #7942, opened 2022-11-09
cleanup spotless workaround code and doc

tisonkun:
This is a follow-up of #6441.


tisonkun:
cc @HTHou

HTHou:
So, we can run `mvn spotless:apply` with jdk 17 now?

tisonkun:
@HTHou I think with #6441 the spotless version already supports doing this. This PR is to clean up workaround settings and incorrect docs.

I ran `mvn spotless:apply` locally with JDK17 and it doesn't complain.

HTHou (approved):

PR #7943, opened 2022-11-09
[IOTDB-4876] Fix PathPatternTree missing information on subpaths

Cpaulyz:
### BUG description
 After appending root.sg.d and root.sg.d.** to an empty pathPatternTree and invoking constructTree(), only root.sg.d.** can be taken by invoking getAllPathPatterns() and root.sg.d is missing.

### Solution
In the original design, only the leaf nodes on the PathPatternTree are considered as measurement. However, the information is lost when there is a subpath for measurement. e.g. root->sg->d(measurement)->**(measurement).

So we use an additional roleType field to identify whether the node in the tree is internal or measurement.

To avoid compatibility problema, new semantics are given to the original valueSet field in the serialization method.


MarcosZyk (approved):

liuminghui233 (approved):
+1

PR #7949, opened 2022-11-09
[IOTDB-4882] NPE appears when using uppercase functionName

RYH61:
see : [https://issues.apache.org/jira/browse/IOTDB-4882](https://issues.apache.org/jira/browse/IOTDB-4882)


liuminghui233:
Fixed by #8108

PR #7954, opened 2022-11-09
[IOTDB-4771]BufferedPipeDataQueueTest.testOfferWhileTakingWithDiscontinuousSerialNumber fail

yschengzi:
problem:
BufferedPipeDataQueueTest.testOfferWhileTakingWithDiscontinuousSerialNumber fail.


server/src/main/java/org/apache/iotdb/db/sync/pipedata/queue/BufferedPipeDataQueue.java
```diff
@@ -194,10 +194,10 @@ public boolean offer(PipeData pipeData) {
...
       }
     }
-    if (!inputDeque.offer(pipeData)) {
-      return false;
-    }
     synchronized (waitLock) {
+      if (!inputDeque.offer(pipeData)) {
+        return false;
+      }
       waitLock.notifyAll();
```
HTHou, line 201: If return false here, we don't need to call waitLock.notifyAll()?

PR #7957, opened 2022-11-10
[IOTDB-4574] Implement Mpp object query and Refine ClusterSchemaFetcher

MarcosZyk:
### Motivation

In current implementation, the schemaTree taken from schema fetch always experiences serialization and deserialization no matter whether the query is only executed locally. This severely affects the query performance for 1C1D.

### Solution

In fact, this problem is a sub problem of mpp based object query and the tsblock doesn't support transport object. We implement the mpp object query framework in this pr and apply it on schema fetch.

Here is the design doc https://apache-iotdb.feishu.cn/docx/ODd3d1s05o52KJxWkIwc1Ue2nSf

Besides, the path pattern provided by data query without any wildcard now can be used to taken schema from schema cache directly without fetching from remote DataNode.


server/src/main/java/org/apache/iotdb/db/mpp/execution/object/MPPObjectPool.java
```diff
@@ -0,0 +1,69 @@
...
+    private MPPObjectPoolHolder() {}
+  }
+
+  public static MPPObjectPool getInstance() {
+    return MPPObjectPoolHolder.INSTANCE;
+  }
+
+  public synchronized <T extends ObjectEntry> T put(String queryId, T objectEntry) {
+    List<ObjectEntry> queryObjectList =
+        objectPool.computeIfAbsent(queryId, k -> Collections.synchronizedList(new ArrayList<>()));
```
xingtanzjr, line 47: Some behaviors of `synchronizedList` may be different from our expectation.

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/object/ObjectDeserializeOperator.java
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    TsBlock tsBlock = child.next();
+    if (tsBlock == null || tsBlock.isEmpty()) {
+      return null;
+    }
+    ByteBuffer buffer;
+    for (int i = 0; i < tsBlock.getPositionCount() - 1; i++) {
+      buffer = ByteBuffer.wrap(tsBlock.getColumn(0).getBinary(i).getValues());
+      bufferList.add(buffer);
```
xingtanzjr, line 94: Maybe it is not a safe way that sharing the intevmidie value by class level variable.

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/object/ObjectDeserializeOperator.java
```diff
@@ -0,0 +1,215 @@
...
+    DataInputStream dataInputStream = new DataInputStream(segmentedByteInputStream);
+    TsBlockBuilder builder = new TsBlockBuilder(outputDataTypes);
+    try {
+      byte objectRecordSymbol = dataInputStream.readByte();
+      while (objectRecordSymbol == OBJECT_START_SYMBOL) {
+        ObjectEntry objectEntry =
+            ObjectEntryFactory.getObjectEntry(ObjectType.deserialize(segmentedByteInputStream));
+        objectEntry.deserializeObject(dataInputStream);
+        builder.getTimeColumnBuilder().writeLong(0L);
+        builder.getColumnBuilder(0).writeInt(objectPool.put(queryId, objectEntry).getId());
```
xingtanzjr, line 116: When do we remove the object from the pool to release the memory ?

server/src/main/java/org/apache/iotdb/db/mpp/execution/object/MPPObjectPool.java
```diff
@@ -0,0 +1,69 @@
...
+  @SuppressWarnings("unchecked")
+  public <T extends ObjectEntry> T get(String queryId, int objectId) {
+    List<ObjectEntry> queryObjectList = objectPool.get(queryId);
+    if (queryObjectList == null) {
+      return null;
+    }
+    return (T) queryObjectList.get(objectId);
+  }
+
+  public void clear(String queryId) {
```
xingtanzjr, line 62: If the object is cleared only when the query ends, it will use lots of memory

PR #7964, opened 2022-11-10
Fix problem in greedySelectLeader

Beyyes:
fix problem in greedySelectLeader, only use DataRegion but not DataRegion + SchemaRegion

Before fix:
image

After fix:
image


CRZbulabula (approved):
LGTM～

PR #7971, opened 2022-11-13
[IOTDB-4896] Fix error in closing a TsFileProcessor with an empty memtable

MrQuansy:
Now, when executing some inserting requests and then deleting all data, there will be an empty memtable in the tsfileProcessor. When executing flush request and closing this tsfileProcessor,  this request will be blocked.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileProcessor.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileProcessor.java
@@ -1273,7 +1273,7 @@ public class TsFileProcessor {
     logger.info("Start to end file {}", tsFileResource);
     long closeStartTime = System.currentTimeMillis();
     writer.endFile();
-    // tsFileResource.serialize();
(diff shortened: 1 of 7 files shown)


HTHou (commented):
Add more UT for this bug.

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileProcessor.java
```diff
@@ -1270,7 +1273,7 @@ private void endFile() throws IOException, TsFileProcessorException {
     logger.info("Start to end file {}", tsFileResource);
     long closeStartTime = System.currentTimeMillis();
     writer.endFile();
-    tsFileResource.serialize();
+    // tsFileResource.serialize();
```
HTHou, line 1276: Why do you remove this?

MrQuansy: fix

HTHou (approved):

PR #7973, opened 2022-11-14
[IOTDB-4923] Enable metric module in default config

SpriCoder:
In order to implement some functions, such as load balance, we need to enable metric module in default level.
1. Start no reporter in default config which means metrics can only be collected by API.
2. Minimize and specific core level metric.


SpriCoder:
The list of core level metrics in confignode are as follows:
1. The number of confignode in different status.
2. The number of datanode in different status.
3. The load of cpu of process: from OperatingSystemMXBean::getProcessCpuLoad
4. The time of cpu of process: from OperatingSystemMXBean::getProcessCpuTime
5. The max size of memory of process: from Runtime::maxMemory
6. The total size of memory of process: from Runtime::totalMemory
7. The free size of memory of process: from Runtime::freeMemory
8. The status of process: whether this thread is alive
9. The load of cpu: from OperatingSystemMXBean::getSystemCpuLoad
10. The number of cores of cpu
11.  The total size of physical memory: from OperatingSystemMXBean::getTotalPhysicalMemorySize
12. The free size of physical memory: from OperatingSystemMXBean::getFreePhysicalMemorySize
13. The total size of swap space: from OperatingSystemMXBean::getTotalSwapSpaceSize
14. The free size of swap space: from OperatingSystemMXBean::getFreeSwapSpaceSize
15. The size of committed virtual memory: from OperatingSystemMXBean::getCommittedVirtualMemorySize
16. The total size of system disk
17. The free size of system disk

SpriCoder:
Then I will test in 1C1D mode to compare the difference of performance.

qiaojialin (approved):

PR #7977, opened 2022-11-14
[IOTDB-4815] Apply SchemaCache for explicit timeseries query

MarcosZyk:
### Motivation

Currently, the DataNodeSchemaCache is only used for data insert. However, some data queries don't take patterns with wildcard and the target timeseries might be cached locally.

### Modification

If all the input patterns in a data query are explicit path without any wildcards and all the target timeseries can be taken from local cache, directly return the schemaTree taken from cache.


server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/ClusterSchemaFetcher.java
```diff
@@ -104,10 +106,67 @@ public ClusterSchemaTree fetchSchemaWithTags(PathPatternTree patternTree) {
...
+            new ArrayList<>(schemaPartition.getSchemaPartitionMap().keySet()));
+        return schemaTree;
+      }
     }
-    return executeSchemaFetchQuery(new SchemaFetchStatement(patternTree, templateMap, withTags));
+
+    schemaTree =
+        executeSchemaFetchQuery(new SchemaFetchStatement(patternTree, templateMap, withTags));
+
+    schemaCache.takeReadLock();
```
JackieTien97, line 154: write lock?

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/ClusterSchemaFetcher.java
```diff
@@ -104,10 +106,67 @@ public ClusterSchemaTree fetchSchemaWithTags(PathPatternTree patternTree) {
...
+            break;
+          } else {
+            schemaTree.mergeSchemaTree(cachedSchema);
+          }
+        }
+      } finally {
+        schemaCache.releaseReadLock();
+      }
+      if (isAllCached) {
+        SchemaPartition schemaPartition = partitionFetcher.getSchemaPartition(patternTree);
```
JackieTien97, line 144: We only need storage groups in `SchemaPartition`, I think it's too heavy to fetch it from partitionFetcher especially when cache miss, it may also need to fetch from remote config.
And even if cache hit, there is no need for us to do too much calculation in `partitionFetcher.getSchemaPartition`

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/ClusterSchemaFetcher.java
MarcosZyk: The schema cache is implemented based on Caffeine which is thread safe. The readLock is added to ensure that the schema cache clean operation, taking writeLock,  executed by delete timesereis or deactivate template will be effective.

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/ClusterSchemaFetcher.java
MarcosZyk: The storage group now is cached in SchemaCacheEntry too.

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/last/UpdateLastCacheOperator.java
```diff
@@ -106,7 +107,14 @@ public TsBlock next() {
 
     if (needUpdateCache) {
       TimeValuePair timeValuePair = new TimeValuePair(lastTime, lastValue);
-      lastCache.updateLastCache(fullPath, timeValuePair, false, Long.MIN_VALUE);
+      lastCache.updateLastCache(
+          ((DataDriverContext) operatorContext.getInstanceContext().getDriverContext())
+              .getDataRegion()
+              .getStorageGroupName(),
```
JackieTien97, line 113: since sg name won't change, you should make it as a field of operator instance instead of getting it through function call each time.

server/src/main/java/org/apache/iotdb/db/metadata/cache/DataNodeSchemaCache.java
```diff
@@ -100,34 +103,61 @@ public void releaseWriteLock() {
    */
   public ClusterSchemaTree get(PartialPath devicePath, String[] measurements) {
     ClusterSchemaTree schemaTree = new ClusterSchemaTree();
+    List<String> storageGroupList = new ArrayList<>();
```
JackieTien97, line 106: List or Set?

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/ClusterSchemaFetcher.java
```diff
@@ -104,10 +105,70 @@ public ClusterSchemaTree fetchSchemaWithTags(PathPatternTree patternTree) {
...
+          }
+        }
+        if (isAllCached) {
+          schemaTree.setStorageGroups(new ArrayList<>(storageGroupSet));
+          return schemaTree;
+        }
+      }
+
+      schemaTree =
+          executeSchemaFetchQuery(new SchemaFetchStatement(patternTree, templateMap, withTags));
```
JackieTien97, line 155: There is no need to put this time-costing operation in lock, remember to keep lock scope as small as possible.

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/ClusterSchemaFetcher.java
MarcosZyk: The schema fetch operation from cache or remote shall be locked together to guarantee that the fetched schema, maybe turned into blacklist after fetch during deleting timeseries, could be cleaned totally when cleaning DataNode schema cache during deleting timeseries.

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/last/UpdateLastCacheOperator.java
MarcosZyk: fixed

server/src/main/java/org/apache/iotdb/db/metadata/cache/DataNodeSchemaCache.java
MarcosZyk: fixed

server/src/main/java/org/apache/iotdb/db/mpp/common/schematree/ClusterSchemaTree.java
```diff
@@ -292,8 +293,8 @@ public static ClusterSchemaTree deserialize(InputStream inputStream) throws IOEx
    * @return database in the given path
    */
   @Override
-  public String getBelongedStorageGroup(String pathName) {
-    for (String storageGroup : storageGroups) {
+  public String getBelongedDatabase(String pathName) {
+    for (String storageGroup : databases) {
```
JackieTien97, line 297:
```suggestion
    for (String database : databases) {
```

JackieTien97 (changes requested):

JackieTien97 (approved):

PR #7995, opened 2022-11-15
[IOTDB-4209] Use .pst to accelerate recovery in SchemaFile mode 

Cpaulyz:
## Description

In the original process, recovery in SchemaFile mode will overwrite the .pst file and redo all the plans in MLog. 

However, it causes unnecessary additional I/O cost. Therefore, the records in the mlog can be removed after flushing the .pst file. When recovering, we only need to redo the plans in the mlog that have not been persisted into the .pst file.

### Environment 

Benchmark Config:

* DEVICE_NUMBER=100000
* SENSOR_NUMBER=10
* GROUP_NUMBER=1
* CLIENT_NUMBER=5

IoTDB Config:
* schemaEngineMode = "Schema_File"
* cachedMNodeSizeInSchemaFileMode = 50
* schemaRegionConsensusProtocolClass = ConsensusFactory.SIMPLE_CONSENSUS

### Do not use check point

Create timeseries ≈ 41.59 second
Recover [root.test.g_0.SchemaRegion[0]] spend: 30612 ms

### Use check point
 
Create timeseries ≈ 41.56 second
Recover [root.test.g_0.SchemaRegion[0]] spend: 1309 ms


server/src/main/java/org/apache/iotdb/db/metadata/schemaregion/SchemaRegionSchemaFileImpl.java
```diff
@@ -265,29 +304,42 @@ public void forceMlog() {
...
+      if (logDescriptionFile.exists()) {
+        try (FileInputStream fileInputStream = new FileInputStream(logDescriptionFile)) {
+          MLogDescription mLogDescription = new MLogDescription();
+          mLogDescription.deserialize(fileInputStream);
+          startIndex = mLogDescription.getCheckPoint();
+          logger.info("MLog recovery check point: {}", startIndex);
+        } catch (IOException e) {
+          logger.warn("Can not get check point in MLogDescription file, use default value 0.");
+        }
+      }
```
MarcosZyk, line 331: encapsulate this into ```MLogDescriptionReader```

server/src/main/java/org/apache/iotdb/db/metadata/schemaregion/SchemaRegionSchemaFileImpl.java
```diff
@@ -297,20 +349,26 @@ private int initFromLog() throws IOException {
...
-        idx++;
+        nextIndex++;
       } catch (Exception e) {
         logger.error("Parse mlog error at lineNumber {} because:", idx, e);
         break;
       }
-      if (plan == null) {
+      if (nextIndex <= startIndex || plan == null) {
         continue;
       }
```
MarcosZyk, line 373: The logs before checkpoint will not be redo but still read from disk. What about recording the position/offset of checkpoint in mlog? The read operation of logs before checkpoint will be saved. 

If this takes effort to implement such mechanism. What about we archive the executed logs into another log file, like ```mlog_archive_xxx```, and they won't be read or executed in next recover process, since if we successfully recover, all the schema has already been stored in schema file.

server/src/main/java/org/apache/iotdb/db/metadata/schemaregion/SchemaRegionSchemaFileImpl.java
Cpaulyz: done

server/src/main/java/org/apache/iotdb/db/metadata/schemaregion/SchemaRegionSchemaFileImpl.java
Cpaulyz: I use position as checkpoint now. Recovery time has been reduced from 3000ms to 1300ms.

PR #7997, opened 2022-11-15
[IOTDB-4401] DataNode should not be stopped  when any regions are migrated failed

Beyyes:
When executed remove-datanode.sh, if some regions are migrated failed, the DataNode should not be stopped（Always be Removing status）.


CRZbulabula (approved):
LGTM~

MiniSho:
LGTM

PR #8011, opened 2022-11-16
[IOTDB-4414] On using the CLI with OpenID Connect Auth enabled on Server side fails

CloudWise-Lukemiao:
[Cluster| NewServer]On using the CLI with OpenID Connect Auth enabled on Server side fails


HTHou:
Hi, please fix tests and conflicts.

HTHou (approved):

PR #8012, opened 2022-11-16
[IOTDB-4727] use AutoScalingBufferWriteTransport to avoid byte buffer overflow

SzyWilliam:
see https://issues.apache.org/jira/browse/IOTDB-4727
see also https://issues.apache.org/jira/browse/RATIS-1743

When TSStatus serialization size exceeds 1024 Bytes, the serialization will fail and the ByteBuffer returned is null. We should use the AutoScalingBufferWriteTransport to support arbitrary TSStatus size.


OneSizeFitsQuorum (approved):
LGTM

PR #8016, opened 2022-11-16
[IOTDB-4850] Disable first election feature

SzyWilliam:
## What is First Election
First election is a new feature recently introduced by Ratis 2.4.0, the original issue is https://issues.apache.org/jira/browse/RATIS-1638. The motivation of first election is to reduce unavailable time during start up.
However, this new feature is experimental. There are some known issues when applying in IoTDB. For example, when first election timeout is 50ms, migrate region will first call createPeer then call addPeer. If election timeouts before addPeer (after createPeer), the leader will tell it to shutdown and the migration will fail.
failure examples can be seen at https://github.com/apache/iotdb/actions/runs/3478206562/jobs/5815284797. 

Therefore, this PR proposes to disable first election feature.

This feature may be re-opened when addNewNodeToExistedGroup is implemented and stand the test of tests/production env.


OneSizeFitsQuorum (approved):
LGTM

PR #8019, opened 2022-11-16
[IOTDB-4957] Add check for create pipesink and optimize re-connection

Cpaulyz:
## Description

https://issues.apache.org/jira/browse/IOTDB-4957

1. Add check for create pipesink. For IoTDBPipeSink, it will response exception if ip address is not standard IPv4 address.
2. To avoid busy waiting and large amount of duplicate logs. I optimize logic of re-connection. If sender lost connection to receiver, it will hang up until scheduled heartbeat task successfully reconnect to receiver. Default interval of heartbeat is 30 seconds. 

image

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/sync/transport/client/SenderManager.java
+++ b/server/src/main/java/org/apache/iotdb/db/sync/transport/client/SenderManager.java
@@ -36,9 +36,14 @@ import org.slf4j.LoggerFactory;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
+import java.util.concurrent.LinkedBlockingQueue;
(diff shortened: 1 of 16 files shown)


yschengzi (approved):
LGTM

server/src/main/java/org/apache/iotdb/db/sync/transport/client/SenderManager.java
```diff
@@ -111,36 +145,87 @@ public void close() throws PipeException {
...
+                                "Transfer PipeData %s error, skip it.",
+                                pipeData.getSerialNumber())));
+              }
+              pipe.commit(dataRegionId);
             }
-            pipe.commit(dataRegionId);
+          } catch (SyncConnectionException e) {
+            // If failed to connect to receiver, it will hang up until scheduled heartbeat task
+            // successfully reconnect to receiver.
+            logger.error(String.format("Connect to receiver %s error, because %s.", pipeSink, e));
```
HTHou, line 224:
```suggestion
            logger.error("Connect to receiver {} error, because ", pipeSink, e);
```

HTHou (approved):

PR #8024, opened 2022-11-17
[IOTDB-4969] Fix the distribution plan for AggregationQuery with align by device

xingtanzjr:
## Description 
Before this change, the distribution plan is not correct when one device locates in more than one DataRegions.
This PR fixed this issue and add more tests for raw aggregation and slidingWndow aggregation

## Test 
Test raw Aggregation 
image

Test  Aggregation with SlidingWindow:
image


Wei-hao-Li (approved):

liuminghui233 (approved):
LGTM :-)

PR #8028, opened 2022-11-17
[IOTDB-4970] Enhanced sync safety management

Cpaulyz:
## Description

Before the sync sender can send data, a handshake with the receiver is required. The current handshake policy is to verify the sender's IP address based on the receiver's white list. However, the current sender's ip address is provided by the sender and may be falsified leading to security issues.

 

1. This ISSUE will get the ip address from the thrift connection.
2. The default whitelist is set to 127.0.0.1/32, which rejects all the sender's ip except localhost.
3. Some optimizations have been made to handle cases of handshake failure.

See test case in https://apache-iotdb.feishu.cn/docx/NsLPdhkC1oa1QCxKOMlcj9RxntK#IM0wdcUQ8oSuMCxaVPXcxjlunlg chapter 2.4~2.5.


yschengzi (approved):
LGTM

HTHou (approved):

PR #8031, opened 2022-11-17
[To rel/0.13] upgrade audit log

caozj1011:
1.  added whether to enable audit logs during open session.
2. replace the original AUDIT_LOGGER.
3. when the Audit log level is set to LOGGER, the write operation is changed from DEBUG to INFO.


server/src/main/java/org/apache/iotdb/db/service/IoTDB.java
```diff
@@ -194,6 +195,13 @@ private void setUp() throws StartupException, QueryProcessException {
     registerManager.register(ContinuousQueryService.getInstance());
     registerManager.register(MetricService.getInstance());
 
+    if (IoTDBDescriptor.getInstance().getConfig().isEnableExternalService()) {
+      ServiceLoader<IService> iServices = ServiceLoader.load(IService.class);
+      for (IService iService : iServices) {
+        registerManager.register(iService);
+      }
+    }
+
```
JackieTien97, line 204: Why here re-register all the services again?

metrics/interface/src/main/java/org/apache/iotdb/metrics/config/MetricConfigDescriptor.java
```diff
@@ -122,6 +123,13 @@ private String getPropsUrl() {
         }
       }
     }
+    if (url == null) {
+      URL uri = MetricConfig.class.getResource("/" + MetricConstant.CONFIG_NAME);
+      logger.info("MetricConfig.class.getResource,{}", uri);
+      if (uri != null) {
+        return uri.getPath();
+      }
+    }
```
JackieTien97, line 132: Add some comments about this if-branch, actually I can't understand its motivation.

session/src/main/java/org/apache/iotdb/session/SessionConnection.java
```diff
@@ -141,7 +141,7 @@ private void init(EndPoint endPoint) throws IoTDBConnectionException {
     openReq.setPassword(session.password);
     openReq.setZoneId(zoneId.toString());
     openReq.putToConfiguration("version", session.version.toString());
-
+    openReq.putToConfiguration("enableAudit", String.valueOf(session.enableAudit));
```
JackieTien97, line 144: same as above

server/src/main/java/org/apache/iotdb/db/service/basic/ServiceProvider.java
```diff
@@ -206,21 +206,26 @@ public BasicOpenSessionResp login(
...
+
     } else {
       openSessionResp.setMessage(loginMessage != null ? loginMessage : "Authentication failed.");
       openSessionResp.setCode(TSStatusCode.WRONG_LOGIN_PASSWORD_ERROR.getStatusCode());
-      AUDIT_LOGGER.info("User {} opens Session failed with an incorrect password", username);
+      session.setUsername(username);
+      if (!AuditLogUtils.LOG_LEVEL_NONE.equals(CONFIG.getAuditLogStorage())) {
+        AuditLogUtils.writeAuditLog(
+            String.format("User %s opens Session failed with an incorrect password", username),
+            true);
```
JackieTien97, line 224: It means that we need to record all the failed login info?

server/src/main/java/org/apache/iotdb/db/service/basic/ServiceProvider.java
```diff
@@ -206,21 +206,26 @@ public BasicOpenSessionResp login(
...
 
-      SESSION_MANAGER.supplySession(session, username, zoneId, clientVersion);
-      LOGGER.info(
-          "{}: Login status: {}. User : {}, opens Session-{}",
-          IoTDBConstant.GLOBAL_DB_NAME,
-          openSessionResp.getMessage(),
-          username,
-          session);
+      SESSION_MANAGER.supplySession(session, username, zoneId, clientVersion, enableAudit);
+      if (!AuditLogUtils.LOG_LEVEL_NONE.equals(CONFIG.getAuditLogStorage())) {
```
JackieTien97, line 210: It seems that it will be constant value after iotdb starting, no need to do this judgement each time.

server/src/main/java/org/apache/iotdb/db/service/thrift/impl/TSServiceImpl.java
```diff
@@ -345,6 +351,14 @@ private IoTDBConstant.ClientVersion parseClientVersion(TSOpenSessionReq req) {
     return IoTDBConstant.ClientVersion.V_0_12;
   }
 
+  private boolean parseEnableAudit(TSOpenSessionReq req) {
+    Map<String, String> configuration = req.configuration;
+    if (configuration != null && configuration.containsKey("enableAudit")) {
```
JackieTien97, line 356: use constant value instead of this raw string.

server/src/main/java/org/apache/iotdb/db/service/thrift/impl/TSServiceImpl.java
```diff
@@ -345,6 +351,14 @@ private IoTDBConstant.ClientVersion parseClientVersion(TSOpenSessionReq req) {
...
+      return Boolean.parseBoolean(configuration.get("enableAudit"));
```
JackieTien97, line 357: same as above

server/src/main/java/org/apache/iotdb/db/service/thrift/impl/TSServiceImpl.java
```diff
@@ -345,6 +351,14 @@ private IoTDBConstant.ClientVersion parseClientVersion(TSOpenSessionReq req) {
...
+    }
+    return true;
```
JackieTien97, line 359: default true? shouldn't be false?

server/src/main/java/org/apache/iotdb/db/service/thrift/impl/TSServiceImpl.java
```diff
@@ -222,8 +224,10 @@ public TSExecuteStatementResp call() throws Exception {
       plan.setLoginUserName(username);
 
       QUERY_FREQUENCY_RECORDER.incrementAndGet();
-      AUDIT_LOGGER.debug("Session {} execute Query: {}", session, statement);
-
+      if (!AuditLogUtils.LOG_LEVEL_NONE.equals(conf.getAuditLogStorage())) {
```
JackieTien97, line 227: same as above, you check it all throught your pr and I won't point it out in the remaing reviews.

server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
```diff
@@ -875,11 +875,15 @@ public class IoTDBConfig {
...
   // determines whether audit logs are written to log files or IoTDB
-  private String auditLogStorage = AuditLogUtils.LOG_LEVEL_NONE;
+  private String auditLogStorage = AuditLogUtils.LOG_LEVEL_LOGGER;
 
   // determines whether audit logs record IoTDB write operation
   private boolean enableAuditLogWrite = false;
 
+  private int remoteConfigPort = 8667;
+
+  private boolean enableExternalService = false;
```
JackieTien97, line 885: It seems that these two fields are totally not used.

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/service/basic/ServiceProvider.java
caozj1011: yes, the user wants to know which illegitimate users are connected to the database

caozj1011 (commented):
change review suggest

caozj1011 (commented):
fix review suggest

JackieTien97 (approved):

PR #8037, opened 2022-11-18
[IOTDB-4668] Add Database name length constraint

MarcosZyk:
## Description

Since the name of database will be used as directory name to create directories in file system, the length shall not exceed the max length of file name. The length of database name in mysql shall not exceed 64, which is an immutable config, thus we add this constraint to our database too.


server/src/main/java/org/apache/iotdb/db/mpp/plan/execution/config/executor/ClusterConfigTaskExecutor.java
```diff
@@ -197,6 +199,13 @@ public static ClusterConfigTaskExecutor getInstance() {
   public SettableFuture<ConfigTaskResult> setStorageGroup(
       SetStorageGroupStatement setStorageGroupStatement) {
     SettableFuture<ConfigTaskResult> future = SettableFuture.create();
+    String storageGroupPath = setStorageGroupStatement.getStorageGroupPath().getFullPath();
+    if (storageGroupPath.length() > MAX_DATABASE_NAME_LENGTH) {
+      future.setException(
+          new IllegalPathException(
+              storageGroupPath, "the length of database name shall not exceed 64."));
```
HTHou, line 206:
```suggestion
              storageGroupPath, "the length of database name shall not exceed " + MAX_DATABASE_NAME_LENGTH));
```

MarcosZyk: Fixed.

qiaojialin (approved):

PR #8053, opened 2022-11-18
[IOTDB-4809] Broadcast the RegionRouteMap to all DataNodes except the unknown ones

CRZbulabula:
Fix bug that the Removing and ReadOnly DataNodes can't receive the latest RegionRouteMap


Beyyes (approved):
LGTM

PR #8055, opened 2022-11-19
[IOTDB-3671] Make thread group of ProcedureExecutor warn

Beyyes:
## Description

Make the log level of `thread group in ProcedureExecutor` as warn.
These threads cannot be stopped may be used by other module such as ASyncClientManger, and it's no need to destroy it.


wangchao316 (approved):
LGTM

PR #8076, opened 2022-11-21
[IOTDB-5008] Fix SHUTDOWN errors in ratis create peer

SzyWilliam:
When adding a new member to an existed group, the correct order is:
1. call createPeer on new member with **EMPTY GROUP**.
2. call addPeer on original group with full group.

This PR enables RatisConsensus to handle EMPTY GROUP create peer requests.


Beyyes (approved):
LGTM

PR #8083, opened 2022-11-22
[IOTDB-5014] fix ratis param

SzyWilliam:
Fix some params that don't take effect.

### DIFF
--- a/node-commons/src/assembly/resources/conf/iotdb-common.properties
+++ b/node-commons/src/assembly/resources/conf/iotdb-common.properties
@@ -997,7 +997,6 @@
 # config_node_ratis_snapshot_trigger_threshold=400000
 # schema_region_ratis_snapshot_trigger_threshold=400000
 # data_region_ratis_snapshot_trigger_threshold=400000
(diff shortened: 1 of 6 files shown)


MarcosZyk (approved):

node-commons/src/assembly/resources/conf/iotdb-common.properties
```diff
@@ -997,7 +997,6 @@
 # config_node_ratis_snapshot_trigger_threshold=400000
 # schema_region_ratis_snapshot_trigger_threshold=400000
 # data_region_ratis_snapshot_trigger_threshold=400000
-# config_node_simple_consensus_snapshot_trigger_threshold=400000
```
wangchao316, line 1000: why delete trigger threshold?

SzyWilliam: `config_node_simple_consensus_trigger_threshold` seems to be mistakenly placed under ratis configuration section. It is not used anywhere in code, so I deleted it.

PR #8086, opened 2022-11-22
[IOTDB-4999] Configure maximum of connection for config node internal service

MrQuansy:


node-commons/src/main/java/org/apache/iotdb/commons/client/ClientManager.java
```diff
@@ -56,7 +56,10 @@ public V borrowClient(K node) throws IOException {
       throw e;
     } catch (Exception e) {
       // external doesn't care of other exceptions
-      String errorMessage = String.format("Borrow client from pool for node %s failed.", node);
+      String errorMessage =
+          String.format(
+              "Borrow client from pool for node %s failed， you need to increase max_connection_for_internal_service.",
```
HTHou, line 61:
```suggestion
              "Borrow client from pool for node %s failed, you need to increase max_connection_for_internal_service.",
```

PR #8089, opened 2022-11-22
[IOTDB-4926] Add error message when the key of parameter attributes is not quoted

lancelly:


integration-test/src/test/java/org/apache/iotdb/db/it/udf/IoTDBUDTFAlignByTimeQueryIT.java
```diff
@@ -156,6 +156,36 @@ private static void registerUDF() {
...
+  public void queryWithUnquotedAttributes() {
+    try (Connection connection = EnvFactory.getEnv().getConnection();
+        Statement statement = connection.createStatement()) {
+      try {
+        statement.executeQuery("select sum_sec(s1, 'interval'=3) from root.udf.d1;");
+      } catch (SQLException e) {
+        assertTrue(e.getMessage().contains("Attributes of functions should be quoted"));
+      }
+      try {
+        statement.executeQuery("select sum_sec(s1, interval=3) from root.udf.d1;");
```
liuminghui233, line 169: This SQL is legal in some cases. For example, the second parameter of the function is a sequence of Boolean types.

I think we need to add checks on function signatures.

liuminghui233 (changes requested):

liuminghui233 (approved):
LGTM :-)

PR #8097, opened 2022-11-22
[IOTDB-5021] Using Junit parameterized test to optimize schema IT framework

Cpaulyz:
## Description

In order to make the integrate tests cover Schema_File mode and Memory mode, we originally needed to create two classes for each IT, such as IoTDBMetadataFetchIT and IoTDBMetadataFetchSchemaFileIT, which would result in a large number of classes.

This ISSUE introduces a JUnit parametric test approach. An `AbstractSchemaIT` abstract class is created for all schema test, and its subclasses will automatically run tests in both schema engine modes.

> reference:
> https://blog.csdn.net/w605283073/article/details/80051229
> https://stackoverflow.com/questions/27745691/how-to-combine-runwith-with-runwithparameterized-class

image


integration-test/src/main/java/org/apache/iotdb/it/framework/IoTDBTestRunnerWithParameters.java
```diff
@@ -0,0 +1,59 @@
...
+
+import org.junit.runner.Description;
+import org.junit.runner.notification.RunNotifier;
+import org.junit.runners.model.FrameworkMethod;
+import org.junit.runners.model.InitializationError;
+import org.junit.runners.parameterized.BlockJUnit4ClassRunnerWithParameters;
+import org.junit.runners.parameterized.TestWithParameters;
+import org.slf4j.Logger;
+
+public class IoTDBTestRunnerWithParameters extends BlockJUnit4ClassRunnerWithParameters {
```
ericpai, line 31: Can this runner be merged to `IoTDBTestRunner` to simplify the framework usage, which means that 
`IoTDBTestRunner` can be used with an optional parameter.

ericpai (changes requested):

integration-test/src/test/java/org/apache/iotdb/db/it/schema/AbstractSchemaIT.java
```diff
@@ -0,0 +1,97 @@
...
+        SchemaTestMode.SchemaFileFullMemory,
+        SchemaTestMode.SchemaFilePartialMemory,
+        SchemaTestMode.SchemaFileNonMemory);
+  }
+
+  public AbstractSchemaIT(SchemaTestMode schemaTestMode) {
+    this.schemaTestMode = schemaTestMode;
+  }
+
+  @Before
```
ericpai, line 61: My guess about the unexpected test failures due to 'cluster can't work' is that we declare `@Before` and `@After` in both super class and child class. Just use them in child one.

ericpai (changes requested):

integration-test/src/main/java/org/apache/iotdb/it/framework/IoTDBTestRunnerWithParameters.java
Cpaulyz: It seems difficult to implement because Parameize requires the runner to be a subclass of `BlockJUnit4ClassRunnerWithParameters`, but `IoTDBRunner` is a subclass of `BlockJUnit4ClassRunner`.

ericpai (approved):

MarcosZyk (approved):

PR #8106, opened 2022-11-23
[IOTDB-5027] The RegionRouteMap doesn't update after the leader Node is Unknown

CRZbulabula:
Add new feature to LEADER route policy: the leader RegionReplica no longer has the highest priority if its status isn't Running.


Beyyes (approved):
LGTM

PR #8108, opened 2022-11-23
[IOTDB-5018] Fix NPE when using uppercase function name in GROUP BY LEVEL

liuminghui233:
see IOTDB-5018 & IOTDB-4882


wangchao316 (approved):
LGTM

PR #8134, opened 2022-11-24
[IOTDB-5031] Make snapshot taking in shutdown hook parallel

THUMarkLau:
See [IOTDB-5031](https://issues.apache.org/jira/browse/IOTDB-5031).

Trigger a snapshot when shutting down IoTDB could help reserve the compaction result, this pr makes it parallel to accelerate thisp process.


OneSizeFitsQuorum (approved):
LGTM

PR #8140, opened 2022-11-24
[IOTDB-4066] Allocate new RegionGroups through double keyword sort

CRZbulabula:
Embed the double keyword sort algorithm into the RegionGroup allocation process:

1. First key: select the DataNode that currently has the minimum number of Regions.
2. Second key: select the DataNode that currently has the maximum free disk space.

You can see the [RegionGroup allocation algorithm](https://apache-iotdb.feishu.cn/docs/doccnHuDkDD6RiwLjaej3hkA9Vb) doc to get more details of this algorithm.

To varify this test, I also enhance the GreedyRegionGroupAllocatorTest. But the ClusterIT framework is hard to simulate this scenario since we can't specific the DataNode's disk space in ClusterIT framework currently.


Beyyes (approved):
LGTM

PR #8143, opened 2022-11-24
[IOTDB-5039] Add retry for dispatching load plan node

yschengzi:
problem:
it's hard to execute a successful load statement when the data set is large and the network is shit.

solution:
Add retry for dispatching load plan node.


qiaojialin (changes requested):
It's none of the business of the network, just the endFIle() is time-consuming, pay attention to your code.

PR #8146, opened 2022-11-25
[IOTDB-5034] Change log level

SzyWilliam:
# Why change log level

Currently RatisConsensus treats failed consensus operations all as error level. However, some failures are expected and can be retries by caller. So I changed the level to INFO, leaving the ERROR/WARN decision to caller.


Beyyes (approved):

qiaojialin:
check if need the  stack

PR #8170, opened 2022-11-25
[IOTDB-5049] Add parameters enable leader balance for specifed consensus protocol class

CRZbulabula:
Add parameters:

1.  **enable_auto_leader_balance_for_ratis:** enable balance the RegionGroups' leaders for Ratis consensus protocol, default is false since the Ratis is unstable for this function.
2. **enable_auto_leader_balance_for_multileader:** enable balance the RegionGroups' leaders for MultiLeader consensus protocol, default is true since the MultiLeader is depends on this function to distribute leader.


Beyyes (approved):
LGTM

PR #8175, opened 2022-11-25
[To rel/1.0] [IOTDB-4902]Optimize process logic for aggregation when there is only one data region

Wei-hao-Li:
Doc is here[https://apache-iotdb.feishu.cn/docs/doccnuuXYJo0AeM34YhCq61o6Ce#]( here);
Two bugs about 'STATIC' step of Aggregator has been fixed:
1. `set final` method  of Accumulator has not been implemented correctly;
2. Serialize & de of Aggregator ignore 'STATIC' step.


server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/distribution/SourceRewriter.java
```diff
@@ -907,45 +933,94 @@ private boolean isAggColumnMatchExpression(String columnName, Expression express
     return columnName.contains(expression.getExpressionString());
   }
 
-  private List<SeriesAggregationSourceNode> splitAggregationSourceByPartition(
-      PlanNode root, DistributionPlanContext context) {
+  /**
+   * This method is used for rewriting second step AggregationNode like GroupByLevelNode and
+   * GroupByTagNode, so the first step AggregationNode's step will be {@link
```
JackieTien97, line 938: also for `SlidingWindowAggregationNode`, remember to add it.

server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/distribution/SourceRewriter.java
```diff
@@ -907,45 +933,94 @@ private boolean isAggColumnMatchExpression(String columnName, Expression express
...
+          .getAggregationDescriptorList()
+          .forEach(
+              d -> {
+                if (isSingle) {
+                  d.setStep(AggregationStep.SINGLE);
+                } else {
+                  d.setStep(AggregationStep.PARTIAL);
+                  LogicalPlanBuilder.updateTypeProviderByPartialAggregation(
+                      d, context.queryContext.getTypeProvider());
+                }
```
JackieTien97, line 998: Does `SINGLE` need to updateTypeProvider?

JackieTien97: Confirm it with @liuminghui233

server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/distribution/SourceRewriter.java
```diff
@@ -588,61 +589,91 @@ public PlanNode visitSlidingWindowAggregation(
...
+      // if the aggregated series have only one region,
+      // upstream will give the final aggregate result,
+      // the step of this series' aggregator will be `STATIC`
+      List<SeriesAggregationSourceNode> sources = new ArrayList<>();
+      Map<PartialPath, Integer> regionCountPerSeries = new HashMap<>();
+      sourceGroup = splitAggregationSourceByPartition(root, context, sources, regionCountPerSeries);
+
+      if (sources.size() == root.getChildren().size()) {
+        newRoot = new VerticallyConcatNode(context.queryContext.getQueryId().genPlanNodeId());
+      } else {
```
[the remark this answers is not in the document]
JackieTien97: Even thought it may be right to use `sources.size() == root.getChildren().size()`, I still think we shouldn't rely on that, use a boolean parameter as we discussed tonight may be better.

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/mpp/aggregation/FirstValueAccumulator.java
```diff
@@ -130,7 +130,6 @@ public void addStatistics(Statistics statistics) {
   @Override
   public void setFinal(Column finalResult) {
     reset();
-    hasCandidateResult = true;
     firstValue.setObject(finalResult.getObject(0));
```
JackieTien97, line 133: BTW, I think it's better to optimize this method in this pr to avoid unbox and box. You can refer to `addIntermediate` and `addInput`, using switch-case to get from exact type.
Remember to optimize all the implementation class of `Accumulator`.

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/distribution/SourceRewriter.java
liuminghui233: When the step output is not partial, no new aggregation functions are introduced and we don't need to update the TypeProvider.

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/join/VerticallyConcatOperator.java
```diff
@@ -0,0 +1,206 @@
...
+    // build each ValueColumn in every inputTsBlock
+    int valueBuilderIndex = 0; // indicate which valueColumnBuilder should use
+    for (int i = 0; i < inputOperatorsCount; i++) {
+      int currTsBlockIndex = inputIndex[i];
+      for (Column column : inputTsBlocks[i].getValueColumns()) {
+        for (int row = 0; row < maxRowCanBuild; row++) {
+          if (column.isNull(currTsBlockIndex)) {
+            valueColumnBuilders[valueBuilderIndex].appendNull();
+          } else {
+            valueColumnBuilders[valueBuilderIndex].writeObject(column.getObject(currTsBlockIndex));
```
JackieTien97, line 131: Remember that `writeObject` is ineffective, never use that if possible.

JackieTien97, line 131:
```suggestion
            valueColumnBuilders[valueBuilderIndex].write(column, currTsBlockIndex);
```

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/join/VerticallyConcatOperator.java
```diff
@@ -0,0 +1,206 @@
...
+    for (int row = 0; row < maxRowCanBuild; row++) {
+      // build TimeColumn
+      timeColumnBuilder.writeLong(firstTimeColumn.getLong(inputIndex[0]));
+      tsBlockBuilder.declarePosition();
+    }
+
+    // build each ValueColumn in every inputTsBlock
+    int valueBuilderIndex = 0; // indicate which valueColumnBuilder should use
+    for (int i = 0; i < inputOperatorsCount; i++) {
+      int currTsBlockIndex = inputIndex[i];
```
JackieTien97, line 125: Actually, I didn't see  that you increase this `currTsBlockIndex`, it seems that you always append the same value in the for-loop

JackieTien97: If so, add a new UT for this Operator.

JackieTien97 (changes requested):

PR #8177, opened 2022-11-25
[IOTDB-5004] [Metrics] Fix the seq file size in grafana is inconsistent with the actual query

THUMarkLau:
See [IOTDB-5004](https://issues.apache.org/jira/browse/IOTDB-5004).

The reason for this bug is that, when reducing the size of deleted tsfile after compaction, the deleted files return a size of zero. This pr fixes it by calculating the total size of deleted file before they are deleted, and reducing it after deletion of the files.


SpriCoder (approved):
LGTM~

qiaojialin (approved):

PR #8223, opened 2022-11-28
[IOTDB-5068] Fix ComapctionSchedulerTest

THUMarkLau:
See [IOTDB-5068](https://issues.apache.org/jira/browse/IOTDB-5068).

The assertion at line 432 is not reasonable. Before this assertion, the test waits for all sequence space compaction task finish. After that, it asserts all unsequence files is compacted to sequence space, which doesn't have to happend, so it may fail some times.  We don't need this assertion, just remove it and the test has already assertted the num of sequence files and unsequence files in the end of the test.


qiaojialin:
will merge rel/1.0 later

PR #8230, opened 2022-11-28
[To rel/1.0] [IOTDB-4940] Optimize query fetch data partition process while containg unclosed time filter

JackieTien97:
Optimize sqls like
where time < 10
where time > 20
where time < 10 or time>20


node-commons/src/main/java/org/apache/iotdb/commons/partition/DataPartitionQueryParam.java
```diff
@@ -28,12 +28,31 @@ public class DataPartitionQueryParam {
   private String devicePath;
   private List<TTimePartitionSlot> timePartitionSlotList = new ArrayList<>();
 
+  // it will be set to true in query when there exist filter like: time <= XXX
+  // (-oo, timePartitionSlotList.get(0)]
+  private boolean needLeftAll = false;
+
+  // it will be set to true query when there exist filter like: time >= XXX
+  // [timePartitionSlotList.get(timePartitionSlotList.size() - 1), +oo)
+  private boolean needRightAll = false;
```
CRZbulabula, line 37: The annotation shows that these two parameters indicate two close intervals. However in the implementation of SeriesPartitionTable, these two parameters indicate two open intervals.

CRZbulabula (commented):
PTAL~

node-commons/src/main/java/org/apache/iotdb/commons/partition/DataPartitionQueryParam.java
Wei-hao-Li: The margin partition will be added （if need）in  previous process code. 
image

CRZbulabula (approved):
LGTM～

PR #8231, opened 2022-11-28
[To rel/1.0] [IOTDB-5074] Add storage group name and region id to the path of snapshot

THUMarkLau:
See [IOTDB-5074](https://issues.apache.org/jira/browse/IOTDB-5074).

Current snapshot path contains only snapshot-id, which is log index in ratis, and it may be duplicated. The snapshot path should be unique for each snapshot. This pr fix it by adding storage group name and region id to the path of snapshot.


OneSizeFitsQuorum (approved):
LGTM

PR #8234, opened 2022-11-29
[To rel/1.0][IOTDB-5045] fix delete database, wal and tsfile still left

HTHou:
## Description

Details see IOTDB-5045


server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
```diff
@@ -176,7 +176,7 @@ public class IoTDBConfig {
 
   // region Write Ahead Log Configuration
   /** Write mode of wal */
-  private volatile WALMode walMode = WALMode.ASYNC;
+  private volatile WALMode walMode = WALMode.DISABLE;
```
HTHou, line 179:
```suggestion
  private volatile WALMode walMode = WALMode.ASYNC;
```

qiaojialin (approved):

PR #8236, opened 2022-11-29
[To rel/1.0] [IOTDB-5059] Remove redundant database check in SchemaEngine

MarcosZyk:
## Description

Since the database name has already been checked on configNode when creating database, there's no need to check it again in SchemaEngine. We suppose all the internal requests are legal and SchemaEngine only need to execute them.

The async physical deletion of schemaRegion affects the creation of new region of new database with nested name, e.g. old db is root.udf and new db is root.udf.load, which results in the creation failure of new region and this shall not happen.


bigreybear (approved):
LGTM~

PR #8240, opened 2022-11-29
[To rel/1.0] [IOTDB-5047] Fix iotdb.jmx.port missing from data-env.sh

MiniSho:
Add JMX_PORT to system properties in data-env.sh/bat
ConfigNode:
0c85b466fe5c01579ee3e66210f532e

DataNode:
1417e2f7ccdc1f2ff337331f0117f89


jixuan1989:
Only when JMX_LOCAL=false, the port can be set.

jixuan1989:
A simple solution:

Change IOTDB_JMX_PORT from  "iotdb.jmx.port" to "com.sun.management.jmxremote.port"
then there is no need to modify the scripts.
```
-  public static final String IOTDB_JMX_PORT = "iotdb.jmx.port";
+  public static final String IOTDB_JMX_PORT = "com.sun.management.jmxremote.port";
```

jixuan1989:
Besides, the StartChecker is incorrect as even when we start a confignode, it also says ".....datanode-env.sh"

jixuan1989:
> A simple solution:
> 
> Change IOTDB_JMX_PORT from "iotdb.jmx.port" to "com.sun.management.jmxremote.port" then there is no need to modify the scripts.
> 
> ```
> -  public static final String IOTDB_JMX_PORT = "iotdb.jmx.port";
> +  public static final String IOTDB_JMX_PORT = "com.sun.management.jmxremote.port";
> ```

Better solution:

set `JMX_LOCAL` as a variable to IoTDB (modify the -env scripts). Then in StartChecker.java, you should check whether the user wants to monitor JMX locally or open its remote port. In this way, we can give users more clear tips.

Beyyes (approved):
LGTM

PR #8246, opened 2022-11-29
[To rel/1.0] Add CUSTOM policy for schema region group

Beyyes:


node-commons/src/assembly/resources/conf/iotdb-common.properties
```diff
@@ -84,6 +84,18 @@
 # Datatype: Double
 # schema_region_per_data_node=1.0
 
+# The policy of extension SchemaRegionGroup for each Database.
+# These policies are currently supported:
+# 1. CUSTOM(Each Database will allocate schema_region_group_per_database RegionGroups as soon as created)
+# 2. AUTO(Each Database will automatically extend SchemaRegionGroups based on the data it has)
+# Datatype: String
+# schema_region_group_extension_policy=AUTO
```
CRZbulabula, line 92: Will it be better if we move the parameter `schema_region_group_extension_policy` ahead of the `schema_region_per_data_node`?

integration-test/src/test/java/org/apache/iotdb/confignode/it/partition/IoTDBCustomRegionGroupExtensionIT.java
```diff
@@ -88,6 +92,12 @@ public void setUp() throws Exception {
 
     originalTimePartitionInterval = CONF.getTimePartitionInterval();
 
+    originalSchemaRegionGroupExtensionPolicy = CONF.getSchemaRegionGroupExtensionPolicy();
+    CONF.setSchemaRegionGroupExtensionPolicy(testSchemaRegionGroupExtensionPolicy);
+
+    originalSchemaRegionGroupPerDatabase = CONF.getSchemaRegionGroupPerDatabase();
+    CONF.setSchemaRegionGroupPerDatabase(testSchemaRegionGroupPerDatabase);
```
CRZbulabula, line 99: Need to add ClusterIT for these to parameters

CRZbulabula (changes requested):
PTAL~

qiaojialin (approved):

PR #8259, opened 2022-11-30
[To rel/1.0][IOTDB-4730] Database name check when passed in Session

MarcosZyk:
## Description

There's no check of raw string of database name passed in Session, which may result in some internal execution error. 
Add the check when generating statement.


HTHou (approved):

MarcosZyk:
Replaced by https://github.com/apache/iotdb/pull/8266

PR #8264, opened 2022-11-30
[To rel/1.0] add IT for IoTDBCustomRegionGroupExtension

Beyyes:


integration-test/src/test/java/org/apache/iotdb/confignode/it/partition/IoTDBCustomRegionGroupExtensionIT.java
```diff
@@ -125,7 +132,7 @@ public void tearDown() {
 
   @Test
   public void testCustomDataRegionGroupExtensionPolicy()
```
CRZbulabula, line 134:
```suggestion
  public void testCustomRegionGroupExtensionPolicy()
```

CRZbulabula (commented):
PTAL～

CRZbulabula (approved):
LGTM～

PR #8265, opened 2022-11-30
[To multi_tenancy][IOTDB-5091] add space quota: device and timeseries num limit

RYH61:
add device and timeseries num limit


confignode/src/main/java/org/apache/iotdb/confignode/manager/ClusterQuotaManager.java
```diff
@@ -46,10 +49,16 @@ public class ClusterQuotaManager {
 
   private final IManager configManager;
   private final QuotaInfo quotaInfo;
+  private final Map<Integer, Integer> deviceNum;
+  private final Map<Integer, Integer> timeSeriesNum;
```
neuyilan, line 53: It better uses a concurrent map, it will be updated concurrently during a heartbeat（because the callback thread may run concurrently.）.

neuyilan (approved):
LGTM

PR #8279, opened 2022-12-01
[To rel/1.0] [IOTDB-4978] Fix deadLock caused by blocked operation in IntoOperator

liuminghui233:
see [IOTDB-4978](https://issues.apache.org/jira/browse/IOTDB-4978).


server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/AbstractIntoOperator.java
```diff
@@ -100,27 +138,59 @@ protected void insertMultiTabletsInternally(boolean needCheck) {
...
+    if (writeOperationFuture == null) {
+      return true;
     }
 
-    for (InsertTabletStatementGenerator generator : insertTabletStatementGenerators) {
-      generator.reset();
+    try {
+      if (!writeOperationFuture.isDone()) {
+        return false;
+      }
```
JackieTien97, line 169: if writeOperationFuture is not done, we should never call next method of into operator, so here if you just want to do a safety check, you should throw a IIlegalStateException.

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/AbstractIntoOperator.java
```diff
@@ -162,42 +232,62 @@ public OperatorContext getOperatorContext() {
...
+    ListenableFuture<?> childBlocked = child.isBlocked();
+    if (writeOperationDone() && childBlocked.isDone()) {
+      return NOT_BLOCKED;
+    } else if (!writeOperationDone() && childBlocked.isDone()) {
+      return writeOperationFuture;
+    } else if (writeOperationDone() && !childBlocked.isDone()) {
+      return childBlocked;
+    } else {
+      return successfulAsList(Arrays.asList(writeOperationFuture, childBlocked));
+    }
```
JackieTien97, line 254:
```suggestion
    ListenableFuture<?> childBlocked = child.isBlocked();
    boolean writeDone = writeOperationDone();
    if (writeDone && childBlocked.isDone()) {
      return NOT_BLOCKED;
    } else if (childBlocked.isDone()) {
      return writeOperationFuture;
    } else if (writeDone) {
      return childBlocked;
    } else {
      return successfulAsList(Arrays.asList(writeOperationFuture, childBlocked));
    }
```

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/IntoOperator.java
```diff
@@ -48,40 +49,70 @@ public IntoOperator(
...
+      if (insertMultiTabletsInternally(false)) {
+        return null;
+      }
+
+      finished = true;
       return constructResultTsBlock();
     }
   }
 
+  /** Return true if write task is submitted during processing */
```
JackieTien97, line 95: Is this java doc right?

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/AbstractIntoOperator.java
liuminghui233: fixed

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/AbstractIntoOperator.java
liuminghui233: fixed

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/IntoOperator.java
liuminghui233: fixed

PR #8285, opened 2022-12-01
[To rel/1.0][IOTDB-5101] Last flush time may replaced by previous files when recovering

HeimingZ:
image

Tested locally. We have verified that after this fix, the unsealed file's update for `LastFlushTime` will always be executed at the last step.


xingtanzjr:
Local UT Test from Haiming
image

PR #8307, opened 2022-12-02
[TO rel/1.0] upgrade audit log module

caozj1011:
1. added whether to enable audit logs during open session.
2. replace the original AUDIT_LOGGER.
3. when the Audit log level is set to LOGGER, the write operation is changed from DEBUG to INFO.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/client/DataNodeInternalClient.java
+++ b/server/src/main/java/org/apache/iotdb/db/client/DataNodeInternalClient.java
@@ -78,7 +78,8 @@ public class DataNodeInternalClient {
           session,
           sessionInfo.getUserName(),
           sessionInfo.getZoneId(),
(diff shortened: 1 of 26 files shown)


server/src/main/java/org/apache/iotdb/db/client/DataNodeInternalClient.java
```diff
@@ -78,7 +78,8 @@ public DataNodeInternalClient(SessionInfo sessionInfo) {
           session,
           sessionInfo.getUserName(),
           sessionInfo.getZoneId(),
-          IoTDBConstant.ClientVersion.V_0_13);
+          IoTDBConstant.ClientVersion.V_0_13,
+          true);
```
HTHou, line 82: Does `select into` need the audit logs?  @JackieTien97

PR #8328, opened 2022-12-05
[IOTDB-5090] Add npe check in DataNode

MiniSho:
When fail to start a Datanode, `SchemaRegionConsensusImpl` and `DataRegionConsensusImpl` will be null instead of being init , so need to add a check whether they are nullpoints when stop the datanode.


Beyyes (approved):
LGTM

PR #8338, opened 2022-12-05
[IOTDB-5122] forward exception when Ratis request failed

SzyWilliam:
See https://issues.apache.org/jira/browse/IOTDB-5122.


Beyyes (approved):
LGTM

We can add some test screenshots.

SzyWilliam:
@Beyyes A new unit test is added to verify the exception is indeed forwarded when createPeer failed. PTAL, thanks!

PR #8339, opened 2022-12-05
[IOTDB-5123] fix ratis disk control param mismatch

SzyWilliam:
See https://issues.apache.org/jira/browse/IOTDB-5123


confignode/src/main/java/org/apache/iotdb/confignode/manager/ConsensusManager.java
```diff
@@ -166,7 +166,7 @@ private void setConsensusLayer(ConfigNodeRegionStateMachine stateMachine) throws
                                       .setClientRetryMaxSleepTimeMs(
                                           CONF.getConfigNodeRatisMaxSleepTimeMs())
                                       .setTriggerSnapshotFileSize(
-                                          CONF.getConfigNodeRatisLogMaxMB() * 1024 * 1024)
```
OneSizeFitsQuorum, line 169: * 1024 * 1024？

OneSizeFitsQuorum (approved):
LGTM~
Please cherry-pick to rel/1.0

PR #8348, opened 2022-12-06
[To rel/1.0][IOTDB-5118] FileMetric opens too many file descriptors

HeimingZ:
FileUtils.listFiles in the commons io doesn't use Files.walk with try-with-resources contract, this error is same as [JDK-8250785](https://bugs.openjdk.org/browse/JDK-8250785)


SpriCoder (approved):
LGTM~

OneSizeFitsQuorum (approved):
LGTM

PR #8349, opened 2022-12-06
Set fast compaction as default compaction performer for cross space compaction and unseq space compaction

THUMarkLau:


choubenson:
Set performer to read_point and read_chunk in `CompactionSchedulerTest`, because there is already existed `CompactionSchedulerWithFastPerformerTest`.

PR #8359, opened 2022-12-07
update nodenames if necessary in ClientRpcServiceImpl

lancelly:
Previously we forget to update the checked measurements as node name  like \`aa\` should be stored as aa.

### DIFF
--- a/node-commons/src/main/java/org/apache/iotdb/commons/utils/PathUtils.java
+++ b/node-commons/src/main/java/org/apache/iotdb/commons/utils/PathUtils.java
@@ -26,6 +26,7 @@ import org.apache.iotdb.tsfile.exception.PathParseException;
 import org.apache.iotdb.tsfile.read.common.parser.PathNodesGenerator;
 import org.apache.iotdb.tsfile.read.common.parser.PathVisitor;
 
+import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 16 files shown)


node-commons/src/main/java/org/apache/iotdb/commons/utils/PathUtils.java
```diff
@@ -48,118 +49,131 @@ public static String[] splitPathToDetachedNodes(String path) throws IllegalPathE
...
+        && measurement.endsWith(TsFileConstant.BACK_QUOTE_STRING)) {
+      if (checkBackQuotes(measurement.substring(1, measurement.length() - 1))) {
+        return removeBackQuotesIfNecessary(measurement);
+      } else {
+        throw new IllegalPathException(measurement);
+      }
+    }
+    if (IoTDBConstant.reservedWords.contains(measurement.toUpperCase())
+        || isRealNumber(measurement)
+        || !TsFileConstant.NODE_NAME_PATTERN.matcher(measurement).matches()) {
```
JackieTien97, line 173: what's the difference with `!TsFileConstant.IDENTIFIER_PATTERN.matcher(unWrapped).matches()`

node-commons/src/main/java/org/apache/iotdb/commons/utils/PathUtils.java
```diff
@@ -48,118 +49,131 @@ public static String[] splitPathToDetachedNodes(String path) throws IllegalPathE
...
+    return res;
   }
 
   /**
-   * check whether measurement is legal according to syntax convention measurement can only be a
+   * check whether measurement is legal according to syntax convention. Measurement can only be a
    * single node name, use set to skip checking duplicated measurements
    */
-  public static void checkLegalSingleMeasurementsAndSkipDuplicate(
+  public static List<String> checkLegalSingleMeasurementsAndSkipDuplicate(
```
JackieTien97, line 82: we may need to change set to map, key is original string, value is formatted string.

JackieTien97 (changes requested):

PR #8368, opened 2022-12-07
[IOTDB-5135] Sync snapshot fails when sync file is empty.

Cpaulyz:
## Description

This problem is caused by the fact that when `ClusterSyncInfo` is empty, `SyncLogWriter` will not generate a tmpFile and `tmpFile.renameTo(snapshotFile)` will return false.

## Solution

Force the file to be initialised when taking a snapshot.


Beyyes (approved):
LGTM

PR #8385, opened 2022-12-08
[IOTDB-5153] add snapshot transfer IT & UT for RatisConsensus

SzyWilliam:
When removing a datanode, the data(logs/snapshot) will be transferred to a new datanode to maintain 3-copy HA.
Currently no IT/UT covers this scenario, so I added corresponding UT & IT.


OneSizeFitsQuorum (approved):
LGTM

PR #8387, opened 2022-12-08
[IOTDB-5119] fix bug that M4 throws exception when the window is empty

LeiRui:
Add the logic "`if (rowWindow.windowSize() > 0)`" and proceed transforming the window only when the window size is larger than 0.


lancelly (approved):
LGTM～

PR #8406, opened 2022-12-09
[IOTDB-5170] stop-confignode.bat, datanode will be closed at the same time

xiaoyekanren:
Before repair:
image

After repair:
image


Beyyes (approved):
LGTM

PR #8428, opened 2022-12-13
[IOTDB-5191] Rename RatisConfig.RatisConsensus

SzyWilliam:
see https://issues.apache.org/jira/browse/IOTDB-5191.

The config class name of RatisConsensus is the same with RatisConsensus Main class. It may lead to misunderstandings.


OneSizeFitsQuorum (approved):
LGTM

PR #8449, opened 2022-12-14
[IOTDB-5061] Add message in RatisRequestFailed

SzyWilliam:
Currently cause is not serialized and returned to confignode. Add detailed message so that the cause is contained in the error message.


OneSizeFitsQuorum (approved):
LGTM

PR #8461, opened 2022-12-14
[IOTDB-5209] Limit the read rate of compaction execution

THUMarkLau:
See [IOTDB-5209](https://issues.apache.org/jira/browse/IOTDB-5209).

This PR limits both the read rate and write rate of compaction execution.


choubenson (approved):
LGTM

PR #8466, opened 2022-12-14
[IOTDB-5222] Fix the load of metric related configuration

SpriCoder:
See: https://issues.apache.org/jira/browse/IOTDB-5222


OneSizeFitsQuorum (approved):
LGTM

PR #8458, opened 2022-12-14
[IOTDB-5174] Use filename format such as NodeID-Index rather than Endpoint-Index to track follower sync progress

OneSizeFitsQuorum:
see [jira5174](https://issues.apache.org/jira/browse/IOTDB-5174)

PR #8468, opened 2022-12-15
[To rel/1.0][IOTDB-5174] Use filename format such as NodeID-Index rather than End…

OneSizeFitsQuorum:
cherry-pick [pull/8458](https://github.com/apache/iotdb/pull/8458)


SpriCoder (approved):
LGTM~

PR #8470, opened 2022-12-15
[IOTDB-5153] Fix dead lock in confignode

SzyWilliam:
1. In register datanode, the trigger lock and udf lock are acquired sequentially.
2. In snapshot, this sequential order is not guaranteed. We have to guarantee the order to avoid dead lock.


OneSizeFitsQuorum (approved):
LGTM~We need to ensure correctness first and then performance

Beyyes (approved):
LGTM

PR #8469, opened 2022-12-15
[IOTDB-5223] Check data directories cross-disk under RatisConsensus

Cpaulyz:
## Description

Check data directories cross-disk under RatisConsensus

PR #8441, opened 2022-12-13
[IOTDB-5201] Extracting getConfig from the registerDataNode interface

Cpaulyz:
## Description

Extracting getConfig from the registerDataNode interface. 

https://apache-iotdb.feishu.cn/docs/doccnMgSBOgHqeT9gjs4RUrqTob

PR #8471, opened 2022-12-15
[To rel/1.0][IOTDB-5201][IOTDB-5223] Extract DataNode getConfig and add check for cross-disk directories 

Cpaulyz:
## Description

Cherry pick from #8441 and #8469


OneSizeFitsQuorum (approved):
LGTM

PR #8483, opened 2022-12-16
[IOTDB-4630] Implement observer pattern in LoadBalancing framework

MiniSho:
when the Cluster-LoadStatistics-Service in ConfigNode-leader finds that the heartbeat sample statistcs has changed (node status is switched), the observer needs to broadcast the change event to Sync, Template, Trigger and CQ modules

1C2D:
after shuting down one datanode
211127fbd12054e226b6cc80b48fd03


confignode/src/main/java/org/apache/iotdb/confignode/manager/load/LoadManager.java
```diff
@@ -185,20 +197,26 @@ private void updateLoadStatistics() {
...
         .getNodeCacheMap()
         .forEach(
             (nodeId, nodeCache) -> {
+              NodeStatistics preNodeStatistics = nodeCache.getPreviousStatistics().deepCopy();
               if (nodeCache.periodicUpdate()) {
                 // Update and record the changed NodeStatistics
-                differentNodeStatisticsMap.put(nodeId, nodeCache.getStatistics());
+                differentNodeStatisticsMap.put(
+                    nodeId, new NodeStatistics[] {nodeCache.getStatistics(), preNodeStatistics});
+                LOGGER.info("previous NodeStatistics: {}", preNodeStatistics);
```
Beyyes, line 213: If it will print many logs?

confignode/src/main/java/org/apache/iotdb/confignode/manager/load/LoadManager.java
```diff
@@ -185,20 +197,26 @@ private void updateLoadStatistics() {
     // Broadcast the RegionRouteMap if some LoadStatistics has changed
     boolean isNeedBroadcast = false;
 
-    // Update NodeStatistics
-    Map<Integer, NodeStatistics> differentNodeStatisticsMap = new ConcurrentHashMap<>();
+    // Update NodeStatistics:
+    // NodeStatistics[]:index 0 means the current NodeStatistics, index 1 means the previous
+    // NodeStatistics
+    Map<Integer, NodeStatistics[]> differentNodeStatisticsMap = new ConcurrentHashMap<>();
```
Beyyes, line 203: What about using Pair<NodeStatistics, NodeStatistics> to replace NodeStatistics[].

confignode/src/main/java/org/apache/iotdb/confignode/manager/load/LoadManager.java
MiniSho: fixed

confignode/src/main/java/org/apache/iotdb/confignode/manager/load/LoadManager.java
MiniSho: done.

confignode/src/main/java/org/apache/iotdb/confignode/manager/ClusterSchemaManager.java
```diff
@@ -713,6 +715,12 @@ public synchronized TSStatus dropSchemaTemplate(String templateName) {
     return getConsensusManager().write(new DropSchemaTemplatePlan(templateName)).getStatus();
   }
 
+  @Subscribe
+  public void handleNodeStatistics(
+      Map<Integer, Pair<NodeStatistics, NodeStatistics>> differentNodeStatisticsMap) {
+    // TODO
+  }
```
MarcosZyk, line 722: Please add  javadoc and param description to all such interfaces.

confignode/src/main/java/org/apache/iotdb/confignode/manager/load/LoadManager.java
```diff
@@ -185,20 +198,26 @@ private void updateLoadStatistics() {
     // Broadcast the RegionRouteMap if some LoadStatistics has changed
     boolean isNeedBroadcast = false;
 
-    // Update NodeStatistics
-    Map<Integer, NodeStatistics> differentNodeStatisticsMap = new ConcurrentHashMap<>();
+    // Update NodeStatistics:
+    // NodeStatistics[]:index 0 means the current NodeStatistics, index 1 means the previous
+    // NodeStatistics
+    Map<Integer, Pair<NodeStatistics, NodeStatistics>> differentNodeStatisticsMap =
+        new ConcurrentHashMap<>();
```
MarcosZyk, line 205: Maybe encapsulate the node state difference as a Class is a better choice, and some common logic can be implemented inside such class in future work.

confignode/src/main/java/org/apache/iotdb/confignode/manager/load/LoadManager.java
```diff
@@ -185,20 +198,26 @@ private void updateLoadStatistics() {
...
                 // Update and record the changed NodeStatistics
-                differentNodeStatisticsMap.put(nodeId, nodeCache.getStatistics());
+                differentNodeStatisticsMap.put(
+                    nodeId, new Pair<>(nodeCache.getStatistics(), preNodeStatistics));
               }
             });
     if (!differentNodeStatisticsMap.isEmpty()) {
       isNeedBroadcast = true;
       recordNodeStatistics(differentNodeStatisticsMap);
+      eventBus.post(differentNodeStatisticsMap);
```
MarcosZyk, line 220: Is there any concurrency control on some other methods that may change the node's state, such as register DataNode, when post the node state difference.

MarcosZyk, line 220: If all the listeners want to interact with the target nodes, is there any way to allocate info and deliver once, just like DataNode start-up check.

MarcosZyk (commented):
PTAL.

Beyyes (approved):
LGTM

PR #8489, opened 2022-12-16
[IOTDB-5176] Change ratis snapshot dir with datadir 

SzyWilliam:
Currently, when user assigns a customized data_dir in iotdb-datanode.properties, the ratis snapshot dir is not updated with this customized value, so that there will be inconsistent snapshot dirs.
In this pr I propose to update ratis snapshot dir togather with data_dir.


OneSizeFitsQuorum (approved):
LGTM

neuyilan:
Why not cherry-pick to rel/1.0?

PR #8494, opened 2022-12-16
[IOTDB-5173][To rel/1.0] Add statistic of flush point

SpriCoder:
Add the metric of flush point in important level with internal reporter support


OneSizeFitsQuorum (approved):
LGTM

PR #8496, opened 2022-12-17
[IOTDB-5173][To rel/0.13] Make metric related doc consistent with implementation

SpriCoder:
Make metric related doc consistent with implementation


OneSizeFitsQuorum (approved):
LGTM

PR #8506, opened 2022-12-19
[IOTDB-5233] Correct the message of unsupported thrift interface of schema template

MarcosZyk:
## Description

see IOTDB-5233


Cpaulyz (approved):
LGTM～

PR #8508, opened 2022-12-19
[To rel/1.0][IOTDB-5233] Correct the message of unsupported thrift interface of schema template

MarcosZyk:
## Description

see [IOTDB-5233](https://issues.apache.org/jira/browse/IOTDB-5233)


Cpaulyz (approved):
LGTM～

PR #8514, opened 2022-12-19
Simplify RegionLeaderBalancingIT

CRZbulabula:
Remove SchemaRegionGroup(which uses Ratis consensus protocol) check


Beyyes (approved):
LGTM

PR #8523, opened 2022-12-19
[IOTDB-5231]Add the handle of exception when metric reporter failed to start

SpriCoder:
1. Add the handle of exception when metric reporter fails
2. Specify the code of start/stop process


OneSizeFitsQuorum (approved):
LGTM

PR #8545, opened 2022-12-20
[IOTDB-5232] Add schema memory metric

MarcosZyk:
## Description

Add the following metric
1. schema region memory usage
2. schema region memory remaining


SpriCoder (approved):
LGTM~

PR #8547, opened 2022-12-20
Start ConfigNode and DataNode successfully when cluster_name is missed

Beyyes:
I have test the upgrate from 1.0 to master.

system file of 1.0

image

image


CRZbulabula (approved):
LGTM～

PR #8551, opened 2022-12-21
[IOTDB-5240] Add template id to schema region snapshot

MarcosZyk:
### Modification
1. Refactor the constructor of Template to decuple it with CreateSchemaTemplateStatement
2. Add template id to schema region memory mode snapshot
3. Add template id to schema region schema_file mode snapshot
    1. refine the activation process
    2. fix the snapshot load process
    3. fix the show timeseries without templateMap
    4. add template id to schema file


Cpaulyz (approved):
LGTM

PR #8555, opened 2022-12-21
[IOTDB-4804] Cpp Client SDK bug-fix

LoopJump:
[IOTDB-4804] Cpp Client SDK bugfix
### Fix two bugs of Cpp Client SDK in IOTDB 1.0.
Bug1: IOTDB1.0 may return to client the status `REDIRECTION_RECOMMEND`, this should be treated as SUCCESS.
Bug2: in some 32bit arch CPU such as armv7l, when doing type casting, it may raises BUS ERROR if the address to be casted is not 4Byte-alined.

### Add a compile script to build sdk library and demo code.
Add compile_iotdb_session_lib.sh. Enable -DARCH32 if you want to compile for 32bit arch CPU.
Add demo.cpp, demonstrate how to use this sdk library.

### Fix sessionIT.
Bug1: use `select s1,s2,s3` instead of `select *` to make the output stable.
Bug2: IOTDB1.0 limits the response packet size in FI, which make a testcase fail in sessionIT. We change the count of rows to test. Beside we make the output pretty.

### Fix example code.
fix SessionExample.cpp


client-cpp/compile_iotdb_session_lib.sh
```diff
@@ -0,0 +1,57 @@
...
+$thrift_bin --gen cpp -o src/main/ -I $common_thrift_dir $client_thrift_dir/client.thrift
+
+cd src/main
+LD_LIBRARY_PATH=$thrift_lib g++ -Igen-cpp $CXX_FLAGS $CXX_SO_FLAGS $CXX_FLAGS_DEFINED -o libiotdb_session.so $source_file
+
+# back to iotdb/client-cpp
+cd ../../
+mkdir iotdb_cpp_client_sdk
+mkdir iotdb_cpp_client_sdk/include iotdb_cpp_client_sdk/lib
+cp $thrift_lib/libthrift-0.14.0.so  iotdb_cpp_client_sdk/lib/
```
HTHou, line 49: We are using thrift 0.14.1 in IoTDB,  why 0.14.0 here?

HTHou (approved):

PR #8579, opened 2022-12-22
[IOTDB-5266] Refine the code of cross selector and fix the bug when using FileTimeIndex 

xingtanzjr:
# Description
1. refine the code of cross selector
2. fix the bug that some seq files may be lost when selecting cross compaction task if the seq file using FileTimeIndex

See this issue for details https://issues.apache.org/jira/browse/IOTDB-5266


server/src/main/java/org/apache/iotdb/db/engine/compaction/cross/rewrite/CrossSpaceCompactionCandidate.java
```diff
@@ -0,0 +1,245 @@
...
+    protected boolean selected;
+    protected boolean isValidCandidate;
+    private Map<String, DeviceInfo> deviceInfoMap;
+
+    protected TsFileResourceCandidate(TsFileResource tsFileResource) {
+      this.resource = tsFileResource;
+      this.selected = false;
+      // although we do the judgement here, the task should be validated before executing because
+      // the status of file may be changed after the task is submitted to queue
+      this.isValidCandidate = tsFileResource.isClosed() && tsFileResource.getTsFile().exists();
```
THUMarkLau, line 188: What if this tsfile resource is compaction candidate or compacting?

THUMarkLau (approved):

PR #8599, opened 2022-12-23
Turn down the max capacity of tsblock from scanOperator

Alima777:
In default, pageSizeInByte is 64K, but maxTsBlockSizeInBytes is 1M, which brings much memory pressure since we have reserver memory space for it in advance.
Therefore, I suggest to turn down this config to 128K, which can contain 2 pages at most.

For more detail, refer to https://apache-iotdb.feishu.cn/docx/O5XCdvEHPoDdtNxFHlXcV6xlnoh

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/mpp/execution/memory/MemoryPool.java
+++ b/server/src/main/java/org/apache/iotdb/db/mpp/execution/memory/MemoryPool.java
@@ -25,10 +25,11 @@ import org.apache.iotdb.tsfile.utils.Pair;
 import com.google.common.util.concurrent.AbstractFuture;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
-import javax.annotation.Nullable;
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/apache/iotdb/db/mpp/execution/memory/MemoryPool.java
```diff
@@ -154,6 +154,14 @@ public Pair<ListenableFuture<Void>, Boolean> reserve(
         bytesToReserve > 0L && bytesToReserve <= maxBytesPerFragmentInstance,
         "bytes should be greater than zero while less than or equal to max bytes per fragment instance: %d",
         bytesToReserve);
+    if (bytesToReserve > maxBytesCanReserve) {
+      LOGGER.error(
```
lancelly, line 158: warn level log

Alima777: Why? It's an error which means it will never make it that reserve enough memory, and supposed to throw an exception.

PR #8614, opened 2022-12-26
[IOTDB-5160] [Metric] Fix the file count of datanode become negative number

THUMarkLau:
See [IOTDB-5160](https://issues.apache.org/jira/browse/IOTDB-5160), [IOTDB-5279](https://issues.apache.org/jira/browse/IOTDB-5279).

The reason for this bug is the program fails to traverse the whole file list in recovering. As a result, the number and size of the files in metrics are smaller than the actual value.


SpriCoder (approved):
LGTM~ Please cherry-pick to rel/1.0

PR #8615, opened 2022-12-26
[IOTDB-5286]Fix target files should not be select to participate in other compaction task

choubenson:
**Description**
When the system is restarted, a compaction error is reported: the source file is deleted and the target file is also deleted, which will set allowCompaction to false.

**Reason**
Target files are selected to participate in other compaction tasks before the end of the current task, causing unexpected errors.

**Solution**
Set target file to COMPACTING until the end of task, so that it can be selected to participate in other task after this task finishes.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
@@ -179,6 +179,14 @@ public class TsFileResource {
     this.timeIndex = CONFIG.getTimeIndexLevel().getTimeIndex();
   }
 
+  /** Used for compaction to create target files. */
+  public TsFileResource(File file, TsFileResourceStatus status) {
+    this.file = file;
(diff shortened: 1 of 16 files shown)


THUMarkLau (approved):

server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResource.java
```diff
@@ -179,6 +179,14 @@ public TsFileResource(File file) {
     this.timeIndex = CONFIG.getTimeIndexLevel().getTimeIndex();
   }
 
+  /** Used for compaction to create target files. */
+  public TsFileResource(File file, TsFileResourceStatus status) {
+    this.file = file;
```
xingtanzjr, line 184: using `this(file)` to reduce duplicated codes

choubenson: Resolved.

PR #8631, opened 2022-12-27
[To rel/1.0][IOTDB-5278] fix connection problem of dbeaver and supports more functions

ycycse:
The dbeaver can connect normally after the PR.
And the dbeaver can show the device and column now.
image

Can be accessed through:
[iotdb-jdbc-1.0.1-SNAPSHOT-jar-with-dependencies.zip](https://github.com/apache/iotdb/files/10306840/iotdb-jdbc-1.0.1-SNAPSHOT-jar-with-dependencies.zip)

### DIFF
--- a/jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBDatabaseMetadata.java
+++ b/jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBDatabaseMetadata.java
@@ -1510,9 +1510,9 @@ public class IoTDBDatabaseMetadata implements DatabaseMetaData {
             TSDataType.INT32,
             TSDataType.TEXT);
 
-    String sg = "";
-    if (catalog != null) sg = catalog;
-    else if (schema != null) sg = schema;
+    String database = "";
+    if (catalog != null) database = catalog;
+    else if (schema != null) database = schema;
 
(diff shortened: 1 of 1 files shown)


jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBDatabaseMetadata.java
```diff
@@ -1674,6 +1509,11 @@ public ResultSet getPrimaryKeys(String catalog, String schema, String table) thr
             TSDataType.TEXT,
             TSDataType.INT32,
             TSDataType.TEXT);
+
+    String sg = "";
```
JackieTien97, line 1513: never call it `sg`, we've already renamed it to `db` in new version.

JackieTien97: check it in the following code.

JackieTien97 (approved):

jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBDatabaseMetadata.java
ycycse: the change has done.

PR #8640, opened 2022-12-28
[To rel/0.13][IOTDB-5307] Failed to get TsFile input of file: NoSuchFileException

lpf4254302:
## Description
因TsFileResource状态修改存在并发问题，导致ttl删除文件和合并冲突，最终导致合并定时任务卡死，原因是未捕获到npe异常

当前PR实现内容：
1、捕获文件不存在异常
2、捕获跨区合并Exception异常

未解决内容：
修改TsFileResource状态时需要增加并发控制

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/compaction/cross/rewrite/RewriteCrossSpaceCompactionSelector.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/compaction/cross/rewrite/RewriteCrossSpaceCompactionSelector.java
@@ -30,13 +30,11 @@ import org.apache.iotdb.db.engine.compaction.inner.utils.InnerSpaceCompactionUti
 import org.apache.iotdb.db.engine.compaction.task.AbstractCompactionTask;
 import org.apache.iotdb.db.engine.storagegroup.TsFileManager;
 import org.apache.iotdb.db.engine.storagegroup.TsFileResource;
-import org.apache.iotdb.db.exception.MergeException;
 import org.apache.iotdb.db.rescon.SystemInfo;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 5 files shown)


THUMarkLau (approved):

choubenson (approved):

server/src/main/java/org/apache/iotdb/db/engine/compaction/cross/rewrite/RewriteCrossSpaceCompactionSelector.java
```diff
@@ -137,7 +135,7 @@ public void selectAndSubmit() {
             mergeResource.getUnseqFiles().size());
       }
 
-    } catch (MergeException | IOException | InterruptedException e) {
+    } catch (Exception e) {
```
xingtanzjr, line 138: It is unnecessary to merge these 3 types of exceptions into `Exception`

tsfile/src/main/java/org/apache/iotdb/tsfile/read/TsFileSequenceReader.java
```diff
@@ -134,6 +134,9 @@ public TsFileSequenceReader(String file, boolean loadMetadataSize) throws IOExce
     }
     this.file = file;
     tsFileInput = FSFactoryProducer.getFileInputFactory().getTsFileInput(file);
```
xingtanzjr, line 136: Current change is OK to resolve the issue. 
But to make it more safe, it's better to change the method `getTsFileInput(file)` to avoid making `null` as a possible returned value.
We can let the method `getTsFileInput` throw `IOException` if the file doesn't exist

xingtanzjr (approved):

PR #8646, opened 2022-12-28
[IOTDB-5285] TimePartition may be error when restarting with different time partition configuration

HeimingZ:
If time partition interval is changed, error log is like below:
error

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/conf/IoTDBStartCheck.java
+++ b/server/src/main/java/org/apache/iotdb/db/conf/IoTDBStartCheck.java
@@ -299,11 +299,11 @@ public class IoTDBStartCheck {
     // check whether upgrading from <=v0.9
     if (!properties.containsKey(IOTDB_VERSION_STRING)) {
       logger.error(
-          "DO NOT UPGRADE IoTDB from v0.9 or lower version to v0.12!"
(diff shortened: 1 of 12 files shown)


server/src/main/java/org/apache/iotdb/db/conf/IoTDBStartCheck.java
```diff
@@ -57,95 +55,91 @@ public class IoTDBStartCheck {
...
-  private static final String mppDataExchangePort = String.valueOf(config.getMppDataExchangePort());
-
   private static final String SCHEMA_REGION_CONSENSUS_PORT = "dn_schema_region_consensus_port";
-  private static final String schemaRegionConsensusPort =
-      String.valueOf(config.getSchemaRegionConsensusPort());
-
   private static final String DATA_REGION_CONSENSUS_PORT = "dn_data_region_consensus_port";
-  private static final String dataRegionConsensusPort =
-      String.valueOf(config.getDataRegionConsensusPort());
+  private static final Map<String, Supplier<String>> variableParamValueTable = new HashMap<>();;
```
HTHou, line 117:
```suggestion
  private static final Map<String, Supplier<String>> variableParamValueTable = new HashMap<>();
```

xingtanzjr (approved):

PR #8650, opened 2022-12-28
[IOTDB-5308]  Limit the output tsblock's size of each operator[Part1]

Alima777:
Currently, **almost all operators don't limit the size of its output tsblock, which 1. have side effect in the CPU performance,
2. bring much pressure to memory control. Therefore I suggest we limit the output tsblock's size.**

image

According the paper that proposes vertorized processing[1], **when the vector can't fit in the L1 and L2 cache of the CPU, the performance will be degrading.** In the above figure, the horizontal ordinate represents the number of rows in vector array, with one tuple of 40 bytes.

The size of L1 is about 16 ~ 64K or more typically, while the size of L2 is 256K or more. Currently, our page size is 64K, **turn tsblock's size to 128K may get a balance bewteen interpration cost and performance loss caused by memory bottleneck.**[2] When there are 1 ~ 10 value columns in one tuple, the vector size will be 2000 ~ 10000 if total is 128K, which shows good performance.

[1] https://www.cidrdb.org/cidr2005/papers/P19.pdf
[2] https://github.com/apache/iotdb/pull/8599


server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/join/RowBasedTimeJoinOperator.java
```diff
@@ -198,14 +207,35 @@ public TsBlock next() {
...
+    retainedTsBlock = tsBlockBuilder.build();
+    LOGGER.info("Current tsBlock size is : {}", retainedTsBlock.getRetainedSizeInBytes());
+    return getResultTsBlock();
+  }
+
+  private TsBlock getResultTsBlock() {
+    long maxSizePerTsBlock = TSFileDescriptor.getInstance().getConfig().getMaxTsBlockSizeInBytes();
+    long maxLength = maxSizePerTsBlock / (1 + outputColumnCount);
+    TsBlock resultTsBlock;
+    if (retainedTsBlock.getPositionCount() - retainedStartOffset < maxLength) {
```
JackieTien97, line 219: maxLength is size in bytes, but `retainedTsBlock.getPositionCount() - retainedStartOffset` is line number, why comparing these two?

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/process/join/TimeJoinOperator.java
```diff
@@ -197,14 +206,35 @@ public TsBlock next() {
...
+    retainedTsBlock = tsBlockBuilder.build();
+    LOGGER.debug("Current tsBlock size is : {}", retainedTsBlock.getRetainedSizeInBytes());
+    return getResultTsBlock();
+  }
+
+  private TsBlock getResultTsBlock() {
+    long maxSizePerTsBlock = TSFileDescriptor.getInstance().getConfig().getMaxTsBlockSizeInBytes();
+    long maxLength = maxSizePerTsBlock / (1 + outputColumnCount);
+    TsBlock resultTsBlock;
+    if (retainedTsBlock.getPositionCount() - startOffset < maxLength) {
```
JackieTien97, line 218: same as above

server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/source/AlignedSeriesScanOperator.java
```diff
@@ -77,13 +84,35 @@ public OperatorContext getOperatorContext() {
...
     builder.reset();
-    return block;
+    return getResultTsBlock();
+  }
+
+  private TsBlock getResultTsBlock() {
+    long maxSizePerTsBlock = TSFileDescriptor.getInstance().getConfig().getMaxTsBlockSizeInBytes();
+    long length = maxSizePerTsBlock / (1 + seriesScanUtil.getAllSensorsSize());
+    TsBlock resultTsBlock;
+    if (retainedTsBlock.getPositionCount() - startOffset < length) {
```
JackieTien97, line 99: same as above

JackieTien97 (changes requested):
What about `SeriesScanOperator`?

PR #8651, opened 2022-12-29
[To rel/1.0] Change read dispatch from async to sync

JackieTien97:
Even for query, we still blocked to wait for dispatch result, so there is no need to use sperated thread pool to do dispatching.


OneSizeFitsQuorum (commented):
The thread executing this function is ClientRPC-Processor, and the maximum number of ClientRPC-Processor threads in threadPool is 65535. However, The coreSize of executor threadPool is 20.

Prior to this change, if the concurrency was high enough,  a maximum of 20 RPCs can be sent in parallel, and the remaining requests are blocked, and the higher the network latency, the more obvious this problem is. However, with this change, you can only send RPCS synchronously. If many instances are involved, it may take a long time to send RPCS multiple times.

For these reasons, I recommend using Thrift AsyncClient. It can both send multiple RPCs in parallel within the current thread without introducing any thread pool, and it can lay the foundation for more refined thread resource control in the future

Although cumbersome, the significant performance gains on this critical path are well worth it.

PR #8654, opened 2022-12-29
[IOTDB-5312] Consolidate ClientManagers in Datanodes for unified management

OneSizeFitsQuorum:
see [jira5312](https://issues.apache.org/jira/browse/IOTDB-5312)


node-commons/src/main/java/org/apache/iotdb/commons/client/async/AsyncConfigNodeHeartbeatServiceClient.java
```diff
@@ -53,41 +55,44 @@ public AsyncConfigNodeHeartbeatServiceClient(
...
   }
 
-  /**
-   * This method will be automatically called by the thrift selector thread, and we'll just simulate
-   * the behavior in our test
-   */
   @Override
   public void onError(Exception e) {
     super.onError(e);
+    ThriftClient.resolveException(e, this);
```
CRZbulabula, line 67: It's better not to log any exception when cluster heartbeat fails down. Otherwise the user might think it's the ConfigNode-leader who has errors.

node-commons/src/main/java/org/apache/iotdb/commons/client/async/AsyncDataNodeHeartbeatServiceClient.java
```diff
@@ -53,41 +55,44 @@ public AsyncDataNodeHeartbeatServiceClient(
...
   }
 
-  /**
-   * This method will be automatically called by the thrift selector thread, and we'll just simulate
-   * the behavior in our test
-   */
   @Override
   public void onError(Exception e) {
     super.onError(e);
+    ThriftClient.resolveException(e, this);
```
CRZbulabula, line 67: It's better not to log any exception when cluster heartbeat fails down. Otherwise the user might think it's the ConfigNode-leader who has errors.

node-commons/src/main/java/org/apache/iotdb/commons/conf/CommonConfig.java
```diff
@@ -99,13 +99,14 @@ public class CommonConfig {
...
-          : 1;
+  private int selectorNumOfClientManager = 1;
 
   /** whether to use thrift compression. */
-  private boolean isCnRpcThriftCompressionEnabled = false;
+  private boolean isRpcThriftCompressionEnabled = false;
+
+  private int maxTotalClientForEachNode = 300;
+
+  private int maxIdleClientForEachNode = 200;
```
CRZbulabula, line 109: These updated parameters should be also removed in ConfigNodeConfig, IoTDBConfig, iotdb-confignode.properties and iotdb-datanode.properties. And should be added in iotdb-commons.properties.

CRZbulabula (changes requested):
PTAL~

node-commons/src/main/java/org/apache/iotdb/commons/client/async/AsyncConfigNodeHeartbeatServiceClient.java
OneSizeFitsQuorum: So far this is just debug log, what do you think?

node-commons/src/main/java/org/apache/iotdb/commons/client/async/AsyncDataNodeHeartbeatServiceClient.java
OneSizeFitsQuorum: So far this is just debug log, what do you think?

node-commons/src/main/java/org/apache/iotdb/commons/conf/CommonConfig.java
OneSizeFitsQuorum: Current thrift related parameters are placed in a separate configuration file for confignode/datanode. This ensures that confignode and datanode can use different thrift parameters in a distribution package.

To keep up with the current status, I load different parameters for the same variable in CommonDescriptor, which is ugly but ensures that any process can set the parameters correctly to initialize the clientPool with the correct configuration.

node-commons/src/main/java/org/apache/iotdb/commons/client/async/AsyncConfigNodeHeartbeatServiceClient.java
CRZbulabula: Debug log is suitable for this~

node-commons/src/main/java/org/apache/iotdb/commons/client/async/AsyncDataNodeHeartbeatServiceClient.java
CRZbulabula: Debug log is suitable for this~

CRZbulabula (approved):
LGTM~

PR #8666, opened 2022-12-29
[IOTDB-5320] Add executeAggregationQuery interface in session API (java)

liuminghui233:
Add an efficient interface for aggregation query that can save SQL parsing cost.

- Support specified query time range: The specified query time range is a left-closed right-open interval, including the start time but not the end time.
- Support GROUP BY TIME.

```java
SessionDataSet executeAggregationQuery(List<String> paths, List<Aggregation> aggregations);

SessionDataSet executeAggregationQuery(
    List<String> paths, List<Aggregation> aggregations, long startTime, long endTime);

SessionDataSet executeAggregationQuery(
    List<String> paths,
    List<Aggregation> aggregations,
    long startTime,
    long endTime,
    long interval);

SessionDataSet executeAggregationQuery(
    List<String> paths,
    List<Aggregation> aggregations,
    long startTime,
    long endTime,
    long interval,
    long slidingStep);
```


thrift/src/main/thrift/client.thrift
```diff
@@ -545,6 +560,8 @@ service IClientRPCService {
 
   TSExecuteStatementResp executeLastDataQuery(1:TSLastDataQueryReq req);
 
+  TSExecuteStatementResp executeAggregationQuery(1:TSAggregationQueryReq req);
+
```
JackieTien97, line 564: we never need this, because old version session won't see this new interface.

isession/src/main/java/org/apache/iotdb/isession/util/Aggregation.java
```diff
@@ -0,0 +1,38 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iotdb.isession.util;
+
+public enum Aggregation {
```
JackieTien97, line 22: Make it a thrift enum class, and delete `AggregationType` in server and instead use the new enum, we should unify the usage of this enum.
In this way, all other languages can use `Aggregation` enum.

thrift/src/main/thrift/client.thrift
```diff
@@ -331,11 +331,24 @@ struct TSLastDataQueryReq {
...
+  6: optional bool enableRedirectQuery
+  7: optional bool jdbcQuery
   8: optional i64 timeout
 }
 
+struct TSAggregationQueryReq {
+  1: required i64 sessionId
+  2: required i64 statementId
+  3: required list<string> paths
+  4: required list<string> aggregations
```
JackieTien97, line 343: use thrift-defined enum instead of string

JackieTien97 (changes requested):

isession/src/main/java/org/apache/iotdb/isession/util/Aggregation.java
liuminghui233: fixed

thrift/src/main/thrift/client.thrift
liuminghui233: fixed

JackieTien97 (approved):

JackieTien97 (approved):

integration-test/src/test/java/org/apache/iotdb/session/it/IoTDBSessionQueryIT.java
```diff
@@ -0,0 +1,327 @@
...
+import static org.apache.iotdb.db.it.utils.TestUtils.assertResultSetEqual;
+import static org.junit.Assert.fail;
+
+@RunWith(IoTDBTestRunner.class)
+@Category({LocalStandaloneIT.class, ClusterIT.class})
+public class IoTDBSessionQueryIT {
+
+  @BeforeClass
+  public static void setUp() throws Exception {
+    System.setProperty(IoTDBConstant.IOTDB_CONF, "src/test/resources/");
```
ericpai, line 54: 后续 PR  SessionIT 中这个 setProperty 可以省略了。

PR #8683, opened 2022-12-30
[IOTDB-5245] Fix bug in last query

JackieTien97:
While there exist multi data regions for one series slot, we will add LastMergeSortNode to merge all last query result from different data regions, it will select max time for the same time-series.
However, if so, we still need to add default order by parameter for each LastQueryNode in each DataRegion and sort its children by timeseries asc, because `LastQueryMergeOperator` will assume that all its children returning TsBlock has already been sorted.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/distribution/SourceRewriter.java
+++ b/server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/distribution/SourceRewriter.java
@@ -546,18 +546,19 @@ public class SourceRewriter extends SimplePlanNodeRewriter<DistributionPlanConte
     if (root instanceof LastQueryNode) {
       LastQueryNode lastQueryNode = (LastQueryNode) root;
       lastQueryNode.setMergeOrderParameter(orderByParameter);
+      // sort children node
       lastQueryNode.setChildren(
           lastQueryNode.getChildren().stream()
               .sorted(
                   Comparator.comparing(
                       child -> {
-                        String device = "";
+                        String fullPath = "";
                         if (child instanceof LastQueryScanNode) {
-                          device = ((LastQueryScanNode) child).getSeriesPath().getDevice();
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/distribution/SourceRewriter.java
```diff
@@ -525,13 +528,46 @@ public List<PlanNode> visitLastQuery(LastQueryNode node, DistributionPlanContext
...
+      }
       root.getChildren().forEach(newRoot::addChild);
       return Collections.singletonList(newRoot);
     } else {
       return Collections.singletonList(root);
     }
   }
 
+  private void addSortForEachLastQueryNode(PlanNode root, OrderByParameter orderByParameter) {
+    if (root instanceof LastQueryNode) {
```
lancelly, line 546: It's possible that the children of LastQueryNode are still LastQueryNodes and these children won't be handled by this method.
截屏2023-01-01 17 30 24

server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/distribution/SourceRewriter.java
```diff
@@ -525,13 +528,46 @@ public List<PlanNode> visitLastQuery(LastQueryNode node, DistributionPlanContext
...
+                          fullPath = ((AlignedLastQueryScanNode) child).getSeriesPath().getDevice();
+                        }
+                        return fullPath;
+                      }))
+              .collect(Collectors.toList()));
+    } else {
+      for (PlanNode child : root.getChildren()) {
+        addSortForEachLastQueryNode(child, orderByParameter);
+      }
+    }
```
lancelly, line 568:
```suggestion
    } 
      for (PlanNode child : root.getChildren()) {
        addSortForEachLastQueryNode(child, orderByParameter);
      }
```

lancelly (approved):
LGTM～

PR #8684, opened 2022-12-30
Fix typo in query metric

liuminghui233:


SpriCoder (changes requested):
Maybe Update Doc at the same time

PR #8708, opened 2023-01-03
[IOTDB-5326] Add cluster_name parameter

CRZbulabula:
Add cluster_name parameter for which the user can distinguish different clusters. The cluster_name is configurable as shown below:

image

I also enhance ClusterIT to test when the cluster_name are inconsist. And the user docs are updated by the way.

### DIFF
--- a/confignode/src/assembly/resources/conf/iotdb-confignode.properties
+++ b/confignode/src/assembly/resources/conf/iotdb-confignode.properties
@@ -18,6 +18,14 @@
 #
 
 ####################
+### Cluster Configuration
+####################
+
+# Used for indicate cluster name and distinguish different cluster.
+# Datatype: string
+cluster_name=defaultCluster
+
+####################
 ### Config Node RPC Configuration
 ####################
 
--- a/confignode/src/assembly/resources/conf/iotdb-confignode.properties
+++ b/confignode/src/assembly/resources/conf/iotdb-confignode.properties
@@ -18,14 +18,6 @@
 #
 
 ####################
-### Cluster Configuration
-####################
-
-# Used for indicate cluster name and distinguish different cluster.
-# Datatype: string
-cluster_name=defaultCluster
-
-####################
 ### Config Node RPC Configuration
 ####################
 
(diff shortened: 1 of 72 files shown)


confignode/src/assembly/resources/conf/iotdb-confignode.properties
```diff
@@ -17,6 +17,14 @@
 # under the License.
 #
 
+####################
+### Cluster Configuration
+####################
+
+# Used for indicate cluster name and distinguish different cluster.
+# Datatype: string
+cluster_name=defaultCluster
```
ericpai, line 26: All the keys in iotdb-confignode.properties have `cn_` prefix. It's better to change `cluster_name` to `cn_cluster_name`, or just move it to iotdb-commom.properties ?

server/src/assembly/resources/conf/iotdb-datanode.properties
```diff
@@ -17,6 +17,14 @@
 # under the License.
 #
 
+####################
+### Cluster Configuration
+####################
+
+# Used for indicate cluster name and distinguish different cluster.
+# Datatype: string
+cluster_name=defaultCluster
```
ericpai, line 26: Both iotdb-datanode.properties and iotdb-confignode.properties have this parameter, which violates our [design](https://apache-iotdb.feishu.cn/docx/doxcnjzB2xd9sScIlkNYJUJA1zb) about the configuration files. Move it to iotdb-common.properties instead.

ericpai (changes requested):

server/src/assembly/resources/conf/iotdb-datanode.properties
CRZbulabula: Fixed~

confignode/src/assembly/resources/conf/iotdb-confignode.properties
CRZbulabula: I've moved it into the iotdb-common.properties~

ericpai (approved):

Beyyes (approved):
LGTM

PR #8713, opened 2023-01-03
Fix cluster_name definition conflict

CRZbulabula:
Changes:

1. Remove `CLUSTER_NAME` definition in IoTDBStartupCheck.
2. Definition `CLUSTER_NAME = "cluster_name"` is in IoTDBConstant, which will be used for reading configuration files.
3. Definition `COLUMN_CLUSTER_NAME = "ClusterName"` is in ColumnHeaderConstant, which will be used for indicating a row result of SQL: `SHOW VARIABLES`.


Beyyes (approved):
LGTM

PR #8715, opened 2023-01-03
Rename SQL show cluster parameters to show variables

CRZbulabula:
Rename and update related user docs. Test result:

image


Beyyes (approved):
LGTM

PR #8721, opened 2023-01-03
[IOTDB-5342][IOTDB-5359][IOTDB-5360] Some optimization in AbstractTreeVisitor

Cpaulyz:
## Description

1. [IOTDB-5342] Separate filtering and business logic in AbstractTreeVisitor.
```java
  /**
   * Internal-match means the node matches an internal node name of the given path pattern. root.sg
   * internal match root.sg.**(pattern). This method should be implemented according to concrete
   * tasks.
   *
   * <p>Return whether the subtree of given node should be processed. If return true, the traversing
   * process will keep traversing the subtree. If return false, the traversing process will skip the
   * subtree of given node.
   */
  protected abstract boolean shouldVisitSubtreeOfInternalMatchedNode(N node);

  /**
   * Full-match means the node matches the last node name of the given path pattern. root.sg.d full
   * match root.sg.**(pattern) This method should be implemented according to concrete tasks.
   *
   * <p>Return whether the subtree of given node should be processed. If return true, the traversing
   * process will keep traversing the subtree. If return false, the traversing process will skip the
   * subtree of given node.
   */
  protected abstract boolean shouldVisitSubtreeOfFullMatchedNode(N node);

  /** Only accepted nodes will be considered for hasNext() and next() */
  protected abstract boolean acceptInternalMatchedNode(N node);

  /** Only accepted nodes will be considered for hasNext() and next() */
  protected abstract boolean acceptFullMatchedNode(N node);
```

2. [IOTDB-5360] Injecting logic for releasing node resources in the AbstractTreeVistor framework. In ClusterSchemaTreeTest, a Mock test class `MockSchemaTreeMeasurementVisitor` is used to simulate the pin and unpin process to test this releasing logic.

See https://apache-iotdb.feishu.cn/docx/Vs88d6ZmkoNimVxQkS0cAh4Nntx#Y0w2dYIEiooeOSxC1hrc7xdlntb for more detail

3. [IOTDB-5359] Use decorator pattern to refactor SchemaTreeVisitorWithLimitOffset


node-commons/src/main/java/org/apache/iotdb/commons/schema/tree/AbstractTreeVisitor.java
```diff
@@ -189,17 +195,26 @@ protected void getNext() throws Exception {
...
 
       if (currentStateMatchInfo.hasFinalState()) {
-        shouldVisitSubtree = processFullMatchedNode(node);
+        if (acceptNode(nextTempNode)) {
+          nextMatchedNode = nextTempNode;
+        }
+        shouldVisitSubtree = processFullMatchedNode(nextTempNode);
       } else {
-        shouldVisitSubtree = processInternalMatchedNode(node);
+        shouldVisitSubtree = processInternalMatchedNode(nextTempNode);
```
MarcosZyk, line 206: Maybe we should set up the following abstract interfaces:
1. ```acceptInternalMatchedNode``` and ```acceptFullMatchedNode```
2. ```shouldVisitSubtreeOfInternalMatchedNode``` and ```shouldVisitSubtreeOfFullMatchedNode```

node-commons/src/main/java/org/apache/iotdb/commons/schema/tree/AbstractTreeVisitor.java
```diff
@@ -274,11 +292,12 @@ protected final N getParentOfNextMatchedNode() {
...
-  // Release a child with the given childName.
-  protected void releaseChild(N child) {}
+  // Release a child node.
+  protected void releaseNode(N node) {}
 
-  // Release an iterator of all children.
-  protected void releaseChildrenIterator(Iterator<N> childrenIterator) {}
+  // Release an iterator. It is not necessary to deal with all the elements in the iterator.
+  // Only the elements that have been fetched but not returned by next() need to be released.
+  protected void releaseNodeIterator(Iterator<N> nodeIterator) {}
```
MarcosZyk, line 300: seems no invokation

MarcosZyk (approved):

PR #8725, opened 2023-01-04
[To rel/0.13][IOTDB-5346] Fix log error in MemtableFlushTask when recovering

HTHou:
## Description

See IOTDB-5346


HeimingZ (approved):
LGTM

PR #8728, opened 2023-01-04
[IOTDB-5347] Implement iterating query for devices and timeseries schema query

MarcosZyk:
### Motivation

The origin implementation for devices and timeseries schema query may cost huge memory peak occupation, since all the results should be retrieved and temporarily cached.

### Modification

Implement iterating query based on schema reader in DeviceSchemaScanOperator, TimeSeriesSchemaScanOperator and PathsUsingTemplateScanOperator.


Cpaulyz (approved):
LGTM

PR #8746, opened 2023-01-04
[To rel/1.0][IOTDB-5356]Rename SeriesSlotId to SeriesSlotNum in show regions (#8326)

CRZbulabula:


server/src/main/java/org/apache/iotdb/db/mpp/common/header/ColumnHeaderConstant.java
```diff
@@ -97,8 +97,8 @@ public class ColumnHeaderConstant {
   public static final String REGION_ID = "RegionId";
   public static final String TYPE = "Type";
   public static final String DATA_NODE_ID = "DataNodeId";
-  public static final String SERIES_SLOT_ID = "SeriesSlotId";
-  public static final String TIME_SLOT_ID = "TimeSlotId";
+  public static final String SERIES_SLOT_NUM = "SeriesSlotNum";
```
Beyyes, line 100: SeriesSlotId TimeSlotId are also needed

PR #8770, opened 2023-01-06
[To rel/1.0] [IOTDB-5312] Consolidate ClientManagers in Datanodes for unified management

OneSizeFitsQuorum:
cherry-pick [pull/8654](https://github.com/apache/iotdb/pull/8654)

### DIFF
--- a/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
+++ b/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
@@ -21,12 +21,11 @@ package org.apache.iotdb.consensus.ratis;
 
 import org.apache.iotdb.common.rpc.thrift.TEndPoint;
 import org.apache.iotdb.common.rpc.thrift.TSStatus;
-import org.apache.iotdb.commons.client.ClientFactoryProperty;
 import org.apache.iotdb.commons.client.ClientManager;
-import org.apache.iotdb.commons.client.ClientPoolProperty;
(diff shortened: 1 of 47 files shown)


CRZbulabula (approved):
LGTM～

consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
```diff
@@ -812,11 +813,7 @@ public KeyedObjectPool<RaftGroup, RatisClient> createClientPool(
         ClientManager<RaftGroup, RatisClient> manager) {
       return new GenericKeyedObjectPool<>(
           new RatisClient.Factory(
-              manager,
-              new ClientFactoryProperty.Builder().build(),
-              properties,
-              clientRpc,
-              MemoizedSupplier.valueOf(() -> config.getImpl())),
+              manager, properties, clientRpc, MemoizedSupplier.valueOf(config::getImpl)),
```
SzyWilliam, line 816: `config::getImpl` is computed during initialization, which will construct a closure that includes uninitialized config. we should use `() -> config.getImpl())` here to postpone its computation to the first call.

SzyWilliam (changes requested):

consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
OneSizeFitsQuorum: I've handled this by adjusting the initial location of the clientManager ~

PR #8776, opened 2023-01-07
Revert the commit 5126711

xingtanzjr:
## Description
Revert the commit 5126711 because this commit made the performance change regarding write/read


Beyyes (approved):
LGTM

PR #8792, opened 2023-01-09
[IOTDB-5386] Implement SchemaSource Component to Simplify Schema Operators

MarcosZyk:
### Motivation
Schema query has several result types, including device, timeseries and MTree node. Currently, we implemented scan operators for each type's info query, which share the common query logic and only vary on SchemaInfo-2-TsBlock transformation logic.

We want to avoid the operator explosion resulted from the multiplication of query logic and result types. The operator shall focus on the common logic of result processing rather than specified format transformation, which should be implemented as a component of the operator.

### Design

See  https://apache-iotdb.feishu.cn/docx/WmPUdeLLeoG5BjxQXYscLmPSnNg

### Modification

1. Implement the SchemaSource 
2. Eliminate DeviceCountOperator, TimeSeriesCountOperator, DeviceSchemaScanOperator, TimeSeriesScanOperator, PathsUsingTemplateScanOperator, and NodePathsScanOperator.
3. Rename LevelTimeSeriesCountOperator to CountGroupByLevelScanOperator, and refactor it to support all types' count group by level.


Cpaulyz (approved):
LGTM

PR #8798, opened 2023-01-09
[IOTDB-5389] Cause DataNode startup to fail when wal_mode is disabled in IoTConsensus

OneSizeFitsQuorum:
see [jira5389](https://issues.apache.org/jira/browse/IOTDB-5389)


Beyyes (approved):
LGTM

PR #8802, opened 2023-01-10
Eliminate LocalSchemaProcessor and stale code

MarcosZyk:
## Description

Since LocalSchemaProcessor is useless in business implementation and only used for UT, even taking no effect, eliminate it.

Remove NodePathsSchemaScanOperator, merged by mistake.


HeimingZ (approved):
The WAL module LGTM.

PR #8826, opened 2023-01-11
[IOTDB-5399] Implement batch auto create schema

MarcosZyk:
### Motivation

This is the first step of [IOTDB-5399](https://issues.apache.org/jira/browse/IOTDB-5399)

The request of auto create schema shall not be executed one device after one, which brings heavy RPC burden. Such request should be implemented as one MPP task, and at least the schema of same SchemaRegion can be executed in one RPC process.

### Modification
1. Refactor the process of schema check and auto create schema
2. Implement InternalBatchActivateTemplate
3. Implement InternalCreateMultiTimeSeries

### Test Result

The time, used for auto creating total 5000 timeseries of one insertRecords request, with 5 dbs, 100 devices in each db and 10 sensors in each device:
1. decreases from 6800ms to 1800ms in 1C3D, including auto create db and regions.
2. decreates from 4500ms  to 300ms in 1C3D, only auto create timeseries.


server/src/main/java/org/apache/iotdb/db/mpp/plan/scheduler/FragmentInstanceDispatcherImpl.java
```diff
@@ -118,20 +124,67 @@ private Future<FragInstanceDispatchResult> dispatchRead(List<FragmentInstance> i
...
-      } catch (FragmentInstanceDispatchException e) {
-        return immediateFuture(new FragInstanceDispatchResult(e.getFailureStatus()));
-      } catch (Throwable t) {
-        logger.warn("[DispatchFailed]", t);
-        return immediateFuture(
-            new FragInstanceDispatchResult(
-                RpcUtils.getStatus(
-                    TSStatusCode.INTERNAL_SERVER_ERROR, "Unexpected errors: " + t.getMessage())));
+      futureList.add(
+          writeOperationExecutor.submit(
```
OneSizeFitsQuorum, line 131: I suspect that such a change would result in a performance rollback in a multi-node production cluster with high concurrent writes.

The thread executing this function is `ClientRPC-Processor`, and the maximum number of `ClientRPC-Processor` threads in threadPool is 65535. However, The coreSize of `writeOperationExecutor` threadPool is `50`. 

Prior to this change, if the concurrency was high enough, the OS could schedule up to 65535 threads despite a lot of synchronous IO, so sending RPCs could still be done in parallel. However, with this change, a maximum of 50 RPCs can currently be sent in parallel, and the remaining requests are blocked, and the higher the network latency, the more obvious this problem is.

For these reasons, I recommend using Thrift `AsyncClient`. Although cumbersome, the significant performance gains on this critical path are well worth it.

OneSizeFitsQuorum (commented):
I am familiar with the current problems at `dispatch`, so I offer my advice. However, I am not particularly familiar with the code of other files, and I did not see any problems after reading them.

server/src/main/java/org/apache/iotdb/db/mpp/plan/scheduler/FragmentInstanceDispatcherImpl.java
MarcosZyk: Such implementation has been reverted, the async client will be implemented in next pr.

OneSizeFitsQuorum (approved):
LGTM

PR #8830, opened 2023-01-11
[IOTDB-5398] Handle failure in schema query if there is exception occurs during the iteration

Cpaulyz:
## Description

* Handle failure in schema query if there is exception occurs during the iteration.
* Add UT to mock exception and check whether Operator throws runtime exception correctlly.


server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/schema/source/PathsUsingTemplateSource.java
```diff
@@ -114,5 +114,15 @@ public boolean hasNext() {
...
+
+    @Override
+    public boolean isSuccess() {
+      return currentDeviceReader.isSuccess();
+    }
+
+    @Override
+    public Throwable getFailure() {
+      return currentDeviceReader.getFailure();
+    }
```
MarcosZyk, line 126: There will be multi readers used for show paths using template. It seems the judgement only on current reader is not effective for the whole process.

MarcosZyk (commented):
PTAL

MarcosZyk (approved):

PR #8835, opened 2023-01-11
[IOTDB-5401] Reduce the estimated size of memory retained after calling next in SeriesAggregationScanOperator

liuminghui233:
Reduce the estimated size of memory retained after calling next.


JackieTien97 (changes requested):
You may need to set inputTsBlock in whose index that already been consumed up to null in AggregationOperator, otherwise, it will occupy the memory.

JackieTien97 (approved):

PR #8862, opened 2023-01-12
Fix a regression issue that the max Region number is not calculated correctly

xingtanzjr:
## Description
This PR is used to fix a regression issue that the max Region number is not calculated correctly. This issue does not exist in release-1.0 because it is introduced by the commit [358be12](https://github.com/apache/iotdb/commit/358be12ca728bf6ff704a000e8d8f8d36f95b4bf).

So we doesn't need to add more items in `Release Notes`


Beyyes (approved):
LGTM

PR #8892, opened 2023-01-17
Add distance-based outlier detection funtion

iotdb-lsmod:
Add distance-based outlier detection funtion and the corresponding statements.


lancelly (approved):
LGTM～

lancelly (approved):

library-udf/src/main/java/org/apache/iotdb/library/anomaly/UDTFOutlier.java
```diff
@@ -0,0 +1,77 @@
+package org.apache.iotdb.library.anomaly;
```
lancelly, line 1: Add license for this file

PR #8908, opened 2023-01-28
[IOTDB-5431] Fix concurrent bug caused by using synchronizedSet() in DriverScheduler

JackieTien97:
more details about this bug can be seen in this [issue](https://issues.apache.org/jira/browse/IOTDB-5431)


wangchao316 (approved):
LGTM

PR #8919, opened 2023-01-29
[IOTDB-5419] Add Region status in DataNode Heartbeat

Caideyipi:
## IOTDB-5419
1. Report region status in DataNode heartbeat,  instead of inheriting the DataNode status. May help distinguish readonly and running regions when regions are in different disks.
2. Set region status to removing when it's migrating.
image
image

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/DataNodeRemoveHandler.java
```diff
@@ -155,6 +157,39 @@ public TDataNodeLocation findDestDataNode(TConsensusGroupId regionId) {
     return newNode.get();
   }
 
+  public TSStatus setRegionStatus(TConsensusGroupId regionId, TDataNodeLocation originalDataNode) {
```
Beyyes, line 160:
```suggestion
  public TSStatus setRegionStatusAsRemoving(TConsensusGroupId regionId, TDataNodeLocation originalDataNode) {
```

thrift/src/main/thrift/datanode.thrift
```diff
@@ -526,6 +532,10 @@ service IDataNodeRPCService {
   */
   common.TSStatus stopDataNode()
 
+  /**
+    * Config node will set a region's status to Data node.
```
CRZbulabula, line 536:
```suggestion
    * ConfigNode will set a Region's status to DataNode.
```

confignode/src/main/java/org/apache/iotdb/confignode/client/async/handlers/heartbeat/DataNodeHeartbeatHandler.java
```diff
@@ -72,8 +72,9 @@ public void onComplete(THeartbeatResp heartbeatResp) {
                       new RegionHeartbeatSample(
                           heartbeatResp.getHeartbeatTimestamp(),
                           receiveTime,
-                          // Region will inherit DataNode's status
-                          RegionStatus.parse(heartbeatResp.getStatus())));
+                          // Each region will be updated by its own status
```
CRZbulabula, line 75:
```suggestion
                          // Each Region will be updated by its own status
```

server/src/main/java/org/apache/iotdb/db/service/thrift/impl/DataNodeInternalRPCServiceImpl.java
```diff
@@ -945,6 +949,27 @@ private PathPatternTree filterPathPatternTree(PathPatternTree patternTree, Strin
...
+      } else {
+        CommonDescriptor.getInstance()
+            .getConfig()
+            .getSchemaRegionStatusMap()
+            .put(
+                new SchemaRegionId(req.getRegionId().getId()), RegionStatus.parse(req.getStatus()));
+      }
+      return new TSStatus(TSStatusCode.SUCCESS_STATUS.getStatusCode());
+    } catch (Exception e) {
+      return new TSStatus(TSStatusCode.SET_REGION_STATUS_ERROR.getStatusCode());
```
CRZbulabula, line 969: What scenario will trigger this error? It's better not to just use catch(Exception e) clause and don't write any error messages, which will cause confuse. For example, the most likely error is a failed to pares RegionStatus. Therefore, we should catch(IllegalException e) and set message "Unknown RegionStatus".

node-commons/src/main/java/org/apache/iotdb/commons/conf/CommonConfig.java
```diff
@@ -20,20 +20,27 @@
...
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 
 public class CommonConfig {
 
   public static final String CONFIG_NAME = "iotdb-common.properties";
   private static final Logger logger = LoggerFactory.getLogger(CommonConfig.class);
 
+  private final Map<SchemaRegionId, RegionStatus> schemaRegionStatusMap = new ConcurrentHashMap<>();
+  private final Map<DataRegionId, RegionStatus> dataRegionStatusMap = new ConcurrentHashMap<>();
```
CRZbulabula, line 43: Is it better to manage these two parameters on DataNodeRegionManager instead of CommonConfig? Since they are only maintained by DataNode, but both ConfigNode and DataNode will cache a CommonConfig instance.

confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/DataNodeRemoveHandler.java
```diff
@@ -155,6 +157,39 @@ public TDataNodeLocation findDestDataNode(TConsensusGroupId regionId) {
...
+                .sendSyncRequestToDataNodeWithGivenRetry(
+                    originalDataNode.getInternalEndPoint(),
+                    req,
+                    DataNodeRequestType.SET_REGION_STATUS,
+                    1)
+            : SyncDataNodeClientPool.getInstance()
+                .sendSyncRequestToDataNodeWithRetry(
+                    originalDataNode.getInternalEndPoint(),
+                    req,
+                    DataNodeRequestType.SET_REGION_STATUS);
```
CRZbulabula, line 176: I notice that this interface is only used in RegionMigrateProcedure. However, is it safe if we migrate a Region that located at a Unknown DataNode? If not, we can just refuse this migration but not try to modify the RegionStatus with 1 retry.

CRZbulabula (changes requested):
PTAL~

confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/DataNodeRemoveHandler.java
Caideyipi: Because a region migration may also be caused by a DataNode removement, which is currently supported on unknown DataNode.

CRZbulabula (approved):
LGTM

PR #8920, opened 2023-01-30
[IOTDB-5337] Parallelization of write operation in FragmentInstanceDispatcher

HeimingZ:


OneSizeFitsQuorum (commented):
What about the test result?

HeimingZ:
> What about the test result?

Not obvious when testing with the dbms.

OneSizeFitsQuorum:
> Not obvious when testing with the dbms.

Whether performance is improved depends on the scenario. For example, the insertRecords and insertTablets write interfaces for different devices may benefit. Because in this scenario, it is profitable to send RPC concurrently

MarcosZyk:
If the load already reaches the cluster's concurrency ceil, such modification does bring no benefit. If the load is low, for example, the number of clients is less than the number of DataNode, such modification will take effects.

server/src/main/java/org/apache/iotdb/db/mpp/plan/scheduler/FragmentInstanceDispatcherImpl.java
```diff
@@ -118,10 +128,25 @@ private Future<FragInstanceDispatchResult> dispatchRead(List<FragmentInstance> i
     return immediateFuture(new FragInstanceDispatchResult(true));
   }
 
-  private Future<FragInstanceDispatchResult> dispatchWriteSync(List<FragmentInstance> instances) {
```
xingtanzjr, line 121: Let's keep original Sync method

server/src/main/java/org/apache/iotdb/db/mpp/plan/scheduler/FragmentInstanceDispatcherImpl.java
```diff
@@ -210,6 +265,29 @@ private void dispatchRemote(FragmentInstance instance, TEndPoint endPoint)
     }
   }
 
+  public void dispatchRemoteAsync(
```
xingtanzjr, line 268: Can we define a seperate class to handle the whole logic

server/src/main/java/org/apache/iotdb/db/mpp/plan/scheduler/FragmentInstanceDispatcherImpl.java
HeimingZ: Fixed.

server/src/main/java/org/apache/iotdb/db/mpp/plan/scheduler/FragmentInstanceDispatcherImpl.java
HeimingZ: Add AsyncPlanNodeSender class.

xingtanzjr (approved):

OneSizeFitsQuorum (approved):
LGTM

PR #8923, opened 2023-01-30
[IOTDB-5140][To Rel/1.0] Add metrics for compaction deserializing pages or writing chunks

THUMarkLau:
See [IOTDB-5140](https://issues.apache.org/jira/browse/IOTDB-5140).

This PR add serveral metrics for compaction execution, especially for the method that the chunk is processed. There are three types of chunk process method, which are directly-flush, merged, and deserialized into points. Monitoring the method the chunk is compacted can benifit the tuning of compaction.

3b39e1bbca488ba3add588ab65204c9


SpriCoder (approved):
LGTM~

PR #8947, opened 2023-01-31
[To rel/1.0] Add setLeastDataRegionGroupNumSetByUser judgement

Beyyes:
## Description

cherry-pick some code of https://github.com/apache/iotdb/pull/8901

### DIFF
--- a/confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeConfig.java
+++ b/confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeConfig.java
@@ -101,6 +101,9 @@ public class ConfigNodeConfig {
   /** The least number of DataRegionGroup for each Database. */
   private volatile int leastDataRegionGroupNum = 5;
 
+  /** Indicate whether the leastDataRegionGroupNum is set by user, if true, lock it. */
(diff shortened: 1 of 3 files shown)


CRZbulabula (approved):
LGTM～

confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeConfig.java
```diff
@@ -101,6 +101,9 @@ public class ConfigNodeConfig {
   /** The least number of DataRegionGroup for each Database. */
   private volatile int leastDataRegionGroupNum = 5;
 
+  /** Indicate whether the leastDataRegionGroupNum is set by user, if true, lock it. */
+  private volatile boolean leastDataRegionGroupNumSetByUser = false;
```
Caideyipi, line 105: image
This change should be cherry-picked too, otherwise the IT will fail.

Beyyes: The codes in rel/1.0 is different from master. There is no `testLeastDataRegionGroupNum` variable in the test class.

PR #8951, opened 2023-02-01
[IOTDB-5441] Fix NPE while fetch schema that is not in template used by related device

MarcosZyk:
### Cause

While fetching a normal timeseries under a device using template, since there's no template provided for traverser, which is brought by the mpp  or schema fetcher analyzing stage result, current implementation still tries to check the template and result in npe.

### Solution

Skip the template check while there's no template provided for traverser.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/mtree/traverser/Traverser.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/mtree/traverser/Traverser.java
@@ -96,6 +96,7 @@ public abstract class Traverser<R> extends AbstractTreeVisitor<IMNode, R> {
     }
     if (!isSuccess()) {
       Throwable e = getFailure();
+      e.printStackTrace();
       throw new MetadataException(e.getMessage(), e);
     }
   }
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/iotdb/db/metadata/mtree/traverser/Traverser.java
```diff
@@ -106,15 +112,21 @@ protected IMNode getChild(IMNode parent, String childName) throws MetadataExcept
     if (parent.isAboveDatabase()) {
       child = parent.getChild(childName);
     } else {
-      if (parent.getSchemaTemplateId() != NON_TEMPLATE) {
-        if (!skipPreDeletedSchema || !parent.getAsEntityMNode().isPreDeactivateTemplate()) {
+      child = store.getChild(parent, childName);
```
Cpaulyz, line 115: Perhaps still trying to fetch from the template first and then calling store.getChild would reduce some disk I/O in schemaFile mode. Just need to avoid the NPE when fetching from the template.

MarcosZyk: Fixed.

Cpaulyz (approved):
LGTM

PR #8953, opened 2023-02-01
Correct DataPartiton Fetch request parameter contruction

JackieTien97:
In this pr, we do four things:
1. Fix previous TDataPartitionReq contruction bug. Previously, while calculating TTimeSlot for each SeriesSlot, we mistakenly replace previous one with current one which cause the TTimeSlot is incomplete for each SeriesSlot. In write case, we may get IndexOutOfBound error in the later phase.
2. Accelerate the distribution plan phase for `group by level query`, reducing time complexity from O(N^2) to O(N) for each data region
3. Correct the memory estimation for AggregationOperator.
4. Correct the default value for `seriesPartitionExecutorClass` from `APHashExecutor` to `BKDRHashExecutor` which is same as the one in config file.


server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/distribution/SourceRewriter.java
```diff
@@ -953,21 +965,43 @@ private void calculateGroupByLevelNodeAttributes(
...
+        //            descriptorExpressions.add(originalDescriptor.getOutputExpression());
+        //            continue;
+        //          }
+        //          for (Expression exp : originalDescriptor.getInputExpressions()) {
+        //            if (isAggColumnMatchExpression(childColumn, exp)) {
+        //              descriptorExpressions.add(exp);
+        //            }
+        //          }
+        //        }
+
```
HTHou, line 1004: Remove this?

HTHou (approved):

PR #8954, opened 2023-02-01
[To rel/1.0][IOTDB-5446] Add V_1_0 version and update version used in IoTDB

SpriCoder:
cherry picked from #8952


HTHou (approved):

HTHou (approved):

client-cpp/src/main/Session.cpp
```diff
@@ -793,6 +793,8 @@ string Session::getVersionString(Version::Version version) {
             return "V_0_12";
         case Version::V_0_13:
             return "V_0_13";
+        case Version::V_1_0,
```
HTHou, line 796:
```suggestion
        case Version::V_1_0:
```

PR #8957, opened 2023-02-01
[IOTDB-5449] Allow query scheduler to receive more tasks

Alima777:
Currently, the MAX_CAPACITY of query scheduler is 1000. But after pipelining, the number of query tasks maybe larger. Therefore, I turn up the size of task ready queue. 
And a problem is fixed in this pr. Blocked task or runnning task is polled out from ready queue, but they can't join the ready queue again if it's full.


server/src/main/java/org/apache/iotdb/db/mpp/execution/schedule/queue/IndexedBlockingQueue.java
```diff
@@ -67,31 +73,38 @@ public synchronized E poll() throws InterruptedException {
...
     return output;
   }
 
   /**
    * Push an element to the queue. The new element position is determined by the implementation. If
-   * the queue size has been reached the maxCapacity, or the queue has already contained an element
-   * with the same ID, an {@link IllegalStateException} will be thrown. If the element is null, an
-   * {@link NullPointerException} will be thrown.
+   * the queue has already contained an element with the same ID, an {@link IllegalStateException}
+   * will be thrown. If the queue size has been reached the maxCapacity, the element will wait a
```
lancelly, line 83: If the queue size has been reached => If the queue size has reached

lancelly:
> Currently, the MAX_CAPACITY of query scheduler is 1000. If size of queue is over 1000, the task will be rejected. We can let this task wait until there is a free space for it.

It seems that in this pr, we will continuously try to push the DriverTask into the queue when it is full and it is possible that the Task is pushed into the queue after it reaches TIME_OUT limit. I wonder if it is a problem.

Alima777:
> It seems that in this pr, we will continuously try to push the DriverTask into the queue when it is full and it is possible that the Task is pushed into the queue after it reaches TIME_OUT limit. I wonder if it is a problem.

Good question. Before pushing the task to the queue, READY status will be checked. So it will not be a problem.

server/src/main/java/org/apache/iotdb/db/mpp/execution/schedule/DriverScheduler.java
```diff
@@ -198,15 +200,34 @@ public void submitDrivers(QueryId queryId, List<IDriver> drivers, long timeOut)
...
+    SettableFuture<?> isBlocked = readyQueue.push(task);
+    if (isBlocked.isDone()) {
+      timeoutQueue.push(task);
+      task.setLastEnterReadyQueueTime(System.nanoTime());
+    } else {
+      isBlocked.addListener(
+          () -> {
+            task.lock();
+            try {
+              if (task.getStatus() == DriverTaskStatus.READY) {
```
JackieTien97, line 220: it seems that it will always be true

server/src/main/java/org/apache/iotdb/db/mpp/execution/schedule/queue/IndexedBlockingQueue.java
```diff
@@ -40,6 +44,7 @@
   private final int MAX_CAPACITY;
   private final E queryHolder;
   private int size;
+  private Deque<SettableFuture<?>> blockedTaskFutures;
```
JackieTien97, line 47: we need to control the size of this queue

JackieTien97 (changes requested):
We need a timeout mechanism to stop task from heaping up

lancelly (approved):

server/src/main/java/org/apache/iotdb/db/mpp/execution/schedule/queue/IndexedBlockingReserveQueue.java
```diff
@@ -0,0 +1,86 @@
...
+  /**
+   * RePush an element which is polled out for running or blocked before to the queue. In this case,
+   * we don't increase the current size of queue, since we have reserved a place for it.
+   */
+  public synchronized void repush(E element) {
+    if (element == null) {
+      throw new NullPointerException("pushed element is null");
+    }
+    pushToQueue(element);
+    reservedSize--;
```
JackieTien97, line 76: needn't do `size++`?

server/src/main/java/org/apache/iotdb/db/mpp/execution/schedule/queue/IndexedBlockingReserveQueue.java
```diff
@@ -0,0 +1,92 @@
...
+    return output;
+  }
+
+  public synchronized void push(E element) {
+    if (element == null) {
+      throw new NullPointerException("pushed element is null");
+    }
+    Preconditions.checkState(
+        !contains(element),
+        "The queue has already contained the element: " + element.getDriverTaskId());
```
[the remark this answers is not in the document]
JackieTien97: same as the method in `IndexedBlockingQueue `

server/src/main/java/org/apache/iotdb/db/mpp/execution/schedule/DriverScheduler.java
```diff
@@ -79,20 +80,22 @@ public static DriverScheduler getInstance() {
   private final AtomicInteger nextDriverTaskHandleId = new AtomicInteger(0);
   private IMPPDataExchangeManager blockManager;
 
-  private static final int MAX_CAPACITY =
+  private static final int QUERY_MAX_CAPACITY =
       IoTDBDescriptor.getInstance().getConfig().getMaxAllowedConcurrentQueries();
   private static final int WORKER_THREAD_NUM =
       IoTDBDescriptor.getInstance().getConfig().getQueryThreadCount();
+  private static final int TASK_MAX_CAPACITY = QUERY_MAX_CAPACITY * WORKER_THREAD_NUM * 2;
```
JackieTien97, line 87: Is there some reasons about this parameter setter?

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/mpp/execution/schedule/queue/IndexedBlockingReserveQueue.java
Alima777: Fixed.

server/src/main/java/org/apache/iotdb/db/mpp/execution/schedule/queue/IndexedBlockingReserveQueue.java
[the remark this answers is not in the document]
Alima777: Yes, I think so.

JackieTien97 (approved):

PR #8958, opened 2023-02-01
[IOTDB-5393] Show Region creation time when execute show regions

CRZbulabula:
image


OneSizeFitsQuorum (commented):
Do we need to consider time zones?

CRZbulabula:
> Do we need to consider time zones?

Yes, I've added this feature~

OneSizeFitsQuorum (commented):
Should we create a separate `createTime` column? Otherwise it seems ambiguous...

CRZbulabula:
> Should we create a separate `createTime` column? Otherwise it seems ambiguous...

Created

OneSizeFitsQuorum (approved):
LGTM

xingtanzjr (approved):

PR #8963, opened 2023-02-02
[IOTDB-5399] Collect complete dispatch result for return

MarcosZyk:
## Description

Currently, one FI dispatch failure will result in the dispatch of rest FI being skipped, which cannot bring complete result for mpp client.

Therefore, improve the dispatch process to dispatch all FI of one mpp task and collect all the result for return.


HeimingZ (approved):
LGTM

PR #8966, opened 2023-02-02
[IOTDB-5459] Memory control cross SchemaRegion

Cpaulyz:
## Description

The shortcoming of the current implementation is that a schemaRegion can only trigger the release of its own occupied memory and cannot affect other schemaRegions, lacking global memory control. It may lead to an incorrect SeriesOverflowException

For example, if there is a SchemaRegion1 and a SchemaRegion2, the IMnode of schemaRegion1 needs to be considered if writes to schema region2 cause the total memory to exceed the release threshold and trigger an internal/external memory swap. 

Therefore, a global CacheMemoryManager needs to be introduced for memory control.

https://apache-iotdb.feishu.cn/docx/IVtgdbJNPoIjpdxlQTocOuTTnLh


server/src/main/java/org/apache/iotdb/db/metadata/mtree/store/disk/cache/CacheMemoryManager.java
```diff
@@ -0,0 +1,125 @@
...
+  private volatile boolean hasFlushTask;
+
+  private final MTreeReleaseTaskManager releaseTaskManager = MTreeReleaseTaskManager.getInstance();
+  private volatile boolean hasReleaseTask;
+  private int releaseCount = 0;
+
+  public ICacheManager createLRUCacheManager(CachedMTreeStore store) {
+    ICacheManager cacheManager = new LRUCacheManager();
+    storeList.add(store);
+    return cacheManager;
```
MarcosZyk, line 47: This method should be synchronized or use a thread-safe list, in case of concurrent creating SchemaRegion.

server/src/main/java/org/apache/iotdb/db/metadata/mtree/store/disk/cache/CacheMemoryManager.java
```diff
@@ -0,0 +1,125 @@
...
+      registerReleaseTask();
+    }
+  }
+
+  private synchronized void registerReleaseTask() {
+    if (hasReleaseTask) {
+      return;
+    }
+    hasReleaseTask = true;
+    releaseTaskManager.submit(this::tryExecuteMemoryRelease);
```
MarcosZyk, line 61: It seems there won' be more than two release tasks being executed concurrently. The efficiency is not much high enough, and it's a waste to prepare an independent thread pool for such tasks.

server/src/main/java/org/apache/iotdb/db/metadata/mtree/store/disk/cache/CacheMemoryManager.java
```diff
@@ -0,0 +1,125 @@
...
+      registerFlushTask();
+    }
+  }
+
+  private synchronized void registerFlushTask() {
+    if (hasFlushTask) {
+      return;
+    }
+    hasFlushTask = true;
+    flushTaskManager.submit(this::tryFlushVolatileNodes);
```
MarcosZyk, line 93: Same problem as release task. Maybe these two boolean field should be attributes of store rather than a global unique attribute.

server/src/main/java/org/apache/iotdb/db/metadata/mtree/store/disk/cache/CacheMemoryManager.java
Cpaulyz: fixed

server/src/main/java/org/apache/iotdb/db/metadata/mtree/store/disk/cache/CacheMemoryManager.java
Cpaulyz: replaced by SingleThreadExecutor

server/src/main/java/org/apache/iotdb/db/metadata/mtree/store/disk/cache/CacheMemoryManager.java
Cpaulyz: I think the two global boolean variables are used to avoid duplicate release/flush requests during the release/flush process, which is needed since the release/flush thread is currently globally unique. When hasFlushTask is false, all store's hasFlushTask is also false, so there seems to be no need to migrate to store?

node-commons/src/main/java/org/apache/iotdb/commons/concurrent/ThreadName.java
```diff
@@ -61,8 +61,9 @@ public enum ThreadName {
   ASYNC_DATANODE_HEARTBEAT_CLIENT_POOL("AsyncDataNodeHeartbeatServiceClientPool"),
   ASYNC_CONFIGNODE_CLIENT_POOL("AsyncConfigNodeIServiceClientPool"),
   ASYNC_DATANODE_MPP_DATA_EXCHANGE_CLIENT_POOL("AsyncDataNodeMPPDataExchangeServiceClientPool"),
-
-  ASYNC_DATANODE_IOT_CONSENSUS_CLIENT_POOL("AsyncDataNodeMPPDataExchangeServiceClientPool");
+  ASYNC_DATANODE_IOT_CONSENSUS_CLIENT_POOL("AsyncDataNodeMPPDataExchangeServiceClientPool"),
+  MTREE_FLUSH_THREAD_POOL("MTree-flush-task"),
+  MTREE_RELEASE_THREAD_POOL_NAME("MTree-release-task");
```
MarcosZyk, line 66: Rename to ```SchemaRegion-Flush-Task``` and ```SchemaRegion-Release-Task```.

Cpaulyz: done

server/src/main/java/org/apache/iotdb/db/metadata/mtree/store/disk/cache/CacheMemoryManager.java
Cpaulyz: Use CompleteFuture to release and flush concurrently.

server/src/main/java/org/apache/iotdb/db/metadata/mtree/store/disk/cache/CacheMemoryManager.java
Cpaulyz: Use CompleteFuture to release and flush concurrently.

MarcosZyk (approved):

PR #8981, opened 2023-02-05
Accelerate Count TimeSeries

MarcosZyk:
## Description

Currently, the time series reader is implemented based on ```ShowTimeSeriesResult```, which is used as temporary schema result carrier and costs construction time. The construction time of ```ShowTimeSeriesResult``` can be saved via a light implementation generating fewest extra temporary objects.

Count timeseries operation benefits from the modification most. Its time consumption decreased to 1/4~1/3.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/schemaregion/ISchemaRegion.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/schemaregion/ISchemaRegion.java
@@ -258,6 +258,10 @@ public interface ISchemaRegion {
   ISchemaReader<IDeviceSchemaInfo> getDeviceReader(IShowDevicesPlan showDevicesPlan)
       throws MetadataException;
 
+  /**
+   * The iterated result shall be consumed before calling reader.haxNext() or read.next(). Its
+   * implementation is based on the reader's process context.
+   */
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/apache/iotdb/db/metadata/schemaregion/ISchemaRegion.java
```diff
@@ -258,6 +258,10 @@ long countPathsUsingTemplate(int templateId, PathPatternTree patternTree)
   ISchemaReader<IDeviceSchemaInfo> getDeviceReader(IShowDevicesPlan showDevicesPlan)
       throws MetadataException;
 
+  /**
+   * The iterated result shall be consumed before calling reader.haxNext() or read.next(). Its
+   * implementation is based on the reader's process context.
+   */
```
Cpaulyz, line 264:
```suggestion
   * The iterated result shall be consumed before calling reader.hasNext() or reader.next(). Its
   * implementation is based on the reader's process context.
   */
```

Cpaulyz (approved):
LGTM

server/src/main/java/org/apache/iotdb/db/metadata/schemaregion/ISchemaRegion.java
MarcosZyk: Fixed.

PR #8993, opened 2023-02-07
[IOTDB-5477] Refactor the interaction between InsertNode and SchemaFetcher

MarcosZyk:
### Motivation

The schema cache read, schemaTree construction and insertNode.validateSchema repeat the process of scanning device path and measurements of input timeseries, which is redundant and can be simplified into one time process for all.

### Modification

Refer to the design of ```Map.compute```, we set the similar interfaces for schema fetcher, ```SchemaFetcher.computeSchemaWithAutoCreate```. During reading schema cache and checking the fetched schema tree, the specific validation operation provided by InsertNode will be executed without repeated scan.

### Result

When all the schema for data insertion has been cached, the time cost of schema validation nearly all focus on cache.get.

image


Cpaulyz (approved):
LGTM

PR #8997, opened 2023-02-07
[IOTDB-5479] Optimize the wal impl of confignode simple consensus

Beyyes:
## Description

TODO: need more experiments

1. disk io comparison:  ratis 1 replica v.s. simple

### DIFF
--- a/confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeConfig.java
+++ b/confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeConfig.java
@@ -287,6 +287,8 @@ public class ConfigNodeConfig {
   /** The getOrCreatePartitionTable interface will log new created Partition if set true */
   private boolean isEnablePrintingNewlyCreatedPartition = false;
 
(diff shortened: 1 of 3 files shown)


OneSizeFitsQuorum (commented):
Perhaps confignode wal should not use the trikky method of timed or fixed threshold flush. Because the confignode stores partition information, it does not tolerate data loss in the limit case like wal of Datanodes. Maybe we don't have to fsync every time, but we should at least write to the page cache every time

CRZbulabula (approved):
LGTM～

CRZbulabula:
> Perhaps confignode wal should not use the trikky method of timed or fixed threshold flush. Because the confignode stores partition information, it does not tolerate data loss in the limit case like wal of Datanodes. Maybe we don't have to fsync every time, but we should at least write to the page cache every time

I agree with that. Nevertheless, how about we merge this improvement first, and I'll add a test code tomorrow. Therefore, we can have a basic efficiency view between Ratis consensus protocol and Simple consensus protocol about ConfigNode.

PR #8998, opened 2023-02-07
[IOTDB-5487] Fix the problem that the timestamp is "null" when using jdbc

RYH61:
see [IOTDB-3936]Add an interface in IClientRPCService to directly return bytebuffer instead of TSQueryDataSet

This pr is modified to add `lastReadWasNull = false;` of the `constructOneRow` method to the `constructOneTsBlock` method, resulting in a call to the next() method that meets the first of the following criteria, The `lastReadWasNull` field was not reset to false.

image

### DIFF
--- a/service-rpc/src/main/java/org/apache/iotdb/rpc/IoTDBRpcDataSet.java
+++ b/service-rpc/src/main/java/org/apache/iotdb/rpc/IoTDBRpcDataSet.java
@@ -271,6 +271,7 @@ public class IoTDBRpcDataSet {
 
   public boolean next() throws StatementExecutionException, IoTDBConnectionException {
     if (hasCachedBlock()) {
+      lastReadWasNull = false;
       constructOneRow();
       return true;
     }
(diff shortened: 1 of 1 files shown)


HTHou:
No changed line?

RYH61:
> No changed line?
This is deleted by mistake

RYH61:
@ycycse please have a look~

neuyilan (approved):
LGTM

ycycse:
Sorry I've just seen the message.

`lastReadWasNull` will be set after a method of `getBoolean`/`getString`/... is invoked.  Actually, the get-method will return a default value even if it's null, so we need `leadReadWasNull` for a judgement. So I think it'll be overrided eventually.

Maybe you can explain more clearly or give an exmaple of the scene you face?

ycycse (approved):
Sorry I've just seen the message... I should check my notification more often.

`getTimestamp()` won't override the `lastReadWasNull`, so it needs a default value. Thank you for pointing out that.

PR #9001, opened 2023-02-07
[CLIENT-PY]Optimize array concatenation performance

fuwei3140:
[CLIENT-PY]When fetching data from the server in batches, each np.concatenate() will create a new array and copy the stock array. To optimize performance, after parsing all batches of arrays, execute np.concatenate() only once.


client-py/tests/test_todf.py
```diff
@@ -178,7 +178,7 @@ def test_with_null_query():
         df_output = df_output[df_input.columns.tolist()]
 
         session.close()
-    assert_frame_equal(df_input, df_output)
+    assert_frame_equal(df_input, df_output, check_dtype=False)
```
SteveYurongSu, line 181: Why we have to set check_dtype=False?

HTHou, line 181:
```suggestion
    assert_frame_equal(df_input, df_output)
```

HTHou: Fixed

client-py/iotdb/utils/IoTDBRpcDataSet.py
```diff
@@ -266,18 +262,17 @@ def resultset_to_pandas(self):
...
-                    else:
-                        result[column_name] = np.concatenate(
-                            (result[column_name], data_array), axis=0
-                        )
+                result[column_name].append(data_array)
+
         for k, v in result.items():
-            if v is None:
-                result[k] = []
+            if v[0].dtype == "Int32":
```
HTHou, line 268:
```suggestion
            if v is None or len(v) < 1 or v[0] is None:
                result[k] = []
            elif v[0].dtype == "Int32":
```

client-py/iotdb/utils/IoTDBRpcDataSet.py
```diff
@@ -266,18 +262,19 @@ def resultset_to_pandas(self):
...
+            if v is None or len(v) < 1 or v[0] is None:
                 result[k] = []
+            elif v[0].dtype == "Int32":
+                result[k] = pd.Series(np.concatenate(v, axis=0)).astype("Int32")
+            elif v[0].dtype == "Int64":
+                result[k] = pd.Series(np.concatenate(v, axis=0)).astype("Int64")
+            elif v[0].dtype == "boolean":
+                result[k] = pd.Series(np.concatenate(v, axis=0)).astype("boolean")
+            else:
+                result[k] = np.concatenate(v, axis=0)
```
SteveYurongSu, line 277: What about the float number type? Should the float type be considered too?

SteveYurongSu (approved):
LGTM

PR #9009, opened 2023-02-07
[IOTDB-5509] Add level control for cross compaction selecting

xingtanzjr:
## Description
In this change, we add a limitation for cross compaction selector to increase the compaction efficiency. That is, only the unseqFile whose level larger that 0 can be selected as the candidate of cross compaction.

So, the small unseq file will be compacted inner before doing cross space compaction


choubenson (approved):
LGTM

Beyyes (approved):

xingtanzjr:
This change is completed in another PR[9059]. close it now

PR #9015, opened 2023-02-08
[IOTDB-5497] Test for ConfigNode consensus efficiency

CRZbulabula:
The result of test will be formatted as follows:

image


Beyyes (approved):
LGTM

PR #9016, opened 2023-02-08
Fix IT error by parallel suites

ericpai:
Disable thread-level parallel test in IT. And refine some test codes.


ericpai:
@Cpaulyz I set `<parallel>` to `none` to disable thread-level parallet test in one JVM fork. We can still achieve parallel test through setting `<forkCount>`. 
@MarcosZyk Some test codes has been refined
- The LastQueueNonCache case is not robust.
- It's better to use `try-with-resource` block to allocate JDBC connections.

MarcosZyk (approved):

PR #9018, opened 2023-02-08
[IOTDB-5482][To rel/0.12]Unsafe deserialize map in Sync Tool

yschengzi:
The owner file will converted into a Map object. in this
deserialization process, a dangerous function readObject is used.
If device_Owner is a malicious file, it would cause executing
arbitrary code issue in the target database.


SteveYurongSu:
rel/0.12.x is out of life cycle, we will NOT fix any issues at branch rel/0.12.x.

PR #9024, opened 2023-02-09
[IOTDB-5504] Fix illegal argument exception occurs when scheduling compaction

THUMarkLau:
See [IOTDB-5504](https://issues.apache.org/jira/browse/IOTDB-5504).

The reason for this bug is that integer overflow occurs when casting long to integer in `executeCompaction`.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/DataRegion.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/DataRegion.java
@@ -115,6 +115,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/DataRegion.java
```diff
@@ -2120,11 +2120,21 @@ private void closeUnsealedTsFileProcessorCallBack(TsFileProcessor tsFileProcesso
...
+          (o1, o2) -> {
+            long diff = o1 - o2;
+            if (diff > 0) {
+              return 1;
+            } else if (diff < 0) {
+              return -1;
+            } else {
+              return 0;
+            }
+          });
```
JackieTien97, line 2137:
```suggestion
      timePartitions.sort(Comparator.reverseOrder());
```

JackieTien97 (changes requested):

JackieTien97 (approved):

JackieTien97 (approved):

PR #9026, opened 2023-02-09
[ Metrics ] The repair leader distribution panel

23931017wu:
### Leadership Distribution
The reason why nodes are not limited before is that the leader distribution can be observed on the non-leader confignode. However, when there are multiple iotdb clusters in prom, the panel will display the leader distribution of all clusters, causing the cluster display to be unclear. Therefore, it is better to add instance in the attribute to limit.

Before : 
图片

After:
图片

### Other small problems

图片
图片

### DIFF
--- a/docs/UserGuide/Monitor-Alert/Apache-IoTDB-ConfigNode-Dashboard.json
+++ b/docs/UserGuide/Monitor-Alert/Apache-IoTDB-ConfigNode-Dashboard.json
@@ -487,7 +487,7 @@
           "interval": "",
           "legendFormat": "Total Region",
           "range": true,
-          "refId": "C"
+          "refId": "A"
         },
         {
           "datasource": {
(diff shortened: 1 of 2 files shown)


docs/UserGuide/Monitor-Alert/Apache-IoTDB-ConfigNode-Dashboard.json
```diff
@@ -1190,11 +1190,11 @@
             "uid": "${DS_PROMETHEUS}"
           },
           "editorMode": "code",
-          "expr": "sum(rate(process_cpu_time{instance=\"$instance\",name=\"process\"}[1m]))*60",
```
neuyilan, line 1193: for cpu time, FYI

https://stackoverflow.com/questions/58998202/does-multithreading-increase-cpu-time

so it cannot be divided by the number of cpu cores

23931017wu: I see. It has been withdrawn.

neuyilan (approved):
LGTM

neuyilan:
cherry-pick to master

PR #9033, opened 2023-02-09
[IOTDB-5470] Refactor compaction metrics

THUMarkLau:
See [IOTDB-5470](https://issues.apache.org/jira/browse/IOTDB-5470).

The original way of collecting compaction tasks metrics is too complicated, which easily leads to errors. This PR refactored CompactionMetrics to use statistics in a similar way to FileMetrics.

image


SpriCoder (approved):
LGTM~

PR #9040, opened 2023-02-10
[IOTDB-5513] Fix duplicate LastQueryNode due to change of partition slot algorithm

lancelly:
see https://issues.apache.org/jira/browse/IOTDB-5513


server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/distribution/ExchangeNodeAdder.java
```diff
@@ -447,9 +447,7 @@ private PlanNode processMultiChildNode(MultiChildProcessNode node, NodeGroupCont
         child -> {
           // If the child's region is NOT_ASSIGNED, it means the child do not belong to any
```
ericpai, line 448: It seems that this comment is not satisified the following logic any more, just remove it.

ericpai (approved):

JackieTien97 (approved):

PR #9041, opened 2023-02-10
[IOTDB-5514] Fix race condition in MemoryPool without synchronized block

ericpai:
See JIRA: https://issues.apache.org/jira/browse/IOTDB-5514


ericpai:
@lancelly @JackieTien97 PTAL

lancelly:
LGTM~

PR #9050, opened 2023-02-13
[IOTDB-5474] Refactor DataNodeSchemaCache Structure

MarcosZyk:
### Motivation

The flattened cache structure, full path -> schema cache entry, is not efficient enough in terms of memory usage and request execution when encountering most iot scenarios, requests organized as device -> measurement -> data.

### Design

https://apache-iotdb.feishu.cn/docx/Ruz7d7wacoEVuRx9vnxcFeOxn0f

### Implementation

Implement ```DualKeyCache```, which support structure like, k1->k2->value. 

Refactor the Implementation of DataNodeSchemaCache based on ```DualKeyCache```.

### Test Results

1C1D, 8G memory

#### BM Test

bm config: 500 device; 1000 sensor/device; insertRecords; batch size 1; loop 50
before this pr: 20w points/s
this pr: 34w points/s

#### InsertRecords Scenario Simulation Test

 15000 device; 40 sensor/device

master: 110w lines/minute; 0.78 cache hit rate
this pr: 140w lines/minute; 0.98 cache hit rate


bigreybear (approved):
Highly abstracted design, and excellent solid implementation!
Only one trivial suggestion that it might be more clear if `cacheEntryManager` was named or commented  about second key manager of which it plays the role.

PR #9054, opened 2023-02-13

THUMarkLau:
See [IOTDB-5517](https://issues.apache.org/jira/browse/IOTDB-5517).
# Overview
This PR add metrics for disk i/o and process i/o status. When IoTDB runs in different os, the way to monitor the i/o status is different. **Currently only disk i/o monitor in Linux is implemented, monitor in Windows and MacOS may be implemented later.**

c7146d18f01a050208356deb58e7d8b

# Metrics
For disk i/o, now monitor traces the following metrics.
## Disk IO Busy Rate
image
### What It Means
This metric indicates how busy the disk I/O is, which represents how much time per second the disk is engaged in I/O. **But according to this [blog](https://bean-li.github.io/dive-into-iostat/), this metric can be misleading in some cases.** But here we still use this metric to give a rough indication of how busy the disk is.
### How To Get This Metric
_/proc/diskstats_ stores some running information of the disk, in which _io_ticks_ is a field that indicates how much time is used the running disk i/o. We calculate the percentage of disk I/O time based on the amount of change in this value over a given period of time, divided by the total length of that period.

## Disk IO Rate
image
### What It Means
This metric indicates the disk I/O rate, and we trace the read rate and write rate for each disk separately.
### How To Get This Metric
Field _rd_sectors_ and _wr_sectors_ in _/proc/diskstats_ mean the number of sectors that the disk reads or writes. The size of each sector in the disk is 512 bytes, we calculate the i/o rate by multiplying the amount of change in this value by 512 bytes.

## Disk IOPS
image

### What It Means
This metric indicates how many disk i/o command is executed in per second. For HDD, the maximum iops is about 100-200. For SSD, the maximum iops is about 10000-20000.

### How To GetThis Metric
Field _rd_ios_ and _wr_ios_ mean the number of read or write operation executed. We calculate the iops by the change of these value.

## Disk IO Time
image
### What It Means
This metric indicates how much time all read or write operations take. Because it includes the time that the i/o request is waitting in the queue, so the value of it may be greater than 1s.
### How To Get It
Field _rd_ticks_ and _wr_ticks_ have the same meaning of this metric.

## Disk IO Time For Each Op
image
### What It Means
This metric indicates how much time a read or write operation takes on average, aka the latency of an io operation.

### How To Get It
We get it by the total waiting time over the operation counts. 

## Disk IO Size
image
### What It Means
This metric indicates the average size of an i/o operation.
### How To Get It
We get it by the total i/o size over the operation counts.

## Process I/O System Call
image
### What It Means
This metrics indicates the rate that the process calls the read system call or write system call.
### How To Get It
_/proc/[PID]/io_ stores some information of process i/o status, in which field _syscr_ and _syscw_ mean the calling count of read/write system call.

## Process IO Rate
image
### What It Means
This metrics indicates the i/o rate of the process.
## How To Get It
The field _rchar_ and _wchar_ mean the data size that the process has caused to read or write. The field _read_bytes_ and _write_bytes_ mean the data size that the process attempt to read or write.


SpriCoder (approved):
LGTM~

metrics/interface/src/main/java/org/apache/iotdb/metrics/metricsets/disk/LinuxDiskMetricsManager.java
```diff
@@ -0,0 +1,397 @@
...
+  private static final int DISK_WRITE_TIME_COST_OFFSET = 11;
+  private static final int DISK_QUEUE_SIZE_OFFSET = 12;
+  private static final int DISK_IO_TOTAL_TIME_OFFSET = 13;
+  private static final long UPDATE_SMALLEST_INTERVAL = 10000L;
+  private Set<String> diskIdSet;
+  private long lastUpdateTime = 0L;
+  private long updateInterval = 1L;
+
+  // Disk IO status structure
+  private final Map<String, Long> lastReadOperationCountForDisk = new HashMap<>();
```
OneSizeFitsQuorum, line 80: How about use `getDiskIds().size()` to be the initial capacity of these maps as `DEFAULT_INITIAL_CAPACITY` in HashMap is 16, which is usually bigger than the number of disks

metrics/interface/src/main/java/org/apache/iotdb/metrics/metricsets/disk/LinuxDiskMetricsManager.java
```diff
@@ -0,0 +1,397 @@
...
+  }
+
+  @Override
+  public Map<String, Long> getWriteCostTimeForDisk() {
+    return lastWriteTimeCostForDisk;
+  }
+
+  @Override
+  public Map<String, Long> getIoUtilsPercentage() {
+    Map<String, Long> utilsMap = new HashMap<>();
```
OneSizeFitsQuorum, line 158: init capacity with `incrementIoBusyTimeForDisk.size()`?

PR #9055, opened 2023-02-14
[IOTDB-5526] Fix NPE while construct timeseries black list in SchemaRegion

MarcosZyk:
## Description

When constructing timeseries black list in SchemaRegion, there won't be any schema template provided, which brings npe when trying to get child node from template. Add judgement before getting child node will fix this.


Cpaulyz (approved):
LGTM

PR #9058, opened 2023-02-14
[IOTDB-5531] Update IoT Consensus Metric from ms precision to ns precision

SpriCoder:
https://issues.apache.org/jira/browse/IOTDB-5531


OneSizeFitsQuorum (approved):
LGTM

PR #9064, opened 2023-02-15
[IOTDB-5112] Fixed IoTConsensus synchronization stuck under low load or during restart

BUAAserein:
### The log synchronization logic of the receiving end of IoTConsensus is judged by nextSyncIndex, which may cause a 10-second delay when the iotdb is initially started and under low load.

Solution: Set the nextSyncIndex when timeout.

Please see details in [IOTDB-5112](https://issues.apache.org/jira/browse/IOTDB-5112) and [feishu cloud doc](https://apache-iotdb.feishu.cn/docx/TekkdtRvTo181UxI2x0c42iTnsN).


OneSizeFitsQuorum (approved):
LGTM

PR #9074, opened 2023-02-16
[IOTDB-5512] Fixed IoTConsensus may repeatedly send some log when restarting

BUAAserein:
### The persistence of log synchronization is not timely, resulting in repeated sending of logs when restarting

Solution: Extend the persistence strategy of log synchronization

Please see details in [IOTDB-5112](https://issues.apache.org/jira/browse/IOTDB-5512) and [feishu cloud doc](https://apache-iotdb.feishu.cn/docx/DddIdK9bnojYbsxQHFbcmkx7nKc)

### DIFF
--- a/consensus/src/main/java/org/apache/iotdb/consensus/iot/logdispatcher/IndexController.java
+++ b/consensus/src/main/java/org/apache/iotdb/consensus/iot/logdispatcher/IndexController.java
@@ -70,7 +70,7 @@ public class IndexController {
     restore();
   }
 
-  public long updateAndGet(long index) {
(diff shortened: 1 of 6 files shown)


SpriCoder (approved):
Good job!

OneSizeFitsQuorum (commented):
Good Job~
BTW, you can persist the index before shutting down the node, and you can call the close method of the consensus layer in IoTDBShutDownHook and persist each group there

OneSizeFitsQuorum (approved):
LGTM

PR #9082, opened 2023-02-17
[To rel/0.13] Fix alias check when CreateAlignedTimeSeries

MarcosZyk:
## Description

Currently, IoTDB v0.13 doesn't support creating aligned timeseries with alias. Eliminate this field passed from client to server via SDK/API.


bigreybear (approved):
LGTM!

PR #9085, opened 2023-02-17
[IOTDB-5533] Make IoTDB Internal Reporter as default

SpriCoder:
In order to support the query of total write point, we need to make IoTDB Internal Reporter as default to collect the points of memtables.
See: IOTDB-5533


OneSizeFitsQuorum (approved):
LGTM

JackieTien97 (approved):

PR #9094, opened 2023-02-18

THUMarkLau:
First PR for this issue is https://github.com/apache/iotdb/pull/9054.

In the previous code, the disk sector size was set to a fixed 512 bytes length. In practice, however, some disks may have a sector size of 4096 bytes (such as some SSDS or large-capacity HDDS). Instead of setting the disk's sector size to a fixed 512 bytes, this PR gets the disk sector size at startup by looking at the /sys/ block/ [DEV]/queue/hw_sector_size file.


OneSizeFitsQuorum (approved):
LGTM

PR #9103, opened 2023-02-20
[IOTDB-5147]Optimize compaction schedule when priority is BALANCE

choubenson:
### The changes in this pr are as follows:
1. Change the size of the `candidateCompactionTaskQueue` from 1000 to 50. The advantage is to avoid tasks blocked in the queue occupying files for a long time, resulting in no files being selected for cross-space compaction. But another impact is that once the number of files is large, because the number of tasks in the queue shrinks, the inner-space compaction will not be able to compact the latest files.
2. Priority comparison of tasks: Inner-space compaction tasks no longer distinguish between seq and unseq, but are sorted according to unified rules; under the configuration of BALANCE, cross-space compaction tasks and inner-space compaction tasks follow the first-come-first-executed rule . One thing to note is that since the priority sorting of the queue is implemented through the minimum heap sorting, and each task is not in partial order, the actual result is not strictly that the one with the smaller seriesID will be executed first.


server/src/main/java/org/apache/iotdb/db/engine/compaction/selector/impl/SizeTieredCompactionSelector.java
```diff
@@ -142,12 +147,25 @@ private boolean selectLevelTask(
     // if next time partition exists
     // submit a merge task even it does not meet the requirement for file num or file size
     if (hasNextTimePartition && selectedFileList.size() > 1) {
-      taskPriorityQueue.add(new Pair<>(new ArrayList<>(selectedFileList), selectedFileSize));
+      addOneTaskToQueue(taskPriorityQueue, selectedFileList, selectedFileSize);
```
THUMarkLau, line 150: The return value is not used.

server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
```diff
@@ -488,6 +488,9 @@ public class IoTDBConfig {
 
   private boolean enableCompactionValidation = true;
 
+  /** The number of candidate compaction task queue. */
+  private int candidateCompactionTaskQueueSize = 50;
```
THUMarkLau, line 492: This param is appeared in IoTDBConfig but not in iotdb-common.properties.

server/src/main/java/org/apache/iotdb/db/engine/compaction/selector/impl/SizeTieredCompactionSelector.java
choubenson: It is unnecessary to use return value of `addOneTaskToQueue` here, because no matter what it is, `selectLevelTask` method will return false here.

server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
choubenson: Resolved.

node-commons/src/assembly/resources/conf/iotdb-common.properties
```diff
@@ -724,7 +728,7 @@ cluster_name=defaultCluster
 
 # The maximum number of data points in a page, default 1024*1024
```
THUMarkLau, line 729: This line of comment should be edited too.

THUMarkLau (changes requested):

node-commons/src/assembly/resources/conf/iotdb-common.properties
choubenson: Resolved

xingtanzjr (approved):

xingtanzjr:
Let's complete the verification of BALANCE for inner compaction between seq and unseq by Benchmark with 50% out-of-order data

PR #9108, opened 2023-02-21
[IOTDB-5516] Try sync delete schema region when dropping database

MarcosZyk:
## Description

After executing ```drop database``` command, we want the resources of dropped database been cleared as much as possible.

This pr implements sync deleting schema region group on DataNode to release memory resources.


CRZbulabula (approved):
LGTM～

PR #9111, opened 2023-02-21
[IOTDB-5512] Fixed same flushIndex situation

BUAAserein:
### Fixed same flushIndex situation which will cause a move file exception.

Please see details in https://github.com/apache/iotdb/pull/9074


OneSizeFitsQuorum (approved):
LGTM

PR #9107, opened 2023-02-20
[IOTDB-5562] Change the data type of AutoGuage from long to double in metric module

OneSizeFitsQuorum:
see [jira5562](https://issues.apache.org/jira/browse/IOTDB-5562)

PR #9113, opened 2023-02-21

THUMarkLau:
Since https://github.com/apache/iotdb/pull/9107 changes the Gauge data type from long to double, so we do not need to multiply the disk monitoring indicators to reduce the accuracy loss.


OneSizeFitsQuorum (approved):
LGTM

CRZbulabula (approved):
LGTM～

PR #9121, opened 2023-02-22
fix return success when insert exception

dh-cloud:
## Description

fix bug when insert error, but can't catch exception, and return success, this scenario may mislead data loss

<hr>

This PR has:
- [x] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


HTHou:
Hi, do you have any test cases to reproduce this bug?

dh-cloud:
> Hi, do you have any test cases to reproduce this bug?

client:
int res[] = statement.executeBatch()
if (res[0] == 200)
{count++;}

client continue insert value , then one node poweroff (or breakpoint), the insert count is inconsistent with query from iotdb

PR #9137, opened 2023-02-24
[IOTDB-5581] Missing reset of aggregators in RawDataAggregationOperator

ycycse:
Current framework will reset aggretors when a new window is initialized.

However, if there is no more window will be initialized, the infomation in aggregator would be appended twice because it wasn't reset after last appending.


ycycse:
PTAL @JackieTien97

JackieTien97 (approved):

PR #9140, opened 2023-02-24
[IOTDB-5582] Add region type in ratis metric prefix

BUAAserein:
### The current ratis metric prefix cannot distinguish metric items from different region types.

Solution: Add a region type to the metric prefix.

Please see details in [IOTDB-5582](https://issues.apache.org/jira/browse/IOTDB-5582).


consensus/src/main/java/org/apache/iotdb/consensus/ratis/metrics/RatisMetricSet.java
```diff
@@ -18,21 +18,36 @@
...
+
+  public void setConsensusGroupType(TConsensusGroupType consensusGroupType) {
+    if (consensusGroupType == TConsensusGroupType.ConfigNodeRegion) {
+      this.consensusGroupType = "confignode";
+    } else if (consensusGroupType == TConsensusGroupType.DataRegion) {
+      this.consensusGroupType = "data_region";
+    } else {
+      this.consensusGroupType = "schema_region";
+    }
+  }
```
SpriCoder, line 52: May be use enum class can be a good idea

SpriCoder (changes requested):
Good Job~ PTAL

consensus/src/main/java/org/apache/iotdb/consensus/ratis/metrics/RatisMetricSet.java
SzyWilliam: We can use `consensusGroupType.toString()` here

consensus/src/main/java/org/apache/iotdb/consensus/ratis/metrics/RatisMetricSet.java
```diff
@@ -18,21 +18,36 @@
...
 
+import org.apache.iotdb.common.rpc.thrift.TConsensusGroupType;
 import org.apache.iotdb.metrics.AbstractMetricService;
 import org.apache.iotdb.metrics.metricsets.IMetricSet;
 
 import org.apache.ratis.metrics.MetricRegistries;
 
 public class RatisMetricSet implements IMetricSet {
   private MetricRegistries manager;
+  private String consensusGroupType;
```
SzyWilliam, line 29: We can pass the consensusGroupType as a constructor param, instead of setting it manually.

SzyWilliam (commented):
Thanks a lot for @BUAAserein for working on this! Just some small comments inlined.

consensus/src/main/java/org/apache/iotdb/consensus/ratis/metrics/RatisMetricSet.java
BUAAserein: Already changed.

consensus/src/main/java/org/apache/iotdb/consensus/ratis/metrics/RatisMetricSet.java
BUAAserein: Already changed.

SpriCoder (approved):
LGTM~

OneSizeFitsQuorum (approved):
LGTM for now~
I will be doing a refactoring of the consensus layer API recently, which may change here.

PR #9141, opened 2023-02-26
[IOTDB-5563] Extract and decouple the logic of window segmentation in Aggregator

ycycse:
design doc: https://apache-iotdb.feishu.cn/docx/Wx6fdSc3qoSAptxwgiKcKdd0ncf


server/src/main/java/org/apache/iotdb/db/mpp/aggregation/slidingwindow/SlidingWindowAggregator.java
```diff
@@ -54,7 +54,7 @@ protected SlidingWindowAggregator(
   }
 
   @Override
```
liuminghui233, line 56: consider not letting SlidingWindowAggregator extend Aggregator, maybe the code will be clearer

liuminghui233 (approved):
LGTM :-)

server/src/main/java/org/apache/iotdb/db/mpp/aggregation/slidingwindow/SlidingWindowAggregator.java
ycycse: If SlidingWindowAggregator doesn't extends Aggregator, there will be more common codes appear in SlidingWindowAggregator (and SlidingWindowAggregationOperator needs to be modified, too). So I just choose to delete the sliding-window-method in Aggregator and add a type cast in SlidingWindowAggregationOperator now.

PR #9146, opened 2023-02-27
[IOTDB-5584] Fix wrong intialization in GROUP BY CONDITION

ycycse:
https://issues.apache.org/jira/browse/IOTDB-5584

When the controlColumn of first data point is false, conditon window wouldn't skip it because needSkip is intialized to false which leads to a redundant row in resultSet.

This PR fix the problem.


server/src/main/java/org/apache/iotdb/db/mpp/execution/operator/window/SeriesWindowManager.java
```diff
@@ -49,7 +49,7 @@ public class SeriesWindowManager implements IWindowManager {
 
   public SeriesWindowManager(SeriesWindowParameter seriesWindowParameter) {
     this.seriesWindow = new SeriesWindow(seriesWindowParameter);
-    this.needSkip = false;
+    this.needSkip = true;
```
JackieTien97, line 52: add some java doc about `needSkip` field

JackieTien97 (changes requested):

PR #9148, opened 2023-02-27
[IOTDB-5587] Add dataRegionId into Memtable and update flush point metrics

SpriCoder:
Add regionId into points


OneSizeFitsQuorum (approved):
LGTM

PR #9151, opened 2023-02-27
[IOTDB-5516] Fix DatabaseNotExistsException caused by pre deleted database during drop database

MarcosZyk:
## Description

During dropping database, the target database will be marked as pre-deleted in PartitionInfo, which brings DatabaseNotExistsException during adjustMaxRegionGroupNum.

Filter and ignore the pre-deleted database during adjustMaxRegionGroupNum.


CRZbulabula (approved):
LGTM～

PR #9157, opened 2023-02-28
[IOTDB-5592]Fix unexpected error when use full Path in having/where

Wei-hao-Li:
cause: error use of code about 'inWithoutNull'
image


liuminghui233 (approved):
LGTM :-)

PR #9167, opened 2023-02-28
Add write metrics and dashboard

shuwenwei:
## Description
Add some metrics about writing.
https://github.com/shuwenwei/iotdb/blob/write-metrics/docs/UserGuide/Monitor-Alert/Metric-Tool.md#4217-write-performance


server/src/main/java/org/apache/iotdb/db/wal/buffer/WALBuffer.java
```diff
@@ -423,14 +431,14 @@ public SyncBufferTask(
 
     @Override
     public void run() {
+      long start = System.nanoTime();
       currentWALFileWriter.updateFileStatus(fileStatus);
 
-      if (logger.isDebugEnabled()) {
```
xingtanzjr, line 428: Please verify whether it is ok for efficiency if this check is removed

xingtanzjr: Confirmed with Haiming, it is ok

xingtanzjr (approved):

PR #9170, opened 2023-02-28
[IOTDB-5516] Fix delete schema region bug during drop database

MarcosZyk:
## Description

When prepare the requests for deleting schema regions, the origin implementation only deletes one replica of one raft group.

Fix this to delete all replicas.


CRZbulabula (approved):
LGTM～

PR #9175, opened 2023-02-28
[IOTDB-5598] Pipe Plugins Management: from SQL to CN

MiniSho:
we provide the interfaces to support managing Pipe Plugins:
1.  `CREATE PIPEPLUGIN <PLUGIN-NAME> AS <FULL CLASS NAME> USING <URI>` 
2. `DROP PIPEPLUGIN <PLUGIN-NAME>`
3. `SHOW PIPEPLUGINS`

And the implements will be supported in next pr.


SteveYurongSu (approved):
Thanks for your pretty work! I have pushed some commits to your branch, please take a look. Now everything looks good to me :)

PR #9188, opened 2023-03-01
 Use ConfigurableTByteBuffer instead of TByteBuffer in ThriftCommonsSerDeUtils

lancelly:
It is possible that we need to deserialize a quite large bytebuffer whose capacity reaches the MaxMessageSize of thirft.

### DIFF
--- a/node-commons/src/main/java/org/apache/iotdb/commons/utils/ThriftCommonsSerDeUtils.java
+++ b/node-commons/src/main/java/org/apache/iotdb/commons/utils/ThriftCommonsSerDeUtils.java
@@ -43,6 +43,8 @@ import java.nio.ByteBuffer;
 /** Utils for serialize and deserialize all the data struct defined by thrift-commons */
 public class ThriftCommonsSerDeUtils {
 
+  private static final long THRIFT_DEFAULT_MAX_MESSAGE_SIZE = 104857600;
+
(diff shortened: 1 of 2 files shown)


node-commons/src/main/java/org/apache/iotdb/commons/utils/ThriftCommonsSerDeUtils.java
```diff
@@ -278,4 +280,13 @@ public static TSchemaNode deserializeTSchemaNode(ByteBuffer buffer) {
     }
     return schemaNode;
   }
+
+  private static TByteBuffer generateTByteBuffer(ByteBuffer buffer) throws TTransportException {
+    // TByteBuffer uses the default TConfiguration which limits the MaxMessageSize
+    // we use ByteBuffer.slice() to update the origin buffer if buffer.capacity() >= MaxMessageSize
+    if (buffer.capacity() >= THRIFT_DEFAULT_MAX_MESSAGE_SIZE) {
+      buffer.slice();
```
OneSizeFitsQuorum, line 288: It looks like this function doesn't do anything?
image

JackieTien97: update the  capacity

JackieTien97 (approved):

PR #9218, opened 2023-03-06
[IOTDB-5620] Fix flush stuck when there is a lot of time partitions in each DataRegion

HTHou:
## Description

When set Max_heap_size=2G and insert data of 7 years. The flush thread will be stuck. 

There is a simple way to reproduce. 

Set max_heap_size of data node to 200M, set write_memory_proportion to 10000000:1.

Then start standalone iotdb and execute the following sqls:

```sql
insert into root.sg.d1(time,s1) values(1,2);
flush;
insert into root.sg.d1(time,s1) values(2,2);
insert into root.sg.d1(time,s1) values(604800001,2);
flush;
```
Finally, you will see the last flush cannot execute successfully.

## Reason

The `evictOldPartition` method in `TimePartitionManager` always peek the first TimePartitionInfo in the TreeSet in the while loop. However, the memSize of the TimePartitionInfo may be 0. In this case, the loop won't be ended.

## Solution

Change the `treeSet.first()` to `treeSet.pollFirst()`.

### DIFF
--- a/integration-test/src/test/java/org/apache/iotdb/db/it/IoTDBInsertMultiPartitionIT.java
+++ b/integration-test/src/test/java/org/apache/iotdb/db/it/IoTDBInsertMultiPartitionIT.java
@@ -1,0 +1,71 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iotdb.db.it;
+
+import org.apache.iotdb.it.env.EnvFactory;
+import org.apache.iotdb.it.framework.IoTDBTestRunner;
+import org.apache.iotdb.itbase.category.ClusterIT;
+import org.apache.iotdb.itbase.category.LocalStandaloneIT;
+import org.apache.iotdb.itbase.category.RemoteIT;
+
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+
+import java.sql.Connection;
+import java.sql.Statement;
+
+import static org.junit.Assert.fail;
+
+@RunWith(IoTDBTestRunner.class)
+@Category({LocalStandaloneIT.class, ClusterIT.class, RemoteIT.class})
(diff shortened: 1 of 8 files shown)


integration-test/src/test/java/org/apache/iotdb/db/it/IoTDBInsertMultiPartitionIT.java
```diff
@@ -0,0 +1,71 @@
...
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+
+import java.sql.Connection;
+import java.sql.Statement;
+
+import static org.junit.Assert.fail;
+
+@RunWith(IoTDBTestRunner.class)
+@Category({LocalStandaloneIT.class, ClusterIT.class, RemoteIT.class})
```
JackieTien97, line 40:
```suggestion
@Category({LocalStandaloneIT.class, ClusterIT.class})
```

server/src/main/java/org/apache/iotdb/db/conf/IoTDBDescriptor.java
```diff
@@ -1606,13 +1606,14 @@ private void initStorageEngineAllocate(Properties properties) {
 
     String allocationRatioForWrite = properties.getProperty("write_memory_proportion", "19:1");
     proportions = allocationRatioForWrite.split(":");
-    int proportionForMemTable = Integer.parseInt(proportions[0].trim());
+    int proportionForMemTable = Integer.parseInt(proportions[0].replace("\\", "").trim());
     int proportionForTimePartitionInfo = Integer.parseInt(proportions[1].trim());
+    logger.info("proportionForTimePartitionInfo = {}", proportionForTimePartitionInfo);
```
JackieTien97, line 1611:
```suggestion
```

server/src/main/java/org/apache/iotdb/db/conf/IoTDBDescriptor.java
```diff
@@ -1606,13 +1606,14 @@ private void initStorageEngineAllocate(Properties properties) {
 
     String allocationRatioForWrite = properties.getProperty("write_memory_proportion", "19:1");
     proportions = allocationRatioForWrite.split(":");
-    int proportionForMemTable = Integer.parseInt(proportions[0].trim());
+    int proportionForMemTable = Integer.parseInt(proportions[0].replace("\\", "").trim());
```
JackieTien97, line 1609: replace("\\", "") is for what?

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/conf/IoTDBDescriptor.java
HTHou: image

The serialize method of property will add `\` for `:`.

integration-test/src/test/java/org/apache/iotdb/db/it/IoTDBInsertMultiPartitionIT.java
HTHou: `RemoteIT.class` is also in other IT cases. It is never used in actions.

server/src/main/java/org/apache/iotdb/db/conf/IoTDBDescriptor.java
```diff
@@ -1606,13 +1606,13 @@ private void initStorageEngineAllocate(Properties properties) {
...
```
HTHou, line 1609:
```suggestion
    int proportionForMemTable = Integer.parseInt(proportions[0].trim());
```

PR #9222, opened 2023-03-06
[IOTDB-5646] Support insert bytes directly by Python API

HTHou:
Java Client supports insert Binary form bytes with TsDataType.TEXT. Should support it in Python API.


SteveYurongSu (approved):
LGTM

PR #9226, opened 2023-03-06
[IOTDB-5630] Make Function cast a built-in function

lancelly:
Cast is a UDTF function before this pr, the user doc can be found at:
https://iotdb.apache.org/zh/UserGuide/Master/Operators-Functions/Conversion.html

### DIFF
--- a/integration-test/src/test/java/org/apache/iotdb/db/it/builtinfunction/scalar/IoTDBCastFunctionIT.java
+++ b/integration-test/src/test/java/org/apache/iotdb/db/it/builtinfunction/scalar/IoTDBCastFunctionIT.java
@@ -1,0 +1,33 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iotdb.db.it.builtinfunction.scalar;
+
+public class IoTDBCastFunctionIT {
+
+    // test without target type
+
+    // test old transformer
+        // test int -> Binary
+
+    // test column transformer
+        // test int -> Binary
+
+    // test nested
+}
--- a/integration-test/src/test/java/org/apache/iotdb/db/it/builtinfunction/scalar/IoTDBCastFunctionIT.java
+++ b/integration-test/src/test/java/org/apache/iotdb/db/it/builtinfunction/scalar/IoTDBCastFunctionIT.java
@@ -19,15 +19,1083 @@
 
 package org.apache.iotdb.db.it.builtinfunction.scalar;
 
+import org.apache.iotdb.it.env.EnvFactory;
+import org.apache.iotdb.it.framework.IoTDBTestRunner;
+import org.apache.iotdb.itbase.category.ClusterIT;
+import org.apache.iotdb.itbase.category.LocalStandaloneIT;
+
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.sql.Statement;
+
+import static org.apache.iotdb.db.it.utils.TestUtils.prepareData;
+import static org.apache.iotdb.db.it.utils.TestUtils.resultSetEqualTest;
+import static org.apache.iotdb.itbase.constant.TestConstant.TIMESTAMP_STR;
+import static org.junit.Assert.fail;
+
+@RunWith(IoTDBTestRunner.class)
+@Category({LocalStandaloneIT.class, ClusterIT.class})
 public class IoTDBCastFunctionIT {
 
-    // test without target type
+  // test without target type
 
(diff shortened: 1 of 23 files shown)


server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/ExpressionAnalyzer.java
```diff
@@ -1324,8 +1324,8 @@ public static boolean isDeviceViewNeedSpecialProcess(Expression expression) {
     } else if (expression instanceof UnaryExpression) {
       return isDeviceViewNeedSpecialProcess(((UnaryExpression) expression).getExpression());
     } else if (expression instanceof FunctionExpression) {
-      if (((FunctionExpression) expression).isBuiltInFunction()
-          && BuiltinFunction.DEVICE_VIEW_SPECIAL_PROCESS_FUNCTIONS.contains(
+      if (((FunctionExpression) expression).isBuiltInScalarFunction()
```
JackieTien97, line 1327: Previously, we named it as BuiltInScalarFunction, but I've forgot why Weihao Li changed it, you need to discuss with him @Wei-hao-Li

server/src/main/java/org/apache/iotdb/db/mpp/transformation/dag/column/unary/scalar/CastFunctionColumnTransformer.java
```diff
@@ -0,0 +1,227 @@
...
+package org.apache.iotdb.db.mpp.transformation.dag.column.unary.scalar;
+
+import org.apache.iotdb.db.mpp.transformation.dag.column.ColumnTransformer;
+import org.apache.iotdb.db.mpp.transformation.dag.column.unary.UnaryColumnTransformer;
+import org.apache.iotdb.tsfile.read.common.block.column.Column;
+import org.apache.iotdb.tsfile.read.common.block.column.ColumnBuilder;
+import org.apache.iotdb.tsfile.read.common.type.Type;
+import org.apache.iotdb.tsfile.utils.Binary;
+
+public class CastFunctionColumnTransformer extends UnaryColumnTransformer {
```
JackieTien97, line 29: use freemarker to gen the code for different data types combinations.

server/src/main/java/org/apache/iotdb/db/mpp/transformation/dag/column/unary/scalar/CastFunctionColumnTransformer.java
```diff
@@ -0,0 +1,227 @@
...
+        throw new UnsupportedOperationException(
+            String.format("Unsupported target dataType: %s", returnType.getTypeEnum()));
+    }
+  }
+
+  private void cast(ColumnBuilder columnBuilder, Binary value) {
+    String stringValue = value.getStringValue();
+    switch (returnType.getTypeEnum()) {
+      case INT32:
+        returnType.writeInt(columnBuilder, (int) Double.parseDouble(stringValue));
```
JackieTien97, line 204: I found that if we directly use Integer.parseInt for string value `1.1`, it will throw exception, so I wonder that how pg or mysql(standard sql) deal with this case?

JackieTien97 (changes requested):

JackieTien97 (changes requested):
Discussed with @qiaojialin , we decided to use `cast(XX as DataType)` like mysql and pg, the legacy UDF syntax can be removed totally.

JackieTien97 (changes requested):
The DataType cast can be seen in https://apache-iotdb.feishu.cn/docx/ESuHdezVOo6TehxXYUbcCfKBnmb

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/ExpressionAnalyzer.java
lancelly: we decided to use BuiltinScalar.

integration-test/src/test/java/org/apache/iotdb/db/it/builtinfunction/scalar/IoTDBCastFunctionIT.java
```diff
@@ -0,0 +1,1101 @@
...
+public class IoTDBCastFunctionIT {
+
+  // test without target type
+
+  // test old transformer
+  // test int -> Binary
+
+  // test column transformer
+  // test int -> Binary
+
```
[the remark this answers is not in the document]
lancelly: done.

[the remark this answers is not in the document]
lancelly: It is a notice for myself hh. I've added cases including each type as the source type and each type as the target type. Special cases are added, too.

server/src/main/java/org/apache/iotdb/db/mpp/transformation/dag/column/unary/scalar/CastFunctionColumnTransformer.java
lancelly: > The DataType cast can be seen in https://apache-iotdb.feishu.cn/docx/ESuHdezVOo6TehxXYUbcCfKBnmb

Compared with that of PostgreSQL. Now the cast function should behave as that of PostgreSQL.

lancelly:
> Discussed with @qiaojialin , we decided to use `cast(XX as DataType)` like mysql and pg, the legacy UDF syntax can be removed totally.

Done.

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/mpp/plan/expression/multi/builtin/BuiltInScalarFunctionHelper.java
```diff
@@ -0,0 +1,74 @@
...
+import org.apache.iotdb.tsfile.file.metadata.enums.TSDataType;
+
+/**
+ * This interface defines the methods that FunctionExpression may use if it is a FunctionExpression
+ * representing a built-in function
+ */
+public interface BuiltInScalarFunctionHelper extends BuiltInFunctionHelper {
+  /**
+   * Check if the input TsDataType is correct. For example, function DIFF only supports numeric
+   * input data. Throw {@link SemanticException} if the type is not correct.
```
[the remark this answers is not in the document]
lancelly: done.

PR #9234, opened 2023-03-07
[To rel/1.1] [IOTDB-5487] Fix the problem that the timestamp is "null" when using jdbc

RYH61:
see [[IOTDB-3936](https://issues.apache.org/jira/browse/IOTDB-3936)]Add an interface in IClientRPCService to directly return bytebuffer instead of TSQueryDataSet

This pr is modified to add lastReadWasNull = false; of the constructOneRow method to the constructOneTsBlock method, resulting in a call to the next() method that meets the first of the following criteria, The lastReadWasNull field was not reset to false.

image


neuyilan (approved):
LGTM

wangchao316:
hi, This pr  nned submit master branch.

wangchao316 (approved):
LGTM

PR #9240, opened 2023-03-08
Set some runtime fields in FIConetxt to null when the FI is done

JackieTien97:
We will keep FIContext in memory for 5 mins, so we only keep necessary infos in FIContext for memory cost.


Alima777 (approved):

wangchao316 (approved):
LGTM

PR #9242, opened 2023-03-08
[IOTDB-5644] Fix unexpected result when there are no select expressions after analyzed in query

Wei-hao-Li:
cause: no select expressions after analyzed, but was planned because  of expressions  in `where/having/groupBy`

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/AnalyzeVisitor.java
+++ b/server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/AnalyzeVisitor.java
@@ -496,6 +496,7 @@ public class AnalyzeVisitor extends StatementVisitor<Analysis, MPPQueryContext> 
     ColumnPaginationController paginationController =
         new ColumnPaginationController(
             queryStatement.getSeriesLimit(), queryStatement.getSeriesOffset(), false);
+    Set<PartialPath> noMeasurementDevices = new HashSet<>(deviceSet);
 
     for (ResultColumn resultColumn : queryStatement.getSelectComponent().getResultColumns()) {
       Expression selectExpression = resultColumn.getExpression();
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/AnalyzeVisitor.java
```diff
@@ -303,6 +307,9 @@ public Analysis visitQuery(QueryStatement queryStatement, MPPQueryContext contex
       } else {
         Map<Integer, List<Pair<Expression, String>>> outputExpressionMap =
             analyzeSelect(analysis, queryStatement, schemaTree);
+        if (outputExpressionMap.values().stream().allMatch(List::isEmpty)) {
+          return finishQuery(analysis);
+        }
```
liuminghui233, line 312: It's better to check `outputExpressions` below. That way we don't need to traverse the outputExpressionMap twice.

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/AnalyzeVisitor.java
```diff
@@ -350,6 +357,11 @@ public Analysis visitQuery(QueryStatement queryStatement, MPPQueryContext contex
     return analysis;
   }
 
+  private Analysis finishQuery(Analysis analysis) {
+    analysis.setFinishQueryAfterAnalyze(true);
```
liuminghui233, line 361: We need to set empty headers in different formats.
```suggestion
if (queryStatement.isSelectInto()) {
          analysis.setRespDatasetHeader(
            DatasetHeaderFactory.getSelectIntoHeader(queryStatement.isAlignByDevice()));
        }
        if (queryStatement.isLastQuery()) {
       analysis.setRespDatasetHeader(DatasetHeaderFactory.getLastQueryHeader());
        }
    analysis.setFinishQueryAfterAnalyze(true);
```

liuminghui233 (changes requested):

liuminghui233 (approved):
LGTM :-)

PR #9247, opened 2023-03-08
[IOTDB-5240] Fix ConfigMTree snapshot deserialization while using template

MarcosZyk:
## Description

The origin implementation of ConfigMTree#deserialization doesn't consider the node beneath database node. Fix it.


Cpaulyz (approved):
LGTM

PR #9248, opened 2023-03-08
[To rel/1.1][IOTDB-5240] Fix ConfigMTree snapshot deserialization while using template

MarcosZyk:
## Description

same as [[IOTDB-5240] Fix ConfigMTree snapshot deserialization while using template](https://github.com/apache/iotdb/pull/9247)


Cpaulyz (approved):
LGTM

PR #9261, opened 2023-03-09
[IOTDB-5652] Acclerate delete schema via AsyncClient

MarcosZyk:
### Motivation

The current implementation of Deleting Schema, including deleting timeseries, deactivating template and unsetting template, do process the request on DataNodes without any concurrency, which brings low performance and may bring client time out.

### Modification

Refactor the implementation of deleting timeseries, deactivating template and unsetting template via new DataNodeRegionTaskExecutor which is a framework implementing parallel request dispatching and failure retry.


CRZbulabula (approved):
LGTM～

PR #9274, opened 2023-03-09
[IOTDB-5659] Fix dead lock condition in shutdown hook

SzyWilliam:
Current stop hook produces dead lock situations that prevent the process to terminate.

Ratis waits applyIndex to advance to commitIndex. However, it cannot be done when node is ReadOnly.


OneSizeFitsQuorum (approved):
LGTM~
As ugly as I think this is, we really don't have a better idea right now, and hope that someone in the future will save us

BUAAserein (approved):
LGTM~

PR #9243, opened 2023-03-08
[IOTDB-5674] Multi-Mode IMNode Management

Cpaulyz:
## Description

https://apache-iotdb.feishu.cn/docx/JZhudIX6Vob7kxx62t2cWeeLnbb

PR #9285, opened 2023-03-10
[IOTDB-5660] Update deprecated storage group name

Cpaulyz:
## Description

Extract some rename modification from https://github.com/apache/iotdb/pull/9243.


confignode/src/main/java/org/apache/iotdb/confignode/consensus/request/write/storagegroup/SetDataReplicationFactorPlan.java
```diff
@@ -29,22 +29,22 @@
 
 public class SetDataReplicationFactorPlan extends ConfigPhysicalPlan {
 
-  private String storageGroup;
+  private String Database;
```
MarcosZyk, line 32: rename to ```database```

MarcosZyk (approved):

PR #9306, opened 2023-03-13
[IOTDB-5519] Improve the concurrency control of DatabaseSchema

CRZbulabula:
The management of cluster DatabaseSchemas still have some concurrency control problems. This PR will improve the following problems:

1. Distinguish the Database management of ClusterSchemaManager and PartitionManager: the ClusterSchemaManager will focus on the validity of DatabaseSchema while the PartitionManager will focus on the existence of Database. More detaild, the ClusterSchemaManager provides interface `isDatabaseNameValid` to check whether a DatabaseName is already exists or is a prefix of another Database. And the PartitionManager provides interface `isDatabaseExist` to check whether a Database is deleted or pre-deleted.
2. Fix some interface bugs in ClusterSchemaManager and PartitionManager: these interface will return immediatly as long as a Database doens't exist. However, it's better to continue the process and print a warn log.
3. Improve log in DeleteDatabaseProcedure: the DeleteDatabaseProcedure now will print a info log when deletion success.
4. Fix invalidate cache bug: the invalidate cache process should invalidate PartitionCache before SchemaCache.


confignode/src/main/java/org/apache/iotdb/confignode/manager/ClusterSchemaManager.java
```diff
@@ -158,53 +158,41 @@ public TSStatus setDatabase(DatabaseSchemaPlan databaseSchemaPlan) {
...
+    if (!isDatabaseExist(databaseSchema.getName())) {
       // Reject if Database doesn't exist
       result = new TSStatus(TSStatusCode.DATABASE_NOT_EXIST.getStatusCode());
       result.setMessage(
-          "Failed to alter database. The Database "
-              + storageGroupSchema.getName()
-              + " doesn't exist.");
+          "Failed to alter database. The Database " + databaseSchema.getName() + " doesn't exist.");
       return result;
     }
```
MarcosZyk, line 169: I think the PartitionManager only takes care of whether a database occupies physical resources in cluster, rather than namely ```isDatabaseExist```. A database exists as only as it exists in clusterSchemaInfo and can be queried by ```show databases```.

MarcosZyk (commented):
PTAL.

MarcosZyk (approved):

PR #9322, opened 2023-03-14
[IOTDB-5662] Fix BufferedUnderflowException occurs in inner space compaction

THUMarkLau:
See [IOTDB-5662](https://issues.apache.org/jira/browse/IOTDB-5662).

The reason for this bug is that the encoding type of chunks that merged together is different with each other.


choubenson (approved):
Fix UT.

choubenson (approved):

PR #9323, opened 2023-03-14
[To rel/1.1] Fix dispatch result collection logic

MarcosZyk:
## Description

There was dispatch result collection logic in ```FragmentDispatcherImpl#dispatchWriteSync```, while it has been missed in ```FragmentDispatcherImpl#dispatchWriteAsync```, which may results in incomplete failure msg returned to client or unexpected failure during auto creating schema.


HeimingZ (approved):
LGTM

PR #9344, opened 2023-03-16
[IOTDB-5682] Fix the update of nodeId in Metrics

SpriCoder:
Fix the update of nodeId Metrics: we need to update nodeId after registry


confignode/src/main/java/org/apache/iotdb/confignode/service/ConfigNode.java
```diff
@@ -169,6 +174,7 @@ public void active() {
           ConfigNodeConstant.GLOBAL_NAME,
           CONF.getConfigNodeId(),
           CONF.getClusterName());
+      setUpMetricService();
```
OneSizeFitsQuorum, line 177: duplicated with line 152?

BUAAserein (approved):
LGTM~

OneSizeFitsQuorum (approved):
LGTM

PR #9353, opened 2023-03-16
[To rel/1.1][IOTDB-5684] Standardize log folder of ConfigNode's Simple consensus protocol

CRZbulabula:
Now the Simple consensus protocol of ConfigNode willl use the same folder as Ratis consensus protocol


OneSizeFitsQuorum (approved):
LGTM

MarcosZyk (approved):

PR #9354, opened 2023-03-16
[To rel/1.1][IOTDB-5685] Fix error msg of failing to create a timeseries on an existing path when ReadOnly state

MarcosZyk:
## Description

The error msg thrown from consensus layer was not correctly returned to client. Fix it.

```
IoTDB> set system to readonly
IoTDB> create timeseries root.db2.device2.s3 int32
Msg: 600: Fail to do non-query operations because system is read-only.
```


CRZbulabula (approved):
LGTM～

SzyWilliam (approved):
+1 the changes look good to me.

PR #9337, opened 2023-03-15
[IOTDB-5679] Implement model storage on MLNode

WenWeiTHU:
Add mlnode model storager 
* portable model format (torchscript)
* save/delete pytorch model with model configs
* lru cache for model load

Usage:
```
from iotdb.mlnode.algorithm.model_factory import create_forecast_model
from iotdb.mlnode.storage.model_storager import modelStorager

model, model_cfg = create_forecast_model(model_name='nbeats')
modelStorager.save_model(model, model_cfg, model_id: int, trial_id: int)
model, model_cfg = modelStorager.load_model(model_id: int, trial_id: int)
modelStorager.delete_trial(model_id: int, trial_id: int) #
modelStorager.delete_model(model_id: int) # delete all model with responding trials
```

PR #9365, opened 2023-03-17
[IOTDB-5696] Implement client to connect ConfigNode/DataNode

WenWeiTHU:


mlnode/iotdb/mlnode/client.py
```diff
@@ -15,12 +15,45 @@
...
+        self.data_node_address = data_node_address
+        self.config_node_address = config_node_address
+
+    def borrow_data_node_client(self):
+        return DataNodeClient(host=self.data_node_address.ip,
+                              port=self.data_node_address.port)
+
+    def borrow_config_node_client(self):
+        return DataNodeClient(host=self.config_node_address.ip,
+                              port=self.config_node_address.port)
```
ycycse, line 56: Why `DataNodeClient`?

mlnode/iotdb/mlnode/client.py
```diff
@@ -40,12 +73,117 @@ def __init__(self, host, port):
...
+        self.__port = port
+
+        transport = TTransport.TFramedTransport(
+            TSocket.TSocket(self.__host, self.__port)
+        )
+        if not transport.isOpen():
+            try:
+                transport.open()
+            except TTransport.TTransportException as e:
+                logger.exception("TTransportException!", exc_info=e)
```
ycycse, line 115: What will happen if `transport.open()` causes exception? Should we raise the exception?

mlnode/iotdb/mlnode/model_storage.py
ycycse, line 1: These codes already in the pr of storage should be removed or merge the master after https://github.com/apache/iotdb/pull/9337 is committed because they are duplicate.

mlnode/test/test_client.py
```diff
@@ -0,0 +1,25 @@
...
+# under the License.
+#
+
+
+from iotdb.mlnode.client import client_manager
+
+
+def test_client_manager():
+    assert client_manager.borrow_data_node_client()
+    assert client_manager.borrow_config_node_client()
```
ycycse, line 25: How will the test codes work if there isn' a working IoTDB? Maybe try IoTDBContainer.

ycycse (changes requested):

mlnode/test/test_client.py
WenWeiTHU: Currently, the namespace of mlnode and iotdb package is the same, making IoTDBContainer not avaible

mlnode/iotdb/mlnode/client.py
WenWeiTHU: we keep the same behavior like iotdb session: when the transport is not open, use the logger.exception

liuminghui233 (approved):

PR #9370, opened 2023-03-18
chore: remove unused methods

tisonkun:
Also, these methods don't override any one from superclasses and generate warnings.


tisonkun:
Closing...

It seems later version Ratis contains these methods. It can be my local env issue.

PR #9375, opened 2023-03-19
Filter unsatisfied TimePartition while preparing resource list

JackieTien97:
One dataregion may contain many time partition, previously, we need to scan all time partition's tsfiles even if the time partition in that dataregion doesn't contains the device's data.


wangchao316:
Good Job

PR #9376, opened 2023-03-19
[IOTDB-5700] Clean temporary files created by UDF query after it finishes

lancelly:
https://issues.apache.org/jira/browse/IOTDB-5700


server/src/main/java/org/apache/iotdb/db/mpp/transformation/dag/udf/UDTFContext.java
```diff
@@ -44,6 +45,8 @@ public void constructUdfExecutors(Expression[] outputExpressions) {
   }
 
   public void finalizeUDFExecutors(long queryId) {
+    // close and delete UDF temp files
+    TemporaryQueryDataFileService.getInstance().deregister(queryId);
```
SteveYurongSu, line 49: Wouldn't it be better to put this on the last line of the method?

SteveYurongSu (approved):

PR #9381, opened 2023-03-20
[IOTDB-5701] Optimize the management of performance overview related metrics

SpriCoder:
We merge manager and related metric sets to avoid the access of same concurrentHashmap in MetricManager


OneSizeFitsQuorum (approved):
LGTM

PR #9382, opened 2023-03-20
[IOTDB-5685] Fix error msg of failing to create a timeseries on an existing path when ReadOnly state

MarcosZyk:
## Description

same as https://github.com/apache/iotdb/pull/9354


wangchao316 (approved):
LGTM

PR #9386, opened 2023-03-20
[IOTDB-5705] Replace data_region_per_processor by data_region_per_data_node

CRZbulabula:
The parameter data_region_per_data_node is a empirical value from reall production environmennt. Replace data_region_per_processor by data_region_per_data_node will improve the efficiency of default cluster and simplify testing process.

### DIFF
--- a/confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeConfig.java
+++ b/confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeConfig.java
@@ -101,7 +101,7 @@ public class ConfigNodeConfig {
   private int defaultDataRegionGroupNumPerDatabase = 2;
 
   /** The maximum number of DataRegions expected to be managed by each DataNode. */
-  private double dataRegionPerProcessor = 1.0;
+  private double dataRegionPerDataNode = 5.0;
 
   /** RegionGroup allocate policy. */
   private RegionBalancer.RegionGroupAllocatePolicy regionGroupAllocatePolicy =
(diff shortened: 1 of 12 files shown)


confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeConfig.java
```diff
@@ -101,7 +101,7 @@ public class ConfigNodeConfig {
   private int defaultDataRegionGroupNumPerDatabase = 2;
 
   /** The maximum number of DataRegions expected to be managed by each DataNode. */
-  private double dataRegionPerProcessor = 1.0;
+  private double dataRegionPerDataNode = 5.0;
```
xingtanzjr, line 104: Change to 3 will be more safe

OneSizeFitsQuorum: image
We tested the write performance comparison of different number of regions (1/2/4/8/16/32) under a typical load on a fit machine (16core), and felt that it was not always optimal when the value was small

xingtanzjr (approved):

confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeConfig.java
xingtanzjr: That's right. Because our BM uses InsertTablet to do the test, which relies less on disk but more on CPU

PR #9392, opened 2023-03-21
[IOTDB-5707] Support Activate Template for Session API

MarcosZyk:
### Motivation

When using session API, there should be an interface to provide service like ```create timeseries of schema template ```.

### Modification

Add interface ```createTimeseriesOfSchemaTemplate``` to Session API, which support batch activating template on different device path.


bigreybear (approved):
LGTM~

Cpaulyz (approved):

PR #9409, opened 2023-03-22
[IOTDB-5714] fix some IoTDB C++ SDK bugs

jamber001:
https://issues.apache.org/jira/browse/IOTDB-5714

### DIFF
--- a/client-cpp/src/main/Session.cpp
+++ b/client-cpp/src/main/Session.cpp
@@ -20,8 +20,16 @@
 #include "Session.h"
 #include <algorithm>
 #include <memory>
+#include <time.h>
 
 using namespace std;
+
+/**
+* Timeout of query can be set by users.
+* A negative number means using the default configuration of server.
+* And value 0 will disable the function of query timeout.
+*/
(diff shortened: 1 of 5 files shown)


jamber001:
> hi @jamber001
> 
> Could you please explain why the server may crash when executing 'select s1, s1, s2 from root.**' and why TSFetchResultsReq may return incorrect data or server crash?
> 
> Additionally, could you add some unit tests for the new interfaces executeRawDataQuery? Thank you.

Please refer to the change of SessionDataSet::SessionDataSet(..) and SessionDataSet::hasNext()，
This bug's main reason is above 2 funcitons' logic is wrong. They do not understand the logic of TSFetchResultsReq and TSExecuteStatementReq, especially about the relation of columns, columnNameIndexMap，queryDataSet.
This bug will let C++ SDK read wrong position of queryDataSet or  TSQueryDataSet in some data cases, which cause core-dump or wrong data result.

I will check IT case for executeRawDataQuery.

LoopJump (approved):
This PR looks good to me. @jamber001 

hi @qiaojialin please check if it can be merged.

PR #9412, opened 2023-03-22
[IOTDB-5715] Improve the performance of query order by time desc

JackieTien97:
more about the issue can be seen in https://issues.apache.org/jira/browse/IOTDB-5715.
The reason it that we directly use `currentTime < currentEndTime` to break the loop, however we should use `currentTime > currentEndTime` in order by time desc.

Now the query speed pf `order by time desc` is the same as `order by time asc`.
image
image


liuminghui233 (approved):
LGTM :-)

PR #9415, opened 2023-03-22
disable multi value selection for disk dashboard

THUMarkLau:
Disable multi value selection for disk dashboard


OneSizeFitsQuorum (approved):
LGTM

PR #9417, opened 2023-03-22
[IOTDB-5716] Wrong dependency when pipeline consumeOneByOneOperator

Alima777:
For example, We have a plan
DeviceViewOperator

- AggregationOperator1
  - SeriesAggregationScanOperator1
  - ExchangeOperator1
- AggregationOperator2
  - SeriesAggregationScanOperator2
  - ExchangeOperator2

If dop  = 3, AggregationOperator1 and SeriesAggregationScanOperator1 will both be a pipeline. 

Howerver, due to the traverseorder, the first pipeline of this group will be the SeriesAggregationScanOperator1 rather than AggregationOperator1.

In before logic, I took the first pipeline as the dependencyPipeline.

But actually AggregationOperator2 and  SeriesAggregationScanOperator2 must have dependency on the parent pipeline, i.e. pipeline of AggregationOperator1.


lancelly (approved):
LGTM

PR #9435, opened 2023-03-23
[IOTDB-5717] Fix incorrect result when querying with limit push-downing & order by time desc

JackieTien97:
cp from https://github.com/apache/iotdb/pull/9434


liuminghui233 (approved):
LGTM :-)

PR #9436, opened 2023-03-23
[To rel/1.1] [IOTDB-5717] Fix incorrect result when querying with limit push-downing & order by time desc

liuminghui233:
see IOTDB-5717.

**Before:**
8344c7b407747922001574dc144832a

**After:**
ddbb763736675a3f42b7d980b021a44


liuminghui233:
see #9434

PR #9439, opened 2023-03-24
[IOTDB-5725] Make internal report recording measurements asynchronous

OneSizeFitsQuorum:
The flush time has been greatly reduced, and the system reject error has disappeared.
image

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/service/metrics/IoTDBInternalLocalReporter.java
+++ b/server/src/main/java/org/apache/iotdb/db/service/metrics/IoTDBInternalLocalReporter.java
@@ -150,39 +150,43 @@ public class IoTDBInternalLocalReporter extends IoTDBInternalReporter {
 
   @Override
   protected void writeMetricToIoTDB(Map<String, Object> valueMap, String prefix, long time) {
-    try {
-      TSInsertRecordReq request = new TSInsertRecordReq();
-      List<String> measurements = new ArrayList<>();
-      List<TSDataType> types = new ArrayList<>();
-      List<Object> values = new ArrayList<>();
-      for (Map.Entry<String, Object> entry : valueMap.entrySet()) {
-        String measurement = entry.getKey();
-        Object value = entry.getValue();
-        measurements.add(measurement);
-        types.add(inferType(value));
-        values.add(value);
-      }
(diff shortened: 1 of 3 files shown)


SpriCoder (approved):
LGTM~

server/src/main/java/org/apache/iotdb/db/service/metrics/IoTDBInternalLocalReporter.java
```diff
@@ -150,39 +150,43 @@ public ReporterType getReporterType() {
...
-      List<Object> values = new ArrayList<>();
-      for (Map.Entry<String, Object> entry : valueMap.entrySet()) {
-        String measurement = entry.getKey();
-        Object value = entry.getValue();
-        measurements.add(measurement);
-        types.add(inferType(value));
-        values.add(value);
-      }
-      ByteBuffer buffer = SessionUtils.getValueBuffer(types, values);
+    service.execute(
```
xingtanzjr, line 153: From the call stack, it seems that the `service.execute()` will be called inside the thread pool, which has potential risk leading dead lock for thread pool `service`

xingtanzjr: Yes, you're correct. Actually my point of previous comment is that there is `potential risk` in current implementation at the same time it won't lead to dead-lock. This is just a talk about whether submit a new task inside a thread pool to itself is a the best practice or not.

OneSizeFitsQuorum: > This is just a talk about whether submit a new task inside a thread pool to itself is a the best practice or not.

OneSizeFitsQuorum: > This is just a talk about whether submit a new task inside a thread pool to itself is a the best practice or not.
Yeah, I don't want to do that either. But if the cost of best practices is to introduce an extra thread, I'm inclined to start from a thread count reduction perspective for now (we want to reduce thread count in IoTDB within the next year).

> shall we add some limitation to it such as change the method name to something like xxxxxAsync.
Good advice. It's been changed.

xingtanzjr (approved):

PR #9444, opened 2023-03-24
[IOTDB-5727] Complete the pipe task logic at config node

Caideyipi:
## IOTDB-5727
1、Completed the pipe procedures at config node.
2、Deleted some useless and conflicting files.
3、PipeTaskMeta and corresponding classes.

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/sync/CreatePipeProcedure.java
```diff
@@ -18,38 +18,24 @@
...
-import java.util.ArrayList;
 import java.util.HashSet;
-import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 
-public class CreatePipeProcedure extends AbstractOperatePipeProcedure {
-  private static final Logger LOGGER = LoggerFactory.getLogger(CreatePipeProcedure.class);
+// Empty procedure for old sync, restored only for compatibility
+public class CreatePipeProcedure extends Procedure<ConfigNodeProcedureEnv> {
```
wangchao316, line 38: why extends Procedure?

Caideyipi: Because the old AbstractOperatePipeProcedure is already deleted... Just for simplicity, I think it is redundant because we don't need it to execute serialization or deserialization. Besides, extending procedure is a must for the old Procedure to deserialize.

PR #9452, opened 2023-03-25
[IOTDB-5729] Fix RatisConsensus ResponseMessage conversion error

SzyWilliam:
An error occurred during a recent run:
```shell
org.apache.iotdb.rpc.StatementExecutionException: 305: 
class org.apache.ratis.protocol.Message$1 cannot be cast to class org.apache.iotdb.consensus.ratis.ResponseMessage 
(org.apache.ratis.protocol.Message$1 and org.apache.iotdb.consensus.ratis.ResponseMessage are in unnamed module of loader 'app')
```

The error is originated from a incorrect return value in `applicationStateMachineProxy.write`, see the details in PR.


OneSizeFitsQuorum (approved):
LGTM

PR #9461, opened 2023-03-27
[To rel/1.1][IOTDB-5730] Fix use a hard link when loading a SchemaFile snapshot 

Cpaulyz:
## Description

Using a hard link when loading a SchemaFile snapshot on the current system can lead to corruption of snaptshot. Files.copy should be used instead.


wangchao316 (approved):
LGTM

PR #9476, opened 2023-03-29
[IOTDB-5683] Add special case process for aggregation function Mode

Wei-hao-Li:
case1: If all the elements have the same number of occurrences, that is no Mode, return the value with earliest time; 
case2: If there are many Modes, return the Mode with earliest time.


server/src/main/codegen/templates/ModeAccumulator.ftl
```diff
@@ -96,7 +104,7 @@ public class ${className} implements Accumulator {
 
     // Step of ModeAccumulator is STATIC,
     // countMap only need to record one entry which key is finalResult
-    countMap.put(finalResult.get${type.dataType?cap_first}(0), 0L);
+    countMap.put(finalResult.get${type.dataType?cap_first}(0), new Pair<>(0L, 0L));
```
JackieTien97, line 107:
```suggestion
    countMap.put(finalResult.get${type.dataType?cap_first}(0), new Pair<>(0L, Long.MIN_VALUE));
```

JackieTien97 (changes requested):

PR #9485, opened 2023-03-30
Fix WrappedSegment extension with big name on edge case

bigreybear:
## Description

(same as https://github.com/apache/iotdb/pull/9484)

In determining the size of the pre-allocated segment for an EntityNode, an accurate calculation is performed by examining the children of the EntityNode if it has fewer than 20 children. Alternatively, the size is estimated using a hard-coded step list based on the cardinality of the EntityNode's children.

During the flushing process of the EntityNode, the segment size is assessed prior to the insertion of each child, and the segment is re-allocated or extended according to the anticipated size. If the segment is unable to accommodate the inserted child even after an extension, an exception will be raised, indicating that the record is too large or the extension has failed.

This issue arises in an edge case where the EntityNode has fewer than 20 children, each with a large size (approximately 1000 bytes in total) when calculated for pre-allocation. However, before the EntityNode is flushed, a new child with a large name and a small buffer, which includes aliases and other schema data but measures under 20 bytes, for instance, is inserted. The segment size assessment may be misled, resulting in a minor increment for the segment extension, e.g., from 1000 bytes to 1022 bytes, leading to an overflow since the segment cannot accommodate the large name.

This hotfix takes into account the length of the name, bytes of the name, and the record buffer offset when assessing the expected size of the WrappedSegment. This issue will not occur with AliasIndexPage or InternalPage, as their lengths will never be assessed multiple times due to their fixed length. The root cause of the original issue, as well as potential performance improvements, can be attributed to the design of the WrappedSegment increment, which is determined by a hard-coded step list as aforementioned.

### DIFF
--- a/server/src/test/java/org/apache/iotdb/db/metadata/mtree/schemafile/SchemaFileTest.java
+++ b/server/src/test/java/org/apache/iotdb/db/metadata/mtree/schemafile/SchemaFileTest.java
@@ -505,6 +505,57 @@ public class SchemaFileTest {
     sf.close();
   }
 
+  /**
+   * This test case is designed to examine a scenario in which, every time a SegmentedPage inserts a
+   * record, the process of checking whether there is enough space in the segment takes into account
+   * not only the size of the record's buffer, but also the size of the key and related pointers.
+   *
+   * @throws Exception
+   */
+  @Test
+  public void testEstimateSegSizeWithBigName() throws Exception {
+    ICachedMNode sgNode = nodeFactory.createDatabaseMNode(null, "mma", 111111111L).getAsMNode();
+    ICachedMNode d1 =
+        fillChildren(
+            sgNode,
+            300,
+            "device中文abcdefahiklmnoparstuwwxyz1ABCDEFGHUKLMNOPORSTUVWXYZ",
+            this::supplyEntity);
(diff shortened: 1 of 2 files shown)


server/src/test/java/org/apache/iotdb/db/metadata/mtree/schemafile/SchemaFileTest.java
```diff
@@ -505,6 +505,57 @@ public void testUpdateOnFullPageSegment() throws MetadataException, IOException
...
+   * @throws Exception
+   */
+  @Test
+  public void testEstimateSegSizeWithBigName() throws Exception {
+    ICachedMNode sgNode = nodeFactory.createDatabaseMNode(null, "mma", 111111111L).getAsMNode();
+    ICachedMNode d1 =
+        fillChildren(
+            sgNode,
+            300,
+            "device中文abcdefahiklmnoparstuwwxyz1ABCDEFGHUKLMNOPORSTUVWXYZ",
```
wangchao316, line 522: Chinese characters are not recommended. Garbled characters are easy to be displayed.

bigreybear: removed now

PR #9495, opened 2023-03-31
[To rel/0.13][IOTDB-5726]Select the last sealed seq file for nonOverlap unseq files to compact in cross compaction

choubenson:
**Description**
Cross space compaction select one unseq file but none seq file, which causes cross compaction under this time partition can never be performed.

**Reason**
Unseq file does not overlap with any seq files under this time partition, including two cases: 1. The devices under the unseq files do not exist in the seq files. 2. The device of the unseq file exists in the seq file, but unseqDevice.startTime>seqDevice.endTime. This can happen if the data is deleted and an seq inner space compaction has been performed.

**Solution**
For nonOverlap unseq files, select the last sealed seq file, if it is valid (status is closed), then select it, otherwise stop the selection.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/compaction/cross/rewrite/selector/RewriteCompactionFileSelector.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/compaction/cross/rewrite/selector/RewriteCompactionFileSelector.java
@@ -191,7 +191,7 @@ public class RewriteCompactionFileSelector implements ICrossSpaceMergeFileSelect
       if (seqSelectedNum != resource.getSeqFiles().size()) {
         selectOverlappedSeqFiles(unseqFile);
       }
-      if (!checkUnseqFileHasOverlapOrNot(unseqFile)) {
+      if (!hasFoundSeqFiles(unseqFile)) {
         tmpSelectedSeqFiles.clear();
         break;
       }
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/apache/iotdb/db/engine/compaction/cross/rewrite/selector/RewriteCompactionFileSelector.java
```diff
@@ -189,6 +191,10 @@ void selectFiles() throws IOException {
       if (seqSelectedNum != resource.getSeqFiles().size()) {
         selectOverlappedSeqFiles(unseqFile);
       }
+      if (!hasFoundSeqFiles(unseqFile)) {
+        tmpSelectedSeqFiles.clear();
```
xingtanzjr, line 195: This line is unnecessary

server/src/main/java/org/apache/iotdb/db/engine/compaction/cross/rewrite/selector/RewriteCompactionFileSelector.java
```diff
@@ -229,6 +235,29 @@ void selectFiles() throws IOException {
...
+   * for it to compact with. Notice: If the data is deleted and then start an inner space
+   * compaction, it may cause unseqStartTime > seqEndTime or partial devices are in the unseq files
+   * but not in seq files, which will cause the unseq files to not overlap with any seq file
+   */
+  private boolean hasFoundSeqFiles(TsFileResource unseqFile) {
+    if (!tmpSelectedSeqFiles.isEmpty()) {
+      return true;
+    }
+    // the current unseq file does not overlap with any seq files
+    logger.info("Unseq file {} does not overlap with seq files.", unseqFile);
```
xingtanzjr, line 249: Suggest to make line 250-257 as a separate method such as `trySelectLatestSealedSeqFile()`. And then change the judgement above into `if (!hasFoundSeqFiles(unseqFile) && !trySelectLatestSealedSeqFile())`

Issue #5886, opened 2022-05-12
This class overrides "equals()" and should therefore also override "hashCode()".


MinaQin:
Code Path:server/src/main/java/org/apache/iotdb/db/metadata/mnode/container/MNodeContainers.java
<img width="839" alt="image" src="https://user-images.githubusercontent.com/55030443/168059561-63d55054-dbdf-4631-b6e2-448e11c6fefb.png">
According to the Java Language Specification, there is a contract between equals(Object) and hashCode():

If two objects are equal according to the equals(Object) method, then calling the hashCode method on each of the two objects must produce the same integer result.

It is not required that if two objects are unequal according to the equals(java.lang.Object) method, then calling the hashCode method on each of the two objects must produce distinct integer results.

However, the programmer should be aware that producing distinct integer results for unequal objects may improve the performance of hashtables.

In order to comply with this contract, those methods should be either both inherited, or both overridden.

selectbook:
We Open Star

azhsmesos:
please assigned it to me

GuanchuShen:
[WeOpen Star]I would like to help

PR #9496, opened 2023-03-31
[ISSUE-5886] Implement hashCode for MNodeContainers 

GYXkeep:
## Description
The parent class of this class is an array, so the hash code of the class can be represented by the hash code of the parent class。


GYXkeep:
https://github.com/apache/iotdb/issues/5886

PR #9503, opened 2023-04-03
[IOTDB-5691] The first level-0 tsfile too large to trigger "Unsequence InnerSpaceComapction" and "CrossSpaceCompaction"

choubenson:
The first unseq file is 0 level and it is too large to trigger unseq inner space compaction and cross space compaction.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/compaction/selector/impl/RewriteCrossSpaceCompactionSelector.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/compaction/selector/impl/RewriteCrossSpaceCompactionSelector.java
@@ -182,11 +182,12 @@ public class RewriteCrossSpaceCompactionSelector implements ICrossSpaceSelector 
       throws IOException {
     TsFileNameGenerator.TsFileName unseqFileName =
         TsFileNameGenerator.getTsFileName(unseqFile.getTsFile().getName());
-    // we add a hard limit for cross compaction that selected unseqFile should be compacted in inner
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/iotdb/db/engine/compaction/selector/impl/RewriteCrossSpaceCompactionSelector.java
```diff
@@ -185,7 +185,8 @@ private boolean canAddToTaskResource(
     // we add a hard limit for cross compaction that selected unseqFile should be compacted in inner
     // space at least once. This is used to make to improve the priority of inner compaction and
     // avoid too much cross compaction with small files.
-    if (unseqFileName.getInnerCompactionCnt() < config.getMinCrossCompactionUnseqFileLevel()) {
+    if (unseqFileName.getInnerCompactionCnt() < config.getMinCrossCompactionUnseqFileLevel()
```
xingtanzjr, line 188: Suggest to add more comments here.  Current comment is not aligned with the new implementation here

xingtanzjr (approved):

server/src/main/java/org/apache/iotdb/db/engine/compaction/selector/impl/RewriteCrossSpaceCompactionSelector.java
choubenson: Resolved.

PR #9517, opened 2023-04-04
[To rel/1.1] [ISSUE-9508] Fix the bug of UDF Median and Percentile

543202718:
The bug is caused by the service load failure of MutableListFactoryImpl, which is used by MutableList. Thus, MutableList is replaced by ArrayList to solve the problem.


donraab:
Hi, I just noticed this PR and associated issue. If any other ServiceLoader issues are encountered using Eclipse Collections, upgrading to Eclipse Collections 11.1 should remedy the issue. Using the `Lists` factory from the `org.eclipse.collections.impl.factory` package instead of `api` package may also address the issue if upgrading to 11.1 is not possible.

PR #9531, opened 2023-04-04
Fix a memory free bug and delete useless O(N^2) check in DriverContext

JackieTien97:
In MemoryPool, FragmentInsatnceId doesn't contain queryId.

And I delete the useless operator Id check in DriverContext, because it will cause O(N^2) time complexity which is a problem while having many PlanNodes.


Alima777 (approved):
LGTM

PR #9540, opened 2023-04-06
Fix Context release source NPE

JackieTien97:
For schema related query FI, `closedFilePaths` and `unClosedFilePaths` in FragmentInstanceContext are null, so while calling `context.releaseResource()` in statemachine callback, it will throw NPE and the remaining codes won't be executed like `deRegisterFragmentInstanceFromMemoryPool` and so on which will cause resource leaky.

And in this pr, I also add the error log while there is exception throwed in the release process.


Alima777 (approved):
LGTM

PR #9549, opened 2023-04-06
Fix Performance Overview Dashboard localScheduler record inaccurately

OneSizeFitsQuorum:
The elapsed time of the Local Scheduler may be recorded repeatedly on the Performance Overview panel. That is, a timer is recorded even when no instance is executed locally. As a result, the elapsed time may be inaccurate.


wangchao316 (approved):
LGTM

PR #9602, opened 2023-04-13
[To rel/1.1] Fix SchemaEngine memory control concurrent bug

MarcosZyk:
### Release-Task Lock Priority

We've implemented ```StampedWriterPreferredLock``` to enable memory release when a flush task is waiting for write lock. However, the priority is not set when release task try taking a read lock. Fix it.

### Memory-Control Boundary

Currently, the variable ```allowToCreateNewSeries``` is not maintained in a thread-safe manner. Use synchronized and double-check to protect it and ensure accuracy.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/rescon/MemSchemaEngineStatistics.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/rescon/MemSchemaEngineStatistics.java
@@ -46,6 +46,8 @@ public class MemSchemaEngineStatistics implements ISchemaEngineStatistics {
 
   private volatile boolean allowToCreateNewSeries = true;
 
+  private final Object allowToCreateNewSeriesLock = new Object();
+
   @Override
   public boolean isAllowToCreateNewSeries() {
     return allowToCreateNewSeries;
@@ -69,7 +71,7 @@ public class MemSchemaEngineStatistics implements ISchemaEngineStatistics {
   public void requestMemory(long size) {
     memoryUsage.addAndGet(size);
     if (memoryUsage.get() >= memoryCapacity) {
-      synchronized (this) {
+      synchronized (allowToCreateNewSeriesLock) {
         if (allowToCreateNewSeries && memoryUsage.get() >= memoryCapacity) {
           logger.warn("Current series memory {} is too large...", memoryUsage);
           allowToCreateNewSeries = false;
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/apache/iotdb/db/metadata/rescon/MemSchemaEngineStatistics.java
```diff
@@ -69,19 +69,27 @@ public long getMemoryUsage() {
   public void requestMemory(long size) {
     memoryUsage.addAndGet(size);
     if (memoryUsage.get() >= memoryCapacity) {
-      logger.warn("Current series memory {} is too large...", memoryUsage);
-      allowToCreateNewSeries = false;
+      synchronized (this) {
```
Cpaulyz, line 72: The same

server/src/main/java/org/apache/iotdb/db/metadata/rescon/MemSchemaEngineStatistics.java
```diff
@@ -69,19 +69,27 @@ public long getMemoryUsage() {
...
   public void releaseMemory(long size) {
     memoryUsage.addAndGet(-size);
-    if (!allowToCreateNewSeries && memoryUsage.get() < memoryCapacity) {
-      logger.info(
-          "Current series memory {} come back to normal level, total series number is {}.",
-          memoryUsage,
-          totalSeriesNumber);
-      allowToCreateNewSeries = true;
+    if (memoryUsage.get() < memoryCapacity) {
+      synchronized (this) {
```
Cpaulyz, line 84:
```suggestion
      synchronized (allowToCreateNewSeriesLock) {
```

It is better to use a new object rather than this

Cpaulyz (changes requested):

server/src/main/java/org/apache/iotdb/db/metadata/rescon/MemSchemaEngineStatistics.java
MarcosZyk: Fixed.

server/src/main/java/org/apache/iotdb/db/metadata/rescon/MemSchemaEngineStatistics.java
MarcosZyk: Fixed.

Cpaulyz (approved):
LGTM

PR #9607, opened 2023-04-13
distinguish whether an exception has occurred when releasing the QueryExecution resource

caozj1011:
The current situation where the resultHandle.abort() method is eventually executed when a client request completes, regardless of whether an exception occurred, is not accurate. The right way to release resources is to say,
1. To end a normal operation, run resultHandle.close();
2. To terminate the exception, run resultHandle.abort().


caozj1011 (commented):
review

JackieTien97 (approved):

PR #9609, opened 2023-04-13
[To rel/1.1] Accelerate count all schema via schema statistic

MarcosZyk:
## Description

When registering timeseries, we usually want to check current registered timeseries num via ```count timeseries```. 

Since there's a statistic value recording the total timeseries num, we can directly return such statistic value without traversing the MTree.


server/src/main/java/org/apache/iotdb/db/metadata/mtree/store/disk/cache/CacheMemoryManager.java
```diff
@@ -266,6 +266,7 @@ private void tryFlushVolatileNodes() {
         .join();
     if (engineMetric != null) {
       engineMetric.recordFlush(System.currentTimeMillis() - startTime);
+      logger.info("Schema flush takes {}ms", System.currentTimeMillis() - startTime);
```
Cpaulyz: remove it

Cpaulyz (changes requested):

PR #9621, opened 2023-04-16
[To rel/0.13] change default concurrent_sub_rawQuery_thread and concurrent_query_thread to 0

qiaojialin:
Previous:
concurrent_sub_rawQuery_thread is 8, we use the min(concurrent_sub_rawQuery_thread, cpu cores) as sub query thread.
concurrent_query_thread is 16, we use the min(concurrent_query_thread, cpu cores) as concurrent query thread.

Problem: When users use a machine with cpu cores larger than 8 and 16, the default query thread will be limited to 8 and 16.

Currently:
concurrent_sub_rawQuery_thread is 0, default is cpu cores, but users could change it, which could exceeds the cpu cores.
concurrent_query_thread is 0, default is cpu cores, but users could change it, which could exceeds the cpu cores.

If uses' cpu core is lower than 8, the upgrade has no effect.
If uses' cpu core is larger than 8, the concurrent_sub_rawQuery_thread will be larger(cpu cores).
If uses' cpu core is larger than 16, the concurrent_query_thread will be larger(cpu cores).


Alima777 (approved):
LGTM

server/src/main/java/org/apache/iotdb/db/query/pool/RawQueryReadTaskPoolManager.java
```diff
@@ -40,10 +40,7 @@ public class RawQueryReadTaskPoolManager extends AbstractPoolManager {
   private static final Logger LOGGER = LoggerFactory.getLogger(RawQueryReadTaskPoolManager.class);
 
   private RawQueryReadTaskPoolManager() {
-    int threadCnt =
-        Math.min(
-            Runtime.getRuntime().availableProcessors(),
-            IoTDBDescriptor.getInstance().getConfig().getConcurrentSubRawQueryThread());
+    int threadCnt = IoTDBDescriptor.getInstance().getConfig().getConcurrentSubRawQueryThread();
```
bigreybear, line 43: May lead to an IllegalArgumentException while initializing ThreadPoolExecutor run/debug within IDEA, but packaged jar works as expected.

PR #9625, opened 2023-04-17
[To rel/1.1][IOTDB-5762][Metric] remove session metrics when close

SpriCoder:
remove session metrics when close session


OneSizeFitsQuorum (approved):
LGTM

PR #9631, opened 2023-04-17
[To rel/1.1] Resolve quadratic complexity issue when flushing numerous Internal/Entity nodes in SchemaFile

bigreybear:
## Description
During a flush of `SchemaFile` for numerous Entity/Internal Nodes, a surge in dirty pages occurs due to the pre-allocation mechanism. Retrieval for a suitable `SegmentedPage` may require a full iteration if each of these pages has too little space spare to accommodate another segment. The quadratic time overhead may appear to be stucked while handling a massive number of child nodes, e.g., 10 million child nodes.

## Solution
An array of LinkedLists has been introduced to index dirty pages into a tiered structure based on their spare size, considering the minimum segment and other step sizes. When a `SegmentedPage` is marked as dirty, it will be integrated into one of the LinkedLists if its spare size is larger than the minimum segment.

## Further Improvement
Although the size of `pageInstCache` is constrained by the config file, and the complexity of retrieval from it will not surge like the dirty pages do, there is a performance pitfall if the application frequently writes small amounts of data after reading almost full pages. This will lead to an optimizable full iteration over the `pageInstCache`. This issue requires a refactor of the cache mechanism or a complete spare page index and should be addressed in future versions.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/mtree/store/disk/schemafile/SchemaFileConfig.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/mtree/store/disk/schemafile/SchemaFileConfig.java
@@ -68,7 +68,7 @@ public class SchemaFileConfig {
 
   // region Segment Configuration
 
-  public static final int SEG_HEADER_SIZE = 25; // in bytes
+  public static final int SEG_HEADER_SIZE = 25; // (bytes) minimum segment size de facto
   public static final short SEG_OFF_DIG =
       2; // length of short, which is the type of segment offset and index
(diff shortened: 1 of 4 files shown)


Cpaulyz:
After testing, time cost of flushing 11149000 devices is about 50s.

bigreybear:
> After testing, time cost of flushing 11149000 devices under the same db node is about 50s.

My SSD took around 3-4 seconds flushing 1million devices, seems like a linear relationship though. Any specific performance target? Further improvement may need larger refactor..

MarcosZyk (approved):
Nice work!

PR #9533, opened 2023-04-04
[IOTDB-5727] pipe task management at config node 

Caideyipi:
## IOTDB-5727
1、Completed the pipe procedures at config node.
2、Deleted some useless and conflicting files.
3、PipeTaskMeta and corresponding classes.

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
    - [ ] concurrent read and write 
- [ ] added documentation for new or modified features or behaviors.
- [ ] added Javadocs for most classes and all non-trivial methods. 
- [ ] added or updated version, __license__, or notice information
- [ ] added comments explaining the "why" and the intent of the code wherever would not be obvious 
  for an unfamiliar reader.
- [ ] added unit tests or modified existing tests to cover new code paths, ensuring the threshold 
  for code coverage.
- [ ] added integration tests.
- [ ] been tested in a test IoTDB cluster.

<hr>

PR #9638, opened 2023-04-18
[IOTDB-5727][hotfix] hide PipeInfo from PipeManager

SteveYurongSu:
See #9533 for more detail.


SteveYurongSu:
The GH Actions are down AGAIN...

PR #9640, opened 2023-04-18
[IOTDB-5786] Fix potential deadlock in DriverScheduler

lancelly:
This pr aims to fix potential deadlock in DriverScheduler caused by concurrently calling abortFramentInstance and toAborted.
See https://apache-iotdb.feishu.cn/docx/KFsvdMwL4orqkCxrdARc37W8nxf for more details.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/mpp/execution/schedule/DriverScheduler.java
+++ b/server/src/main/java/org/apache/iotdb/db/mpp/execution/schedule/DriverScheduler.java
@@ -299,13 +299,9 @@ public class DriverScheduler implements IDriverScheduler, IService {
       for (Set<DriverTask> fragmentRelatedTasks : queryRelatedTasks.values()) {
         if (fragmentRelatedTasks != null) {
           for (DriverTask task : fragmentRelatedTasks) {
-            task.lock();
-            try {
-              task.setAbortCause(DriverTaskAbortedException.BY_QUERY_CASCADING_ABORTED);
-              clearDriverTask(task);
-            } finally {
-              task.unlock();
-            }
+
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/apache/iotdb/db/mpp/execution/schedule/DriverScheduler.java
```diff
@@ -299,13 +299,8 @@ public void abortQuery(QueryId queryId) {
...
         if (fragmentRelatedTasks != null) {
           for (DriverTask task : fragmentRelatedTasks) {
-            task.lock();
-            try {
-              task.setAbortCause(DriverTaskAbortedException.BY_QUERY_CASCADING_ABORTED);
-              clearDriverTask(task);
-            } finally {
-              task.unlock();
-            }
+            task.setAbortCause(DriverTaskAbortedException.BY_QUERY_CASCADING_ABORTED);
```
Alima777, line 302: There is a little problem. If we set abort cause without lock, the origin cause may be occupied.

Alima777: Oh, even though we set it with lock, it can be occupied too. Maybe a status check is necessary, or just delete the abort cause, it  seems we don't care it...

lancelly: > 

Yes, it seems that the last write wins

lancelly: > Oh, even though we set it with lock, it can be occupied too. Maybe a status check is necessary, or just delete the abort cause, it seems we don't care it...

For now we just need it not to be NULL

JackieTien97 (approved):

PR #9644, opened 2023-04-19
Add a cache to avoid initialize duplicated device id object in write process

HTHou:
In write process, especially write by insertRecords interface, the PartialPath Object of deviceId may init too many times. This PR is adding a DataNodeDevicePathCache to avoid initialize duplicated PartialPath.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/cache/DataNodeDevicePathCache.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/cache/DataNodeDevicePathCache.java
@@ -21,17 +21,21 @@ package org.apache.iotdb.db.metadata.cache;
 
 import org.apache.iotdb.commons.exception.IllegalPathException;
 import org.apache.iotdb.commons.path.PartialPath;
+import org.apache.iotdb.db.conf.IoTDBConfig;
+import org.apache.iotdb.db.conf.IoTDBDescriptor;
 
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
 
-/** This class is for reducing duplicated Device PartialPath splits in write process. */
+/** This cache is for reducing duplicated DeviceId PartialPath initialization in write process. */
 public class DataNodeDevicePathCache {
+
+  private static final IoTDBConfig config = IoTDBDescriptor.getInstance().getConfig();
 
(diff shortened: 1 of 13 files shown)


server/src/main/java/org/apache/iotdb/db/metadata/cache/DataNodeDevicePathCache.java
```diff
@@ -0,0 +1,69 @@
...
+    return DataNodeDevicePathCache.DataNodeDevicePathCacheHolder.INSTANCE;
+  }
+
+  /** singleton pattern. */
+  private static class DataNodeDevicePathCacheHolder {
+    private static final DataNodeDevicePathCache INSTANCE = new DataNodeDevicePathCache();
+  }
+
+  public PartialPath getPartialPath(String deviceId) throws IllegalPathException {
+    try {
```
xingtanzjr, line 51: In current implementation, the `IllegalPathException` seems not be threw by this method ?

server/src/main/java/org/apache/iotdb/db/metadata/cache/DataNodeDevicePathCache.java
```diff
@@ -0,0 +1,69 @@
...
+      return devicePathCache.get(
+          deviceId,
+          path -> {
+            try {
+              return new PartialPath(path);
+            } catch (IllegalPathException e) {
+              throw new IllegalArgumentException(e);
+            }
+          });
+    } catch (IllegalArgumentException e) {
```
xingtanzjr, line 61: Why shall we catch `IllegalArgumentException` here

HTHou: image

The `get` method is not allowed to throw `IllegalPathException` directly. We have to throw an allowed `IllegalArgumentException` first and catch the exception outside, then re-throw `IllegalPathException`.

server/src/main/java/org/apache/iotdb/db/metadata/cache/DataNodeDevicePathCache.java
HTHou: Eh... actually I add a UT for this case, and it will throw `IllegalPathException` indeed...

image

xingtanzjr (approved):

PR #9645, opened 2023-04-19
[IOTDB-5791] Construct Cluster-LoadPublisher-Thread and IClusterStatusSubscriber

CRZbulabula:
New feature in ConfigNode-leader's LoadManager:

- Cluster-LoadPublisher-Thread: publish cluster events
- IClusterStatusSubscriber: subscribe cluster status events

See detailed design at [LoadManager](https://apache-iotdb.feishu.cn/docx/doxcnZ2tquzJKINQWi9nnYECPCb)


confignode/src/main/java/org/apache/iotdb/confignode/manager/load/balancer/RouteBalancer.java
```diff
@@ -189,10 +193,10 @@ private boolean updateRegionLeaderMap() {
             isLeaderChanged.set(true);
           }
         });
-    return isLeaderChanged.get();
+    isLeaderChanged.get();
```
OneSizeFitsQuorum, line 196: do we need this line?

CRZbulabula: I forgot to delete this line. Fix now~

PR #9658, opened 2023-04-20
[IOTDB-5789] Fix the time result column of Distinct function is incorrect

Wei-hao-Li:
fix: save min time for each distinct element.


library-udf/src/main/java/org/apache/iotdb/library/dprofile/UDTFDistinct.java
```diff
@@ -52,6 +55,8 @@ public class UDTFDistinct implements UDTF {
   private HashSet<String> stringSet;
   private Type dataType;
 
+  private Map<Object, Long> minTimeMemo;
```
JackieTien97, line 58: use `XXXLongHashMap` instead of previous `XXXHashSet` and your newly added `Map<Object, Long>`

JackieTien97 (changes requested):

PR #9667, opened 2023-04-21
[IOTDB-5792] Parallel encoding in MemTable flush

jt2594838:
For the issue, please refer to: 
https://issues.apache.org/jira/browse/IOTDB-5792

For the design and evaluation, please refer to:
https://apache-iotdb.feishu.cn/docx/TAu9dKrFioYQxmxhgUVcVrmmnWf


client-go
HTHou, line 1: Should update your local submodule code?

jt2594838: Updated.

neuyilan:
Fantastic job, only minor question, why not just using thread pool like this `ThreadPoolExecutor(int corePoolSize, int maximumPoolSize, long keepAliveTime, TimeUnit unit, BlockingQueue<Runnable> workQueue, ThreadFactory threadFactory)` rather than recreate the DynamicThread?

node-commons/src/main/java/org/apache/iotdb/commons/concurrent/pipeline/TaskRunner.java
```diff
@@ -0,0 +1,82 @@
...
+          break;
+        }
+      }
+
+      cleanUp.run();
+    } finally {
+      Thread.currentThread().setName(baseName);
+    }
+  }
+}
```
neuyilan, line 82: If the task in the input queue is completely fetched, it becomes blocked, and the thread is blocked in `input.take()`, so it cannot calculate the idle ratio and cannot exit. If there are many tasks, then the thread should not be idle. So in actual operation, the scenario where threads automatically exit mainly occurs when the production speed of producers cannot keep up with the consumption speed of consumers? For example, the speed of sort task generation cannot keep up with the speed of sort task processing; The speed of encoding task production cannot keep up with the speed of encoding task consumption？

server/src/main/java/org/apache/iotdb/db/engine/flush/MemTableFlushTask.java
```diff
@@ -218,159 +265,59 @@ public void syncFlushMemTable() throws ExecutionException, InterruptedException
...
+  }
 
-    StartFlushGroupIOTask(String deviceId) {
-      this.deviceId = deviceId;
-    }
+  private void cleanIOThread() {
+    metricFlush();
+    WRITING_METRICS.recordFlushCost(WritingMetrics.FLUSH_STAGE_IO, allContext.getIoTime().get());
+    WRITING_METRICS.recordFlushTsFileSize(storageGroup, allContext.getWriter().getFile().length());
   }
```
neuyilan, line 322: why call` metricFlush() `after each task pipeline?

neuyilan (approved):
Excellent job, I noticed that your design documentation and code implementation both mentioned: 

> "The previous version flushes devices and timeseries on each device in the lex order. To guarantee the sameflushing, the task should be properly divided so that the final l0 order can be preserved, while sort tasks and encoding tasks can be fully parallel"



I'm not sure if the lex order is based on what design when flush data is written to disk, but if we don't follow this order, there may be performance improvements in DeviceIOTask,

node-commons/src/assembly/resources/conf/iotdb-common.properties
```diff
@@ -541,6 +541,26 @@ cluster_name=defaultCluster
...
 # device_path_cache_size=500000
 
+# When flushing a MemTable, the range of thread number that will be available for each pipeline state.
+# Set to 1 when less than or equal to 0.
+# Datatype: int
+# flush_min_sub_thread_num=1
+# flush_max_sub_thread_num=16
+
+# If the idle ratio of a DynamicThread is below this value, it will try to add a new thread in
+# its group if there are fewer threads than flushMemTableMaxSubThread.
```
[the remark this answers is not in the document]
jt2594838: fixed

node-commons/src/assembly/resources/conf/iotdb-common.properties
```diff
@@ -541,6 +541,26 @@ cluster_name=defaultCluster
...
+# Datatype: double
+#dynamic_min_idle_ratio=0.1
+
+# If the idle ratio of a DynamicThread is over this value, it will try to exit if there are more
+# threads than flushMemTableMinSubThread.
```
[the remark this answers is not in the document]
jt2594838: fixed

node-commons/src/main/java/org/apache/iotdb/commons/concurrent/dynamic/DynamicThread.java
```diff
@@ -0,0 +1,117 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(DynamicThread.class);
+  private DynamicThreadGroup threadGroup;
+  private long idleStart;
+  private long runningStart;
+  private long idleTimeSum;
+  private long runningTimeSum;
+  // TODO: add configuration for the values
+  private double maximumIdleRatio = 0.5;
+  private double minimumIdleRatio = 0.1;
+  private long minimumRunningTime = 10_000_000_000L;
```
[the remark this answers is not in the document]
jt2594838: fixed

server/src/main/java/org/apache/iotdb/db/engine/flush/MemTableFlushTask.java
jt2594838: The original code did this at the end of each stage. I am not sure of its functionality, so I keep it.

node-commons/src/main/java/org/apache/iotdb/commons/concurrent/pipeline/TaskRunner.java
jt2594838: It is a good idea. However, the main intention of DynamicThread is to reduce the thread contention instead of the number of sleeping threads. The cost of contention is more significant as it consumes CPU resources to switch threads, while the sleeping threads do not consume any CPU resources.
That is, if the queue is exhausted, the thread will fall asleep permanently until the main task cancels it, which is more affordable than a living thread competing with others.

More importantly, it is very difficult to record the speed accurately and respond to it. Due to the existence of thread being switched out and garbage collection, the speed is not likely to be stable. 
The judgment of keeping up or not may be inaccurate or even erroneous, causing unwanted thread creation and destruction. 
If you have any idea that may precisely record and compare speed across tasks, I will be more than happy to make the improvement.

jt2594838:
> Excellent job, I noticed that your design documentation and code implementation both mentioned:
> 
> > "The previous version flushes devices and timeseries on each device in the lex order. To guarantee the sameflushing, the task should be properly divided so that the final l0 order can be preserved, while sort tasks and encoding tasks can be fully parallel"
> 
> I'm not sure if the lex order is based on what design when flush data is written to disk, but if we don't follow this order, there may be performance improvements in DeviceIOTask,

Agreed. But I am not sure what side effect it may cause, so we may discuss about it in another issue.

server/src/main/java/org/apache/iotdb/db/engine/flush/MemTableFlushTask.java
neuyilan: I see the previous code implementation only calls metricFlush at the end of the entire flush task. Maybe you need to confirm.

neuyilan:
> Agreed. But I am not sure what side effect it may cause, so we may discuss about it in another issue.

OK~

server/src/main/java/org/apache/iotdb/db/engine/flush/MemTableFlushTask.java
jt2594838: Now I only retain the one in the IOTask.

PR #9674, opened 2023-04-23
[IOTDB-5803] Improve query performance by reducing cpu consuming

Beyyes:
## Description
Specific improvements are as follow:
1. Using data structure which is more efficient.
2. Add query_metrics_enable switch
3. Avoiding modFile.exist judgement when there is no delete operation globally


server/src/main/java/org/apache/iotdb/db/mpp/plan/parser/StatementGenerator.java
```diff
@@ -208,7 +208,7 @@ public static Statement createStatement(TSAggregationQueryReq req, ZoneId zoneId
     SelectComponent selectComponent = new SelectComponent(zoneId);
     List<PartialPath> selectPaths = new ArrayList<>();
     for (String pathStr : req.getPaths()) {
-      selectPaths.add(new PartialPath(pathStr));
+      selectPaths.add(new PartialPath(pathStr.split("\\.")));
```
JackieTien97, line 211: We cannot directly use `String.split` unless user point it out.

node-commons/src/assembly/resources/conf/iotdb-common.properties
```diff
@@ -366,6 +366,10 @@ cluster_name=defaultCluster
 # Datatype: boolean
 # meta_data_cache_enable=true
 
+# Whether to enable query metrics monitoring or not. Enable this may cause a little query performance degradation.
+# Datatype: boolean
+# query_metric_enable = true
```
JackieTien97, line 371: I remember that @OneSizeFitsQuorum has already added a new level metric level which means no metrics at all, i think you can use that instead.

JackieTien97 (changes requested):

node-commons/src/assembly/resources/conf/iotdb-common.properties
SpriCoder: Yes, just use DO_NOTNING level

SpriCoder (changes requested):
I think we can use metric sets to "hold" these metrics and manage the life cycle of metrics.
You can see this PR: https://github.com/apache/iotdb/pull/9381

node-commons/src/assembly/resources/conf/iotdb-common.properties
Beyyes: Change the level of metric to `DO_NOTHING` will also take some costs in cpu bound situation(TSBS-benchAnts), the query cost of QueryMetricsManager is (4+1+3+2+2+1+3+2+3+3+2+1+2+2+1+1+15+2+2+3+1+3+15+9+12)/3884=2.4%.

image

JackieTien97 (approved):

PR #9685, opened 2023-04-24
[IOTDB-5810] Result handle in memory pool is not be cleaned properly

Alima777:
When Fragment instance is finished, related source handle and sink channels in memory pool will be all cleaned. However, result handle uses virtual fragment instance, so it can not be cleaned properly.


JackieTien97 (approved):

OneSizeFitsQuorum (approved):
LGTM

PR #9686, opened 2023-04-24
Remove unecessary synchronized on decrementCnt of SinkListener

lancelly:
At first, we add synchronized on decrementGet method to avoid concurrent modification on cnt or we may close the shuffleSinkHandle wrongly. However, using AtomicBoolean is enough to ensure that one channel only decrement the cnt once.
Moreover, adding synchronized on decrementGet method could lead to dead lock. Thread A invokes onFinsh of ISinkLitsener and holds the lock of it. Then A tries to lock the corresponding SinkChannel while Thread B invokes close of SinkChannel first and holds the lock of SinkChannel and then try to lock ISinkListener.
So we need to remove the synchonized keyword on decrementGet method.


lancelly:
> Add some description maybe..

done

PR #9687, opened 2023-04-24
Combine DataNodeSchemaCache of Template and Non-Template Scenarios 

MarcosZyk:
## Description

Currently, the DataNodeSchemaCache manages the timeseries not using template cache and DataNodeTemplateSchemaCache manages devices using template. It is hard to manage the total cache stats when keeping these separated. 

Combine these two cache.


bigreybear (approved):
LGTM

PR #9691, opened 2023-04-24
[IOTDB-5812] Reduce useless create of PartialPath in auth module

SpriCoder:
Reduce useless create of PartialPath in auth module


SpriCoder:
image

Already solve the performance problem

THUMarkLau (approved):

PR #9698, opened 2023-04-25
[IOTDB-5819] Fix npe when booting net metrics

THUMarkLau:
See [IOTDB-5819](https://issues.apache.org/jira/browse/IOTDB-5819).

In general, the files in /sys/class/net are considered to represent a network card, as shown in the figure below. 
image

However, in some environments, the /sys/class/net folder may also include non-network card files.
7ce6a129ea259417455cdda3f662c91

 NetMetrics uses a Map from the name of a network card to the status of a network card, and NPE occurs when the file name of a non-network card is used to get the status of a network card. To fix this bug, we use a `getOrDefault` method instead of `get` method to avoid NPE.


OneSizeFitsQuorum (approved):
LGTM

Cpaulyz:
Don't forget to cherry-pick this to rel/1.1 😃

PR #9699, opened 2023-04-25
[To rel/1.1][IOTDB-5820] Exception when use PIPE if sender has more than one data dir on different disks

Cpaulyz:
## Description

Solution: When the data directory and syncDir cross the file system, create a temporary directory in the data directory to hold the hard connection.

Note that when dropping the pipe, all dataDirs need to be scanned to clear the temporary directory.


Wei-hao-Li (approved):
has been verified

PR #9706, opened 2023-04-26
Sort the result of AlignedUpdateLastCacheOperator for LastQueryMergeOperator

lancelly:
LastQueryMergeOperator assumes that the TsBlocks its children resturn are sorted. However,  AlignedUpdateLastCacheOperator does not sort its result accoring to measurements.


server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/OperatorTreeGenerator.java
```diff
@@ -2292,6 +2293,7 @@ public Operator visitLastQueryMerge(LastQueryMergeNode node, LocalExecutionPlanC
         (items.isEmpty() || items.get(0).getOrdering() == Ordering.ASC)
             ? ASC_BINARY_COMPARATOR
             : DESC_BINARY_COMPARATOR;
+    context.setLastQueryComparator(comparator);
```
JackieTien97, line 2296: won't take effect because children have been visited before.

JackieTien97 (changes requested):

JackieTien97 (approved):

PR #9707, opened 2023-04-26
Use two params to distinguish fsync delay in the wal sync mode and async mode

HeimingZ:
Use two params to distinguish fsync delay in the wal sync mode and async mode

```
wal_async_mode_fsync_delay_in_ms=1000
wal_sync_mode_fsync_delay_in_ms=3
```


OneSizeFitsQuorum (commented):
update user doc?

BTW, wal_dirs now supports configuring multiple disks, but the user documentation has not been updated.

HeimingZ:
> update user doc?
> 
> BTW, wal_dirs now supports configuring multiple disks, but the user documentation has not been updated.

wal_dirs param always supports configuring multiple disks. But it seems that someone used to rename it to dn_wal_dir, which is confusing.

OneSizeFitsQuorum (approved):
LGTM

PR #9708, opened 2023-04-26
[IOTDB-5823] wal pipe handler for the pipe module

HeimingZ:


server/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
```diff
@@ -539,6 +539,10 @@ public class IoTDBConfig {
 
   /** Memory allocated proportion for time partition info */
   private long allocateMemoryForTimePartitionInfo = allocateMemoryForStorageEngine * 50 / 1001;
+
+  /** Memory allocated proportion for wal pipe cache */
+  private long allocateMemoryForWALPipeCache = allocateMemoryForConsensus / 10;
```
xingtanzjr, line 544: If we borrow memory from allocateMemoryForConsensus, the memory used by original part should also be decrease ?

xingtanzjr: Confirm with @HeimingZ , this will be changed by what the memory it used. @SteveYurongSu

PR #9717, opened 2023-04-26
[IOTDB-5815] Fix Npe when UDF spilling data to disk

lancelly:
https://issues.apache.org/jira/browse/IOTDB-5815


server/src/main/java/org/apache/iotdb/db/query/control/QueryResourceManager.java
```diff
@@ -74,7 +72,7 @@ public void endQuery(long queryId) {
     filePathsManager.removeUsedFilesForQuery(queryId);
 
     // close and delete UDF temp files
-    TemporaryQueryDataFileService.getInstance().deregister(queryId);
+    // TemporaryQueryDataFileService.getInstance().deregister(queryId);
```
SteveYurongSu, line 75: ?

SteveYurongSu (approved):

PR #9718, opened 2023-04-26
Change default multi_dir_strategy to SequenceStrategy and fix original bug

xingtanzjr:
## Description
1. change default multi_dir_strategy to SequenceStrategy
2. fix original bug in SequenceStrategy where one folder won't used if others space is limited
3. use `diskSpaceWarningThreshold` to decide whether a folder is full or not

## Test

Tested using 1C1D in fit16 with 1 DataRegion. See the snapshot below. The threshold is set to `0.85` in this test.
5BpUzfYisk

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/conf/directories/strategy/SequenceStrategy.java
+++ b/server/src/main/java/org/apache/iotdb/db/conf/directories/strategy/SequenceStrategy.java
@@ -18,7 +18,6 @@
  */
 package org.apache.iotdb.db.conf.directories.strategy;
 
-import com.sun.media.jfxmedia.logging.Logger;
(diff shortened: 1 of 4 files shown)


HeimingZ (approved):
LGTM. But the codes need running mvn spotless.

server/src/main/java/org/apache/iotdb/db/conf/directories/strategy/SequenceStrategy.java
```diff
@@ -18,6 +18,7 @@
  */
 package org.apache.iotdb.db.conf.directories.strategy;
 
+import com.sun.media.jfxmedia.logging.Logger;
```
HTHou, line 21:
```suggestion
```

PR #9726, opened 2023-04-27
[IOTDB-5827][To rel/1.1] Change default multi_dir_strategy to SequenceStrategy and fix original bug #9718

xingtanzjr:
## Description
1. change default multi_dir_strategy to SequenceStrategy
2. fix original bug in SequenceStrategy where one folder won't used if others space is limited
3. use `diskSpaceWarningThreshold` to decide whether a folder is full or not

## Test

Tested using 1C1D in fit16 with 1 DataRegion. See the snapshot below. The threshold is set to `0.85` in this test.
5BpUzfYisk


HeimingZ (approved):
LGTM

PR #9745, opened 2023-04-28
[rest service] support select into & cq clauses

CloudWise-Lukemiao:
Rest service supports select into or cq


server/src/main/java/org/apache/iotdb/db/query/control/clientsession/RestClientSession.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  @Override
+  public int getClientPort() {
+    return 0;
+  }
+
+  @Override
+  TSConnectionType getConnectionType() {
+    return TSConnectionType.MQTT_BASED;
```
SteveYurongSu, line 50:
```suggestion
    return TSConnectionType.REST_BASED;
```

PR #9753, opened 2023-05-03
[IOTDB-5828] Optimize the performance of some parts in metrics, and correcting the metrics count of temporal file size in inner space compaction

THUMarkLau:
See [IOTDB-5828](https://issues.apache.org/jira/browse/IOTDB-5828).

This PR optimize the performance of `JVMThreadMetrics` by caching the result of thread infos, and optimize the performance of `FileMetrics` by counting file in system directory(/proc/[PID]/fd) instead of calling Runtime.exec() in Linux. This PR refactor the way of getting file size and count of temporal files in compaction. The way is changed from updating size incrementally to recording by `CompactionTaskSummary`.


OneSizeFitsQuorum (approved):
LGTM

PR #9755, opened 2023-05-04
[IOTDB-5824] Fix show devices with * cannot display satisfied devices

MarcosZyk:
## Description

When traversing a tree structure, the current algorithm takes an greedy strategy that try match the node with states close to final state as much as possible, but not ensure a node due to match a final state do match a final state.

Add final state check and continue the traversing and state matching process until a candidate node match a final state or check all possible states.

PR #9761, opened 2023-05-05
[To rel/1.1][IOTDB-5824] Fix show devices with * cannot display satisfied devices

MarcosZyk:
## Description

same as #9755


Cpaulyz (approved):
LGTM

PR #9762, opened 2023-05-05
[To rel/1.1][IOTDB-5831]Fix drop database won't delete totally files in disk during data insertion

MarcosZyk:
## Description

same as #9754


CRZbulabula (approved):
LGTM

PR #9763, opened 2023-05-05
[IOTDB-5831]Fix drop database won't delete totally files in disk during data insertion

MarcosZyk:
## Description

When dropping database, the database schema shall be removed before database partition table, otherwise the pre-deleted check won't take effect.


CRZbulabula (approved):
LGTM～

PR #9768, opened 2023-05-05
[IOTDB-5840] Enhance the problem that the insertRecords interface may cause the number of threads to balloon when there are too many data regions

OneSizeFitsQuorum:
see [jira](https://issues.apache.org/jira/browse/IOTDB-5840)

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/mpp/plan/scheduler/AsyncPlanNodeSender.java
+++ b/server/src/main/java/org/apache/iotdb/db/mpp/plan/scheduler/AsyncPlanNodeSender.java
@@ -30,7 +30,6 @@ import org.apache.iotdb.mpp.rpc.thrift.TSendSinglePlanNodeReq;
 import org.apache.iotdb.mpp.rpc.thrift.TSendSinglePlanNodeResp;
 import org.apache.iotdb.rpc.RpcUtils;
 import org.apache.iotdb.rpc.TSStatusCode;
-import org.apache.iotdb.tsfile.utils.Pair;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -52,7 +51,7 @@ public class AsyncPlanNodeSender {
       asyncInternalServiceClientManager;
   private final List<FragmentInstance> instances;
 
-  private final Map<TEndPoint, Pair<List<Integer>, TSendBatchPlanNodeReq>> batchRequests;
+  private final Map<TEndPoint, BatchRequestWithIndex> batchRequests;
   private final Map<Integer, TSendSinglePlanNodeResp> instanceId2RespMap;
   private final AtomicLong pendingNumber;
   private final long startSendTime;
@@ -66,14 +65,12 @@ public class AsyncPlanNodeSender {
     this.instances = instances;
     this.batchRequests = new HashMap<>();
     for (int i = 0; i < instances.size(); i++) {
-      Pair<List<Integer>, TSendBatchPlanNodeReq> value =
-          this.batchRequests.computeIfAbsent(
+      this.batchRequests
+          .computeIfAbsent(
               instances.get(i).getHostDataNode().getInternalEndPoint(),
-              x -> new Pair<>(new ArrayList<>(), new TSendBatchPlanNodeReq()));
-      value.getLeft().add(i);
-      value
-          .getRight()
-          .addToRequests(
+              x -> new BatchRequestWithIndex())
+          .addSinglePlanNodeReq(
+              i,
               new TSendSinglePlanNodeReq(
                   new TPlanNode(
                       instances.get(i).getFragment().getPlanNodeTree().serializeToByteBuffer()),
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/apache/iotdb/db/service/thrift/impl/DataNodeInternalRPCServiceImpl.java
```diff
@@ -289,18 +290,25 @@ public TSendFragmentInstanceResp sendFragmentInstance(TSendFragmentInstanceReq r
...
-    RegionWriteExecutor executor = new RegionWriteExecutor();
-    TSendPlanNodeResp resp = new TSendPlanNodeResp();
-    RegionExecutionResult executionResult = executor.execute(groupId, planNode);
-    resp.setAccepted(executionResult.isAccepted());
-    resp.setMessage(executionResult.getMessage());
-    resp.setStatus(executionResult.getStatus());
-    return resp;
+  public TSendBatchPlanNodeResp sendBatchPlanNode(TSendBatchPlanNodeReq req) {
+    List<TSendSinglePlanNodeResp> responses =
+        req.getRequests().stream()
```
xingtanzjr, line 295: Although the count of RPCs decreased, the parallelism also decreased here because the FIs towards different DataRegions in need to be executed one by one. 

Does it impact the write speed in some scenario ?

server/src/main/java/org/apache/iotdb/db/mpp/plan/scheduler/FragmentInstanceDispatcherImpl.java
```diff
@@ -268,11 +270,15 @@ private void dispatchRemote(FragmentInstance instance, TEndPoint endPoint)
           }
           break;
         case WRITE:
-          TSendPlanNodeReq sendPlanNodeReq =
-              new TSendPlanNodeReq(
-                  new TPlanNode(instance.getFragment().getPlanNodeTree().serializeToByteBuffer()),
-                  instance.getRegionReplicaSet().getRegionId());
-          TSendPlanNodeResp sendPlanNodeResp = client.sendPlanNode(sendPlanNodeReq);
+          TSendBatchPlanNodeReq sendPlanNodeReq =
```
xingtanzjr, line 273: It seems that this change is unnecessary because only one Req is processed in this batch. Maybe we can still use original way to do that

server/src/main/java/org/apache/iotdb/db/mpp/plan/scheduler/AsyncPlanNodeSender.java
```diff
@@ -43,38 +46,53 @@
...
           asyncInternalServiceClientManager,
       List<FragmentInstance> instances) {
+    this.startSendTime = System.nanoTime();
     this.asyncInternalServiceClientManager = asyncInternalServiceClientManager;
     this.instances = instances;
-    this.instanceId2RespMap = new ConcurrentHashMap<>();
-    this.pendingNumber = new AtomicLong(instances.size());
+    this.batchRequests = new HashMap<>();
+    for (int i = 0; i < instances.size(); i++) {
+      Pair<List<Integer>, TSendBatchPlanNodeReq> value =
```
xingtanzjr, line 69: Replacing `Pair` with an inner class may make this code more concise and understandable

server/src/main/java/org/apache/iotdb/db/service/thrift/impl/DataNodeInternalRPCServiceImpl.java
OneSizeFitsQuorum: This may affect performance in some scenarios, but I think it is reasonable.

For a write request, does the server need to use parallel techniques to reduce latency? To this question, I think the most important thing is to provide a coordinated and unified answer. If parallelism is not used, the delay cannot be reduced at this time, but it can ensure that the request consumes at most one thread on one node to execute the request. To improve the throughput, the user can directly increase the concurrency. If parallelism is used, then the request should be able to use multiple threads on any related node to reduce latency, although there may be some performance costs when concurrency is high.

However, the code in the current dispatch phase is not clear. If you look at the code [here](https://github.com/apache/iotdb/blob/master/server/src/main/java/org/apache/iotdb/db/mpp/plan/scheduler/FragmentInstanceDispatcherImpl.java#L185-L202), it is currently executing in serial for localInstance, while executing in parallel for remoteInstance, which I think is very strange. It does not provide users with a coherent answer.

Perhaps ideally, both localInstance and remoteInstance should be executed in parallel, so that there is a coherent answer. But this would introduce a thread pool, which I don't really want to do right now.

In summary, I currently think it is a better and clearer choice to provide a non-parallel semantics for write requests that users can understand and adjust concurrency to improve throughput. After we reduce the number of thread pools and are able to do some scheduling in the application state, perhaps we can parallel both local and remote execution, allowing latency to fall further in low-concurrency scenarios.

server/src/main/java/org/apache/iotdb/db/mpp/plan/scheduler/FragmentInstanceDispatcherImpl.java
OneSizeFitsQuorum: I think batch means that there is more than or equal to one. Writing two interfaces in thrift for the same logic results in more redundant code on the sender and receiver...

server/src/main/java/org/apache/iotdb/db/mpp/plan/scheduler/AsyncPlanNodeSender.java
OneSizeFitsQuorum: OK ~ I'll do it later tonight~

server/src/main/java/org/apache/iotdb/db/mpp/plan/scheduler/FragmentInstanceDispatcherImpl.java
xingtanzjr: make sense

server/src/main/java/org/apache/iotdb/db/service/thrift/impl/DataNodeInternalRPCServiceImpl.java
xingtanzjr: Got it. Currently the `dispatchLocally` is a SYNC operation. Maybe we can talk about it later. I have noted this point on BenchAnts performance enhancement doc.

xingtanzjr (approved):

PR #9783, opened 2023-05-08
[To rel/1.1] Refactor compaction task metrics

THUMarkLau:
This PR refactor the way to update compaction task metrics. The previous way to update the metrics is updating the counting value when compaction task is submitted or starts to run, but it could be not accurate because the task submission could fail. This pr change the way of counting task by getting running task list and task waiting queue from task manager, and count them when the metrics framework tries to report the value.


xingtanzjr:
The whole logic is OK but not sure whether its efficiency is OK because it seems that each metrics pull will trigger one invoke regarding `getCompactionTaskStatistic()`

xingtanzjr:
> The whole logic is OK but not sure whether its efficiency is OK because it seems that each metrics pull will trigger one invoke regarding `getCompactionTaskStatistic()`

@THUMarkLau has confirmed there is no more cpu cost

PR #9800, opened 2023-05-09
[IOTDB-5831] Fix create region failure after recreate db

MarcosZyk:
### Problem
c94696e4fb690a9ba78087b4381b521

### Cause

The DataNode fetched the predeleted database and carries on to create regions, while the database is deleted, which results in the region creation failure.

### Solution

Add predeleted check on ConfigNode when return database msg to DataNode.


SpriCoder (approved):
LGTM~

CRZbulabula (approved):
LGTM～

PR #9818, opened 2023-05-10
[To rel/1.1][IOTDB-5859]Compaction error when using Version as first sort dimension

choubenson:
**Description**
Starting from version 1.0, using the load function to import data and then compact them will result in overlapped data in sequential files.

**Reason**
After version 1.0, using the load function may cause files with the same timestamp to appear in the system, and the version corresponding to the sequential file with the same timestamp is incremented, and the version of the file under a time partition is unique. Compaction use `version` as first sort dimension, resulting in an incorrect order.

**Solution**
Use `timestamp` as first sort dimension and `version` as second sort dimension.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResourceList.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResourceList.java
@@ -143,43 +143,45 @@ public class TsFileResourceList implements List<TsFileResource> {
       return false;
     }
     if (tail == null) {
+      // empty list
       header = newNode;
       tail = newNode;
       count++;
     } else {
       // find the position to insert of this node
       // the list should be ordered by file timestamp
-      long versionOfNewNode =
-          TsFileNameGenerator.getTsFileName(newNode.getTsFile().getName()).getVersion();
+      TsFileNameGenerator.TsFileName newTsFileName =
+          TsFileNameGenerator.getTsFileName(newNode.getTsFile().getName());
+      long timestampOfNewNode = newTsFileName.getTime();
+      long versionOfNewNode = newTsFileName.getVersion();
 
-      if (TsFileNameGenerator.getTsFileName(header.getTsFile().getName()).getVersion()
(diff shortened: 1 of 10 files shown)


server/src/main/java/org/apache/iotdb/db/engine/storagegroup/TsFileResourceList.java
```diff
@@ -143,43 +143,43 @@ public boolean keepOrderInsert(TsFileResource newNode) throws IOException {
...
-        // the timestamp of new node is greater than the tail node
-        // insert it after the tail
+      TsFileNameGenerator.TsFileName newTsFileName =
+          TsFileNameGenerator.getTsFileName(newNode.getTsFile().getName());
+      long timestampOfNewNode = newTsFileName.getTime();
+      long versionOfNewNode = newTsFileName.getVersion();
+
+      // find the position to insert, the list should be ordered by file timestamp. If timestamp is
+      // equal to each other, then list should be ordered by file version.
+      boolean isNewNodeAtTail = true;
```
xingtanzjr, line 158: Actually this var is unnecessary. We can use curNode's null value to infer whether the new node should be inserted at tail or not

server/src/test/java/org/apache/iotdb/db/engine/storagegroup/TsFileResourceListTest.java
```diff
@@ -212,4 +213,113 @@ public void testCount() {
         tsFileResourceList.getHeader(), tsFileResources.get(tsFileResources.size() - 1));
     Assert.assertEquals(tsFileResourceList.getTail(), tsFileResources.get(0));
   }
+
+  /** Seq files: 1-1, 1-5, 1-6, 2-4, 3-3, 4-7, 4-9, 5-8 */
+  @Test
+  public void testKeepOrderInsertWithSameTimestampAndDifferentVersion() throws IOException {
```
xingtanzjr, line 219: It is better to add more test cases here

xingtanzjr (approved):

PR #9819, opened 2023-05-10
[IOTDB-5843] Stall write requests when system shutting down

SzyWilliam:
# Changes proposed
See https://issues.apache.org/jira/browse/IOTDB-5843.
In RatisConsensus
1. reject client write-requests when system-readonly.
2. stall state machine applying when system-readonly (not because of shutting down)


OneSizeFitsQuorum (approved):
LGTM

wangchao316 (approved):
LGTM

PR #9844, opened 2023-05-14
[IOTDB-5872] Reduce some IoTConsensus index logs

BUAAserein:
#### In a long-running cluster, most of the logs are related to the index, reducing the number of index logs can help reduce the number of logs.
- Change the print length of the index log segment from 1000 to 10000.
- Deleted index persistence when synchronization was completed.


OneSizeFitsQuorum (approved):
LGTM

PR #9858, opened 2023-05-16
[IOTDB-5882] Support create empty schema template

MarcosZyk:
### Motivation

When user cannot define specific measurements before data insertion, it's necessary to set up an empty schema template and auto extend it during data insertion.

### Modification

Session already support create empty schema template, thus only antlr is modified to support create empty schema template via SQL.


Cpaulyz (approved):
LGTM

PR #9864, opened 2023-05-16
[IOTDB-5884] Throw Exception when alter template with duplicate measurement

MarcosZyk:
### Problem

When there's duplicate measurement in one template extension request, there's no exception returned but the request is executed successfully and the registered measurement is the first one.

### Cause

When extending schema template on configNode, the measurement duplication will be ignored and the first one takes effect.

### Solution

Add duplication check on dataNode before sending RPC request to configNode.
Add deduplication in template auto extension during data insertion with insert records or tablets or TsFile load.

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/metadata/template/alter/TemplateExtendInfo.java
+++ b/server/src/main/java/org/apache/iotdb/db/metadata/template/alter/TemplateExtendInfo.java
@@ -28,7 +28,9 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 public class TemplateExtendInfo extends TemplateAlterInfo {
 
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/apache/iotdb/db/metadata/template/alter/TemplateExtendInfo.java
```diff
@@ -102,6 +104,24 @@ public void addMeasurement(
...
+      if (set.contains(measurements.get(i))) {
+        continue;
+      }
+      set.add(measurements.get(i));
+      result.addMeasurement(
+          measurements.get(i), dataTypes.get(i), encodings.get(i), compressors.get(i));
+    }
+    return result;
+  }
+
```
Cpaulyz, line 124:
```suggestion
  public boolean isDuplicated() {
    if (measurements != null) {
      Set<String> set = new HashSet<>(measurements);
      return set.size()!=measurements.size();
    }
    return false;
  }
```

server/src/main/java/org/apache/iotdb/db/mpp/plan/execution/config/executor/ClusterConfigTaskExecutor.java
```diff
@@ -1382,6 +1387,28 @@ public SettableFuture<ConfigTaskResult> dropSchemaTemplate(
...
+          future.setException(
+              new MetadataException(
+                  String.format(
+                      "Duplicated measurement [%s] in schema template alter request",
+                      measurement)));
+          return future;
+        } else {
+          existingMeasurements.add(measurement);
+        }
+      }
```
Cpaulyz, line 1409:
```suggestion
      TemplateExtendInfo templateExtendInfo =
          (TemplateExtendInfo) alterSchemaTemplateStatement.getTemplateAlterInfo();
      if (templateExtendInfo.isDuplicated()) {
          future.setException(
              new MetadataException(
                  String.format(
                      "Duplicated measurement [%s] in schema template alter request",
                      measurement)));
          return future;
      }
```

Cpaulyz (commented):
PTAL

server/src/main/java/org/apache/iotdb/db/mpp/plan/execution/config/executor/ClusterConfigTaskExecutor.java
MarcosZyk: Already add ```TemplateExtendInfo#getFirstDuplicateMeasurement``` and use it.

server/src/main/java/org/apache/iotdb/db/metadata/template/alter/TemplateExtendInfo.java
MarcosZyk: Already add ```TemplateExtendInfo#getFirstDuplicateMeasurement``` and use it.

Cpaulyz (approved):
LGTM

PR #9871, opened 2023-05-17
Support schema cache when querying all measurement of devices using template

MarcosZyk:
## Description

Currently, a device using schema template may be cached in DataNodeSchemaCache, but the cache is only used for data insertion.

Enable data query to benefit from such device cache when querying all measurements of the device, e.g. select * from root.db.d, with schema template t1 set on db.

### DIFF
--- a/node-commons/src/main/java/org/apache/iotdb/commons/path/PartialPath.java
+++ b/node-commons/src/main/java/org/apache/iotdb/commons/path/PartialPath.java
@@ -557,8 +557,8 @@ public class PartialPath extends Path implements Comparable<Path>, Cloneable {
           // if without MULTI_LEVEL_PATH_WILDCARD, scan and check
           if ((PathPatternUtil.hasWildcard(nodes1[i - 1])
                   && PathPatternUtil.isNodeMatch(nodes1[i - 1], nodes2[j - 1]))
-              || (PathPatternUtil.hasWildcard(nodes2[i - 1])
-                  && PathPatternUtil.isNodeMatch(nodes2[i - 1], nodes1[j - 1]))
+              || (PathPatternUtil.hasWildcard(nodes2[j - 1])
+                  && PathPatternUtil.isNodeMatch(nodes2[j - 1], nodes1[i - 1]))
               || nodes1[i - 1].equals(nodes2[j - 1])) {
             // if nodes1[i-1] and nodes[2] is matched, dp[i][j] = dp[i-1][j-1]
             dp[i][j] |= dp[i - 1][j - 1];
--- a/node-commons/src/main/java/org/apache/iotdb/commons/path/PartialPath.java
+++ b/node-commons/src/main/java/org/apache/iotdb/commons/path/PartialPath.java
@@ -40,7 +40,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
-import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import static org.apache.iotdb.commons.conf.IoTDBConstant.MULTI_LEVEL_PATH_WILDCARD;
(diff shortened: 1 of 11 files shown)


node-commons/src/main/java/org/apache/iotdb/commons/path/PartialPath.java
```diff
@@ -107,13 +107,25 @@ public PartialPath(String path, boolean needSplit) {
...
 
+  // e.g. root.db.d.s, root.db.d.*, root.db.d.s*, not include patterns like root.db.d.**
+  public boolean hasExplicitDevice() {
+    for (int i = 0; i < nodes.length - 1; i++) {
+      // *, ** , d*, *d*
+      if (PathPatternUtil.hasWildcard(nodes[i])) {
+        return false;
+      }
+    }
+    return !nodes[nodes.length - 1].equals(MULTI_LEVEL_PATH_WILDCARD);
```
Cpaulyz, line 126:
```suggestion
    if(nodes[nodes.length - 1].equals(MULTI_LEVEL_PATH_WILDCARD)){
        return false;
    }
    for (int i = 0; i < nodes.length - 1; i++) {
      // *, ** , d*, *d*
      if (PathPatternUtil.hasWildcard(nodes[i])) {
        return false;
      }
    }
    return true;
```

node-commons/src/main/java/org/apache/iotdb/commons/path/PathPatternUtil.java
```diff
@@ -0,0 +1,47 @@
...
+public class PathPatternUtil {
+
+  private PathPatternUtil() {}
+
+  /**
+   * The input string is a single node of a path pattern. Return true if the node may be a
+   * patternNode that can match batch explicit node names. e.g. *, e.g. *, **, d*, *d*.
+   */
+  public static boolean hasWildcard(String node) {
+    return node.startsWith(ONE_LEVEL_PATH_WILDCARD) || node.endsWith(ONE_LEVEL_PATH_WILDCARD);
```
Cpaulyz, line 36:
```suggestion
    return node.contains(ONE_LEVEL_PATH_WILDCARD);
```

node-commons/src/main/java/org/apache/iotdb/commons/path/PathPatternUtil.java
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  /** The input patternNode shall be a string starts or ends with *, e.g. *, **, d*, *d* */
+  public static boolean isNodeMatch(String patternNode, String nodeName) {
```
Cpaulyz, line 40:
```suggestion
  /** 
  * Determine if a wildcard matches a node name.
  *
  * @param wildcardPattern must be a string starts or ends with *, e.g. *, **, d*, *d* 
  * @param nodeName node to match
  */
  public static boolean isWildcardMatch(String wildcardPattern, String nodeName) {
```

Cpaulyz (commented):
Some code smells and execution optimization. PTAL~

node-commons/src/main/java/org/apache/iotdb/commons/path/PartialPath.java
MarcosZyk: Fixed.

node-commons/src/main/java/org/apache/iotdb/commons/path/PathPatternUtil.java
MarcosZyk: The new judgement will be kept and all other related logic has been modified and rely on this method.

node-commons/src/main/java/org/apache/iotdb/commons/path/PathPatternUtil.java
MarcosZyk: Fixed.

Cpaulyz (approved):
LGTM

PR #9872, opened 2023-05-17
[IOTDB-5889] [To rel/1.1] Fix not releasing readlock of TsFile when fail to allocate memory for compaction

THUMarkLau:
See [IOTDB-5889](https://issues.apache.org/jira/browse/IOTDB-5889).

The reason for this bug is that when the compaction task failed to allocate enough memory to starts, the system will throw an exception and just abort the task. But when the task is aborted, the status of tsfile resource doesn't change from COMPACTING to NROMAL, and the read lock doesn't get released. As a result of that, TTL thread failed to get the write lock of the tsfile and fail to delete the file.

This PR fix the bug by changing the order of compaction task starting process from 
```Getting Readlock -> Changing file status -> Allocating memory```
to
```Allocating memory -> Getting readlock -> Changing file status```


xingtanzjr (approved):

choubenson (approved):
LGTM

PR #9886, opened 2023-05-18
[To rel/1.1][IOTDB-5897] Fix NullPointerException in compaction

THUMarkLau:
See [IOTDB-5897](https://issues.apache.org/jira/browse/IOTDB-5897).

The reason for this bug is that when reading chunk that contains 0 points, the list will set the chunk metadata to null.


choubenson (approved):
LGTM. Add some tests that contain empty chunks and empty pages in the align timeseries.

xingtanzjr (approved):

PR #9895, opened 2023-05-19
Change cost time unit from milli to nano & merge operator metric together in one driver

JackieTien97:
1. Change cost time unit from milli to nano
2. merge operator metric together in one driver
3. correct the bloomfilter cache hit ratio and timeseries metadata cache hit ratio


SpriCoder (approved):
LGTM~

PR #9896, opened 2023-05-19
[IOTDB-5890] [IOTDB-5891] Fix messages problem in order by

ycycse:
1. change the default size of sort buffer
2. add more semantic check in order by


JackieTien97 (changes requested):
default 1MB or 10MB?

PR #9898, opened 2023-05-20
[IOTDB-5848] Allow ratis-client retry when gRPC IO Unavailable

SzyWilliam:
## Changes proposed
Current retry policy blocks retry when IOExceptions except RaftExceptions are thrown. However, when gRPC channel is disconnected, another type of IOException (StatusRuntimeException) is thrown. We should allow it to be retried.


OneSizeFitsQuorum (approved):
LGTM

PR #9940, opened 2023-05-25
[IOTDB-5928] Refactor TsFile status transition module

xingtanzjr:
## Description
Before this change, a `TsFileResource`'s status can be set to any value by different layer, which may cause concurrent issue when comparing and setting the status of TsFileResources.

In this PR, we made two changes:
1. make `status` of TsFileResource as an atomic variable and support `compareAndSet` method
2. Separate read/write of `status` with `TsFileResource`'s read/write
3. Add a state-machine-kind transition method for `setStatus()` which will return `true` or `false` to represent whether the status is changed successfully

After above changes, we optimize some lock-related logic and fix a potential dead-lock issue when delete a TsFile by TTL and Compaction module


HeimingZ (approved):
LGTM

PR #9952, opened 2023-05-26
Fix memory control when creating timeseries using schema template

MarcosZyk:
## Description

The memory control is missed when creating timeseries using schema template, which results in unlimited memory usage.

Add memory status check.


bigreybear (approved):
LGTM!

PR #9962, opened 2023-05-29
[IOTDB-5929] Enable DataPartition inherit policy

CRZbulabula:
We've found in real production environment that queries across different DataRegionGroups are time-consuming. Therefore we should enable and enhance the DataPartition inherit policy. The new created DataPartition will inherit its neighbor(predecessor or successor)'s allocate result if the neighbor is allocated before.


OneSizeFitsQuorum (approved):
LGTM

SpriCoder (approved):
LGTM~

PR #9967, opened 2023-05-29
[IOTDB-5930] Support OrderBy in Last Query

ycycse:
design doc: https://apache-iotdb.feishu.cn/docx/A0dSdumbKoIxEDxU2WFcqxExnlb


server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/LogicalPlanBuilder.java
```diff
@@ -1215,6 +1213,16 @@ private LogicalPlanBuilder planSingleShowQueries(TDataNodeLocation dataNodeLocat
...
   }
 
+  public LogicalPlanBuilder planOrderBy(List<SortItem> sortItemList) {
+    OrderByParameter orderByParameter = new OrderByParameter(sortItemList);
+    if (orderByParameter.isEmpty()) {
+      return this;
+    }
+
+    this.root = new SortNode(context.getQueryId().genPlanNodeId(), root, orderByParameter);
+    return this;
```
JackieTien97, line 1223:
```suggestion
    if (sortItemList.isEmpty()) {
      return this;
    }

    this.root = new SortNode(context.getQueryId().genPlanNodeId(), root, new OrderByParameter(sortItemList););
    return this;
```

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/Analysis.java
```diff
@@ -195,6 +196,8 @@ public class Analysis {
 
   private OrderByParameter mergeOrderParameter;
 
+  private Ordering timeseriesOrderingForLastQuery = null;
```
JackieTien97, line 199: add some comments for this field, like: when there only exist order by timeseries clause in last query, this field will be set, otherwise it will be null.

server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/OperatorTreeGenerator.java
```diff
@@ -2233,7 +2226,7 @@ public Operator visitLastQuery(LastQueryNode node, LocalExecutionPlanContext con
 
     int initSize = cachedLastValueAndPathList != null ? cachedLastValueAndPathList.size() : 0;
     // no order by clause
```
JackieTien97, line 2228:
```suggestion
    // no need to order by timeseries at first
```

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/LogicalPlanBuilder.java
ycycse: done

server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/OperatorTreeGenerator.java
ycycse: done

JackieTien97 (approved):

PR #9981, opened 2023-05-30
Update schema memory control config

MarcosZyk:
### Motivation

The lastcache parameter in ```schema_memory_allocate_proportion``` is useless, thus should be removed. 

Besides, we need to keep compatible with previous properties file.

### Modification

Introduce new config ```schema_memory_proportion```, which keep consistency with other parameters on naming style and only cover schema region, schema cache and partition cache. 

The old config ```schema_memory_allocate_proportion``` is removed from properties file, but still can take effect if an IoTDB is upgrade from previous version.

### DIFF
--- a/node-commons/src/assembly/resources/conf/iotdb-common.properties
+++ b/node-commons/src/assembly/resources/conf/iotdb-common.properties
@@ -186,7 +186,7 @@ cluster_name=defaultCluster
 # storage_query_schema_consensus_free_memory_proportion=3:3:1:1:2
 
 # Schema Memory Allocation Ratio: SchemaRegion, SchemaCache, and PartitionCache.
-# The parameter form is a:b:c, where a, b and d are integers. for example: 1:1:1 , 6:2:1
+# The parameter form is a:b:c, where a, b and c are integers. for example: 1:1:1 , 6:2:1
 # schema_memory_proportion=5:4:1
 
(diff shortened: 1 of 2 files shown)


node-commons/src/assembly/resources/conf/iotdb-common.properties
```diff
@@ -185,9 +185,9 @@ cluster_name=defaultCluster
 # If you have high level of writing pressure and low level of reading pressure, please adjust it to for example 6:1:1:1:2
 # storage_query_schema_consensus_free_memory_proportion=3:3:1:1:2
 
-# Schema Memory Allocation Ratio: SchemaRegion, SchemaCache, PartitionCache and LastCache.
-# The parameter form is a:b:c:d, where a, b, c and d are integers. for example: 1:1:1:1 , 6:2:1:1
-# schema_memory_allocate_proportion=5:3:1:1
+# Schema Memory Allocation Ratio: SchemaRegion, SchemaCache, and PartitionCache.
+# The parameter form is a:b:c, where a, b and d are integers. for example: 1:1:1 , 6:2:1
```
HTHou, line 189:
```suggestion
# The parameter form is a:b:c, where a, b and c are integers. for example: 1:1:1 , 6:2:1
```

MarcosZyk: Fixed.

PR #9990, opened 2023-05-30
[IOTDB-5934] Let series_slot_num compatible

CRZbulabula:
Now the parameters in system.properties will force cover the user configurations.

image

### DIFF
--- a/confignode/src/main/java/org/apache/iotdb/confignode/conf/SystemPropertiesUtils.java
+++ b/confignode/src/main/java/org/apache/iotdb/confignode/conf/SystemPropertiesUtils.java
@@ -117,6 +117,9 @@ public class SystemPropertiesUtils {
       }
     }
 
+    final String format =
+        "[SystemProperties] The parameter \"{}\" can't be modified after first startup."
+            + " Your configuration: {} will be forced update to: {}";
     // Consensus protocol configuration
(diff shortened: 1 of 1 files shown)


confignode/src/main/java/org/apache/iotdb/confignode/conf/SystemPropertiesUtils.java
```diff
@@ -124,11 +124,11 @@ public static void checkSystemProperties() throws IOException, ConfigurationExce
       needReWrite = true;
     } else if (!configNodeConsensusProtocolClass.equals(
         conf.getConfigNodeConsensusProtocolClass())) {
-      throw new ConfigurationException(
-          "config_node_consensus_protocol_class",
+      LOGGER.warn(
+          "[SystemProperties] The parameter \"config_node_consensus_protocol_class\" can't be modified after first startup. Your configuration: {} will be forced update to: {}",
```
OneSizeFitsQuorum, line 128: Please fix the code smell of these long one-character lines, and we can also extract these same strings to unify them

CRZbulabula: Fixed～

OneSizeFitsQuorum (approved):
LGTM

PR #9991, opened 2023-05-30
Support using alias view in last query & optimize code

liuminghui233:


server/src/main/java/org/apache/iotdb/db/mpp/plan/planner/LogicalPlanBuilder.java
```diff
@@ -207,22 +207,33 @@ public LogicalPlanBuilder planLast(
...
-        sourceExpressions.stream()
-            .map(expression -> ((TimeSeriesOperand) expression).getPath())
-            .collect(Collectors.toList());
+    Map<PartialPath, List<String>> selectedPathToOutputSymbolsMap = new LinkedHashMap<>();
+    for (Expression sourceExpression : sourceExpressions) {
+      PartialPath selectedPath = ((TimeSeriesOperand) sourceExpression).getPath();
+      selectedPathToOutputSymbolsMap
+          .computeIfAbsent(selectedPath, k -> new ArrayList<>())
+          .add(sourceExpression.getOutputSymbol());
+    }
```
JackieTien97, line 216: That's not right, we need to make sure the order of measurement,  in this way, you break the order.

server/src/main/java/org/apache/iotdb/db/mpp/plan/analyze/ExpressionUtils.java
```diff
@@ -76,19 +60,16 @@ public static List<Expression> reconstructFunctionExpressions(
...
+      resultExpressions.add(reconstructFunctionExpression(expression, functionExpressions));
     }
     return resultExpressions;
   }
 
   public static Expression reconstructFunctionExpression(
       FunctionExpression expression, List<Expression> childExpressions) {
-    return new FunctionExpression(
-        expression.getFunctionName(), expression.getFunctionAttributes(), childExpressions);
+    FunctionExpression copiedFunctionExpression = (FunctionExpression) expression.copy();
```
JackieTien97, line 70: remove the copy method in each `Expression` and then set viewPath path in this class.

JackieTien97 (changes requested):
We need to make sure

JackieTien97 (approved):

JackieTien97:
d7l4hqWh8Y

Run CI in ClusterMode on my local laptop. These two IT failed because I increase the fork count, the load may be too heavy for my own laptop and when I rerun these two, it succeed.

PR #9992, opened 2023-05-30
Release resource of FI after all drivers have been closed

lancelly:
We can release resource of FI until all drivers have been closed, otherwise there could be memory leak/file handle leak.


server/src/main/java/org/apache/iotdb/db/mpp/execution/fragment/FragmentInstanceContext.java
```diff
@@ -352,6 +355,26 @@ private void addFilePathToMap(TsFileResource tsFile, boolean isClosed) {
     }
   }
 
+  public void initializeNumOfDrivers(Integer numOfDrivers) {
```
JackieTien97, line 358:
```suggestion
  public void initializeNumOfDrivers(int numOfDrivers) {
```

server/src/main/java/org/apache/iotdb/db/mpp/execution/fragment/FragmentInstanceContext.java
```diff
@@ -352,6 +355,26 @@ private void addFilePathToMap(TsFileResource tsFile, boolean isClosed) {
...
+  }
+
+  public void releaseResourceWhenAllDriversAreClosed() {
+    try {
+      allDriversClosed.await();
+    } catch (InterruptedException e) {
+      LOGGER.warn(
+          "Interrupted when await on allDriversClosed, FragmentInstance Id is {}", this.getId());
+      throw new RuntimeException(e);
+    }
```
JackieTien97, line 374:
```suggestion
    while (true) {
       try {
          allDriversClosed.await();
          break;
        } catch (InterruptedException e) {
          LOGGER.warn(
              "Interrupted when await on allDriversClosed, FragmentInstance Id is {}", this.getId());
        }
    }
```

JackieTien97 (changes requested):

server/src/main/java/org/apache/iotdb/db/mpp/execution/fragment/FragmentInstanceContext.java
lancelly: done

JackieTien97 (approved):

server/src/main/java/org/apache/iotdb/db/mpp/execution/fragment/FragmentInstanceContext.java
lancelly: done

PR #10012, opened 2023-06-01
Fix data type error during data insertion

MarcosZyk:
## Description

Some process on data type in PrimitiveArrayManager are left out when add new data type Unknown, which results in UnsupportedException


HTHou (approved):

Cpaulyz (approved):
LGTM

PR #10022, opened 2023-06-01
Fix ConcurrentModificationException in updating FileMetrics

THUMarkLau:
0d149246ba47731955f4741d86db11a

The reason is the map in FileMetrics is not concurrent-safe type.


OneSizeFitsQuorum (approved):
LGTM

OneSizeFitsQuorum (approved):
LGTM

PR #10025, opened 2023-06-01
[To rel/1.2] Added ITs for logical views. Fix duplicate rows in show views.

citrusreticulata:
### Added ITs for logical views.

Added ITs for logical views.

### Fix duplicate rows in show views.

There is a bug in PR https://github.com/apache/iotdb/pull/10000 cased duplicate rows in show views query.
Now it's fixed.


citrusreticulata:
This PR is not clean, so let's close it.

PR #10035, opened 2023-06-02
Get canonical path when using IOTDB_DATA_HOME and IOTDB_HOME

HeimingZ:
Current addDataHomeDir method in the IoTDBConfig cannot fulfill its responsibilities. We should get canonical path when using IOTDB_DATA_HOME and IOTDB_HOME
截屏2023-06-02 15 27 56


node-commons/src/main/java/org/apache/iotdb/commons/utils/PathUtils.java
```diff
@@ -202,4 +203,16 @@ private static boolean checkBackQuotes(String src) {
     }
     return src.length() == (src.replace("``", "").length() + num);
   }
+
+  /** Add a prefix to a relative path file */
+  public static String addPrefix2FilePath(String prefix, String file) {
```
lancelly, line 208: Perhaps we should put this method in FileUtils. PathUtils is for PartialPath.

HeimingZ: fixed

PR #10039, opened 2023-06-02
[To rel/1.2] Fix SchemaTree.hasView check during query analysis

MarcosZyk:
## Description

When constructing a SchemaTree, there's mistake on setting value ```hasViewMeasurement```, which results in the useless check in query without view.

Fix ```hasViewMeasurement``` updation.


citrusreticulata (approved):
LGTM :-) Excellent work!

PR #10040, opened 2023-06-02
Fix SchemaTree.hasView check during query analysis

MarcosZyk:
## Description

same as https://github.com/apache/iotdb/pull/10039


citrusreticulata (approved):
LGTM :-) Excellent work!

PR #10051, opened 2023-06-05
[To rel/1.1] Change default charset encoding to UTF-8 in start-datanode.bat 

xingtanzjr:
## Description
1. Change default charset encoding to UTF-8 in `start-datanode.bat`
2. Add log when DataNode/ConfigNode starts to print the default charset

image

NOTICE: Before this change, the explicit setting has only been added in `start-confignode.bat`. To keep the setting is consistent between DataNode and ConfigNode, we add the explicit charset encoding setting in script for windows.


HeimingZ (approved):
LGTM

PR #10071, opened 2023-06-07
[IOTDB-5975] Fix NPE in cpu metrics

THUMarkLau:
See [IOTDB-5975](https://issues.apache.org/jira/browse/IOTDB-5975).

The reason for this bug is that, when a thread is no longer alive, the result of `ThreadMXBean.getThreadInfoById` will return null for this thread. To solve this bug, I check the return value and filter the null value out of the result.


server/src/main/java/org/apache/iotdb/db/service/metrics/DataNodeMetricsHelper.java
```diff
@@ -75,16 +77,20 @@ public static void bind() {
...
-    List<String> pools = new ArrayList<>();
-    Arrays.stream(ThreadName.values()).forEach(x -> pools.add(x.name()));
-    MetricService.getInstance()
-        .addMetricSet(
-            new CpuUsageMetrics(
-                threadModules,
-                pools,
-                x -> ThreadName.getModuleTheThreadBelongs(x).toString(),
-                x -> ThreadName.getThreadPoolTheThreadBelongs(x).name()));
+    if (MetricLevel.higherOrEqual(
```
OneSizeFitsQuorum, line 80: How will we hot-load metricSet like this? @SpriCoder

OneSizeFitsQuorum (approved):
LGTM

PR #10073, opened 2023-06-07
[To rel/1.2] [IOTDB-5975] Fix NPE in cpu metrics

THUMarkLau:
See [IOTDB-5975](https://issues.apache.org/jira/browse/IOTDB-5975).

The reason for this bug is that, when a thread is no longer alive, the result of `ThreadMXBean.getThreadInfoById` will return null for this thread. To solve this bug, I check the return value and filter the null value out of the result.


OneSizeFitsQuorum (approved):
LGTM

PR #10089, opened 2023-06-08
[To rel/1.2][IOTDB-5980] Pipe: drop not existed pipe will return successful result

yschengzi:
This PR contains:
- Change return msg when drop pipe not existed.
  now, IoTDB will return a prompt 
```
String.format(
    "Failed to drop pipe %s. Failures: %s does not exist.", pipeName, pipeName);
```

cherry pick from #10088


SteveYurongSu (changes requested):
keep the same with it on master

SteveYurongSu (approved):

PR #10123, opened 2023-06-12
[To rel/1.2][IOTDB-5938] Fix select into alias series

MarcosZyk:
## Description

There's no source series fetch and check when selecting into alias series. 

Add related schema refetch and data type check.


liuminghui233 (approved):
LGTM :-)

PR #10126, opened 2023-06-12
[metrics] Refactor compaction read write throughput metrics

THUMarkLau:
In the existing Compaction Read-Write metrics, metrics are obtained by recording I/O data through the associated interfaces within the compaction code, after reading or writing a chunk. This process necessitates the inclusion of such monitoring code in all possible I/O operations involved in compaction, which can easily lead to oversight and consequently, inaccuracies in the metrics. This Pull Request establishes compaction-specific writer and reader mechanisms based on TsFileIOWriter and TsFileSequenceReader. These mechanisms automatically quantify I/O data volume before and after the activation of compaction I/O interfaces. Not only does this approach record the data in the metrics system, but it also controls resource utilization during compaction through RateLimiter regulation. Moreover, this PR provides a comprehensive categorization of I/O data, including aligned data, non-aligned data, and metadata. For each type of compaction task, the I/O status of these three types of data is logged individually.

1eb760c6d1222c519f4daa1216314d9


server/src/main/java/org/apache/iotdb/db/engine/compaction/io/CompactionTsFileReader.java
```diff
@@ -0,0 +1,119 @@
...
+
+  @Override
+  public TsFileDeviceIterator getAllDevicesIteratorWithIsAligned() throws IOException {
+    long before = readDataSize.get();
+    TsFileDeviceIterator iterator = super.getAllDevicesIteratorWithIsAligned();
+    long dataSize = readDataSize.get() - before;
+    CompactionMetrics.getInstance()
+        .recordReadInfo(compactionType, CompactionIoDataType.METADATA, dataSize);
+    return iterator;
+  }
```
choubenson, line 106: When recording the amount of metadata read here, there will be less records. The process of compaction reads the device iteratively. At the beginning, it traverses and reads the entire index tree to obtain the offset of the leaf device node, and then reads all the device information in the node according to the offset when compaction.

server/src/main/java/org/apache/iotdb/db/engine/compaction/io/CompactionTsFileWriter.java
```diff
@@ -0,0 +1,120 @@
...
+    long beforeSize = this.getPos();
+    super.endFile();
+    long writtenDataSize = this.getPos() - beforeSize;
+    if (writtenDataSize > 0) {
+      CompactionTaskManager.getInstance().getMergeWriteRateLimiter().acquire((int) writtenDataSize);
+    }
+    CompactionMetrics.getInstance()
+        .recordWriteInfo(type, CompactionIoDataType.METADATA, writtenDataSize);
+  }
+}
```
choubenson, line 120: It seems that there is no record of the amount of data written to the temporary file during the compaction process.

server/src/main/java/org/apache/iotdb/db/engine/compaction/execute/utils/MultiTsFileDeviceIterator.java
```diff
@@ -112,8 +116,13 @@ public MultiTsFileDeviceIterator(
     Collections.sort(
         this.tsFileResourcesSortedByDesc, TsFileResource::compareFileCreationOrderByDesc);
     this.readerMap = readerMap;
+    CompactionType type =
+        !seqResources.isEmpty() && !unseqResources.isEmpty()
+            ? CompactionType.CROSS_COMPACTION
+            : CompactionType.INNER_UNSEQ_COMPACTION;
```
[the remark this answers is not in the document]
THUMarkLau: resolve

PR #10129, opened 2023-06-12
Fix SQL parse of privileges on view and pipe

MarcosZyk:
## Description

same as https://github.com/apache/iotdb/pull/10128


SteveYurongSu (approved):
LGTM

PR #10146, opened 2023-06-14
Validate tsfiles after compaction

shuwenwei:
## Description
Add configuration item 'compaction_validation_level'.
```
# The level of validation after compaction
# The details of these three levels are as follows:
# 1. NONE: the validation after compaction is disabled.
# 2. OVERLAP_ONLY: the validation after compaction check overlap only.
# 3. OVERLAP_AND_FILE: the validation after compaction check overlap and file.
# Datatype: String
# compaction_validation_level=NONE
```

### DIFF
--- a/server/src/main/java/org/apache/iotdb/db/engine/compaction/constant/CompactionValidationLevel.java
+++ b/server/src/main/java/org/apache/iotdb/db/engine/compaction/constant/CompactionValidationLevel.java
@@ -1,0 +1,26 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iotdb.db.engine.compaction.constant;
+
+public enum CompactionValidationLevel {
+  NONE,
+  OVERLAP_ONLY,
+  OVERLAP_AND_FILE
+}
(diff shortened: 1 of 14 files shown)


server/src/main/java/org/apache/iotdb/db/engine/compaction/execute/task/InnerSpaceCompactionTask.java
```diff
@@ -189,9 +189,9 @@ protected boolean doCompaction() {
           compactionLogger.logFile(targetTsFileResource, CompactionLogger.STR_DELETED_TARGET_FILES);
         }
 
-        if (IoTDBDescriptor.getInstance().getConfig().isEnableCompactionValidation()
-            && !CompactionUtils.validateTsFileResources(
-                tsFileManager, storageGroupName, timePartition)) {
+        CompactionValidator validator = CompactionValidator.getInstance();
+        if (!validator.validateCompaction(
+            tsFileManager, targetTsFileList, storageGroupName, timePartition)) {
           LOGGER.error(
```
xingtanzjr, line 195: Add more logs to illustrate why the validation is false.

server/src/main/java/org/apache/iotdb/db/engine/compaction/execute/utils/validator/NoneCompactionValidator.java
```diff
@@ -0,0 +1,47 @@
...
+public class NoneCompactionValidator implements CompactionValidator {
+
+  private NoneCompactionValidator() {}
+
+  public static NoneCompactionValidator getInstance() {
+    return NoneCompactionValidatorHolder.INSTANCE;
+  }
+
+  @Override
+  public boolean validateCompaction(
```
xingtanzjr, line 36: Maybe this method can return a class such as `ValidationResult` to represent the validation-result and detailed message if the validation is false. So that the invoker can add more error handle logic when validation is not passed such as print detailed logs

server/src/main/java/org/apache/iotdb/db/engine/compaction/constant/CompactionValidationLevel.java
```diff
@@ -0,0 +1,26 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iotdb.db.engine.compaction.constant;
+
+public enum CompactionValidationLevel {
```
xingtanzjr, line 22: @qiaojialin Suggest to rename it to `NONE`、`RESOURCE_ONLY`、`RESOURCE_AND_TSFILE`

PR #10162, opened 2023-06-15
[IOTDB-5998] Fix device query bugs in show slot APIs

CRZbulabula:
The show slot APIs with parameter Device=devicePath should filter Database firstly.


l2280212 (approved):
LGTM!

PR #10165, opened 2023-06-15
[To rel/1.2] Refactor Alter View

MarcosZyk:
## Description

Refactor the implementation of alter view to fix the following bugs:
1. NPE when alter an not existed view
2. view will be deleted when altering operation failed
3. return success when some of the target view in one request not exist.

The new implementation is based on procedure and a new WritePlanNode, which won't result existing view deleted.


citrusreticulata (approved):
Excellent work!

PR #10220, opened 2023-06-19
[To rel/1.2][IOTDB-6011] Move default database from IoTDBMetricsUtils to IoTDBConfig

SpriCoder:
It's more suitable to manage default database in IoTDBConfig instead of IoTDBMetricsUtils


THUMarkLau (approved):
LGTM

PR #10231, opened 2023-06-20
[To rel/1.2] Refactor SchemaScanOperator to do asynchronous fetch and return isBlock

Cpaulyz:
## Description

Refactor SchemaScanOperator to do asynchronous fetch and return isBlock.

Modify ISchemaReader's hasNext interface to:
```java
  /**
   * Returns a future that will be completed when the schemaReader becomes unblocked. It may be
   * called several times before next and will return the same value.
   *
   * @return isDone is false if is Blocked.
   */
  ListenableFuture<?> isBlocked();

  boolean hasNext();

  /**
   * The next interface is responsible for consuming the next result.
   *
   * @throws java.util.NoSuchElementException if the current iterator hasNext is false
   */
  T next();
```
And modify modify Operator's isBlocked to an asynchronous fetch procedure if necessary.

* Pros: low overhead of thread switching for mpp scheduling. 
* Cons: 
    * More code modifications.
    * The original hasNext returned a boolean, the hasNextFuture now returns a ListenableFuture<Boolean>, which requires an additional get step. **The impact is minimal.**


server/src/main/java/org/apache/iotdb/db/metadata/query/reader/SchemaReaderLimitOffsetWrapper.java
```diff
@@ -64,19 +60,50 @@ public void close() throws Exception {
...
+    } catch (Exception e) {
+      throw new RuntimeException(e);
     }
   }
 
   @Override
   public T next() {
-    if (!hasNext()) {
-      throw new NoSuchElementException();
-    }
```
MarcosZyk, line 79: Keep this hasNext check

server/src/main/java/org/apache/iotdb/db/metadata/query/reader/ISchemaReader.java
```diff
@@ -21,9 +21,17 @@
...
+import static com.google.common.util.concurrent.Futures.immediateFuture;
+import static com.google.common.util.concurrent.Futures.immediateVoidFuture;
+
+public interface ISchemaReader<T extends ISchemaInfo> extends AutoCloseable {
+
+  ListenableFuture<Boolean> NOT_BLOCKED_TRUE = immediateFuture(true);
+  ListenableFuture<Boolean> NOT_BLOCKED_FALSE = immediateFuture(false);
+  ListenableFuture<Void> NOT_BLOCKED = immediateVoidFuture();
 
-public interface ISchemaReader<T extends ISchemaInfo> extends Iterator<T>, AutoCloseable {
```
MarcosZyk, line 26: Keep extends Iterator<T>

server/src/main/java/org/apache/iotdb/db/metadata/query/reader/SchemaReaderLimitOffsetWrapper.java
```diff
@@ -32,20 +34,14 @@
   private final boolean hasLimit;
 
   private int count = 0;
-  int curOffset = 0;
+  private int curOffset = 0;
+  private ListenableFuture<?> hasNextFuture = null;
```
MarcosZyk, line 38: Rename to isBlocked to keep consistent that in Operators

server/src/main/java/org/apache/iotdb/db/metadata/query/reader/SchemaReaderLimitOffsetWrapper.java
```diff
@@ -64,19 +60,50 @@ public void close() throws Exception {
...
-  public boolean hasNext() {
+  public ListenableFuture<?> isBlocked() {
+    if (hasNextFuture != null) {
+      return hasNextFuture;
+    }
+    hasNextFuture = hasNextFuture();
+    return hasNextFuture;
+  }
+
+  private ListenableFuture<?> hasNextFuture() {
```
MarcosZyk, line 71: Rename to tryGetNext

server/src/main/java/org/apache/iotdb/db/metadata/query/reader/TimeseriesReaderWithViewFetch.java
```diff
@@ -45,12 +51,13 @@
...
 
 public class TimeseriesReaderWithViewFetch implements ISchemaReader<ITimeSeriesSchemaInfo> {
-
+  private static final Logger LOGGER = LoggerFactory.getLogger(TimeseriesReaderWithViewFetch.class);
   private final Traverser<ITimeSeriesSchemaInfo, ?> traverser;
   private final Queue<ITimeSeriesSchemaInfo> cachedViewList = new ArrayDeque<>();
   private ITimeSeriesSchemaInfo next = null;
   private boolean consumeView = false;
   private final SchemaFilter schemaFilter;
+  private ListenableFuture<Boolean> hasNextFuture = null;
```
MarcosZyk, line 60: Rename to isBlocked.

server/src/main/java/org/apache/iotdb/db/metadata/schemaregion/ISchemaRegion.java
```diff
@@ -293,8 +293,8 @@ ISchemaReader<IDeviceSchemaInfo> getDeviceReader(IShowDevicesPlan showDevicesPla
       throws MetadataException;
 
   /**
-   * The iterated result shall be consumed before calling reader.hasNext() or reader.next(). Its
-   * implementation is based on the reader's process context.
+   * The iterated result shall be consumed before calling reader.hasNextFuture() or reader.next().
```
MarcosZyk, line 296: Fix this annotation. Describe these three interfaces usage, isBlocked, hasNext and next.

MarcosZyk (commented):
PTAL

server/src/main/java/org/apache/iotdb/db/metadata/query/reader/ISchemaReader.java
Cpaulyz: We only use the interfaces hasNext and next, with the addition of isBlocked and no support for operations such as reset, so I think we can leave the Iterator unimplemented.

Also, this makes it easier to do static call analysis in IDEA.

server/src/main/java/org/apache/iotdb/db/metadata/schemaregion/ISchemaRegion.java
Cpaulyz: move interfaces usage description to `ISchemaReader`

MarcosZyk (approved):

PR #10261, opened 2023-06-21
Adapt superset's time aggregation queries to iotdb syntax

linxt20:
## Description:
This pull request modifies the behavior of superset when handling time columns in the x-axis or aggregate columns. Currently, superset defaults to using time-based aggregation segments for aggregate functions, employing the syntax "GROUP BY TRUNC(interval, 'Time')". However, iotdb utilizes a different syntax for time-based aggregation segments, which is "GROUP BY ([startTime, endTime), interval)". This creates a conflict in syntax between superset and iotdb. This PR addresses this issue by making the necessary modifications to iotdb, enabling compatibility with time-based aggregation queries in superset.

## Effect:
image

## Implementation:
The implementation involves adding a flag in the group_by_clause of sqlalchemy/IOTDBSQLcompiler.py to indicate the time-based aggregation segment operation as "GROUP BY ([startTime, endTime), 1s)". Additionally, it converts the interval in superset's syntax to iotdb's syntax. Furthermore, in the execute function of dbapi/Cursor.py, the SQL statement is intercepted before it is sent to the iotdb server. The code then retrieves the maximum timestamp of the base table using the execution function, with the aggregate functions changed to MAX_VALUE. Similarly, it retrieves the minimum timestamp by switching the functions to MIN_VALUE. The startTime and endTime placeholders in the flagged statement are then replaced with the respective timestamps using regex matching. Finally, the modified GROUP BY statement is appended to the query SQL statement.

### DIFF
--- a/iotdb-client/client-py/iotdb/dbapi/Cursor.py
+++ b/iotdb-client/client-py/iotdb/dbapi/Cursor.py
@@ -18,7 +18,8 @@
 
 import logging
 import warnings
-
+import re
+import pandas as pd
 from iotdb.Session import Session
 
 from .Exceptions import ProgrammingError
@@ -112,6 +113,9 @@ class Cursor(object):
 
         time_index = []
         time_names = []
+        time_group_sql = ""
+        time_group_flag = False
+
         if self.__sqlalchemy_mode:
             sql_seqs = []
             seqs = sql.split("\n")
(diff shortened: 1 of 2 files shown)


iotdb-client/client-py/iotdb/dbapi/Cursor.py
```diff
@@ -125,10 +129,49 @@ def execute(self, operation, parameters=None):
...
+                max_sql_seqs = []
+                temp_seqs = temp_sql.split("\n")
+                for seq in temp_seqs:
+                    if seq.find("SELECT") >= 0:
+                        min_seq = seq
+                        max_seq = seq
+                        aggregate_functions = ["count", "max_value", "min_value", "sum", "avg"]
+                        for func in aggregate_functions:
+                            max_seq = re.sub(r"\b" + func + r"\b", "MAX_TIME",  max_seq)
+                            min_seq = re.sub(r"\b" + func + r"\b", "MIN_TIME",  min_seq)
```
Cpaulyz, line 151: This should be case insensitive.
```suggestion
                        aggregate_functions = ["count", "max_value", "min_value", "sum", "avg"]
                        for func in aggregate_functions:
                            max_seq = re.sub(r"\b" + func + r"\b", "MAX_TIME",  max_seq, flags=re.IGNORECASE)
                            min_seq = re.sub(r"\b" + func + r"\b", "MIN_TIME",  min_seq, flags=re.IGNORECASE)
```

iotdb-client/client-py/iotdb/dbapi/Cursor.py
```diff
@@ -125,10 +129,49 @@ def execute(self, operation, parameters=None):
...
+                min_sql = "\n".join(min_sql_seqs)
+                max_sql = "\n".join(max_sql_seqs)
+                min_data_set = self.__session.execute_statement(min_sql)
+                min_data = pd.DataFrame(min_data_set.todf())
+                min_value = min_data.iloc[0, 0]
+                max_data_set = self.__session.execute_statement(max_sql)
+                max_data = pd.DataFrame(max_data_set.todf())
+                max_value = max_data.iloc[0, 0]
+                time_group_sql = re.sub(r"startTime", str(min_value), time_group_sql)
+                time_group_sql = re.sub(r"endTime", str(max_value), time_group_sql)
```
Cpaulyz, line 166: Other aggregations also need to find starttime and endtime, taking the smallest starttime and endtime as the result for replacement.

For example,
```
SELECT AVG(_93741819) AS `AVG(_93741819)`,
       AVG(_24646171) AS `AVG(_24646171)`,
       AVG(_03380912) AS `AVG(_03380912)`
FROM root.view.speed.hebei
```
In the current implementation, we will only translate the first one (`if seq.find("SELECT") >= 0`). That is, set starttime according to MIN_TIME(_93741819)

However, we should translate it to 
```
select MIN_TIME(_93741819), MIN_TIME(_24646171), MIN_TIME(_03380912) from root.view.fuel.hebei
```
And then choose the smallest time as starttime.

Cpaulyz (changes requested):
Bravo work! Thanks for contributing to IoTDB, here are two suggestions, PTAL~

PR #10276, opened 2023-06-23
[IOTDB-5937] Fix wrong measurement check when using view to import csv

MarcosZyk:
## Description

same as https://github.com/apache/iotdb/pull/10275


citrusreticulata (commented):
If there is no need to check duplications of measurements in tablets or rows, then there is no need to use the function `validateInsertTabletList`. 

This is because the duplication of each `insertRowStatement` or `insertTabletStatement` has already been checked through the function `validateMapFromDeviceToMeasurement` in function `getMapFromDeviceToMeasurementAndIndex`.

MarcosZyk:
> If there is no need to check duplications of measurements in tablets or rows, then there is no need to use the function `validateInsertTabletList`.
> 
> This is because the duplication of each `insertRowStatement` or `insertTabletStatement` has already been checked through the function `validateMapFromDeviceToMeasurement` in function `getMapFromDeviceToMeasurementAndIndex`.

Thanks for suggestion. The modification has already been updated as removing useless check cross insert statements.

PR #10277, opened 2023-06-23
[To rel/1.2] Fix slow creation of view when using batch creation sql

MarcosZyk:
### Cause

When constructing pattern tree for partition fetch from configNode, the interface invoked was appendPathPattern, which takes n^2 time complexity.

### Solution

When the source clause of createView is a query statement, use the schema tree fetched during analyze query of createViewStatement, and this already contains all source timeseries.

When  the source clause of createView is not a query statement, use appendFullPath to construct pattern tree, since all the given paths are precise timeseries paths.


citrusreticulata (approved):
Excellent work!

PR #10278, opened 2023-06-23
Fix slow creation of view when using batch creation sql

MarcosZyk:
## Description

same as https://github.com/apache/iotdb/pull/10277


citrusreticulata (approved):
Excellent work!

PR #10295, opened 2023-06-25
Revert "[To rel/1.2][IOTDB-6022] Modify the default parameters of iotconsensus"

qiaojialin:
Reverts apache/iotdb#10267. The write performance degrades after 24h does not due to this change, but the environment(jdk and gc strategy).


qiaojialin:
Do not need revert.

PR #10307, opened 2023-06-25
Add cpu usage metrics for confignode

THUMarkLau:
image

This PR add cpu usage metrics for confignode, which enables us to gain insight of the cpu occupation of each thread pool and modules in confignode.


metrics/interface/src/main/java/org/apache/iotdb/metrics/metricsets/cpu/CpuUsageMetrics.java
```diff
@@ -39,6 +42,7 @@
 import java.util.stream.Collectors;
 
 public class CpuUsageMetrics implements IMetricSet {
+  private static final Logger log = LoggerFactory.getLogger(CpuUsageMetrics.class);
```
OneSizeFitsQuorum, line 45: user logger

metrics/interface/src/main/java/org/apache/iotdb/metrics/metricsets/cpu/CpuUsageMetrics.java
```diff
@@ -211,6 +217,11 @@ private void updateCpuUsage() {
     lastThreadCpuTime.putAll(currentThreadCpuTime);
     lastThreadUserTime.clear();
     lastThreadUserTime.putAll(currentThreadUserTime);
+    long timeCost = System.nanoTime() - startTime;
+    updateCount.incrementAndGet();
+    if (updateCount.get() % 10 == 0) {
+      log.info("Time for update cpu usage is {} ns", timeCost);
```
OneSizeFitsQuorum, line 223: I recommend changing it to debug logs. If we need to check the time, we can adjust the logback, otherwise the log is too much

CRZbulabula (approved):
LGTM! Learned a lot from Dr.Liu's code

OneSizeFitsQuorum (approved):
LGTM

PR #10317, opened 2023-06-26
Update TsFileOutputFormat.java

asia-zengtao:
Exception handling is improved by providing specific error messages when dealing with URISyntaxException


.github/workflows/build.yml
HTHou, line 1: Consider remove this workflow file?

pom.xml
```diff
@@ -124,6 +124,8 @@
     </modules>
     <!-- Properties Management -->
     <properties>
+        <sonar.organization>asia-zengtao</sonar.organization>
+        <sonar.host.url>https://sonarcloud.io</sonar.host.url>
```
HTHou, line 128: remove it?

asia-zengtao: Ok I will remove it! thanks !

.github/workflows/build.yml
HTHou, line 1: Empty file?

PR #10327, opened 2023-06-26
[IOTDB-6028] Fix SonarCloud code smells in consensus module

BUAAserein:


consensus/src/main/java/org/apache/iotdb/consensus/ratis/ApplicationStateMachineProxy.java
```diff
@@ -163,7 +164,7 @@ public CompletableFuture<Message> applyTransaction(TransactionContext trx) {
       } catch (InterruptedException i) {
         logger.warn("{} interrupted when retry sleep", this);
         Thread.currentThread().interrupt();
-      } catch (Throwable rte) {
```
SzyWilliam, line 166: Throwable here is better. We need to catch RunTimeExceptions so that the SM can still works under some rare scenarios.

SzyWilliam (changes requested):

PR #10361, opened 2023-06-28
[IOTDB-6033] Fix SonarCloud code smells in confignode module

BUAAserein:
#### Fix SonarCloud code smells in confignode module.


OneSizeFitsQuorum (approved):
LGTM

PR #10390, opened 2023-06-29
UT: add unit tests for sync/pipe/load packages in datanode & node-commons

DanielWang2035:


iotdb-core/datanode/src/test/java/org/apache/iotdb/db/pipe/extractor/PipeRealtimeExtractTest.java
```diff
@@ -102,12 +104,11 @@ public void tearDown() {
   public void testRealtimeExtractProcess() {
     // set up realtime extractor
 
-    try (PipeRealtimeDataRegionHybridExtractor extractor1 =
-            new PipeRealtimeDataRegionHybridExtractor();
+    try (PipeRealtimeDataRegionLogExtractor extractor1 = new PipeRealtimeDataRegionLogExtractor();
```
DanielWang2035, line 107: Will this change the logic of this test?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/load/LoadSingleTsFileNode.java
```diff
@@ -151,7 +152,7 @@ public TRegionReplicaSet getRegionReplicaSet() {
 
   @Override
   public List<PlanNode> getChildren() {
-    return null;
+    return Collections.emptyList();
```
DanielWang2035, line 155: I found that if this function returns null here:
- In Windows, it returns null.
- In Windows Server and Unix, it returns an empty list.

So I made it return an empty list explicitly.

iotdb-core/datanode/src/test/java/org/apache/iotdb/db/pipe/extractor/PipeRealtimeExtractTest.java
SteveYurongSu: No. I think this is a really GOOD way to do tests :)

SteveYurongSu (approved):

PR #10417, opened 2023-07-03
[IOTDB-5589]The "IoTDB-Compaction_Schedule" thread count needs to be controlled

shuwenwei:
## Description
This pull request solve the conflicts of https://github.com/apache/iotdb/pull/10257

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/tsfile/TsFileManager.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/tsfile/TsFileManager.java
@@ -19,10 +19,10 @@
 
 package org.apache.iotdb.db.storageengine.dataregion.tsfile;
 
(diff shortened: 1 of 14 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/tsfile/TsFileManager.java
```diff
@@ -319,12 +322,26 @@ public Set<Long> getTimePartitions() {
...
+    CompactionTaskManager.getInstance().register(this);
+  }
+
   public boolean isAllowCompaction() {
     return allowCompaction;
   }
 
   public void setAllowCompaction(boolean allowCompaction) {
     this.allowCompaction = allowCompaction;
+    if (!allowCompaction) {
```
xingtanzjr, line 341: What if the `setAllowCompaction(true)` is invoked after the thread pool is shutdown ?

PR #10418, opened 2023-07-03
Fix semantic incompatibility when AtomicReference replaces volatile

OneSizeFitsQuorum:
The previous code might cause an error like this
img_v2_1cd40617-e4b9-4602-b083-4bb3d3b20bcg


BUAAserein (approved):
LGTM

PR #10423, opened 2023-07-03
[IOTDB-6042] Pipe: fix CreationTime display error using non-default time_precision when executing show pipes

DanielWang2035:
see https://timechor.feishu.cn/docx/Hhqbd84saoPTmHxCXuCcHQKgnBg
see https://issues.apache.org/jira/browse/IOTDB-6042
after (timestamp_precision=ns):
image

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/DateTimeUtils.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/DateTimeUtils.java
@@ -665,17 +665,21 @@ public class DateTimeUtils {
     }
   }
 
-  public static String convertLongToDate(long timestamp) {
-    String timePrecision = CommonDescriptor.getInstance().getConfig().getTimestampPrecision();
-    switch (timePrecision) {
+  public static String convertLongToDate(long timestampInMs) {
+    return convertLongToDate(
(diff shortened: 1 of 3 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/DateTimeUtils.java
```diff
@@ -665,17 +665,21 @@ public static long currentTime() {
...
 
-  public static String convertLongToDate(long timestamp) {
-    String timePrecision = CommonDescriptor.getInstance().getConfig().getTimestampPrecision();
-    switch (timePrecision) {
+  public static String convertLongToDate(long timestampInMs) {
+    return convertLongToDate(
+        timestampInMs, CommonDescriptor.getInstance().getConfig().getTimestampPrecision());
+  }
+
+  public static String convertLongToDate(long timestampInMs, String targetPrecision) {
```
SteveYurongSu, line 673: It seems that the name of timestampInMs is not correct?

DanielWang2035: Fixed.

SteveYurongSu (approved):
LGTM

PR #10430, opened 2023-07-04
[IOTDB-5576] Some node missed out when deleting after updating

Cpaulyz:
## Description

Fix bug occurs in `SchemaStatisticsTest`

https://apache-iotdb.feishu.cn/docx/Ykibdd0ehox4qSxsPOAcc8JGnyg

image

After troubleshooting, this was caused by PBTree memory management in the corner case of update then delete.

When updating measurement, PBTree management will remove the device from the cache and add it to the buffer.

When removing measurement, PBTree management removes the device's node from the device's updated buffer.

Therefore, when flushing, although the device node is regarded as volatile not, it will not be collected because it has no children. We need to add it back to the cache, otherwise the device node will not be unpinable, resulting in incorrect statistics.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/impl/pbtree/cache/CacheManager.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/impl/pbtree/cache/CacheManager.java
@@ -287,17 +287,20 @@ public abstract class CacheManager implements ICacheManager {
   }
 
   private void collectVolatileNodes(ICachedMNode node, List<ICachedMNode> nodesToPersist) {
-    if(node.isMeasurement()){
+    if (node.isMeasurement()) {
       return;
     }
     Iterator<ICachedMNode> bufferIterator =
         getCachedMNodeContainer(node).getChildrenBufferIterator();
-    nodesToPersist.add(node);
-
-    ICachedMNode child;
-    while (bufferIterator.hasNext()) {
-      child = bufferIterator.next();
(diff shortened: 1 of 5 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/impl/pbtree/cache/CacheManager.java
```diff
@@ -287,17 +287,26 @@ public List<ICachedMNode> collectVolatileMNodes() {
...
+        collectVolatileNodes(child, nodesToPersist);
+      }
+    } else {
+      // add back to cache
+      CacheEntry cacheEntry = getCacheEntry(node);
+      while (cacheEntry != null && !isInNodeCache(cacheEntry)) {
+        addToNodeCache(cacheEntry, node);
+        node = node.getParent();
+        cacheEntry = getCacheEntry(node);
+      }
```
MarcosZyk, line 309: Maybe we can do this when executing deleting node.

MarcosZyk (commented):
Excellent work! Only one potential optimization. PTAL

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/impl/pbtree/cache/CacheManager.java
MarcosZyk: It's costing to judge whether a node has some child in buffer. Update the status during flushing is the fastest way.

MarcosZyk (approved):

PR #10444, opened 2023-07-04
[To rel/1.2] Fix npe exception when invalid in metric module

SpriCoder:
Fix npe exception when invalid in metric module


OneSizeFitsQuorum (approved):
LGTM

PR #10448, opened 2023-07-05
[IOTDB-6046] Refactor Ratis metrics Reporting in read/write

SzyWilliam:
Refactor Ratis metrics Reporting in read/write. Use AutoClosable to simplify code.

### DIFF
--- a/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/metrics/RatisMetricsManager.java
+++ b/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/metrics/RatisMetricsManager.java
@@ -26,12 +26,50 @@ import org.apache.iotdb.commons.service.metric.enums.Tag;
 import org.apache.iotdb.metrics.utils.MetricLevel;
 
 import java.util.concurrent.TimeUnit;
+import java.util.function.BiConsumer;
 
 public class RatisMetricsManager {
+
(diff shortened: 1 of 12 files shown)


iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/metrics/RatisMetricsManager.java
```diff
@@ -26,12 +26,50 @@
...
 
 public class RatisMetricsManager {
+
+  static class TimeKeeper implements AutoCloseable {
+    private final long startMoment;
+    private final TConsensusGroupType groupType;
+    private final BiConsumer<Long, TConsensusGroupType> reporter;
+
+    private TimeKeeper(
+        BiConsumer<Long, TConsensusGroupType> reporter, TConsensusGroupType groupType) {
```
OneSizeFitsQuorum, line 39: Can we find a way to avoid the cost of boxing or unboxing long？

SzyWilliam: primitive types are not allowed for Functional interfaces. Besides, the boxing & unboxing costs are trivial in metric reporting operation.

OneSizeFitsQuorum (approved):
LGTM

PR #10450, opened 2023-07-05
Fix metric service cannot start

HTHou:
## Description

https://github.com/apache/iotdb/pull/10429 removed some netty dependencies. However, some of them are necessary for metrics module. When open the PROMETHEUS reporter, the confignode cannot start.
1nqXPHuwNi

Therefore, this PR is recovering the necessary dependencies.


OneSizeFitsQuorum (approved):
LGTM

PR #10457, opened 2023-07-05
[IOTDB-6051] Fixed concurrency error in IoTConsensus UT when stopping cluster

OneSizeFitsQuorum:
see [jira](https://issues.apache.org/jira/browse/IOTDB-6051)


BUAAserein (approved):
LGTM

PR #10458, opened 2023-07-05
[To rel/1.1] Fix not decreasing file size after ttl delete file

THUMarkLau:
# Overview
In IoTDB, the size and count of all deleted TsFiles should be deducted in TsFileMetrics. Users have noticed discrepancies between the TsFile size and count displayed on the monitoring panel and the actual file size and count on the disk. This inconsistency is due to the fact that the changes in file size and count were not recorded in TsFileMetrics when files were deleted due to TTL (Time-To-Live) expiry.
# How to reproduce the bug 
Launch an IoTDB instance and write data to it while setting the TTL. Once the TTL takes effect, you will notice the mismatch between the data reflected in the monitor and the actual data on the disk.


OneSizeFitsQuorum (approved):
LGTM

PR #10480, opened 2023-07-07
added the selection of files to be merged based on mods file size in the inner compaction selector

caozj1011:
If the size of the corresponding mods file is larger than 50M, we will preferentially select this file to submit to the global priority queue. If a suitable task is selected, the hierarchical selection will not be executed. In the inner sorting rules, sorting by mods file size has the highest priority.

See documentation link for testing procedure：
https://apache-iotdb.feishu.cn/docx/HLCPdkY5LohWN9xTZIZcEE0yn0g

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/selector/impl/SizeTieredCompactionSelector.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/selector/impl/SizeTieredCompactionSelector.java
@@ -67,6 +67,7 @@ public class SizeTieredCompactionSelector
   protected boolean sequence;
   protected TsFileManager tsFileManager;
   protected boolean hasNextTimePartition;
+  private static final long MODS_FILE_SIZE_THRESHOLD = 1024 * 1024 * 50L;
 
   public SizeTieredCompactionSelector(
       String storageGroupName,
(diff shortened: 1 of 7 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/selector/impl/SizeTieredCompactionSelector.java
```diff
@@ -165,12 +170,17 @@ public List<List<TsFileResource>> selectInnerSpaceTask(List<TsFileResource> tsFi
     PriorityQueue<Pair<List<TsFileResource>, Long>> taskPriorityQueue =
         new PriorityQueue<>(new SizeTieredCompactionTaskComparator());
     try {
-      int maxLevel = searchMaxFileLevel();
-      for (int currentLevel = 0; currentLevel <= maxLevel; currentLevel++) {
-        if (!selectLevelTask(currentLevel, taskPriorityQueue)) {
-          break;
+      selectMaxModsFileTask(taskPriorityQueue);
```
xingtanzjr, line 173: One suggestion is that it is better to make every method as `pure function`, that is, do not pass the variable `taskPriorityQueue` into the method and add element into it inside the method. 

We can change it to something like `taskPriorityQueue.addAll(selectMaxModsFileTask())`

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/selector/impl/SizeTieredCompactionSelector.java
```diff
@@ -183,6 +193,18 @@ public List<List<TsFileResource>> selectInnerSpaceTask(List<TsFileResource> tsFi
     return Collections.emptyList();
   }
 
+  private void selectMaxModsFileTask(
+      PriorityQueue<Pair<List<TsFileResource>, Long>> taskPriorityQueue) {
+    for (TsFileResource tsFileResource : tsFileResources) {
+      ModificationFile modFile = tsFileResource.getModFile();
+      if (!Objects.isNull(modFile) && modFile.getSize() > 1024 * 1024 * 50) {
```
xingtanzjr, line 200: Use a static final var to represent the threshold rather than hard code here.

PR #10487, opened 2023-07-07
[IOTDB-6056] Pipe: Failed to load tsfile with empty pages (NPE occurs when loading)

MiniSho:
**bug**：
image
image

**reason**：
we didn't handle empty pages in aligned chunk

**solution**:
add `if` statement to judge the page is empty or not, and write the entire page when it is not an empty page


yschengzi (approved):
LGTM

SteveYurongSu (approved):

PR #10514, opened 2023-07-11
Fix the preoblem when using select into with view path which is same as base path.

Beyyes:
## Description

The situation is as below, when the source columns have both base and view paths, and all these paths indicate same base paths.

a4746553-4455-45b0-a2ed-59d5a2006b12

image


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/process/IntoOperator.java
```diff
@@ -42,6 +44,8 @@ public class IntoOperator extends AbstractIntoOperator {
 
   private final List<Pair<String, PartialPath>> sourceTargetPathPairList;
 
+  private final List<Pair<String, String>> sourceColumnToViewList;
```
JackieTien97, line 47: It seems that `List<String> sourceColumnToViewList` is enough.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/parameter/IntoPathDescriptor.java
```diff
@@ -47,6 +49,9 @@ public class IntoPathDescriptor {
   // List<(sourceColumn, targetPath)>
   private List<Pair<String, PartialPath>> sourceTargetPathPairList;
 
+  // sourceColumn -> viewPath, if the viewPath is not empty
+  private List<Pair<String, String>> sourceColumnToViewList;
```
JackieTien97, line 53: It seems that `List<String> sourceColumnToViewList` is enough.

JackieTien97 (changes requested):
Think more about the align by device + select into

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/parameter/IntoPathDescriptor.java
Beyyes: fixed

JackieTien97 (approved):

PR #10526, opened 2023-07-12
Fix toString method of FunctionViewExpression

Beyyes:
## Description

Before the tostring is like `FUNC(type=INT32, key=value)(root.db.device.s01, root.ln.d.s01)`,
Now it's like `FUNC(root.db.device.s01, root.ln.d.s01, type=INT32, key=value)`


Cpaulyz (approved):
LGTM～～

PR #10532, opened 2023-07-12
Use customer timeout in schema fetcher 

Cpaulyz:
## Description

Use customer timeout in schema fetcher


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/analyze/schema/ClusterSchemaFetchExecutor.java
```diff
@@ -46,42 +52,54 @@
...
+  private ExecutionResult executionStatement(
+      Long queryId, Statement statement, MPPQueryContext context) {
+    return coordinator.execute(
+        statement,
+        queryId,
+        context == null ? null : context.getSession(),
+        "",
+        ClusterPartitionFetcher.getInstance(),
+        schemaFetcher,
+        context == null || context.getQueryType().equals(QueryType.WRITE)
```
JackieTien97, line 85: write should never  have a timeout.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/analyze/schema/AutoCreateSchemaExecutor.java
```diff
@@ -54,21 +59,33 @@
...
     this.templateManager = templateManager;
-    this.statementExecutor = statementExecutor;
+    this.schemaFetcher = schemaFetcher;
+  }
+
+  private ExecutionResult executeStatement(Statement statement) {
+    return coordinator.execute(
+        statement,
+        SessionManager.getInstance().requestQueryId(),
+        null,
```
JackieTien97, line 84: try to gey `SessionInfo` here.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/analyze/schema/AutoCreateSchemaExecutor.java
```diff
@@ -54,21 +59,33 @@
...
+        "",
+        ClusterPartitionFetcher.getInstance(),
+        schemaFetcher,
+        config.getQueryTimeoutThreshold());
```
JackieTien97, line 88: get timeout from context.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/analyze/schema/ClusterSchemaFetchExecutor.java
JackieTien97: Actually, we have already set the timeout field in `Context` for write to `Long.MAX_VALUE` in Coordinator.

JackieTien97 (changes requested):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/analyze/schema/ClusterSchemaFetchExecutor.java
Cpaulyz: > write should never have a timeout.

Yes, the statement must be query here but context may be WRITE type when inserting. So I use default timeout here.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/analyze/schema/AutoCreateSchemaExecutor.java
Cpaulyz: This method will only be called when writing (for automatic schema creation).The type of context is WRITE, and its timeout is set to LONG.MAX_VALUE, which doesn't make any sense.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/analyze/schema/AutoCreateSchemaExecutor.java
Cpaulyz: The statement here must be write type. SessionInfo seems to be useless in the write process.

JackieTien97 (approved):

PR #10554, opened 2023-07-14
Pipe: Add a configuration parameter to control the number of rows of the tablet used within the pipe module

DanielWang2035:
Currently, tablets in pipe transfer are created with default rowSize. If data fransfer load is high, there  will be too many thrift requests of tablet transfer. So we can add a knob to specify rowSize of pipe tablets.

### DIFF
--- a/iotdb-core/node-commons/src/assembly/resources/conf/iotdb-common.properties
+++ b/iotdb-core/node-commons/src/assembly/resources/conf/iotdb-common.properties
@@ -972,6 +972,9 @@ cluster_name=defaultCluster
 # Noted that: this should be less than or equals to realtimeExtractorPendingQueueCapacity
 # pipe_extractor_pending_queue_tablet_limit=64
 
+# The row size of tablets created in the extractor stage.
+# pipe_extractor_tablet_row_size=65536
+
 # The buffer size used for reading file during file transfer.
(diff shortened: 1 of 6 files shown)


iotdb-core/node-commons/src/assembly/resources/conf/iotdb-common.properties
```diff
@@ -972,6 +972,9 @@ cluster_name=defaultCluster
 # Noted that: this should be less than or equals to realtimeExtractorPendingQueueCapacity
 # pipe_extractor_pending_queue_tablet_limit=64
 
+# The row size of tablets created in the extractor stage.
+# pipe_extractor_tablet_row_size=65536
```
SteveYurongSu, line 976:
```suggestion
# pipe_datastructure_tablet_row_size=65535
```

SteveYurongSu (changes requested):
Only a minor issue about naming

iotdb-core/node-commons/src/assembly/resources/conf/iotdb-common.properties
DanielWang2035: changed

SteveYurongSu (approved):

SteveYurongSu:
Please cp this commit to rel/1.2, thanks!

PR #10556, opened 2023-07-14
[IOTDB-6068] Pipe: Failed to load tsfile with first page is empty in a chunk

MiniSho:
in previous pr: https://github.com/apache/iotdb/pull/10487, We only considered the case where the empty page is at the end of an aligned chunk, and it's not well thought out. Empty page is skipped, causing problems with rewriting tsfile on datanode

case not considered: 
image

in this pr, we consider all the position of empty pages in an aligned chunk, and handle the empty page through write an empty page into the chunk.


yschengzi (approved):
LGTM

SteveYurongSu (approved):

PR #10558, opened 2023-07-16
[IOTDB-6069] Support SessionPool in client-py

ycycse:
This PR implement a simplify  session pool for python client.
```python
"""
 PoolConfig(host: str, ip: str, user_name: str, password: str, node_urls: list = None,
                 fetch_size: int = DEFAULT_FETCH_SIZE, time_zone: str = DEFAULT_TIME_ZONE,
                 max_retry: int = DEFAULT_MAX_RETRY)
"""
pool_config = PoolConfig(db.get_ip_address(), db.get_port(), [], "root", "root", 1024, "Asia/Shanghai", 3)

"""
create_session_pool(pool_config: PoolConfig, max_pool_size: int, wait_timeout_in_ms: int)
"""
session_pool = create_session_pool(pool_config, max_pool_size, 3000)

# Try to get session from pool, if there isn't enough connection, a error about timeout will be raised
try:
    session = session_pool.get_session()
except TimeoutError as e:
    raise e
session.open(False)
session_pool.put_back(session)
```

### DIFF
--- a/iotdb-client/client-py/iotdb/SessionPool.py
+++ b/iotdb-client/client-py/iotdb/SessionPool.py
@@ -18,7 +18,7 @@
 
 import multiprocessing
 import time
-from queue import Queue
+from multiprocessing import Queue
 from threading import Lock
 
 from iotdb.Session import Session
@@ -114,6 +114,7 @@ class SessionPool(object):
         while not self.__queue.empty():
             session = self.__queue.get(block=False)
             session.close()
+            self.__pool_size -= 1
 
 
 def create_session_pool(pool_config: PoolConfig, max_pool_size: int, wait_timeout_in_ms: int) -> SessionPool:
--- a/iotdb-client/client-py/iotdb/SessionPool.py
+++ b/iotdb-client/client-py/iotdb/SessionPool.py
@@ -1,0 +1,121 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+import multiprocessing
+import time
+from queue import Queue
+from threading import Lock
+
+from iotdb.Session import Session
+
+
+class PoolConfig:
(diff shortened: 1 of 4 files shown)


HTHou (approved):

iotdb-client/client-py/iotdb/SessionPool.py
```diff
@@ -0,0 +1,122 @@
...
+                self.__pool_size += 1
+                should_create = True
+                self.__lock.release()
+                break
+            else:
+                if time.time() - start > self.__wait_timeout_in_ms:
+                    self.__lock.release()
+                    raise TimeoutError("Wait to get session timeout in SessionPool, current pool size: {0}"
+                                       .format(self.__max_pool_size))
+                time.sleep(1)
```
JackieTien97, line 90: sleep for 1 second? Make sure default time unit in python sleep is second.

iotdb-client/client-py/iotdb/SessionPool.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+class SessionPool(object):
+
+    def __init__(self, pool_config: PoolConfig, max_pool_size: int, wait_timeout_in_ms: int):
+        self.__pool_config = pool_config
+        self.__max_pool_size = max_pool_size
+        self.__wait_timeout_in_ms = wait_timeout_in_ms / 1000
+        self.__pool_size = 0
+        self.__queue = Queue(max_pool_size)
```
JackieTien97, line 55: Is this `Queue` a thread-safe class?

iotdb-client/client-py/iotdb/SessionPool.py
```diff
@@ -0,0 +1,122 @@
...
+            self.__queue.put(session)
+        else:
+            self.__lock.acquire()
+            self.__pool_size -= 1
+            self.__lock.release()
+
+    def close(self):
+        while not self.__queue.empty():
+            session = self.__queue.get(block=False)
+            session.close()
```
JackieTien97, line 116: don't need to update `__pool_size`?

JackieTien97 (changes requested):

iotdb-client/client-py/iotdb/SessionPool.py
ycycse: Yes, the unit of time.sleep() is second. 
https://docs.python.org/3/library/time.html?highlight=sleep#time.sleep

iotdb-client/client-py/iotdb/SessionPool.py
ycycse: It's thread-safe, but I found it not process-safe, so it should be changed.

iotdb-client/client-py/iotdb/SessionPool.py
ycycse: This closes the sessionPool, so I think it's not necessary to update pool_size.

JackieTien97: But there is no guarantee that users won't use the closed SessionPool. What if they use it after closing it?

ycycse: I see. I have added a `closed` status for SessionPool

JackieTien97 (approved):

PR #10566, opened 2023-07-17
Optimize the speed of delete by pattern

HTHou:
## Description

When IoTDB has data of 100,000 devices, 20k devices in each tsfile, 163 tsfiles total, execute a delete sql like `delete from root.db.** where time<1970-1-01T08:01:01.000+08:00`. No data actually matches the time filter.  The execution will takes over **20 hours**. 

The reason of this issue is the process of matching DeviceId in TsFileResource and the pattern `root.db.**` takes too match time. For this scenario , it will match 20k * 163 = 3,260,000 times.

This PR is fixing the above issue. We can use a hashset to record the match deviceId info for this delete operation. The maximum matching times will be reduced to the device number 100,000.

Plus, we can use a file level time filter to skip the device check before the deviceId matching.

After this fixing, the execution only takes **8 seconds**.
img_v2_5221de63-ecfa-469c-aad5-42b9c7f615dg


caozj1011 (approved):
LGTM!

xingtanzjr (approved):

PR #10568, opened 2023-07-17
Fix default labels in prometheus reporter when metrics has no tag

SpriCoder:
Fix default labels in prometheus reporter when metrics has no tag


OneSizeFitsQuorum (approved):
LGTM

PR #10569, opened 2023-07-17
add version and build info in starting configNode

l2280212:
add version and build info in starting configNode


OneSizeFitsQuorum (approved):
LGTM

PR #10577, opened 2023-07-18
Pipe: record root cause when exception is reported

DanielWang2035:


SteveYurongSu:
Plz cp this commit to rel/1.2. Thanks!

PR #10588, opened 2023-07-18
Add READ and WRITE into keyword

SpriCoder:
Add READ and WRITE into keyword


iotdb-core/antlr/src/main/antlr4/org/apache/iotdb/db/qp/sql/IdentifierParser.g4
```diff
@@ -174,6 +174,7 @@ keyWords
     | QUOTA
     | RANGE
     | READONLY
+    | READ
```
lancelly, line 177: READONLY is after READ in lexicographic order(:

PR #10597, opened 2023-07-19
[IOTDB-5557] [RatisConsensus] Support linearizable read during recovery

SzyWilliam:
The default level is set to "query statemachine directly”, which, while latency-friendly, has led to user-reported bugs. Specifically, these bugs relate to the production of inconsistent results in subsequent SQL queries during a restart, creating a phantom read problem that may be confusing for our users.

To address this issue, I propose that we temporarily increase the read consistency level to linearizable read during restarts. This will ensure that we maintain data consistency during the critical recovery period. Once the cluster has successfully finished recovering from previous logs, we can then revert to the default consistency level.


lancelly:
> The default level is set to "query statemachine directly”, which, while latency-friendly, has led to user-reported bugs. Specifically, these bugs relate to the production of inconsistent results in subsequent SQL queries during a restart, creating a phantom read problem that may be confusing for our users.
> 
> To address this issue, I propose that we temporarily increase the read consistency level to linearizable read during restarts. This will ensure that we maintain data consistency during the critical recovery period. Once the cluster has successfully finished recovering from previous logs, we can then revert to the default consistency level.

LGTM~ Song is true master of Ratis indeed🤩

CRZbulabula (approved):
LGTM! Let me introduce William Song, a shining star in both Apache IoTDB and Apache Ratis community.

Cpaulyz:
Excellent work! Thank you for working day and night! This is a great contribution for IoTDB. 🤟💘

BUAAserein:
Awesome work! Thanks William Song @SzyWilliam very much for working on this! 🤟💘

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
```diff
@@ -803,6 +838,11 @@ public RaftServer getServer() {
     return server;
   }
 
+  @TestOnly
+  public void allowStaleRead(ConsensusGroupId consensusGroupId) {
+    canServeStaleRead.computeIfAbsent(consensusGroupId, id -> new AtomicBoolean(true));
```
OneSizeFitsQuorum, line 843: canServeStaleRead.computeIfAbsent(consensusGroupId, id -> new AtomicBoolean(true)).set(true) ?

iotdb-core/consensus/src/test/java/org/apache/iotdb/consensus/ratis/RecoverReadTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+    miniCluster.getStateMachines().stream()
+        .map(SlowRecoverStateMachine.class::cast)
+        .forEach(SlowRecoverStateMachine::setStall);
+
+    // shutdown and restart the cluster
+    miniCluster.restart();
+
+    // query during redo: get exception that ratis is under recovery
+    final ConsensusReadResponse readResponse = TestUtils.doRead(miniCluster.getServer(0), gid);
```
OneSizeFitsQuorum, line 173: The execution time for 10 logs is 5s, and the query timeout is 30s. Why is there an exception?

OneSizeFitsQuorum (commented):
Awesome work！
Although the Ratis community does not currently support lease read, this work has allowed us to avoid a lot of unwanted anomalies.
I was wondering if we could avoid this kind of anomalies when the new leader takes over, so that some of the upper confingode modules don't see the old state. I think RatisConsensus and ApplicationStateMachineProxy of Boolean variables can be shared to meet this requirement

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
SzyWilliam: Good catch!

iotdb-core/consensus/src/test/java/org/apache/iotdb/consensus/ratis/RecoverReadTest.java
SzyWilliam: It is caused by no-leader. No active leader will result in immediate return of failure.

SzyWilliam:
@lancelly @Cpaulyz @CRZbulabula @BUAAserein Thanks a lot for reviewing at this PR! You are the true heroes of IoTDB Community!

SzyWilliam:
@OneSizeFitsQuorum Thanks for the careful review! Made corresponding changes on code. PTAL.
How about keep this PR to 'restart' and propose 'leader-switch' related changes in a new PR?

OneSizeFitsQuorum:
> @OneSizeFitsQuorum Thanks for the careful review! Made corresponding changes on code. PTAL. How about keep this PR to 'restart' and propose 'leader-switch' related changes in a new PR?

Good Idea！After fixing suspected errors in CI, I will merge this PR. Looking forward to your next PR!

OneSizeFitsQuorum (approved):
LGTM

PR #10601, opened 2023-07-19
Fix a bug of cross space selection

shuwenwei:
## Description
If an unsequence file is generated by a load operation or some files in sequence space is deleted, some devices in unsequence file may not overlap with any files in sequence space, these devices' startTime may larger than the latest endTime of it in sequence space. When performing a cross-space compaction, the selected sequence file may not contain these devices, causing these devices to be allocated to the first target file when performing the compaction. 
This PR ensures that if there have some sequence files before the time range of the unsequence device , the last sequence file will also be selected unless it is unclosed.
截屏2023-07-19 18 15 04


caozj1011 (approved):
LGTM!

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/selector/utils/CrossSpaceCompactionCandidate.java
```diff
@@ -113,15 +116,39 @@ private boolean prepareNextSplit() throws IOException {
...
+      // not contain these devices, causing these devices to be allocated to the
+      // first target file when performing the compaction.
+      // Here ensures that if there have some sequence files before the time range
+      // of the unsequence device , the last sequence file will also be selected unless
+      // it is unclosed.
+      boolean seqSpaceHasCurrentUnseqDeviceButNotOverlap =
+          (!atLeastOneSelectedSeqFileContainsCurrentDevice
+              && lastSeqFileWhoseDeviceTimeBeforeCurrentUnseqDevice != null);
+      if (!seqSpaceHasCurrentUnseqDeviceButNotOverlap
+          || lastSeqFileWhoseDeviceTimeBeforeCurrentUnseqDevice.unsealed()
```
xingtanzjr, line 146: which scenario does this line `lastSeqFileWhoseDeviceTimeBeforeCurrentUnseqDevice.unsealed()` avoid ?

shuwenwei: If there is a unclosed sequence file which contains this device and it is not overlapped with the unsequence file, it will become this lastSeqFileWhoseDeviceTimeBeforeCurrentUnseqDevice. However, a unclosed file should not be selected in any case.

xingtanzjr (approved):

PR #10604, opened 2023-07-19
[To rel/1.1] Change the threshold of candidate size of seq tsfile in cross space

xingtanzjr:
## Description
The threshold of candidate size of seq tsfile in cross space compaction is changed to `target_compaction_file_size * 1.5` in this PR.
Before this change, some cross compaction task won't be executed because of this limitation, which lead to accumulation of  lots of unseq files.


Beyyes (approved):
LGTM

PR #10620, opened 2023-07-20
[IOTDB-6081] Pipe: use HybridExtractor instead of LogExtractor when realtime mode is set to log to avoid OOM under heavy insertion load

DanielWang2035:
Problem: https://jira.infra.timecho.com:8443/browse/TIMECHODB-229
Fix: 
- When user specifies 'extractor.realtime.mode'='log', we use hybrid extractors which can automatically degrade to file mode when load is too heavy.
- Add a parameter: 'extractor.realtime.mode'='forced-log', the pipe will always use log mode and will never degrade to file mode.


SteveYurongSu (approved):

SteveYurongSu:
Please cheery-pick it to rel/1.2

PR #10624, opened 2023-07-20
[IOTDB-6072] Load: workaround to CPU 100% after loading tsfile when using JDK8 runtime

MiniSho:
## Description
After loading tsfile, it causes 100% load on one CPU core.

## Reason
it was a bug in JDK1.8.

[JDK-8129861](https://bugs.openjdk.java.net/browse/JDK-8129861)
[JDK-8022642](https://bugs.openjdk.java.net/browse/JDK-8022642)

## Solution
1. set corePoolSize to 1 to avoid high processor load
2. lazy initialization: init loadTsFileManager  on first call


SteveYurongSu:
LGTM :)

PR #10630, opened 2023-07-20
[To rel/1.1] Remove multi-IP initialization in session pool of client-py

ycycse:
Session.init_from_node_urls() is not supported in rel/1.1, so we should remove it in session pool.

### DIFF
--- a/client-py/iotdb/SessionPool.py
+++ b/client-py/iotdb/SessionPool.py
@@ -27,6 +27,7 @@ DEFAULT_MULTIPIE = 5
 DEFAULT_FETCH_SIZE = 5000
 DEFAULT_MAX_RETRY = 3
 DEFAULT_TIME_ZONE = "Asia/Shanghai"
+HAS_SESSION_POOL = False
 logger = logging.getLogger("IoTDB")
 
 
@@ -53,6 +54,8 @@ class SessionPool(object):
         self.__queue = Queue(max_pool_size)
         self.__lock = Lock()
         self.__closed = False
+        global HAS_SESSION_POOL
+        HAS_SESSION_POOL = True
 
     def __construct_session(self) -> Session:
         return Session(self.__pool_config.host, self.__pool_config.port, self.__pool_config.user_name,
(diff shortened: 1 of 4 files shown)


JackieTien97 (approved):

client-py/iotdb/SessionPool.py
```diff
@@ -55,22 +53,23 @@ def __init__(self, pool_config: PoolConfig, max_pool_size: int, wait_timeout_in_
...
-        else:
-            return Session(self.__pool_config.host, self.__pool_config.ip, self.__pool_config.user_name,
-                           self.__pool_config.password, self.__pool_config.fetch_size, self.__pool_config.time_zone)
+        return session
 
     def __poll_session(self) -> Session | None:
-        if self.__queue.empty():
-            return None
-        return self.__queue.get(block=False)
+        self.__q_lock.acquire()
```
THUMarkLau, line 67: It's better to rewrite it as
```
with self.__q_lock:
    # Do something
```
Which will automatically aquire and release lock.

client-py/iotdb/SessionPool.py
```diff
@@ -114,17 +112,23 @@ def put_back(self, session: Session):
             raise ConnectionError("SessionPool has already been closed, please close the session manually.")
 
         if session.is_open():
+            self.__q_lock.acquire()
```
THUMarkLau, line 115: Same as above

client-py/iotdb/SessionPool.py
```diff
@@ -114,17 +112,23 @@ def put_back(self, session: Session):
...
             self.__queue.put(session)
+            self.__q_lock.release()
         else:
             self.__lock.acquire()
             self.__pool_size -= 1
             self.__lock.release()
 
     def close(self):
+        self.__q_lock.acquire()
```
THUMarkLau, line 124: Same as above

client-py/iotdb/SessionPool.py
```diff
@@ -18,32 +18,30 @@
 import logging
 import multiprocessing
 import time
-from multiprocessing import Queue
-from threading import Lock
+from queue import Queue
```
THUMarkLau, line 21: `queue.Queue` is used for multi-threading, and `multiprocessing.Queue` is used for multi-processing. `threading.Lock` is used for multi-threading, and `multiprocessing.Lock` is used for multi-processing.  I think we should use `queue.Queue` and `threading.Lock` if the session-pool is designed for the usage in multi-threading environment, else we should use `multiprocessing.Queue` and `multiprocessing.Lock` if it's designed for the usage of multi process environment.

THUMarkLau (approved):

PR #10651, opened 2023-07-22
fix inaccurate statistics for monitoring the number of unseq Tsfile on cross compaction

caozj1011:
When an exception occurs in the execution of a cross-space task, when some source files exist and all target files have been generated, the system will decide to delete the remaining source files. When the source files are deleted, the number of files in the monitoring system is not updated, and it should be updated

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/utils/CompactionUtils.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/utils/CompactionUtils.java
@@ -466,7 +466,7 @@ public class CompactionUtils {
       sequenceFileNames.add(tsFileResource.getTsFile().getName());
       tsFileResource.remove();
       logger.info(
-          "[Compaction] delete unSequence file :{}", tsFileResource.getTsFile().getAbsolutePath());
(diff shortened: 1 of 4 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/utils/CompactionUtils.java
```diff
@@ -436,4 +436,38 @@ public static boolean validateSingleTsFiles(TsFileResource resource) {
...
+    deleteSourceTsFileAndUpdateFileMetrics(sourceSeqResourceList);
+
+    // delete unSeq file
+    long[] unSequenceFileSize = new long[sourceUnseqResourceList.size()];
+    List<String> unSequenceFileNames = new ArrayList<>();
+    for (int i = 0; i < sourceUnseqResourceList.size(); i++) {
+      TsFileResource tsFileResource = sourceUnseqResourceList.get(i);
+      unSequenceFileSize[i] = tsFileResource.getTsFileSize();
+      unSequenceFileNames.add(tsFileResource.getTsFile().getName());
+      tsFileResource.remove();
```
xingtanzjr, line 452: We need to handle whether the remove operation is successful or not

xingtanzjr (approved):

PR #10665, opened 2023-07-24
Fixes #10676: Refactoring and warning messages improvement for ConfigNodeClient

### LINKED ISSUE #10676: ConfigNodeClient’s exception log needs improvement
Currently, whenever a DataNode sends a request to a ConfigNode and an TException is caught due to any reason, an "Failed to connect to ConfigNode" exception is logged in the system. 
However, in reality, the exception may not necessarily be a connection issue. 
This could be misleading and cause confusion during issue investigation, as developers might falsely assume that the issue is related to connecting to the ConfigNode, when it may not be the case.

liyuheng55555:
resolve this issue: https://github.com/apache/iotdb/issues/10676
doc: https://apache-iotdb.feishu.cn/docx/Gd1SdLZaIoYBsSxJ1MTc6AN5niZ

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/client/ConfigNodeClient.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/client/ConfigNodeClient.java
@@ -721,291 +721,100 @@ public class ConfigNodeClient implements IConfigNodeRPCService.Iface, ThriftClie
 
   @Override
   public TSStatus createTrigger(TCreateTriggerReq req) throws TException {
-    for (int i = 0; i < 5; i++) {
-      try {
-        TSStatus status = client.createTrigger(req);
-        if (!updateConfigNodeLeader(status)) {
-          return status;
-        }
-      } catch (TException e) {
-        logger.warn(
-            MSG_RECONNECTION_DATANODE_FAIL,
-            configNode,
-            config.getAddressAndPort(),
-            Thread.currentThread().getStackTrace()[1].getMethodName());
-        configLeader = null;
-      }
-      waitAndReconnect();
-    }
-    throw new TException(MSG_RECONNECTION_FAIL);
+    return executeRemoteCallWithRetry(
+            () -> client.createTrigger(req), (status) -> !updateConfigNodeLeader(status)
+    );
   }
 
   @Override
   public TSStatus dropTrigger(TDropTriggerReq req) throws TException {
-    for (int i = 0; i < RETRY_NUM; i++) {
-      try {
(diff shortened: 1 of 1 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/client/ConfigNodeClient.java
```diff
@@ -1065,150 +634,50 @@ public TSStatus setDataNodeStatus(TSetDataNodeStatusReq req) throws TException {
...
   public TShowDataNodesResp showDataNodes() throws TException {
-    for (int i = 0; i < RETRY_NUM; i++) {
-      try {
-        TShowDataNodesResp showDataNodesResp = client.showDataNodes();
-        showDataNodesResp.setStatus(showDataNodesResp.getStatus());
-        if (!updateConfigNodeLeader(showDataNodesResp.getStatus())) {
+    return executeRemoteCallWithRetry(
+        () -> {
+          TShowDataNodesResp showDataNodesResp = client.showDataNodes();
+          showDataNodesResp.setStatus(showDataNodesResp.getStatus());
```
OneSizeFitsQuorum, line 658: seems redundant

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/client/ConfigNodeClient.java
```diff
@@ -315,27 +315,48 @@ private boolean updateConfigNodeLeader(TSStatus status) {
...
       try {
-        TSystemConfigurationResp resp = client.getSystemConfiguration();
-        if (!updateConfigNodeLeader(resp.status)) {
-          return resp;
+        T result = call.execute();
+        if (!check.apply(result)) {
+          return result;
         }
       } catch (TException e) {
         logger.warn(
```
OneSizeFitsQuorum, line 336: maybe we should override `void warn(String var1, Throwable var2);`

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/client/ConfigNodeClient.java
liyuheng55555: deleted.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/client/ConfigNodeClient.java
liyuheng55555: done

OneSizeFitsQuorum (approved):
LGTM

PR #10677, opened 2023-07-25
[To rel/1.2] Fixes #10676: Refactoring and warning messages improvement for Config…

### LINKED ISSUE #10676: ConfigNodeClient’s exception log needs improvement

liyuheng55555:
cherry-pick of https://github.com/apache/iotdb/pull/10665


OneSizeFitsQuorum (approved):
LGTM

PR #10691, opened 2023-07-26
[To rel/1.2] add memory estimator on inner space compaction

shuwenwei:
## Description
For in-space merge, the memory evaluation process is added. At the beginning of task execution, the system evaluates the memory size of the task to determine whether the evaluation result is smaller than the configured maximum memory size. If the evaluation result is smaller than the configured maximum memory size, the task can be executed normally. Added two memory estimators, the ReadChunkEstimator and the FastCompactionEstimator, which will be used when performer of the merge task is ReadChunk. Otherwise, use FastCompactionEstimator
see https://github.com/apache/iotdb/pull/10688


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/task/InnerSpaceCompactionTask.java
```diff
@@ -456,9 +476,24 @@ public boolean checkValidAndSetMerging() {
...
+      if (e instanceof InterruptedException) {
+        LOGGER.warn("Interrupted when allocating memory for compaction", e);
+        Thread.currentThread().interrupt();
+      } else if (e instanceof CompactionMemoryNotEnoughException) {
+        LOGGER.info("No enough memory for current compaction task {}", this, e);
+      } else if (e instanceof CompactionFileCountExceededException) {
+        LOGGER.info("No enough file num for current compaction task {}", this, e);
+        SystemInfo.getInstance().resetCompactionMemoryCost(memoryCost);
+      }
+      resetCompactionCandidateStatusForAllSourceFiles();
```
xingtanzjr, line 494: This line is unnecessary

PR #10696, opened 2023-07-26
Do not split pipeline for IdentitySinkNode which only has one child

lancelly:


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/OperatorTreeGenerator.java
```diff
@@ -1924,6 +1924,10 @@ public Operator visitIdentitySink(IdentitySinkNode node, LocalExecutionPlanConte
     sinkHandle.setMaxBytesCanReserve(context.getMaxBytesOneHandleCanReserve());
     context.getDriverContext().setSink(sinkHandle);
 
+    if (node.getChildren().size() == 1) {
+      // if IdentitySinkNode only has one child, we do not need to split pipeline for it.
+      context.setDegreeOfParallelism(1);
+    }
```
JackieTien97, line 1930: We just want the IdentityOperator won't take a seperated pipeline itself when there is only one child operator, not forbidding the parallelism of its child.

JackieTien97 (changes requested):

PR #10697, opened 2023-07-26
[IOTDB-6084] Pipe: support node-urls in connector-v1

DanielWang2035:
Currently, connector-v1 only supports ip and port, while connector-v2 only supports node-urls. We should make connector-v1 and v2 support both ip & port and node-urls.


SteveYurongSu:
Thanks for your great contributions ;) 

Some of your code has been merged in PR https://github.com/apache/iotdb/pull/10749. I will close this PR now.

PR #10706, opened 2023-07-27
[IOTDB-6085] NullPointerException in readAsync when Ratis leader is changing

BUAAserein:
see [jira](https://issues.apache.org/jira/browse/IOTDB-6085)


OneSizeFitsQuorum (approved):
LGTM

PR #10709, opened 2023-07-27
build: Cleaned up the poms to match the new directory structure

chrisdutz:
## Description

Even if the directory structure is a lot cleaner now, still the structure of the maven reactor was flat. 
By inserting intermediate parent poms in each of the subdirectories and adjusting the parent relation, now the project should be nicely structured in Maven enabled IDEs.

I also streamlined the profiles as some were duplicates and moved the integration-test module execution into a separate "with-integration-test" profile as it's execution takes unreasonably long and should probably not be run locally.


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/executable/ExecutableManager.java
```diff
@@ -227,6 +228,8 @@ protected void saveToDir(ByteBuffer byteBuffer, String destination) throws IOExc
     try {
       Path path = Paths.get(destination);
       Files.deleteIfExists(path);
+      // Ensure the target directory exists.
+      SystemFileFactory.INSTANCE.makeDirIfNecessary(destination);
```
HTHou, line 232: The destination String is not a Dir path, it is a file path I believe.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/file/SystemFileFactory.java
```diff
@@ -64,18 +64,21 @@ public File getFile(File parent, String child) {
...
+    File directory = getFile(dir);
+    if (directory.isFile()) {
+      directory = directory.getParentFile();
+    }
+    if (directory.exists()) {
       return;
     }
-    FileUtils.forceMkdir(file);
+    FileUtils.forceMkdir(directory);
   }
```
HTHou, line 83: Then the non-existed file path will be made as a directory by this method.

chrisdutz:
Unfortunately I've encountered a number of problems with the CPP builds and will be cleaning up ab bit.

HTHou:
> Unfortunately I've encountered a number of problems with the CPP builds and will be cleaning up ab bit.

yes, I think the latest CMake has a bug on macos. It cannot find the correct homebrew path.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/executable/ExecutableManager.java
chrisdutz: Yeah... I've noticed that and implemented fixes and am currently trying my changes on mac/linux/win each for x86_64 and aarch64... Just doing this in a fork in order to not spam the making list.

chrisdutz:
Cancelling this PR and will be re-submitting one from my Fork (as this one's slightly out of sync)

PR #10741, opened 2023-07-31
Cleaned up in the projects maven configuration

chrisdutz:
## Description

After the refactoring of the project by grouping together projects in subdirectories, the maven build didn't reflect this change. 

This PR introduces intermediate parent poms for each of these new module groups.

The java related profiles had the problem of not activating on Java 20. So I changed the version range to be open-ended, so we don't need to update the version range every time a new Java version is released.

I also renamed some of the profiles:
- In general self-activating profile (such as OS detection, Java-Version , Thrift-Generation) are now all prefixed with a "." to indicate these are not meant to be manually activated or deactivated.
- Profiles adding new modules to the reactor, I renamed to "with-{xyz}"
- I would propose to rename profiles customizing the build with something like "option-{xyz}" (like "option-get-jar-with-dependencies")

When reviewing the pom content of many projects, I noticed a lot of the configuration (especially the cpp stuff) was mainly copied (I guess from PLC4X' build) Many design assumptions I did in PLC4X however don't apply to IoTDB. I cleaned up the maven config for these parts quite dramatically and hereby greatly simplified things.

Also did I notice that the project contains a number of dependency loops, such as:
- client-cpp needs deployment for running tests
- deployment bundles client-cpp

This is the main reason you can't run a maven build with tests if you've just checked it out or run clean on the project.

This PR resolves all of these issues, except the logical cycle of depending on "distribution", as I'd like to address that by introducing a "test-server" module, that runs IoTDB locally and inside the current modules target directory.

There are more dependencies in the projects, which are not modeled in a Maven way and Maven therefore can't know about them. Especially, in the cpp modules, the CMakeLists.txt or the poms quite regularly use paths, that navigate outside of the current module to copy things from other maven modules. 

Also did I have quite a lot of problems running the build on my ARM64 Mac (M2) in any of our supported OSes (Mac, Linux, Windows). I worked hard on getting the build to run on all of these:
- Linux-x86_64
- Linux-aarch64
- Mac-x86_64
- Mac-aarch64
- Windows-x86_64
- Windows-aarch64
(Currently building the cpp-client in windows is still a bit of a Work-in-Progress, Thrift however builds nicely)

The thrift compiler module in "iotdb-client" furned out to also be needed in all protocol modules, where we use the thrift compiler to compile the Thrift specs (Even if we used a download from our second git repo to hide that dependency). I moved that part to a "tools" directory in the root of the project. I would however propose to move this part into a separate repository, as it actually doesn't have even the slightest thing to do with IoTDB. As some modules were not only depending on the executable, but also the shared libraries and the header files, I refactored that module to produce an assembly containing all of this with a classifier for each supported hardware and os platform we support.


chrisdutz:
Wrong branches ...

PR #10742, opened 2023-07-31
Cleaned up in the projects maven configuration

chrisdutz:
After the refactoring of the project by grouping together projects in subdirectories, the maven build didn't reflect this change.

This PR introduces intermediate parent poms for each of these new module groups.

The java related profiles had the problem of not activating on Java 20. So I changed the version range to be open-ended, so we don't need to update the version range every time a new Java version is released.

I also renamed some of the profiles:

In general self-activating profile (such as OS detection, Java-Version , Thrift-Generation) are now all prefixed with a "." to indicate these are not meant to be manually activated or deactivated.
Profiles adding new modules to the reactor, I renamed to "with-{xyz}"
I would propose to rename profiles customizing the build with something like "option-{xyz}" (like "option-get-jar-with-dependencies")
When reviewing the pom content of many projects, I noticed a lot of the configuration (especially the cpp stuff) was mainly copied (I guess from PLC4X' build) Many design assumptions I did in PLC4X however don't apply to IoTDB. I cleaned up the maven config for these parts quite dramatically and hereby greatly simplified things.

Also did I notice that the project contains a number of dependency loops, such as:

client-cpp needs deployment for running tests
deployment bundles client-cpp
This is the main reason you can't run a maven build with tests if you've just checked it out or run clean on the project.

This PR resolves all of these issues, except the logical cycle of depending on "distribution", as I'd like to address that by introducing a "test-server" module, that runs IoTDB locally and inside the current modules target directory.

There are more dependencies in the projects, which are not modeled in a Maven way and Maven therefore can't know about them. Especially, in the cpp modules, the CMakeLists.txt or the poms quite regularly use paths, that navigate outside of the current module to copy things from other maven modules.

Also did I have quite a lot of problems running the build on my ARM64 Mac (M2) in any of our supported OSes (Mac, Linux, Windows). I worked hard on getting the build to run on all of these:

Linux-x86_64
Linux-aarch64
Mac-x86_64
Mac-aarch64
Windows-x86_64
Windows-aarch64
(Currently building the cpp-client in windows is still a bit of a Work-in-Progress, Thrift however builds nicely)
The thrift compiler module in "iotdb-client" furned out to also be needed in all protocol modules, where we use the thrift compiler to compile the Thrift specs (Even if we used a download from our second git repo to hide that dependency). I moved that part to a "tools" directory in the root of the project. I would however propose to move this part into a separate repository, as it actually doesn't have even the slightest thing to do with IoTDB. As some modules were not only depending on the executable, but also the shared libraries and the header files, I refactored that module to produce an assembly containing all of this with a classifier for each supported hardware and os platform we support.


chrisdutz:
Ok ... so finally I think I've ironed out all quirks ... however no Idea why the integration tests only seldomly seem to pass ... haven't quite detected a pattern yet.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/executable/ExecutableManager.java
```diff
@@ -227,6 +228,8 @@ protected void saveToDir(ByteBuffer byteBuffer, String destination) throws IOExc
     try {
       Path path = Paths.get(destination);
       Files.deleteIfExists(path);
+      // Ensure the target directory exists.
+      SystemFileFactory.INSTANCE.makeDirIfNecessary(destination);
```
HTHou, line 232: > Ok ... so finally I think I've ironed out all quirks ... however no Idea why the integration tests only seldomly seem to pass ... haven't quite detected a pattern yet.

The failed integration tests are related to Trigger. And the reason is what I mentioned in your previous PR #10709. The Line 232 is unnecessary.

chrisdutz: Thanks for re-reporting that ... now that you mention it, you did report that on the old PR, that I dropped, because I wanted to keep the noise on the list to a mimimum, but in the end couldn't completely as my fork didn't run the Workflows :-( ... I wasn't ignoring you intentionally ;-)

chrisdutz:
@HTHou thanks for re-reporting that problem ... finally the integration-tests are green again (sorry for having you double-report that).
Now I hope I'll get that python test green again.

chrisdutz:
YAY ... only had to re-run it 3 times ;-)

HTHou:
@chrisdutz I am trying to fix the python problem, and fortunately the reason is found out. :)
See #10760.

chrisdutz:
Just updated the PR with the latest updates from upstream

chrisdutz:
So I guess I'll consider using lazy consensus on this pr... Or should I wait for some formal approval?

Gesendet von Outlook für Android<https://aka.ms/AAb9ysg>
________________________________

pom.xml
```diff
@@ -1109,88 +1115,116 @@
...
             <properties>
-                <os.classifier>windows-x86_64</os.classifier>
-                <thrift.download-url>http://archive.apache.org/dist/thrift/${thrift.version}/thrift-${thrift.version}.exe</thrift.download-url>
-                <thrift.executable>thrift-${thrift.version}-win-x86_64.exe</thrift.executable>
-                <thrift.skip-making-executable>true</thrift.skip-making-executable>
-                <thrift.exec-cmd.executable>echo</thrift.exec-cmd.executable>
-                <thrift.exec-cmd.args>"Do nothing"</thrift.exec-cmd.args>
+                <os.classifier>linux-x86_64</os.classifier>
+                <cmake.generator>Unix Makefiles</cmake.generator>
+                <thrift.executable>thrift</thrift.executable>
```
HTHou, line 1133: Well, I have a question about this part. If we do this modification, anyone who want to build an IoTDB binary package to need to install a lot of extra tools like boost and bison, which is very inconvenient. However currently if we would like to do that, only we need is JDK and maven....

chrisdutz: There are two profiles: 
- with-cpp with will add the cpp-client and cpp-example
- with-tools which will add the compilation of thrift
The thing that I've changed, was that in the past we only had the thrift executable and not the shared libraries available as download. I've changed the thrift build to build an assembly that contains the executable, includes and shared libraries.
If a user wants to compile without bison, boost etc, the build will - instead of downloading from git - will load a maven dependency: 
`                                        <artifactItem>
                                            <artifactId>iotdb-tools-thrift</artifactId>
                                            <version>${project.version}</version>
                                            <classifier>${os.classifier}</classifier>
                                            <type>zip</type>
                                            <overWrite>true</overWrite>
                                            <outputDirectory>${project.build.directory}/thrift</outputDirectory>
                                        </artifactItem>
`
So he doesn't need to compile it on his machine. I think you still don't ned anything more than Java (cause you could even use the maven-wrapper and then not even need Maven). But I could double-check this as I have some VMs that I can reset to a clean OS installations for checks like this.

This worked fine for mac and linux for both x86_64 and aarch64 ... just on Windows did I run into some issues.

HTHou (approved):
No more questions from my side. LGTM!

pom.xml
chrisdutz: Was worried at first as I got an error on my clean system with the architecture not being right, till I noticed I hadn't switched to my branch (So I have confirmed, that on aarch64 (M2 Mac) systems I currently couldn't build without building thrift on my machine first). After switching to my branch, it built nicely:

    [INFO] ------------------------------------------------------------------------
    [INFO] BUILD SUCCESS
    [INFO] ------------------------------------------------------------------------
    [INFO] Total time:  03:25 min
    [INFO] Finished at: 2023-08-03T14:31:15+02:00
    [INFO] ------------------------------------------------------------------------

chrisdutz: And I checked ... bison wasn't installed ... ssh-develop wasn't, boost wasn't ... just Git and Java

PR #10762, opened 2023-08-02
[IOTDB-6094] Load: Fix error in if/else judgment of tsfileResource exists or not & Reduce warn log when timeseries exist

MiniSho:
## Description
In a benchmark scenario with 500 devices and 50 sensors, the load gives a No Such File error. 
image

## Reason
`constructTsFileResource` only update the tsfile resource information and does not create a file, then every time it determines whether the file exists, causes it to go to the wrong branch and add resource to LoadTsfileStatement wrongly and repeatedly. As a result, multiple identical tsfile resources are stored in the `resources` of  `LoadTsFileStatement`.After loading the file once, it will be deleted in tmp dir and NoSuchFileException will be reported in the subsequent load.

constructTsFileResource 只会更新tsfileResource而不会创建一个新文件，导致每次对tsfile resource 逻辑判断是否存在该文件时会走入错误的分支而重复添加该 tsfile resource到 LoadTsfileStatement 中。在第一次load后，存储在临时文件夹里的tsfile会被删掉，而导致后续重复的load找不到这个文件而报错 NoSuchFIleException.


lancelly:
Nobody knows load better than you🙌

yschengzi:
You Make Load Great Again!!!

SteveYurongSu (approved):
MLGA!

PR #10765, opened 2023-08-03
Refactoring DeleteOutdatedFileTask in WalNode

caozj1011:
This commit reconstructs the execution logic of DeleteOutdatedFileTask, abstracting the collation logic into the following steps:
1. Delete obsolete wal files
2. Update the valid information ratio
3. Determine whether snapshot or flush is required based on the valid information ratio
4. Summarize the execution result and export it

In addition, there are two updates:
1. The initial assignment process of the property is moved to the constructor, which makes the code look cleaner.
2. If the retry process after snapshot or flush is deleted, it is difficult to execute statistics containing multiple retry results at a time.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/node/WALNode.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/node/WALNode.java
@@ -63,6 +63,7 @@ import java.io.File;
 import java.io.FileNotFoundException;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.LinkedList;
@@ -75,6 +76,7 @@ import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 /**
  * This class encapsulates {@link IWALBuffer} and {@link CheckpointManager}. If search is enabled,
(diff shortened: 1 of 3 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/node/WALNode.java
```diff
@@ -215,38 +218,64 @@ public void deleteOutdatedFiles() {
   }
 
   private class DeleteOutdatedFileTask implements Runnable {
-    private static final int MAX_RECURSION_TIME = 5;
     // .wal files whose version ids are less than first valid version id should be deleted
     private long firstValidVersionId;
     // the effective information ratio
-    private double effectiveInfoRatio;
-    // recursion time of calling deletion
-    private int recursionTime = 0;
```
HeimingZ, line 224: It's unsuitable to remove the retry procedure, because outdated wal files may be kept by several memtables and the retry procedure can snapshot/flush more outdated memtables in one task.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/node/WALNode.java
```diff
@@ -255,91 +284,108 @@ public void run() {
...
+              .append(System.getProperty("line.separator"));
+        }
+        if (fileIndexAfterFilterSafelyDeleteIndex < filesShouldDelete.length) {
+          summary.append(
+              String.format(
+                  "- MemTable contained in the wal file was flushed, but the data in the file was not consumed by the consensus group.current searchIndex is %d,safelyDeleteIndex is %d",
+                  getCurrentSearchIndex(), safelyDeletedSearchIndex));
+        }
+        String summaryLog = summary.toString();
+        logger.warn(summaryLog);
```
HeimingZ, line 342: Your summary log is very long because of the filesShouldDelete, why use warn log level here?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/node/WALNode.java
```diff
@@ -255,91 +284,108 @@ public void run() {
...
+                    firstValidVersionId,
+                    System.getProperty("line.separator")));
+
+        if (!pinnedMemTableIds.isEmpty()) {
+          summary
+              .append("- MemTable has been flushed but pinned by PIPE,the MemTableId list is : ")
+              .append(StringUtils.join(pinnedMemTableIds, ","))
+              .append(".")
+              .append(System.getProperty("line.separator"));
+        }
```
HeimingZ, line 334: it seems that you will warn users when the PIPE pins some flushed memtables even if these operations aren't dangerous.

HeimingZ (approved):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/node/WALNode.java
```diff
@@ -219,34 +222,75 @@ private class DeleteOutdatedFileTask implements Runnable {
...
     // the effective information ratio
-    private double effectiveInfoRatio;
-    // recursion time of calling deletion
+    private double effectiveInfoRatio = 0d;
+
+    private List<Long> pinnedMemTableIds;
+
+    private File[] filesShouldDelete;
+
+    private int fileIndexAfterFilterSafelyDeleteIndex = Integer.MAX_VALUE;
```
HeimingZ, line 231: set this to Integer.MAX_VALUE in the init() method

caozj1011: This problem can be ignored after confirmation

xingtanzjr (approved):

PR #10838, opened 2023-08-11
[To rel/1.2] Feature add insertRecords method to rest service

CloudWise-Lukemiao:
[To rel/1.2] Feature add insertRecords method to rest service

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/rest/v1/handler/StatementConstructionHandler.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/rest/v1/handler/StatementConstructionHandler.java
@@ -19,14 +19,22 @@ package org.apache.iotdb.db.protocol.rest.v1.handler;
 
 import org.apache.iotdb.commons.exception.IllegalPathException;
 import org.apache.iotdb.db.exception.WriteProcessRejectException;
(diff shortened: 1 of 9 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/rest/v1/handler/StatementConstructionHandler.java
```diff
@@ -164,4 +172,55 @@ public static InsertTabletStatement constructInsertTabletStatement(
...
+
+    for (int i = 0; i < insertRecordsRequest.getDeviceIds().size(); i++) {
+      InsertRowStatement statement = new InsertRowStatement();
+      statement.setDevicePath(
+          DataNodeDevicePathCache.getInstance()
+              .getPartialPath(insertRecordsRequest.getDeviceIds().get(i)));
+      statement.setMeasurements(
+          insertRecordsRequest.getMeasurementsList().get(i).toArray(new String[0]));
+      statement.setTime(insertRecordsRequest.getTimestamps().get(i));
+      statement.fillValues(valuesList.get(i));
```
HTHou, line 214: I think convert the values to byte buffer and transfer back to objects is not a good idea. We can just set values directly to the insertStatement.

HTHou (approved):

PR #10846, opened 2023-08-14
[IOTDB-6112] Fix Limit & Offset push down doesn't take effect while there exist time filter

JackieTien97:
Previously, for aligned time series, we only use statistics to skip offset while the queryFilter is null, however if the query filter is time filter, it can also use that.


liuminghui233 (approved):
LGTM :-)

PR #10854, opened 2023-08-14
[IOTDB-6113] Load: Fix setDatabase does not print logs and set failure except DATABASE_ALREADY_EXISTS

MiniSho:
The previous if/else logic judgment will cause exceptions which excepts DATABASE_ALREADY_EXISTS to be set to success


SteveYurongSu (approved):
Sho makes load great again :))

PR #10858, opened 2023-08-14
[To rel/1.2][IOTDB-6111] Fix Region creation bugs

CRZbulabula:
**Updations:**

1. Both the Running and Unknown DataNodes can be used when making CreateRegionGroupsPlan. Since we have IoTDB-Region-Maintainer thread in PartitionManager to ensure the creation at Unknown DataNodes will finally successed.
2. Distinguish re-creation and re-deletion in IoTDB-Region-Maintainer. So that these corner cases wouldn't stuck into infinite loop.
3. A RegionGroup will always been Removing status as soon as it has been removed.


OneSizeFitsQuorum (approved):
LGTM

PR #10862, opened 2023-08-15
Fix Region migrate bug

CRZbulabula:
The RegionReplicaSet was not updated successfully after migration


OneSizeFitsQuorum (approved):
LGTM

PR #10873, opened 2023-08-16
reduce wal log output level

caozj1011:
info logs are often printed unnecessarily when no outdated WAL files have been deleted. This fix will reduce the log level to debug. In addition, some optimization of the printed log content.

image


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/node/WALNode.java
```diff
@@ -309,8 +309,8 @@ private void updateEffectiveInfoRationAndUpdateMetric() {
 
     private void summarizeExecuteResult() {
       if (filesShouldDelete.length == 0) {
-        logger.info(
-            "wal node-{}:no wal file was found that should be deleted,current first valid version id is {}",
+        logger.debug(
```
SteveYurongSu, line 312: if (logger.isDebugEnabled()) {
  xxx.
}

xingtanzjr (approved):

PR #10875, opened 2023-08-16
[IOTDB-6117] Pipe: Optimize RPC requests from cn to dn

DanielWang2035:
Currently, cn sends pipeMeta of all existing pipes to each dn upon create/start/stop/drop pipe, which may be time-comsuming. In this PR, cn will send exactly one pipeMeta to each dn upon create/start/stop/drop pipe.


SteveYurongSu (approved):
Pretty!!!

PR #10877, opened 2023-08-16
[IOTDB-6115] Fix Limit & Offset push down doesn't take effect while there exist null value

JackieTien97:
If we query all sensors of one device or any queried sensor has the same value count as the time column, we can use the statistics to skip.


Beyyes (approved):
LGTM

PR #10886, opened 2023-08-17
Fix dataMigrationExample wrong concurrency error

l2280212:
The example for data migration may generate warnings, causing the data import to fail. Now, I have made corrections to the example and added comments.
image


OneSizeFitsQuorum (approved):
LGTM

PR #10887, opened 2023-08-17
[IOTDB-6119] Add ConfigNode leader service check

CRZbulabula:
The ConfigNode-leader can't provide services until all of its leader services are ready. Therefore there should add another AtomicBoolean to mark the leader services' status.


CRZbulabula:
> LGTM for now~ But we may need lease read in the future to solve the problem completely

Yes. I've added "TODO" in corresponding code

SzyWilliam:
We can discuss further on lease read @OneSizeFitsQuorum @CRZbulabula. I think its quite important for confignode to do the 'right' decision based on the 'correct' informations.

PR #10892, opened 2023-08-17
Upgrade guava to avoid vulnerability issue

HTHou:
## Description

There are some vulnerability issues about early versions of guava. This PR is upgrading it to latest version.
https://mvnrepository.com/artifact/com.google.guava/guava
image


wangchao316 (approved):
LGTM

PR #10893, opened 2023-08-17
Feature/test externalized iotdb tools thrift

chrisdutz:
This is a test-drive of the newly staged external thrift artifacts.


pom.xml
```diff
@@ -1291,6 +1299,17 @@
                                     <generator>java</generator>
                                 </configuration>
                             </execution>
+                            <execution>
+                                <id>generate-thrift-sources-nodejs</id>
```
HTHou, line 1303: Any reason for adding it?

chrisdutz: Admittedly this branch was never meant for being merged ... I used it to test my externalized thrift binaries and as I updated them to also contain the JS compiler, this was more added as a test for that.

HTHou (approved):

PR #10894, opened 2023-08-17
Override existing jar instead of deleting it when registering UDF

lancelly:
On some OS systems(eg.windows), it is not possible to delete a file when it is open or in use. The Jar file will be used by the acitive UDFClassLoader and can not be deleted On windows, which will lead to the failure of the CreateFunctionTask.


lancelly:
image

JackieTien97 (approved):

PR #10900, opened 2023-08-18
[To rel/1.2] fix wal npe when memTable has flushed.

caozj1011:
wal obtains IMemTable from memTableInfo twice when deleting obsolete files. The IMemTable is not null when it is obtained the first time, but the imemtable may have been flushed when it is obtained the second time. As a result, the obtained value is null, resulting in NPE

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/node/WALNode.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/node/WALNode.java
@@ -381,7 +381,9 @@ public class WALNode implements IWALNode {
         return false;
       }
       IMemTable oldestMemTable = oldestMemTableInfo.getMemTable();
-
+      if (oldestMemTable == null) {
+        return false;
+      }
       // get memTable's virtual database processor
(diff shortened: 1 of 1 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/node/WALNode.java
```diff
@@ -457,8 +455,8 @@ private void snapshotMemTable(DataRegion dataRegion, File tsFile, MemTableInfo m
           "CheckpointManager$DeleteOutdatedFileTask.snapshotOrFlushOldestMemTable");
       try {
         // make sure snapshot is made before memTable flush operation
-        synchronized (memTable) {
-          if (memTable.getFlushStatus() != FlushStatus.WORKING) {
+        synchronized (this) {
```
wangchao316, line 458: sync this class, Does the WAL write performance be affected?

caozj1011: Your question makes sense. I changed it to memTableInfo

PR #10906, opened 2023-08-18
Feature/simplified tsfile examples

chrisdutz:
The examples were more complicated than they needed to be, I simplified them a bit.


HTHou (approved):
LGTM

PR #10907, opened 2023-08-18
Feature/streamlined grafana plugin profile

chrisdutz:
Till now the way the grafana plugin was added to the main project reactor was different to that of the other optional modules. 

With this PR it matches that of the others.


HTHou:
https://github.com/apache/iotdb/blob/e9fe0b1082dff04815e75a84b7778126af55d2eb/iotdb-connector/grafana-plugin/backend-compile.sh#L31-L33

To fix the error, you may need to modify these lines.

chrisdutz:
Good catch ... but a bit odd as previously it was 1.13.0 ... so in that case it couldn't have worked with 1.14.0 in the script.

HTHou:
Should we also move this in the pom.xml file of distribution to a profile like we do for cpp client?

https://github.com/apache/iotdb/blob/e9fe0b1082dff04815e75a84b7778126af55d2eb/distribution/pom.xml#L112-L117

distribution/pom.xml
```diff
@@ -125,24 +117,71 @@
...
             <groupId>org.apache.iotdb</groupId>
             <artifactId>spark-iotdb-connector_2.11</artifactId>
             <version>${project.version}</version>
-            <type>jar</type>
         </dependency>
         <dependency>
             <groupId>org.apache.iotdb</groupId>
             <artifactId>spark-iotdb-connector_2.12</artifactId>
             <version>${project.version}</version>
-            <type>jar</type>
```
HTHou, line 140: Why do you remove the type?

chrisdutz: Usually you don't specify it, if it's "jar" as that is the default.

PR #10908, opened 2023-08-18
[IOTDB-6122] Pipe: Progress Index markers an incorrect value when load a tsfile with tsfileResource

MiniSho:
## Description
When we load a tsfile , if it has a tsFileResource, there are two cases in tsFileResource: 
 1. it doesn't have a progress index, we just give it one from system.
 2. it has a progress index, the index shouldn't be loaded, it should be **recovered** instead of updating.


SteveYurongSu (commented):
MLGA!

SteveYurongSu (approved):

PR #10913, opened 2023-08-19
[IOTDB-6123] Pipe: reduce the de/ser to slove the slow log transfer problem

MiniSho:
## Description
As you can see from the graph, WAL's serialization and deserialization consume a lot of CPU, about 25%, which becomes the focus of our optimization attention.
image

## Solution
We perform the following optimizations for log mode:
* When no pattern is set, i.e., the extractor's default pattern is used, we don't need to serialize and deserialize the WAL, but instead transfer the read binary data directly to the receiver
* Expansion of allocated memory


SzyWilliam:
MPGA!

Cpaulyz:
MPGA!

SteveYurongSu (approved):

SteveYurongSu (approved):
MPGA!

PR #10918, opened 2023-08-21
Fix NPE issue in show cluster when adding datanodes

OneSizeFitsQuorum:
img_v2_141f6e9a-8a43-42d0-8697-3e8e7568b5ag


OneSizeFitsQuorum:
May not be the root cause

PR #10919, opened 2023-08-21
Fix point priority reader array index out of bound exception

shuwenwei:
## Description
Add hasNextTimeValuePair() check before call nextTimeValuePair() when creating a new instance of PointElement.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/utils/executor/fast/SeriesCompactionExecutor.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/utils/executor/fast/SeriesCompactionExecutor.java
@@ -244,8 +244,11 @@ public abstract class SeriesCompactionExecutor {
           || firstPageElement.needForceDecoding) {
         // has overlap or modified pages, then deserialize it
         summary.pageOverlapOrModified += 1;
-        pointPriorityReader.addNewPage(firstPageElement);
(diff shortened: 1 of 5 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/utils/executor/fast/SeriesCompactionExecutor.java
```diff
@@ -276,7 +277,9 @@ private void compactWithNonOverlapPage(PageElement pageElement)
     } else {
       // unsealed page is not large enough or page.endTime > file.endTime, then deserialze it
       summary.pageNoneOverlapButDeserialize += 1;
```
xingtanzjr, line 279: Should this summary be considered inside the following `if` ?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/utils/executor/fast/SeriesCompactionExecutor.java
```diff
@@ -244,8 +244,9 @@ private void compactPages()
           || firstPageElement.needForceDecoding) {
         // has overlap or modified pages, then deserialize it
         summary.pageOverlapOrModified += 1;
```
xingtanzjr, line 246: Should this summary be considered inside the following `if` ?

xingtanzjr (approved):

PR #10938, opened 2023-08-23
[IOTDB-6125] Fix DataPartition allocation bug when insert big batch data

CRZbulabula:
The DataPartitions will be allocated un-balanced if the user insert a big batch data to the cluster firstly. Because the RegionGroup quickly-extension-policy doesn't extend enough RegionGroups for the special case.

In this PR, the RegionGroup extension policy will always ensure the database have enough RegionGroups for allocate new Partitions balancely.

And this PR also add some concise log to record the DataAllotTable


OneSizeFitsQuorum (approved):
LGTM

PR #10991, opened 2023-08-29
[IOTDB-6061] Fix the instability failure caused by initServer in IoTConsensus UT not binding to the corresponding port 

OneSizeFitsQuorum:


BUAAserein (approved):
LGTM

PR #10999, opened 2023-08-30
Lazy load chunk in FastCompaction

shuwenwei:
## Description
When an aligned chunk can be directly flushed into target file writer, read the data of chunk from disk only when it is necessary


xingtanzjr:
Need talk more before merging

shuwenwei:
Another version (Lazy load chunk and page): https://github.com/apache/iotdb/compare/master...shuwenwei:iotdb:lazy_load_chunk_and_page_in_fast_compaction?expand=1

PR #10988, opened 2023-08-29
[IOTDB-6129] ConfigNode restarts without relying on Seed-ConfigNode 

liyuheng55555:
https://apache-iotdb.feishu.cn/docs/doccnSfXXrSbLu3ETmDWLY5tmwu#H8SOcB

PR #11003, opened 2023-08-31
[To rel/1.2][IOTDB-6129] ConfigNode restarts without relying on Seed-ConfigNode

liyuheng55555:
same as https://github.com/apache/iotdb/pull/10988


OneSizeFitsQuorum (approved):
LGTM

PR #11020, opened 2023-09-01
Fix possible NPE while executing show cluster or show cluster details

JackieTien97:
We need check whether the value is null while using ColumnBuilder.appendXXValue function, if it's null we need to use appendNull instead, otherwise we may encounter NPE while serializing this TsBlock.
img_v2_e6e85a2d-67dc-44e2-988c-310b0ce937dg


OneSizeFitsQuorum (commented):
LGTM~
However, we will find out the root cause of this problem further in ConfigNode

PR #11026, opened 2023-09-02
[To rel/1.2] Fix possible NPE while executing show cluster or show cluster details

JackieTien97:
cp from https://github.com/apache/iotdb/pull/11020


OneSizeFitsQuorum (approved):
LGTM~
However, we will find out the root cause of this problem further in ConfigNode~

PR #11034, opened 2023-09-04
Fix mods file contains writable view

Cpaulyz:
## Description

When analysing the delete data statement, the deletion of a writable view should be converted to an operation on the original timeseries.

For example:
```
insert into root.db.d1(time, s1) values(1,1);
flush;
create view root.db.d1.vs1 as select s1 from root.db.d1;
delete from root.db.d1.vs1 where time <= 1;
```

modfication file should not contain `root.db.d1.vs1`


SzyWilliam:
@Cpaulyz You did made a significant change to IoTDB. What a bravo PR!

JackieTien97 (approved):

PR #11045, opened 2023-09-05
[RatisConsensus] retry cache expiration time should be longer than retriable-client wait duration

SzyWilliam:
see https://github.com/apache/ratis/pull/915


OneSizeFitsQuorum (approved):
LGTM!

SzyWilliam:
The good news is that we can now rule out the `apply twice` problem and guarantee exactly-once semantic!

PR #11046, opened 2023-09-05
Release chunk data buffer after copy page data from it in aligned series FastCompaction 

shuwenwei:
## Description
When FastCompaction deserializes a overlapped Chunk into the Page queue, the ByteBuffer.get() method is called from the Chunk's Buffer to copy the data of each Page. After copy, the Page is wrapped into a PageElement and put into the pageQueue, but the timeChunk and valueChunks are passed into a new instance of AlignedChunkReader, which causes the clearChunks method to be called at the end of the method is not effective. The reference of each Chunk is still held by AlignedChunkReader and cannot be released. At this time, there is 2 times the chunk size of data in the memory.
## Fix
Release chunk data buffer after copy page data from it.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/utils/executor/fast/element/PageElement.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/utils/executor/fast/element/PageElement.java
@@ -113,6 +113,9 @@ public class PageElement {
     if (this.iChunkReader instanceof ChunkReader) {
       this.batchData = ((ChunkReader) iChunkReader).readPageData(pageHeader, pageData);
     } else {
+      // In order to save memory, AlignedChunkReader is not used here.
+      // If the variable 'iChunkReader' is null, it means that it is
+      // currently an aligned series.
       this.pointReader = getAlignedPagePointReader();
     }
   }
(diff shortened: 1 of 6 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/utils/executor/fast/element/PageElement.java
```diff
@@ -95,20 +103,84 @@ public PageElement(
...
     this.isLastPage = isLastPage;
     this.needForceDecoding = chunkMetadataElement.needForceDecoding;
   }
 
   public void deserializePage() throws IOException {
-    if (iChunkReader instanceof AlignedChunkReader) {
-      this.pointReader =
-          ((AlignedChunkReader) iChunkReader)
-              .getPagePointReader(pageHeader, valuePageHeaders, pageData, valuePageDatas);
+    if (this.iChunkReader == null) {
```
xingtanzjr, line 113: What does `iChunkReader == null` mean ?

shuwenwei: The variable 'iChunkReader' was not set during construction when the series of this PageElement is aligned.

PR #11078, opened 2023-09-07
Pipe: include creation time in PipeProcessorSubtask#taskID to avoid task scheduling issues after 1000+ pipes' creating and dropping

DanielWang2035:
To reproduce:
- create 1000 devices and write some data
- create and start 1000 pipes, their patterns matching the devices
- drop all pipes
- delete all data on receiver side
- create and start the same pipes again
- execute queries on receiver side, and the data are NOT transferred.


SteveYurongSu (approved):
What a nice catch :)

PR #11098, opened 2023-09-08
Fix/addressed issues with rat not being able to process the source bundle

chrisdutz:
Also made sure the thrift pom is not copied to the lib directory of the assemblies

The main issue of not being able to scan the source-bundle seems to have been related to having used a very old assembly plugin version. Updating resolved the issue.


HTHou (approved):

chrisdutz:
@HTHou do you have an idea, why all ITs are failing?

HTHou:
> @HTHou do you have an idea, why all ITs are failing?

I guess that because of the change of `maven.assembly.version`, but I don't know the reason. Having a try to change it back?

chrisdutz:
Thing is that the reason for me updating that version was that the release produced slightly corrupted archives ... updating to the latest fixed that.

In general we should ensure our plugins are always updated to the most recent versions, as only this way will be get support.

HTHou:
I find a code error in ConfigExecution.java.
https://github.com/apache/iotdb/blob/3c24fa31750623182ab1b94df018a321f1680085/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/execution/config/ConfigExecution.java#L43-L45

L44 should be `com.google.common.util.concurrent.SettableFuture;`

After fixing that, the tests would pass.

iotdb-client/service-rpc/pom.xml
```diff
@@ -109,7 +109,7 @@
         <dependency>
             <groupId>org.apache.iotdb</groupId>
             <artifactId>tsfile</artifactId>
-            <version>${project.version}</version>
+            <version>1.3.0-SNAPSHOT</version>
```
HTHou, line 112: Why do you change `${project.version}` to `1.3.0-SNAPSHOT`?

iotdb-client/service-rpc/pom.xml
```diff
@@ -121,7 +121,7 @@
         <dependency>
             <groupId>org.apache.iotdb</groupId>
             <artifactId>iotdb-thrift</artifactId>
-            <version>${project.version}</version>
+            <version>1.3.0-SNAPSHOT</version>
```
HTHou, line 124: Same point as above.

iotdb-client/service-rpc/pom.xml
chrisdutz: Well actually that's a best practice ... the release plugin replaces all of them so that's no extra work.

iotdb-client/service-rpc/pom.xml
chrisdutz: I just stumbled of the the usage in those two cases, but would lean towards doing that for all.

iotdb-client/service-rpc/pom.xml
HTHou: If I want using `mvn versions:set -DnewVersion=xxx` to change the project version, will the version here also change to the new one?

iotdb-client/service-rpc/pom.xml
chrisdutz: Undid the change ...

HTHou (approved):
LGTM

PR #11134, opened 2023-09-13
[IOTDB-6152] Add IoTConsensus Sync Log Timer

BUAAserein:
## Description

When the IoT follower receives a syncLogEntries request, it will go through three stages: deserialize, sort and apply. This pr adds this three timers.

### DIFF
--- a/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/iot/IoTConsensusServerImpl.java
+++ b/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/iot/IoTConsensusServerImpl.java
@@ -904,8 +904,7 @@ public class IoTConsensusServerImpl {
           }
         }
         long sortTime = System.nanoTime();
-        ioTConsensusServerMetrics.recordSortCost(
-            (sortTime - insertStartTime) / request.getInsertNodes().size());
+        ioTConsensusServerMetrics.recordSortCost(sortTime - insertStartTime);
         logger.debug(
             "source = {}, region = {}, queue size {}, startSyncIndex = {}, endSyncIndex = {}",
             sourcePeerId,
@@ -917,8 +916,7 @@ public class IoTConsensusServerImpl {
         for (IConsensusRequest insertNode : request.getInsertNodes()) {
           subStatus.add(stateMachine.write(insertNode));
         }
-        ioTConsensusServerMetrics.recordApplyCost(
-            (System.nanoTime() - sortTime) / request.getInsertNodes().size());
+        ioTConsensusServerMetrics.recordApplyCost(System.nanoTime() - sortTime);
         queueSortCondition.signalAll();
         return new TSStatus().setSubStatus(subStatus);
       } finally {
--- a/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/iot/IoTConsensusServerImpl.java
+++ b/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/iot/IoTConsensusServerImpl.java
@@ -740,6 +740,10 @@ public class IoTConsensusServerImpl {
     return retryService;
   }
 
+  public IoTConsensusServerMetrics getIoTConsensusServerMetrics() {
+    return this.ioTConsensusServerMetrics;
+  }
+
   public boolean isReadOnly() {
     return stateMachine.isReadOnly();
   }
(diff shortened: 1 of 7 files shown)


SpriCoder (approved):
LGTM~ Please update dashboard at the same time

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/iot/IoTConsensusServerImpl.java
```diff
@@ -909,6 +916,7 @@ private TSStatus cacheAndInsertLatestNode(DeserializedBatchIndexedConsensusReque
         for (IConsensusRequest insertNode : request.getInsertNodes()) {
           subStatus.add(stateMachine.write(insertNode));
         }
+        ioTConsensusServerMetrics.recordApplyCost(System.nanoTime() - sortTime);
```
OneSizeFitsQuorum, line 919: Should we count the time for each IConsensusRequest, otherwise it's not the time of applying the state machine for a single request

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/iot/service/IoTConsensusRPCServiceProcessor.java
```diff
@@ -115,8 +115,11 @@ public void syncLogEntries(
                             : ByteBufferConsensusRequest::new)
                     .collect(Collectors.toList())));
       }
+      long buildRequestTime = System.nanoTime();
       IConsensusRequest deserializedRequest =
           impl.getStateMachine().deserializeRequest(logEntriesInThisBatch);
+      impl.getIoTConsensusServerMetrics()
```
OneSizeFitsQuorum, line 121: Do we need to divide by the number of logs?

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/iot/IoTConsensusServerMetrics.java
```diff
@@ -36,23 +37,44 @@ public class IoTConsensusServerMetrics implements IMetricSet {
...
+
+  public void recordApplyCost(long costTimeInNanos) {
+    applyTimer.updateNanos(costTimeInNanos);
+  }
+
   @Override
   public void bindTo(AbstractMetricService metricService) {
     bindAutoGauge(metricService);
     bindStageHistogram(metricService);
+    bindSyncLogTimer(metricService);
```
OneSizeFitsQuorum, line 70: I was wondering if we could unify this into Histogram or timer, otherwise it seems like there would be some difference in the way they are currently computed? May cause the time does not correspond exactly, right? @SpriCoder What's your opinion?

OneSizeFitsQuorum (commented):
Awesome work!
PTAL

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/iot/IoTConsensusServerImpl.java
BUAAserein: Yes. Now it's the time of one Batch. We should divide it to calculate for every single request.

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/iot/service/IoTConsensusRPCServiceProcessor.java
BUAAserein: Yes.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/service/metric/enums/Metric.java
```diff
@@ -45,6 +45,7 @@ public enum Metric {
   // consensus related
   STAGE("stage"),
   IOT_CONSENSUS("iot_consensus"),
+  IOT_SYNC_LOG("iot_sync_log"),
```
OneSizeFitsQuorum, line 48: iot_receive_log?

BUAAserein: ok

OneSizeFitsQuorum (approved):
LGTM

SpriCoder (approved):
LGTM~

PR #11142, opened 2023-09-14
[IOTDB-6154] Pipe: better algorithm for hybrid mode to switch log/tsfile extraction in iotdb-extractor

MiniSho:
In this PR, I've made two improvements：
1. HybridExtractor will first try to do extraction in log mode, and then choose log or tsfile mode to continue extracting, but if (leader data regions num * Wal size) > (maximum size of wal buffer) will cause OOM, so we need to use tsfile mode if the size exceeds the threshold.
2. PipeRealtimeDataRegionHybridExtractor don't need to use log mode until the historical extractor has consumed all the hitorical data.


CRZbulabula (approved):
I can't describe how elegant these codes are!

Cpaulyz (approved):
MPGA！

SteveYurongSu (approved):
MPGA!

PR #11150, opened 2023-09-15
Refactor/managing dependencies

chrisdutz:
Enabled the enforcer plugin as well as the dependency plugin to make sure every used dependency is declared and we're not using transitive dependencies directly. 

Started replacing javax packages by their jakarta counterparts

Strarted streamlining the dependencies, so we're no longer pulling in 5 different versions of the same dependency.


chrisdutz:
So I guess I'm done .... please ensure this PR can be merged ASAP, as keeping it up to date costs a lot of time.

distribution/pom.xml
```diff
@@ -27,7 +27,43 @@
         <version>1.3.0-SNAPSHOT</version>
     </parent>
     <artifactId>iotdb-distribution</artifactId>
+    <packaging>pom</packaging>
     <name>IoTDB: Distribution</name>
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.iotdb</groupId>
+            <artifactId>iotdb-server</artifactId>
+            <version>1.3.0-SNAPSHOT</version>
```
HTHou, line 36: <version>${project.version}</version> ?

chrisdutz: Sorry .. shouldn't have replied on my phone ... especially if it resolved the discussion and threw away my text ... so here goes again:

Technically there's no difference between using the one or the other. I actually find it easier to type the real version when adding a new dependency than requiring 3 special characters. But beyond that when switching between versions and branches, I usually find it beneficial to use the real versions as then you don't need to scroll to the top to see which version we're in.

However if it's project consensus to use ${project.version}, I'll change it to that. If it hasn't been discussed, then I'd like to start a discussion to reach consensus on it.

chrisdutz: And did you stop reviewing after this, or can I interpret it as that this is the only objection you have?

qiaojialin: If the version could be changed in one mvn command, using real version is better.

qiaojialin:
Hi, after this pr, is there anything that contributors need to pay attention or execute when commit code?

qiaojialin (commented):
Hi, after this pr, is there anything that contributors need to pay attention or execute when commit code?

distribution/pom.xml
chrisdutz: The maven-release-plugin as well as them maven-versions-plugin perfectly handle this case ... there are issues with some non-standard plugins, that if the version is used in the configuration of there, that they are not found, so I strongly encourage using ${project.version} in plugin configurations, however for dependencies I have never seen these plugins miss anything.

chrisdutz:
I have put a lot more strict rules in place ... these will prevent us from having diverging versions again and will inform us if there are CVEs reported for dependencies we use. 

In general the rule is: don't define the versions of plugins and third-party dependencies locally, if you add a new third party dependency or plugin, add that to the pluginManagement or dependencyManagement of the main pom and then use notation without version in sub-modules.

Another rule is, that if you use something in your code, you need to add a dependency to that artifact in the correct scope, otherwise the build will fail. The reasoning behind this is, that if you rely on transitive dependencies, these possibly change without us noticing causing hard to debug problems.

chrisdutz:
On the first look it does seem as if this process adds a bit more work for the developer, which it does. However long-term it pays off 10 time as managing dependency problems becomes really easy and it has saved us many days of tracing down tricky problems. 

For example in the current "Master" branch we use different versions of ratis ... unfortunately versions that are API incompatible. So depending on which modules get added to the distribution in which order will define which module works, which I don't know if we have encountered this problem before, however I am sure, if we had run into it ... that would have taken us quite some time to track down.

distribution/pom.xml
chrisdutz: And when adding new inner-project dependencies, intelliJ auto-fills it with the current version ... which you manually have to replace with the variable (I should probably add to the list of benefits ;-) )

HTHou: > And did you stop reviewing after this, or can I interpret it as that this is the only objection you have?

Yeah, it is the only question I have.

qiaojialin:
So how does these rules work? Will they take effect when we do mvn package?

chrisdutz:
The rules are part of the enforcer plugin executions that are bound to the "validation" phase, which comes after "package". So a "package" build will not be affected ... I also actually didn't even add the rules, they were already in there I simply switched them to fail the build if the rules are not met. I was seeing the warnings while building all the time ... so now there should no longer be any warnings from them because I addressed what they were complaining about.

JackieTien97 (approved):

qiaojialin (approved):

SpriCoder:
This pr cause the failure of manager loading in metric module. So it cause the unavailable of metric module. Let me take a look

PR #11153, opened 2023-09-15
[IOTDB-6151] Move datanode system.properties

liyuheng55555:
1. For the initial startup, the system.properties file is generated under data/system instead of data/system/schema.
2. During startup, check if there is a system.properties file under data/system/schema. If it exists, move it to data/system.


OneSizeFitsQuorum (approved):
LGTM

PR #11184, opened 2023-09-20
[mlnode] Logging module, shut down function and the fixing of a bug about multi process training

ZDandsomSP:
### Description
Logging module, shut down function and the fixing of a bug about multi process training
We have developed new functions for the mlnode module of IoTDB, which mainly includes the following three parts:

#### Logging module
Developed a logging module for mlnode, supporting standard function calls and localized log information for multi-process security

#### Shut down function
Completed the addition of the mlnode shutdown function, and implemented the direct shutdown of the mlnode main process based on the port number on the Linux system

#### Bug fixing
Fixed the issue involved in related work, which is that the original process pool does not support creating nums_ Issues with worker subprocesses

#### Logging module
We provide 5 levels of functions including debug, info, warn, error, and the module can localize logs according to the log level. In subsequent development, we only need to use a form similar to calling the Python standard logging module to use it.

#### Stop function
Since our start function is designed in the form of pip package, our stop function is developed in a similar form. If user start the mlnode in a terminal, we can stop the node in another one with the command "mlnode stop".

### DIFF
--- a/iotdb-core/mlnode/iotdb/mlnode/log.py
+++ b/iotdb-core/mlnode/iotdb/mlnode/log.py
@@ -41,9 +41,8 @@ class LoggerFilter(logging.Filter):
         pid = os.getpid()
         process_name = multiprocessing.current_process().name
 
-        # 格式化函数调用栈信息
         stack_info = ""
-        frame_info = stack_trace[7]  # 忽略前两个栈帧
+        frame_info = stack_trace[7]
         stack_info += f"File: {frame_info.filename}, Line: {frame_info.lineno}, Function: {frame_info.function}"
 
         return f"PID: {pid} PName: {process_name} Stack Trace:{stack_info}"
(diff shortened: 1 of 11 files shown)


iotdb-core/mlnode/iotdb/mlnode/log.py
```diff
@@ -17,18 +17,106 @@
...
+    def custom_log_info():
+        frame = inspect.currentframe()
+        stack_trace = inspect.getouterframes(frame)
+
+        pid = os.getpid()
+        process_name = multiprocessing.current_process().name
+
+        # 格式化函数调用栈信息
+        stack_info = ""
+        frame_info = stack_trace[7]  # 忽略前两个栈帧
```
HTHou, line 46: Please change the comments to English.

ZDandsomSP: We are already aware of these issues, and these fixes are already in our plan. We will correct these obvious deficiencies in subsequent commits

PR #11186, opened 2023-09-21
[IOTDB-6165] Pipe: cache device metadata in TsFile reader to avoid redundant IO

DanielWang2035:
device metadata is deserialized from the file twice in the constructor of TsFileDataContainer, causing an IO overhead.
before:
image
after:
image


SteveYurongSu (approved):
One line diff, Several times inprovement! How magical!

PR #11191, opened 2023-09-21
Pipe: use file mode in forced-log extractor when pipe is stopped to avoid OOM

DanielWang2035:
When pipe is stopped and data is inserted continuously, forced-log extractor will keep queueing tablet insertion events, which may lead to OOM. So we may switch to file mode when pipe is stopped and the queue is reaching capacity.


SteveYurongSu:
We will use hard-link based wal pin strategy to avoid issues caused by wal not deleted. Thanks for working for this.

PR #11196, opened 2023-09-22
Sl fix auth bugs

shuolinl:


wangchao316:
hi, It is recommended to hang the jira order number or ISSUES number.

shuolinl:
> hi, It is recommended to hang the jira order number or ISSUES number.

Good idea.  Before merging this pr, I will squash this two commit into one, and add jira number.

shuolinl:
In this pr:
1. avoid NPE when create view like: create view xx as select xx from xx.
2. user cannot create role with name "root" anymore.
3. when DBA try to delete user/role that not exists, he will get an error.
4. Allow user create user/role with "!@#$%&^*()_+-=", and refine error message.
5. refine TSStatus code used by auth.

JackieTien97 (approved):

PR #11197, opened 2023-09-22
Optimize merge selection process and increase disk space availability logic

caozj1011:
## Description
This PR mainly has the following upgrades:
1. Added the selection criteria for merging tasks in the space. When the availability of disk space is insufficient, TsFile containing mods file is preferentially selected.
2. Compaction adds the CompactionTaskType attribute to a task and compacts the task to `NORMAL` and `MOD_SETTLE`. `NORMAL` is the default value. When the disk size of the mod file exceeds 50M or the disk space availability is insufficient, the mods file sorting task is preferred. In this case, the task type is `MOD_SETTLE`. When a task is executed, the system determines the disk space availability. If the task type is `NORMAL` and the disk space availability is insufficient, the task is discarded from the priority queue to avoid unexpected errors.
3. Optimized the return value of the `selectInnerSpaceTask()` method. The previous return value was not convenient to wrap the CompactionTask object, so we made the logic of this method more pure, the purpose is to return a list of tasks that can be executed, which can be put directly into the task queue or directly executed.
4. Remove the priority queue in the in-space selection step. The `CompactionTaskManager` priority queue already contains all the priority decisions, so the priority queue in the selection stage is meaningless.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/task/CrossSpaceCompactionTask.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/task/CrossSpaceCompactionTask.java
@@ -366,7 +366,7 @@ public class CrossSpaceCompactionTask extends AbstractCompactionTask {
       resetCompactionCandidateStatusForAllSourceFiles();
       return false;
     }
-    if (compactionTaskType == CompactionTaskType.NORMAL && CompactionUtils.isDiskHaveSpace()) {
+    if (compactionTaskType == CompactionTaskType.NORMAL && !CompactionUtils.isDiskHaveSpace()) {
       LOGGER.debug(
           "cross task start check failed because disk free ratio is less than disk_space_warning_threshold");
       return false;
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/task/CrossSpaceCompactionTask.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/task/CrossSpaceCompactionTask.java
@@ -366,6 +366,11 @@ public class CrossSpaceCompactionTask extends AbstractCompactionTask {
       resetCompactionCandidateStatusForAllSourceFiles();
       return false;
     }
+    if (compactionTaskType == CompactionTaskType.NORMAL && CompactionUtils.isDiskHaveSpace()) {
+      LOGGER.debug(
+          "cross task start check failed because disk free ratio is less than disk_space_warning_threshold");
+      return false;
+    }
     try {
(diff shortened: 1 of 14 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/task/CrossSpaceCompactionTask.java
```diff
@@ -366,6 +366,11 @@ public boolean checkValidAndSetMerging() {
       resetCompactionCandidateStatusForAllSourceFiles();
       return false;
     }
+    if (compactionTaskType == CompactionTaskType.NORMAL && !CompactionUtils.isDiskHaveSpace()) {
+      LOGGER.debug(
```
xingtanzjr, line 370: using INFO is OK

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/task/CrossSpaceCompactionTask.java
```diff
@@ -366,6 +366,11 @@ public boolean checkValidAndSetMerging() {
       resetCompactionCandidateStatusForAllSourceFiles();
       return false;
     }
+    if (compactionTaskType == CompactionTaskType.NORMAL && !CompactionUtils.isDiskHaveSpace()) {
```
xingtanzjr, line 369: Maybe we can make this judgement as a method of parent class to get rid of duplicated codes in InnerSpaceCompaction

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/utils/CompactionUtils.java
```diff
@@ -466,4 +472,29 @@ public static void deleteSourceTsFileAndUpdateFileMetrics(
       FileMetrics.getInstance().deleteFile(fileSizes, seq, fileNames);
     }
   }
+
+  public static boolean isDiskHaveSpace() {
```
xingtanzjr, line 476: Maybe rename it to `isDiskHasSpace` or using another description ? 😄

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/task/CrossSpaceCompactionTask.java
caozj1011: good idea

caozj1011: good idea

shuwenwei (approved):

shuwenwei (approved):

PR #11204, opened 2023-09-24
Ensure concurrency safety when updating last cache

Cpaulyz:
## Description

Ensure concurrency safety when updating last cache


JackieTien97 (changes requested):
add getAndInit method

JackieTien97 (approved):

PR #11220, opened 2023-09-26
upgrade compression ratio file name format and solve the problem of high compression ratio caused by repeated data writing

caozj1011:
## Description
This PR has three optimizations:
1. The naming format of the compression Ratio file is changed from `Ratio-compressRatio-flushCount`to `ratio-totalMemoryDataSize-totalDiskDataSize`
2. The rule for calculating the compression ratio from the `sum(memTableSize/flushSize) / count(flushTimes)` to ` totalMemoryDataSize / totalDiskDataSize `
3. Add the `memTableSize` deduplication logic. If no deduplication is performed, the compression ratio calculated is very large, resulting in inaccurate estimation of the combined memory. Therefore, in flush encode, if data is duplicated, the size of the duplicate is removed from `memTableSize`.

## Snapshot
image


shuwenwei:
Does AlignedWritableMemChunk also need to be modified?

PR #11226, opened 2023-09-27
SchemaCache supports precise eviction

Cpaulyz:
## Description

SchemaCache supports precise eviction


Cpaulyz:
There are some experiment results for this optimization. The workload is designed as below:

```java
        session.executeNonQueryStatement("select ** from root.db");
        long timeCost = 0;
        for (int i = 0; i < 1000; i++) {
            session.deleteData(Collections.singletonList("root.db.d1.s1"), 0, Long.MAX_VALUE);
            session.executeQueryStatement("select s0,s1,s2,s3,s4,s5,s6,s7,s8,s9 from root.db.d2");
            timeCost += System.currentTimeMillis() - startTime;
        }
        System.out.println(timeCost);
```

Originally all metadata caches were cleared when deleting data. Therefore, it needs fetchSchema and put into cache for every query statement. 
In the new implementation, deleting data only evict the exact cache enrty. In this example, only root.db.d1.s1 will be evicted, so queries about root.db.d2.* will hit the cache every time.

profile before optimization:
image

profile after optimization:
image

In this example, optimised 10-20% of the time cost.

PR #11228, opened 2023-09-27
Add database and region tag into tsfile metrics

SpriCoder:
Add database and region tag into tsfile metrics


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/metrics/FileMetrics.java
```diff
@@ -333,149 +331,236 @@ private long getOpenFileHandlersNumber() {
...
   // following are update functions for tsfile metrics
-  public void addFile(long size, boolean seq, String name) {
-    updateGlobalCountAndSize(size, 1, seq);
+  public void addFile(String database, String regionId, long size, boolean seq, String name) {
+    updateGlobalCountAndSize(database, regionId, size, 1, seq);
     try {
       TsFileNameGenerator.TsFileName tsFileName = TsFileNameGenerator.getTsFileName(name);
       int level = tsFileName.getInnerCompactionCnt();
-      updateLevelCountAndSize(size, 1, seq, level);
+      updateLevelCountAndSize(size, 1, seq, level, database, regionId);
```
OneSizeFitsQuorum, line 339: Put the database and regionID arguments first, just like all the other functions such as `updateGlobalCountAndSize`

OneSizeFitsQuorum (commented):
Awesome work！
PTAL

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/metrics/FileMetrics.java
SpriCoder: Get it~

THUMarkLau (approved):
Apart from the suggestion made by Master Tan, everything else seems to be fine

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/metrics/FileMetrics.java
```diff
@@ -344,138 +336,161 @@ public void addFile(long size, boolean seq, String name) {
...
+  public <T> void deleteRegionFromMap(
+      Map<String, Map<String, T>> map, String database, String regionId) {
+    Map<String, T> innerMap =
+        map.computeIfPresent(
+            database,
+            (k, v) -> {
+              v.remove(regionId);
+              return v;
+            });
+    if (innerMap == null) {
```
OneSizeFitsQuorum, line 464: When does it become null? It looks like this should be isEmpty?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/metrics/FileMetrics.java
```diff
@@ -344,138 +336,161 @@ public void addFile(long size, boolean seq, String name) {
...
+      map.remove(database);
```
OneSizeFitsQuorum, line 465: Also, atomicity is not guaranteed here, and there may be some concurrency issues. Feel like you need to make a judgment in the function expression and return null if isEmpty?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/metrics/FileMetrics.java
```diff
@@ -114,49 +135,20 @@ public void bindTo(AbstractMetricService metricService) {
...
-        Tag.NAME.toString(),
-        "seq");
-    metricService.createAutoGauge(
-        Metric.FILE_SIZE.toString(),
-        MetricLevel.CORE,
-        this,
-        o -> o.getFileSize(false),
-        Tag.NAME.toString(),
-        "unseq");
     metricService.createAutoGauge(
```
OneSizeFitsQuorum, line 138: We can add that mods files tend to be small, so the benefit of expanding to region metric is too small, so we'll leave it at that

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/metrics/FileMetrics.java
```diff
@@ -344,138 +340,240 @@ public void addFile(long size, boolean seq, String name) {
...
+      updateLevelGauge(
+          entry.getKey(), entry.getValue(), unseqLevelSizeGaugeMap, UNSEQUENCE, FILE_LEVEL_SIZE);
     }
   }
 
-  public long getFileSize(boolean seq) {
-    return seq ? seqFileSize.get() : unseqFileSize.get();
-  }
-
   public long getFileNum(boolean seq) {
```
OneSizeFitsQuorum, line 568: is this a testonly function now?

OneSizeFitsQuorum (approved):
LGTM~
After this refactoring, the code is much clearer

THUMarkLau (approved):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/metrics/file/TsFileMetrics.java
```diff
@@ -0,0 +1,413 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class TsFileMetrics implements IMetricSet {
+  private static final Logger LOGGER = LoggerFactory.getLogger(TsFileMetrics.class);
+  private AbstractMetricService metricService = null;
```
OneSizeFitsQuorum, line 44: use AtomicReference？

PR #11231, opened 2023-09-28
Optimize result set iteration in Python Client

HTHou:
## Description

This PR is optimizing the result set iteration in Python Client. 

* When build RowRecord for each line, the Field list can be reused.
* When deserialize the data points from the bytes, we can use numpy to do the deserialization and pandas.DataFrame to 
cache the fetched result. numpy and pandas are accelerated by C.

## How to test
First using this script to insert data.
```python
from datetime import datetime
import numpy as np

from iotdb.Session import Session
from iotdb.utils.IoTDBConstants import TSDataType, TSEncoding, Compressor
from iotdb.utils.NumpyTablet import NumpyTablet

# creating session connection.
ip = "127.0.0.1"
port_ = "6667"
username_ = "root"
password_ = "root"
session = Session(ip, port_, username_, password_, fetch_size=1024, zone_id="UTC+8", enable_redirection=False)

session.open(False)

device = "root.sg.d"
measurements_ = []
ts_path_lst_ = []
values_ = []
for i in range(440):
    measurement = "s" + str(i)
    measurements_.append(measurement)
    ts_path_lst_.append(device + "." + measurement)
    values_.append(1.1 * i)

data_type_lst_ = [TSDataType.FLOAT for _ in range(440)]
encoding_lst_ = [TSEncoding.GORILLA for _ in range(440)]
compressor_lst_ = [Compressor.LZ4 for _ in range(440)]
# 创建单设备 440 Float 类型序列
session.create_multi_time_series(
    ts_path_lst_, data_type_lst_, encoding_lst_, compressor_lst_
)

# insertRecord 插入10000行数据
startTime = int(datetime.now().timestamp() * 1000)
for i in range(10000):
    session.insert_record(device, 1692761800000 + i, measurements_, data_type_lst_, values_)
print("InsertRecord 10000 rows cost: " + str(int(datetime.now().timestamp() * 1000) - startTime) + "ms")

# insertRecords 插入10000行数据耗时
measurements_list_ = []
values_list_ = []
data_type_list_ = []
device_ids_ = []

for i in range(100):
    device_ids_.append(device)
    data_type_list_.append(data_type_lst_)
    measurements_list_.append(measurements_)
    values_list_.append(values_)
# 插入100次 每次100行
startTime = int(datetime.now().timestamp() * 1000)
for i in range(100):
    time_list = []
    for j in range(100):
        time_list.append(1692761900000 + i * 100 + j)
    session.insert_records(
        device_ids_, time_list, measurements_list_, data_type_list_, values_list_
    )
print("InsertRecords 10000 rows cost: " + str(int(datetime.now().timestamp() * 1000)- startTime) + "ms")

# insertTablet 插入10000行数据
np_values_ = []
for i in range(440):
    np_values_.append(np.array([1.1 * i for _ in range(100)], TSDataType.FLOAT.np_dtype()))

# 插入100次 每次100行
startTime = int(datetime.now().timestamp() * 1000)
for i in range(100):
    np_timestamps_ = np.array([1692762000000 + i * 100 + j for j in range(100)], TSDataType.INT64.np_dtype())
    np_tablet_ = NumpyTablet(
        device, measurements_, data_type_lst_, np_values_, np_timestamps_
    )
    session.insert_tablet(np_tablet_)
print("InsertTablet by numpy 10000 rows cost: " + str(int(datetime.now().timestamp() * 1000) - startTime) + "ms")
print("All executions done!!")
```

Then execute the following code to run the test.
```python
from datetime import datetime
import numpy as np

from iotdb.Session import Session
from iotdb.utils.IoTDBConstants import TSDataType, TSEncoding, Compressor
from iotdb.utils.NumpyTablet import NumpyTablet

# creating session connection.
ip = "127.0.0.1"
port_ = "6667"
username_ = "root"
password_ = "root"
session = Session(
    ip,
    port_,
    username_,
    password_,
    fetch_size=1024,
    zone_id="UTC+8",
    enable_redirection=False,
)
session.open()
# startTime = int(datetime.now().timestamp() * 1000)
# with session.execute_query_statement("select ** from root") as data_set:
#     data_set.todf()
#
# print("todf cost: " + str(int(datetime.now().timestamp() * 1000) - startTime) + "ms")
startTime = int(datetime.now().timestamp() * 1000)
with session.execute_query_statement("select * from root.**") as data_set:
    data_set.get_column_names()
    data_set.get_column_types()
    while data_set.has_next():
        data_set.next()
        # print(data_set.next())
print("Query cost: " + str(int(datetime.now().timestamp() * 1000) - startTime) + "ms")
session.close()
exit(0)

```

## Test result

Test with Python 3.10

Before this PR: Query cost: **29955ms**

After this PR: Query cost: **1983ms**

Test with Python 3.11

Before this PR: Query cost: **19374ms**

After this PR: Query cost: **1494ms**


iotdb-client/client-py/iotdb/utils/IoTDBRpcDataSet.py
```diff
@@ -51,62 +55,70 @@ def __init__(
...
         self.__is_closed = False
         self.__empty_resultSet = False
-        self.__has_cached_record = False
+        self.has_cached_record = False
         self.__rows_index = 0
+        self.is_null_info = [
+            False for _ in range(len(self.column_type_deduplicated_list))
+        ]
 
     def close(self):
```
ycycse, line 123: Should we release the reference to memoryview so that the garbage collection in python can collect it?

ycycse (approved):
LGTM! But I have some questions:

iotdb-client/client-py/iotdb/utils/Field.py
```diff
@@ -62,70 +57,72 @@ def is_null(self):
         return self.__data_type is None
 
     def set_bool_value(self, value):
```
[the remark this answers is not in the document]
HTHou: Fixed

iotdb-client/client-py/iotdb/utils/IoTDBRpcDataSet.py
HTHou: Fixed

iotdb-client/client-py/iotdb/utils/Field.py
```diff
@@ -61,71 +58,73 @@ def get_data_type(self):
...
+    def set_int_value(self, value: int):
+        self.value = value
 
     def get_int_value(self):
         if self.__data_type is None:
             raise Exception("Null Field Exception!")
-        return self.__int_value
+        if self.__data_type != TSDataType.INT32:
+            return None
+        return np.int32(self.value)
```
ycycse, line 79: Should we check if `self.value` is None and raise Exception (or return None, `np.int32(None)` would cause unexpected TypeError)?

iotdb-client/client-py/iotdb/utils/Field.py
```diff
@@ -61,71 +58,73 @@ def get_data_type(self):
...
+    def set_long_value(self, value: int):
+        self.value = value
 
     def get_long_value(self):
         if self.__data_type is None:
             raise Exception("Null Field Exception!")
-        return self.__long_value
+        if self.__data_type != TSDataType.INT64:
+            return None
+        return np.int64(self.value)
```
ycycse, line 89: The same as `np.int32`

iotdb-client/client-py/iotdb/utils/Field.py
```diff
@@ -61,71 +58,73 @@ def get_data_type(self):
...
+    def set_float_value(self, value: float):
+        self.value = value
 
     def get_float_value(self):
         if self.__data_type is None:
             raise Exception("Null Field Exception!")
-        return self.__float_value
+        if self.__data_type != TSDataType.FLOAT:
+            return None
+        return np.float32(self.value)
```
ycycse, line 99: `np.float32` would return `np.nan`. Is this an expected result?

iotdb-client/client-py/iotdb/utils/Field.py
```diff
@@ -134,22 +133,19 @@ def get_object_value(self, data_type):
         """
         :param data_type: TSDataType
         """
-        if self.__data_type is None:
+        if self.__data_type is None or self.value is None or self.value is pd.NA:
```
ycycse, line 136: Do you mean `np.NaN`? I don't found the calculation about pandas in this file.
```suggestion
        if self.__data_type is None or self.value is None or self.value is np.NaN:
```

iotdb-client/client-py/iotdb/utils/IoTDBRpcDataSet.py
```diff
@@ -142,30 +137,141 @@ def close(self):
...
             return False
         if self.fetch_results():
-            self.construct_one_row()
+            self.construct_one_data_frame()
             return True
         return False
 
-    def has_cached_result(self):
-        return (self.__query_data_set is not None) and (
-            len(self.__query_data_set.time) != 0
```
ycycse, line 147: Should we keep len(self.__query_data_set.time) != 0? Maybe the resultSet is already an emptySet.

iotdb-client/client-py/iotdb/utils/Field.py
HTHou: Fixed

iotdb-client/client-py/iotdb/utils/Field.py
HTHou: Fixed

iotdb-client/client-py/iotdb/utils/Field.py
HTHou: No. If self.value == None, I think the result should be None. Added check for the self.value.

iotdb-client/client-py/iotdb/utils/Field.py
HTHou: Not the `np.NaN`.
pd.NA will appear when the dataframe has some null value. For example, there is a dataset looks like this.
image
In Python client, pd.NA may need to process.
image

iotdb-client/client-py/iotdb/utils/IoTDBRpcDataSet.py
HTHou: Fixed

xingtanzjr (approved):

PR #11243, opened 2023-10-08
Capture build scans on ge.apache.org to benefit from deep build insights

clayburn:
@chrisdutz - It was nice to meet you at Community over Code today. This is the PR we discussed to publish Build Scans to ge.apache.org.

## Description

This PR publishes a build scan for every CI build and for every local build from an authenticated Apache committer. The build will not fail if publishing fails.

The build scans of the Apache iotdb project are published to the Develocity instance at [ge.apache.org](https://ge.apache.org/), hosted by the Apache Software Foundation and run in partnership between the ASF and Gradle. This Develocity instance has all features and extensions enabled and is freely available for use by the Apache iotdb project and all other Apache projects.

On this Develocity instance, Apache iotdb will have access not only to all of the published build scans but other aggregate data features such as:

- Dashboards to view all historical build scans, along with performance trends over time
- Build failure analytics for enhanced investigation and diagnosis of build failures
- Test failure analytics to better understand trends and causes around slow, failing, and flaky tests

Please let me know if there are any questions about the value of Develocity or the changes in this pull request and I’d be happy to address them.

This PR has:
- [x] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

### DIFF
--- a/.github/workflows/client-go.yml
+++ b/.github/workflows/client-go.yml
@@ -28,6 +28,7 @@ concurrency:
 
 env:
   MAVEN_OPTS: -Dhttp.keepAlive=false -Dmaven.wagon.http.pool=false -Dmaven.wagon.http.retryHandler.class=standard -Dmaven.wagon.http.retryHandler.count=3
+  GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GE_ACCESS_TOKEN }}
 
 jobs:
   unix:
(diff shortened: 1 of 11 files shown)


chrisdutz:
So mabe some additional information on this ... I learned at CommunityOverCode that we have a gradle-enterprise instance at Apache, which also is ale to enrich Maven builds. The probably most interesting thing is that with these extensions we should be able to detect which tests need to be run to test changed code of a commit and it allows running these first. I thought we could probably give that a try ... I probably need to check with Infra, how we can use this on our self-hosted runners though.

clayburn:
For what it's worth, the `GE_ACCESS_TOKEN` is an org-wide secret managed by ASF Infra. I know this is in place and available to use. GitHub Actions does not allow PRs from forks to access secrets, so you won't see any publishing from this PR.

I'm not a GitHub Actions expert, but I believe that secrets are propagated to self-hosted runners. If I remember correctly, this is how Apache Beam do it. Still a good conversation that we can have with Infra today to verify.

PR #11261, opened 2023-10-10
Load: skipping instead of throwing exceptions when loading empty tsfiles

ppppoooo:
When the loaded TsFile is empty, continue to load the next


MiniSho (approved):
LGTM！:)

SteveYurongSu (approved):

PR #11273, opened 2023-10-10
[IOTDB-6185] Fix show cluster null status bug

CRZbulabula:
Some Node's status might be `null` if the ConfigNode can't send heartbeat request to them. Therefore, the ConfigNode will initiate all Node's status to `Unknown` and make sure all Node's status are either exist or set to `Unknown` by default.


OneSizeFitsQuorum (approved):
LGTM

PR #11287, opened 2023-10-11
Fix potential loss of metrics when restart confignode

SpriCoder:
1. Move add metrics from init to leader ready
2. Move the management of replication factor metric into PartitionMetrics

### DIFF
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/partition/PartitionMetrics.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/partition/PartitionMetrics.java
@@ -263,7 +263,7 @@ public class PartitionMetrics implements IMetricSet {
 
     List<String> databases = getClusterSchemaManager().getDatabaseNames();
     for (String database : databases) {
-      unbindDatabasePartitionMetricsWhenUpdate(metricService, database);
+      unbindDatabaseRelatedMetricsWhenUpdate(metricService, database);
     }
   }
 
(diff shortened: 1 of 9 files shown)


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/schema/ClusterSchemaManager.java
```diff
@@ -261,25 +240,11 @@ public TSStatus alterDatabase(DatabaseSchemaPlan databaseSchemaPlan) {
...
-      metricService
-          .getOrCreateGauge(
-              Metric.REPLICATION_FACTOR.toString(),
-              MetricLevel.CORE,
-              Tag.TYPE.toString(),
-              SCHEMA,
-              Tag.DATABASE.toString(),
-              databaseSchemaPlan.getSchema().getName())
-          .set(databaseSchemaPlan.getSchema().schemaReplicationFactor);
+      PartitionMetrics.bindDatabaseReplicationFactorMetrics(
```
OneSizeFitsQuorum, line 243: If changing the number of replica isn't currently supported, maybe we shouldn't do it here

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/partition/PartitionMetrics.java
```diff
@@ -274,23 +326,11 @@ public static void bindDatabasePartitionMetrics(
         database,
         Tag.TYPE.toString(),
         TConsensusGroupType.DataRegion.toString());
+    bindDatabaseReplicationFactorMetrics(
```
OneSizeFitsQuorum, line 329: If put this line in it, we should revise the function bindDatabasePartitionMetricsWhenUpdate name?

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/schema/ClusterSchemaManager.java
SpriCoder: But maybe I think we need to do that here?

OneSizeFitsQuorum (approved):
LGTM~
Awesome work!

PR #11291, opened 2023-10-12
[IOTDB-6187] IoTConsense WAL Interval Reading Strategy Optimization

BUAAserein:
#### Description

IoTConsensus WAL always resets the insertNodes when WAL skip to a new targetIndex. But in most writting process, targetIndex is sequentially increasing, which means we may use the previous insertNodes as a cache and doesn't need to reset it everytime.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/node/WALNode.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/node/WALNode.java
@@ -611,7 +611,7 @@ public class WALNode implements IWALNode {
     /** true means filesToSearch and currentFileIndex are outdated, call updateFilesToSearch */
     private boolean needUpdatingFilesToSearch = true;
     /** batch store insert nodes */
-    private final List<IndexedConsensusRequest> insertNodes = new LinkedList<>();
+    private final LinkedList<IndexedConsensusRequest> insertNodes = new LinkedList<>();
     /** iterator of insertNodes */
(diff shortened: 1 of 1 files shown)


caozj1011 (approved):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/node/WALNode.java
```diff
@@ -833,6 +833,21 @@ public void skipTo(long targetIndex) {
             targetIndex,
             targetIndex);
       }
+
+      if (itr != null
+          && itr.hasNext()
+          && insertNodes.get(itr.nextIndex()).getSearchIndex() <= targetIndex
+          && targetIndex <= insertNodes.get(insertNodes.size() - 1).getSearchIndex()) {
```
OneSizeFitsQuorum, line 840: How about changing the type of insertNodes to LinkedList, so you can use `insertNodes.getLast().getSearchIndex()` here

OneSizeFitsQuorum (approved):
LGTM~
Please do some performance testing and post the result to this pr~

BUAAserein:
After testing the long-term and short-term performance, we found that the write performance decreased while the sync performance increased, which means that the sync performance was improved. Writing performance dropped by approximately 0.61%, and synchronization performance increased by approximately 0.29%.

OneSizeFitsQuorum:
see [doc](https://apache-iotdb.feishu.cn/docx/YJDldH3eoo2hffx70oxcuE1XnUg) for details

PR #11264, opened 2023-10-10
[IOTDB-6182] Add fsync in ConfigNode persistence modules

CRZbulabula:
Use fsync when the ConfigNodes need to take snapshot, which is safer than previous implementation.

PR #11296, opened 2023-10-12
[To rel/1.2][IOTDB-6182] Add fsync in ConfigNode persistence modules

CRZbulabula:
Use fsync when the ConfigNodes need to take snapshot, which is safer than previous implementation.

cherry-pick [pull](https://github.com/apache/iotdb/pull/11264)


OneSizeFitsQuorum (approved):
LGTM

PR #11297, opened 2023-10-12
Fix the problem of inaccurate GC monitor detection at the beginning and adjusting the alert threshold.

Pengzna:
1. Fix the problem of inaccurate GC monitor detection at the beginning.

- When the Metric Service starts after the jvm, the gc time before the Metric Service is started will also be counted, causing the GC percentage within the initial time window to be significantly higher than expected, as shown in this figure:
修复前

- Fix this bug by only calculating GC time within the time window. The effect after fixing this bug is shown in this figure:
修复后

2.   Lowered the gc percentage alert threshold from 50% to 20%.


OneSizeFitsQuorum (approved):
LGTM

PR #11301, opened 2023-10-13
[To Rel/1.2] Fix the problem of inaccurate GC monitor detection at the beginning and adjusting the alert threshold.

Pengzna:
cherry picked from #11297


OneSizeFitsQuorum (approved):
LGTM

Issue #11133, opened 2023-09-13
[Bug] 


cqzhang7:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/iotdb/issues) and found nothing similar.

### Version

操作系统：CentOS 7.9
IoTDB：1.12

### Describe the bug and provide the minimal reproduce step

![36b286f9031b417a41072392cbcdba0](https://github.com/apache/iotdb/assets/49841963/ab0f98f8-1641-4273-ae51-efab7ce17bf5)
使用iotdb的分段分组聚合，指定聚合窗口大小和步长为1y，发现聚合结果步长是固定的365天，这并没有考虑到闰年为366天的事实。
我查询了基于月是可以实现每28|29|30|31天作为自然月来分组的
![image](https://github.com/apache/iotdb/assets/49841963/288e70b7-ac58-4a68-868d-86b7569484a1)
这应当是一个错误，请尽快解决一下

### What did you expect to see?

期望的查询结果应是
+-----------------------------+----+----+-------+------+
|                         Time|  pv|load|selfuse|charge|
+-----------------------------+----+----+-------+------+
|2023-01-01T00:00:00.000+08:00|  39|   3|      0|    35|
|2024-01-01T00:00:00.000+08:00|null|null|   null|  null|
|2025-01-01T00:00:00.000+08:00|null|null|   null|  null|
|2026-01-01T00:00:00.000+08:00|null|null|   null|  null|
+-----------------------------+----+----+-------+------+

### What did you see instead?

实际的查询结果是
+-----------------------------+----+----+-------+------+
|                         Time|  pv|load|selfuse|charge|
+-----------------------------+----+----+-------+------+
|2023-01-01T00:00:00.000+08:00|  39|   3|      0|    35|
|2024-01-01T00:00:00.000+08:00|null|null|   null|  null|
|2024-12-31T00:00:00.000+08:00|null|null|   null|  null|
|2025-12-31T00:00:00.000+08:00|null|null|   null|  null|
+-----------------------------+----+----+-------+------+

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #11309, opened 2023-10-14
[IOTDB-6191] Fix group by year

Wei-hao-Li:
github issue https://github.com/apache/iotdb/issues/11133
doc https://apache-iotdb.feishu.cn/docx/TNBPd8nZUoDsJMxjdOncOUkcnVA

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/thrift/impl/DataNodeInternalRPCServiceImpl.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/thrift/impl/DataNodeInternalRPCServiceImpl.java
@@ -1274,10 +1274,10 @@ public class DataNodeInternalRPCServiceImpl implements IDataNodeRPCService.Iface
   private void sampleDiskLoad(TLoadSample loadSample) {
     final CommonConfig commonConfig = CommonDescriptor.getInstance().getConfig();
 
-    double freeDisk =
+    double availableDisk =
         MetricService.getInstance()
             .getAutoGauge(
-                SystemMetric.SYS_DISK_FREE_SPACE.toString(),
+                SystemMetric.SYS_DISK_AVAILABLE_SPACE.toString(),
                 MetricLevel.CORE,
                 Tag.NAME.toString(),
                 SYSTEM)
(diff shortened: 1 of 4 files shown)


liuminghui233:
It is a good idea to add duration unit fields in `GroupByComponent`. But I think we should go one step further - define a `TimeDuration` class like this: 
```java
public class TimeDuration {
    private final long duration;
    private final DurationUnit unit;
}
```

The `GroupByComponent` will then contain the following fields:
```java
// [startTime, endTime)
private long startTime;
private long endTime;

// time interval
private TimeDuration interval;

// sliding step
private TimeDuration slidingStep;

// if it is left close and right open interval
private boolean leftCRightO;
```

I believe this will make the code clearer and more readable.

What do you think? @Wei-hao-Li @JackieTien97

JackieTien97:
> It is a good idea to add duration unit fields in `GroupByComponent`. But I think we should go one step further - define a `TimeDuration` class like this:
> 
> ```java
> public class TimeDuration {
>     private final long duration;
>     private final DurationUnit unit;
> }
> ```
> 
> The `GroupByComponent` will then contain the following fields:
> 
> ```java
> // [startTime, endTime)
> private long startTime;
> private long endTime;
> 
> // time interval
> private TimeDuration interval;
> 
> // sliding step
> private TimeDuration slidingStep;
> 
> // if it is left close and right open interval
> private boolean leftCRightO;
> ```
> 
> I believe this will make the code clearer and more readable.
> 
> What do you think? @Wei-hao-Li @JackieTien97

good idea,  that would be more clear.

cqzhang7:
This PR conflicts with mine,I think the focus should be to evaluate whether this change improves the calculation performance, otherwise why not replace the input '1y' to '12mo', which would be a simple solution, and doesn't introduce any additional computational burden  @HTHou  @JackieTien97

liuminghui233:
> This PR conflicts with mine,I think the focus should be to evaluate whether this change improves the calculation performance, otherwise why not replace the input '1y' to '12mo', which would be a simple solution, and doesn't introduce any additional computational burden @HTHou @JackieTien97

Yes, replacing '1y' with '12mo' might be a simple solution. @Wei-hao-Li  you can think about it.

Wei-hao-Li:
> This PR conflicts with mine,I think the focus should be to evaluate whether this change improves the calculation performance, otherwise why not replace the input '1y' to '12mo', which would be a simple solution, and doesn't introduce any additional computational burden @HTHou @JackieTien97

Your think is right.

Wei-hao-Li:
@cqzhang7 The fixed PR using your method is [here](https://github.com/apache/iotdb/pull/11323) , thanks for your contribution.

PR #11320, opened 2023-10-16
[IOTDB-6192] Remove redundant mergeSortOperator in 'ORDER BY TIME ALIGN BY DEVICE'

ycycse:
https://issues.apache.org/jira/browse/IOTDB-6192
If there is only one device, there is no need to use mergeSortOperator.


Beyyes (approved):
LGTM

PR #11350, opened 2023-10-19
[IOTDB-6198] Pipe: Memory Management Framework

MiniSho:
## Description
Pipe module memory is divided into two main sections: Static Memory and Dynamic Memory.
* static memory: the lifecycle is bending to a specific pipe.
* dynamic memory: related to Pipe framework or user's custom logic operation.

So if we want to manage pipe memory, we have to build a framework which implements the allocation and release of memory within the Pipe module.
The framework provides `allocate()` & `release` methods.


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/conf/IoTDBDescriptor.java
```diff
@@ -1709,15 +1709,18 @@ private void initMemoryAllocate(Properties properties) {
             maxMemoryAvailable * Integer.parseInt(proportions[1].trim()) / proportionSum);
         conf.setAllocateMemoryForSchema(
             maxMemoryAvailable * Integer.parseInt(proportions[2].trim()) / proportionSum);
-        conf.setAllocateMemoryForConsensus(
+        conf.setAllocateMemoryForPipe(
             maxMemoryAvailable * Integer.parseInt(proportions[3].trim()) / proportionSum);
+        conf.setAllocateMemoryForConsensus(
+            maxMemoryAvailable * Integer.parseInt(proportions[4].trim()) / proportionSum);
```
lancelly, line 1715: It seems that pipe memory should be proportions[4].

SteveYurongSu (approved):

PR #11357, opened 2023-10-20
Use bash instead of sh in integration test.

Beyyes:
## Description

Some CI environment is ubuntu, but `sh` in ubuntu is indeed `dash` which is incompatible from `bash`.

So change `sh` to `bash` in integration-test.


wangchao316 (approved):
LGTM

PR #11368, opened 2023-10-23
Fix the failure of metric module

SpriCoder:
Fix the failure of metric module. Because we use SPI in metric module, so we shouldn't set metric related dependency's scope as test which may lead to failure


OneSizeFitsQuorum (commented):
LGTM~
BTW, I think we can add an IT for this to avoid such things in the future

OneSizeFitsQuorum (approved):

PR #11378, opened 2023-10-24
[IOTDB-6206] Metric: merge metric algorithm library and optimize some core components

Pengzna:
[IOTDB-6206](https://issues.apache.org/jira/projects/IOTDB/issues/IOTDB-6206?filter=allissues)
1. Compare the performance of the metric algorithm library, merge them, and remove the library selection parameter.
2. Try to optimize the performance of some core components, such as counter, rate.
3. Detailed docs: https://apache-iotdb.feishu.cn/docx/P5svdKZTxoVtqax0mgvcCeYInQe

### DIFF
--- a/iotdb-core/datanode/pom.xml
+++ b/iotdb-core/datanode/pom.xml
@@ -260,6 +260,16 @@
             <artifactId>jetty-servlet</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.glassfish.jersey.inject</groupId>
+            <artifactId>jersey-hk2</artifactId>
+            <scope>runtime</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <scope>runtime</scope>
+        </dependency>
+        <dependency>
             <groupId>com.github.ben-manes.caffeine</groupId>
             <artifactId>caffeine</artifactId>
         </dependency>
@@ -316,11 +326,6 @@
         <dependency>
             <groupId>org.apache.httpcomponents</groupId>
             <artifactId>httpcore</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-databind</artifactId>
             <scope>test</scope>
         </dependency>
         <dependency>
@@ -523,6 +528,8 @@
                         <!-- We need these dependencies as they provide the metric managers used in tests -->
                         <usedDependency>org.apache.iotdb:dropwizard-metrics</usedDependency>
                         <usedDependency>org.apache.iotdb:micrometer-metrics</usedDependency>
+                        <!-- This dependency is required at runtime, when esnabling the rest service -->
+                        <usedDependency>org.glassfish.jersey.inject:jersey-hk2</usedDependency>
                     </usedDependencies>
                 </configuration>
             </plugin>
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/rest/filter/AuthorizationFilter.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/rest/filter/AuthorizationFilter.java
@@ -155,7 +155,8 @@ public class AuthorizationFilter implements ContainerRequestFilter, ContainerRes
   public void filter(
       ContainerRequestContext requestContext, ContainerResponseContext responseContext)
       throws IOException {
-    if (SESSION_MANAGER.getSessionInfo(SESSION_MANAGER.getCurrSession()) != null) {
+    if (SESSION_MANAGER.getCurrSession() != null
+        && SESSION_MANAGER.getSessionInfo(SESSION_MANAGER.getCurrSession()) != null) {
       SESSION_MANAGER.removeCurrSession();
     }
   }
--- a/pom.xml
+++ b/pom.xml
@@ -1730,6 +1730,11 @@
                 <artifactId>orc-shims</artifactId>
                 <version>${orc-core.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.glassfish.jersey.inject</groupId>
+                <artifactId>jersey-hk2</artifactId>
+                <version>${jersey.version}</version>
+            </dependency>
         </dependencies>
     </dependencyManagement>
     <build>


iotdb-core/datanode/src/test/java/org/apache/iotdb/db/metric/MetricServiceTest.java
```diff
@@ -55,30 +50,31 @@ public class MetricServiceTest {
...
-      testNormalSituation();
-
-      testOtherSituation();
-
-      // stop metric module
-      metricService.stopService();
-    }
-  }
+  //  @Test
+  //  public void testMetricService() {
```
OneSizeFitsQuorum, line 54: Do not comment test.
BTW, I think this test(line58 - line 75) is still useful

iotdb-core/metrics/core/src/main/java/org/apache/iotdb/metrics/core/type/IoTDBCounter.java
```diff
@@ -17,29 +17,35 @@
...
 import org.apache.iotdb.metrics.type.Counter;
 
-public class MicrometerCounter implements Counter {
+import java.util.concurrent.atomic.LongAdder;
+
+public class IoTDBCounter implements Counter {
+  private final LongAdder count;
+  // Here we retain the micrometer counter variable in order to use its naming system within
+  // micrometer.
   io.micrometer.core.instrument.Counter counter;
```
OneSizeFitsQuorum, line 30: If we just want to use its naming system, do we need to hold a reference here? Seems that it is never used?

iotdb-core/metrics/core/src/main/java/org/apache/iotdb/metrics/core/uitls/IoTDBCachedGauge.java
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  private boolean shouldLoad() {
+    while (true) {
+      final long time = clock.getTick();
+      final long current = reloadAt.get();
+      if (current > time) {
+        return false;
+      }
+      if (reloadAt.compareAndSet(current, time + timeoutNS)) {
```
OneSizeFitsQuorum, line 102: Perhaps when time is much larger than current (e.g., many times we have not come to get), we still only have cas once instead of multiple times

iotdb-core/metrics/interface/src/test/java/org/apache/iotdb/metrics/config/MetricConfigTest.java
```diff
@@ -54,7 +53,7 @@ public void testConfigNodeMetricConfig() {
     MetricConfig metricConfig = MetricConfigDescriptor.getInstance().getMetricConfig();
 
     assertEquals(3, metricConfig.getMetricReporterList().size());
-    assertEquals(MetricFrameType.DROPWIZARD, metricConfig.getMetricFrameType());
+    //    assertEquals(MetricFrameType.DROPWIZARD, metricConfig.getMetricFrameType());
```
OneSizeFitsQuorum, line 56: remove this line?

iotdb-core/metrics/interface/src/test/java/org/apache/iotdb/metrics/config/MetricConfigTest.java
```diff
@@ -93,7 +92,7 @@ public void testDataNodeMetricConfig() {
...
```
OneSizeFitsQuorum, line 95: remove this line?

iotdb-core/metrics/pom.xml
```diff
@@ -31,8 +31,8 @@
     <name>IoTDB: Core: Metrics</name>
     <description>Metrics interfaces for IoTDB</description>
     <modules>
+        <module>core</module>
         <module>interface</module>
-        <module>micrometer-metrics</module>
         <module>dropwizard-metrics</module>
```
OneSizeFitsQuorum, line 36: How about just removing whole dropwizard-metrics module

OneSizeFitsQuorum (commented):
Awesome work！Through this work, we will become more familiar with the difference between micrometer and dropwizard and then find the most suitable algorithm library implementation for IoTDB itself.
I have left some minor issue, PTAL

OneSizeFitsQuorum (commented):
Awesome work！Through this work, we will become more familiar with the difference between micrometer and dropwizard and then find the most suitable algorithm library implementation for IoTDB itself.
I have left some minor issue, PTAL

iotdb-core/datanode/src/test/java/org/apache/iotdb/db/metric/MetricServiceTest.java
SpriCoder: +1, maybe we should leave the test of micrometer implement

iotdb-core/metrics/core/src/main/resources/META-INF/services/org.apache.iotdb.metrics.AbstractMetricManager
```diff
@@ -15,4 +15,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-org.apache.iotdb.metrics.micrometer.MicrometerMetricManager
```
SpriCoder, line 18: Maybe there is no need to use SPI if we choose micrometer as implementation

iotdb-core/metrics/pom.xml
SpriCoder: +1, and maybe there is no need to split core and micrometer. What's your opinion?

SpriCoder (changes requested):
Great Job! Some problems PTAL~
At the same time, please check UT

iotdb-core/metrics/core/src/main/resources/META-INF/services/org.apache.iotdb.metrics.AbstractMetricManager
Pengzna: good idea!

iotdb-core/datanode/src/test/java/org/apache/iotdb/db/metric/MetricServiceTest.java
Pengzna: Has fixed.

iotdb-core/metrics/interface/src/test/java/org/apache/iotdb/metrics/config/MetricConfigTest.java
Pengzna: Has fixed.

iotdb-core/metrics/interface/src/test/java/org/apache/iotdb/metrics/config/MetricConfigTest.java
Pengzna: Has fixed.

PR #11383, opened 2023-10-25
Avoid modification causing OOM while there is much deletion entry in mods file

JackieTien97:
More details can be seen in https://jira.infra.timecho.com:8443/browse/TIMECHODB-349.

In this pr:
1. For loading from disk case, we delete `filePathModCache` in `QueryContext`,  save List<Modification> in `IChunkMetadataLoader`.
2. for loading from memory, we generating `TimeSeriesMetadata` from each `ChunkMetadata`, so we can directly use isModified function of `ChunkMetadata` to set modified of `TimeSeriesMetadata`.

In this way, we can safely delete `filePathModCache` without sacrificing query performance.


Beyyes (approved):
LGTM

PR #11392, opened 2023-10-26
add tsfile correctness validation for write compaction and load

caozj1011:
At present, the system has the basis that write, merge and Load will generate TsFile, and whether the system can run normally when the generated file is correct. This PR has optimized TsFile verification and added the verification of TsFile correctness.


iotdb-core/node-commons/src/assembly/resources/conf/iotdb-common.properties
```diff
@@ -300,6 +300,24 @@ cluster_name=defaultCluster
...
+# write_tsfile_validation_level=RESOURCE_ONLY
+
+# The level of validation after compaction or load generate target file
+# The details of these three levels are as follows:
+# 1. NONE: the validation after compaction is disabled.
+# 2. RESOURCE_ONLY: validate resource file is correct.
+# 3. TSFILE_INTEGRITY: validate TsFile and resource file is complete.
+# 4. TSFILE_DATA_CORRECTNESS: validate TsFile and resource file is complete and correct.
+# Datatype: String
+# rewrite_tsfile_validation_level=RESOURCE_ONLY
```
HTHou, line 319: Move these to Storage Engine part?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/utils/CompactionUtils.java
```diff
@@ -493,4 +497,225 @@ public static boolean isDiskHasSpace(double redundancy) {
...
+            int dataSize = header.getDataSize();
+            if (dataSize == 0) {
+              break;
+            }
+
+            boolean isHasStatistic = (header.getChunkType() & 0x3F) == MetaMarker.CHUNK_HEADER;
+            Decoder defaultTimeDecoder =
+                Decoder.getDecoderByType(
+                    TSEncoding.valueOf(TSFileDescriptor.getInstance().getConfig().getTimeEncoder()),
+                    TSDataType.INT64);
```
HeimingZ, line 546: Perhaps it‘s unsafe to use default decoder here if users change time encoder when restarting, can we get the decoder type from the tsfile? But I find TSFileDescriptor.getInstance().getConfig().setTimeEncoder is only used in the test methods, so it's temporarily safe to use like this.

shuwenwei (approved):

xingtanzjr (approved):

xingtanzjr:
image

RUN IT locally

xingtanzjr:
For failed tests, rerun succeeded.

image

xingtanzjr:
For previous failed UT in this CI https://github.com/apache/iotdb/actions/runs/6705382433/job/18219698479?pr=11392#logs

image

xingtanzjr:
image

xingtanzjr (approved):

PR #11396, opened 2023-10-26
Accelerate datanode recovery

HTHou:
## Description

This PR is optimizing two points.

1. #5800 added a strategy. When execute stop scripts, the data in memory will flush to disk. However, this strategy was skipped.
2. Compaction threads may take too much CPU resource, which caused recovery being slow. We can start compaction after all data region recovered.


wangchao316 (approved):
LGTM

PR #11399, opened 2023-10-27
Add reason to set ReadOnly status when compaction recovery is failed

shuwenwei:
## Description
Add reason to set ReadOnly status when compaction recovery is failed


wangchao316 (approved):
LGTM

HTHou (approved):

PR #11404, opened 2023-10-27
Fix TsFileResource endTime error when insert empty tablet and flush

HTHou:
## Description

When data is written using the insertTablet() interface and both values are null, exactly the data written is flushed. An empty TsFile file will be generated, the space cannot be deleted, and the generated resource file will be incorrect, and the endTime of the device will be Long.MIN_VALUE. This in turn affects query and merge mechanisms.

Based on this, this PR has the following two optimizations:
1. In the flush process, it is added that if the resource file is empty, the generated TsFile and.resource file are deleted
2. Enhance the judgment logic when there is no valid data in MemTable. If there is no valid data, delete the related empty Device data from the resource.

## How to test
Execute the following code.

```java
    session =
        new Session.Builder()
            .host(LOCAL_HOST)
            .port(6667)
            .username("root")
            .password("root")
            .version(Version.V_1_0)
            .build();
    session.open(false);
    insertTablet();
    session.executeNonQueryStatement("flush");
    session.close();
  }


  private static void insertTablet() throws IoTDBConnectionException, StatementExecutionException {
    // The schema of measurements of one device
    // only measurementId and data type in MeasurementSchema take effects in Tablet
    List<MeasurementSchema> schemaList = new ArrayList<>();
    schemaList.add(new MeasurementSchema("s1", TSDataType.INT64));
    schemaList.add(new MeasurementSchema("s2", TSDataType.INT64));
    schemaList.add(new MeasurementSchema("s3", TSDataType.INT64));

    Tablet tablet = new Tablet(ROOT_SG1_D1, schemaList, 100);

    // Method 1 to add tablet data

      for (int s = 0; s < 3; s++) {
        tablet.addValue(schemaList.get(s).getMeasurementId(), rowIndex, null);
      }
        tablet.reset();
      }
      timestamp++;
    }

    if (tablet.rowSize != 0) {
      session.insertTablet(tablet);
      tablet.reset();
    }
  }

```

Then print the TsFileResource file, you will see the endTime is Long.MIN_VALUE.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/DataRegion.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/DataRegion.java
@@ -2053,7 +2053,7 @@ public class DataRegion implements IDataRegionForQuery {
     closeQueryLock.writeLock().lock();
     try {
       tsFileProcessor.close();
-      if (tsFileProcessor.isEmpty()) {
+      if (tsFileProcessor.isEmpty() || tsFileProcessor.getTsFileResource().isEmpty()) {
         try {
           fsFactory.deleteIfExists(tsFileProcessor.getTsFileResource().getTsFile());
           tsFileManager.remove(tsFileProcessor.getTsFileResource(), tsFileProcessor.isSequence());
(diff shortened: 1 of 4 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/DataRegion.java
```diff
@@ -2053,7 +2053,7 @@ private void closeUnsealedTsFileProcessorCallBack(TsFileProcessor tsFileProcesso
     closeQueryLock.writeLock().lock();
     try {
       tsFileProcessor.close();
-      if (tsFileProcessor.isEmpty()) {
+      if (tsFileProcessor.isEmpty() || tsFileProcessor.getTsFileResource().isEmpty()) {
```
wangchao316, line 2056: hi , tsfile exsit and resource does not exsit,  will delete tsfile?

HTHou: In this case, tsfile contains no data and resource file exists. 
I'll update the PR description later.

wangchao316: ok

wangchao316 (approved):
LGTM

xingtanzjr (approved):

caozj1011 (approved):

PR #11411, opened 2023-10-29
[IOTDB-6219] Fix the display problem of explain that the print result is not aligned

Beyyes:
## Description

Example - 1:
Before fix:
image

After fix:
image

Example -2:
Before fix:
image

After fix:
image


wangchao316 (approved):
LGTM

PR #11414, opened 2023-10-30
Add InsertionCrossSpaceCompaction to move non overlapped unsequence file to sequence space directly

shuwenwei:
## Description
1. Fix the bug of LastFlushTime
2. Add new CompactionTask type: InsertionCrossSpaceCompaction

### Fix the bug of LastFlushTime
LastFlushTime will not always be maintained in memory. When the memory is insufficient, this value in the memory may disappear. When used again, this value will be restored by scanning the TsFileResource in the time partition. When performing a Load operation, there is no recovery when updating this value, resulting in the updated value being incorrect. This results in overlapping data that may appear under normal writes.
### InsertionCrossSpaceCompaction
The data generated by Load will enter unsequence space, and these data will be rewritten into the sequence space through CrossSpaceCompaction. However, this data may not overlap with the data of the files in the sequence space. In this case, the new compaction type move such unsequence files directly to the sequence space to avoid repeated writing.
截屏2023-10-30 11 48 07


choubenson (approved):

xingtanzjr:
IT local run result
image

xingtanzjr:
RUN 5 failed IT，they  all succeeded
image

PR #11415, opened 2023-10-30
[IOTDB-6151] Move datanode system.properties safer

liyuheng55555:
Previous pr https://github.com/apache/iotdb/pull/11153 has some problem. 

Now we use replace-existing-copy. If copy success, then delete the original file.


OneSizeFitsQuorum (approved):
LGTM

PR #11419, opened 2023-10-30
degrade log level of pbtree memory status monitor

MarcosZyk:
## Description

In pressure tests, the schema memory usage always wander around the memory capacity, which results in tremendous log revealing memory status changes. 

Such changes are normal system behavior and only used for system status check. Therefore, we do not need to print these logs in industrial scenarios. It is necessary to change the log level from info or warn to debug.


Cpaulyz (approved):

bigreybear (approved):
LGTM~

PR #11432, opened 2023-10-31
Implement iterative volatile node process during pbtree flush

MarcosZyk:
### Motivation

The original flush implementation will collect all volatile node into a list for flush. Such mechanism will bring extra space and time costs, like duplicate reference storage and list scan.

### Modification

Implement an iterative process for volatile flush. Now the Cachemanager will return a iterator, used to iteratively visit volatile subtree in memory, and the CachedMTreeStore will use it to write nodes into file.


Cpaulyz (approved):
LGTM

PR #11451, opened 2023-11-01
[IOTDB-6227] Remove unused retry policy in IStateMachine

SzyWilliam:
The retry policy is never used in IStateMachine, yet it brings complexity in Ratis write while may incur poerformance penalties. We should remove it.
see https://issues.apache.org/jira/browse/IOTDB-6227


OneSizeFitsQuorum (approved):
LGTM

Cpaulyz:
What a great contribution! Thx!

PR #11454, opened 2023-11-01
Check empty target file after sequence InnerSpaceCompactionTask

shuwenwei:
## Description
1. Mark target file as DELETED status if the target file is empty in ReadChunkCompactionPerformer
2. Skip the target file which is marked as DELETED status when validate compaction result

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/performer/impl/ReadChunkCompactionPerformer.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/performer/impl/ReadChunkCompactionPerformer.java
@@ -92,6 +92,9 @@ public class ReadChunkCompactionPerformer implements ISeqCompactionPerformer {
         targetResource.updatePlanIndexes(tsFileResource);
       }
       writer.endFile();
+      if (writer.isEmptyTargetFile()) {
+        targetResource.forceMarkDeleted();
+      }
     }
   }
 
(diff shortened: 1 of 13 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/performer/impl/ReadChunkCompactionPerformer.java
```diff
@@ -92,6 +92,9 @@ public void perform()
         targetResource.updatePlanIndexes(tsFileResource);
       }
       writer.endFile();
+      if (writer.isEmptyTargetFile()) {
```
caozj1011, line 95: Do you need to judge other performers as well？

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/task/AbstractCompactionTask.java
```diff
@@ -372,21 +374,56 @@ public boolean isDiskSpaceCheckPassed() {
...
+  protected void validateCompactionResult(
+      List<TsFileResource> sourceSeqFiles,
+      List<TsFileResource> sourceUnseqFiles,
+      List<TsFileResource> targetFiles)
+      throws CompactionValidationFailedException {
+    // skip TsFileResource which is marked as DELETED status
+    List<TsFileResource> validTargetFiles =
+        targetFiles.stream().filter(resource -> !resource.isDeleted()).collect(Collectors.toList());
+    CompactionTaskType taskType = getCompactionTaskType();
+    boolean needToValidateTsFileCorrectness = taskType != CompactionTaskType.INSERTION;
```
caozj1011, line 386: These **CompactionTaskType** are supposed to be specified by task new or another step, but I can't find a place to assign a value to the task, make sure the task corresponds to its task type

PR #11469, opened 2023-11-03
Disable spotless for JDK 21

HTHou:
## Description

The current version of spotless plugin we are using cannot run with JDK 21. Therefore I disabled the spotless check for JDK 21. 

After this fixing, we can run `mvn clean package -pl distribution -am -DskipTests` with JDK21 directly.  
20d10792-c604-4cbc-9b6b-5b3867bb460d


CritasWang:
LGTM.

CritasWang (approved):

PR #11489, opened 2023-11-07
Pipe: Introduce PipeEventCommitManager to manage event commit order to avoid losing data during parallel connector scheduling

DanielWang2035:


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/protocol/IoTDBConnector.java
```diff
@@ -53,6 +61,11 @@ public abstract class IoTDBConnector implements PipeConnector {
 
   protected boolean isTabletBatchModeEnabled = true;
 
+  protected final AtomicLong commitIdGenerator = new AtomicLong(0);
```
Caideyipi, line 64: To ensure the ordering of the commiting across connectors, it may be better to make the 3 variables static, to let all the connectors share one commitId

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/protocol/IoTDBConnector.java
```diff
@@ -125,4 +138,39 @@ protected Set<TEndPoint> parseNodeUrls(PipeParameters parameters) {
 
     return givenNodeUrls;
   }
+
+  /**
```
Caideyipi, line 142: I deem it better to do another abstract upon the "IoTDBConnector", and make all the connectors need commiting extend it. Therefore we can avoid duplicateing the commit logic.

Caideyipi (commented):
PTAL

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/protocol/IoTDBConnector.java
DanielWang2035: how about connectors with same attribute string share one commit id?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/protocol/IoTDBConnector.java
DanielWang2035: i think this is good

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/protocol/IoTDBConnector.java
Caideyipi: This will not affect the correctness, for the global commit order ensures the local commit order of each pipe sharing the same connector.

> how about connectors with same attribute string share one commit id?

Caideyipi:
I finally discovered that when an event is collected in a connector, it's sequence is eliminated....Maybe the commitId must be assigned before any subtask picking this event, and the commitId may better be something like "static Map<connectorAttributes, commitId>" or "static Map<pendingQueue, commitId>"....

Caideyipi:
> I finally discovered that when an event is collected in a connector, it's sequence is eliminated....Maybe the commitId must be assigned before any subtask picking this event, and the commitId may better be something like "static Map<connectorAttributes, commitId>" or "static Map<pendingQueue, commitId>"....

It may to some extent be my fault... And ConcurrentHashMap is OK since we do not need to ensure the sequentiality across pipes.

DanielWang2035:
> I finally discovered that when an event is collected in a connector, it's sequence is eliminated....Maybe the commitId must be assigned before any subtask picking this event, and the commitId may better be something like "static Map<connectorAttributes, commitId>" or "static Map<pendingQueue, commitId>"....

Maybe we can assign the commitId atomically when the connectorSubtask polls an event. But some events are created inside transfer() when parsing pattern, do they need a commitId?

Caideyipi:
> > I finally discovered that when an event is collected in a connector, it's sequence is eliminated....Maybe the commitId must be assigned before any subtask picking this event, and the commitId may better be something like "static Map<connectorAttributes, commitId>" or "static Map<pendingQueue, commitId>"....
> 
> Maybe we can assign the commitId atomically when the connectorSubtask polls an event. But some events are created inside transfer() when parsing pattern, do they need a commitId?

Well, to extreme, after a connector subtask polls an event, it may just sleep until another subtask has commited the subsequent events... Besides, the events created inside transfer do not need commitId, though they(rawTablets) typically do nothing if you commit them, thus arbitrary order is allowed. In a nutshell, I suggest that we assign the Id before the events are put in to the pending queue, like in PipeEventCollector.

Caideyipi:
> > I finally discovered that when an event is collected in a connector, it's sequence is eliminated....Maybe the commitId must be assigned before any subtask picking this event, and the commitId may better be something like "static Map<connectorAttributes, commitId>" or "static Map<pendingQueue, commitId>"....
> 
> Maybe we can assign the commitId atomically when the connectorSubtask polls an event. But some events are created inside transfer() when parsing pattern, do they need a commitId?

I missed that, you can also use lock to make sure a connector assign an commitID right after poll... Then the sequence can be sure. It depents on your choice to realize it.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/commit/PipeEventCommitManager.java
```diff
@@ -0,0 +1,124 @@
...
+        || !eventCommitterMap.containsKey(committerKey)
+        || commitId <= EnrichedEvent.NO_COMMIT_ID) {
+      return;
+    }
+
+    eventCommitterMap.get(committerKey).commit(event);
+  }
+
+  private PipeEventCommitManager() {
+    committableConnectors.put(
```
Caideyipi, line 95: Maybe consider "ImmutableSet"?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/protocol/thrift/sync/IoTDBThriftSyncConnector.java
```diff
@@ -262,8 +262,9 @@ public void transfer(TabletInsertionEvent tabletInsertionEvent) throws Exception
 
     try {
       if (isTabletBatchModeEnabled) {
-        if (tabletBatchBuilder.onEvent(tabletInsertionEvent)) {
-          doTransfer(client);
+        if (tabletBatchBuilder.onEvent(
+            tabletInsertionEvent, ((EnrichedEvent) tabletInsertionEvent).getCommitId())) {
```
Caideyipi, line 266: If the commitId is built in the event itself, maybe it's adequate to pass only the event in.

Caideyipi (changes requested):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/commit/PipeEventCommitManager.java
DanielWang2035: fixed

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/commit/PipeEventCommitManager.java
```diff
@@ -0,0 +1,124 @@
...
+    if (pipeName == null || !COMMITTABLE_CONNECTORS.contains(pipePluginName)) {
+      return;
+    }
+
+    final String committerKey = generateCommitterKey(pipeName, dataRegionId);
+    if (eventCommitterMap.containsKey(committerKey)) {
+      LOGGER.warn("Pipe with same name is already registered, overwriting: {}", committerKey);
+    }
+    eventCommitterMap.put(committerKey, new PipeEventCommitter());
+    LOGGER.info("Pipe committer registered for pipe: {}", committerKey);
```
Caideyipi, line 48: For pipe on dataRegion?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/commit/PipeEventCommitManager.java
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  public void deregister(String pipeName, int dataRegionId) {
+    final String committerKey = generateCommitterKey(pipeName, dataRegionId);
+    eventCommitterMap.remove(committerKey);
+    LOGGER.info("Pipe committer deregistered for pipe: {}", committerKey);
```
Caideyipi, line 54: For pipe on dataRegion?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/payload/evolvable/builder/PipeTransferBatchReqBuilder.java
```diff
@@ -98,6 +100,49 @@ protected PipeTransferBatchReqBuilder(PipeParameters parameters) {
...
+    }
+
+    final TPipeTransferReq req = buildTabletInsertionReq(event);
+    final long requestCommitId = ((EnrichedEvent) event).getCommitId();
+
+    if (requestCommitIds.isEmpty()
+        || !requestCommitIds.get(requestCommitIds.size() - 1).equals(requestCommitId)) {
+      reqs.add(req);
+      events.add(event);
+      requestCommitIds.add(requestCommitId);
```
Caideyipi, line 121: Maybe eliminate this "requestCommitId" too?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/protocol/thrift/async/IoTDBThriftAsyncConnector.java
```diff
@@ -176,13 +171,15 @@ public void transfer(TabletInsertionEvent tabletInsertionEvent) throws Exception
...
+
     if (isTabletBatchModeEnabled) {
-      final long requestCommitId = commitIdGenerator.incrementAndGet();
-      if (tabletBatchBuilder.onEvent(tabletInsertionEvent, requestCommitId)) {
+      if (tabletBatchBuilder.onEvent(tabletInsertionEvent)) {
         final PipeTransferTabletBatchEventHandler pipeTransferTabletBatchEventHandler =
             new PipeTransferTabletBatchEventHandler(tabletBatchBuilder, this);
 
-        transfer(requestCommitId, pipeTransferTabletBatchEventHandler);
+        transfer(commitId, pipeTransferTabletBatchEventHandler);
```
Caideyipi, line 181: Can all those "commitId"s packed into the its event?

SteveYurongSu (approved):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/protocol/thrift/async/IoTDBThriftAsyncConnector.java
DanielWang2035: I think this may be inconvenient

Caideyipi: This may have a point..

SteveYurongSu (approved):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/commit/PipeEventCommitManager.java
DanielWang2035: fixed

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/commit/PipeEventCommitManager.java
DanielWang2035: fixed

SteveYurongSu (approved):
MPGA!

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/task/subtask/connector/PipeConnectorSubtaskLifeCycle.java
```diff
@@ -66,14 +66,23 @@ public synchronized void register() {
...
    * @return true if the subtask is out of life cycle, indicating that the subtask should never be
    *     used again
    * @throws IllegalStateException if aliveTaskCount <= 0
    */
-  public synchronized boolean deregister() {
+  public synchronized boolean deregister(String pipeNameToDeregister) {
+    // The stop() here keeps runningTaskCount consistent, otherwise there might be runningTaskCount
+    // > aliveTaskCount.
+    // Remember not to stop twice.
+    stop();
```
SteveYurongSu, line 78: Think twice here

SteveYurongSu (changes requested):

SteveYurongSu (approved):

PR #11491, opened 2023-11-07
Fix aggregation error when inserting negative timestamp

Beyyes:
## Description

Before fix, the result of select count(*) from root.** is always 0, now it will be the right value.

See details in TIMECHODB-434.


HTHou (approved):

wangchao316 (approved):
LGTM

PR #11502, opened 2023-11-08
Optimize batch fetch schema for data insertion

Cpaulyz:
## Description

Optimize batch fetch schema for data insertion

image


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/path/PathPatternTree.java
```diff
@@ -296,16 +309,19 @@ public boolean isOverlapWith(PathPatternTree patternTree) {
   public void serialize(PublicBAOS outputStream) throws IOException {
     constructTree();
     root.serialize(outputStream);
+    ReadWriteIOUtils.write(containWildcard, outputStream);
```
MarcosZyk, line 312: Some module may persist patternTree on disk, thus this modification may result in compatibility problems. Since this field is only used for runtime, make this field initialized or generated during pattrenTree construction.

MarcosZyk (commented):
Excellent work! Only one problem left. PTAL~

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/path/PathPatternTree.java
Cpaulyz: > Some module may persist patternTree on disk, thus this modification may result in compatibility problems. Since this field is only used for runtime, make this field initialized or generated during pattrenTree construction.

Fix. PTAL

MarcosZyk (approved):

PR #11504, opened 2023-11-09
Pipe IT: adjust -DintegrationTest.forkCount & -DDataNodeMaxHeapSize in pipe-it-2cluster.yml to provide more stable env

SteveYurongSu:
As title.


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/conf/CommonConfig.java
```diff
@@ -185,7 +185,7 @@ public class CommonConfig {
 
   private int pipeMaxAllowedPendingTsFileEpochPerDataRegion = 2;
 
-  private boolean pipeMemoryManagementEnabled = true;
+  private boolean pipeMemoryManagementEnabled = false;
```
SteveYurongSu, line 188:
```suggestion
  private boolean pipeMemoryManagementEnabled = true;
```

PR #11509, opened 2023-11-09
Pipe: skip pipe CN heartbeat and sync when `PipeTaskCoordinatorLock` is held by another thread

VGalaxies:
## Description

When the DN processing logic is blocked during the create (start) pipe, due to the timeout and retry mechanism, the write locks of CN's `PipeTaskCoordinatorLock` and DN's `PipeMetaKeeper` will be held for a long period of time, causing the timed `heartbeat` logic and `sync` logic to make no progress at all during this period, which in turn leads to a large number of related procedures and RPCs piling up. This PR partially sacrifices the liveness of `heartbeat` and `sync` to avoid the above situation.

---

Minimum Reproduction:

```sql
INSERT INTO root.db.d(time, m) values (1, 1);

create pipe a2b
with connector (
    'connector'='iotdb-air-gap-connector',
    'connector.ip'='192.0.2.0',
    'connector.port'='12345',
    'connector.air-gap.handshake-timeout-ms'='400000'
);
```

Here, the IP is a reserved address, so the socket connect process in `iotdb-air-gap-connector` will continue to be blocked until timeout. 

The client can observe the error message `Msg: 1107: Procedure execution timed out.`. Meanwhile, CN will be unable to process additional pipe requests due to the accumulation of procedures and RPCs.

---

Detailed Explanation for added skip logic:

For the `heartbeat` operation, if it is found that `PipeTaskCoordinatorLock` is held by another thread before execution, according to the calling relationship, there are no more than the following situations:

- create / start / stop / drop pipe, these procedures will call the `pushSinglePipeMeta` rpc of DN in the `OPERATE_ON_DATA_NODES` step (`org.apache.iotdb.confignode.procedure.env.ConfigNodeProcedureEnv#pushSinglePipeMetaToDataNodes/dropSinglePipeOnDataNodes`), these two steps will acquire the write lock of `PipeMetaKeeper` (DN) on the DN side. Notice that `heartbeat` will also call the `pipeHeartbeat` rpc of DN, and the read lock of `PipeMetaKeeper` will be acquired on the DN side. If the previous process is blocked, the latter process will inevitably be blocked
- parse heartbeat, indicating that the previous round of `heartbeat` has not been processed (impossible, guaranteed by `synchronized`?)
- auto restart / handle successful restart, indicating that `sync` is being processed at this time, these two steps will acquire the write lock of `PipeMetaKeeper` (CN) (No effect)
- `PipeMetaSyncProcedure` / `PipeHandleMetaChangeProcedure` / `PipeHandleLeaderChangeProcedure`, these procedures will call the `pushPipeMeta` rpc of DN, analysis of the same as above

In this way, when performing the `heartbeat` operation, if it is found that `PipeTaskCoordinatorLock` is held by another thread, if the other thread is blocked, the `heartbeat` will **most likely** be blocked, so just skip.

The same analysis can be performed for `sync` operation.

<hr>

This PR has:
- [x] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


SteveYurongSu (approved):
Make pipe GREAT again!!!

PR #11510, opened 2023-11-10
Check and update searchIndex when restarting dataNode if necessary

OneSizeFitsQuorum:
Description：
- The searchIndex for some regions may be smaller than the safelyDeletedSearchIndex when the node is restarted, which means that the data of the current node is not flushed by the wal, but has been synchronized to other nodes. At this point, different replicas have been inconsistent and cannot be automatically recovered. 

Solution：
- To prevent subsequent logs from marking smaller searchIndex and exacerbating the inconsistency, we manually set the searchIndex to safelyDeletedSearchIndex here to reduce the impact of this problem in the future


wangchao316 (approved):
LGTM, GOOD

BUAAserein (approved):
LGTM

PR #11516, opened 2023-11-12
Feat/config option os metrics enabled

chrisdutz:
Currently there is no option to disable the metrics collection on config-nodes.
This causes problems on CtlrX core as the app tries to access resources outside of it's sandbox ("/sys/block/")

This new config option allows programatically disabling the DiskMetrics and NetMetrics.


OneSizeFitsQuorum:
Hi，chris ,we have added a off level to close system metric, you could take a look for this pr :https://github.com/apache/iotdb/pull/11530

chrisdutz:
Will have a look after my trip, next week. Thanks

PR #11526, opened 2023-11-13
Remove useless variable outputDeviceToQueriedDevicesMap in Analysis

Beyyes:


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/analyze/AnalyzeVisitor.java
```diff
@@ -1789,10 +1785,7 @@ private void analyzeDataPartition(
       Analysis analysis, QueryStatement queryStatement, ISchemaTree schemaTree) {
     Set<String> deviceSet = new HashSet<>();
     if (queryStatement.isAlignByDevice()) {
-      deviceSet =
```
Beyyes, line 1792: optimize it, it no LogicalView, just use analysis.getDeviceList().

PR #11532, opened 2023-11-13
Pipe: skip pipe CN meta sync after successful synchronization when having no pipe

VGalaxies:
## Description

In order to fix the issue that `sync` operation logs are still printed when no pipe exists in CN, like:

```text
2023-11-13 21:01:36,412 [ProcExecWorker-1] INFO  o.a.i.c.p.ProcedureExecutor:394 - pid=1, state=SUCCESS; org.apache.iotdb.confignode.procedure.impl.pipe.runtime.PipeMetaSyncProcedure finished in 55ms successfully.
```

a simple idea is to, in the next round of `sync` after a successful `sync` operation, check whether the current version of `PipeTaskInfo` is the same as the version synced before (empty `PipeTaskInfo`), and whether the current `PipeTaskInfo` is empty (indicating no need for a restart).

In terms of specific implementation:

1. Maintain the version of `PipeTaskInfo` synced by CN in the last round. In this context, the version is considered as the number of times the `PipeTaskInfo` write lock is acquired.
    - If a restart occurs, it is the version after `PipeHandleMetaChangeProcedure`; otherwise, it is the version after `PipeMetaSyncProcedure`.
    - Update in the lock release phase of these procedures to ensure that the recorded version at this point is free from concurrency issues.
    - Considering that `PipeHandleMetaChangeProcedure` may also be submitted by other logic, in order to avoid introducing new variables that may affect the serialization and deserialization logic in `PipeHandleMetaChangeProcedure`, we only update the version of `PipeTaskInfo` synced by `PipeMetaSyncProcedure` (partially affecting the liveness of **skipping sync** but not affecting correctness).
2. Maintain whether the last sync was successful in `PipeMetaSyncer`.
    - To check if the `sync` operation succeeded, it is also necessary to block and wait for the execution result of `PipeHandleMetaChangeProcedure` (refer to #11475).

Since querying `PipeTaskInfo` before each round of `sync` operations is done without locking, there may be other procedures modifying `PipeTaskInfo` at this time. After verification, it is confirmed that this only partially affects the liveness of **sync** but does not affect correctness.

<hr>

This PR has:
- [x] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/pipe/runtime/PipeMetaSyncer.java
```diff
@@ -104,16 +117,39 @@ private synchronized void sync() {
       pipeAutoRestartRoundCounter.set(0);
     }
 
-    final TSStatus status = procedureManager.pipeMetaSync();
+    final TSStatus metaSyncStatus = procedureManager.pipeMetaSync();
 
     if (somePipesNeedRestarting
```
Caideyipi, line 122: The "SUCCESS_STATUS" may be compacted to a single "if", and the "!" may be an "else".

Caideyipi: Besides, if there are "stopped" pipe, these pipe may be automatically restarted. Yet, there will be no pipe to restart given that there are no pipes in the cluster.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/pipe/runtime/PipeMetaSyncer.java
```diff
@@ -104,16 +117,39 @@ private synchronized void sync() {
...
+        && metaSyncStatus.getCode() == TSStatusCode.SUCCESS_STATUS.getStatusCode()) {
+      final boolean isRestartSuccessful = handleSuccessfulRestartWithLock();
+      final TSStatus handleMetaChangeStatus =
+          procedureManager.pipeHandleMetaChangeWithBlock(true, false);
+      if (!configManager.getPipeManager().getPipeTaskCoordinator().hasAnyPipe()
+          && isRestartSuccessful
+          && handleMetaChangeStatus.getCode() == TSStatusCode.SUCCESS_STATUS.getStatusCode()) {
+        LOGGER.info(
+            "set isLastEmptyPipeSyncSuccessful to true, skip next round of sync if no pipe");
+        isLastEmptyPipeSyncSuccessful = true;
```
Caideyipi, line 132: The user behaviour may cause "ABA" problem, such as "empty sync, create, start, drop(partially succeed on some DN)", and the subsequent syncs may be skipped, then the failed DN may never know the drop command and will keep sending data forever.
I think a meta sync cannot be skipped if there are user manual commands after last empty sync.

Caideyipi (changes requested):
PTAL~

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/pipe/runtime/PipeMetaSyncer.java
VGalaxies: > Besides, if there are "stopped" pipe, these pipe may be automatically restarted. Yet, there will be no pipe to restart given that there are no pipes in the cluster.

When there are "stopped" pipes in the cluster, should the result of `PipeTaskCoordinator.hasAnyPipe()` be true?

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/pipe/runtime/PipeMetaSyncer.java
VGalaxies: Good catch, I added an extra variable to determine if `PipeTaskInfo` has been modified between two empty sync.

VGalaxies:
The PR description needs to be updated later...

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/pipe/runtime/PipeMetaSyncer.java
```diff
@@ -87,9 +94,21 @@ public synchronized void start() {
   }
 
   private synchronized void sync() {
+    if (!configManager.getPipeManager().getPipeTaskCoordinator().hasAnyPipe()
+        && isLastEmptyPipeSyncSuccessful
+        && lastEmptyPipeSyncWriteLockAcquiredCount
```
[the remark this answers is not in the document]
Caideyipi: And the pipe heartbeat won't produce "pipeTaskInfoLockCount" if it has nothing to do with the update of the CN pipe task metas(*^_^*)

Caideyipi (changes requested):
PTAL2~

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/pipe/runtime/PipeMetaSyncer.java
Caideyipi: Actually if the "successfullyRestart" is true, at least one pipe has a high probablity to exist... Namely, an empty pipe task won't let the "sync" method to reach "Successfully restart". Also, deleting the judgment won't cost much, at most the extra sync.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/pipe/PipeTaskInfo.java
```diff
@@ -80,12 +84,51 @@ private void releaseReadLock() {
...
+    public void increaseLatestVersion() {
+      latestVersion.incrementAndGet();
+    }
+
+    public void updateLastSyncedVersion() {
+      lastSyncedVersion = latestVersion.get();
+      isLastSyncedPipeTaskInfoEmpty = pipeMetaKeeper.isEmpty();
+    }
+
+    public boolean isEmptySynced() {
```
Caideyipi, line 117: isEmptySynced infers not only the very last empty sync, but the empty pipeMeta now. Maybe simply call it "canSkipNextSync"?

Caideyipi (approved):
LGTM

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/pipe/PipeTaskInfo.java
```diff
@@ -80,12 +84,51 @@ private void releaseReadLock() {
...
+    pipeTaskInfoVersion.increaseLatestVersion();
   }
 
   private void releaseWriteLock() {
     pipeMetaKeeper.releaseWriteLock();
   }
 
+  /////////////////////////////// Version ///////////////////////////////
+
+  private class PipeTaskInfoVersion {
```
SteveYurongSu, line 96: It is recommended to remove this class from PipeTaskInfo, as Info records content that needs to be persisted to the consensus layer. PipeTaskInfoVersion can be placed in the PipeTaskCoordinator class (or its package)

SteveYurongSu (approved):
LGTM

PR #11534, opened 2023-11-14
Support high available while meeting timeout exception

JackieTien97:
1. If TimeoutException happened while sending write FI, we need to retry sending this FI to another follower.
img_v3_0256_2972aaeb-f965-4b30-b33c-921c0970a45g
b676fdc9-7f45-444d-b0f8-9a7199632833

2. We need to stop rpc service firstly in shut down hook, because we don't want to receive more client requests which will also be refused in the later process.


OneSizeFitsQuorum (approved):
LGTM

PR #11543, opened 2023-11-15
Return WRITE_PROCESS_ERROR when catching BatchProcessException in writing process

xingtanzjr:
## Description
Before this change, some write operations will return `QUERY_PROCESS_ERROR` when catching BatchProcessException. The status seems wired for write operations. And especially in IoTConsensus, it won't trigger operation retry when the response code is `QUERY_PROCESS_ERROR`.

Thus, to fix this issue, we will return WRITE_PROCESS_ERROR if write operation catches BatchProcessException


HTHou (approved):
LGTM

PR #11545, opened 2023-11-15
ISchemaTree supports template interface

Cpaulyz:
## Description

ISchemaTree supports template interface:
```java
void appendTemplateDevice(PartialPath devicePath, boolean isAligned, int templateId);

/**
 * Check whether this schema tree has normal series(not template series).
 * @return true if it has normal series, else false
 */
boolean hasNormalTimeSeries();

/**
 * Get all templates being used in this schema tree.
 * @return template list
 */
List<Template>getUsingTemplates();
```

## Docs
https://apache-iotdb.feishu.cn/docx/QDJ4dAf7eow3BSxN580cu8PPnlh

image


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/common/schematree/ClusterSchemaTree.java
```diff
@@ -44,23 +46,31 @@
...
 
 import static org.apache.iotdb.commons.conf.IoTDBConstant.PATH_ROOT;
 import static org.apache.iotdb.commons.schema.SchemaConstant.ALL_MATCH_PATTERN;
+import static org.apache.iotdb.commons.schema.SchemaConstant.NON_TEMPLATE;
 import static org.apache.iotdb.db.queryengine.common.schematree.node.SchemaNode.SCHEMA_ENTITY_NODE;
 import static org.apache.iotdb.db.queryengine.common.schematree.node.SchemaNode.SCHEMA_MEASUREMENT_NODE;
 
 public class ClusterSchemaTree implements ISchemaTree {
+  private static final ClusterTemplateManager templateManager =
+      ClusterTemplateManager.getInstance();
```
MarcosZyk, line 62: Pass this into the ```deserialize``` method as a local param rather than a global static attribute.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/common/schematree/ISchemaTree.java
```diff
@@ -71,6 +70,20 @@ Pair<List<MeasurementPath>, Integer> searchMeasurementPaths(
 
   void mergeSchemaTree(ISchemaTree schemaTree);
 
+  /**
+   * Check whether this schema tree has normal series(not template series).
+   *
+   * @return true if it has normal series, else false
+   */
+  boolean hasNormalTimeSeries();
```
MarcosZyk, line 78: Seems TestOnly?

MarcosZyk (commented):
Excellent work! Only a few problem left. PTAL~

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/common/schematree/ISchemaTree.java
Cpaulyz: > Seems TestOnly?

Query optimization will use this interface.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/common/schematree/ClusterSchemaTree.java
Cpaulyz: > Pass this into the `deserialize` method as a local param rather than a global static attribute.

seems that static attribute can save some memory usgae?

MarcosZyk: The build in static attribute may not friendly to unit test. Anyway, related modification can be delayed to when it is needed.

MarcosZyk (approved):

PR #11546, opened 2023-11-15
Refactor CachedMTreeStore.updateMNode interface usage and implementation in PBTree

MarcosZyk:
### Motivation
In the original implementation, the attribute modification is not protected by the read lock, therefore a node already in buffer may experience redundant flush while the updated info may already be synced to disk.

Here's a simple example. 

MNode A is already in buffer since it may be newly added to MTree or recently updated. 

Thread A tries to modify some attribute in it while the flush thread starts a flush process and take the write lock in advance. 

In this case, the modified info will be synced to disk by this flush process, but thread A may still invoke the updateMNode method to add the node to buffer after this flush. 

The result is a node with nothing new will experience another flush process.

### Modification

Refactor the updateMNode method to require the invokers to pass the specific update process as a consumer, which will be protected by read lock.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/IMTreeStore.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/IMTreeStore.java
@@ -28,6 +28,7 @@ import org.apache.iotdb.db.schemaengine.template.Template;
 
 import java.io.File;
 import java.util.Map;
+import java.util.function.Consumer;
 
 /**
  * This interface defines the basic access methods of an MTreeStore.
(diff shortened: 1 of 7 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/IMTreeStore.java
```diff
@@ -73,7 +74,7 @@ IMNodeIterator<N> getTraverserIterator(
 
   void deleteChild(N parent, String childName) throws MetadataException;
 
-  void updateMNode(N node) throws MetadataException;
+  void updateMNode(N node, Consumer<N> consumer) throws MetadataException;
```
Cpaulyz, line 77:
```suggestion
  /**
   * Update the mnode under the guarantee of latch.
   *
   * @param node node to be updated
   * @updation update operation
   */
  void updateMNode(N node, Consumer<N> updation) throws MetadataException;
```

Cpaulyz (approved):
LGTM

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/IMTreeStore.java
MarcosZyk: Java doc added.

PR #11548, opened 2023-11-15
Increase retry timeout for ratis retry mechanism

SzyWilliam:
Previous we only retry 3 times, waiting 500ms each.
We should align this parameter with connection timeout.


OneSizeFitsQuorum (approved):
LGTM

PR #11553, opened 2023-11-16
Add last time check for inner space compaction task

shuwenwei:
## Description
Add last time check for inner space compaction task. 
When the source file has overlap data, compaction task will throw a CompactionLastTimeCheckFailedException.


xingtanzjr (approved):

caozj1011 (approved):

choubenson (approved):
LGTM

PR #11561, opened 2023-11-16
Using the cumulative execution time of all FIs of a Query as the basis for calculating the level in MultiLevelPriorityQueue of the related DriverTasks

lancelly:
If a query has many FIs, and each FI is divided into a large number of DriverTasks, using the cumulative execution time of FIs as the basis for calculating the level of DriverTasks might result in the current query being allocated more CPU resources than expected.
The insertion and removal of elements in queryIdToDriverTaskHandleMap are consistent with queryMap, so memory leak issues are not expected.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/schedule/DriverScheduler.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/schedule/DriverScheduler.java
@@ -65,7 +65,6 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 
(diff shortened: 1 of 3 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/schedule/DriverScheduler.java
```diff
@@ -105,6 +108,7 @@ private DriverScheduler() {
         new L1PriorityQueue<>(
             QUERY_MAX_CAPACITY, new DriverTask.TimeoutComparator(), new DriverTask());
     this.queryMap = new ConcurrentHashMap<>();
+    this.queryIdToDriverTaskHandleMap = new ConcurrentHashMap<>();
```
SzyWilliam, line 111: Why not initializing when declaring?

PR #11566, opened 2023-11-17
Optimize fetch template schema

Cpaulyz:
## Description

Currently, if a template is activated in the SchemaRegion, all the mnode sequences are traversed when fetchSchema. For example, 100,000 devices with 30 measurements per device will traverse 3,000,000 mnode.

This PR is optimised so that for devices that use templates. Interface `skipTemplateChildren(IDeviceMNode device)` is provided by traverser framework. In the example above, only 100,000 mnode need to be traversed.

## Test
```java
@Test
public void testFetchPerformance() throws Exception {
  ISchemaRegion schemaRegion = getSchemaRegion("root.sg", 0);
  List<String> measurements = new ArrayList<>();
  List<TSEncoding> encodings = new ArrayList<>();
  List<CompressionType> compressors = new ArrayList<>();
  for (int i = 0; i < 30; i++) {
    measurements.add("s" + i);
    dataTypes.add(TSDataType.INT64);
    encodings.add(TSEncoding.PLAIN);
    compressors.add(CompressionType.SNAPPY);
  }
  Template template = new Template("t1", measurements, dataTypes, encodings, compressors);
  template.setId(1);

  for (int i = 0; i < 100000; i++) {
    schemaRegion.activateSchemaTemplate(
        SchemaRegionWritePlanFactory.getActivateTemplateInClusterPlan(
            new PartialPath("root.sg.d" + i), 1, template.getId()),
        template);
  }
  System.out.println("activateSchemaTemplate cost: " + (System.currentTimeMillis() - startTime));
  startTime = System.currentTimeMillis();
  ISchemaTree schemaTree =
      schemaRegion.fetchSchema(ALL_MATCH_SCOPE, Collections.singletonMap(1, template), false);
  System.out.println("fetchSchema cost: " + (System.currentTimeMillis() - startTime));
}
```

Before:

activateSchemaTemplate cost: 382
fetchSchema cost: 520

After

activateSchemaTemplate cost: 382
fetchSchema cost: 122


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/traverser/Traverser.java
```diff
@@ -73,6 +74,9 @@ public abstract class Traverser<R, N extends IMNode<N>> extends AbstractTreeVisi
   // default false means fullPath pattern match
   protected boolean isPrefixMatch = false;
 
+  private IMNodeIterator<N> currentChildrenIterator;
```
MarcosZyk, line 77: Should this attribute be updated during traverse process? What about getting the current iterator via peeking the IteratorStack in AbstractTreeVisitor?

MarcosZyk (approved):

PR #11569, opened 2023-11-17
Remove wrong TimeSliceAllocator

JackieTien97:
We never need to split the time slice for each operator in one DriverTask because we're using pull-based execution model, even if one child consume all the time slice of one dirver, its parent should detect that and then stop current run immediately  instead of keeping runing next child operator.


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/OperatorContext.java
```diff
@@ -90,8 +96,8 @@ public Duration getMaxRunTime() {
     return maxRunTime;
   }
 
-  public void setMaxRunTime(Duration maxRunTime) {
-    this.maxRunTime = maxRunTime;
+  public static void setMaxRunTime(Duration maxRunTime) {
+    OperatorContext.maxRunTime = maxRunTime;
```
lancelly, line 100: why not remove this method since its nearly only used in tests and ClientRpcServiceImpl? I think there's no need to call this method both in tests and ClientRpcServiceImpl, too.
截屏2023-11-18 10 37 28

JackieTien97: Actually, I wanna keep this method for future usage like we want to change the time slice for running DN without restarting it.

lancelly: If I'm not mistaken, changing the time slice length without restarting should be done through calling a Session interface or executing a SQL . However, setting OperatorMaxRuntime doesn't seem to work, and it seems that what should be changed is the value in the Config, right?

JackieTien97: Yes, but in current situation, each operator's max run time is got from OperatorContext. And calling a Session interface or executing a SQL is to change the value in OperatorContext.

lancelly: I see.

lancelly (approved):
LGTM

PR #11573, opened 2023-11-19
Use `Expression` to represent predicate in FE & Refactor `tsfile/read/filter` module

liuminghui233:
In this PR, I
- Use `Expression` to represent value filter in `PlanNode`. (Field name rename `valueFilter` to `pushDownPredicate`)
- Set `globalTimeFilter` as a query-level property rather than a field of the `PlanNode`.
   - Remove `timeFilter` field from all scan node.
   - Save the analyzed `globalTimePredicate` (`Expression` form) in `Analysis`.
   - Save a `globalTimeFilter` (`Filter` form) in `MPPContext` for filtering the partition table on the FE side.
   - Save a `globalTimePredicate` (`Expression` form) in each FI, convert it to `globalTimeFilter` (`Filter` form) on BE and save it in `FragmentInstanceContext`, which is used to (1) initialize `QueryDataSource`; (2) initialize `ScanOperator`.
- Sorted out the code of the Filter module:
   - Inherits `Filter` and implements more interfaces, including `ITimeFilter`, `IValueFilter`, `IDisableStatisticsTimeFilter`, `IDisableStatisticsValueFilter`, and `IStatefulFilter`. When implementing an operator, choose the interface to implement based on the nature of the operator.
   - Added base classes for various operators: `ColumnCompareFilter`, `ColumnRangeFilter`, `ColumnSetFilter`, `ColumnPatternMatchFilter`, `BinaryLogicalFilter`.
   - Separated implementation of `TimeFilter` and `ValueFilter` (removed `FilterType`).
   - The factory class is clarified. All external uses of `Filter` are created through the factory class.
   - Removed the implementation of `Like` and converted it directly to `Regex`.


iotdb-core/tsfile/src/main/java/org/apache/iotdb/tsfile/utils/RegexUtils.java
```diff
@@ -0,0 +1,92 @@
...
+
+  private RegexUtils() {
+    // util class
+  }
+
+  /**
+   * The main idea of this part comes from
+   * https://codereview.stackexchange.com/questions/36861/convert-sql-like-to-regex/36864
+   */
+  public static String parseLikePatternToRegex(String likePattern) {
```
JackieTien97, line 35: What's the difference between this and previous implementation, and why you want to do this change. I found that someone said that this method is still incorrect in this article's comments.https://codereview.stackexchange.com/questions/36861/convert-sql-like-to-regex/36864

JackieTien97 (changes requested):

PR #11602, opened 2023-11-23
Improve BinaryStatistics deserialization

JackieTien97:
the first value and last value initialization process in BinaryStatistics is time consuming, and there is no need for us to do it.
image


wangchao316 (approved):
LGTM

PR #11603, opened 2023-11-23
Rename THeartbeat to TDataNodeHeart

liyuheng55555:
Now we have TConfigNodeHeartbeat, so old THeartbeat's name also needs changing.


OneSizeFitsQuorum (approved):
LGTM

PR #11604, opened 2023-11-23
Fix Some Code Smells of Node Commons Module

cuibo01:
Fix Some Code Smells of Node Commons Module


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/auth/authorizer/BasicAuthorizer.java
```diff
@@ -338,7 +338,7 @@ public Map<String, Boolean> getAllUserWaterMarkStatus() {
       try {
         userWaterMarkStatus.put(user, isUserUseWaterMark(user));
       } catch (AuthException e) {
-        LOGGER.error(String.format(NO_SUCH_USER_EXCEPTION, user));
+        LOGGER.error("No such user: {}", user, e);
```
wangchao316, line 341: authexception does not print, have Security question

wangchao316 (approved):
LGTM

PR #11609, opened 2023-11-23
[IOTDB-6255] Scatter width metric

CRZbulabula:
https://issues.apache.org/jira/browse/IOTDB-6255

Also updated the Apache IoTDB ConfigNode Dashboard:

overview

info

### DIFF
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/node/NodeInfo.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/node/NodeInfo.java
@@ -224,6 +224,24 @@ public class NodeInfo implements SnapshotProcessor {
     return result;
   }
 
+  /** Return the number of registered Nodes. */
+  public int getRegisteredNodeCount() {
+    int result;
+    dataNodeInfoReadWriteLock.readLock().lock();
+    try {
+      result = registeredDataNodes.size();
+    } finally {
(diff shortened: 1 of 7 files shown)


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/node/NodeInfo.java
```diff
@@ -224,6 +224,24 @@ public DataNodeConfigurationResp getDataNodeConfiguration(
...
+    int result;
+    dataNodeInfoReadWriteLock.readLock().lock();
+    try {
+      result = registeredDataNodes.size();
+    } finally {
+      dataNodeInfoReadWriteLock.readLock().unlock();
+    }
+    configNodeInfoReadWriteLock.readLock().lock();
+    try {
+      result += registeredConfigNodes.size();
```
OneSizeFitsQuorum, line 238: do we need to add confignode size?

CRZbulabula: Yes, this makes the statistics of scatter width easier. Otherwise, we should take another map step when processing DataNodeId.

OneSizeFitsQuorum: got it~

OneSizeFitsQuorum (approved):
LGTM

PR #11615, opened 2023-11-27
[RatisConsensus] Cache group id in the write/read critical path

SzyWilliam:
We need to compute the consensus group id to raft group id mapping every time. This is cpu-consuming and we can avoid them simply by adding a weak-value cache.

### DIFF
--- a/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/utils/Utils.java
+++ b/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/utils/Utils.java
@@ -38,6 +38,8 @@ import org.apache.ratis.protocol.RaftPeer;
 import org.apache.ratis.protocol.RaftPeerId;
 import org.apache.ratis.server.RaftServerConfigKeys;
 import org.apache.ratis.server.protocol.TermIndex;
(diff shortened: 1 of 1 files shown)


iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/utils/Utils.java
```diff
@@ -57,6 +60,8 @@ public class Utils {
   private static final String DATA_REGION_GROUP = "group-0001";
   private static final String SCHEMA_REGION_GROUP = "group-0002";
   private static final CommonConfig config = CommonDescriptor.getInstance().getConfig();
+  private static final Cache<ConsensusGroupId, RaftGroupId> cache =
+      CacheBuilder.newBuilder().weakValues().build();
```
OneSizeFitsQuorum, line 64: Add the default expiration time of 5min? This will also work in the future with region migrations

SzyWilliam: Captain, oh my captain!

OneSizeFitsQuorum (approved):
LGTM~
Thanks William for this detailed optimization

PR #11622, opened 2023-11-27
Fix concurrent bug in PartitonCache

JackieTien97:
more details can be seen in https://jira.infra.timecho.com:8443/browse/TIMECHODB-447

we can find that if the cache is cleared between these two line, we will try to create an already existing database.

img_v3_0258_cc4d98db-8808-4bd1-bc97-2828b309e3eg


OneSizeFitsQuorum (approved):
LGTM

PR #11628, opened 2023-11-28
Adjust series_partition_slot_num to series_slot_num in the response message

OneSizeFitsQuorum:
The series_partition_slot_num parameter has been replaced by series_slot_num, and the error message needs to be updated


CritasWang (approved):
LGTM

PR #11240, opened 2023-10-07
Pipe: introduce global cache for TsFile metadatas and support TTL of Pipe TsFile

DanielWang2035:
Currently, if a TsFile is processed by N pipes, its metadata will be constructed N times, the overhead will be high when N is very big. This PR added a global cache to reduce repeated creation and redundant IO.

This PR also introduces TTL of TsFiles pinned by pipe engine, just like WALs.

PR #11637, opened 2023-11-29
[IOTDB-6260] Pipe: Wait for the hardlink/copied tsfile to finish being transfered before deleting

MiniSho:
## Description
When dropping the pipe, if there is a tsfile being reading or transfering, it will throw an error by calling Pipe Plugin‘s `close()` method and the file will not be unpinned.

so we add a check before deleting the tsfile, wait until the hardlink/copied file has been transfered.

image


SteveYurongSu:
Thanks for your contribution.

NO NEED TO FIX: the issue is already fixed by PR https://github.com/apache/iotdb/pull/11240

PR #11640, opened 2023-11-29
delete data and files together.

shuolinl:


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/DataRegion.java
```diff
@@ -2067,6 +2112,141 @@ private void deleteDataInFiles(
...
+    // can be deleted by files
+    for (TsFileResource tsFileResource : deletedByFiles) {
+      if (!tsFileResource.isClosed()) {
+        tsFileResource
+            .getProcessor()
+            .deleteDataInMemory(deletion, new HashSet<>(pathToDelete.getDevicePathPattern()));
+        tsFileResource.close();
+      }
+
+      tsFileResource.setStatus(TsFileResourceStatus.DELETED);
```
JackieTien97, line 2196: You shouldn't ignore the return result of setStatus, because we cannot directly delete tsfile while it's being compacted.

JackieTien97 (changes requested):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/DataRegion.java
```diff
@@ -2067,6 +2112,141 @@ private void deleteDataInFiles(
...
+      }
+    }
+
+    // can be deleted by files
+    for (TsFileResource tsFileResource : deletedByFiles) {
+      if (!tsFileResource.isClosed()) {
+        tsFileResource
+            .getProcessor()
+            .deleteDataInMemory(deletion, new HashSet<>(pathToDelete.getDevicePathPattern()));
+        tsFileResource.close();
```
JackieTien97, line 2193:
```suggestion
```
why close the resource here?

JackieTien97 (changes requested):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/DataRegion.java
shuolinl: Make sure that other thread does not modify this file anymore.

PR #11641, opened 2023-11-29
Pipe: avoid useless pipe meta sync (stopped status) to DN

VGalaxies:
## Description

After #11279, when a pipe encounters a critical exception on DN, it stops all DR's pipe tasks related to that pipe and modifies the status of the relevant pipe to exception stopped on this DN. Subsequently, the CN collects this information through the heartbeat mechanism and pushes the changes to all DNs. This causes other DNs with running normal pipe tasks to be auto-restarted unnecessarily. This PR addresses this issue.

From now on, there will be a subtle change in the semantics of the CN-side pipe status. When the status is exception stopped, it indicates that the pipe tasks on some (not necessarily all) DRs (DNs) of this pipe encountered a critical exception.

<hr>

This PR has:
- [x] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/pipe/runtime/PipeHeartbeatParser.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/pipe/runtime/PipeHeartbeatParser.java
@@ -224,7 +224,7 @@ public class PipeHeartbeatParser {
               runtimeMeta.setIsStoppedByRuntimeException(true);
 
               needWriteConsensusOnConfigNodes.set(true);
-              needPushPipeMetaToDataNodes.set(true);
+              needPushPipeMetaToDataNodes.set(false);
 
               LOGGER.warn(
                   "Detect PipeRuntimeCriticalException {} from DataNode, stop pipe {}.",
@@ -253,7 +253,7 @@ public class PipeHeartbeatParser {
                             runtimeMeta.setIsStoppedByRuntimeException(true);
 
                             needWriteConsensusOnConfigNodes.set(true);
-                            needPushPipeMetaToDataNodes.set(true);
+                            needPushPipeMetaToDataNodes.set(false);
 
                             LOGGER.warn(
                                 String.format(


VGalaxies:
Another question: When a pipe task encounters an exception, why is it necessary to rely on `lastEvent` instead of directly calling the `org.apache.iotdb.db.pipe.agent.runtime.PipeRuntimeAgent#report` method? See https://github.com/apache/iotdb/blob/master/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/task/subtask/PipeSubtask.java#L182

VGalaxies:
The `needPushPipeMetaToDataNodes` parameter of `PipeHandleMetaChangeProcedure` is currently always set to false.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/pipe/task/PipeTaskCoordinator.java
```diff
@@ -118,11 +120,18 @@ public TSStatus startPipe(String pipeName) {
   public TSStatus stopPipe(String pipeName) {
     final boolean isStoppedByRuntimeException = pipeTaskInfo.isStoppedByRuntimeException(pipeName);
     final TSStatus status = configManager.getProcedureManager().stopPipe(pipeName);
-    if (status == RpcUtils.SUCCESS_STATUS && isStoppedByRuntimeException) {
-      LOGGER.info(
-          "Pipe {} has stopped successfully manually, stop its auto restart process.", pipeName);
-      pipeTaskInfo.setIsStoppedByRuntimeExceptionToFalse(pipeName);
-      configManager.getProcedureManager().pipeHandleMetaChange(true, true);
+    if (status == RpcUtils.SUCCESS_STATUS) {
+      if (isStoppedByRuntimeException) {
```
SteveYurongSu, line 124:
```suggestion
      if (!isStoppedByRuntimeException) {
```

VGalaxies: No need, if modified in this way, the logic under this branch becomes meaningless 😖

PR #11647, opened 2023-11-30
Fix close method of TopOperator and make close method of Operator an interface

lancelly:
This PR adds a close method to TokOperator, ensuring that its child Operators are properly closed when it is closed, and guaranteeing that the SourceHandle and SinkChannel of the data transmission module are correctly cleaned up even when the query times out.
The close method of the Operator should be an interface, and the implementation of a new Operator must include logic for resource cleanup.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/exchange/sink/LocalSinkChannel.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/exchange/sink/LocalSinkChannel.java
@@ -57,8 +57,8 @@ public class LocalSinkChannel implements ISinkChannel {
       DataExchangeCostMetricSet.getInstance();
 
   public LocalSinkChannel(SharedTsBlockQueue queue, SinkListener sinkListener) {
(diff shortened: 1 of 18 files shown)


Beyyes (approved):
LGTM~

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/exchange/sink/LocalSinkChannel.java
```diff
@@ -57,8 +57,8 @@ public class LocalSinkChannel implements ISinkChannel {
       DataExchangeCostMetricSet.getInstance();
 
   public LocalSinkChannel(SharedTsBlockQueue queue, SinkListener sinkListener) {
-    this.sinkListener = Validate.notNull(sinkListener);
-    this.queue = Validate.notNull(queue);
+    this.sinkListener = Validate.notNull(sinkListener, "sinkListener can not be null.");
```
Beyyes, line 60: Use Validate.method(xx, string) may be time-consuming in cpu-bound situation such as benchant/tsbs, because it must construct a string object.

PR #11651, opened 2023-11-30
Ignore MismatchedType caused by delete

JackieTien97:
Now delete has concurrent bug which may cause some tsfile may not be deleted(the reason can be seen in https://apache-iotdb.feishu.cn/docx/FBqwd8562of2BJxDnQUcu3tRnsW, and we are trying to fix it but it may take some time), in such case, if we insert the same name sensor with different data type, while querying, it will throw excption:
img_v3_025m_858b41ee-692f-456f-a0fc-7bed1685025g


liuminghui233 (approved):
LGTM :-)

PR #11662, opened 2023-12-05
Fix tsfile counted twice in TsFileMetrics

HTHou:
## Description
image

In the recovery process of datanode, the updateRemainData method of TsFileMetrics will update the tsfile size and count twice than the real values.

After fixing, the number will be correct.

nrlm1UN4dZ


SpriCoder (approved):
LGTM~

PR #11669, opened 2023-12-06
Opt for only unpacking tsfile when it's really needed

JackieTien97:
Previously, we always unpack one seq and one unseq tsfile in hasNextFile.
image

In this pr, we try to unpack tsfile which we really need instead of unpacking at least one seq and one unseq timeseries metadata each time. In some case that has limit clasue, if the seq and unseq timeseries metadata are not overlapped, we can save one disk IO(if cache missed).


wangchao316 (approved):
LGTM,GOOD

PR #11679, opened 2023-12-08
[RatisConsensus] Retry Read Requests on StatusRuntimeException

SzyWilliam:
StatusRuntimeException will be thrown if the leader is offline (gRPC cannotconnect to the peer)
We can still retry in case it's a temporary network partition.


OneSizeFitsQuorum (approved):
LGTM

PR #11680, opened 2023-12-08
Pipe schema: RPC request & response datastructures

DanielWang2035:


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/connector/payload/request/TransferConfigPlanReq.java
```diff
@@ -0,0 +1,24 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iotdb.commons.pipe.connector.payload.request;
+
+import org.apache.iotdb.service.rpc.thrift.TPipeTransferReq;
+
+public abstract class TransferConfigPlanReq extends TPipeTransferReq {}
```
Caideyipi, line 24: May better not use this abstract class, since no class else will need this. Make configNode's PipeTransferConfigPlanReq directly extend TPipeTransferReq.

iotdb-core/datanode/src/test/java/org/apache/iotdb/db/pipe/connector/PipeDataNodeThriftRequestTest.java
```diff
@@ -167,4 +233,9 @@ public void testPIpeTransferFilePieceResp() throws IOException {
     Assert.assertEquals(resp.getStatus(), deserializeResp.getStatus());
     Assert.assertEquals(resp.getEndWritingOffset(), deserializeResp.getEndWritingOffset());
   }
+
+  @Test
+  public void testPIpeTransferSnapshotPieceResp() throws IOException {
+    // TODO: add test
```
Caideyipi, line 239: Plz complete this~

Caideyipi (commented):
PTAL~

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/connector/payload/request/TransferConfigPlanReq.java
DanielWang2035: PipeTransferConfigPlanReq is in CN package, so receiver on DN won't be able to directly import it. This abstract class is intended to be imported in the receiver.

iotdb-core/datanode/src/test/java/org/apache/iotdb/db/pipe/connector/PipeDataNodeThriftRequestTest.java
DanielWang2035: fixed

SteveYurongSu (approved):

PR #11682, opened 2023-12-08
Optimized wal file deletion algorithm

caozj1011:
### before this PR:

1. Select all less than unFlushMemTableMinVersion wal file, remember to CandidateFiles
2. In CandidateFiles, select all files smaller than safelyDeleteIndex, denoted as MustDeletedFiles
3. Delete MustDeletedFiles

### after this PR:
1. Obtain all wal files in a WalNode except the last file, which are called AllWalFiles
2. Obtain CheckPointManager. All not flush or be pin memTableId2Info MemTable, remember to ActiveMemTable
3. Get the set of MEMtables that each wal file contains, and refer to it as MemTableIdsOfWalMap
4. Traverse AllWalFiles and determine whether the deletion conditions in the optimization idea are met according to ActiveMemTable and MemTableIdsOfWalMap. If yes, delete the current file.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/node/WALNode.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/node/WALNode.java
@@ -402,8 +402,8 @@ public class WALNode implements IWALNode {
         WALFileStatus walFileStatus = WALFileUtils.parseStatusCode(currentWal.getName());
         long versionId = WALFileUtils.parseVersionId(currentWal.getName());
         if (canDeleteFile(searchIndex, walFileStatus, versionId)) {
+          long fileSize = currentWal.length();
           if (currentWal.delete()) {
-            long fileSize = currentWal.length();
             deleteFileSize += fileSize;
             Long memTableRamCostSum = walFileVersionId2MemTablesTotalCost.remove(versionId);
             if (memTableRamCostSum != null) {
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/node/WALNode.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/node/WALNode.java
@@ -457,7 +457,7 @@ public class WALNode implements IWALNode {
         return false;
       }
       // find oldest memTable
-      MemTableInfo oldestMemTableInfo = checkpointManager.getOldestMemTableInfo();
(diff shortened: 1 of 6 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/node/WALNode.java
```diff
@@ -252,20 +254,24 @@ public DeleteOutdatedFileTask() {
     }
 
     private void init() {
-      this.firstValidVersionId = initFirstValidWALVersionId();
```
HeimingZ, line 255: We don‘t need firstValidWALVersionId any more, but I think we should keep initFirstValidWALVersionId method because the delete procedure will try to delete current file in this method.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/node/WALNode.java
```diff
@@ -252,20 +254,24 @@ public DeleteOutdatedFileTask() {
...
-      this.filesShouldDelete = logDirectory.listFiles(this::filterFilesToDelete);
-      if (filesShouldDelete == null) {
-        filesShouldDelete = new File[0];
+      File[] allWalFilesOfOneNode = WALFileUtils.listAllWALFiles(logDirectory);
+      if (allWalFilesOfOneNode == null) {
+        allWalFilesOfOneNode = new File[0];
       }
+      WALFileUtils.ascSortByVersionId(allWalFilesOfOneNode);
+      this.sortedWalFilesExcludingLast =
+          Arrays.copyOfRange(allWalFilesOfOneNode, 0, allWalFilesOfOneNode.length - 1);
```
HeimingZ, line 263: Exception occurs when allWalFilesOfOneNode.length==0

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/node/WALNode.java
```diff
@@ -583,22 +578,28 @@ private void snapshotMemTable(DataRegion dataRegion, File tsFile, MemTableInfo m
...
+      if (memTableIdsOfCurrentWal.isEmpty()) {
+        return true;
+      }
+      for (MemTableInfo memTableInfo : activeOrPinnedMemTables) {
+        for (Long memTableId : memTableIdsOfCurrentWal) {
+          if (memTableId.equals(memTableInfo.getMemTableId())) {
+            return true;
+          }
         }
       }
```
HeimingZ, line 595: May we use Collections.disjoint() here?

HeimingZ (approved):

PR #11710, opened 2023-12-13
[RatisConsensus] Fix Precondition failure on consensus read

SzyWilliam:
image
Ratis required clientId serverId and groupId when constructing reply messge.


OneSizeFitsQuorum (approved):
LGTM

PR #11690, opened 2023-12-11
Pipe: automatically start pipe upon creation & allow to start RUNNING pipe & allow to stop STOPPED pipe 

VGalaxies:
## Description

As title, set the pipe status to `RUNNING` in create pipe procedure, so that users do not need to manually start the pipe after creating it.

**TODO:**

- [ ] This may lead to some IT failures.
- [x] Some comments need to be modified accordingly.

<hr>

This PR has:
- [x] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
    - [ ] concurrent read and write 
- [ ] added documentation for new or modified features or behaviors.
- [ ] added Javadocs for most classes and all non-trivial methods. 
- [ ] added or updated version, __license__, or notice information
- [ ] added comments explaining the "why" and the intent of the code wherever would not be obvious 
  for an unfamiliar reader.
- [ ] added unit tests or modified existing tests to cover new code paths, ensuring the threshold 
  for code coverage.
- [ ] added integration tests.
- [x] been tested in a test IoTDB cluster.

<hr>

PR #11713, opened 2023-12-13
Pipe: enhance the idempotence of start / stop pipe 

VGalaxies:
## Description

As title. This PR is a follow-up to #11690, from now onwards, no-op start pipe and stop pipe will directly skip the entire procedure execution, avoiding the impact of rollback and other exceptional situations. At the same time, the procedure result is set to facilitate differentiation on CN side for identifying no-op start pipe and stop pipe.

It is important to note that when starting a RUNNING pipe or stopping a STOPPED pipe, we need to consider the `isStoppedByRuntimeException` status of the pipe. Otherwise, certain procedures might be mistakenly skipped.

**TODO:**

- [x] IT
- [ ] transfer no-op msg to client side

<hr>

This PR has:
- [x] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


SteveYurongSu (approved):
LGTM

PR #11717, opened 2023-12-14
Improved stability of Ratis addMemberToGroup and testThresholdSnapshot UT

OneSizeFitsQuorum:
For addMemberToGroup UT:
- The previous UT operates the underlying state machine only after the leader is elected, and the state machine may not be fully restored at this time, thus making some subsequent judgments wrong. This PR changes to ensure that the Leader's state machine is not read until the leader is ready (i.e., the state machine is restored) to ensure that the CI environment does not fail

For testThresholdSnapshot UT:
- In CI environment, some threads are scheduled slowly, requiring 3 chances to generate Snapshot may lead to instability, so this PR is changed to 12 times, which is as low as possible in CI environment with weak resource environment


BUAAserein (approved):
LGTM

SzyWilliam (approved):
+1.
Mr. OneSizeFitsQuorum is the true master of code!

PR #11719, opened 2023-12-14
Add more query metrics in the fragment instance level

Beyyes:
## Description

1. Before, fragment_instance_context_size and fragment_instance_execution_size use AutoGauge, it will be updated every 15s using method `getInstanceContextSize` and `getInstanceExecutionSize` of `FragmentInstanceManager`.
Its latency is too high, I change to Histgram which will be updated every invoking of `execDataQueryFragmentInstance`

2. And I add a Timer to record the execution time of each FI

image

3. I update the load-timeseries-metadata timer, to record the total execution count and execution time of load aligned/non-aligned timeseries but not the single execution time. 

image
image

4. Add count and time statistics of construct chunk readers in FI level.
image

5. Add count and time statistics of page readers decompress in FI level.
image


iotdb-core/confignode/src/test/resources/confignode1conf/iotdb-common.properties
```diff
@@ -20,8 +20,8 @@
 timestamp_precision=ms
 data_region_consensus_protocol_class=org.apache.iotdb.consensus.iot.IoTConsensus
 schema_region_consensus_protocol_class=org.apache.iotdb.consensus.ratis.RatisConsensus
-schema_replication_factor=3
```
Beyyes, line 23: remove it later

iotdb-core/datanode/src/test/resources/datanode1conf/iotdb-datanode.properties
```diff
@@ -39,4 +39,4 @@ sort_tmp_dir=target/datanode1/tmp
 dn_metric_reporter_list=PROMETHEUS
 dn_metric_level=IMPORTANT
 dn_metric_async_collect_period=5
-dn_metric_prometheus_reporter_port=9097
```
Beyyes, line 42: remove these files later

JackieTien97 (approved):

PR #11722, opened 2023-12-14
Pipe: fix losing points when enable batch mode introduced by progress commit framework refactor

VGalaxies:
## Description

As title. With the introduction of `PipeEventCommitter`, each DR assigns commit id independently to events. Events with the same commit id in different DRs may be lost if using commit id as a condition for duplication check.

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


DanielWang2035:
LGTM. The reason is that one connector may transfer events from multiple dataregions, and events from different dataregions may have the same commit id. So we should use equals() instead of comparing commit ids.

SteveYurongSu (approved):

PR #11727, opened 2023-12-15
[IOTDB-5706] Data inconsistency between IoT protocol replications

BUAAserein:
## Description
The status code returned for handling BatchProcessException is `WRITE_PROCESS_ERROR`, but this status code was not included in the retry judgment of IoT, so the batch that caused BatchProcessException did not retry, resulting in inconsistent data between IoT replications.

To fix the problem, we add the status code `WRITE_PROCESS_ERROR` in IoT Retry judgment.

Besides, the atomicity issue of storage engine write operations can also lead to data inconsistency, which will be fixed in subsequent work.


OneSizeFitsQuorum (approved):
LGTM

PR #11728, opened 2023-12-15
Take snapshot for mods using copy instead of hard link

THUMarkLau:
Currently, a hard link of mods file is created when taking snapshot for data region, but the mods file is not immutable, which means the snapshot could be updated. We need to make sure the snapshot is immutable too, so a copy of mods file should be created instead of hard link.


OneSizeFitsQuorum (approved):
LGTM

PR #11737, opened 2023-12-18
[IOTDB-6266] Add the ability to flush syncIndex and update reader periodically for IoTConsensus

OneSizeFitsQuorum:
do not set SafelyDeletedSearchIndex as it is Long.MAX_VALUE when replica is 1.


caozj1011 (approved):

BUAAserein (approved):
LGTM

PR #11741, opened 2023-12-18
Refactor PBTree memory-management-related classes

MarcosZyk:
## Description

Separate NodeCache from original CacheManager, and it has two implementations, LRUNodeCache and PlainNodeCache.

Rename CacheManager to MemoryManager.

Rename MemManager to MemoryStatistics.

Refactor the original cache package to memory -> cache+buffer


Cpaulyz (approved):
LGTM!

PR #11742, opened 2023-12-18
Fix bug using null as measurement name

linxt20:
This modification is based on the bug identified in timechoDB—TIMECHO-483

## Issue Description
The IoTDB's cli session can run normally when receiving null as a template sequence name. The SQL test is as follows: create schema template nullt aligned (null TEXT, s_status BOOLEAN);

However, users found that when adding a measurement node with null as the measurement name to an existing template null_t using IoTDB's Java native interface, an NPE occurred during createSchemaTemplate by the session.

## Cause Description:
This is because the cli session of IoTDB processes the input SQL in the ASTVisitor by using parseNodeNameWithoutWildCard to handle the measurement name part as a String. Correspondingly, the Java native interface creates a measurementNode with null as the name, and at this point, the name as a String is assigned as null. Subsequently, the null value will be continuously passed to the server, but the measurement name cannot be null, resulting in this error.

## Issue Fix:
This problem fix is based on the principle that the measurement name cannot be null, and an error will be given for any occurrence of null received by the API.

Regarding the problem encountered in the Java native interface, considering that similar issues might exist in other language native interfaces, modifications were made to the server's code. In the StatementGenerator within CreateSchemaTemplateStatement, if the measurement name read is null, an error will be thrown.


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/parser/ASTVisitor.java
```diff
@@ -3313,6 +3314,12 @@ public Statement visitCreateSchemaTemplate(IoTDBSqlParser.CreateSchemaTemplateCo
       List<CompressionType> compressors = new ArrayList<>();
       for (IoTDBSqlParser.TemplateMeasurementClauseContext templateClauseContext :
           ctx.templateMeasurementClause()) {
+        if (templateClauseContext
+            .nodeNameWithoutWildcard()
+            .getText()
+            .equalsIgnoreCase(STRING_NULL)) {
+          throw new SemanticException("Device Template: measurement's name can not be 'null'.");
```
MarcosZyk, line 3321: Refactor the error msg to "The name of a measurement in schema template shall not be null."

MarcosZyk:
Please add some IT to validate this modification.

PR #11743, opened 2023-12-18
Fix some problems of FastCompactionPerformer

shuwenwei:
## Description
1. skip seal page if the next page cannot be flushed directly.
2. Fix the compaction metric if FastCompactionPerformer is used for sequence InnerSpaceCompaction


choubenson (approved):
LGTM

PR #11763, opened 2023-12-21
Pipe Schema: schema events

DanielWang2035:
This PR:
- Moved EnrichedEvent to node-commons package
- Defined Events for SchemaRegion and ConfigRegion
- Provided a farmework to pin and unpin Snapshots on CN and DN


Caideyipi (commented):
Some common parts can still be extracted, like in the other execution PR.

PR #11772, opened 2023-12-22
[IOTDB-6275] Disable limit/offset push down in ORDER BY

ycycse:
We cannot push limit down to scan if we have order by clause. 
Limit/Offset only works on the sorted results.


liuminghui233 (approved):
LGTM :-)

PR #11773, opened 2023-12-22
fix insert&delete cases dead data.

shuolinl:
In this PR:
1. remove useless write lock on schema cache when execute deletion
2. coordinator will acquire a read lock on schema cache, which will block deletion's cache invalidation.
3. solves the DEAD data problem: these data have no metadata but only data.


shuolinl:
You can reproduce the problem using the steps below:
add code at the end of function validateSchema() 
```java
    try {
      Thread.sleep(10 * 1000);
      // nothing
    }
```
run create and insert in one cli:
```java
CREATE TIMESERIES root.db.d1.s1 WITH DATATYPE=INT32, ENCODING=PLAIN
INSERT INTO root.db.d1(timestamp,s1) values(1,1)
```
run delete on another cli:
```java
delete timeseries  root.db.d1.s1
```

JackieTien97 (approved):

PR #11785, opened 2023-12-26
Ratis：transfer leader improvement

SzyWilliam:
After https://github.com/apache/ratis/pull/807 is introduced, now we can transfer leadership between nodes with same priorities. 
In this PR, I refactored the transferLeader implementation, remove the step where we first need to change the priority.
It brings a new benefit: since the priority is not changed, the original leader will not step down. If the transfership fails, the original leader can still be the leader and not affect the normal operations.


OneSizeFitsQuorum (approved):
LGTM

PR #11789, opened 2023-12-26
[IOTDB-6280] Parallel RegionGroup activation

CRZbulabula:
https://issues.apache.org/jira/secure/RapidBoard.jspa?rapidView=492&view=detail&selectedIssue=IOTDB-6280&quickFilter=2189

The RegionGroup activation is now started simultaneously

image

### DIFF
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/statemachine/CreateRegionGroupsProcedure.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/statemachine/CreateRegionGroupsProcedure.java
@@ -173,12 +173,9 @@ public class CreateRegionGroupsProcedure
         // Build RegionGroupCache immediately to make these successfully built RegionGroup available
         createRegionGroupsPlan
             .getRegionGroupMap()
-            .entrySet()
-            .parallelStream()
             .forEach(
-                regionReplicaEntry ->
-                    regionReplicaEntry
-                        .getValue()
+                (database, regionReplicaSets) ->
+                    regionReplicaSets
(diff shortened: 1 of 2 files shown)


OneSizeFitsQuorum (approved):
LGTM~
The current code is consistent with the current style, but may need to be changed to an asynchronous style in the future to avoid synchronous sleep occupancy of threads

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/statemachine/CreateRegionGroupsProcedure.java
```diff
@@ -173,40 +173,47 @@ protected Flow executeFromState(ConfigNodeProcedureEnv env, CreateRegionGroupsSt
         // Build RegionGroupCache immediately to make these successfully built RegionGroup available
         createRegionGroupsPlan
             .getRegionGroupMap()
+            .entrySet()
+            .parallelStream()
```
OneSizeFitsQuorum, line 177: I suggest we remove the parallelism here, as it's very unlikely that multiple sg regions will be created at the same time, but parallel here and below might introduce deadlock issues in some scenarios

OneSizeFitsQuorum (commented):
PTAL

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/statemachine/CreateRegionGroupsProcedure.java
CRZbulabula: Fixed

PR #11798, opened 2023-12-28
[IOTDB-6281] Enhance Procedure metrics

BUAAserein:
## Description

The execution status of the Procedure after submission is difficult to detect, and the execution time of the Procedure may be too long. So, we will enhance Procedure metrics in this pr.

We want to enhance Procedure metrics on three levels:

- Thread pool level:
   - Total thread pool size
   - Number of active threads
   - Queue length
   
- Procedure level:
   - Submmit Rate/Failed Rate Counter for different types of Procedures
   - The completion time of different types of Procedures is P50/P99/AVG
   
- Log level: 
   -  Do some Log optimization

### DIFF
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/ConfigManager.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/ConfigManager.java
@@ -102,7 +102,6 @@ import org.apache.iotdb.confignode.persistence.partition.PartitionInfo;
 import org.apache.iotdb.confignode.persistence.pipe.PipeInfo;
 import org.apache.iotdb.confignode.persistence.quota.QuotaInfo;
 import org.apache.iotdb.confignode.persistence.schema.ClusterSchemaInfo;
-import org.apache.iotdb.confignode.procedure.ProcedureMetrics;
 import org.apache.iotdb.confignode.rpc.thrift.TAlterLogicalViewReq;
 import org.apache.iotdb.confignode.rpc.thrift.TAlterSchemaTemplateReq;
 import org.apache.iotdb.confignode.rpc.thrift.TAuthizedPatternTreeResp;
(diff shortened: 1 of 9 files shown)


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/ConfigManager.java
```diff
@@ -248,6 +249,8 @@ public class ConfigManager implements IManager {
 
   private static final String DATABASE = "\tDatabase=";
 
+  private ProcedureMetrics procedureMetrics;
```
OneSizeFitsQuorum, line 252: why we put this into ConfigManager rather than ProcedureManager?

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/Procedure.java
```diff
@@ -900,4 +895,41 @@ public static boolean haveSameParent(Procedure<?> a, Procedure<?> b) {
...
   }
+
+  /**
+   * This function will be called just when procedure is submitted for execution.
+   *
+   * @param env The environment passed to the procedure executor
+   */
+  protected void updateMetricsOnSubmit(Env env) {
+    ProcedureMetrics metrics = null;
+    if (env instanceof ConfigNodeProcedureEnv) {
```
OneSizeFitsQuorum, line 906: Why we need this judgement?

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/ConfigManager.java
BUAAserein: Referring to NodeMetrics and PartitionMetrics. Put it in ConfigManager for easy access to addMetrics and removeMetrics

BUAAserein: fixed

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/Procedure.java
BUAAserein: removed

BUAAserein:
image

OneSizeFitsQuorum (approved):
LGTM

PR #11808, opened 2023-12-29
Optimize log according to real scenarios

SpriCoder:
We analyze logs of iotdb in many scenarios, and we decide to change level of some logs and remove some useless logs, more details in: https://apache-iotdb.feishu.cn/docx/ACRUd32DPozuUTxK7MBcPWCBn0d


OneSizeFitsQuorum (approved):
LGTM

PR #11813, opened 2024-01-02
Move path related antlr defined files to tsfile module

JackieTien97:
In this pr, I do some prepare works for tsfile module moving out:

1. move some Path related antlr defined g4 file to tsfile module, so that tsfile module won't depend on iotdb-antlr module
2. move `into` Pattern related code from tsfile module to datanode module.


qiaojialin (approved):

wangchao316 (approved):
LGTM

PR #11815, opened 2024-01-02
Add log for NotEnoughDataNodeException

CRZbulabula:
Add log for NotEnoughDataNodeException. Therefore when the cluster DataNode is not enough, we can compare the detailed information between registered DataNodes and the replicationFactor.


OneSizeFitsQuorum (approved):
LGTM

PR #11816, opened 2024-01-02
Feture collection iotdb info

CloudWise-Lukemiao:
Feture collection iotdb info


CritasWang (approved):
LGTM

Some operating systems also have compatibility issues, such as Windows Server 2012

PR #11819, opened 2024-01-02
Clear status in InsertNodes to fix consensus endless retry

HTHou:
When an InsertTabletsNode contains the system reject status, the consensus will retry it endlessly.


OneSizeFitsQuorum (approved):
LGTM

PR #11822, opened 2024-01-02
try fix SessionPoolTest UT

JackieTien97:
In Session & SessionPool UT, we should set `enableAutoFetch` to `false` to avoid NPE in close.
image


CritasWang (approved):
LGTM

PR #11823, opened 2024-01-02
Fix timeout doesn't take effect in schema fetch bug

JackieTien97:
more details can be seen in https://jira.infra.timecho.com:8443/browse/TIMECHODB-523


Cpaulyz (approved):
LGTM～

PR #11827, opened 2024-01-02
Print Exception only when encountering unexpected TSStatus

JackieTien97:
PATH_ALREADY_EXIST and ALIAS_ALREADY_EXIST is what we expect, we should never print it as Exception


OneSizeFitsQuorum (approved):
LGTM

PR #11835, opened 2024-01-03
Optimize log according to real scenarios after checking

SpriCoder:
Optimize log according to real scenarios after checking


OneSizeFitsQuorum (approved):
LGTM

PR #11842, opened 2024-01-04
Return when leader election waiting is interrupted

CRZbulabula:
There is no need to pay more time when leader election was interrupted. Since all the information just stored in memory. The ConfigNode will re-calculate them when restart.


OneSizeFitsQuorum (approved):
LGTM

PR #11849, opened 2024-01-04
PBTree: fix deed lock in ReleaseFlushMonitor and Scheduler

Cpaulyz:
## Description

Dead lock happened when:
* Thread A try to flush all, hold Scheduler's synchronized lock, acquire store read lock.
* Thread A try to schedule flush, hold store read lock, acquire Scheduler's synchronized lock.

Solution: Make store all read&write locks are acquired after Scheduler's synchronized lock.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/impl/pbtree/flush/Scheduler.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/impl/pbtree/flush/Scheduler.java
@@ -95,8 +95,12 @@ public class Scheduler {
                     entry ->
                         CompletableFuture.runAsync(
                             () -> {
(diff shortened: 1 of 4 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/impl/pbtree/flush/Scheduler.java
```diff
@@ -228,12 +246,12 @@ public synchronized void scheduleFlush(List<Pair<Integer, Long>> regionIds) {
                 LOGGER.debug("It takes {}ms to flush MTree in SchemaRegion {}", time, regionId);
               }
               store.recordFlushMetrics(time, flushNodeNum.get(), flushMemSize.get());
-              lockManager.globalStampedReadUnlock(pair.getRight());
+              lockManager.globalReadLock();
```
MarcosZyk, line 249: This should be unlock operation.

PR #11865, opened 2024-01-09
[IOTDB-6190] Increase the threshold for Ratis to shut itself down if it detects that a process is stuck

OneSizeFitsQuorum:
After Ratis was updated to version 3.0, the way this parameter is set has changed, and the previous [PR](https://github.com/apache/iotdb/pull/11416/files) is no longer valid and needs to be modified


SzyWilliam (approved):
+1 LGTM. Thanks @OneSizeFitsQuorum for the patch!

PR #11871, opened 2024-01-10
[RatisConsensus] Enable Leader Transfer UT

SzyWilliam:
Enable Leader Transfer UT for preparation on load balancing.

100x tests passed locally.

image
image


OneSizeFitsQuorum (approved):
LGTM

PR #11883, opened 2024-01-11
Exclude gc time from cpu usage percentage for thread pool and modules

THUMarkLau:
# Background
In past CPU monitoring, the CPU usage of each thread/thread pool/module equaled the CPU time it occupied divided by the total CPU time of all threads, multiplied by the process's CPU utilization. However, this method has a flaw. Since we cannot obtain the CPU time of the GC thread, the calculated thread CPU utilization is inaccurate. When extensive GC occurs, it may seem like all threads' CPU utilization has risen because their ratios remain unchanged, but the process's CPU utilization has increased. This is actually due to GC. To eliminate this influence, this PR removes the impact of GC when calculating CPU utilization. Experimental results show that this PR has achieved the expected effect.
# Experiment
To verify this PR, I added a piece of code to the server side of IoTDB. Each time an insertRecords request is processed, 2MB of memory is allocated and stored in a global list to ensure that this memory is not reclaimed. After running for a while, the system will start to experience intense GC. However, at this time, the CPU utilization of other threads should remain the same or decrease. Therefore, we should observe a decrease in the process's CPU utilization, while the CPU utilization of each module remains unchanged or decreases.
image
image
image
image
As shown in the figure above, when intense GC occurs, the CPU utilization of the process increases, but the utilization of the ClientRPC thread decreases, which is consistent with expectations.


OneSizeFitsQuorum (approved):
LGTM

PR #11885, opened 2024-01-11
Feat/cli with overridable streams

chrisdutz:
In order to provide a WebCLI, I needed to be able to pass in InputStream and OutputStreams to the CLI, so I could redirect input and output to a WebSocket.

I updated the CLI so when using the "main" it simply initializes the CliContext with System.stdin and System.stdout, but also allows calling a static runCli method where I can pass in a manually constructed CliContext.


CritasWang:
image

chrisdutz:
Thanks for spotting this ... issue now seems to be solved.

OneSizeFitsQuorum (commented):
Hi, Chris, How about splitting this cli PR with another resetting static final field PR into 2 merge requests?

chrisdutz:
There is one for just the static stuff ... Just for my current work, I need both changes, so I based this PR off the other one. 

https://github.com/apache/iotdb/pull/11870

chrisdutz:
However I think there might be one singleton that I learned too late that it is needed ...

CritasWang:
image

hi @chrisdutz 
This change will cause the CLI to not work properly.

For example:
The input statement will be repeatedly output
Unable to auto-complete keywords
Unable to select SQL statement by up and down keys

chrisdutz:
This is really odd ... as it shouldn't change anything, just instead of using "System.in" and "System.out" you use variables that are initialized with them. 

But I'll have a look, if I can spot what's causing this.

chrisdutz:
But I can confirm, that the code seems to be behaving differently, even if it shouldn't be. I just switched to master, rebuilt and ran the CLI and can see what you were referring to. I'll try to figure out what's causing the difference.

chrisdutz:
Ok ... was a bit tricky, but in the end I pushed a change that should fix it. There were actually two issues: 
1. I was running the CLI in IntelliJ (not from the console) ... here both version behaved identically (Even when running the CLI from master in IntelliJ directly, it will echo the commands and not have auto-complete).
2. I was expecting to setting "in" and "out" to System.in and System.out woudln't be any change, but it turns out in this case the internal variables "in" and "out" are not null and the code will take a wrong turn.

I updated the code to simply not set the "streams" if "in == System.in && out == System.out" ... that seemed to have worked.

iotdb-client/cli/src/main/java/org/apache/iotdb/cli/IoTDBSyntaxHighlighter.java
```diff
@@ -67,12 +67,14 @@ public AttributedString highlight(LineReader reader, String buffer) {
 
   @Override
   public void setErrorPattern(Pattern errorPattern) {
-    IoTPrinter.print("");
+    // TODO: Not quite sure what this code should actually do
```
OneSizeFitsQuorum, line 70: This class was temporarily disabled in [PR](https://github.com/apache/iotdb/blob/master/iotdb-client/cli/src/main/java/org/apache/iotdb/cli/utils/JlineUtils.java#L90).

Perhaps we can remove this class, or leave it at that? Adding these comments does not seem to help

iotdb-client/cli/src/main/java/org/apache/iotdb/cli/IoTDBSyntaxHighlighter.java
```diff
@@ -67,12 +67,14 @@ public AttributedString highlight(LineReader reader, String buffer) {
...
+    new IoTPrinter(System.out).print("");
   }
 
   @Override
   public void setErrorIndex(int errorIndex) {
-    IoTPrinter.print("");
+    // TODO: Not quite sure what this code should actually do
```
OneSizeFitsQuorum, line 76: same as above

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/DataNode.java
```diff
@@ -917,6 +925,9 @@ public void stop() {
       if (dataRegionConsensusStarted) {
         DataRegionConsensusImpl.getInstance().stop();
       }
+      /*if (IoTDBRestServiceDescriptor.getInstance().getConfig().isEnableRestService()) {
```
OneSizeFitsQuorum, line 928: seems we can remove this line? or just put this uncommented line to `deactivate()`?

OneSizeFitsQuorum (commented):
Hi, Chris, thanks for your awesome work! The PR contains some work to change the singleton schema, but I don't feel it is complete, such as `DataRegionConsensusImpl`. I suggest we go ahead to see if there are any missing singletons.

BTW, our current change would break our default inner private class-based singleton pattern (due to the loss of `final`), which would have no real impact, but the code style would look a bit strange. What do you think about it? Giving way to code style for functionality seemed acceptable to me, as I couldn't think of a better solution

iotdb-client/cli/src/main/java/org/apache/iotdb/cli/IoTDBSyntaxHighlighter.java
chrisdutz: Yeah ... I added comments to code that I didn't understand ... in this particular case I couldn't see what the code actually did, so I added a comment.

chrisdutz:
Hi,
yeah ... it surely doesn't address all singletons, just the ones I needed for my work. Of course adjusting all singletons is the more complete work, but I wanted to change as little as possible. 

In general do I think we should completely get rid of any form of static variables for anything but constant definitions. 

Currently I would be in strong favor of changing things to compose things using constructors, similar to Spring Constructor Injection, but to manually weave the components together, so we can generally use IoTDB in any form of application, no matter the component framework being used.

However do I know this is a big change and possibly one that could break compatibility with extensions built for older versions, so I think we should do the "resettable statics" first and focus on a more general refactoring later on.

OneSizeFitsQuorum (approved):
LGTM

PR #11887, opened 2024-01-12
Repair unsorted file by compaction

shuwenwei:
## Description
Add a new system maintainance command `repair data (ON (LOCAL | CLUSTER))`. 
When execute this command, the datanode will start a repair schedule task. This task check all files and will submit a repair task if the file has internal unsorted data or the file is overlapped with other files.
When the repair task is running, all compaction schedule task are blocked until the repair schedule task is finished.
If a restart occurred during the repair, it will recover from a repair log file in system/repair directory and skip the time partitions which is repaired.
If a file can not be repaired by repair task, it will be marked as CANNOT_REPAIR in memory. To keep the mark available after restart, these files are also recorded in repair log file.


shuwenwei:
> 1. I have a question. this will read all tsfile in a storage group ? .
> 2. if have too much data (10T), how long does repair compaction?
> 3. if need repair data is too much, does this occur oom?

1. This command will read all tsfile in all data directory because we can't know which file has unsorted data.
2. It can take a long time to scan all file. So, the repair task scheduler has a log file to keep the progress after restart. 
3. Every file which has unsorted data will be repaired by submitting a compaction task.  It acquire memory from allocated compaction memory and will estimated the required memory cost before execute repairing.

caozj1011 (changes requested):

HTHou (approved):

PR #11901, opened 2024-01-15
Clear useless query resource ASAP

JackieTien97:
In some big query, like select * from root.db.** that query all devices.

1. the query plan will be memory-consuming and actually LogicalPlan and DistributionPlan is useless now after we dispatch all FIs.

2. these queries' Physical Operator Tree will also be very large and we need to set to null as soon as possible. And in most operators (like TopKOperator),  we've already done that, we will set i-th child to null if we detect there will be no data for that child. However, I found that we still save all SourceOperators in DataDriverContext which just used for initializing `QueryDataSource`. So we should set them to null after `QueryDataSource` being initialized.


Beyyes (approved):
LGTM

PR #11914, opened 2024-01-17
[IOTDB-6291] Update RegionGroup configuration when update DataNode configuration

CRZbulabula:
https://issues.apache.org/jira/browse/IOTDB-6291

Local simple test:

Before update:
before

After update:
after


OneSizeFitsQuorum (approved):
LGTM

PR #11915, opened 2024-01-17
[IOTDB-6292] Add the internal address column in the result of show regions

CRZbulabula:
https://issues.apache.org/jira/browse/IOTDB-6292

image


OneSizeFitsQuorum (approved):
LGTM

PR #11924, opened 2024-01-18
Load: Introduce LoadTsFileManager.CleanupTask to force close writer manager after exception occurring

SteveYurongSu:
As title. 

This can avoid load clean thread accumulate continously.

image


yschengzi (approved):
LGTM

PR #11928, opened 2024-01-18
Fix the issue where ConfigNode reports 'ConsensusGroupAlreadyExist' error during startup in some scenarios

OneSizeFitsQuorum:
img_v3_0277_e93e2168-74f4-4e9c-a2fc-8ffb2325c43g

In some operating systems, creating a consensus group takes longer than the default ratis client timeout of 10 seconds, so the current method of using RaftClient to create a consensus group locally will trigger a retry, resulting in a log error.

This PR solves the problem by changing RPC to LPC


CRZbulabula (approved):
LGTM

PR #11930, opened 2024-01-18
Fixed the issue that IoTConsensus synchronization is stuck when WAL file is corrupted

OneSizeFitsQuorum:
For a corrupted WAL file, the storage engine will skip the file on restart without affecting the restart. However, IoTConsensus will be trapped in wireless retry when it encounters a corrupted WAL file causing index jump, which will lead to synchronization deadlock.


HeimingZ (approved):

BUAAserein (approved):
LGTM

PR #11944, opened 2024-01-22
Pipe: result of show pipes statement is incorrect when it contains back quoted pipe names

xuanronaldo:
1. Filter back quote in pipe name, but it's still necessary to and back quote if there is illegal character in pipe name.
2. Fix bug:  result of pipe's query statement contains back quote is incorrect.


SteveYurongSu (approved):

SteveYurongSu (changes requested):
Plz fix IT

SteveYurongSu (approved):

SteveYurongSu (approved):

PR #11945, opened 2024-01-22
Fix PointPriorityReader calculation of aligned series null value num

shuwenwei:
## Description
Fix PointPriorityReader calculation of aligned series null value num.
The decreasing of null value num may repeated in some cases, which cause some series lost point.
If the result of null value num is negative, point won't lost through the calculation of this field is wrong.
image
If the result equals 0, the loop break directly and some point may not merged into current time value pair.
image


choubenson (approved):
LGTM

PR #11965, opened 2024-01-24
Pipe: Move parsing event logic into PipeProcessorSubtask to separate CPU intensive and IO intensive tasks

DanielWang2035:
Changes:
- moved parse event logic from connector plugins to `PipeProcessorSubtask`.
- added a buffer queue in `PipeProcessorSubtask` to store parsed events.

Note:
- now the sequence of an event is: extractor -> parse -> processor -> connector
- now commit-ids are assigned to **parsed** events, not the ones before parsing

For memory control, parsed events are stored in two places: `PipeProcessorSubtask#parsedEventQueue` and `PipeEventCollector#bufferQueue`. The former one will only store events parsed from one source event, and `PipeProcessorSubtask` will not parse new events if the latter one is not empty. So there won't be too much parsed events in queue..


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/task/subtask/processor/PipeProcessorSubtask.java
```diff
@@ -199,6 +225,58 @@ public int hashCode() {
     return taskID.hashCode();
   }
 
+  //////////////////////////// Parsing events ////////////////////////////
+
+  /** Parse an event by pattern and time, then add parsed events into {@code parsedEventQueue} */
+  private void parseEventByPatternAndTime(Event sourceEvent) throws Exception {
```
Caideyipi, line 231: The parsed event's reference count shall be decreased, and the newly generated events' reference count shall be increased, to avoid wal/tsfile resource leak.

Caideyipi: If a tsFile event generates no tablets, it shall be reported immediately; If not, the report shall be postponed to the sent time of the last raw tablet.

Caideyipi (changes requested):
PTAL~

SteveYurongSu (approved):

PR #11969, opened 2024-01-24
Fixed issue where ConfigNode might recover incorrectly under SimpleConsensus

BUAAserein:
## Description

SimpleConsensus logs are restored in the wrong order during startup.


OneSizeFitsQuorum (approved):
LGTM

PR #11971, opened 2024-01-25
Pipe: stop pipe using restarting strategy to unpin the wal's reference count to avoid WAL stacking

MiniSho:
## Description
Change the execution flow of the `stop pipe` : 
1. drop the pipe task firstly 
2. then create a new pipe task with the same pipeMeta.

When the pipe is dropped and re-created with the same pipeMeta, it will re-trigger the collection of historical data, and most of the files it pins are tsfile files not wal, because the wal files are flushed. By taking advantage of this property, we can make sure that the subsequent pins are new files, not old wal files, thus avoiding the wal stacking.

### DIFF
--- a/iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/agent/task/PipeTaskAgent.java
+++ b/iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/agent/task/PipeTaskAgent.java
@@ -555,34 +555,29 @@ public abstract class PipeTaskAgent {
   }
 
   protected void stopPipe(String pipeName, long creationTime) {
-    final PipeMeta existedPipeMeta = pipeMetaKeeper.getPipeMeta(pipeName);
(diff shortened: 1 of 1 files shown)


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/agent/task/PipeTaskAgent.java
```diff
@@ -555,34 +555,29 @@ private void startPipe(String pipeName, long creationTime) {
...
+    // 1. Drop the pipe task
     final long startTime = System.currentTimeMillis();
-    pipeTasks.values().parallelStream().forEach(PipeTask::stop);
+    dropPipe(pipeName);
+
+    // 2. Set pipe meta status to STOPPED
+    pipeMeta.getRuntimeMeta().getStatus().set(PipeStatus.STOPPED);
+
+    // 3. create a new pipe with the same pipeMeta
+    createPipe(pipeMeta);
```
SteveYurongSu, line 575: image

SteveYurongSu (approved):

PR #12005, opened 2024-01-29
Introduce UDAF(User-Defined Aggregation Functions)

Sh-Zh-7:
## Description

This PR we introduce a new feature: **UDAF(User-Defined Aggregation Functions)**.
Now users can customize their aggregation functions and seamlessly integrate them into IoTDB powerful aggregation framework. 🎉🎉

More details can be seen: https://apache-iotdb.feishu.cn/docx/HNawdvH1Goo5sRxtmX9cjLHyniF
The User Manual will come later soon.

<hr>

This PR has:
- [x] been self-reviewed.
  for code coverage.

<hr>


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/AggregationUtil.java
```diff
@@ -58,6 +60,9 @@ public class AggregationUtil {
 
   private static final String PARTIAL_SUFFIX = "_partial";
 
+  private static final List<String> builtinAggregationNames =
```
JackieTien97, line 63: Use `BuiltinAggregationFunction.NATIVE_FUNCTION_NAMES` instead.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/parameter/AggregationDescriptor.java
```diff
@@ -63,18 +64,29 @@ public class AggregationDescriptor {
 
   protected String parametersString;
 
+  private List<String> builtinAggregationNames;
+
```
JackieTien97, line 68:
```suggestion
```

never used.

JackieTien97 (changes requested):
Remember to add a UDAF example in `example/udf` submodule.

After this pr being merged, you need also update the docs in website. submit a pr to this repo https://github.com/apache/iotdb-docs

iotdb-api/udf-api/src/main/java/org/apache/iotdb/udf/api/customizer/config/UDAFConfigurations.java
```diff
@@ -0,0 +1,38 @@
...
+
+package org.apache.iotdb.udf.api.customizer.config;
+
+import org.apache.iotdb.udf.api.type.Type;
+
+import java.time.ZoneId;
+
+public class UDAFConfigurations extends UDFConfigurations {
+
+  protected final ZoneId zoneId;
```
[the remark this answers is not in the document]
lancelly: "UDTFConfiguration retains ZoneId, but it is not used in the internal implementation or built-in UDTFs. It may simply be kept for users who need it."

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/aggregation/UDAFAccumulator.java
```diff
@@ -0,0 +1,219 @@
...
+  @Override
+  public void setFinal(Column finalResult) {
+    throw new UnsupportedOperationException(getClass().getName());
+  }
+
+  @Override
+  public void outputIntermediate(ColumnBuilder[] columnBuilders) {
+    checkArgument(columnBuilders.length == 1, "partialResult of UDAF should be 1");
+
+    byte[] bytes = state.serialize();
```
lancelly, line 157: This method serializes the initial state into bytes[] (even if the accumulator hasn't had any values added, it will still serialize the initial state?), but in other Accumulator implementations, it checks whether an initial value has already been set, and if so, serializes the binary data; otherwise, it appends Null. Currently, it seems that this method doesn't check whether an initial value has been initialized. I wonder if this could potentially have an impact on the result.

Sh-Zh-7: I think NULL value here is just to save some space. While all built-in aggregators use additional field (like `initResult`) to indicate whether the state is initialized, we can't force users to do so. For simplicity, we just serialize the whole state object into binary.

The problem is, some built-in aggregators will append NULL in columnBuilder if the state is uninitialized, and the NULL value would be skipped in `addIntermediate()` and `removeIntermediate()` method. Since UDAF won't produce NULL intermediate output, shall we remove the **skip the NULL logic** in both methods?

IMHO, if the logic is removed and UT/IT is still passed, we can consider to safely apply this change.

Sh-Zh-7: If uninitialized state is needed, users shall add a field like `isInit` in their state class, and serialize the whole state object along with the field into binary, then handle this field in `combineState()` method. You can refer to this example: [UDAFSum.java](https://github.com/apache/iotdb/pull/12005/files#diff-e9b1b5efc75765bc2ddc3ad158368fdf40cc0b6a716c5c01b1a75899c849740dR111)

iotdb-api/udf-api/src/main/java/org/apache/iotdb/udf/api/UDAF.java
```diff
@@ -0,0 +1,70 @@
...
+
+  /** Create and initialize state. You may bind some resource in this method. */
+  State createState();
+
+  /**
+   * Batch update state with data columns. You shall iterate columns and update state with raw
+   * values
+   *
+   * @param state state to be updated
+   * @param column input columns from IoTDB TsBlock
```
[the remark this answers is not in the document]
Sh-Zh-7: Done.

iotdb-api/udf-api/src/main/java/org/apache/iotdb/udf/api/UDAF.java
```diff
@@ -0,0 +1,70 @@
...
+   * @param bitMap define some filtered position in columns
+   */
+  void addInput(State state, Column[] column, BitMap bitMap);
```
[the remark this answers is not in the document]
Sh-Zh-7: Done.

iotdb-api/udf-api/src/main/java/org/apache/iotdb/udf/api/customizer/config/UDAFConfigurations.java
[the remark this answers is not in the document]
Sh-Zh-7: Temporarily remove this field. We can add it when we actually need it in the future,

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/AggregationUtil.java
Sh-Zh-7: Done.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/parameter/AggregationDescriptor.java
Sh-Zh-7: Done.

JackieTien97 (approved):

lancelly (approved):

PR #12006, opened 2024-01-30
Avoid deleting wal files pinned by the iot consensus

HeimingZ:
image
image

wal file will be deleted when its startSearchIndex is smaller SafelyDeletedSearchIndex, but this operation may delete the wal file which contains nodes for iot consensus. To fix this, we revert some logic to the previous version.


caozj1011 (approved):

BUAAserein (approved):
LGTM

OneSizeFitsQuorum (approved):
LGTM

PR #12008, opened 2024-01-30
Implement PageLifecycleManager for global page control

Cpaulyz:
## Description

Implement PageLifecycleManager for global page control


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/impl/pbtree/CachedMTreeStore.java
```diff
@@ -556,6 +557,7 @@ public void clear() {
         }
       }
       file = null;
+      PageLifecycleManager.getInstance().removePagePool(schemaRegionId);
```
MarcosZyk, line 560: Should this be placed in file.clear()?

Cpaulyz: > Should this be placed in file.clear()?

Yes. But schema region id is invisible in file...

PR #12015, opened 2024-02-01
Fix non time queries with time column issues

CloudWise-Lukemiao:
Fix non time queries with time column issues


CritasWang (approved):
LGTM

PR #12018, opened 2024-02-01
Pipe: support replace and modify mode for alter pipe sql

VGalaxies:
## Description

As title.

<hr>

This PR has:
- [x] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/integration-test/src/test/java/org/apache/iotdb/pipe/it/IoTDBPipeAlterIT.java
+++ b/integration-test/src/test/java/org/apache/iotdb/pipe/it/IoTDBPipeAlterIT.java
@@ -22,6 +22,7 @@ package org.apache.iotdb.pipe.it;
 import org.apache.iotdb.commons.client.sync.SyncConfigNodeIServiceClient;
 import org.apache.iotdb.confignode.rpc.thrift.TShowPipeInfo;
 import org.apache.iotdb.confignode.rpc.thrift.TShowPipeReq;
+import org.apache.iotdb.db.it.utils.TestUtils;
 import org.apache.iotdb.it.env.cluster.node.DataNodeWrapper;
 import org.apache.iotdb.it.framework.IoTDBTestRunner;
 import org.apache.iotdb.itbase.category.MultiClusterIT2;
@@ -34,7 +35,10 @@ import org.junit.runner.RunWith;
 import java.sql.Connection;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import static org.junit.Assert.fail;
 
@@ -43,7 +47,7 @@ import static org.junit.Assert.fail;
 public class IoTDBPipeAlterIT extends AbstractPipeDualIT {
 
   @Test
-  public void testBasicPipeAlter() throws Exception {
+  public void testBasicAlterPipe() throws Exception {
     DataNodeWrapper receiverDataNode = receiverEnv.getDataNodeWrapper(0);
 
     // create pipe
@@ -58,15 +62,15 @@ public class IoTDBPipeAlterIT extends AbstractPipeDualIT {
       fail(e.getMessage());
     }
 
-    // show pipe and record first creation time
-    long firstCreationTime;
+    // show pipe and record last creation time
+    long lastCreationTime;
     try (SyncConfigNodeIServiceClient client =
         (SyncConfigNodeIServiceClient) senderEnv.getLeaderConfigNodeConnection()) {
       List<TShowPipeInfo> showPipeResult = client.showPipe(new TShowPipeReq()).pipeInfoList;
       Assert.assertEquals(1, showPipeResult.size());
       Assert.assertTrue(showPipeResult.get(0).pipeConnector.contains("batch.enable=false"));
       Assert.assertEquals("RUNNING", showPipeResult.get(0).state);
-      firstCreationTime = showPipeResult.get(0).creationTime;
+      lastCreationTime = showPipeResult.get(0).creationTime;
     }
 
     // stop pipe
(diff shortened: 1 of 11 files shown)


integration-test/src/test/java/org/apache/iotdb/pipe/it/IoTDBPipeAlterIT.java
```diff
@@ -103,16 +104,62 @@ public void testBasicPipeAlter() throws Exception {
...
+          showPipeResult
+              .get(0)
+              .pipeConnector
+              .contains(String.format("node-urls=%s", receiverDataNode.getIpAndPortString())));
+      // alter pipe will keep running status
+      Assert.assertEquals("RUNNING", showPipeResult.get(0).state);
+      // alter pipe will reset pipe creation time
+      Assert.assertTrue(showPipeResult.get(0).creationTime > lastCreationTime);
       // alter pipe will clear exception messages
       Assert.assertEquals("", showPipeResult.get(0).exceptionMessage);
```
Caideyipi, line 157: The exception message may always be empty, whether or not the "alter pipe" clear it.

integration-test/src/test/java/org/apache/iotdb/pipe/it/IoTDBPipeAlterIT.java
```diff
@@ -138,17 +185,60 @@ public void testPipeAlterFailure() {
...
+      statement.execute("alter pipe a2b modify processor ('down-sampling.interval-seconds'='2')");
+    } catch (SQLException e) {
+      fail(e.getMessage());
+    }
+
+    // check data on receiver
+    expectedResSet.clear();
+    expectedResSet.add("1000,1.0,");
+    expectedResSet.add("3000,5.0,");
+    TestUtils.assertDataOnEnv(
```
Caideyipi, line 241: The previous progressIndex is reported, though they might be sent again when flushed to a single tsFile. However, assuring "retransmission" may not be good.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/pipe/PipeTaskInfo.java
```diff
@@ -180,30 +180,36 @@ private void checkBeforeAlterPipeInternal(TAlterPipeReq alterPipeRequest) throws
...
-          String.format(
-              "Failed to alter pipe %s, nothing to alter", alterPipeRequest.getPipeName());
-      LOGGER.warn(exceptionMessage);
-      throw new PipeException(exceptionMessage);
+          copiedPipeStaticMetaFromCoordinator.getConnectorParameters().getAttribute());
+    } else if (!alterPipeRequest.isReplaceAllConnectorAttributes) {
+      alterPipeRequest.setConnectorAttributes(
+          copiedPipeStaticMetaFromCoordinator
+              .getConnectorParameters()
+              .updateEquivalentAttributes(alterPipeRequest.getConnectorAttributes())
```
Caideyipi, line 211: What if the user wants to add some params into an existing plugin?

iotdb-api/pipe-api/src/main/java/org/apache/iotdb/pipe/api/customizer/parameter/PipeParameters.java
```diff
@@ -307,6 +307,24 @@ public boolean isEquivalent(Object obj) {
...
+   * @param attributes Provide the key that needs to be updated along with the new value.
+   * @return this pipe parameters
+   */
+  public PipeParameters updateEquivalentAttributes(Map<String, String> attributes) {
+    attributes.forEach(
+        (k, v) ->
+            this.attributes.entrySet().stream()
+                .filter(
+                    entry ->
+                        entry.getKey().equals(k) || entry.getKey().equals(KeyReducer.reduce(k)))
```
Caideyipi, line 323: What if entry.getKey().equals("sink.ip") and k.equals("connector.ip")?

Caideyipi (changes requested):
PTAL

integration-test/src/test/java/org/apache/iotdb/pipe/it/IoTDBPipeAlterIT.java
Caideyipi: BTW, though the progressIndex is preserved on CN, the DN inherit logic still needs to be preserved to retransmit less data... I did not notice that in the last PR...

iotdb-api/pipe-api/src/main/java/org/apache/iotdb/pipe/api/customizer/parameter/PipeParameters.java
```diff
@@ -281,30 +283,15 @@ public String toString() {
...
-                    (oldValue, newValue) -> oldValue, TreeMap::new))
-            .toString(),
-        that.attributes.entrySet().stream()
-            .collect(
-                Collectors.toMap(
-                    entry -> KeyReducer.reduce(entry.getKey()), Entry::getValue,
-                    (oldValue, newValue) -> oldValue, TreeMap::new))
-            .toString());
+  public PipeParameters addOrReplaceAttributes(PipeParameters that) {
+    that.attributes.forEach((k, v) -> this.attributes.put(KeyReducer.reduce(k), v));
```
Caideyipi, line 293: This attributes may have to remove attributes with the same meaning to ensure the attributes from "that" works. Such as, "sink.ip" may have to clear “connector.ip” and "connector.host"(newly supported).

Caideyipi: How to tackle node-urls is tricky and may need some discussion.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/execution/config/executor/ClusterConfigTaskExecutor.java
```diff
@@ -1663,10 +1663,12 @@ public SettableFuture<ConfigTaskResult> alterPipe(AlterPipeStatement alterPipeSt
 
     // Validate before alteration
     try {
-      if (!alterPipeStatement.getProcessorAttributes().isEmpty()) {
+      if (!alterPipeStatement.getProcessorAttributes().isEmpty()
+          && alterPipeStatement.isReplaceAllProcessorAttributes()) {
         PipeAgent.plugin().validateProcessor(alterPipeStatement.getProcessorAttributes());
```
Caideyipi, line 1668: I think it's better to try validate the "modify" as well if the pipeMeta persists in the PipeTaskMeta. If not, you can do nothing here.

iotdb-api/pipe-api/src/main/java/org/apache/iotdb/pipe/api/customizer/parameter/PipeParameters.java
Caideyipi: A reasonable process may be like this:
1. See if reduced old attribute equals to reduced new attribute.
2. If so, remove the old one and add the new one into parameters.

SteveYurongSu (approved):
Greaaaaaaat

PR #12053, opened 2024-02-20
Cache clusterId in IoTDBConfig

liyuheng55555:
IoTDBDescriptor.getInstance().getConfig().getClusterId()

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
@@ -24,14 +24,10 @@ import org.apache.iotdb.commons.conf.CommonDescriptor;
 import org.apache.iotdb.commons.conf.IoTDBConstant;
 import org.apache.iotdb.commons.utils.FileUtils;
 import org.apache.iotdb.commons.utils.TestOnly;
-import org.apache.iotdb.confignode.rpc.thrift.TGetClusterIdResp;
 import org.apache.iotdb.consensus.ConsensusFactory;
 import org.apache.iotdb.db.audit.AuditLogOperation;
 import org.apache.iotdb.db.audit.AuditLogStorage;
 import org.apache.iotdb.db.exception.LoadConfigurationException;
-import org.apache.iotdb.db.protocol.client.ConfigNodeClient;
-import org.apache.iotdb.db.protocol.client.ConfigNodeClientManager;
-import org.apache.iotdb.db.protocol.client.ConfigNodeInfo;
 import org.apache.iotdb.db.protocol.thrift.impl.ClientRPCServiceImpl;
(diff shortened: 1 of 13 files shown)


Caideyipi (approved):
LGTM

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/runtime/PipeRuntimeAgent.java
```diff
@@ -108,22 +105,6 @@ public ServiceType getID() {
     return ServiceType.PIPE_RUNTIME_AGENT;
   }
 
-  public String getClusterIdIfPossible() {
-    if (clusterId.get() == null) {
```
SteveYurongSu, line 112: remove the useless clusterId

liyuheng55555: Had removed, Chief Su check check ?

SteveYurongSu (approved):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
```diff
@@ -3015,6 +3026,24 @@ public void setClusterName(String clusterName) {
     this.clusterName = clusterName;
   }
 
+  public synchronized String getClusterId() throws LoadConfigurationException {
+    if (clusterId.isEmpty()) {
+      try (ConfigNodeClient client =
+          ConfigNodeClientManager.getInstance().borrowClient(ConfigNodeInfo.CONFIG_REGION_ID)) {
```
wangchao316, line 3032: get cluster id when register data node in restart . "sendRestartRequestToConfigNode and 
 sendRegisterRequestToConfigNode"

liyuheng55555: This might slow down the startup speed of the DataNode.
When the DataNode is registering, the cluster id may not yet be ready. For compatibility reason, the cluster id will be checked and generated after the state machine is fully recovered, and asynchronously written into the consensus layer. This process can sometimes be slow.

OneSizeFitsQuorum: I think the ClusterID should be something that a DataNode has when it starts up, and now it seems a bit odd to need to use a try catch every time datanode uses a clusterID

liyuheng55555: now fetch clusterId when dn start and restart

OneSizeFitsQuorum (approved):
LGTM

wangchao316 (approved):
LGTM

OneSizeFitsQuorum (commented):
These codes will be optimized in next PR.

`while (configManager.getConsensusManager() == null) `

PR #12066, opened 2024-02-22
IT: revert some changes in restart IT to make the error log clear when failed to restart

l2280212:
in https://github.com/apache/iotdb/pull/12022.
I catch exceptions during the restart cluster check process, then return a boolean value to indicate the result of the restart. This approach makes the error log unclear when there is a problem.
now revert it

### DIFF
--- a/integration-test/src/main/java/org/apache/iotdb/it/env/cluster/env/AbstractEnv.java
+++ b/integration-test/src/main/java/org/apache/iotdb/it/env/cluster/env/AbstractEnv.java
@@ -280,7 +280,7 @@ public abstract class AbstractEnv implements BaseEnv {
     testJDBCConnection();
   }
   /**
-   * Returns whether the all nodes' status all match the provided predicate. check nodes with RPC.
+   * check whether all nodes' status match the provided predicate with RPC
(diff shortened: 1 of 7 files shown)


integration-test/src/main/java/org/apache/iotdb/it/env/cluster/env/AbstractEnv.java
```diff
@@ -264,29 +264,29 @@ private Map<String, Integer> countNodeStatus(Map<Integer, String> nodeStatus) {
...
+        nodeStatus -> {
+          Map<String, Integer> count = countNodeStatus(nodeStatus);
+          return count.getOrDefault("Unknown", 0) == 1
+              && count.getOrDefault("Running", 0) == nodeStatus.size() - 1;
+        });
+    testJDBCConnection();
   }
   /**
-   * Returns whether the all nodes' status all match the provided predicate. check nodes with RPC
+   * Returns whether the all nodes' status all match the provided predicate. check nodes with RPC.
```
Caideyipi, line 283: “Returns whether” can be deleted

SteveYurongSu (approved):

PR #12074, opened 2024-02-23
[IOTDB-6300] Support place time column at any column index in insert statement

JackieTien97:
In this pr, we enhance the insert statement in two cases:
1. time column can be placed at any index of the column list:
```
insert into root.db.d1(s1,s2,time) values(1,2,3);  
```

2. we can add `` in time column:
```
insert into root.db.d1(`time`,s1,s2) values(1,2,3);  
```


liuminghui233 (approved):
LGTM :-)

PR #11941, opened 2024-01-21
Add the ability for ClientManager to periodically clean up idle objects

Pengzna:
sub task of https://github.com/apache/iotdb/issues/11949

1. Add eviction thread to evict clients that are idle more than 1 minute.
2. Merge `core_client_num` and `max_client_num` params, only remain `max_client_num`
3. Add eviction UT

Detailed docs: https://timechor.feishu.cn/docx/GJYjd1vuYoxQr8xNJejcC7Q6nTc

PR #12075, opened 2024-02-23
Do not wait to retry when configLeader exists

Pengzna:
After introducing https://github.com/apache/iotdb/pull/11941, when the configNode client is newly created and looking for a leader, it is expected to not wait if there is a leader.

Detailed docs：https://apache-iotdb.feishu.cn/docx/XlHkdnkQvohOCwxAduJc0rutnNh


CRZbulabula (approved):
LGTM!

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/client/ConfigNodeClient.java
```diff
@@ -243,6 +226,15 @@ private void tryToConnect() throws TException {
         logger.warn("The current node may have been down {},try next node", configLeader);
         configLeader = null;
       }
+    } else {
+      try {
+        // Wait to start the next try
+        Thread.sleep(RETRY_INTERVAL_MS);
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        throw new TException(
```
liyuheng55555, line 235: Seems not necessary to throw an exception here ?

liyuheng55555 (changes requested):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/client/ConfigNodeClient.java
Pengzna: fixed

liyuheng55555 (approved):
LGTM~

OneSizeFitsQuorum (approved):
LGTM

PR #12081, opened 2024-02-26
Optimize flushing memtable check

HTHou:
## Description

We found the flush submit task would stack a lot when write multiple time partitions concurrently. 

image

The reason of this phenomenon is that the flush submit task check whether is the TsFileProcessor closing in the **WriteLock**  of DataRegion. This check logic should be move out of the WriteLock. 

Moreover, the closingSequenceTsFileProcessor and closingUnSequenceTsFileProcessor can be simplify to the HashSet instead. 

After this modification, the write performance increased about 10%. 

image


shuwenwei (approved):

OneSizeFitsQuorum (commented):
LGTM

PR #12086, opened 2024-02-26
[IOTDB-6302] Enhance the support of ISO_LOCAL_DATE_TIME timestamp format

HTHou:
## Description

IoTDB doesn't support timestamp format like `2024-02-22T16:36:34.67`, which should be equivalent to `2024-02-22T16:36:34.670`.

image

After referred the code of JDK below, we can simply support it.

image

image


JackieTien97 (approved):

OneSizeFitsQuorum (approved):
LGTM

PR #12095, opened 2024-02-28
fix: IoTDBPreparedStatement implement  addBatch function

CritasWang:
1. IoTDBPreparedStatement implement  addBatch function
2. remove JDBC IoTDBConnection commit  "Does not support commit"


HTHou (approved):
LGTM

PR #12101, opened 2024-02-29
Split configuration and add resetPeerList for region migration

BUAAserein:
## Description

Split configuration and add resetPeerList for region migration, see [feishu-doc](https://apache-iotdb.feishu.cn/docx/MC8gdgTgOoG9eYx9MIpcrtB9n4e) for details.

### DIFF
--- a/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/iot/IoTConsensusServerImpl.java
+++ b/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/iot/IoTConsensusServerImpl.java
@@ -605,22 +605,21 @@ public class IoTConsensusServerImpl {
   public void persistConfigurationUpdate() throws ConsensusGroupModifyPeerException {
     try {
       serializeConfigurationAndFsyncToDisk(CONFIGURATION_TMP_FILE_NAME);
-      Path tmpConfigurationPath =
-          Paths.get(new File(storageDir, CONFIGURATION_TMP_FILE_NAME).getAbsolutePath());
-      Path configurationPath =
-          Paths.get(new File(storageDir, CONFIGURATION_FILE_NAME).getAbsolutePath());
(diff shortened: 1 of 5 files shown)


iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/iot/IoTConsensusServerImpl.java
```diff
@@ -635,20 +623,59 @@ public void recoverConfiguration() {
       // interrupted
       // unexpectedly, we need substitute configuration with tmpConfiguration file
       if (Files.exists(tmpConfigurationPath)) {
-        if (Files.exists(configurationPath)) {
-          Files.delete(configurationPath);
-        }
+        Files.deleteIfExists(configurationPath);
         Files.move(tmpConfigurationPath, configurationPath);
       }
       buffer = ByteBuffer.wrap(Files.readAllBytes(configurationPath));
```
liyuheng55555, line 629: After the configuration.dat file is split, seems Files.readAllBytes(configurationPath) will cause `IOException`, then line 637 will catch this exception and handle it with the new `getConfiguration` method.
Thus, there are some other code also throw `IOException`. This makes the code here somewhat unclear.
image
It is hoped to use a clearer method to determine whether the current processing is a new version file or an old version file (for example, whether the CONFIGURATION_FILE_NAME file exists or not).

liyuheng55555 (changes requested):
PTAL~

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/iot/IoTConsensusServerImpl.java
BUAAserein: ok.

liyuheng55555 (approved):
LGTM

PR #12106, opened 2024-03-01
[IOTDB-6305] Add show current_timestamp statement 

yx-zhang:
issue: https://issues.apache.org/jira/browse/IOTDB-6305

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/execution/memory/StatementMemorySourceVisitor.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/execution/memory/StatementMemorySourceVisitor.java
@@ -39,6 +39,7 @@ import org.apache.iotdb.db.queryengine.plan.statement.metadata.CountNodesStateme
 import org.apache.iotdb.db.queryengine.plan.statement.metadata.CountTimeSeriesStatement;
 import org.apache.iotdb.db.queryengine.plan.statement.metadata.ShowChildNodesStatement;
 import org.apache.iotdb.db.queryengine.plan.statement.metadata.ShowChildPathsStatement;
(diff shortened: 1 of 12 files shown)


JackieTien97 (approved):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/execution/memory/StatementMemorySourceVisitor.java
```diff
@@ -259,4 +260,19 @@ public StatementMemorySource visitShowPathsUsingTemplate(
...
+  public StatementMemorySource visitShowCurrentTimestamp(
+      ShowCurrentTimestampStatement showCurrentTimestampStatement,
+      StatementMemorySourceContext context) {
+    List<TSDataType> outputDataTypes =
+        ColumnHeaderConstant.showCurrentTimestampColumnHeaders.stream()
+            .map(ColumnHeader::getColumnType)
+            .collect(Collectors.toList());
+    TsBlockBuilder tsBlockBuilder = new TsBlockBuilder(outputDataTypes);
+    tsBlockBuilder.getTimeColumnBuilder().writeLong(0L);
+    tsBlockBuilder.getColumnBuilder(0).writeLong(System.currentTimeMillis());
```
HTHou, line 273: Is there any issue when the timestamp_precision is ns or us? @JackieTien97

JackieTien97: It's a long value, how to resolve this long value depends on the user. We can default return `ms`, in stardand sql, there will be a parameter called `precision` in current_timestamp function to decide whether to return `ms`, `us` or `ns`, so I think it's ok to defaultly return `ms` here and this function call's time precision should have nothing to do with our `timestamp_precision` configuration.

Beyyes (approved):
LGTM

HTHou (approved):

PR #12113, opened 2024-03-03
invert insertRecords to insertTablets

l2280212:
This PR will switch from using insertRecords to insertTablets when the number of devices involved in the records is low.

The effect when converting to a tablet and not needing to add null values is as follows.
image
The effect when converting to a tablet and needing to add null values is as follows.
image

the complete test report https://timechor.feishu.cn/docx/RxkadMOuxoEUZNx98YxcrZEYnge


iotdb-client/session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -2737,6 +2774,78 @@ private void updateTSInsertTabletsReq(
...
+          deviceIds.stream()
+              .collect(Collectors.groupingBy(e -> e, Collectors.counting()))
+              .values()
+              .stream()
+              .max(Long::compare)
+              .orElse((long) deviceIds.size());
+      for (int i = 0; i < deviceIds.size(); i++) {
+        String deviceId = deviceIds.get(i);
+        if (tablets.containsKey(deviceId)) {
+          if (measurementsList.get(i) == deviceMeasurement.get(deviceId)) {
```
THUMarkLau, line 2805: Use `equals` instead of `==` here.

iotdb-client/session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -2737,6 +2774,78 @@ private void updateTSInsertTabletsReq(
...
 
+  public void invertToTabletsAndInsert(
+      List<String> deviceIds,
+      List<Long> times,
+      List<List<String>> measurementsList,
+      List<List<TSDataType>> typesList,
+      List<List<Object>> valuesList,
+      boolean isAligned)
+      throws IoTDBConnectionException, StatementExecutionException {
+    while (!deviceIds.isEmpty()) {
```
THUMarkLau, line 2785: Even if all the records are about the same measurement under the same device, if the order of measurements in each record is different, it could lead to one set of records being split into multiple tablets for transmission. In conditions of average network quality, the multiple RPC delays could result in high latency and potentially poor performance. I lean towards placing all records of a device into one tablet to complete the transmission in a single step.

THUMarkLau (changes requested):

OneSizeFitsQuorum (commented):
any other insertRecords interface?

HTHou:
> any other insertRecords interface?

Yes, I think insertRecordsOfOneDevice can also be transformed to InsertTablet.

iotdb-client/session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -2737,6 +2755,85 @@ private void updateTSInsertTabletsReq(
...
+  public void addRecordToTablet(
+      Tablet tablet,
+      Long timestamp,
+      List<String> measurements,
+      List<Object> values,
+      List<String> allMeasurements) {
+    int row = tablet.rowSize++;
+    tablet.addTimestamp(row, timestamp);
+    Map<String, Object> measurementValueMap = new HashMap<>();
+    for (int i = 0; i < measurements.size(); i++) {
```
THUMarkLau, line 2828: Do we need to convert the data into a Map before adding it to the Tablet? Perhaps it can be added directly to the Tablet because the Tablet defaults to filling in nulls.

THUMarkLau (changes requested):

iotdb-client/session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -2737,6 +2706,143 @@ private void updateTSInsertTabletsReq(
...
+    tablet.addTimestamp(row, timestamp);
+    // tablet without null value
+    if (measurements.size() == allMeasurements.size()) {
+      for (int i = 0; i < measurements.size(); i++) {
+        tablet.addValue(measurements.get(i), row, values.get(i));
+      }
+      return;
+    }
+    // tablet with null value
+    Map<String, Object> measurementValueMap = new HashMap<>(measurements.size() + 1, 1);
```
OneSizeFitsQuorum, line 2836: optimize

iotdb-client/session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -2737,6 +2706,143 @@ private void updateTSInsertTabletsReq(
...
+      List<TSDataType> types = typesList.get(rowIndex);
+      for (int colIndex = 0; colIndex < measurements.size(); colIndex++) {
+        if (!measurementType.containsKey(measurements.get(colIndex))) {
+          measurementType.put(measurements.get(colIndex), types.get(colIndex));
+        }
+      }
+      rowMap.merge(device, 1, Integer::sum);
+    }
+    // device -> schema
+    Map<String, List<MeasurementSchema>> schemaMap = new HashMap<>(deviceSize + 1, 1);
```
OneSizeFitsQuorum, line 2783: optimized

OneSizeFitsQuorum (approved):
LGTM

iotdb-client/session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -2744,6 +2780,209 @@ private void updateTSInsertTabletsReq(
     request.addToSizeList(tablet.rowSize);
   }
 
+  // sample some records and judge weather need to add too many null values to convert to tablet.
+  public boolean judgeConvertOfOneDevice(List<List<String>> measurementsList) {
+    int size = measurementsList.size();
+    int sampleNum = (int) (size * SAMPLE_PROPORTION);
+    if (sampleNum < MIN_SAMPLE_SIZE) {
```
THUMarkLau, line 2787: should be MIN_RECORDS_NUM，comparing with the original size.

THUMarkLau (changes requested):

OneSizeFitsQuorum (approved):
LGTM

THUMarkLau (approved):

PR #12132, opened 2024-03-06
Fix fail to read memTable ids from the wal file

HTHou:
## Description

When the walDeleteThread and a flush command delete the outdated wal file concurrently, the following error may appear.
img_v3_028l_e1c4b5ec-c709-4536-91f7-c206f2ab805g

To fix it, the flush command thread should submit the wal delete task to walDeleteThread instead of executing it by itself.


HeimingZ (approved):

OneSizeFitsQuorum (approved):
LGTM

PR #12143, opened 2024-03-08
Fixed ConfigNode startup error in SimpleConsensus

OneSizeFitsQuorum:
SimpleConsensus triggers the start and notifyLeaderReady functions of SimpleConsensusImpl before the consensusGroup in the map has been written. In the initialization logic of some asynchronous calls, if it is executed fast enough, SimpleConsensus may still have a map size of 0, causing an error to fail startup.

The fix is as simple as calling start after a successful put

### DIFF
--- a/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/simple/SimpleConsensus.java
+++ b/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/simple/SimpleConsensus.java
@@ -168,11 +168,13 @@ class SimpleConsensus implements IConsensus {
                     return null;
                   }
 
-                  SimpleConsensusServerImpl impl =
-                      new SimpleConsensusServerImpl(peers.get(0), registry.apply(groupId));
-                  impl.start();
-                  return impl;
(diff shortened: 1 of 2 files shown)


SteveYurongSu (approved):

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/simple/SimpleConsensus.java
```diff
@@ -168,11 +168,13 @@ public void createLocalPeer(ConsensusGroupId groupId, List<Peer> peers)
...
 
-                  SimpleConsensusServerImpl impl =
-                      new SimpleConsensusServerImpl(peers.get(0), registry.apply(groupId));
-                  impl.start();
-                  return impl;
+                  return new SimpleConsensusServerImpl(peers.get(0), registry.apply(groupId));
                 }))
+        .map(
+            impl -> {
+              impl.start();
```
liyuheng55555, line 175: Does IoTConsensus have similar problem ?
image

PR #12145, opened 2024-03-08
Support schema region snapshot parser.

shuolinl:
In this pr:
1. Support translating snapshot file to statements, so you can use schema region snapshot parser to construct a same Schema tree in another cluster.
2. Add some UTs.
3. refine memory use when translating schema region snapshot. The node num depends on the longest path and Its children num. If one path is root.t1.t2.t3, and It has 100 measurements, so the node num will be 4 + 100.

152201eb8e7e6b2425a492d1c8c8c26

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/tools/schema/SchemaRegionSnapshotParser.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/tools/schema/SchemaRegionSnapshotParser.java
@@ -1,0 +1,175 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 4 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/tools/schema/SchemaRegionSnapshotParser.java
```diff
@@ -0,0 +1,501 @@
...
+
+import static org.apache.iotdb.commons.schema.SchemaConstant.ENTITY_MNODE_TYPE;
+import static org.apache.iotdb.commons.schema.SchemaConstant.INTERNAL_MNODE_TYPE;
+import static org.apache.iotdb.commons.schema.SchemaConstant.LOGICAL_VIEW_MNODE_TYPE;
+import static org.apache.iotdb.commons.schema.SchemaConstant.MEASUREMENT_MNODE_TYPE;
+import static org.apache.iotdb.commons.schema.SchemaConstant.STORAGE_GROUP_ENTITY_MNODE_TYPE;
+import static org.apache.iotdb.commons.schema.SchemaConstant.STORAGE_GROUP_MNODE_TYPE;
+import static org.apache.iotdb.commons.schema.SchemaConstant.isStorageGroupType;
+
+public class SchemaRegionSnapshotParser {
```
Cpaulyz, line 72:
```suggestion
/**
 * Only one thread can be parsing snapshots at a time.
 */
@NotThreadSafe
public class SchemaRegionSnapshotParser {
```

Cpaulyz (changes requested):
PTAL

Cpaulyz (approved):
LGTM

shuolinl:
@SteveYurongSu Please take a look at this pr.

SteveYurongSu (approved):

PR #12146, opened 2024-03-09
UDF: add an udf function for envelope demodulation analysis

caozj1011:
## Description

This PR adds a UDF function for envelope demodulation analysis. The function of envelope demodulation analysis is designed to realize the demodulation and envelope extraction of the signal by inputting one-dimensional floating-point numbers and the modulation frequency specified by the user. The goal of demodulation is to extract parts of interest from complex signals and make them easier to understand. For example, demodulation can find the envelope of the signal, that is, the trend of amplitude change. The realization principle is to convert the time domain signal into the frequency domain signal, and the algorithms involved include Hilbert transform (Hilbert), fast Fourier transform (FFT), fast Fourier inverse transform (IFFT) and so on.

### DIFF
--- a/library-udf/src/main/java/org/apache/iotdb/library/frequency/UDFEnvelopeAnalysis.java
+++ b/library-udf/src/main/java/org/apache/iotdb/library/frequency/UDFEnvelopeAnalysis.java
@@ -40,7 +40,7 @@ import java.util.List;
 public class UDFEnvelopeAnalysis implements UDTF {
   private int frequency;
 
-  private final List<Double> list = new ArrayList<>();
+  private final List<Double> signals = new ArrayList<>();
 
   public static final String frequencyConstant = "frequency";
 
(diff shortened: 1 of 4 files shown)


caozj1011 (commented):
self review done.

caozj1011 (commented):
self review done! really!

library-udf/src/main/java/org/apache/iotdb/library/frequency/UDFEnvelopeAnalysis.java
```diff
@@ -0,0 +1,189 @@
...
+            x -> validator.getParameters().getAttributes().size() == 1,
+            "The 'envelope' function takes only 'frequency' as an argument.",
+            validator.getParameters());
+  }
+
+  @Override
+  public void beforeStart(UDFParameters parameters, UDTFConfigurations configurations)
+      throws Exception {
+    configurations.setAccessStrategy(new RowByRowAccessStrategy()).setOutputDataType(Type.DOUBLE);
+    this.frequency = Integer.parseInt(parameters.getStringOrDefault(frequencyConstant, "0"));
```
SteveYurongSu, line 63: Using getIntOrDefault instead

library-udf/src/main/java/org/apache/iotdb/library/frequency/UDFEnvelopeAnalysis.java
```diff
@@ -0,0 +1,189 @@
...
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class UDFEnvelopeAnalysis implements UDTF {
+  private int frequency;
+
+  private final List<Double> signals = new ArrayList<>();
```
SteveYurongSu, line 43: org.eclipse.collections.impl.list.mutable.primitive.DoubleArrayList can be a better choice for performance.

caozj1011 (commented):
self review

library-udf/src/main/java/org/apache/iotdb/library/frequency/UDFEnvelopeAnalysis.java
```diff
@@ -0,0 +1,248 @@
...
+        .validateInputSeriesNumber(1)
+        .validateInputSeriesDataType(0, Type.DOUBLE, Type.FLOAT, Type.INT32, Type.INT64)
+        .validate(
+            x -> (double) x >= 0,
+            "The param 'frequency' must greater than 0.",
+            validator.getParameters().getDoubleOrDefault(FREQUENCY, 0))
+        .validate(
+            x -> (int) x >= 1,
+            "The param 'amplification' must greater than 1.",
+            validator.getParameters().getIntOrDefault(AMPLIFICATION, 1));
```
SteveYurongSu, line 67:
```suggestion
        .validate(
            x -> (double) x > 0,
            "The param 'frequency' must be > 0.",
            validator.getParameters().getDoubleOrDefault(FREQUENCY, 0))
        .validate(
            x -> (int) x >= 1,
            "The param 'amplification' must be >= 1.",
            validator.getParameters().getIntOrDefault(AMPLIFICATION, 1));
```

SteveYurongSu (approved):
LGTM

PR #12155, opened 2024-03-12
Print native request api in datanode_slow_sql.log

Beyyes:
## Description

print example:

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/thrift/impl/ClientRPCServiceImpl.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/thrift/impl/ClientRPCServiceImpl.java
@@ -354,7 +354,7 @@ public class ClientRPCServiceImpl implements IClientRPCServiceWithHandler {
         // record total time cost for one query
         long executionTime = COORDINATOR.getTotalExecutionTime(queryId);
         addQueryLatency(statementType, executionTime > 0 ? executionTime : currentOperationCost);
-        COORDINATOR.cleanupQueryExecution(queryId, t);
+        COORDINATOR.cleanupQueryExecution(queryId, req, t);
       }
       SESSION_MANAGER.updateIdleTime();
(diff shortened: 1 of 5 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/Coordinator.java
```diff
@@ -235,19 +242,81 @@ public void cleanupQueryExecution(Long queryId, Throwable t) {
...
+      TSFetchResultsReq req = (TSFetchResultsReq) request;
+      StringBuilder sb = new StringBuilder();
+      for (int i = 0;
+          i < Math.min(queryExecution.getOutputValueColumnCount(), MAX_SLOW_NATIVE_API_OUTPUT_NUM);
+          i++) {
+        sb.append(i == 0 ? "" : ",")
+            .append(queryExecution.getDatasetHeader().getRespColumns().get(i));
+      }
+      slowContent =
+          String.format(
```
JackieTien97, line 310: total size.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/thrift/impl/ClientRPCServiceImpl.java
```diff
@@ -1103,7 +1103,7 @@ public TSFetchResultsResp fetchResultsV2(TSFetchResultsReq req) {
         long executionTime = COORDINATOR.getTotalExecutionTime(req.queryId);
         addQueryLatency(
             StatementType.QUERY, executionTime > 0 ? executionTime : currentOperationCost);
-        COORDINATOR.cleanupQueryExecution(req.queryId, t);
+        COORDINATOR.cleanupQueryExecution(req.queryId, req, t);
```
JackieTien97, line 1106: also consider the catch exception block

JackieTien97 (changes requested):

PR #12160, opened 2024-03-12
[IOTDB-6310] Optimize for query resource init

JackieTien97:
we use timefilter to filter useless timepartition instead of getting all of them and then filter tsfile in them one by one.

We insert `0~9999` points whose timestamp is also `0~9999` and execute `flush` after each point insert and then insert one point with timestamp now(). As so, we will get two partitions, one contains 10,000 tsfiles, another one contains only one tsfile.

execute the following sql: (only the last tsfile satisfies)
```
select s1 from root.db.d1 where time > now() - 1h;
```

before opt, the time for query init tsfile list are:(about 2.5ms)

```
3057000
2356292
3258834
2415417
2876041
2395750
2564833
```

after opt, the time for query init tsfile list are:(about 0.047 ms)
```
47500
39792
33875
31958
37584
56666
45709
```


liuminghui233 (approved):
LGTM :-)

PR #12176, opened 2024-03-15
Fix wal delete sort method

HeimingZ:
Because the disk usage of wal nodes may change when comparing their delete priority, which violates comparison method's  general contract. So, I cache the disk usage in a map and use the values in the map to sort wal nodes.


HTHou (approved):

OneSizeFitsQuorum (approved):
LGTM

PR #12195, opened 2024-03-18
feat: Refactored the visibility of some methods to make it possible to add BasicAuthorizer implementations in other packages.

chrisdutz:
The visibility of some methods prevented being able to implement custom authenticators outside of the org.apache.iotdb.commons.auth.authorizer package.

This should now be fixed.


wangchao316 (approved):
GOOD, this is easy to extends BasicAuthorizer and implements methods.

PR #12196, opened 2024-03-19
Fix NPE in SessionConnection while first node in nodeurls is unavailable

JackieTien97:
more details can be seen in https://jira.infra.timecho.com:8443/browse/TIMECHODB-611.

If a `SessionConnection`'s first node in its nodeurls is unavailable and it will try to reconnect to other nodes in nodeurls, and it will succeed to connect to other nodes, however currently its `endPointToSessionConnection` and `deviceIdToEndpoint` are not inited, so the logic in reconnect may cause NPE.


Beyyes (approved):
LGTM

PR #12205, opened 2024-03-20
Add Query Metric for TsFileResourceManager

JackieTien97:
Add total resource number, degraded resource number, max index memory size and used index memory size metrics.
image


Beyyes (approved):
LGTM

PR #12209, opened 2024-03-21
[IOTDB-6301] Optimize insert first for single replica

BUAAserein:
## Description
Force update region leader for ratis consensus when Ratis replication factor is 1.


OneSizeFitsQuorum (approved):
LGTM

PR #12220, opened 2024-03-22
Ensure the DataNode can obtain the complete list of ConfigNodes

liyuheng55555:
Over the past period, some initialization work has been brought forward to the DataNode startup phase, which has slightly slowed down the startup speed of the DataNode, and in turn, increased the time between DataNode registration and the start of the RPC service.

Now, if the ConfigNode and DataNode are started simultaneously, a fault may occur where the DataNode cannot obtain a complete list of ConfigNodes.

The occurrence of this fault requires a specific sequence of events:

- When the DataNode registration returns, because the ConfigNode has not completed registration, the DataNode does not get a complete list of ConfigNodes.
- After the ConfigNode registration is complete, it fails to send the complete list to the DataNode via RPC because the DataNode RPC service has not yet started.

A temporary solution to this problem is to move the classLoader() function somewhere after RPC service starting.

By the way, the classLoader function :
```java
  private void classLoader() {
    try {
      // StatementGenerator
      Class.forName(StatementGenerator.class.getName());
      Class.forName(ASTVisitor.class.getName());
      Class.forName(SqlLexer.class.getName());
      Class.forName(CommonTokenStream.class.getName());
      Class.forName(IoTDBSqlParser.class.getName());
      // SourceRewriter
      Class.forName(SourceRewriter.class.getName());
      Class.forName(DistributionPlanContext.class.getName());
      // LogicalPlaner
      Class.forName(LogicalPlanVisitor.class.getName());
      Class.forName(LogicalQueryPlan.class.getName());
      // TsFileProcessor
      Class.forName(TsFileProcessor.class.getName());
    } catch (ClassNotFoundException e) {
      logger.error("load class error: ", e);
    }
  }
```

### DIFF
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/client/async/AsyncDataNodeClientPool.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/client/async/AsyncDataNodeClientPool.java
@@ -163,6 +163,13 @@ public class AsyncDataNodeClientPool {
       if (clientHandler.getRequestIndices().isEmpty()) {
         return;
       }
+
+      try {
+        Thread.sleep(1000);
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        LOGGER.warn("Sleep was interrupted");
+      }
     }
 
     if (!clientHandler.getRequestIndices().isEmpty()) {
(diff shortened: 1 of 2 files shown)


BUAAserein (approved):
LGTM

OneSizeFitsQuorum (commented):
Better to resolve it thorough, or the code looks even uglier

PR #12221, opened 2024-03-22
Refactor MultiTsFileDeviceIterator.MeasurementIterator

shuwenwei:
## Description
Refactor MultiTsFileDeviceIterator.MeasurementIterator


iotdb-core/datanode/src/test/java/org/apache/iotdb/db/storageengine/dataregion/compaction/utils/MultiTsFileDeviceIteratorTest.java
```diff
@@ -82,6 +90,70 @@ public void tearDown() throws IOException, StorageEngineException {
...
+            CompressionType.LZ4);
+      }
+      writer.endChunkGroup();
+      writer.endFile();
+    }
+    resource2.setStatusForTest(TsFileResourceStatus.COMPACTING);
+
+    seqResources.add(resource2);
+
+    Set<String> measurementSet = new HashSet<>(4000);
```
choubenson, line 129: Change it to a list type so that we can detect whether there are unmerged series. Otherwise, this UT (unit test) can still pass as before.

shuwenwei: Resolved

PR #12227, opened 2024-03-23
Optimize SeriesScanUtil by memorizing the order time and satisfied information for each Seq and Unseq Resource

JackieTien97:
We can memorize the order time and satisfied information for each Seq and Unseq Resource instead of getting it from resource ITimeIndex map each time, in which way we can save much time by avoiding time consuming map.get();


liuminghui233 (approved):
LGTM :-)

PR #12232, opened 2024-03-25
Use heartbeat to update DataNode's ConfigNodeList

liyuheng55555:
Over the past period, some initialization work has been brought forward to the DataNode startup phase, which has slightly slowed down the startup speed of the DataNode, and in turn, increased the time between DataNode registration and the start of the RPC service.

Now, if the ConfigNode and DataNode are started simultaneously, a fault may occur where the DataNode cannot obtain a complete list of ConfigNodes.

The occurrence of this fault requires a specific sequence of events:

- When the DataNode registration returns, because the ConfigNode has not completed registration, the DataNode does not get a complete list of ConfigNodes.
- After the ConfigNode registration is complete, it fails to send the complete list to the DataNode via RPC because the DataNode RPC service has not yet started.

To solve this problem, we will use heartbeat to transit ConfigNodes' list. The heartbeat will always carry ConfigNodes' list, until the DataNode confirms this list.

*This pr may slightly increase the latency of ConfigNode list synchronizing (less than 1s). We believe this will do no harm because there doesn't seem to be anything that requires an immediate notification to the DataNode after a ConfigNode's registration.*


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/load/service/HeartbeatService.java
```diff
@@ -150,6 +153,17 @@ private TDataNodeHeartbeatReq genHeartbeatReq() {
     return heartbeatReq;
   }
 
+  private void addConfigNodeLocationsToReq(int dataNodeId, TDataNodeHeartbeatReq req) {
+    Set<TConfigNodeLocation> confirmedConfigNodes =
+        loadCache.getConfirmedConfigNodeLocations(dataNodeId);
+    Set<TConfigNodeLocation> actualConfigNodes =
+        new HashSet<>(getNodeManager().getRegisteredConfigNodes());
+    if (!actualConfigNodes.equals(confirmedConfigNodes)
```
OneSizeFitsQuorum, line 161: always carry this list?

iotdb-protocol/thrift-datanode/src/main/thrift/datanode.thrift
```diff
@@ -252,6 +252,7 @@ struct TDataNodeHeartbeatReq {
   8: optional bool needPipeMetaList
   9: optional i64 deviceQuotaRemain
   10: optional TDataNodeActivation activation
+  11: optional set<common.TConfigNodeLocation> configNodeLocations
```
OneSizeFitsQuorum, line 255: use id is enough?

liyuheng55555: use TEndPoint

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/client/DataNodeRequestType.java
```diff
@@ -32,7 +32,6 @@ public enum DataNodeRequestType {
   STOP_REPAIR_DATA,
   LOAD_CONFIGURATION,
   SET_SYSTEM_STATUS,
-  BROADCAST_LATEST_CONFIG_NODE_GROUP,
```
CRZbulabula, line 35: We could keep the BROADCAST_LATEST_CONFIG_NODE_GROUP interface, since it's never too much to broadcast immediately when the cluster ConfigNode group has been updated. Therefore the services that rely on ConfigNodeList will save the 1s heartbeat interval time. The heartbeat function should be took as a security method to ensure the DataNodes can update their ConfigNode list when BROADCAST_LATEST_CONFIG_NODE_GROUP failed.

CRZbulabula (changes requested):
PTAL~

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/client/DataNodeRequestType.java
liyuheng55555: This is indeed worth considering, and we can have a discussion. I believe there is no need to retain the original RPC notification mechanism here, as the ConfigNode list is not time-sensitive. After a ConfigNode is added, there doesn't seem to be anything that requires an immediate response from the DataNode.

CRZbulabula (approved):
LGTM!

OneSizeFitsQuorum (approved):
LGTM

PR #12234, opened 2024-03-25
[IOTDB-6301] wait leader ready in single replica

BUAAserein:
## Description
In single replica scenario, Ratis leader will be forcibly updated when a region is created, but writing needs to wait for the leader to take effect.


SzyWilliam (approved):
+1 the change looks good. Thanks @BUAAserein for the patch.

OneSizeFitsQuorum (approved):
LGTM

PR #12239, opened 2024-03-26
Remove -XX:-UseBiasedLock in IoTDB JVM options

Pengzna:
We have observed performance losses after adding `-XX:-UseBiasedLock` in some cases. Thus we remove it.

Detailed docs: https://apache-iotdb.feishu.cn/docx/JNvTdo0oOoOJr7xq3QecRe9Wnhc


OneSizeFitsQuorum (approved):
LGTM

PR #12240, opened 2024-03-27
fix: consider more general type of nodes in optimization above LimitNode

ycycse:
In EXPLAIN ANALYZE statement, there will be an IdentityNode above LimitNode to transfer result to ExplainAnalyzeOperator.
However, the implementation in LimitOffsetPushDown considers all the node above is SingleChildNode (IdentityNode is MultiChildSinkNode) which leads to an error: 
`Msg: org.apache.iotdb.jdbc.IoTDBSQLException: 305: [INTERNAL_SERVER_ERROR(305)] Exception occurred: "explain analyze select s1 from root.test.sg limit 2". executeStatement failed. class org.apache.iotdb.db.queryengine.plan.planner.plan.node.sink.IdentitySinkNode cannot be cast to class org.apache.iotdb.db.queryengine.plan.planner.plan.node.process.SingleChildProcessNode (org.apache.iotdb.db.queryengine.plan.planner.plan.node.sink.IdentitySinkNode and org.apache.iotdb.db.queryengine.plan.planner.plan.node.process.SingleChildProcessNode are in unnamed module of loader 'app')
`
This PR fix this problem.


liuminghui233 (approved):
LGTM :-)

PR #12241, opened 2024-03-27
Fix C++ client build problem

liyuheng55555:
For more information, visit https://thrift.apache.org/docs/types and https://apache-iotdb.feishu.cn/docx/doxcnK48Vydl3oYRVwSC9RwS3Wg

> N.B.: For maximal compatibility, the key type for map should be a basic type rather than a struct or container type. There are some languages which do not support more complex key types in their native map types. In addition the JSON protocol only supports key types that are base types.


OneSizeFitsQuorum (approved):
LGTM

HTHou (approved):

PR #12243, opened 2024-03-27
IoTConsensus: Skip retry sending batch caused by TApplicationException

BUAAserein:
## Description

When the wal file is damaged, it will cause follower synchronization failure, causing leader synchronization to get stuck and infinite retries until wal blocks writing. The fix is to skip retrying send batches caused by TApplicationException and print a warn log.


OneSizeFitsQuorum (approved):
LGTM

PR #12255, opened 2024-03-29
Modify compaction task priority comparator

shuwenwei:
## Description
Modify compaction task priority comparator. When compare max file version of compaction task, the data region and time partition should be same.


HTHou (approved):

OneSizeFitsQuorum (approved):
LGTM

PR #12291, opened 2024-04-03
Use UTF-8 to load and store properties file

liyuheng55555:
Currently, IoTDB uses the ISO/IEC 8859-1 encoding to read and write properties files. This encoding, which is the default for Java to handle properties files, does not support Chinese characters. In order to support Chinese characters, IoTDB will switch to using UTF-8 for all read and write operations on properties files.

After this modification, the cluster name can be named in Chinese.

image

*Although a portion of letters in ISO/IEC 8859-1 cannot be automatically parsed by UTF-8 (indicated by the red box), we believe these characters are very rarely used. 
For common English characters (indicated by the green box), UTF-8 is fully compatible.*

image


OneSizeFitsQuorum (approved):
LGTM

PR #12297, opened 2024-04-06
Storage Engine: Fixed inaccurate compression ratio when writing duplicate data

caozj1011:
## Description
This PR fixed inaccurate compression ratio when writing duplicate data,

## Before this PR
The memory size may be reduced to negative values in aligned sequence scenarios and when writing duplicate data

## Before this PR
In the aligned sequence scenario and when writing duplicate data, the memory size will not become negative again


SteveYurongSu:
@HTHou PTAL

PR #12299, opened 2024-04-07
fix delete dead lock.

shuolinl:
The current system has a deadlock problem when the deletion is concurrent with query and insertion:
1. Query worker 's num is limit as core num.
2. Deletion will acquire Dataregion's write lock and Datanode Schemacache's write lock.
3. Insert or query will request additional queries to obtain metadata in the case of acquiring a read lock.

DeadLock cycle like this:
```mermaid
graph BT
A[#QueryWorker# get one core to run query]
B[#QueryWorker# hold read lock on dataregion]
A-->B
C[#DeleteData# hold write lock on dataregion]
D[#DeleteData# hold write lock on datanode schema]
C-->D
E[#InsertData# holdread lock on datanode schema]
F[#InsertData# get one core to run fetch schema]
E-->F
B-->C
F-->A
D-->E
```

In this pr:
1. Optimize the lock order to prevent deadlocks
5. Add invalidate last cache operation.  The deletion operation does not need to clean up all caches, but should clean the corresponding last cache. 
6.  You can use the following patch to reproduce the deadlock problem.
[0001-to-cause-dead-lock-when-delete.patch](https://github.com/apache/iotdb/files/14900114/0001-to-cause-dead-lock-when-delete.patch)
In this patch, datanodeCache will always be empty, so query or insert will fetch remote schema. Besides the query worker's num is limit to 1 and the program will wait about 20s/30s after acquiring the critical lock.
To reproduce the problem, run these three queries in three sessions within 30s:
session1:  `insert into root.sg1.t1(t) values(1);`
session2: ` delete from root.sg1.t1.**;`
session3: `select * from root.sg1.**;`

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/analyze/cache/schema/dualkeycache/impl/DualKeyCacheImpl.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/analyze/cache/schema/dualkeycache/impl/DualKeyCacheImpl.java
@@ -22,10 +22,12 @@ package org.apache.iotdb.db.queryengine.plan.analyze.cache.schema.dualkeycache.i
 import org.apache.iotdb.commons.path.PartialPath;
 import org.apache.iotdb.commons.path.PathPatternUtil;
 import org.apache.iotdb.commons.utils.TestOnly;
+import org.apache.iotdb.db.queryengine.plan.analyze.cache.schema.SchemaCacheEntry;
 import org.apache.iotdb.db.queryengine.plan.analyze.cache.schema.dualkeycache.IDualKeyCache;
 import org.apache.iotdb.db.queryengine.plan.analyze.cache.schema.dualkeycache.IDualKeyCacheComputation;
 import org.apache.iotdb.db.queryengine.plan.analyze.cache.schema.dualkeycache.IDualKeyCacheStats;
 import org.apache.iotdb.db.queryengine.plan.analyze.cache.schema.dualkeycache.IDualKeyCacheUpdating;
+import org.apache.iotdb.db.queryengine.plan.analyze.cache.schema.lastcache.DataNodeLastCacheManager;
 
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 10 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/analyze/cache/schema/lastcache/ILastCacheContainer.java
```diff
@@ -38,5 +38,8 @@ public interface ILastCacheContainer {
   int updateCachedLast(
       TimeValuePair timeValuePair, boolean highPriorityUpdate, Long latestFlushedTime);
 
+  /** Invalidate Last cache. */
+  void invalidateLastCache();
```
JackieTien97, line 42:
```suggestion
  int invalidateLastCache();
```

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/analyze/cache/schema/lastcache/DataNodeLastCacheManager.java
```diff
@@ -65,4 +65,11 @@ public static int updateLastCache(
     }
     return entry.updateLastCache(timeValuePair, highPriorityUpdate, latestFlushedTime);
   }
+
+  public static void invalidateLastCache(SchemaCacheEntry entry) {
```
JackieTien97, line 69:
```suggestion
  public static int invalidateLastCache(SchemaCacheEntry entry) {
```

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/analyze/cache/schema/dualkeycache/impl/DualKeyCacheImpl.java
```diff
@@ -240,6 +243,71 @@ private int evictOneCacheEntry() {
...
+        for (Iterator<Map.Entry<SK, T>> it = entryGroup.getAllCacheEntries(); it.hasNext(); ) {
+          Map.Entry<SK, T> entry = it.next();
+          if (Boolean.TRUE.equals(measurementFilter.apply(entry.getKey()))) {
+            synchronized (entry) {
+              SchemaCacheEntry cacheEntry = (SchemaCacheEntry) entry.getValue().getValue();
+              if (cacheEntry.getLastCacheContainer().getCachedLast() != null) {
+                cacheStats.decreaseMemoryUsage(
+                    cacheEntry.getLastCacheContainer().getCachedLast().getSize());
+              }
+              DataNodeLastCacheManager.invalidateLastCache(cacheEntry);
```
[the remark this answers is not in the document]
JackieTien97: and also what if cacheEntry.getLastCacheContainer() is null, need to consider this in your new `DataNodeLastCacheManager.invalidateLastCache` implementation

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/analyze/cache/schema/dualkeycache/impl/DualKeyCacheImpl.java
```diff
@@ -240,6 +243,71 @@ private int evictOneCacheEntry() {
...
+      if (device.toString().startsWith(database)) {
+        ICacheEntryGroup<FK, SK, V, T> entryGroup = firstKeyMap.get(device);
+        for (Iterator<Map.Entry<SK, T>> it = entryGroup.getAllCacheEntries(); it.hasNext(); ) {
+          Map.Entry<SK, T> entry = it.next();
+          SchemaCacheEntry cacheEntry = (SchemaCacheEntry) entry.getValue().getValue();
+          if (cacheEntry.getLastCacheContainer().getCachedLast() != null) {
+            cacheStats.decreaseMemoryUsage(
+                cacheEntry.getLastCacheContainer().getCachedLast().getSize());
+          }
+          DataNodeLastCacheManager.invalidateLastCache(cacheEntry);
```
JackieTien97, line 305: same as above

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/write/DeleteDataNode.java
```diff
@@ -60,6 +60,8 @@ public class DeleteDataNode extends WritePlanNode implements WALEntryValue {
   private final long deleteStartTime;
   private final long deleteEndTime;
 
+  private final boolean isDeleteTimeseries;
```
JackieTien97, line 63: add some comments here to explain why this field doesn't need to be serialized (because it will always be false except for that in delete timesereis, and in which case we will set this to true in `DataNodeInternalRPCServiceImpl.deleteDataForDeleteSchema`.

```suggestion
  private final transient boolean isDeleteTimeseries;
```

JackieTien97 (changes requested):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/DataRegion.java
```diff
@@ -2539,7 +2540,7 @@ public int compact() {
     }
   }
 
-  private void resetLastCacheWhenLoadingTsFile() throws IllegalPathException {
+  private void resetLastCacheWhenLoadingTsFile() {
```
JackieTien97, line 2543: just call `DataNodeSchemaCache.getInstance().unsetLastCacheInDataRegion(getDatabaseName())`?

JackieTien97 (changes requested):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/analyze/cache/schema/dualkeycache/impl/DualKeyCacheImpl.java
```diff
@@ -240,6 +242,63 @@ private int evictOneCacheEntry() {
...
+    Function<SK, Boolean> measurementFilter = null;
+
+    if (PathPatternUtil.hasWildcard(devicePath.getFullPath())) {
+      deviceFilter = d -> devicePath.matchFullPath((PartialPath) d);
+    }
+    if (PathPatternUtil.isMultiLevelMatchWildcard(measurement)) {
+      measurementFilter = m -> true;
+    }
+    if (deviceFilter == null) {
+      deviceFilter = d -> ((PartialPath) d).equals(devicePath.getFullPath());
```
JackieTien97, line 259:
```suggestion
      deviceFilter = d -> ((PartialPath) d).equals(devicePath);
```

JackieTien97 (changes requested):

JackieTien97 (approved):

PR #12304, opened 2024-04-09
[RatisConsensus] Notify term index change when configuration changes

SzyWilliam:
Currently the statemachine's apply index is not updated for configuration change logs.
This may block the readIndex since the readIndex will compare the applyIndex and currentIndex. During unstable clusters where configuration logs are not followed by statemachine logs, the readIndex may never return.


OneSizeFitsQuorum (changes requested):
image
image
Looks like it's been updated?

SzyWilliam:
@OneSizeFitsQuorum You are right. The term and index are already updated. Thanks for finding it out!

PR #12311, opened 2024-04-10
Do not set read only in compaction module

shuwenwei:
## Description
1. Do not set read only in compaction module.
2. No longer set allowCompaction to false though the compaction exception recovery is failed.


OneSizeFitsQuorum (approved):
LGTM

PR #12315, opened 2024-04-10
Pipe: clear event reference count instead of decreasing when closing to improve idempotence & add creation time for committer key to avoid outdated commit & detect re-increasing reference count to avoid NPE

VGalaxies:
As title.

### DIFF
--- a/iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/progress/PipeEventCommitter.java
+++ b/iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/progress/PipeEventCommitter.java
@@ -66,9 +66,10 @@ public class PipeEventCommitter {
 
       if (e.getCommitId() <= lastCommitId.get()) {
         LOGGER.warn(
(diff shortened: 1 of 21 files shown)


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/progress/PipeEventCommitter.java
```diff
@@ -66,9 +66,10 @@ public synchronized void commit(EnrichedEvent event) {
 
       if (e.getCommitId() <= lastCommitId.get()) {
         LOGGER.warn(
-            "commit id must be monotonically increasing, lastCommitId: {}, event: {}",
+            "commit id must be monotonically increasing, lastCommitId: {}, event: {}, stack trace: {}",
             lastCommitId.get(),
-            e);
+            e,
+            Thread.currentThread().getStackTrace());
```
SteveYurongSu, line 72:
```suggestion
            Thread.currentThread().getStackTrace(),
            e);
```

SteveYurongSu (approved):

PR #12344, opened 2024-04-15
Pipe: clear reference count for events in retry queue & detect re-increasing reference count for other connectors

VGalaxies:
follow up #12315


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/event/EnrichedEvent.java
```diff
@@ -148,6 +152,9 @@ public boolean decreaseReferenceCount(String holderMessage, boolean shouldReport
             Thread.currentThread().getStackTrace());
       }
     }
+    if (!isSuccessful) {
```
Caideyipi, line 155: It seems that the flag is meaningless and the "0" reference count will still go "<0" and succeed.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/connector/PipeReceiverStatusHandler.java
```diff
@@ -133,12 +133,14 @@ public void handle(TSStatus status, String exceptionMessage, String recordMessag
 
           LOGGER.warn(
               "User conflict exception: will retry {}. status: {}",
-              (retryMaxMillisWhenConflictOccurs == Long.MAX_VALUE ? "forever" : "for at least ")
-                  + (retryMaxMillisWhenConflictOccurs
-                          + exceptionFirstEncounteredTime.get()
-                          - System.currentTimeMillis())
-                      / 1000.0
-                  + " seconds",
+              retryMaxMillisWhenConflictOccurs == Long.MAX_VALUE
```
Caideyipi, line 136: Personally I think keeping the "()" is clearer...

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/event/EnrichedEvent.java
VGalaxies: The `isSuccessful` flag is used to determine whether internal increases or decreases were successful. In the case where the reference count is less than zero, we still consider the decrease to be successful.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/connector/PipeReceiverStatusHandler.java
VGalaxies: The previous approach would result in appending "xxx seconds" after "forever".

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/protocol/thrift/sync/IoTDBDataRegionSyncConnector.java
```diff
@@ -182,6 +224,8 @@ private void doTransfer() {
           tabletBatchBuilder.deepCopyEvents().toString());
     }
 
+    tabletBatchBuilder.decreaseEventsReferenceCount(
+        IoTDBDataRegionSyncConnector.class.getName(), false);
```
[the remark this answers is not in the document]
VGalaxies: fixed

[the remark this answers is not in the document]
SteveYurongSu: > Why false here?

Nice catch! @Caideyipi

SteveYurongSu (approved):

PR #12356, opened 2024-04-16
Load: Add LoadTsfilePointRateLimiter to control load tsfile throughput rate

MiniSho:
## Description
LoadTsFileRateLimiter is a global singleton, and control the load writing point rate at 2 places:
1. The second phase of Two-phase Commit: After loading the tsfilePiecesNode
2. Load the tsfile which does not need to be decoded


SzyWilliam (approved):
+1 what a bravo patch!

Caideyipi (approved):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/scheduler/load/LoadTsFileScheduler.java
```diff
@@ -141,6 +143,7 @@ public LoadTsFileScheduler(
     this.allReplicaSets = new HashSet<>();
     this.isGeneratedByPipe = isGeneratedByPipe;
     this.block = LoadTsFileMemoryManager.getInstance().allocateDataCacheMemoryBlock();
+    this.loadTsFileRateLimiter = new LoadTsFileRateLimiter();
```
Caideyipi, line 146: Make the limiter private per scheduler may not be of use because the scheduler is created per statement. It is maybe ideal to make a public static one in loadTsFileManager and use that.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/load/LoadTsFileManager.java
```diff
@@ -80,6 +80,8 @@ public class LoadTsFileManager {
   private final Map<String, CleanupTask> uuid2CleanupTask = new ConcurrentHashMap<>();
   private final PriorityBlockingQueue<CleanupTask> cleanupTaskQueue = new PriorityBlockingQueue<>();
 
+  private static final LoadTsFileRateLimiter loadTsFileRateLimiter = new LoadTsFileRateLimiter();
```
Caideyipi, line 83: If it is a static one I suggest using upper case as variable name instead.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/load/LoadTsFileRateLimiter.java
```diff
@@ -0,0 +1,65 @@
...
+
+import com.google.common.util.concurrent.RateLimiter;
+
+public class LoadTsFileRateLimiter {
+
+  private static final IoTDBConfig CONFIG = IoTDBDescriptor.getInstance().getConfig();
+
+  private RateLimiter loadWriteRateLimiter;
+
+  private double throughoutMbPerSec = CONFIG.getLoadWriteThroughputMbPerSecond();
```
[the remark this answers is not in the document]
Caideyipi: Besides, for the limiter may be accessed concurrently I suggest you use atomicDouble instead. The RateLimiter is checked and won't have any concurrency problems.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/load/LoadTsFileRateLimiter.java
```diff
@@ -0,0 +1,65 @@
...
+
+  public LoadTsFileRateLimiter() {
+    loadWriteRateLimiter = RateLimiter.create(throughoutMbPerSec);
+  }
+
+  private void setWritePointRate(final double throughoutMbPerSec) {
+    double throughout = throughoutMbPerSec * 1024.0 * 1024.0;
+    // if throughout = 0, disable rate limiting
+    if (throughout <= 0) {
+      throughout = Double.MAX_VALUE;
```
Caideyipi, line 43: ......

Caideyipi (commented):
The main logic is fairly OK, however some kind of things may be taken into account.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/load/LoadTsFileRateLimiter.java
Caideyipi: And maybe this means "throughput" too.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/StorageEngine.java
```diff
@@ -852,6 +853,8 @@ public TSStatus writeLoadTsFileNode(
       return status;
     }
 
+    LoadTsFileRateLimiter.getInstance().acquire(pieceNode.getDataSize());
```
yschengzi, line 856: pieceNode.getDataSize() always return 0 here. it's fake on RPC receiver.

yschengzi (commented):
done.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/StorageEngine.java
SteveYurongSu: Really BAD design...

yschengzi (approved):
LGTM

SteveYurongSu (approved):

PR #12357, opened 2024-04-17
Fix compaction worker stopped after drop database

shuwenwei:
## Description
Fix compaction worker stopped after drop database


OneSizeFitsQuorum (approved):
LGTM

PR #12368, opened 2024-04-18
Optimize region migration log && change region status to removing before transferring leader 

liyuheng55555:


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/region/AddRegionPeerProcedure.java
```diff
@@ -132,31 +143,39 @@ protected Flow executeFromState(ConfigNodeProcedureEnv env, AddRegionPeerState s
...
           LOGGER.info(
-              "AddRegionPeerProcedure success, region {} has been added to DataNode {}",
+              "[pid{}][AddRegion] success, region {} has been added to DataNode {}. Procedure took {} (start at {}).",
+              getProcId(),
               consensusGroupId.getId(),
-              destDataNode.getDataNodeId());
+              destDataNode.getDataNodeId(),
+              CommonDateTimeUtils.convertMillisecondToDurationStr(
+                  System.currentTimeMillis() - getSubmittedTime()),
+              DateTimeUtils.convertLongToDate(getSubmittedTime()));
```
HTHou, line 154: Using `DateTimeUtils.convertLongToDate(long timestamp, String sourcePrecision)` instead?
`DateTimeUtils.convertLongToDate(long timestamp)` will use `ns` as the precision when timestamp_precision is `ns`.

liyuheng55555: good point

OneSizeFitsQuorum (approved):
LGTM

PR #12370, opened 2024-04-18
Fix IoTConsensus LogDispatcherThread Stop Using Futrue Cancel

BUAAserein:
## Description
Removing a single LogDispatcherThread does not clean up the threads in the thread pool, which may cause a memory leak.


OneSizeFitsQuorum (approved):
LGTM

PR #12373, opened 2024-04-18
Feature/move stuff to the extras repository take 2

chrisdutz:
Moved all content from examples and connectors to the new iotdb-extras repository.
Split up the distribution into the parts that bundle the core of iotdb and moved the distributions of the connectors to the new repository too.


HTHou (approved):
LGTM

PR #12378, opened 2024-04-19
Fix NPE when concurrently dropping database and deleting wal

HeimingZ:
NPE occurs because drop database operation remove dataregion and wal deletion thread gets null. To fix it, I add if clause to avoid using null.


OneSizeFitsQuorum (approved):
LGTM

PR #12387, opened 2024-04-22
Fixed rest service insertTablet lost data event return SUCCESS_STATUS

CloudWise-Lukemiao:
Fixed rest service insertTablet  lost data event return SUCCESS_STATUS

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/rest/utils/InsertTabletSortDataUtils.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/rest/utils/InsertTabletSortDataUtils.java
@@ -19,6 +19,7 @@ package org.apache.iotdb.db.protocol.rest.utils;
 
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


CloudWise-Lukemiao:
http://127.0.0.1:18080/rest/v2/insertTablet
When inserting with the following parameters, when there are duplicates in the timestamp and more than half of the values in the values are null

{"is_aligned":false,"device":"root.mqtt.topic.abc454","data_types":["TEXT","INT32","FLOAT"],"measurements":["foo","temp","weight"],"timestamps":[1713495895809,1713499707255,1713495893809,1713495892809,1713495891809,1713495890809,1713495889809,1713495888809,1713495887809,1713495893809,1713495892809,1713495891809,1713495890809,1713495889809,1713495888809],"values":[["bar",null,null,null,null,null,null,null,null,null,null,null,null,null,null],[null,41,32,33,34,39,38,37,36,null,null,null,null,null,null],[null,null,null,null,null,null,null,null,null,87.0,87.3,87.3,87.0,87.3,87.3]]}

After repair, as shown in the following figure：
image

HTHou (approved):

PR #12394, opened 2024-04-23
Fix effective ratio calculation method

HeimingZ:
Use wal node disk size instead of wal manager total size.


OneSizeFitsQuorum (approved):
LGTM

PR #12400, opened 2024-04-24
Pipe: distinguish between old and new process subtasks before and after stuck restart to avoid unexpected hashmap override and removal

VGalaxies:
We noticed that when the pipe is stuck and restarted, the hashcode of new and old process subtasks is identical. This may lead to a situation where the key and value are inconsistent in the maintenance of ConcurrentHashMap `subtasks -> subtasks` by `PipeProcessorSubtaskWorker`, thereby causing unexpectedly removed subtasks that are NOT in the close state by `cleanupClosedSubtasksIfNecessary` in `org.apache.iotdb.db.pipe.task.subtask.processor.PipeProcessorSubtaskWorker`.


Caideyipi (approved):

SteveYurongSu (approved):
Nice catch!

PR #12406, opened 2024-04-24
Pipe: timely restart pipes according to the deletion of original TsFiles

DanielWang2035:
In this PR:
When pipe pins a TsFile, its TsFileResource is also maintained in order to keep track of the TsFileResourceStatus. We timely check the total size of pinned TsFiles whose TsFileResourceStatus is DELETED, if this size exceeds a threshold (with respect to total disk size), we restart all pipes that may be pinning TsFiles.


SteveYurongSu (approved):
Brilliant!

PR #12409, opened 2024-04-24
Change log level and opt error msg for source path of view is deleted

JackieTien97:
1. chaneg the log level from error to warn when https://jira.infra.timecho.com:8443/browse/TIMECHODB-689
2. opt error msg for source path of view is deleted https://jira.infra.timecho.com:8443/browse/TIMECHODB-693
3. delete useless log file


Beyyes (approved):
LGTM

PR #12411, opened 2024-04-24
Fix Concurrency Instability When IoTConsensus LogDispatcher Exits

BUAAserein:
## Description

- Immediately check `Thread.interrupted()` for interrupts after `getBatch()`
- Fix configuration logic in ReplicateTest


OneSizeFitsQuorum (approved):
LGTM

PR #12417, opened 2024-04-25
Revert `-XX:GuaranteedSafepointInterval=0` JVM option

Pengzna:
According to https://bugs.openjdk.org/browse/JDK-8304316 and https://bugs.openjdk.org/browse/JDK-8305994, when `-XX:GuaranteedSafepointInterval=0` and certain versions of JDK 15 ~ JDK 17 are used, the JVM's monitor memory may continue to expand, causing Java's non-heap memory to occupy a large amount of RSS. So we revert `-XX:GuaranteedSafepointInterval=0`


OneSizeFitsQuorum (approved):
LGTM~

Without Professor Peng, tuning IoTDB's JVM would be like an eternal long night.

PR #12434, opened 2024-04-26
Load: Execute tsfile serially to avoid concurrent serialization of the resource file

SteveYurongSu:
As title.


yschengzi (approved):
LGTM-

PR #12439, opened 2024-04-28
Fix insertion cross space compaction recover null pointer exception

shuwenwei:
## Description
Fix insertion cross space compaction recover null pointer exception. If all target files is not exist, the variable `targetFile` will be null, which may cause NullPointerException in `rollback()`.


liyuheng55555 (approved):
LGTM

PR #12446, opened 2024-04-29
[IOTDB-6325] Support RegionScan for active metadata queries [FE Part]

ycycse:
If we want to find the metaData which is activate during a period of time, we need to scan the raw data.
Current implementation of seriesScan may cause random access due to the access to one TSFile repeatedly. We could use file scan to eliminate this.
Region scan is the feature in IoTDB to support fileScan, this PR would support activate metadata query through region-scan, optimizing the performance of full scanning.

requirements docs: https://apache-iotdb.feishu.cn/docx/YqcSdnYbSoJ58zxeK0kcnuSzn8c
design docs: https://apache-iotdb.feishu.cn/docx/ZdeYdBJyRoPtvFxtWHvc6VtCnuc


JackieTien97 (changes requested):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/distribution/SourceRewriter.java
```diff
@@ -719,6 +751,43 @@ private List<PlanNode> processRawSeriesScan(
...
+    }
+
+    for (Map.Entry<TRegionReplicaSet, Set<PartialPath>> entry : devicesInRegion.entrySet()) {
+      RegionScanNode split = (RegionScanNode) node.clone();
+      split.setPlanNodeId(context.queryContext.getQueryId().genPlanNodeId());
+      split.setRegionReplicaSet(entry.getKey());
+      split.setDevicePaths(entry.getValue());
+      split.setOutputCount(isAllDeviceOnlyInOneRegion);
+      planNodesForRegions.add(split);
+    }
```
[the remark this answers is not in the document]
JackieTien97: getDevices may be duplicate.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/source/TimeseriesRegionScanNode.java
```diff
@@ -31,25 +33,37 @@
...
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
 public class TimeseriesRegionScanNode extends RegionScanNode {
-  private Map<PartialPath, TimeseriesSchemaInfo> timeseriesToSchemaInfo;
+  // DevicePath -> (MeasurementPath -> TimeseriesSchemaInfo), it will be serialized to
+  // timeseriesToSchemaInfo in the end
+  Map<PartialPath, Map<PartialPath, List<TimeseriesSchemaInfo>>> deviceToTimeseriesSchemaInfo;
```
JackieTien97, line 54: add private

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/source/TimeseriesRegionScanNode.java
```diff
@@ -101,28 +132,57 @@ public int allowedChildCount() {
...
+        if (path instanceof AlignedPath) {
+          for (String measurementName : ((AlignedPath) path).getMeasurementList()) {
+            partialPaths.add(new PartialPath(new PlainDeviceID(path.getDevice()), measurementName));
+          }
+        } else {
+          partialPaths.add(path);
+        }
+      }
+      return partialPaths;
+    }
```
JackieTien97, line 167: What we really need about this function?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/source/TimeseriesRegionScanNode.java
```diff
@@ -101,28 +132,57 @@ public int allowedChildCount() {
 
   public static PlanNode deserialize(ByteBuffer buffer) {
```
JackieTien97, line 133: Does this deserialize method match the serialize method??

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/source/TimeseriesRegionScanNode.java
JackieTien97: what if we don't call serialize and deserialize method in which case that DataRegion is in current node and we directly pass this object.

JackieTien97 (changes requested):

JackieTien97 (approved):

PR #12450, opened 2024-04-29
Fix health_chech bug

CloudWise-Lukemiao:
Fix the issue of obtaining Windows directory permissions,and the issue of allocating IoTDB memory for configuration, and add the -- help command


CritasWang (approved):
LGTM

PR #12454, opened 2024-04-30
Fixed backup path detection and optimized output log file names

CloudWise-Lukemiao:
Fixed backup path detection and optimized output log file names


CritasWang (approved):
LGTM

PR #12469, opened 2024-05-06
Modify the strategy of compaction schedule

shuwenwei:
## Description
Modify the strategy of compaction schedule. 
1. Skip compaction selection if the queue is full. 
2. Skip high level inner compaction when there are zero level inner compaction in compaction queue.


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/task/InnerSpaceCompactionTask.java
```diff
@@ -510,6 +511,13 @@ public int getProcessedFileNum() {
     return selectedTsFileResourceList.size();
   }
 
+  @Override
+  public void handleTaskCleanup() {
+    if (sumOfCompactionCount == 0) {
```
OneSizeFitsQuorum, line 516: decrement when pop queue

PR #12481, opened 2024-05-07
Remove restrictions on the number and length of tags and attributes

linxt20:


JackieTien97 (changes requested):
also remove them in iotdb-common.properties

JackieTien97 (approved):

PR #12487, opened 2024-05-08
Fix redundant lock taking during create multi timeseries

MarcosZyk:
### Problem
image

### Cause
The original implementation caused repeating lock taking for all given devices when checking template compatibility for multi timeseries creation. Therefore, if the count of devices in one creation request exceeds 65535, the state count in ReentrantReadWriteLock will overflow, causing the error shown in the picture.

### Solution
Make the lock taking only execute once when creating multi timeseries.


Beyyes (approved):
LGTM

PR #12489, opened 2024-05-08
Fixed deadlock problem caused by concurrent auto create schema and delete database

OneSizeFitsQuorum:


CRZbulabula (approved):
LGTM!

CRZbulabula (approved):
After detail analysis, the process of producing deadlock is as follows:

1. The PartitionFetcher requires a read lock for SchemaRegions when create database.
2. The DeleteDatabaseProcedure requires a write lock for SchemaRegions when delete database.
3. Both create and delete database process are serialized.

Therefore, the DeleteDatabaseProcedure would never obtain the write lock for SchemaRegions when the read lock is already took by PartitionFetcher.

Our solution is to decompose the serialization of create and delete database interfaces. Thus, the DeleteDatabaseProcedure could obtain the write lock after the creation process finish. We've checked the code to ensure that there are no severe parallel bugs after we cancel the serializition

PR #12493, opened 2024-05-09
Fix tsfile flush error when drop database or remove peer

HTHou:
## Description

When some TsFiles is flushing and execute drop database, there is some error occured.

img_v3_02an_d33a4080-2e93-4207-a034-035b41c6376g

To fix that, we need to wait all closing TsFiles finished their flushing before delete them physically.


OneSizeFitsQuorum (approved):
LGTM

PR #12494, opened 2024-05-09
Add allow compaction check after inner space compaction task start

shuwenwei:
## Description
add allow compaction check after inner space compaction task start


OneSizeFitsQuorum (approved):
LGTM

PR #12496, opened 2024-05-09
Modify direct buffer memory proportion

shuwenwei:
## Description
Modify direct buffer memory proportion 0.5 -> 0.8


OneSizeFitsQuorum (approved):
LGTM

PR #12510, opened 2024-05-11
Degrade configNode leader redirection log level and avoid meaningless log print when ConfigNodeClient is first constructed.

Pengzna:
as title


Pengzna:
@OneSizeFitsQuorum PTAL

OneSizeFitsQuorum (approved):
LGTM

PR #12511, opened 2024-05-11
[IOTDB-6327] Random choosing available nodes to send sql requests

JackieTien97:
more details can be seen in https://apache-iotdb.feishu.cn/docx/FVh3djW9boALTaxVRwqcUGVznLd


iotdb-client/session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -905,6 +906,25 @@ private SessionDataSet executeStatementMayRedirect(String sql, long timeoutInMs)
...
   }
 
+  private SessionConnection getQuerySessionConnection() {
+    Optional<TEndPoint> endPoint =
+        availableNodes == null ? Optional.empty() : availableNodes.getQueryEndPoint();
+    if (!endPoint.isPresent()) {
+      return defaultSessionConnection;
+    }
+    SessionConnection connection =
+        endPointToSessionConnection.computeIfAbsent(
```
VGalaxies, line 916: If both `enableRedirection` and `enableQueryRedirection` are set to false, this will cause an NPE since `endPointToSessionConnection` has not been initialized.

JackieTien97: fixed

CritasWang (approved):
LGTM

PR #12520, opened 2024-05-13
Feature/jakarta migration

chrisdutz:
In this branch I am trying to migrate IoTDB to using the Jakarta namespace instead of the old JavaX.

It's probably going to take a while.


chrisdutz:
So it seems I'm done with transitioning everything to Jakarta. As mentioned above, I had to give up on Java 8 support setting the new base-line at Java 11, which also allowed updating some dependencies for which unfixed CVEs were being reported. I think now the build is completely CVE-free. I however haven't updated all places where I could now update (Thrift etc.).

I haven't migrated the Extras repo however as I thought it wouldn't make sense to do this, if we haven't decided to actually do the migration.

chrisdutz:
Not merging.

PR #12522, opened 2024-05-14
fixed  health check default parameters, output ulimit as empty, remove unused logs

CloudWise-Lukemiao:
fixed health check default parameters, output ulimit as empty, remove unused logs


CritasWang (approved):
LGTM

PR #12523, opened 2024-05-14
Adjust the hard connection error log information to debug level

CloudWise-Lukemiao:
Adjust the hard connection error log information to debug level


CritasWang (approved):
LGTM

PR #12525, opened 2024-05-14
Load: Update TsFile Resource when enable_auto_create_schema and verify schema are FALSE in Analysis Stage

MiniSho:
## Description
As title.


MiniSho (commented):
LGTM：）

PR #12526, opened 2024-05-14
add transformStatus method in TsFileResource for compaction

shuwenwei:
## Description
add transformStatus method in TsFileResource for compaction


OneSizeFitsQuorum (approved):
LGTM

PR #12549, opened 2024-05-18
Load Tsfile Tool: load does not delete the source tsfile by default

caozj1011:
as title.

## Before this PR
load deletes the source tsfile by default

## After this PR
load does not delete the source tsfile by default


SteveYurongSu (approved):
Thanks!

PR #12550, opened 2024-05-19
Fix incorrect query result in GROUP BY LEVEL query with logical view

liuminghui233:


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/distribution/SourceRewriter.java
```diff
@@ -1556,12 +1556,14 @@ private void calculateGroupByLevelNodeAttributes(
...
+          String groupedInputExpressionsString =
+              originalDescriptor.getInputString(groupedInputExpressions);
           List<Expression> inputExpressions =
-              childrenExpressionMap.get(groupedInputExpressionString);
+              childrenExpressionMap.get(groupedInputExpressionsString);
           if (inputExpressions != null && !inputExpressions.isEmpty()) {
-            descriptorExpressions.addAll(inputExpressions);
+            descriptorExpressions.addAll(groupedInputExpressions);
           }
         }
```
JackieTien97, line 1568: use a field to indicate that there exists view, otherwise we keep same as previous.

JackieTien97 (changes requested):

PR #12552, opened 2024-05-20
Modify strategy of wait task time limitation & Fix proc id

liyuheng55555:
Strategy: max disconnection time = 60s + task running time / 60, but will not more than 10 minutes.


OneSizeFitsQuorum (approved):
LGTM

PR #12573, opened 2024-05-22
Introducing a memory control mechanism during the query planning stage

lancelly:
This PR mainly introduces a memory control mechanism during the query planning stage. Currently, the IoTDB query engine does not implement memory control at the FE (Frontend) stage. In scenarios with massive series queries (e.g., select * from root.**), the query plans generated at the FE stage can become excessively large. Roughly estimating, the size of a single SeriesScanNode is about 1/2 KB, which means that two million series corresponding to two million SeriesScanNodes would occupy 1 GB, posing a potential risk of Out-Of-Memory (OOM). In high concurrency scenarios, even if a single query plan is not large, the total memory occupied by multiple query plans can still lead to OOM.
Therefore, it is now desired to introduce memory size control for FE query plans within the query engine.
Related docs：https://apache-iotdb.feishu.cn/docx/F7nidsysKofhO4xTi35cuDhynCh

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/common/MPPQueryContext.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/common/MPPQueryContext.java
@@ -127,6 +127,7 @@ public class MPPQueryContext {
 
   public void prepareForRetry() {
     this.initResultNodeContext();
+    this.releaseMemoryForFrontEnd();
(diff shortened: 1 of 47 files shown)


JackieTien97 (changes requested):
Think about the query retry logic and the last remaining memory(less than 1MB), you can reserve that before dispatch.

JackieTien97 (approved):

PR #12584, opened 2024-05-24
Avoid infinite and unecessary loop when CN RPC processors are killed/interrupted by OS

Pengzna:
We found that in some scenarios, when the `stop-confignode.sh` command is executed, confignode does not exit immediately, but hangs for about 90 seconds, printing more than 3 million "Unexpected interruption during waiting for configNode leader ready." warning logs.

This problem is because in the code modified by this PR, the RPC thread of confignode will continue to sleep in the while loop. When the confignode process is killed, the OS will continuously try to interrupt the sleep thread, causing the sleep thread to be continuously interrupted and print logs in the while loop.

So we check the confignode for all threads that are sleeping in the while loop. If these threads will only be interrupted during shutdown during their life cycle, then they need to jump out of the while loop when they are interrupted to avoid infinite log printing and stop-confignode lags.

for detailed info, see https://jira.infra.timecho.com:8443/browse/TIMECHODB-750

### DIFF
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/ClusterManager.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/ClusterManager.java
@@ -63,7 +63,7 @@ public class ClusterManager {
         Thread.sleep(100);
       } catch (InterruptedException e) {
         Thread.currentThread().interrupt();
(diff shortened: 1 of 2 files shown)


OneSizeFitsQuorum (approved):
LGTM

liyuheng55555 (approved):
LGTM. The detailed description greatly helps to understand the issue and solution !

CRZbulabula (approved):
LGTM!!!

PR #12589, opened 2024-05-24
Feature/updated cmake

chrisdutz:
Updated CMake to a version that works on all potentially supported system types.


iotdb-client/client-cpp/src/test/CMakeLists.txt
```diff
@@ -38,7 +38,7 @@ IF (DEFINED BOOST_INCLUDEDIR)
 ENDIF()
 
 # Link directories are different for Windows and Linux/Mac
-IF(MSVC)
```
HTHou, line 41: Should we revert the change of this file? I see you changed the generator back to the MSVC.

chrisdutz: Thanks for spotting this ... I would really love to be able to use MinGW for compilation ... updated the variable name ... hopefully that helps.

PR #12594, opened 2024-05-27
Fix the bug that iot consensus doesn't adapt InsertRowsNode

HTHou:
## Description

as title.

img_v3_02b9_df50fec6-a539-481e-9a92-60dd67b6f29g


OneSizeFitsQuorum (approved):
LGTM

PR #12597, opened 2024-05-27
Generate compaction target file with tier level

shuwenwei:
## Description
Generate compaction target file with tier level


OneSizeFitsQuorum (approved):
LGTM

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/tsfile/generator/TsFileNameGenerator.java
```diff
@@ -265,6 +270,7 @@ public static TsFileResource getInnerCompactionTargetFileResource(
     long maxVersion = Long.MIN_VALUE;
     long maxInnerMergeCount = Long.MIN_VALUE;
     long maxCrossMergeCount = Long.MIN_VALUE;
+    int maxTierLevel = Integer.MIN_VALUE;
```
HeimingZ, line 273: Maybe use 0 is better.

shuwenwei: resolved

Issue #12591, opened 2024-05-25
[Feature request] Allow Pandas >= 2


rmpinchback:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/iotdb/issues) and found nothing similar.

### Motivation

I ran into a situation where I was trying to install several libraries via pip that had pandas dependencies, and one of them required >= 2 on the pandas version.  While the requirements.txt in source control for the iotdb Python client doesn't indicate a restriction for not using pandas 2+, the metadata in the setup.py file restricts to less than v2.

### Solution

If there isn't a functional reason for the limitation then I'd suggest removing the "<1.99.99" restriction.  Also the minimum required version should probably be updated to agree with what is in the requirements.txt file, which it currently doesn't.

### Alternatives

No real alternative comes to mind.  Pandas v2 has been out for quite awhile now.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

HTHou:
Hi, I don't think we have such "<1.99.99" restriction for pandas. 
https://github.com/apache/iotdb/blob/baf37c3711d907ed15cc7a99f8d9b16e7c7eceea/iotdb-client/client-py/requirements.txt#L20

Moreover, the python CI is using the latest pandas v2.2.2, and it works well.
https://github.com/apache/iotdb/actions/runs/9262251482/job/25478871454

HTHou:
Oh..okay, I know where the problem is...
https://github.com/apache/iotdb/blob/baf37c3711d907ed15cc7a99f8d9b16e7c7eceea/iotdb-client/client-py/resources/setup.py#L44

rmpinchback:
Awesome, ty!  I should have mentioned setup.py, apologies for that.  I had to hunt around a bit for it myself.   I was seeing the effect of the constraint from Jenkins build logs, but it took a little poking around before I could see the root problem.

HTHou:
Any problem of this PR? https://github.com/apache/iotdb/pull/12601 You can give an approval if there is no problem and then I'll merge it.

rmpinchback:
Commented on the PR, thumbs up.  I think you can close this now.

PR #12601, opened 2024-05-28
[PY-client] Unify setup.py and requirement.txt

HTHou:
## Description

https://github.com/apache/iotdb/issues/12591

### DIFF
--- a/iotdb-client/client-py/requirements.txt
+++ b/iotdb-client/client-py/requirements.txt
@@ -21,5 +21,5 @@ pandas>=1.3.5
 numpy>=1.21.4
 thrift>=0.14.1
 # SQLAlchemy Dialect
-sqlalchemy == 1.3.20
-sqlalchemy-utils == 0.36.8
+sqlalchemy<1.5,>=1.4
+sqlalchemy-utils>=0.37.8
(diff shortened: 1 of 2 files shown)


ycycse:
I find the version requirement in requirement.txt is 
`# Pandas Export
pandas>=1.3.5`
Should we unify these?

ycycse:
> I find the version requirement in requirement.txt is `# Pandas Export pandas>=1.3.5` Should we unify these?

By the way, other requirements in requirement.txt are `# Pandas Export
pandas>=1.3.5
numpy>=1.21.4
thrift>=0.14.1
# SQLAlchemy Dialect
sqlalchemy == 1.3.20
sqlalchemy-utils == 0.36.8` which are not the same as the requirement in setup.py.

Generally speaking, the version in `setup.py` determines the specific version of dependencies that users will download when installing our package. The requirements.txt is typically the version that developers pay attention to. Although inconsistency will not affect usage, in order to unify the development and application environment, I think it is better to unify the two (or let requirements.txt include what setup specifies).

HTHou:
> I find the version requirement in requirement.txt is `# Pandas Export pandas>=1.3.5` Should we unify these?

Yes, it makes sense. Fixed.

ycycse (approved):
LGTM~

rmpinchback:
I don't have any opinions on the sqlalchemy version range because I haven't had reason know the v2 changes or how that could relate to the iotdb python client, but as for the pandas and numpy changes, that all looks good to me.  Thanks!

PR #12611, opened 2024-05-29
Abort old task when compaction config is modified

shuwenwei:
## Description
Abort old task when compaction config is modified


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/schedule/CompactionScheduler.java
```diff
@@ -140,6 +140,8 @@ public static int tryToSubmitInnerSpaceCompactionTask(
     String storageGroupName = tsFileManager.getStorageGroupName();
     String dataRegionId = tsFileManager.getDataRegionId();
 
+    long compactionConfigWhenSelectTask =
```
HTHou, line 143: compactionConfigVersionWhenSelectTask

HTHou (approved):

PR #12618, opened 2024-05-30
JDBC: Support different charsets

MiniSho:
## Description
When initializing a connection, you can specify the charset type in the URL.

```java
try (Connection connection =
        DriverManager.getConnection(
            "jdbc:iotdb://127.0.0.1:6667?charset=GB18030", "root", "root");
    Statement statement = connection.createStatement())
```
The default charset of IoTDB is UTF-8, so the charset does't need to be set when the system default charset is UTF-8


CRZbulabula (commented):
The excellent code style & detailed description benifit me a lot :)

PR #12624, opened 2024-05-30
Filter those time series metadata whose endTime is less than startTime

JackieTien97:
if there exists time series metadata whose endTime is less than startTime, it will cause query thread run into dead while-loop


Beyyes (approved):
LGTM

PR #12627, opened 2024-05-30
Support special path for device TTL

choubenson:
This pr accomplishes two things as following:
- support special path for device ttl, eg: `root.db.'test.device'`
- use a more efficient interface for DataNodeTTLCache.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/utils/TsFileResourceUtils.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/utils/TsFileResourceUtils.java
@@ -122,8 +122,8 @@ public class TsFileResourceUtils {
   }
 
   public static boolean validateTsFileIsComplete(TsFileResource resource) {
-    if (!resource.getTsFile().exists()
-        || resource.getTsFile().length()
+    if (resource.getTsFile().exists()
+        && resource.getTsFile().length()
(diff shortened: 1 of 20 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/utils/TsFileResourceUtils.java
```diff
@@ -122,8 +122,8 @@ public static boolean validateTsFileResourceCorrectness(TsFileResource resource)
   }
 
   public static boolean validateTsFileIsComplete(TsFileResource resource) {
-    if (!resource.getTsFile().exists()
-        || resource.getTsFile().length()
+    if (resource.getTsFile().exists()
+        && resource.getTsFile().length()
```
JackieTien97, line 126: why do this change?

JackieTien97 (changes requested):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/utils/TsFileResourceUtils.java
choubenson: Since empty files are allowed, the check here should be based on the premise that the file exists before determining whether it is complete.

JackieTien97 (approved):

PR #12637, opened 2024-06-03
Fix compaction recover skip negative time partition

shuwenwei:
## Description
Fix compaction recover skip negative time partition


OneSizeFitsQuorum (approved):
LGTM

PR #12647, opened 2024-06-03
Subscription: adapt pipe completion signal for automatically drop subscription

VGalaxies:
- Add topic parameter mode
  - query
  - subscribe
- When to commit PipeTerminateEvent
  - After the DN-side subscription sink receives the PipeTerminateEvent, it reduces the reference count and commits directly, marking the corresponding pipe task as complete.
- How to notify the client
  - When the pipe is automatically dropped and the prefetching queue is unbound, the corresponding prefetching queue is marked as complete (unbind).
  - When the client pulls from a prefetching queue marked as complete, it will get an event containing a TerminationPayload. The client will automatically unsubscribe from the topic of this event.
- How to release resources
  - Upon all associated consumers unsubscribing from the associated topic, reclaim the corresponding prefetching queue (remove).
  - Until all consumers unsubscribe from that topic, all prefetching queues on all DN remain.

---

**Example**: `org.apache.iotdb.SubscriptionSessionExample#dataSubscription3`, `org.apache.iotdb.SubscriptionSessionExample#dataSubscription4`


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/broker/SubscriptionBroker.java
```diff
@@ -162,8 +162,30 @@ public void unbindPrefetchingQueue(final String topicName) {
...
+
+  public void removePrefetchingQueue(final String topicName) {
+    final SubscriptionPrefetchingQueue prefetchingQueue =
+        topicNameToPrefetchingQueue.get(topicName);
+    if (Objects.isNull(prefetchingQueue)) {
+      LOGGER.warn("Subscription: prefetching queue bound to topic [{}] does not exist", topicName);
+      return;
+    }
+
+    if (!prefetchingQueue.isCompleted()) {
```
Caideyipi, line 182: It seems that the "completed" is marked when the pipe is dropped. However in the normal scenarios (namely the unsubscriptions is manually triggered by the client) the "drop pipe" is executed after updating consumer group meta. That is, the "isCompleted" may be false when a topic's last subscription is removed.

iotdb-client/session/src/main/java/org/apache/iotdb/session/subscription/consumer/SubscriptionConsumer.java
```diff
@@ -336,18 +336,23 @@ protected List<SubscriptionMessage> poll(final Set<String> topicNames, final lon
...
+            case TERMINATION:
+              final SubscriptionCommitContext commitContext = pollResponse.getCommitContext();
+              final String topicNameToUnsubscribe = commitContext.getTopicName();
+              LOGGER.info(
+                  "Termination occurred when SubscriptionConsumer {} polling topics {}, unsubscribe topic {} on DN {} automatically",
+                  this,
+                  topicNames,
+                  topicNameToUnsubscribe,
+                  commitContext.getDataNodeId());
+              unsubscribe(topicNameToUnsubscribe);
```
Caideyipi, line 353: Note that since the pipe is automatically dropped, the "DropPipeProcedure"s in "DropSubscriptionProcedure" will encounter failure in validation... Need to add some checks or else the procedure will fail.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/subscription/meta/consumer/ConsumerGroupMeta.java
```diff
@@ -74,6 +74,31 @@ public long getCreationTime() {
...
+  public static /* @NonNull */ Set<String> getTopicsUnsubByGroup(
+      final ConsumerGroupMeta currentMeta, final ConsumerGroupMeta updatedMeta) {
+    if (!Objects.equals(currentMeta.consumerGroupId, updatedMeta.consumerGroupId)) {
+      return Collections.emptySet();
+    }
+    if (!Objects.equals(currentMeta.creationTime, updatedMeta.creationTime)) {
+      return Collections.emptySet();
+    }
+    if (!Objects.equals(
+        currentMeta.consumerIdToConsumerMeta, updatedMeta.consumerIdToConsumerMeta)) {
```
Caideyipi, line 86: Consider closing the consumer directly, and the consumerIdToConsumerMeta is altered by the DropConsumerProcedure.... In that case the queue can not be removed.

Caideyipi: TODO: Remove "topic" key in ConsumerGroupMeta#removeConsumer.

integration-test/src/test/java/org/apache/iotdb/subscription/it/dual/IoTDBSubscriptionTopicIT.java
```diff
@@ -543,7 +551,112 @@ public void testTopicInvalidProcessorConfig() throws Exception {
...
+                              QueryExpression.create(Collections.singletonList(path), null));
+                      while (dataSet.hasNext()) {
+                        dataSet.next();
+                        rowCount.addAndGet(1);
+                      }
+                    }
+                  }
+                  consumer.commitSync(messages);
+                }
+                consumer.unsubscribe(topicName);
```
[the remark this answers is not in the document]
VGalaxies: > This will work not only in the query mode since the "unsubscribe" is explicitly called...

Before the Awaitility test passes, the consumer is always in a poll loop, so it can be confirmed that the drop subscription is triggered due to receiving a TerminationPayload.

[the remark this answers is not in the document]
VGalaxies: > This will work not only in the query mode since the "unsubscribe" is explicitly called...

removed this explicit unsubscribe...

iotdb-client/session/src/main/java/org/apache/iotdb/session/subscription/consumer/SubscriptionConsumer.java
VGalaxies: In DropPipeProcedureV2, no exception will be occurred even if the pipe does not exist...

Caideyipi: Okay, checked that there are not.... However there will be a warn log in PipeDataNodeRemainingEventAndTimeMetrics#deregister. Can you remove it incidentally?

VGalaxies: > Okay, checked that there are not.... However there will be a warn log in PipeDataNodeRemainingEventAndTimeMetrics#deregister. Can you remove it incidentally?

added a boolean return value to `dropPipe`...

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/subscription/subscription/DropSubscriptionProcedure.java
```diff
@@ -242,44 +250,47 @@ protected void rollbackFromOperateOnConfigNodes(ConfigNodeProcedureEnv env) {
...
     }
 
-    // Do nothing to rollback dropPipeProcedures
+    // Rollback AlterConsumerGroupProcedure
+    alterConsumerGroupProcedure.rollbackFromOperateOnConfigNodes(env);
   }
 
   @Override
-  protected void rollbackFromOperateOnDataNodes(ConfigNodeProcedureEnv env) throws IOException {
+  protected void rollbackFromOperateOnDataNodes(final ConfigNodeProcedureEnv env)
```
Caideyipi, line 268: There are 3 points for the rollback function:
1. Usually when we do rollback, we roll back from the last executed to the first executed.
2. We do not rollback DataNode and ConfigNode execution of dropping pipe.
3. Whether the drop subscription procedure needs to be rolled back may need to be considered twice.

iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/subscription/exception/SubscriptionTerminationException.java
```diff
@@ -0,0 +1,40 @@
...
+ * under the License.
+ */
+
+package org.apache.iotdb.rpc.subscription.exception;
+
+import java.util.Objects;
+
+public class SubscriptionTerminationException extends SubscriptionRuntimeCriticalException {
+
+  public SubscriptionTerminationException(final String message) {
```
Caideyipi, line 26: Consider whether to use it before merge.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/subscription/subscription/DropSubscriptionProcedure.java
VGalaxies: > Usually when we do rollback, we roll back from the last executed to the first executed.

The DropSubscriptionProcedure has been appropriately modified. The CreateSubscriptionProcedure will be modified in subsequent metadata restructuring.

> Whether the drop subscription procedure needs to be rolled back may need to be considered twice.

Before metadata restructuring, considering that the process of changing the topic meta and consumer group meta involves multiple procedures, there could be failures during execution. Therefore, it is currently necessary to consider rolling back part of the procedure to ensure CN metadata consistency.

In the subsequent metadata restructuring, we will consider mapping the drop subscription procedure to a single procedure. At that time we can consider not performing rollback to prevent the occurrence of create subscription after the rollback of drop subscription.

PR #12653, opened 2024-06-04
Fix new UDTF execution framework dead loop error.

Sh-Zh-7:
## Description

When `cachedConsumed` reaches the end of cached time column, it may occur dead loop. 

This PR try to fix this issue.


JackieTien97 (changes requested):
Add IT about this case

JackieTien97 (approved):

PR #12654, opened 2024-06-04
Feature/predictive test selection

chrisdutz:
## Description
Enabled develocity predictive test-selection for non-master branches.

This will selectively run tests that are most probable to test changed parts and skip tests, that most probably effect parts that have not changed. 

This is only enabled in non-master branches, so on master nothing should have changed.

I also added a configuration option to automatically have surefire re-run tests that failled. Possibly helping no longer requiring to manually re-run gh action jobs.


chrisdutz:
So ... is there any chance that we'll merge this PR?

PR #12655, opened 2024-06-04
Fix Concurrency Instability When IoTConsensus LogDispatcher Exits

BUAAserein:
## Description
- Use semaphore synchronization when LogDispatcher exits
- Remove change configuration logic in ReplicateTest


OneSizeFitsQuorum (approved):
LGTM

PR #12656, opened 2024-06-04
ConfigNode client allow munually choose ConfigNode

liyuheng55555:


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/client/ConfigNodeClient.java
```diff
@@ -1116,4 +1129,37 @@ public boolean validateObject(
...
+    }
+
+    public void setAndLock(int value) {
+      set(value);
+      locked = true;
+    }
+
+    public void unlock() {
+      locked = false;
+    }
```
shuwenwei, line 1163: Is it necessary to change the value back when unlocking?

liyuheng55555: Seems not necessary, because ConfigNodeClient can find leader by it self

PR #12678, opened 2024-06-06
Pipe IT: Ignore failed cases caused by cluster restart failure

SteveYurongSu:
As title.


Caideyipi (commented):
The same logic shall be in Setup#initClusterEnvironment.

PR #12700, opened 2024-06-11
Adapt partition computation for IDeviceID

OneSizeFitsQuorum:


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/path/PathPatternTree.java
```diff
@@ -169,11 +172,12 @@ public boolean isEmpty() {
         && (pathPatternList == null || pathPatternList.isEmpty());
   }
 
-  public List<String> getAllDevicePatterns() {
+  public List<IDeviceID> getAllDevicePatterns() {
```
JackieTien97, line 175: need to talk about it, if we are sure that we need IDeviceId, we should use the way `List<PartialPath> getAllDevicePaths()`

JackieTien97 (changes requested):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/analyze/cache/partition/PartitionCache.java
```diff
@@ -342,16 +342,18 @@ private void getStorageGroupMap(
    * @param userName
    */
   private void getStorageGroupCacheResult(
-      StorageGroupCacheResult<?> result,
-      List<String> devicePaths,
+      StorageGroupCacheResult<?, ?> result,
+      List<IDeviceID> devicePaths,
```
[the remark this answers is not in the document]
OneSizeFitsQuorum: add todo here

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/path/PathPatternTree.java
OneSizeFitsQuorum: Fixed

JackieTien97 (approved):

PR #12705, opened 2024-06-11
Update pom.xml to fix security issue

HTHou:
## Description

https://github.com/apache/iotdb/security/dependabot

image

These dependences are not using in IoTDB currently, just remove them to fix the security warning.

### DIFF
--- a/example/schema/pom.xml
+++ b/example/schema/pom.xml
@@ -32,7 +32,6 @@
         <dependency>
             <groupId>ch.qos.logback</groupId>
             <artifactId>logback-classic</artifactId>
-            <version>1.3.13</version>
         </dependency>
         <dependency>
             <groupId>org.apache.tsfile</groupId>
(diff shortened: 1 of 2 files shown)


chrisdutz (changes requested):
Please also remove the comments in lines 1426-1430 as they referred to the removed dependency.

chrisdutz:
I took the liberty to actually go through all dependencies and clean up the dependencyManagement as well as the properties.

HTHou:
Thank you Chris, I saw you already updated this PR! 
So is it ready to merge?

chrisdutz:
Yeah ... I first thought it needed that update but then I thought: I bet there's a lot more ... so I took the changes I had in my private fork of IoTDB (Wehre I migrated everything to Java 11 and Jakarta) ... turns out there was a lot more that could go.

PR #12719, opened 2024-06-12
Fix NumberFormatException when upgrade from old version

shuwenwei:
## Description
When upgrading from an older version(before 1.3), if only sbin is updated without simultaneously updating the env script in the conf directory, a NumberFormatException may occur.


OneSizeFitsQuorum (approved):
LGTM

PR #12730, opened 2024-06-13
Fix config node use IoTDBDescriptor

shuwenwei:
## Description
Fix config node use IoTDBDescriptor. Using AuthorityChecker.SUPER_USER in the code of confignode will trigger the initialization of ClusterAuthorityFetcher, causing the logic of IoTDBDescriptor initialization to be entered in confignode.


OneSizeFitsQuorum (approved):
LGTM~
Awesome design!

PR #12750, opened 2024-06-17
[IOTDB-6339] Optimize the time slice control of SeriesScanOperator and AlignedSeriesScanOperator

JackieTien97:
Previously, we will scan until we get at least one satisfied row and then we have chance to do time slice control.
However, when we do the predicate push down optimization, we may need much more time to get a satisfied row from tsfile, so we need to change the control processing of SeriesScanOperator and AlignedSeriesScanOperator.

We will do time check after we got one page data from SeriesScanUtil, even if it's an empty TsBlock.


liuminghui233 (approved):
LGTM :-)

PR #12756, opened 2024-06-18
Check duplicate in start-node and stop-node scripts

shuwenwei:
## Description
1. Check duplicate in start-node and stop-node scripts
2. Properties.store is not used as Properties.store may generate '\\' automatically
3. Fix config node send showConfiguration to config node throw ClassCastException


OneSizeFitsQuorum (approved):
LGTM

PR #12757, opened 2024-06-18
Use milliseconds to express TTL

choubenson:
1. Use milliseconds to express TTL. Eg：
image

The TTL values are always displayed and stored in milliseconds. They are converted to the corresponding unit based on the `TimestampPrecision` parameter only during filtering and cleaning processes. See details in [link](https://jira.infra.timecho.com:8443/browse/TIMECHODB-778)

2. Remove ttl column when showing databases.

### DIFF
--- a/iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/utils/CommonDateTimeUtils.java
+++ b/iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/utils/CommonDateTimeUtils.java
@@ -42,7 +42,8 @@ public class CommonDateTimeUtils {
       default:
         break;
     }
-    return result;
+    // To avoid integer overflow
+    return result < 0 ? Long.MAX_VALUE : result;
   }
 
   public static long currentTime() {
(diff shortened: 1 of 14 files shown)


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/utils/CommonDateTimeUtils.java
```diff
@@ -42,7 +42,8 @@ public static long convertMilliTimeWithPrecision(long milliTime, String timePrec
       default:
         break;
     }
-    return result;
+    // To avoid integer overflow
+    return result < 0 ? Long.MAX_VALUE : result;
```
jt2594838, line 46: It is better to print a log no lower than the warning level. It is even better to also print the stack traces.
Otherwise, any problem will be unable to be traced.

jt2594838 (approved):

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/utils/CommonDateTimeUtils.java
choubenson: Resolved. A warning log has been added when integer overflow.

PR #12767, opened 2024-06-19
[RatisConsensus] Synchronize takeSnapshotAsync on RaftGroupID

SzyWilliam:
1. Log every successful snapshot actions
2. Synchronize on takeSnapshotAsync so that the daemons will not affect force actions


OneSizeFitsQuorum (approved):
LGTM

PR #12785, opened 2024-06-21
Enhance leader quantity metric for load scenario

OneSizeFitsQuorum:
1. fix tag list size not equal issue
2.make metric more accurate

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/load/LoadTsFileManager.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/load/LoadTsFileManager.java
@@ -34,8 +34,6 @@ import org.apache.iotdb.db.consensus.DataRegionConsensusImpl;
 import org.apache.iotdb.db.exception.DiskSpaceInsufficientException;
 import org.apache.iotdb.db.exception.LoadFileException;
 import org.apache.iotdb.db.pipe.agent.PipeDataNodeAgent;
-import org.apache.iotdb.db.queryengine.execution.load.LoadTsFileManager.TsFileWriterManager.CleanupTask;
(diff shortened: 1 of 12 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/load/LoadTsFileManager.java
```diff
@@ -435,6 +437,27 @@ private void loadAll(boolean isGeneratedByPipe, ProgressIndex progressIndex)
                           dataRegion.getDataRegionId(),
                           Tag.TYPE.toString(),
                           Metric.LOAD_POINT_COUNT.toString());
+                  // Because we cannot accurately judge who is the leader here,
+                  // we directly divide the writePointCount by the replicationNum to ensure the
+                  // correctness of
+                  // this metric, which will be accurate in most cases
```
SteveYurongSu, line 443:
```suggestion
                  // Because we cannot accurately judge who is the leader here,
                  // we directly divide the writePointCount by the replicationNum to ensure the
                  // correctness of this metric, which will be accurate in most cases           
```

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/load/LoadTsFileManager.java
```diff
@@ -435,6 +437,27 @@ private void loadAll(boolean isGeneratedByPipe, ProgressIndex progressIndex)
...
+                  MetricService.getInstance()
+                      .count(
+                          writePointCount
+                              / DataRegionConsensusImpl.getInstance()
+                                  .getReplicationNum(
```
SteveYurongSu, line 448: Should we consider the case where the return value of getReplicationNum is 0? It may be necessary to prevent a divide-by-zero error.

SteveYurongSu (approved):
LGTM! What a wonderful implementation :)

SteveYurongSu:
Plz apply spotless

PR #12794, opened 2024-06-24
Try to fix set cluster name integration test

liyuheng55555:
Create new statement after datanode restart.


OneSizeFitsQuorum (approved):
LGTM

PR #12796, opened 2024-06-24
Fix WAL resource leak when using IoT series consensus

Pengzna:
As title


Pengzna:
@OneSizeFitsQuorum PTAL

OneSizeFitsQuorum (approved):
LGTM

PR #12830, opened 2024-06-29
Enhance wal compression

OneSizeFitsQuorum:
1. use directBuffer to optimize performance especially for the gc
2. optimize warn log
3. Upgrade in v1 by default for compatibility


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/io/WALInputStream.java
```diff
@@ -226,15 +231,18 @@ private void loadNextSegmentV2() throws IOException {
         throw new IOException("Unexpected end of file");
       }
       compressedBuffer.flip();
-
       IUnCompressor unCompressor = IUnCompressor.getUnCompressor(segmentInfo.compressionType);
       unCompressor.uncompress(compressedBuffer, dataBuffer);
+      MmapUtil.clean(compressedBuffer);
```
shuwenwei, line 236: The direct buffer is not released if an exception occurred in previous code.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/MmapUtil.java
```diff
@@ -32,4 +33,11 @@ public static void clean(MappedByteBuffer mappedByteBuffer) {
     }
     PlatformDependent.freeDirectBuffer(mappedByteBuffer);
   }
+
+  /** we do not need to clean heapByteBuffer manually, so we just leave it alone. */
+  public static void clean(ByteBuffer byteBuffer) {
+    if (byteBuffer != null & byteBuffer instanceof MappedByteBuffer) {
```
HTHou, line 39: Bitwise operator `&` is a little bit weird here.

OneSizeFitsQuorum: Nice catch!
BTW, IDEA reminds me that the later null instanceof any class will be false, so the previous judgement is not needed

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/io/WALInputStream.java
OneSizeFitsQuorum: It should be noted that clean here is only for early release of unwanted off-heap memory, but even if it is not freed, off-heap memory may eventually be freed by a Full GC.

With a non-memory leak like this, I feel like we don't need to add a lot of try finally to make the code less maintainable, what's your opinion?

shuwenwei: ok

HTHou (approved):

shuwenwei (approved):

PR #12834, opened 2024-07-01
Support ResultSetMetaData.getColumnTypeName for new data type

JackieTien97:
more details can be seen in https://jira.infra.timecho.com:8443/browse/TIMECHODB-818


iotdb-client/jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBResultMetadata.java
```diff
@@ -252,7 +278,13 @@ public int getPrecision(int column) throws SQLException {
       case DOUBLE:
         return 308;
       case TEXT:
+      case BLOB:
+      case STRING:
         return Integer.MAX_VALUE;
+      case TIMESTAMP:
```
CritasWang, line 284: This case doesn't seem to make it to the end.
And the result above is 13, here it is 3

CritasWang (approved):
LGTM

PR #12842, opened 2024-07-03
Fix schema query stuck when system is readonly

OneSizeFitsQuorum:
Currently, SchemaRegion has LINEARIZABLE reads set up, and its state machine stops writing when the system is readonly, potentially causing the query to freeze indefinitely.
img_v3_02cd_b5506948-93bf-41a6-b8d1-9c34ed85caeg
img_v3_02cd_53be8b5c-389f-47b6-a2e0-2d804b36a1ag

In fact, we only need to process dataRegion, because the semantics of readonly itself does not prevent metadata from being written. In addition, dataregion is not LINEARIZABLE read by default. Therefore, even if the dataregion is ratis, there will be no stuck problem.

In addition, the problem comes from the earliest [issue](https://issues.apache.org/jira/browse/IOTDB-5843), the current shutdownhook would stop the RPC client service at the beginning, Therefore, this change will not cause any worse effects.


wangchao316 (approved):
LGTM

PR #12864, opened 2024-07-05
fix fail file directory when import tools failed

2b3c511:
When importing data and schema data, if a failure occurs and the failure directory does not exist, the failed file will not be placed in the specified directory


2b3c511:
When importing data and schema data, if a failure occurs and the failure directory does not exist, the failed file will not be placed in the specified directory

PR #12869, opened 2024-07-08
Reduce copy when get page from chunk ByteBuffer

shuwenwei:
## Description
Reduce copy when get page from chunk ByteBuffer


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/utils/executor/fast/reader/CompactionChunkReader.java
```diff
@@ -103,6 +102,22 @@ public List<Pair<PageHeader, ByteBuffer>> readPageDataWithoutUncompressing() thr
     return pages;
   }
 
+  public static ByteBuffer readCompressedPageData(PageHeader pageHeader, ByteBuffer chunkBuffer)
+      throws IOException {
+    int compressedPageBodyLength = pageHeader.getCompressedSize();
+    if (compressedPageBodyLength > chunkBuffer.remaining()) {
+      throw new IOException(
+          "do not has a complete page body. Expected:"
```
jt2594838, line 110: has -> have

jt2594838 (approved):

PR #12876, opened 2024-07-08
Fix last cache metric updating logic

HTHou:
## Description

When last cache is disabled, we shouldn't update its metric. The insertTablet interface missed that logic.


OneSizeFitsQuorum (approved):
LGTM

PR #12882, opened 2024-07-09
Add direct compressed byte buffer in WALBuffer

shuwenwei:
## Description
Add direct compressed byte buffer in WALBuffer


OneSizeFitsQuorum (approved):
LGTM

PR #12889, opened 2024-07-09
PipeConsensus: Fix no such element error when transfer tsfile 

Pengzna:
As title. 

When transmitting tsfile, an exception of no such element was found. After investigation, it was found that when the Leader had a retryEvent and was also sending events in real time, the load on the Follower would exceed the pipeline size, causing an exception in the retrieval process of tsFileWriter.


OneSizeFitsQuorum (approved):
LGTM

PR #12911, opened 2024-07-12
Fix IndexController visibility issues in concurrency

BUAAserein:
## Description

- Remove the lock and use volatile
- Synchronize updates


OneSizeFitsQuorum (approved):
LGTM

PR #12912, opened 2024-07-12
Fix uncaught IllegalArgumentException when reading broken WAL files

jt2594838:
As the title indicates.


OneSizeFitsQuorum (approved):
LGTM

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/io/WALMetaData.java
```diff
@@ -131,38 +131,45 @@ public long getFirstSearchIndex() {
...
+        for (int size : metaData.buffersSize) {
+          channel.position(offset);
+          ByteBuffer buffer = ByteBuffer.allocate(Long.BYTES);
+          channel.read(buffer);
+          buffer.clear();
+          metaData.memTablesId.add(buffer.getLong());
+          offset += size;
+        }
       }
+    } catch (IllegalArgumentException e) {
```
OneSizeFitsQuorum, line 170: maybe we can catch Exception?

jt2594838: Done

OneSizeFitsQuorum (approved):
LGTM

PR #12913, opened 2024-07-12
Pipe: Fix async client returning self too early when downgrade from handshake v2 to v1

luoluoyuyu:
As title.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/client/IoTDBDataNodeAsyncClientManager.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/client/IoTDBDataNodeAsyncClientManager.java
@@ -199,6 +199,7 @@ public class IoTDBDataNodeAsyncClientManager extends IoTDBClientManager
           }
         };
 
+    client.setShouldReturnSelf(false);
     // Try to handshake by PipeTransferHandshakeV2Req.
     final HashMap<String, String> params = new HashMap<>();
     params.put(
@@ -233,7 +234,8 @@ public class IoTDBDataNodeAsyncClientManager extends IoTDBClientManager
           callback);
       waitHandshakeFinished(isHandshakeFinished);
     }
-
+    client.setShouldReturnSelf(true);
+    client.returnSelf();
     if (exception.get() != null) {
       throw new PipeConnectionException("Failed to handshake.", exception.get());
     }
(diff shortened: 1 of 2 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/client/IoTDBDataNodeAsyncClientManager.java
```diff
@@ -233,7 +234,7 @@ public void onError(Exception e) {
           callback);
       waitHandshakeFinished(isHandshakeFinished);
     }
-
+    client.setShouldReturnSelf(true);
     if (exception.get() != null) {
       throw new PipeConnectionException("Failed to handshake.", exception.get());
```
Caideyipi, line 239: If there are exceptions, maybe we should return the client, or else it's leaked and cannot be borrowed again...

Caideyipi (approved):
LGTM

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/client/IoTDBDataNodeAsyncClientManager.java
```diff
@@ -199,43 +199,48 @@ public void onError(Exception e) {
...
+        exception.set(null);
+
+        client.setTimeoutDynamically(PipeConfig.getInstance().getPipeConnectorHandshakeTimeoutMs());
+        client.pipeTransfer(
+            PipeTransferDataNodeHandshakeV1Req.toTPipeTransferReq(
+                CommonDescriptor.getInstance().getConfig().getTimestampPrecision()),
+            callback);
+        waitHandshakeFinished(isHandshakeFinished);
+      }
+      if (exception.get() != null) {
```
SteveYurongSu, line 238:
```suggestion
      }
      
      if (exception.get() != null) {
```

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/client/IoTDBDataNodeAsyncClientManager.java
```diff
@@ -199,43 +199,48 @@ public void onError(Exception e) {
...
-          targetNodeUrl.getPort());
-
-      supportModsIfIsDataNodeReceiver = false;
-      isHandshakeFinished.set(false);
-      resp.set(null);
-      exception.set(null);
+    try {
+      client.setShouldReturnSelf(false);
+      // Try to handshake by PipeTransferHandshakeV2Req.
+      final HashMap<String, String> params = new HashMap<>();
```
SteveYurongSu, line 205:
```suggestion
      client.setShouldReturnSelf(false);
      
      // Try to handshake by PipeTransferHandshakeV2Req.
      final HashMap<String, String> params = new HashMap<>();
```

SteveYurongSu (approved):

PR #12914, opened 2024-07-12
Issue/timechodb 841

chrisdutz:
Fix various issues with removing data-nodes.

1) Not being able to remove a data-node by it's id
2) When using an rpc-port of 0.0.0.0:{somePort}, it's impossible to only remove one node.


OneSizeFitsQuorum (commented):
Can we add two IT for remove datanodes by ip:port and id ？

chrisdutz:
That's the reason why It's still in WIP and not marked as finished ... my current changes fully address the issues in the JIRA, but have to find out how to sensibly write a unit-test for this (Doing everything with ITs isn't a good idea in my oppinion)

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/service/thrift/ConfigNodeRPCServiceProcessor.java
```diff
@@ -244,6 +246,24 @@ public TGetClusterIdResp getClusterId() {
 
   @Override
   public TDataNodeRegisterResp registerDataNode(TDataNodeRegisterReq req) {
+    // If the remote server is configured to bind to all ip addresses (0.0.0.0),
+    // update the request to use the remote address we got this request from.
+    if ((req.dataNodeConfiguration.getLocation() != null)
+        && req.dataNodeConfiguration.getLocation().isSetClientRpcEndPoint()
+        && req.dataNodeConfiguration.getLocation().getClientRpcEndPoint().isSetIp()
+        && "0.0.0.0"
```
OneSizeFitsQuorum, line 254: In the current version, we allow an RPCAddress of 0.0.0.0 to join the cluster; for example, the default RPCAddress for 1c1d is 0.0.0.0.
Here we don't have to do this replacement, we can support this node to join, because we can use internalAddress to remove the node without any problems

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/service/thrift/ConfigNodeRPCServiceProcessor.java
```diff
@@ -256,6 +276,24 @@ public TDataNodeRegisterResp registerDataNode(TDataNodeRegisterReq req) {
 
   @Override
   public TDataNodeRestartResp restartDataNode(TDataNodeRestartReq req) {
+    // If the remote server is configured to bind to all ip addresses (0.0.0.0),
+    // update the request to use the remote address we got this request from.
+    if ((req.dataNodeConfiguration.getLocation() != null)
```
OneSizeFitsQuorum, line 281: In the current version, we allow an RPCAddress of 0.0.0.0 to join the cluster; for example, the default RPCAddress for 1c1d is 0.0.0.0.
Here we don't have to do this replacement, we can support this node to join, because we can use internalAddress to remove the node without any problems

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/service/thrift/RequestContext.java
```diff
@@ -0,0 +1,69 @@
...
+import org.apache.thrift.transport.TTransport;
+
+import java.net.InetAddress;
+import java.util.Optional;
+
+/**
+ * Special context based on a thread-local allowing to pass information that is used by only very
+ * few services bia Java's ThreadLocals.
+ */
+public class RequestContext {
```
OneSizeFitsQuorum, line 33: maybe we do not need this class anymore

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/DataNodeServerCommandLine.java
```diff
@@ -71,39 +99,40 @@ protected int run(String[] args) throws Exception {
...
+  private int doRemoveDataNode(String[] args)
       throws BadNodeUrlException, TException, IoTDBException, ClientManagerException {
 
     if (args.length != 2) {
       LOGGER.info("Usage: <node-id>/<ip>:<rpc-port>");
-      return;
+      return -1;
     }
 
+    // REMARK: Don't need null or empty-checks for args[ß] or args[1], as if they were
```
OneSizeFitsQuorum, line 129: args[ß] -> args[0]

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/DataNodeServerCommandLine.java
```diff
@@ -135,57 +165,79 @@ private void doRemoveDataNode(String[] args)
...
+  protected static class NodeCoordinateIP implements NodeCoordinate {
+    private final TEndPoint endPoint;
+
+    public NodeCoordinateIP(TEndPoint endPoint) {
+      this.endPoint = endPoint;
+    }
+
+    @Override
+    public boolean matches(TDataNodeLocation location) {
+      return endPoint.ip.equals(location.getClientRpcEndPoint().getIp())
```
OneSizeFitsQuorum, line 225: use internalRPCEndpoint?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/DataNodeServerCommandLine.java
```diff
@@ -135,57 +165,79 @@ private void doRemoveDataNode(String[] args)
...
+              .collect(Collectors.toList());
+    } catch (TException | ClientManagerException e) {
+      LOGGER.error("Get data node locations failed", e);
+    }
+
+    return dataNodeLocations;
+  }
+
+  protected List<NodeCoordinate> parseCoordinates(String coordinatesString) {
+    // Multiple coordinates are separated by ","
```
OneSizeFitsQuorum, line 194: Looks like it currently supports deleting a datanode node? So the judgment here seems superfluous? 
Of course, it's a good design if it's for future scalability

iotdb-core/datanode/src/test/java/org/apache/iotdb/db/service/DataNodeServerCommandLineTest.java
```diff
@@ -0,0 +1,385 @@
...
+import org.apache.iotdb.rpc.TSStatusCode;
+
+import junit.framework.TestCase;
+import org.junit.Assert;
+import org.mockito.Mockito;
+
+import java.util.Arrays;
+import java.util.Collections;
+
+public class DataNodeServerCommandLineTest extends TestCase {
```
OneSizeFitsQuorum, line 43: Good Test！

OneSizeFitsQuorum (commented):
Awesome work!
Please add at least two ITs for this removing datanode feature by id/internalAddress:port besides UT

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/DataNodeServerCommandLine.java
```diff
@@ -135,57 +165,79 @@ private void doRemoveDataNode(String[] args)
...
-      } catch (TException | ClientManagerException e1) {
-        LOGGER.error("Get data node locations failed", e);
+      // In the other case, we expect it to be a numeric value referring to the node-id
+      else if (NumberUtils.isCreatable(coordinate)) {
+        coordinates.add(new NodeCoordinateNodeId(Integer.parseInt(coordinate)));
       }
     }
-    return dataNodeLocations;
+    return coordinates;
+  }
```
liyuheng55555, line 208: 1. What will UrlUtils.parseTEndPointIpv4AndIpv6Url do if coordinate contains ":" but has some other format problem ?
2. If neither of these two conditions can be met, an error should be reported here.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/DataNodeServerCommandLine.java
```diff
@@ -135,57 +165,79 @@ private void doRemoveDataNode(String[] args)
...
-      try (ConfigNodeClient client =
-          ConfigNodeClientManager.getInstance().borrowClient(ConfigNodeInfo.CONFIG_REGION_ID)) {
-        dataNodeLocations =
-            client.getDataNodeConfiguration(-1).getDataNodeConfigurationMap().values().stream()
-                .map(TDataNodeConfiguration::getLocation)
-                .filter(location -> endPoints.contains(location.getClientRpcEndPoint()))
-                .collect(Collectors.toList());
-      } catch (TException | ClientManagerException e) {
-        LOGGER.error("Get data node locations failed", e);
+    List<NodeCoordinate> endPoints = parseCoordinates(args);
```
liyuheng55555, line 175: Rename this var ? (endpoint has some other meaning in IoTDB)

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/service/thrift/ConfigNodeRPCServiceProcessor.java
liyuheng55555: Agree

liyuheng55555 (changes requested):
PTAL~

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/service/thrift/ConfigNodeRPCServiceProcessor.java
chrisdutz: Well ... the issue I was trying to fix, was that in the case of 3 dataNodes joining the cluster, they all will be listed as "0.0.0.0" ... if you then want to remove it by it's real ip then you will not be able to remove it as it only finds 0.0.0.0 nodes and if you use 0.0.0.0, you will delete all nodes. So I think for a 1C1D it should be 127.0.0.1 and not 0.0.0.0 because one is a real ip and the other one isn't.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/DataNodeServerCommandLine.java
chrisdutz: Nothing else than in the original code ... I theory the parser should throw an exception, but the way it was implemented, it would have been an empty NodeCoordinate. Exactly this was also the reason, why deleting by node didn't cause any problems prior to me adding support for deleting by node id. I now handle one case more that wasn't handled before.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/service/thrift/ConfigNodeRPCServiceProcessor.java
OneSizeFitsQuorum: We can use internalAddress to remove it,and there internalAddress must be different

chrisdutz: And what happens to the second and third data-node that registers? I mean ... two of the 3 issues mentioned in this issue are bound to the fact that 0.0.0.0 is simply not right. I also don't quite understand what's the benefit of using an address that brings no additional information. If we have it the way I implemented it, a user can query the list of data-nodes and instantly knows IP addresses he can connect to. 

A perfect solution would be to list all known public addresses of a data-node, when using 0.0.0.0 in the configuration. This would especially help, if the cluster had a separate network and clients are not part of that. 

I personally think using 0.0.0.0 is the worst possible option.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/service/thrift/ConfigNodeRPCServiceProcessor.java
chrisdutz: We are not preventing the node from joining at all ... we are just ensuring we are using a real IP address to list the node ... the only difference is, that in this case the data-node will probably use 127.0.0.1 as address.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/DataNodeServerCommandLine.java
chrisdutz: I think I didn't quite understand what the internal rpc endpoint is ... it this a second rpc endpoint which is used by the cluster internally? If yes, I would indeed prefer using that.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/service/thrift/ConfigNodeRPCServiceProcessor.java
liyuheng55555: The 0.0.0.0 rpc address (which I think "external address" is a better name) is actually not useful to ConfigNode at all, so theoretically it's ok to replace it to actual ip.

But, this modification should be further considered. What you are trying to do is to correct the poor network configuration of DataNode on ConfigNode, which might cause other issues. 

For example, can ConfigNode ensure that the RequestContext parsed is always consistent? If the RequestContext is inconsistent after two restarts, ConfigNode might perceive it as not the same DataNode. Besides, there is also a need to consider the compatibility with past DataNodes that have joined the cluster via ip 0.0.0.0, and whether ports outside of the external port also need to be corrected. 

We can discuss later whether it is necessary to correct all these addresses, and if it is, this won't be a minor modification. Perhaps it would be better for this PR to limit the scope of changes to fixing the remove DataNode function itself ? That will ensure the atomicity of the PR :)

chrisdutz: If you ask me, I think handling 2 bits of information instead of only 1 would be the better solution.
Currently a user runs "show datanodes" he possibly gets a list of "0.0.0.0" which helps him nothing. From a user perspective I think a list of public-facing ips would be ideal (So returing the list of real IPs that a user can connect to the RPC server to) ... but on the other side we also have the internal network and as far as I understood it, the Config Node(s) use the internal rpc to do the cluster management (These can be on completely isolated newtorks, that a user can't even reach). So having a list of external rpc addresses per node and one internal rpc address seems the best option.

chrisdutz: Yeah ... because I think both are important bits of information.

OneSizeFitsQuorum: How about adding two columns to show datanodes this time, and leaving the 0.0.0.0 replacement as TODO?

chrisdutz: Part of the whole problem this issue wanted to address, was the fact that 0.0.0.0 was used. I think I don't quite understand why it would be good to keep the 0.0.0.0, as it has the least bit of information.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/DataNodeServerCommandLine.java
OneSizeFitsQuorum: I think it's OK for now

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/DataNodeServerCommandLine.java
OneSizeFitsQuorum: Yes, each datanode has an external RPC and an internal RPC. The external RPC is meant to be easy to use, so it's possible to use 0.0.0.0 so that it doesn't have to worry about specific IP addresses, which makes it easier for users to use. However, 0.0.0.0 cannot be used for internal IP, so it can be used to identify nodes

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/service/thrift/ConfigNodeRPCServiceProcessor.java
OneSizeFitsQuorum: Hi, Each datanode has an external RPC and an internal RPC. The external RPC is meant to be easy to use, so it's possible to use 0.0.0.0 so that it doesn't have to worry about specific IP addresses, which makes it easier for users to use. However, 0.0.0.0 cannot be used for internal IP, so it can be used to identify nodes。

What's your opinion?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/DataNodeServerCommandLine.java
chrisdutz: But 0.0.0.0 doesn't help if I want to connect to a node ... assuming I'm connected to one device (I obviously need to know one IP address in order to query it) ... I want to shut down that node in order to do something, I run the "give me the other data nodes data" and then I get a list of 0.0.0.0 ... so how would I connect to the next? 
I do understand that setting it up for 0.0.0.0 and that just means: Listen on all available IP addresses is simple for setup. But not for connecting to. So I would propose to use 0.0.0.0 for setup but internally to maintain a list of IPs in that case.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/service/thrift/ConfigNodeRPCServiceProcessor.java
chrisdutz: As I just wrote in the other place. I would allow 0.0.0.0 for setting up the data-node ... but I would use the technique that I use in this pr for using the internal RPC based on the IP used to connect to the config node and I would resolve the 0.0.0.0 on the external rpc to the list of real IP addresses ... you can't open a connection to 0.0.0.0 from any client (possibly you can only do that on the same node as the data-node is running on)

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/DataNodeServerCommandLine.java
OneSizeFitsQuorum: img_v3_02d3_8170a440-1d04-4e7d-83f3-5486e5deaafg
Above, the server is configured with three network cards
If 127.0.0.1 is configured, it can only be accessed locally
If one of these cards is configured, for example, 172.16.17.15, the network segment on which the other two cards are located cannot be accessed
If you configure 0.0.0.0, all will be accessible

If you want to take the ip of the communication between the cn and the dn, the actual ip is the internal_address rather than client_address. Then it makes less sense for us to separate the two ips.

OneSizeFitsQuorum: In any case, whether or not 0.0.0.0 is allowed to register and Remove datanodes is a separate matter. For the former, it feels like we could start a mailing list discussion to see if we need the next PR; Regarding the latter, we can fix this bug first?

chrisdutz: I used the client_address because it looked as if this was the one being used before (you can't configure 0.0.0.0 as internal_address I was told). 
All in all the IP actually doesn't really matter, as the values are simply used to filter items in the list of nodes returned by the config-node. If we use the Node-ID, that's absolutely fine. If we use anything else, it must be possible to identify a node ... using 0.0.0.0 didn't work as identifier, because all had the same value. 

If you ask me, I think it would be good to track the internal_address and a list of external_addresses. (As you are not guaranteed to be able to connect to the internal_address I think the list of external_addresses would be more important. So in above case, for example the internal_address could be 127.0.0.1, and if the external_address is set to 0.0.0.0, then it should list "11.101.17.0, 172.16.17.0, 172.17.0.0". If a user now used any of these external addresses in the command to remove the data-node, this would be clear which one to remove as only one data node will have the same ip-port combination.

OneSizeFitsQuorum: +1 for identifying nodes solely by nodeid. This way we don't have to worry about the function definition.

Let's wait for the community to react. If the majority is in favor, let's remove the IP identifier from this PR

chrisdutz: +1 to that ... would be the cleaner option.
(However for a user it would still be hard to spot the id of the node to remove as all have the same ip of 0.0.0.0)

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/DataNodeServerCommandLine.java
```diff
@@ -135,57 +163,60 @@ private void doRemoveDataNode(String[] args)
...
+
+  protected List<NodeCoordinate> parseCoordinates(String coordinatesString) {
+    // Multiple nodeIds are separated by ","
+    String[] nodeIdStrings = coordinatesString.split(",");
+    List<NodeCoordinate> nodeIdCoordinates = new ArrayList<>(nodeIdStrings.length);
+    for (String nodeId : nodeIdStrings) {
+      // In the other case, we expect it to be a numeric value referring to the node-id
+      if (NumberUtils.isCreatable(nodeId)) {
+        nodeIdCoordinates.add(new NodeCoordinateNodeId(Integer.parseInt(nodeId)));
+      }
```
OneSizeFitsQuorum, line 199: maybe we can log something to remind user here?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/DataNodeServerCommandLine.java
```diff
@@ -135,57 +163,60 @@ private void doRemoveDataNode(String[] args)
...
+      LOGGER.error("Get data node locations failed", e);
     }
+
     return dataNodeLocations;
   }
+
+  protected List<NodeCoordinate> parseCoordinates(String coordinatesString) {
+    // Multiple nodeIds are separated by ","
+    String[] nodeIdStrings = coordinatesString.split(",");
+    List<NodeCoordinate> nodeIdCoordinates = new ArrayList<>(nodeIdStrings.length);
```
OneSizeFitsQuorum, line 194: 1. can we use set now?
2. if it's empty. maybe we can do something?

chrisdutz: If it was completely empty we'd get an error earlier as the number of arguments wouldn't be right.
I could use a Set, however I used what I saw was used elsewhere. Also not quite sure what this sort of "optimization" would help with. The only thing I see would be if I added the raw ids in there, the way elements are added, the set would still contain duplicates. ArrayLists with fixed sizes are implemented more memory and cpu-efficient. So why should I change it? If the "de-duplicating" would be it, I would need to implement this differently. The benefit would be minimal however, as the way the filter is implemented it would simply match the "andMatch" of the first and still we'd get a de-duplicated list out in "dataNodeLocations".

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/DataNodeServerCommandLine.java
chrisdutz: Done

OneSizeFitsQuorum (approved):
Awesome work！

liyuheng55555 (approved):
LGTM

PR #12928, opened 2024-07-15
Use max bytes to allocate WAL compressed buffer

shuwenwei:
## Description
Use max bytes to allocate WAL compressed buffer


OneSizeFitsQuorum (approved):
LGTM

PR #12931, opened 2024-07-15
PipeConsensus: Avoid pipe task being restarted frequently by Pipe framework

Pengzna:
When the writing pressure is high, the cache queue/pipeline of `PipeConsensusAsyncConnector` may frequently fail to enqueue. The Pipe framework will catch the exception and throw `PipeRuntimeConnectorCriticalException`, causing the consensus pipe to be stopped. But the period task of the consensus layer will restart the consensus pipe, causing the consensus pipe to be restarted continuously. This pr fixes this problem.


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/exception/pipe/PipeConsensusConnectorEnqueueException.java
```diff
@@ -0,0 +1,50 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.iotdb.commons.exception.pipe;
+
+import org.apache.iotdb.pipe.api.exception.PipeException;
+
+import java.util.Objects;
+
+public class PipeConsensusConnectorEnqueueException extends PipeException {
```
Caideyipi, line 25: Why not use PipeRuntimeConnectorRetryTimesConfigurableException and set an unlimited(or a reasonably large) retry times?

Pengzna: Good catch，fixed

Caideyipi (approved):

OneSizeFitsQuorum (approved):
LGTM

SteveYurongSu (approved):

PR #12934, opened 2024-07-16
Fix python client dependency issue

HTHou:
## Description

The maven module of the client-py is missing the dependency of iotdb-thrift. It will cause an unexpected issue when package the python client. 

When we released 1.3.2 python client, we included a local thrift file of the master branch instead of building it again.

A new package https://pypi.org/project/apache-iotdb/1.3.2.post0/ released.


CritasWang (approved):
LGTM

PR #12937, opened 2024-07-16
PipeConsensus: Fix circular replicate in kill -9/restart cases

Pengzna:
as title. Any data synchronized from a replica should not be replicated again.

### DIFF
--- a/iotdb-api/pipe-api/src/main/java/org/apache/iotdb/pipe/api/event/dml/insertion/TsFileInsertionEvent.java
+++ b/iotdb-api/pipe-api/src/main/java/org/apache/iotdb/pipe/api/event/dml/insertion/TsFileInsertionEvent.java
@@ -34,7 +34,4 @@ public interface TsFileInsertionEvent extends Event, AutoCloseable {
    * @return {@code Iterable<TabletInsertionEvent>} the list of {@link TabletInsertionEvent}
    */
   Iterable<TabletInsertionEvent> toTabletInsertionEvents();
-
-  /** The method is used to prevent circular replication in PipeConsensus */
-  boolean isGeneratedByPipeConsensus();
 }
--- a/iotdb-api/pipe-api/src/main/java/org/apache/iotdb/pipe/api/event/dml/insertion/TsFileInsertionEvent.java
+++ b/iotdb-api/pipe-api/src/main/java/org/apache/iotdb/pipe/api/event/dml/insertion/TsFileInsertionEvent.java
@@ -34,4 +34,7 @@ public interface TsFileInsertionEvent extends Event, AutoCloseable {
    * @return {@code Iterable<TabletInsertionEvent>} the list of {@link TabletInsertionEvent}
    */
   Iterable<TabletInsertionEvent> toTabletInsertionEvents();
+
+  /** The method is used to prevent circular replication in PipeConsensus */
+  boolean isGeneratedByPipeConsensus();
 }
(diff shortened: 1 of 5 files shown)


iotdb-api/pipe-api/src/main/java/org/apache/iotdb/pipe/api/event/dml/insertion/TsFileInsertionEvent.java
```diff
@@ -34,4 +34,7 @@ public interface TsFileInsertionEvent extends Event, AutoCloseable {
    * @return {@code Iterable<TabletInsertionEvent>} the list of {@link TabletInsertionEvent}
    */
   Iterable<TabletInsertionEvent> toTabletInsertionEvents();
+
+  /** The method is used to prevent circular replication in PipeConsensus */
+  boolean isGeneratedByPipeConsensus();
```
Caideyipi, line 39: This is pipe api and is exposed to users... Do not include this in interface.

Pengzna: ok

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/tsfile/TsFileResource.java
```diff
@@ -161,6 +161,9 @@ public class TsFileResource {
 
   private ProgressIndex maxProgressIndex;
 
+  /** used to prevent circular replication in PipeConsensus */
+  private boolean isGeneratedByPipeConsensus = false;
```
OneSizeFitsQuorum, line 165: do we need to serialize it?

Pengzna: Seems no need. There is no serialization process involved from the generation of tsFileResource by PipeConsensusReceiver to the capture and process of the tsFileInsertionEvent by pipe

Caideyipi: > Seems no need. There is no deserialization process involved from the generation of tsFileResource by PipeConsensusReceiver to the capture and process of the tsFileInsertionEvent by pipe; For other processes involving deserialization of TsFileResource, TsFileResource will have a default value of false for `isGeneratedByPipeConsensus`, which does not affect correctness

Now we do the same thing to "isGeneratedByPipe" method and this is also a flag that will not recover. The consequence is that pipe may resend it after restarts, but seems that it does not affect the correctness(The complete correctness may be ensured after load tsfile is enriched with a progressIndex)

Caideyipi (approved):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/tsfile/TsFileResource.java
Pengzna: @Caideyipi What you said makes sense, I will keep it in mind, thank you!

SteveYurongSu (approved):

luoluoyuyu (approved):

OneSizeFitsQuorum (approved):
LGTM

PR #12946, opened 2024-07-16
Cherry pick Partite Graph Replication

CRZbulabula:
Better region group allocator with smaller time complexity O(n log n) and promised scatter width lower bound


OneSizeFitsQuorum (approved):
LGTM
We will add detailed doc in the future

PR #12856, opened 2024-07-04
Fix IoTConsensus sync stuck problem 

liyuheng55555:
The IoTConsensus protocol requires any write or delete operations to be recorded in the WAL file to synchronize data changes to other nodes. 

Currently, if a delete operation cannot correspond to any MemTable, it will not be recorded in the WAL, causing the synchronization between IoTConsensus nodes to be stuck. 

This PR fixes this issue by treating such delete operations as if they have a MemTable id of -1, and records them in the WAL in the same way. During restart recovery, WALs with a MemTable id of -1 will not be recovered.

PR #12955, opened 2024-07-17
Fix sync stuck problem of IoTConsensus and WAL

liyuheng55555:
In the current Write-Ahead Logging (WAL) mechanism, write operations with the same sequence number may be split into multiple WAL entries and written into different WAL files. IoTConsensus constructs synchronization batches from WAL files for synchronization between replicas. Taking a write operation with sequence number `a` as an example, the sync batch constructed by IoTConsensus for operation `a` needs to **completely** include all WAL entries with sequence number `a`.

The problem lies in how to determine "**completely**".

As the flushing of WAL is asynchronous, IoTConsensus currently waits until a write operation with sequence number `a+1` arrives, or some other operation with a sequence number of `-1` (indicating no need for synchronization between replicas) arrives, before IoTConsensus can confirm that operation `a` can be fully read from the WAL file. **This causes the last write operation to never be synchronized** (Admittedly, if operation `a` can enter pendingEntries, it can be synchronized normally, but pendingEntries is essentially an engineering optimization and should be ignored when thinking outside of IoTConsensus module).

In order to allow the last write operation to be synchronized normally, this PR adds an additional special WAL with a sequence number of -1 after each batch of WAL entries with the same sequence number is written. This special WAL is called **ContinuousSameSearchIndexSeparatorNode**. In this way, IoTConsensus can immediately know this after the current write operation is completed, and the last operation won't be stucked.

Also related to https://github.com/apache/iotdb/pull/12856

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/write/BatchDoneNode.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/write/BatchDoneNode.java
@@ -1,0 +1,42 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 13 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/write/BatchDoneNode.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.apache.iotdb.db.queryengine.plan.planner.plan.node.PlanNodeType;
+import org.apache.iotdb.db.storageengine.dataregion.wal.buffer.IWALByteBufferView;
+import org.apache.iotdb.db.storageengine.dataregion.wal.buffer.WALEntryValue;
+
+/**
+ * For IoTConsensus sync. See <a href="https://github.com/apache/iotdb/pull/12955">github pull
+ * request</a> for details.
+ */
+public class BatchDoneNode implements WALEntryValue {
```
OneSizeFitsQuorum, line 30: WALPlanMarkerNode?

OneSizeFitsQuorum (approved):
LGTM

PR #12960, opened 2024-07-18
Fixed issue where the wal version kept increasing after the write load stopped

OneSizeFitsQuorum:
image

image

img_v3_02cs_06c78a62-0d3b-4352-84cc-a8e96dcfb8cg

Ensure that the wal file does not trigger a file switch if it contains only a Magic String Header of length 6

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/io/LogWriter.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/io/LogWriter.java
@@ -46,7 +46,6 @@ public abstract class LogWriter implements ILogWriter {
   protected final File logFile;
   protected final FileOutputStream logStream;
   protected final FileChannel logChannel;
-  protected long size = 0;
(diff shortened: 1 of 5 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/io/LogWriter.java
```diff
@@ -108,9 +108,6 @@ public double write(ByteBuffer buffer) throws IOException {
       bufferSize = buffer.position();
       buffer.flip();
       compressed = true;
-      size += COMPRESSED_HEADER_SIZE;
-    } else {
-      size += UN_COMPRESSED_HEADER_SIZE;
```
shuwenwei, line 113: Why doesn’t the size need to be increased in this place?

jt2594838 (commented):
Not quite understand where the problem is. Could you please make a further explanation?

shuwenwei (approved):

jt2594838 (approved):

PR #12983, opened 2024-07-19
Use Maven Wrapper in CI

bdemers:
Makes the version of Maven used consistent across all builds.

> [!NOTE]
> Before this change some scripts were using the version of Maven defined by the wrapper, and some _could_ have been using the version that was on the `PATH`.

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/iotdb-client/client-py/release.sh
+++ b/iotdb-client/client-py/release.sh
@@ -27,7 +27,7 @@ rm -Rf dist
 rm -Rf iotdb_session.egg_info
 
 # (Re-)build generated code
-(cd ../..; mvn clean package -pl iotdb-client/client-py -am)
+(cd ../..; ./mvnw clean package -pl iotdb-client/client-py -am)
 
 # Run unit tests
 if [ "$1" == "test" ]; then
--- a/.github/workflows/cluster-it-1c1d.yml
+++ b/.github/workflows/cluster-it-1c1d.yml
@@ -72,7 +72,7 @@ jobs:
         # we do not compile client-cpp for saving time, it is tested in client.yml
         # we can skip influxdb-protocol because it has been tested separately in influxdb-protocol.yml
         run: |
-          mvn clean verify \
+          ./mvnw clean verify \
           -P with-integration-tests \
           -DskipUTs \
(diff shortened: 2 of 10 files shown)


HTHou:
Hi, any advantage of this change?

bdemers:
> Hi, any advantage of this change?

Mostly consistency. Most of the IoTDB CI builds are already using the wrapper, but the ones in this change were not.  
Without all the builds using the same version we could end up in a situation where we see different behavior on some builds and it would be difficult to troubleshoot.

HTHou:
I think the most of IoTDB developers are using the maven on the `PATH` to dev, test, build and release binary packages. We have no report of the problem about the maven version before. From my side, I think it is no necessary to change that for now.  Should rise a discussion on the mail list?

iotdb-client/client-py/release.sh
```diff
@@ -27,7 +27,7 @@ rm -Rf dist
 rm -Rf iotdb_session.egg_info
 
 # (Re-)build generated code
-(cd ../..; mvn clean package -pl iotdb-client/client-py -am)
```
HTHou, line 30: This shell script is only used for releasing python package, not in the CI I believe. The user should be a RM, basically me...

bdemers:
> I think the most of IoTDB developers are using the maven on the `PATH` to dev, test, build and release binary packages. We have no report of the problem about the maven version before. From my side, I think it is no necessary to change that for now. Should rise a discussion on the mail list?

This is totally up to you. However, I'd recommend, being consistent here in using one or the other

This change specifically called out needing to use the wrapper to avoid the type of problems I mentioned
https://github.com/apache/iotdb/commit/09f9bcfedf68d619f5bd6250b4603e3126b6d855

Given that Maven 4 is around the corner, using the wrapper, is another _easy_ way to help with migrating (e.g. single place to update the Maven version). Similar if in the future IoTDB wants to use mvnd (Maven Daemon)

PR #12984, opened 2024-07-22
Fix concurrency problem in Memory Control

SpriCoder:
Fix concurrency problem in Memory Control


shuwenwei (approved):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/rescon/memory/SystemInfo.java
```diff
@@ -196,15 +197,15 @@ public synchronized void resetFlushingMemTableCost(long flushingMemTableCost) {
   }
 
   public boolean addDirectBufferMemoryCost(long size) {
-    while (true) {
-      long memCost = directBufferMemoryCost.get();
+    AtomicBoolean result = new AtomicBoolean(false);
```
shuwenwei, line 200: Is it necessary to use Atomic here?

OneSizeFitsQuorum: Can not use long in lambda

OneSizeFitsQuorum (approved):
LGTM

PR #12986, opened 2024-07-22
Remove UNKNOWN from WALFileVersion which fails recovering V1 WAL file

jt2594838:
A broken V1 wal file will be misjudged as UNKNOWN, truncating the whole file.


OneSizeFitsQuorum (approved):
LGTM~
Awesome design!

PR #12990, opened 2024-07-22
Ensure the flush total points statistic function works correctly when enable_auto_create_schema is false

OneSizeFitsQuorum:
Add automatic creation logic when internal time series fail to write due to metadata issues


SpriCoder (approved):
LGTM~

PR #13001, opened 2024-07-23
[Region Migration] Add retry when the read region does not exist

HxpSerein:
When region migration and read operations are executed concurrently, an error message "The consensus group XXX does not exist" will be reported. The main reason is that the query process obtains partition information in the Analyze stage and then executes it in the Dispatch stage. Therefore, the region may have already been migrated during actual execution, and the solution is to retry these read operations correctly.


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/executor/RegionReadExecutor.java
liyuheng55555, line 1: Beside RegionReadExecutor, IoTDB also has Region**Write**Executor, does it need any changes ?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/scheduler/FragmentInstanceDispatcherImpl.java
```diff
@@ -434,8 +452,13 @@ private void dispatchLocally(final FragmentInstance instance)
...
           LOGGER.warn(readResult.getMessage());
-          throw new FragmentInstanceDispatchException(
-              RpcUtils.getStatus(TSStatusCode.EXECUTE_STATEMENT_ERROR, readResult.getMessage()));
+          if (readResult.isNeedRetry()) {
+            throw new FragmentInstanceDispatchException(
+                RpcUtils.getStatus(TSStatusCode.DISPATCH_ERROR, readResult.getMessage()));
+          } else {
+            throw new FragmentInstanceDispatchException(
+                RpcUtils.getStatus(TSStatusCode.EXECUTE_STATEMENT_ERROR, readResult.getMessage()));
+          }
```
liyuheng55555, line 461: We set two different status code here, but where to handler them ?

liyuheng55555 (changes requested):
PTAL~

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/scheduler/FragmentInstanceDispatcherImpl.java
```diff
@@ -325,7 +327,11 @@ private void dispatchRemoteHelper(final FragmentInstance instance, final TEndPoi
             LOGGER.warn(sendFragmentInstanceResp.message);
             if (sendFragmentInstanceResp.isSetNeedRetry()
                 && sendFragmentInstanceResp.isNeedRetry()) {
-              throw new RatisReadUnavailableException(sendFragmentInstanceResp.message);
+              if (sendFragmentInstanceResp.message.contains("read")) {
```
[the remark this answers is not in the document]
BUAAserein: Use TSStatus instead

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/scheduler/FragmentInstanceDispatcherImpl.java
BUAAserein: Dispatch will determine whether to retry in the outer layer

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/scheduler/FragmentInstanceDispatcherImpl.java
```diff
@@ -325,7 +327,11 @@ private void dispatchRemoteHelper(final FragmentInstance instance, final TEndPoi
...
+                throw new RatisReadUnavailableException(sendFragmentInstanceResp.message);
+              } else {
+                throw new ConsensusGroupNotExistException(sendFragmentInstanceResp.message);
+              }
```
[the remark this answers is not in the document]
BUAAserein: modified

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/executor/RegionReadExecutor.java
BUAAserein: RegionWriteExecutor will retry on the client side

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/executor/RegionReadExecutor.java
```diff
@@ -100,6 +103,10 @@ public RegionExecutionResult execute(
           || t instanceof NotLeaderException
           || t instanceof ServerNotReadyException) {
         resp.setNeedRetry(true);
+        resp.setStatus(new TSStatus(TSStatusCode.RATIS_READ_UNAVAILABLE.getStatusCode()));
+      } else if (t instanceof ConsensusGroupNotExistException) {
+        resp.setNeedRetry(true);
+        resp.setStatus(new TSStatus(TSStatusCode.CONSENSUS_GROUP_NOT_EXIST.getStatusCode()));
```
OneSizeFitsQuorum, line 109: Is it necessary to add these status codes just for retry purposes? Seems to print a log + set the retry flag is ok?

HxpSerein: Because the handling methods for these two exceptions are different, CONSENSUS-GROUP_NOTEXIST does not need to retry again

OneSizeFitsQuorum: OK

OneSizeFitsQuorum (approved):
LGTM

PR #13031, opened 2024-07-25
PipeConsensus: Avoid start pipe by PipeConsensus

Pengzna:
as title.

### DIFF
--- a/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/pipe/PipeConsensusServerImpl.java
+++ b/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/pipe/PipeConsensusServerImpl.java
@@ -276,6 +276,10 @@ public class PipeConsensusServerImpl {
                   consensusGroupId,
                   existedName,
                   expectedStatus);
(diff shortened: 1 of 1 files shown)


iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/pipe/PipeConsensusServerImpl.java
```diff
@@ -276,6 +276,10 @@ public synchronized void checkConsensusPipe(Map<ConsensusPipeName, PipeStatus> e
                   consensusGroupId,
                   existedName,
                   expectedStatus);
+              if (expectedStatus.equals(PipeStatus.RUNNING)) {
+                // Do nothing. Because Pipe framework's metaSync will do that.
```
Caideyipi, line 280: If this pipe already exists on configNode, pipe's meta sync can ensure correctness. Can we assure that the pipe is already known by configNode here?

Pengzna: yes, we can

OneSizeFitsQuorum (approved):
LGTM

PR #13032, opened 2024-07-25
Subscription: fix unexpected cancellation of workers during consumer startup & optimize server-side subscription logs & add synchronized modifier

VGalaxies:
This PR includes the following modification:

1. fix unexpected cancellation of the worker during consumer startup
2. optimize of server-side data subscription logs
3. add synchronized modifier to `onEventInternal`

### DIFF
--- a/iotdb-client/session/src/main/java/org/apache/iotdb/session/subscription/consumer/SubscriptionConsumer.java
+++ b/iotdb-client/session/src/main/java/org/apache/iotdb/session/subscription/consumer/SubscriptionConsumer.java
@@ -234,13 +234,14 @@ abstract class SubscriptionConsumer implements AutoCloseable {
       providers.releaseWriteLock();
     }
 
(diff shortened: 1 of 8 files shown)


SteveYurongSu (approved):

VGalaxies:
Explain why the `onEventInternal` method needs to be synchronized; otherwise, it may cause event loss.

Consider two threads, A and B, both blocking at the `batch.onEvent(event)` call. When thread A calls `batch.onEvent(event)` and the returned events are not empty, it indicates that this batch has already been sealed. At this point, when thread B call `batch.onEvent(event)`, the returned events are empty. Note that the event corresponding to thread B has not entered tht batch. **If thread B's return in `getAndUpdate` executes earlier than thread A**, thread B's `getAndUpdate` will mistakenly believe that the event corresponding to thread B has already entered the batch, causing event loss.

PR #13033, opened 2024-07-25
fix md5 suffix too many

133tosakarin:
the .md5 suffix is not filtered in Path.endWith(".md5"), so we should get the filename of string

### DIFF
--- a/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/SnapshotStorage.java
+++ b/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/SnapshotStorage.java
@@ -166,7 +166,17 @@ public class SnapshotStorage implements StateMachineStorage {
         continue;
       }
       FileInfo fileInfo;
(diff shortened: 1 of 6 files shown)


iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/SnapshotStorage.java
```diff
@@ -155,29 +155,18 @@ SnapshotInfo findLatestSnapshot() {
...
-          fileInfo = new FileInfo(file, null);
-        } else {
-          // For regions that have a separate snapshot installation path, use absolute path
-          fileInfo = new FileInfo(file.toRealPath(), null);
-        }
-      } catch (IOException e) {
-        logger.warn("{} cannot resolve real path of {} due to ", this, file, e);
-        return null;
-      }
+      fileInfo = new FileInfo(file.toPath(), null);
```
OneSizeFitsQuorum, line 169: why remove this?

OneSizeFitsQuorum (approved):
LGTM

PR #13066, opened 2024-07-31
Refactor TsFileValidationTool with abstracting the sequential scan process

jt2594838:
Use TsFileSequenceScan to abstract the process of scanning a TsFile sequentially.
Also add some code refinements.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/tools/validate/TsFileValidationTool.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/tools/validate/TsFileValidationTool.java
@@ -18,25 +18,8 @@
  */
 package org.apache.iotdb.db.tools.validate;
 
-import org.apache.iotdb.db.storageengine.dataregion.tsfile.TsFileResource;
+import org.apache.iotdb.db.tools.utils.TsFileValidationScan;
 
-import org.apache.tsfile.common.conf.TSFileConfig;
-import org.apache.tsfile.common.conf.TSFileDescriptor;
-import org.apache.tsfile.common.constant.TsFileConstant;
-import org.apache.tsfile.encoding.decoder.Decoder;
-import org.apache.tsfile.enums.TSDataType;
-import org.apache.tsfile.file.MetaMarker;
-import org.apache.tsfile.file.header.ChunkGroupHeader;
-import org.apache.tsfile.file.header.ChunkHeader;
-import org.apache.tsfile.file.header.PageHeader;
-import org.apache.tsfile.file.metadata.IDeviceID;
-import org.apache.tsfile.file.metadata.PlainDeviceID;
-import org.apache.tsfile.file.metadata.enums.TSEncoding;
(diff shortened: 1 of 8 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/tools/validate/TsFileValidationTool.java
```diff
@@ -579,12 +209,7 @@ private static boolean checkArgs(String[] args) {
   }
 
   public static void clearMap(boolean resetBadFileNum) {
```
shuwenwei, line 211: Maybe validationScan should be cleaned after scanning all files in one time partition?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/tools/validate/TsFileValidationTool.java
```diff
@@ -189,351 +160,10 @@ public static void main(String[] args) throws IOException {
 
   public static void findUncorrectFiles(List<File> tsFiles) {
```
shuwenwei, line 161: Do we need to check for out-of-order files as well? Out-of-order files do not need to check for overlap between files, but a single file may still have errors(chunk level or page level) checked in TsFileValidationScan

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/tools/validate/TsFileValidationTool.java
jt2594838: Maybe we want to detect cross-partition overlap?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/tools/validate/TsFileValidationTool.java
jt2594838: Added a parameter `ignoreFileOverlap(-if)`. 
When it is true, at the end of the scan of each file, the context maps will be cleared so that overlap between files will not be detected.

iotdb-core/datanode/src/test/java/org/apache/iotdb/db/tools/TsFileValidationScanTest.java
```diff
@@ -0,0 +1,207 @@
...
+
+public class TsFileValidationScanTest {
+
+  public static void main(String[] args) throws IOException {
+    prepareTsFiles();
+  }
+
+  @Test
+  public void testValidation() throws IOException {
+    List<File> files = prepareTsFiles();
```
shuwenwei, line 47: Use '@Before' and '@After'?

shuwenwei (approved):

iotdb-core/datanode/src/test/java/org/apache/iotdb/db/tools/TsFileValidationScanTest.java
jt2594838: Done

PR #13078, opened 2024-08-01
Improve robustness of removing peer step of region migration 

liyuheng55555:
This PR replace some `throw Exception` to `logger.warn`.

Because during removing region peer step, we always want the migration procedure continue, even if there are some mistake.


OneSizeFitsQuorum (approved):
LGTM

PR #13081, opened 2024-08-01
Subscription: java client supports Create Topic If Not Exists and Drop Topic If Exists

luoluoyuyu:
## Description
As Title

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/iotdb-client/session/src/main/java/org/apache/iotdb/session/subscription/SubscriptionSession.java
+++ b/iotdb-client/session/src/main/java/org/apache/iotdb/session/subscription/SubscriptionSession.java
@@ -136,7 +136,7 @@ public class SubscriptionSession extends Session {
     executeNonQueryStatement(sql);
   }
 
-  private void dropTopicIfExists(final String topicName)
+  public void dropTopicIfExists(final String topicName)
       throws IoTDBConnectionException, StatementExecutionException {
     final String sql = String.format("DROP TOPIC IF Exists %s", topicName);
     executeNonQueryStatement(sql);
--- a/iotdb-client/session/src/main/java/org/apache/iotdb/session/subscription/SubscriptionSession.java
+++ b/iotdb-client/session/src/main/java/org/apache/iotdb/session/subscription/SubscriptionSession.java
@@ -86,31 +86,23 @@ public class SubscriptionSession extends Session {
     executeNonQueryStatement(sql);
   }
 
-  public void createTopic(final String topicName, final Properties properties)
+  public void createTopicIfNotExists(final String topicName)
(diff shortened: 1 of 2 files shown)


luoluoyuyu:
> Can the existing createTopic and dropTopic implementations be removed and reuse the newly added code directly?

I mainly considered compatibility. If existing users use the original function, there may be problems, so I added a new function

iotdb-client/session/src/main/java/org/apache/iotdb/session/subscription/SubscriptionSession.java
```diff
@@ -110,12 +110,49 @@ public void createTopic(final String topicName, final Properties properties)
     executeNonQueryStatement(sql);
   }
 
+  public void createTopic(
```
SteveYurongSu, line 113:
```suggestion
  public void createTopicIfNotExists(
```

iotdb-client/session/src/main/java/org/apache/iotdb/session/subscription/SubscriptionSession.java
```diff
@@ -110,12 +110,49 @@ public void createTopic(final String topicName, final Properties properties)
...
+    executeNonQueryStatement(sql);
+  }
+
   public void dropTopic(final String topicName)
       throws IoTDBConnectionException, StatementExecutionException {
     final String sql = String.format("DROP TOPIC %s", topicName);
     executeNonQueryStatement(sql);
   }
 
+  public void dropTopic(final String topicName, final boolean isSetIfExistsCondition)
```
SteveYurongSu, line 147:
```suggestion
  public void dropTopicIfExists(final String topicName, final boolean isSetIfExistsCondition)
```

iotdb-client/session/src/main/java/org/apache/iotdb/session/subscription/SubscriptionSession.java
SteveYurongSu: Add 2 new createTopicIfNotExists methods for 2 createTopic methods

VGalaxies:
> > Can the existing createTopic and dropTopic implementations be removed and reuse the newly added code directly?
> 
> I mainly considered compatibility. If existing users use the original function, there may be problems, so I added a new function

I feel that for this reuse, there is no change in the SDK interface and semantics, so there should be no problem... cc @SteveYurongSu

SteveYurongSu (approved):
Lovely!

PR #13088, opened 2024-08-02
Remove the usage of pthread in cpp example

HTHou:
## Description
glibc 2.31 doesn't support pthread. We can remove it from the cpp example code.
https://www.gnu.org/software/gnulib/manual/html_node/_005f_005flibc_005fsingle_005fthreaded.html

image


CritasWang (approved):
cool

LGTM

PR #13098, opened 2024-08-05
Enhance transferLeader retry frequency when migrating region

OneSizeFitsQuorum:
Avoid frequent transfer leader retries by cn
img_v3_00df_835d888a-a140-482f-a68d-2a0b6cad057g
image


liyuheng55555 (commented):
The change seems reasonable, to sleep for a while before each attempt to retransfer the leader. But why can this solve the problem shown in the first screenshot?

liyuheng55555 (approved):
LGTM

PR #13105, opened 2024-08-07
Add wal compression metric items

shuwenwei:
## Description
Add wal compression metric items


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/wal/io/WALInputStream.java
```diff
@@ -360,6 +365,23 @@ private SegmentInfo getNextSegmentInfo() throws IOException {
     return info;
   }
 
+  private int readWALBufferFromChannel(ByteBuffer buffer) throws IOException {
+    long startTime = System.nanoTime();
+    long startPosition = channel.position();
+    int size = channel.read(buffer);
+    WritingMetrics.getInstance()
+        .recordWALRead(channel.position() - startPosition, System.nanoTime() - startTime);
```
jt2594838, line 373: How about just using size to avoid calling position().

jt2594838 (approved):

jt2594838 (approved):

PR #13112, opened 2024-08-08
Fixed the rest service when executing unsupported SQL, return null

CloudWise-Lukemiao:
Fixed the rest service when executing unsupported SQL, return null


CritasWang (approved):
LGTM

PR #13138, opened 2024-08-12
[region migration] Ignore old region migration procedure during procedure framework updating

liyuheng55555:
image


OneSizeFitsQuorum (approved):
LGTM

PR #13142, opened 2024-08-12
[Region Migration] Convert a Ratis exception to enable read retry

HxpSerein:
## Description
Fix a corner case where the group may not exist during the execution of a Ratis read.

image


OneSizeFitsQuorum (approved):
lgtm

PR #13150, opened 2024-08-13
Share device info in compaction schedule

shuwenwei:
## Description
Share device time index in compaction schedule


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/schedule/CompactionScheduleContext.java
```diff
@@ -37,6 +44,25 @@ public class CompactionScheduleSummary {
 
   // end region
 
+  private final Map<TsFileResource, Map<IDeviceID, DeviceInfo>> partitionFileDeviceInfoCache;
```
jt2594838, line 47: Some monitors should be added to this cache later.

jt2594838 (approved):

Issue #13155, opened 2024-08-13
[Bug] When having more than one pipeline subscription and restarting the server, the second subscription ends in a loop of doom


chrisdutz:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/iotdb/issues) and found nothing similar.

### Version

Mac OS M2 (aarch64)
1.4.0-SNAPHSOT

### Describe the bug and provide the minimal reproduce step

When restarting a system that had more than one subcription for a streaming pipe client, the system gets stuck in an endless loop of logging. 

### What did you expect to see?

The system should startup normally 

### What did you see instead?

Thousands of rows like this in the console:

    14015 [pool-60-IoTDB-Subscription-Executor-Pool-5] WARN  o.a.i.d.s.a.SubscriptionBrokerAgent - Subscription: broker bound to consumer group [weatherServiceGroup] does not exist

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #13156, opened 2024-08-13
[ISSUE-13155] When having more than one pipeline subscription and restarting the server, the second subscription ends in a loop of doom

chrisdutz:
It fixes the issue #13155


VGalaxies (approved):
good catch! thanks~

SteveYurongSu (approved):

PR #13164, opened 2024-08-14
Clear Invalid schema region when restarting

133tosakarin:
Sometimes we shut down some machines in the cluster. Due to the high availability provided by the consensus layer, we can still move data from the stopped machines through the CLI. However, when these stopped machines restart, they still think they are in the corresponding consensus group. In fact, since the data has been migrated during the shutdown, we should clean up these invalid data when restarting. 

In this PR, we fix the problem of not cleaning up the invalid schema region

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/DataNode.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/DataNode.java
@@ -469,24 +469,42 @@ public class DataNode implements DataNodeMBean {
         DataRegionConsensusImpl.getInstance().getAllConsensusGroupIdsWithoutStarting().stream()
             .filter(consensusGroupId -> !dataNodeConsensusGroupIds.contains(consensusGroupId))
             .collect(Collectors.toList());
+    int length = invalidConsensusGroupIds.size();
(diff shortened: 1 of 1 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/DataNode.java
```diff
@@ -469,27 +469,45 @@ private void removeInvalidRegions(List<ConsensusGroupId> dataNodeConsensusGroupI
         DataRegionConsensusImpl.getInstance().getAllConsensusGroupIdsWithoutStarting().stream()
             .filter(consensusGroupId -> !dataNodeConsensusGroupIds.contains(consensusGroupId))
             .collect(Collectors.toList());
+    int length = invalidConsensusGroupIds.size();
+    invalidConsensusGroupIds.addAll(
```
OneSizeFitsQuorum, line 473: Break this into two functions, one to remove schemaregion and one to remove dataregion

133tosakarin: ok :)

OneSizeFitsQuorum (approved):
LGTM

PR #13167, opened 2024-08-14
Pipe: Add TsFileEvent deduplication based on Subscription

Pengzna:
as title.

The difference from Subscription is that the duplicate events are directly discarded here without assigning commit id and reporting (to ensure the order of pipeconsensus). Although duplicate events will not be reported, it does not seem to affect the correctness.

refer:
https://github.com/apache/iotdb/pull/12887
https://github.com/apache/iotdb/pull/12915
https://github.com/apache/iotdb/pull/13061


VGalaxies (commented):
If duplicate tsfiles are filtered out during the process stage, maybe the subscription sink won't need to filter again?

Besides, shouldn't it be impossible to get an event of type PipeRawTabletInsertionEvent from inputEventSupplier? cc @SteveYurongSu

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/task/subtask/processor/PipeProcessorSubtask.java
```diff
@@ -208,6 +225,58 @@ protected boolean executeOnce() throws Exception {
...
+    final boolean isGeneratedByHistoricalExtractor = event.isGeneratedByHistoricalExtractor();
+    final Boolean existedIsGeneratedByHistoricalExtractor =
+        hashCodeToIsGeneratedByHistoricalExtractor.getIfPresent(hashCode);
+    if (Objects.isNull(existedIsGeneratedByHistoricalExtractor)) {
+      hashCodeToIsGeneratedByHistoricalExtractor.put(hashCode, isGeneratedByHistoricalExtractor);
+      return false;
+    }
+    // Multiple PipeRawTabletInsertionEvents parsed from the same PipeTsFileInsertionEvent (i.e.,
+    // with the same isGeneratedByHistoricalExtractor field) are not considered duplicates.
+    if (Objects.equals(existedIsGeneratedByHistoricalExtractor, isGeneratedByHistoricalExtractor)) {
```
luoluoyuyu, line 270: if (existedIsGeneratedByHistoricalExtractor == isGeneratedByHistoricalExtractor ) {

luoluoyuyu (approved):

PR #13178, opened 2024-08-15
RegionMigrate: Fix migrating region with ratisConsensus cause the region is unavailable.

133tosakarin:
If the current region configuration is [a, b, c], we execute the migrate region id from c to d, and the leader is a. During the region migration process, due to the need to catch up with the new node d, if the data is large, it may take a long time.
 
Assuming that the operations before the execution of the Remove phase are successful, the region configuration at this time should be [a, b, c, d], and then start executing RemoveRegion, but due to some reasons, the RemoveRegion operation has not been completed, resulting in the failure of the DELETE_REGION_PEER phase timeout in the executeFromState function of RemoveProcessure, and then enter the DELETE_OLD_REGION_PEER, the new node d has not caught up with a successfully, and the old node c has been deleted. 

If the leaderTransfership is executed midway, such as a to b or a to d, then the entire region group will be in a state without a leader, which is permanent because c thinks it is not in the region group, but other nodes [a, b, d] It is believed that if it is in the region group and the new node d has not caught up with the old leader, d will not execute the vote (due to the lifeCycle of ratis), then this region group will never be elected as the leader because the candidate will not receive the majority vote.

We modify the member change part and wait for the raft member change to complete or throw an exception before returning

In the previous region migration implementation, we randomly selected an available node a and relinquished the existing leader privileges to this available node a. This may result in selecting a new node whose data has not yet caught up with the previous leader. This useless election is wasted because other nodes will never vote for node a.
In this PR, we fixed this issue, and if the migrated node is not the leader, then we do not perform the leader transfer

### DIFF
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
@@ -715,7 +715,6 @@ public class RegionMaintainHandler {
             SyncDataNodeClientPool.getInstance()
                 .changeRegionLeader(
                     regionId, originalDataNode.getInternalEndPoint(), newLeaderNode.get());
-
         if (resp.getStatus().getCode() == TSStatusCode.SUCCESS_STATUS.getStatusCode()) {
           timestamp = resp.getConsensusLogicalTimestamp();
           break;
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
@@ -736,7 +736,7 @@ public class RegionMaintainHandler {
             Collections.singletonMap(
                 regionId,
                 new ConsensusGroupHeartbeatSample(timestamp, newLeaderNode.get().getDataNodeId())));
-    // configManager.getLoadManager().getRouteBalancer().balanceRegionLeaderAndPriority();
+    configManager.getLoadManager().getRouteBalancer().balanceRegionLeaderAndPriority();
 
     LOGGER.info(
         "{}, Change region leader finished, regionId: {}, newLeaderNode: {}",
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
@@ -758,8 +758,7 @@ public class RegionMaintainHandler {
    */
   public Optional<TDataNodeLocation> filterDataNodeWithOtherRegionReplica(
       TConsensusGroupId regionId, TDataNodeLocation filterLocation) {
-    List<TDataNodeLocation> filterLocations = new ArrayList<>();
-    filterLocations.add(filterLocation);
+    List<TDataNodeLocation> filterLocations = Collections.singletonList(filterLocation);
     return filterDataNodeWithOtherRegionReplica(regionId, filterLocations);
   }
 
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
@@ -453,7 +453,7 @@ public class RegionMaintainHandler {
         getIdWithRpcEndpoint(deprecatedLocation),
         status);
     configManager.getLoadManager().removeRegionCache(regionId, deprecatedLocation.getDataNodeId());
-    // configManager.getLoadManager().getRouteBalancer().balanceRegionLeaderAndPriority();
+    configManager.getLoadManager().getRouteBalancer().balanceRegionLeaderAndPriority();
   }
 
   /**
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
@@ -669,8 +669,6 @@ public class RegionMaintainHandler {
       TConsensusGroupId regionId, TDataNodeLocation originalDataNode, TDataNodeLocation coodinator)
       throws ProcedureException, InterruptedException {
     // find new leader
-    final int findNewLeaderTimeLimitSecond = 10;
-    long startTime = System.nanoTime();
     Optional<TDataNodeLocation> newLeaderNode = Optional.empty();
     List<TDataNodeLocation> excludeDataNode = new ArrayList<>();
     excludeDataNode.add(originalDataNode);
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
@@ -665,26 +665,25 @@ public class RegionMaintainHandler {
    * @param regionId The region to be migrated
    * @param originalDataNode The DataNode where the region locates
    */
-  public void transferRegionLeader(TConsensusGroupId regionId, TDataNodeLocation originalDataNode)
-      throws ProcedureException, InterruptedException {
-    List<TDataNodeLocation> excludeDataNode = Collections.singletonList(originalDataNode);
-    transferRegionLeader(regionId, originalDataNode, excludeDataNode);
-  }
-
   public void transferRegionLeader(
-      TConsensusGroupId regionId,
-      TDataNodeLocation originalDataNode,
-      List<TDataNodeLocation> excludeDataNode)
+      TConsensusGroupId regionId, TDataNodeLocation originalDataNode, TDataNodeLocation coodinator)
       throws ProcedureException, InterruptedException {
     // find new leader
     final int findNewLeaderTimeLimitSecond = 10;
     long startTime = System.nanoTime();
     Optional<TDataNodeLocation> newLeaderNode = Optional.empty();
+    List<TDataNodeLocation> excludeDataNode = new ArrayList<>();
+    excludeDataNode.add(originalDataNode);
+    excludeDataNode.add(coodinator);
     while (System.nanoTime() - startTime < TimeUnit.SECONDS.toNanos(findNewLeaderTimeLimitSecond)) {
       newLeaderNode = filterDataNodeWithOtherRegionReplica(regionId, excludeDataNode);
       if (newLeaderNode.isPresent()) {
         break;
       }
+    }
+    if (!newLeaderNode.isPresent()) {
+      // If we have no choice, we use it
+      newLeaderNode = Optional.of(coodinator);
     }
     newLeaderNode.orElseThrow(
         () ->
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
@@ -667,8 +667,7 @@ public class RegionMaintainHandler {
    */
   public void transferRegionLeader(TConsensusGroupId regionId, TDataNodeLocation originalDataNode)
       throws ProcedureException, InterruptedException {
-    List<TDataNodeLocation> excludeDataNode = new ArrayList<>();
-    excludeDataNode.add(originalDataNode);
+    List<TDataNodeLocation> excludeDataNode = Collections.singletonList(originalDataNode);
     transferRegionLeader(regionId, originalDataNode, excludeDataNode);
   }
 
@@ -706,11 +705,9 @@ public class RegionMaintainHandler {
       Integer leaderId = configManager.getLoadManager().getRegionLeaderMap().get(regionId);
 
       if (leaderId != -1) {
-        // The migrated node is not leader, so we specify the transfer leader to current leader node
+        // The migrated node is not leader, so we don't need to transfer temporarily
         if (originalDataNode.getDataNodeId() != leaderId) {
-          newLeaderNode =
-              Optional.of(
-                  configManager.getNodeManager().getRegisteredDataNode(leaderId).getLocation());
+          return;
         }
       }
       while (true) {
@@ -776,9 +773,7 @@ public class RegionMaintainHandler {
 
   public Optional<TDataNodeLocation> filterDataNodeWithOtherRegionReplica(
       TConsensusGroupId regionId, TDataNodeLocation filterLocation, NodeStatus... allowingStatus) {
-    List<TDataNodeLocation> excludeLocations = new ArrayList<>();
-    excludeLocations.add(filterLocation);
-
+    List<TDataNodeLocation> excludeLocations = Collections.singletonList(filterLocation);
     return filterDataNodeWithOtherRegionReplica(regionId, excludeLocations, allowingStatus);
   }
 
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
@@ -675,21 +675,11 @@ public class RegionMaintainHandler {
     List<TDataNodeLocation> excludeDataNode = new ArrayList<>();
     excludeDataNode.add(originalDataNode);
     excludeDataNode.add(coodinator);
-    while (System.nanoTime() - startTime < TimeUnit.SECONDS.toNanos(findNewLeaderTimeLimitSecond)) {
-      newLeaderNode = filterDataNodeWithOtherRegionReplica(regionId, excludeDataNode);
-      if (newLeaderNode.isPresent()) {
-        break;
-      }
-    }
+    newLeaderNode = filterDataNodeWithOtherRegionReplica(regionId, excludeDataNode);
     if (!newLeaderNode.isPresent()) {
       // If we have no choice, we use it
       newLeaderNode = Optional.of(coodinator);
     }
-    newLeaderNode.orElseThrow(
-        () ->
-            new ProcedureException(
-                "Cannot find the new leader after " + findNewLeaderTimeLimitSecond + " seconds"));
-
     // ratis needs DataNode to do election by itself
     long timestamp = System.nanoTime();
     if (TConsensusGroupType.SchemaRegion.equals(regionId.getType())
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
@@ -453,7 +453,7 @@ public class RegionMaintainHandler {
         getIdWithRpcEndpoint(deprecatedLocation),
         status);
     configManager.getLoadManager().removeRegionCache(regionId, deprecatedLocation.getDataNodeId());
-    configManager.getLoadManager().getRouteBalancer().balanceRegionLeaderAndPriority();
+    // configManager.getLoadManager().getRouteBalancer().balanceRegionLeaderAndPriority();
   }
 
   /**
@@ -667,12 +667,22 @@ public class RegionMaintainHandler {
    */
   public void transferRegionLeader(TConsensusGroupId regionId, TDataNodeLocation originalDataNode)
       throws ProcedureException, InterruptedException {
+    List<TDataNodeLocation> excludeDataNode = new ArrayList<>();
+    excludeDataNode.add(originalDataNode);
+    transferRegionLeader(regionId, originalDataNode, excludeDataNode);
+  }
+
+  public void transferRegionLeader(
+      TConsensusGroupId regionId,
+      TDataNodeLocation originalDataNode,
+      List<TDataNodeLocation> excludeDataNode)
+      throws ProcedureException, InterruptedException {
     // find new leader
     final int findNewLeaderTimeLimitSecond = 10;
     long startTime = System.nanoTime();
     Optional<TDataNodeLocation> newLeaderNode = Optional.empty();
     while (System.nanoTime() - startTime < TimeUnit.SECONDS.toNanos(findNewLeaderTimeLimitSecond)) {
-      newLeaderNode = filterDataNodeWithOtherRegionReplica(regionId, originalDataNode);
+      newLeaderNode = filterDataNodeWithOtherRegionReplica(regionId, excludeDataNode);
       if (newLeaderNode.isPresent()) {
         break;
       }
@@ -693,11 +703,22 @@ public class RegionMaintainHandler {
           (CONF.getSchemaRegionRatisRpcLeaderElectionTimeoutMaxMs()
                   + CONF.getSchemaRegionRatisRpcLeaderElectionTimeoutMinMs())
               / 2;
+      Integer leaderId = configManager.getLoadManager().getRegionLeaderMap().get(regionId);
+
+      if (leaderId != -1) {
+        // The migrated node is not leader, so we specify the transfer leader to current leader node
+        if (originalDataNode.getDataNodeId() != leaderId) {
+          newLeaderNode =
+              Optional.of(
+                  configManager.getNodeManager().getRegisteredDataNode(leaderId).getLocation());
+        }
+      }
       while (true) {
         TRegionLeaderChangeResp resp =
             SyncDataNodeClientPool.getInstance()
                 .changeRegionLeader(
                     regionId, originalDataNode.getInternalEndPoint(), newLeaderNode.get());
+
         if (resp.getStatus().getCode() == TSStatusCode.SUCCESS_STATUS.getStatusCode()) {
           timestamp = resp.getConsensusLogicalTimestamp();
           break;
@@ -718,7 +739,7 @@ public class RegionMaintainHandler {
             Collections.singletonMap(
                 regionId,
                 new ConsensusGroupHeartbeatSample(timestamp, newLeaderNode.get().getDataNodeId())));
-    configManager.getLoadManager().getRouteBalancer().balanceRegionLeaderAndPriority();
+    // configManager.getLoadManager().getRouteBalancer().balanceRegionLeaderAndPriority();
 
     LOGGER.info(
         "{}, Change region leader finished, regionId: {}, newLeaderNode: {}",
@@ -742,12 +763,29 @@ public class RegionMaintainHandler {
    */
   public Optional<TDataNodeLocation> filterDataNodeWithOtherRegionReplica(
       TConsensusGroupId regionId, TDataNodeLocation filterLocation) {
+    List<TDataNodeLocation> filterLocations = new ArrayList<>();
+    filterLocations.add(filterLocation);
+    return filterDataNodeWithOtherRegionReplica(regionId, filterLocations);
+  }
+
+  public Optional<TDataNodeLocation> filterDataNodeWithOtherRegionReplica(
+      TConsensusGroupId regionId, List<TDataNodeLocation> filterLocations) {
     return filterDataNodeWithOtherRegionReplica(
-        regionId, filterLocation, NodeStatus.Running, NodeStatus.ReadOnly);
+        regionId, filterLocations, NodeStatus.Running, NodeStatus.ReadOnly);
   }
 
   public Optional<TDataNodeLocation> filterDataNodeWithOtherRegionReplica(
       TConsensusGroupId regionId, TDataNodeLocation filterLocation, NodeStatus... allowingStatus) {
+    List<TDataNodeLocation> excludeLocations = new ArrayList<>();
+    excludeLocations.add(filterLocation);
+
+    return filterDataNodeWithOtherRegionReplica(regionId, excludeLocations, allowingStatus);
+  }
+
+  public Optional<TDataNodeLocation> filterDataNodeWithOtherRegionReplica(
+      TConsensusGroupId regionId,
+      List<TDataNodeLocation> excludeLocations,
+      NodeStatus... allowingStatus) {
     List<TDataNodeLocation> regionLocations = findRegionLocations(regionId);
     if (regionLocations.isEmpty()) {
       LOGGER.warn("Cannot find DataNodes contain the given region: {}", regionId);
@@ -762,11 +800,10 @@ public class RegionMaintainHandler {
             .collect(Collectors.toList());
     Collections.shuffle(aliveDataNodes);
     for (TDataNodeLocation aliveDataNode : aliveDataNodes) {
-      if (regionLocations.contains(aliveDataNode) && !aliveDataNode.equals(filterLocation)) {
+      if (regionLocations.contains(aliveDataNode) && !excludeLocations.contains(aliveDataNode)) {
         return Optional.of(aliveDataNode);
       }
     }
-
     return Optional.empty();
   }
 
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/region/RemoveRegionPeerProcedure.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/region/RemoveRegionPeerProcedure.java
@@ -47,7 +47,6 @@ import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
-import java.util.Optional;
 
 import static org.apache.iotdb.commons.utils.KillPoint.KillPoint.setKillPoint;
 import static org.apache.iotdb.confignode.procedure.state.RemoveRegionPeerState.DELETE_OLD_REGION_PEER;
@@ -61,7 +60,6 @@ public class RemoveRegionPeerProcedure
   private TConsensusGroupId consensusGroupId;
   private TDataNodeLocation coordinator;
   private TDataNodeLocation targetDataNode;
-  private Optional<TDataNodeLocation> destDataNode;
 
   public RemoveRegionPeerProcedure() {
     super();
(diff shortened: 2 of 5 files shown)


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
```diff
@@ -693,11 +703,22 @@ public void transferRegionLeader(TConsensusGroupId regionId, TDataNodeLocation o
           (CONF.getSchemaRegionRatisRpcLeaderElectionTimeoutMaxMs()
                   + CONF.getSchemaRegionRatisRpcLeaderElectionTimeoutMinMs())
               / 2;
+      Integer leaderId = configManager.getLoadManager().getRegionLeaderMap().get(regionId);
+
+      if (leaderId != -1) {
+        // The migrated node is not leader, so we specify the transfer leader to current leader node
```
OneSizeFitsQuorum, line 709: then we do not need to send rpc to datanode?

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
```diff
@@ -453,7 +453,7 @@ public void removeRegionLocation(
         getIdWithRpcEndpoint(deprecatedLocation),
         status);
     configManager.getLoadManager().removeRegionCache(regionId, deprecatedLocation.getDataNodeId());
-    configManager.getLoadManager().getRouteBalancer().balanceRegionLeaderAndPriority();
+    // configManager.getLoadManager().getRouteBalancer().balanceRegionLeaderAndPriority();
```
OneSizeFitsQuorum, line 456: do not comment this

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
```diff
@@ -667,12 +667,22 @@ public void removeDataNodePersistence(TDataNodeLocation dataNodeLocation) {
    */
   public void transferRegionLeader(TConsensusGroupId regionId, TDataNodeLocation originalDataNode)
       throws ProcedureException, InterruptedException {
+    List<TDataNodeLocation> excludeDataNode = new ArrayList<>();
```
OneSizeFitsQuorum, line 670: Collections.singletonList()

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
```diff
@@ -742,12 +763,29 @@ public void transferRegionLeader(TConsensusGroupId regionId, TDataNodeLocation o
...
+      TConsensusGroupId regionId, List<TDataNodeLocation> filterLocations) {
     return filterDataNodeWithOtherRegionReplica(
-        regionId, filterLocation, NodeStatus.Running, NodeStatus.ReadOnly);
+        regionId, filterLocations, NodeStatus.Running, NodeStatus.ReadOnly);
   }
 
   public Optional<TDataNodeLocation> filterDataNodeWithOtherRegionReplica(
       TConsensusGroupId regionId, TDataNodeLocation filterLocation, NodeStatus... allowingStatus) {
+    List<TDataNodeLocation> excludeLocations = new ArrayList<>();
+    excludeLocations.add(filterLocation);
```
OneSizeFitsQuorum, line 780: same as above

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
```diff
@@ -718,7 +739,7 @@ public void transferRegionLeader(TConsensusGroupId regionId, TDataNodeLocation o
             Collections.singletonMap(
                 regionId,
                 new ConsensusGroupHeartbeatSample(timestamp, newLeaderNode.get().getDataNodeId())));
-    configManager.getLoadManager().getRouteBalancer().balanceRegionLeaderAndPriority();
+    // configManager.getLoadManager().getRouteBalancer().balanceRegionLeaderAndPriority();
```
OneSizeFitsQuorum, line 742: do not comment

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
```diff
@@ -667,12 +667,22 @@ public void removeDataNodePersistence(TDataNodeLocation dataNodeLocation) {
    */
   public void transferRegionLeader(TConsensusGroupId regionId, TDataNodeLocation originalDataNode)
       throws ProcedureException, InterruptedException {
+    List<TDataNodeLocation> excludeDataNode = new ArrayList<>();
+    excludeDataNode.add(originalDataNode);
```
OneSizeFitsQuorum, line 671: why not use originDataNode directly?

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
```diff
@@ -578,15 +597,26 @@ public void resetPeerList(ConsensusGroupId groupId, List<Peer> peers) throws Con
   @Override
   public void transferLeader(ConsensusGroupId groupId, Peer newLeader) throws ConsensusException {
     // first fetch the newest information
+    logger.info("transfer consensusGroup {} leader to {}", groupId, newLeader);
     final RaftGroupId raftGroupId = Utils.fromConsensusGroupIdToRaftGroupId(groupId);
     final RaftGroup raftGroup =
         Optional.ofNullable(getGroupInfo(raftGroupId))
             .orElseThrow(() -> new ConsensusGroupNotExistException(groupId));
-
+    if (raftGroup.getPeers() == null) {
```
OneSizeFitsQuorum, line 605: if we forbid leader trasnfer during regionMigration, thus we do not need this judgement?

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
```diff
@@ -578,15 +597,26 @@ public void resetPeerList(ConsensusGroupId groupId, List<Peer> peers) throws Con
...
+      throw new ConsensusException("group " + groupId + " has no peers");
+    }
     final RaftPeer newRaftLeader = Utils.fromPeerAndPriorityToRaftPeer(newLeader, DEFAULT_PRIORITY);
-
     final RaftClientReply reply;
     try {
+      Peer leader = getLeader(groupId);
```
OneSizeFitsQuorum, line 611: same as above, not necessary if we consider this in condignode

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
```diff
@@ -880,6 +936,30 @@ private RaftClientReply sendReconfiguration(RaftGroup newGroupConf)
     return reply;
   }
 
+  private RaftClientReply sendReconfigurationWithRetry(RaftGroup newGroupConf)
```
OneSizeFitsQuorum, line 939: not used?

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
```diff
@@ -112,6 +115,11 @@ class RatisConsensus implements IConsensus {
 
   private static final Logger logger = LoggerFactory.getLogger(RatisConsensus.class);
 
+  // The two fields are used to control the retry times and wait time for setConfiguration
+  // reconfiguration may take many time, so we need to set a long wait time and retry times
+  private final long setConfigurationWaitTime = 10000;
+  private final int setConfigurationRetryTimes = 50;
```
OneSizeFitsQuorum, line 121: not used?

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
```diff
@@ -186,6 +195,16 @@ public RatisConsensus(ConsensusConfig config, IStateMachine.Registry registry) {
                     this.config.getImpl().getRetryWaitMillis(), TimeUnit.MILLISECONDS))
             .build();
 
+    regionMigrateRetryPolicy =
```
OneSizeFitsQuorum, line 198: not used?

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
133tosakarin: ok, i have changed it

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
133tosakarin: I think so, we can't report success

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
133tosakarin: :), ok

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
133tosakarin: I'm considering which one to use

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
133tosakarin: This commit is draft version, I'll delete these code in future commit

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/region/RemoveRegionPeerProcedure.java
```diff
@@ -68,12 +69,25 @@ public RemoveRegionPeerProcedure(
...
+      throws ProcedureException, InterruptedException {
+    LOGGER.info(
+        "[pid{}][RemoveRegion] started, region {} will be removed from DataNode {}.",
+        getProcId(),
+        consensusGroupId.getId(),
+        targetDataNode.getDataNodeId());
+    handler.forceUpdateRegionCache(consensusGroupId, targetDataNode, RegionStatus.Removing);
+    List<TDataNodeLocation> excludeDataNode = new ArrayList<>();
+    excludeDataNode.add(targetDataNode);
+    excludeDataNode.add(coordinator);
```
OneSizeFitsQuorum, line 87: It feels like the semantics here should be allowed, and in the end, if the judgment is the same, do not send rpc?

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
```diff
@@ -742,12 +759,27 @@ public void transferRegionLeader(TConsensusGroupId regionId, TDataNodeLocation o
    */
   public Optional<TDataNodeLocation> filterDataNodeWithOtherRegionReplica(
       TConsensusGroupId regionId, TDataNodeLocation filterLocation) {
+    List<TDataNodeLocation> filterLocations = new ArrayList<>();
```
OneSizeFitsQuorum, line 762: Collections.singletonList

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
```diff
@@ -667,12 +667,21 @@ public void removeDataNodePersistence(TDataNodeLocation dataNodeLocation) {
...
   public void transferRegionLeader(TConsensusGroupId regionId, TDataNodeLocation originalDataNode)
       throws ProcedureException, InterruptedException {
+    List<TDataNodeLocation> excludeDataNode = Collections.singletonList(originalDataNode);
+    transferRegionLeader(regionId, originalDataNode, excludeDataNode);
+  }
+
+  public void transferRegionLeader(
+      TConsensusGroupId regionId,
+      TDataNodeLocation originalDataNode,
+      List<TDataNodeLocation> excludeDataNode)
```
OneSizeFitsQuorum, line 677: seems we only need one node which will be deleted in the future? not a list?

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
OneSizeFitsQuorum: OK, waiting for yuheng's pr

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
```diff
@@ -869,9 +873,34 @@ private RaftClientReply sendReconfiguration(RaftGroup newGroupConf)
     // notify the group leader of configuration change
     RaftClientReply reply;
     try (RatisClient client = getRaftClient(newGroupConf)) {
-      reply =
-          client.getRaftClient().admin().setConfiguration(new ArrayList<>(newGroupConf.getPeers()));
-      if (!reply.isSuccess()) {
+      int basicWaitTime = 500;
+      int maxWaitTime = 10000;
```
OneSizeFitsQuorum, line 877: It feels like we can just check regularly here, without adding idempotent logic, because it doesn't make a big difference.

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
```diff
@@ -869,9 +873,34 @@ private RaftClientReply sendReconfiguration(RaftGroup newGroupConf)
...
+            client
+                .getRaftClient()
+                .admin()
+                .setConfiguration(new ArrayList<>(newGroupConf.getPeers()));
+        if (reply.isSuccess()) {
+          logger.info("reConfiguration success");
+          break;
+        }
+        if (reply.getException() instanceof ReconfigurationInProgressException
+            || reply.getException() instanceof LeaderSteppingDownException
```
OneSizeFitsQuorum, line 891: when we need thses two exceptions? will this cause forever loop?

133tosakarin: it will loop until reply return success or throw other exception.  In fact, I want to use RaftServer.LifeCycle as loop factor,  if majority RaftServer.LifeCycle is RUNNING, then we can return, but I can't get raftServers in newGroupConf's peers

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
133tosakarin: This is also okay.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
```diff
@@ -693,6 +702,14 @@ public void transferRegionLeader(TConsensusGroupId regionId, TDataNodeLocation o
...
                   + CONF.getSchemaRegionRatisRpcLeaderElectionTimeoutMinMs())
               / 2;
+      Integer leaderId = configManager.getLoadManager().getRegionLeaderMap().get(regionId);
+
+      if (leaderId != -1) {
+        // The migrated node is not leader, so we don't need to transfer temporarily
+        if (originalDataNode.getDataNodeId() != leaderId) {
+          return;
+        }
+      }
```
liyuheng55555, line 712: 1. `if (leaderId != -1)` useless check
2. inline leaderId

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/region/RemoveRegionPeerProcedure.java
```diff
@@ -68,12 +69,25 @@ public RemoveRegionPeerProcedure(
...
+    handler.transferRegionLeader(consensusGroupId, targetDataNode, excludeDataNode);
```
liyuheng55555, line 88: If replica = 1, we have to choose coordinator as new leader. Maybe not use `excludedDataNode` only, but use `excluded` and `notSuggested`

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
133tosakarin: these codes have been deleted

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisClient.java
```diff
@@ -167,4 +210,48 @@ public Action handleAttemptFailure(Event event) {
...
+
+  // This policy is used to raft configuration change
+  private static class RatisEndlessRetryPolicy implements RetryPolicy {
+
+    private static final Logger logger = LoggerFactory.getLogger(RatisEndlessRetryPolicy.class);
+    private static final RetryPolicy defaultPolicy;
+
+    static {
+      String str = "";
+      // 50, 500ms, 40, 1000ms, 30, 1500ms, 20, 2000ms, 10, 2500ms
```
liyuheng55555, line 222: What is this comment's meaning? Describe your algorithm more clearly.

And, why define an algorithm in static chunk?

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisClient.java
```diff
@@ -167,4 +210,48 @@ public Action handleAttemptFailure(Event event) {
...
+    @Override
+    public Action handleAttemptFailure(Event event) {
+      // Ratis guarantees that event.getCause() is instance of IOException.
+      // We should allow RaftException or IOException(StatusRuntimeException, thrown by gRPC) to be
+      // retried.
+      Optional<Throwable> unexpectedCause =
+          Optional.ofNullable(event.getCause())
+              .filter(RaftException.class::isInstance)
+              .map(Throwable::getCause)
+              .filter(StatusRuntimeException.class::isInstance);
```
liyuheng55555, line 246: What is `RaftException or IOException(StatusRuntimeException, thrown by gRPC)`? Words between () is confusing, StatusRuntimeException is not a kind of IOException.

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
```diff
@@ -122,6 +122,7 @@ class RatisConsensus implements IConsensus {
   private final RaftClientRpc clientRpc;
 
   private final IClientManager<RaftGroup, RatisClient> clientManager;
+  private final IClientManager<RaftGroup, RatisClient> reconfClientManager;
```
liyuheng55555, line 125: **reconfiguration**ClientManager. We usually not use abbreviation in iotdb.

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
```diff
@@ -864,14 +869,30 @@ private RatisClient getRaftClient(RaftGroup group) throws ClientManagerException
     }
   }
 
+  private RatisClient getConfRaftClient(RaftGroup group) throws ClientManagerException {
```
liyuheng55555, line 872: get**Reconfiguration**RaftClient. Keep name consistent.

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
```diff
@@ -901,12 +922,21 @@ public void allowStaleRead(ConsensusGroupId consensusGroupId) {
 
   private class RatisClientPoolFactory implements IClientPoolFactory<RaftGroup, RatisClient> {
 
+    private boolean isReConfiguration;
+
+    RatisClientPoolFactory(boolean isReConfiguration) {
+      this.isReConfiguration = isReConfiguration;
+    }
```
liyuheng55555, line 929: is**Reconfiguration**

liyuheng55555 (changes requested):
Please take a look.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
liyuheng55555: I reconsidered here, maybe the check is necessary?

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
133tosakarin: what's mean?

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
133tosakarin: ok:)

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
133tosakarin: ok:)

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisClient.java
133tosakarin: I copy from RatisRetryPolicy

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisClient.java
133tosakarin: it's a pairArray, pair such as (n, t),    use the i-th pair, then client will retry n times, each time sleep t  ms.  

I cancat these pair as str, The MultipleLinearRandomRetry will parse it.

I want to calculate only once,  so I think static chunk will only be executed once

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/region/RemoveRegionPeerProcedure.java
133tosakarin: Then I'll change it again.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
133tosakarin: leaderId may be -1, I think we should check it

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
liyuheng55555: C->c

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisClient.java
liyuheng55555: I think just using "50, 500ms, 40, 1000ms, 30, 1500ms, 20, 2000ms, 10, 2500ms" will be more clear than building a String by loop.

But, this strategy seems not actually **endless** ?

133tosakarin: yes, the retry time should give a large number.
if the number is enough large, then the loop is ok

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisClient.java
```diff
@@ -167,4 +216,41 @@ public Action handleAttemptFailure(Event event) {
...
+
+    RatisEndlessRetryPolicy() {
+      // about 1 hour wait Time.
+      defaultPolicy =
+          RetryPolicies.retryForeverWithSleep(TimeDuration.valueOf(2, TimeUnit.SECONDS));
+    }
+
+    @Override
+    public Action handleAttemptFailure(Event event) {
+      if (event.getCause() instanceof ReconfigurationInProgressException) {
```
OneSizeFitsQuorum, line 235: why not use if（A || B || C)

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
```diff
@@ -665,27 +665,26 @@ public void removeDataNodePersistence(TDataNodeLocation dataNodeLocation) {
...
     while (System.nanoTime() - startTime < TimeUnit.SECONDS.toNanos(findNewLeaderTimeLimitSecond)) {
       newLeaderNode = filterDataNodeWithOtherRegionReplica(regionId, excludeDataNode);
       if (newLeaderNode.isPresent()) {
         break;
       }
     }
+    if (!newLeaderNode.isPresent()) {
+      // If we have no choice, we use it
+      newLeaderNode = Optional.of(coodinator);
+    }
```
liyuheng55555, line 687: Better than my thought, nice and clean code!

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisClient.java
```diff
@@ -212,46 +218,39 @@ public Action handleAttemptFailure(Event event) {
   }
 
   // This policy is used to raft configuration change
+  //
```
liyuheng55555, line 221: Use javadoc /** */ instead

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
```diff
@@ -922,18 +922,18 @@ public void allowStaleRead(ConsensusGroupId consensusGroupId) {
 
   private class RatisClientPoolFactory implements IClientPoolFactory<RaftGroup, RatisClient> {
 
-    private boolean isReConfiguration;
+    private final boolean isReconfiguration;
 
     RatisClientPoolFactory(boolean isReConfiguration) {
-      this.isReConfiguration = isReConfiguration;
+      this.isReconfiguration = isReConfiguration;
```
liyuheng55555, line 928: you forget something

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisClient.java
133tosakarin: haha!   I'm lazy here :) . This part is the copilot prompt, and the tab is generated with one click.

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisClient.java
```diff
@@ -212,46 +218,39 @@ public Action handleAttemptFailure(Event event) {
...
-        basicSleep += 500;
-      }
+    private final RetryPolicy defaultPolicy;
 
+    RatisEndlessRetryPolicy() {
+      // about 1 hour wait Time.
       defaultPolicy =
-          MultipleLinearRandomRetry.parseCommaSeparated(str.substring(0, str.length() - 1));
+          RetryPolicies.retryForeverWithSleep(TimeDuration.valueOf(2, TimeUnit.SECONDS));
     }
```
liyuheng55555, line 231: 1. Isn't retry every 2 seconds too frequently ?
2. Why still // about 1 hour wait Time.

liyuheng55555 (changes requested):

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisClient.java
133tosakarin: ok

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisClient.java
133tosakarin: 1. if reply return false, then resend request after 2s.  Instead of sending it every 2 seconds

sorry, I forget delete comment

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisClient.java
OneSizeFitsQuorum: Why not change this?

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
OneSizeFitsQuorum: why not change?

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisClient.java
133tosakarin: I have changed

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
133tosakarin: It's convenient to use to determine whether an object is included

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
133tosakarin: I didn't see it

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
133tosakarin: I have used another policy

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
133tosakarin: ok

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
133tosakarin: I have changed

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
133tosakarin: I have changed

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
```diff
@@ -864,14 +869,30 @@ private RatisClient getRaftClient(RaftGroup group) throws ClientManagerException
     }
   }
 
+  private RatisClient getConfigurationRaftClient(RaftGroup group) throws ClientManagerException {
+    try {
+      return reconfigurationClientManager.borrowClient(group);
```
OneSizeFitsQuorum, line 874: consistent with clientManager

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
OneSizeFitsQuorum: remove it

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RegionMaintainHandler.java
```diff
@@ -665,18 +665,26 @@ public void removeDataNodePersistence(TDataNodeLocation dataNodeLocation) {
...
     // find new leader
     final int findNewLeaderTimeLimitSecond = 10;
     long startTime = System.nanoTime();
     Optional<TDataNodeLocation> newLeaderNode = Optional.empty();
+    List<TDataNodeLocation> excludeDataNode = new ArrayList<>();
+    excludeDataNode.add(originalDataNode);
+    excludeDataNode.add(coodinator);
     while (System.nanoTime() - startTime < TimeUnit.SECONDS.toNanos(findNewLeaderTimeLimitSecond)) {
-      newLeaderNode = filterDataNodeWithOtherRegionReplica(regionId, originalDataNode);
+      newLeaderNode = filterDataNodeWithOtherRegionReplica(regionId, excludeDataNode);
```
OneSizeFitsQuorum, line 679: cause redundant sleep when replica is 1?

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
133tosakarin: ok

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
```diff
@@ -864,14 +868,30 @@ private RatisClient getRaftClient(RaftGroup group) throws ClientManagerException
...
   private RaftClientReply sendReconfiguration(RaftGroup newGroupConf)
       throws RatisRequestFailedException {
     // notify the group leader of configuration change
     RaftClientReply reply;
-    try (RatisClient client = getRaftClient(newGroupConf)) {
-      reply =
-          client.getRaftClient().admin().setConfiguration(new ArrayList<>(newGroupConf.getPeers()));
-      if (!reply.isSuccess()) {
+    try (RatisClient client = getConfigurationRaftClient(newGroupConf)) {
+      while (true) {
```
OneSizeFitsQuorum, line 886: do we need this while(true) now? as we have already done endless retry in raftclient

133tosakarin: I think it's no problem,  but i can remove it

liyuheng55555 (approved):
LGTM

PR #13179, opened 2024-08-15
[region migration] Fix NPE of RegionWriteExecutor

liyuheng55555:


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/executor/RegionWriteExecutor.java
```diff
@@ -147,7 +147,12 @@ public RegionExecutionResult execute(ConsensusGroupId groupId, PlanNode planNode
           new WritePlanNodeExecutionContext(groupId, regionManager.getRegionLock(groupId));
       return planNode.accept(executionVisitor, context);
     } catch (Throwable e) {
-      LOGGER.warn(e.getMessage(), e);
+      String forgivableReason = "";
+      if (e instanceof NullPointerException && regionManager.getRegionLock(groupId) == null) {
```
OneSizeFitsQuorum, line 151: do not use this

PR #13210, opened 2024-08-17
Eventually consistent heartbeat samples

CRZbulabula:
As the term and leader reported by the Ratis interfaces might inconsistent with a small probability, the ConfigNode might end up with caching incorrect leader location. By caching all heartbeat samples, the ConfigNode can ensure the eventually consistency for all leader locations of all RegionGroups use the Ratis consensus protocol.


OneSizeFitsQuorum (commented):
Will this cause redundant event push for same sample?

CRZbulabula:
> Will this cause redundant event push for same sample?

No, the observer only publishes events that have greater nano timestamps

OneSizeFitsQuorum (approved):
LGTM

OneSizeFitsQuorum:
> No, the observer only publishes events that have greater nano timestamps

got it my lord

image
image

PR #13217, opened 2024-08-19
[region migration] Fix regionRWLock NPE when migrating region concurrently

HxpSerein:
## Description

- Capture the NPE below and enable write retry

img_v3_02dt_223cbb1a-5544-4bd2-8765-d35240c8513g

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/executor/RegionExecutionResult.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/executor/RegionExecutionResult.java
@@ -30,6 +30,17 @@ public class RegionExecutionResult {
   private TSStatus status;
   private boolean needRetry;
 
+  private RegionExecutionResult(boolean accepted, String message, TSStatus status) {
+    this.accepted = accepted;
+    this.message = message;
+    this.status = status;
+    this.needRetry = false;
+  }
+
+  public static RegionExecutionResult create(boolean accepted, String message, TSStatus status) {
+    return new RegionExecutionResult(accepted, message, status);
+  }
+
   public boolean isAccepted() {
     return accepted;
   }
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/executor/RegionExecutionResult.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/executor/RegionExecutionResult.java
@@ -28,13 +28,13 @@ public class RegionExecutionResult {
   private String message;
 
   private TSStatus status;
-  private boolean needRetry;
+  private boolean readNeedRetry;
 
   private RegionExecutionResult(boolean accepted, String message, TSStatus status) {
     this.accepted = accepted;
     this.message = message;
     this.status = status;
-    this.needRetry = false;
+    this.readNeedRetry = false;
   }
 
   public static RegionExecutionResult create(boolean accepted, String message, TSStatus status) {
@@ -65,11 +65,11 @@ public class RegionExecutionResult {
     this.status = status;
   }
 
-  public boolean isNeedRetry() {
-    return needRetry;
+  public boolean isReadNeedRetry() {
+    return readNeedRetry;
   }
 
-  public void setNeedRetry(boolean needRetry) {
-    this.needRetry = needRetry;
+  public void setReadNeedRetry(boolean readNeedRetry) {
+    this.readNeedRetry = readNeedRetry;
   }
 }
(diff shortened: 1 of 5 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/executor/RegionWriteExecutor.java
```diff
@@ -242,24 +242,27 @@ public RegionExecutionResult visitPipeEnrichedInsertNode(
...
+
         TSStatus status = fireTriggerAndInsert(context.getRegionId(), insertNode);
         response.setAccepted(TSStatusCode.SUCCESS_STATUS.getStatusCode() == status.getCode());
         response.setMessage(status.message);
         if (!response.isAccepted()) {
           response.setStatus(status);
         }
-        return response;
-      } catch (ConsensusException e) {
+      } catch (NullPointerException | ConsensusException e) {
```
liyuheng55555, line 255: I think it's better to check whether `Lock` is null independently, because `fireTriggerAndInsert` is a complicate function and catch NPE may hidden some other mistake.

HxpSerein: I think it is ok. It will log a warning and retry for other possible NPE.

OneSizeFitsQuorum: do not catch npe,it's ugly.

@liyuheng55555 you can just fix this as we discussed before~

HxpSerein: ok, I will close this pr

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/executor/RegionExecutionResult.java
```diff
@@ -30,6 +30,17 @@ public class RegionExecutionResult {
...
+  private RegionExecutionResult(boolean accepted, String message, TSStatus status) {
+    this.accepted = accepted;
+    this.message = message;
+    this.status = status;
+    this.needRetry = false;
+  }
+
+  public static RegionExecutionResult create(boolean accepted, String message, TSStatus status) {
+    return new RegionExecutionResult(accepted, message, status);
+  }
```
liyuheng55555, line 42: What's the difference between RegionExecutionResult.create and calling new RegionExecutionResult directly?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/executor/RegionWriteExecutor.java
```diff
@@ -241,22 +238,25 @@ public RegionExecutionResult visitPipeEnrichedInsertNode(
 
     private RegionExecutionResult executeDataInsert(
         InsertNode insertNode, WritePlanNodeExecutionContext context) {
-      RegionExecutionResult response = new RegionExecutionResult();
+      if (context.getRegionWriteValidationRWLock() == null) {
+        String message = "Failed to get the lock of the region because the region is not existed.";
+        return RegionExecutionResult.create(
+            false, message, RpcUtils.getStatus(TSStatusCode.WRITE_PROCESS_ERROR, message));
+      }
+
```
liyuheng55555, line 246: Is it necessary to set `needRetry` field?

HxpSerein: Not necessary, `needRetry` field is for read retry and is not required for write.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/executor/RegionExecutionResult.java
HxpSerein: `RegionExecutionResult.create` provides a layer of abstraction for object creation, encapsulating the details of object construction. Directly calling `new RegionExecutionResult` is a straightforward way to create objects, suitable for simpler scenarios. It makes the code more intuitive but offers less flexibility.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/executor/RegionWriteExecutor.java
liyuheng55555: @HxpSerein Then maybe rename it to `readNeedRetry` ?

HxpSerein: Good point. Fixed.

liyuheng55555 (approved):
LGTM. Migration + Write should be more robust now.

OneSizeFitsQuorum (approved):
LGTM

PR #13221, opened 2024-08-19
Avoid the brain split phenomenon in the symmetric network partition scenario

OneSizeFitsQuorum:
The current Ratis module's Leader may step down voluntarily without knowing who the new Leader is, which will not trigger the state machine's notifyLeaderChange callback. As a result, some modules that rely on this interface to determine whether the current node is no longer the Leader might delay resource release, potentially causing split-brain issues with multiple Leaders.

image

image

For example, in a 3-node ConfigNode setup, if a symmetric network partition fault is injected into the Leader node, the other two nodes will elect a new Leader. However, certain services (such as heartbeat, procedure, etc.) on the old Leader will not be cleared, leading to a split-brain scenario, which could cause some unexpected behavior.

image

After this PR, even if the new Leader is unknown, Ratis will still call the notifyNotReady function, thereby preventing split-brain issues from occurring.


liyuheng55555 (approved):
LGTM

SteveYurongSu (approved):

CRZbulabula (approved):
LGTM! Nice work for improving the overall robustness

HxpSerein (approved):
LGTM~

133tosakarin:
LGTM

PR #13234, opened 2024-08-20
PipeConsensus: build separate dir for each receiver thread.

Pengzna:
as title.

After this pr, every thread response for receiving tsFilePiece will has its own writing dir and file.

For example, if there are 5 tsFileWriters(which means the parallel degree/pipeline of PipeConsensusReceiver for receiving tsFile is 5), these 5 tsFileWriters will have its own dir naming from 0 - 4 respectively
image

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/receiver/protocol/pipeconsensus/PipeConsensusReceiver.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/receiver/protocol/pipeconsensus/PipeConsensusReceiver.java
@@ -101,8 +101,7 @@ public class PipeConsensusReceiver {
   private final ConsensusPipeName consensusPipeName;
   private final List<String> receiverBaseDirsName;
   // Used to buffer TsFile when transfer TsFile asynchronously.
-  private final PipeConsensusTsFileWriterPool pipeConsensusTsFileWriterPool =
-      new PipeConsensusTsFileWriterPool();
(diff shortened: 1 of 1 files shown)


Caideyipi (approved):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/receiver/protocol/pipeconsensus/PipeConsensusReceiver.java
```diff
@@ -1112,8 +1123,38 @@ private static class PipeConsensusTsFileWriter {
...
+    public PipeConsensusTsFileWriter(int index, ConsensusPipeName consensusPipeName) {
       this.index = index;
+      this.consensusPipeName = consensusPipeName;
+    }
+
+    public void setFilePath(String receiverBasePath) throws IOException {
+      this.localWritingDirPath = receiverBasePath + File.separator + index;
+      File tsFileWriterDirectory = new File(this.localWritingDirPath);
+      // Remove exists dir
+      if (tsFileWriterDirectory.exists()) {
```
OneSizeFitsQuorum, line 1135: Might this result in frequent directory creation and deletion? If so, it doesn't seem necessary?

Pengzna: @OneSizeFitsQuorum  It will not cause frequent file creation and deletion. This code will only be called when IoTDB is started to ensure that the dir is clean every time IoTDB restarts.

PR #13236, opened 2024-08-20
Make DN wait when CN cluster's leader is not ready or down.

Pengzna:
as title.

there is a case: when **CN1** leader is killed -9, The CN cluster may not be aware of this **immediately**, causing requests sent to the cluster to still be redirected to **CN1**. Due to the lack of waiting machinism in DN, DN may exhaust all retry count(10 times) in a very short time(about 500ms in 16C32G server) and all these retry requests will still be redirected to the 
dead node CN1, causing failure of user request.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/client/ConfigNodeClient.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/client/ConfigNodeClient.java
@@ -173,14 +173,15 @@ public class ConfigNodeClient implements IConfigNodeRPCService.Iface, ThriftClie
 
   private static final Logger logger = LoggerFactory.getLogger(ConfigNodeClient.class);
 
-  private static final int RETRY_NUM = 10;
+  private static final int RETRY_NUM = 15;
 
   public static final String MSG_RECONNECTION_FAIL =
       "Fail to connect to any config node. Please check status of ConfigNodes or logs of connected DataNode";
 
   private static final String MSG_RECONNECTION_DATANODE_FAIL =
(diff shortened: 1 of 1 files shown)


Pengzna:
@OneSizeFitsQuorum @liyuheng55555 PTAL

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/client/ConfigNodeClient.java
```diff
@@ -374,6 +376,22 @@ private <T> T executeRemoteCallWithRetry(final Operation<T> call, final Predicat
...
         configLeader = null;
       }
+
+      // If we have detected all configNodes and still not return
+      if (detectedNodeNum >= configNodes.size()) {
+        // Clear count
+        detectedNodeNum = 0;
+        // Wait to start the next try
+        try {
+          Thread.sleep(RETRY_INTERVAL_MS);
```
liyuheng55555, line 386: If there are 3 ConfigNodes, then the sleep time will be 10➗3 = 3. Maybe 3 seconds sleep is too short?

OneSizeFitsQuorum: Perhaps half of confignode's ratis election time is reasonable

OneSizeFitsQuorum: And we can change the retry num to 15？

Pengzna: fixed

OneSizeFitsQuorum (approved):
LGTM

PR #13250, opened 2024-08-21
Pipe: Fix reference count leak when tasks restart

SteveYurongSu:
As title.


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/event/EnrichedEvent.java
```diff
@@ -108,27 +108,32 @@ protected EnrichedEvent(
...
     synchronized (this) {
       if (isReleased.get()) {
         LOGGER.warn(
             "re-increase reference count to event that has already been released: {}, stack trace: {}",
             coreReportMessage(),
             Thread.currentThread().getStackTrace());
         isSuccessful = false;
-        // Here we still increase the reference count, to remain consistent with the behavior after
-        // internal increase failure.
-        referenceCount.incrementAndGet();
```
VGalaxies, line 120: don't forget to modify the corresponding javadoc...

PR #13254, opened 2024-08-22
Introduce AINode to IoTDB

ycycse:
Introduce AINode Module to IoTDB, it just a preliminary version, and there will be more refactoring and new features introduced in the future.


iotdb-core/pom.xml
```diff
@@ -38,4 +38,12 @@
...
         <module>relational-grammar</module>
     </modules>
+    <profiles>
+        <profile>
+            <id>with-ainode</id>
+            <modules>
+                <module>ainode</module>
+            </modules>
+        </profile>
+    </profiles>
```
HTHou, line 48: Good job!
Is there any plan to add a profile in distribution module? When `with-ainode` is activated, the package can be packed in distribution.

ycycse: That makes sense! I will try that.

ycycse: > Good job! Is there any plan to add a profile in distribution module? When `with-ainode` is activated, the package can be packed in distribution.

Hi! I've found some aspects of this feature that are worth discussing, so I plan to develop it as a new feature after this PR is merged, which will facilitate our discussion.

SteveYurongSu (approved):

PR #13269, opened 2024-08-22
Refactor/singleton 1

chrisdutz:
## Description

This branch should be used to track the progress of moving away from using singletons.

In a first step I wrote some rules and added a new profile "with-architectural-analysis" that spots Singletons and their usage and reports that.

In order to collect this data during the build, be sure to run a build with that profile enabled:

    mvn -P with-architectural-analysis clean verify

When building with enabled profile, the JQAssistant plugin will automatically analyse the project and collect all data in a Neo4J database which is located in "target/jqassistant". After the build with 

    mvn -P with-architectural-analysis jqassistant:report 

A nice HTML report can be generated.
When running:

    mvn -P with-architectural-analysis jqassistant:server

A Neo4j server is started alongside the Neo4J web-UI. The user can then use the browser to analyse all sort of different things in IoTDB using the Neo4J graph query language: Cypher.


liyuheng55555:
Hi Chris, I fetched your branch, tried `mvn -P with-architectural-analysis jqassistant:report`, but cannot found the generated html file. Where is this "nice HTML report" located at?

chrisdutz:
it should be in "target/site/" something ...

liyuheng55555:
@chrisdutz still cannot find it... Which package's target directory are you referring to? Besides, I currently only called mvn -P with-architectural-analysis jqassistant:report. Do I need to build the distribution or execute additional commands beyond this?

chrisdutz:
image

So ... these two steps are needed:

    mvn -P with-architectural-analysis clean verify

To collect the data, then:

    mvn -P with-architectural-analysis jqassistant:report 

To produce the report as seen in my screenshot above.

liyuheng55555:
image

image

@chrisdutz Thanks for the guide, now I find the site dir, and `mvn -P with-architectural-analysis jqassistant:server` is able to connect to the server. By the way, can you provide some example query for neo4j? Chatgpt give me something which seems not working:
```
MATCH (singleton:Type:Singleton)
RETURN singleton
ORDER BY singleton.fqn
```

PR #13282, opened 2024-08-23
[region migration] Ratis delete local peer when resetPeerList is called and myself not in correct peer list

liyuheng55555:


OneSizeFitsQuorum (approved):
LGTM

PR #13286, opened 2024-08-23
[Fix] Add another retryPolicy for RatisClient

133tosakarin:
For RatisClient returns reply,  we need to handle these execption in different policy.

In this pr, we use EndLessRetryPolicy and LimitRetryPlicy


OneSizeFitsQuorum (approved):
LGTM

PR #13290, opened 2024-08-23
Pipe: Avoid req size excceeding the thrift max frame size by slicing the original req automatically

SteveYurongSu:
As title.


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/receiver/protocol/thrift/IoTDBDataNodeReceiver.java
```diff
@@ -148,111 +152,185 @@ public class IoTDBDataNodeReceiver extends IoTDBFileReceiver {
...
     try {
-      long startTime = System.nanoTime();
+      final long startTime = System.nanoTime();
       final short rawRequestType = req.getType();
       if (PipeRequestType.isValidatedRequestType(rawRequestType)) {
-        TPipeTransferResp resp;
-        switch (PipeRequestType.valueOf(rawRequestType)) {
+        final PipeRequestType requestType = PipeRequestType.valueOf(rawRequestType);
+        if (requestType != PipeRequestType.TRANSFER_SLICE) {
+          sliceReqHandler.clear();
```
Caideyipi, line 160: Is it OK to clear here? What about two sequential slices?

Caideyipi:
Better add some ser/de test and ITs.

PR #13302, opened 2024-08-26
Refactor/ai node build update

chrisdutz:
## Description
The current build configuration was quite rough and will probably cause problems especially during a release.
Also did it contain a number of configurations that seem to be relics from previous iterations. I've removed the obsolete things, updated the deprecated ones. Also have I switched the python build environment to using venv to be more robust in various setups.


iotdb-core/ainode/resources/syncPythonVersion.groovy
```diff
@@ -0,0 +1,38 @@
...
+
+import groovy.toml.TomlSlurper
+
+def currentMavenVersion = project.version as String
+def currentPythonVersion = currentMavenVersion
+if(currentMavenVersion.contains("-SNAPSHOT")) {
+    currentPythonVersion = currentMavenVersion.split("-SNAPSHOT")[0] + ".dev"
+}
+
+def pyprojectFile = new File("pyproject.toml")
```
ycycse, line 28: I set this too absolute path because `../pyproject.toml` doesn't work either.

iotdb-core/ainode/pom.xml
```diff
@@ -56,81 +56,73 @@
...
+                    -->
                     <execution>
+                        <id>sync-python-version</id>
+                        <phase>validate</phase>
                         <goals>
-                            <goal>revision</goal>
+                            <goal>execute</goal>
                         </goals>
+                        <configuration>
+                            <source>resources/syncPythonVersion.groovy</source>
```
ycycse, line 108:
```suggestion
                            <source>${basedir}/resources/syncPythonVersion.groovy</source>
```

ycycse (changes requested):
This approach can effectively solved the building problem!
But when I tried to build ainode in my machine, I got problem as following when running `mvn clean package -pl iotdb-core/ainode -am -P with-ainode`:
```
Execution sync-python-version of goal org.codehaus.gmaven:groovy-maven-plugin:2.1.1:execute failed: No such property: resources for class: script1
No such property: resources for class: script1
---
Execution sync-python-version of goal org.codehaus.gmaven:groovy-maven-plugin:2.1.1:execute failed: java.nio.file.NoSuchFileException: pyproject.toml
```
And I got these two problems fixed by modify the path in`iotdb-core/ainode/pom.xml` and `iotdb-core/ainode/resource/syncpythonversion.groovy` (I will mark then in signle comment.)

So we may need to confirm whether it's my usage that's incorrect or if the paths in these two places need to be modified.

iotdb-core/ainode/resources/syncPythonVersion.groovy
chrisdutz: Yeah ... there was an issue that only occured when building with the full reactor ... I fixed that. And extended the script quite a bit.

iotdb-core/ainode/pom.xml
chrisdutz: Yeah ... that was the other thing I found ;-) Should already be fixed, if you update.

ycycse (approved):
LGTM~

.gitignore
```diff
@@ -123,3 +123,4 @@ iotdb-core/tsfile/src/main/antlr4/org/apache/tsfile/parser/gen/
 # Develocity
 .mvn/.gradle-enterprise/
 .mvn/.develocity/
+/iotdb-core/ainode/poetry.lock
```
HTHou, line 126: It seems that the dependabot required a poetry.lock file to create PR. Should we add it?
https://github.com/apache/iotdb/security/dependabot/294

chrisdutz: I just had a look at the PLC4X plc4go module, we do check in the go.sum file (which is something similar) ... I guess another type of project worth having a look would be node-based ones.

But sonsidering the volatility of versions in that space (possibly also in python space) it might be a good idea to check in the file as well.

chrisdutz: Just re-read your comment ... I just added the poetry.lock file and removed it from the gitignore.

PR #13315, opened 2024-08-27
Add clusterId check when restarting nodes

HxpSerein:
## Description

Enhance the correctness check during node restart.

### DIFF
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/node/ClusterNodeStartUtils.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/node/ClusterNodeStartUtils.java
@@ -325,12 +325,11 @@ public class ClusterNodeStartUtils {
               "Reject %s restart. Because the clusterId of the current %s and the target cluster are inconsistent. "
                   + "ClusterId of the current Node: %s, ClusterId of the target cluster: %s."
                   + POSSIBLE_SOLUTIONS
-                  + "\t1. Change the cluster_id parameter in %s to match the target cluster.",
+                  + "\t1. Please check if the node configuration or path is correct.",
               nodeType.getNodeType(),
               nodeType.getNodeType(),
               clusterId,
-              configManager.getClusterManager().getClusterId(),
-              CommonConfig.SYSTEM_CONFIG_NAME));
+              configManager.getClusterManager().getClusterId()));
       return status;
     }
 
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/node/ClusterNodeStartUtils.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/node/ClusterNodeStartUtils.java
@@ -315,7 +315,7 @@ public class ClusterNodeStartUtils {
         break;
     }
 
-    // check clusterId if exists
+    // check clusterId if not empty
     if (clusterId != null
         && !clusterId.isEmpty()
         && !clusterId.equals(configManager.getClusterManager().getClusterId())) {
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/node/ClusterNodeStartUtils.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/node/ClusterNodeStartUtils.java
@@ -211,6 +211,7 @@ public class ClusterNodeStartUtils {
   public static TSStatus confirmNodeRestart(
       NodeType nodeType,
       String clusterName,
+      String clusterId,
       int nodeId,
       Object nodeLocation,
       ConfigManager configManager) {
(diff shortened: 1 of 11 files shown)


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/node/ClusterNodeStartUtils.java
```diff
@@ -314,6 +315,25 @@ public static TSStatus confirmNodeRestart(
...
+    if (clusterId != null
+        && !clusterId.isEmpty()
+        && !clusterId.equals(configManager.getClusterManager().getClusterId())) {
+      status.setCode(TSStatusCode.REJECT_NODE_START.getStatusCode());
+      status.setMessage(
+          String.format(
+              "Reject %s restart. Because the clusterId of the current %s and the target cluster are inconsistent. "
+                  + "ClusterId of the current Node: %s, ClusterId of the target cluster: %s."
+                  + POSSIBLE_SOLUTIONS
+                  + "\t1. Change the cluster_id parameter in %s to match the target cluster.",
```
liyuheng55555, line 328: “Solution" might be not proper here, because we don't want user to modify system.properties. Maybe just suggest the user to check if the node configuration is correct?

liyuheng55555 (changes requested):
I think there is no more need to pass cluster id in DN restart response now, please check 363 line of NodeManager and remove it.

OneSizeFitsQuorum (commented):
Missing confignode?

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/node/ClusterNodeStartUtils.java
HxpSerein: Fixed

HxpSerein:
> Missing confignode?

At present, when the ConfigNode is restarted, it doesn't send an RPC to the ConfigNode leader. I've implemented persistent storage for the clusterId on the ConfigNode and added a TODO during the restart phase, laying the groundwork for further enhancements in the future.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/node/NodeManager.java
```diff
@@ -421,9 +421,11 @@ public TConfigNodeRegisterResp registerConfigNode(TConfigNodeRegisterReq req) {
     int nodeId = nodeInfo.generateNextNodeId();
     req.getConfigNodeLocation().setConfigNodeId(nodeId);
     configManager.getProcedureManager().addConfigNode(req);
+    String clusterId = configManager.getClusterManager().getClusterId();
     return new TConfigNodeRegisterResp()
         .setStatus(ClusterNodeStartUtils.ACCEPT_NODE_REGISTRATION)
-        .setConfigNodeId(nodeId);
+        .setConfigNodeId(nodeId)
+        .setClusterId(clusterId);
```
liyuheng55555, line 428: Not quite necessary here, CN's consensus layer will automatically sync cluster id

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/DataNode.java
```diff
@@ -449,7 +446,7 @@ private void storeRuntimeConfigurations(
     initTTLInformation(runtimeConfiguration.getAllTTLInformation());
 
     /* Store cluster ID */
-    IoTDBDescriptor.getInstance().getConfig().setClusterId(runtimeConfiguration.getClusterId());
+    storeClusterID(runtimeConfiguration.getClusterId());
```
liyuheng55555, line 449: After remove the cluster id from DN restart resp, NPE might happen here

liyuheng55555 (changes requested):
PTAL~

liyuheng55555:
> > Missing confignode?
> 
> At present, when the ConfigNode is restarted, it doesn't send an RPC to the ConfigNode leader. I've implemented persistent storage for the clusterId on the ConfigNode and added a TODO during the restart phase, laying the groundwork for further enhancements in the future.

Even if a wrong CN joining in accidently , I think it won't cause any serious problem. Remove it and add a new empty CN will fix everything. So for now it's fine to only verify DN's restart.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/node/NodeManager.java
HxpSerein: changed.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/DataNode.java
HxpSerein: fixed.

OneSizeFitsQuorum (approved):
LGTM

liyuheng55555 (approved):
LGTM. Those who add the wrong ConfigNode to the cluster, hope god bless them.

PR #13341, opened 2024-08-29
Get root cause to enable read retry

HxpSerein:
## Description

- use `getRootCause` instead of `getCause`


OneSizeFitsQuorum (approved):
LGTM

PR #13353, opened 2024-08-30
Change FileTimeIndexCache to region level

HTHou:
## Description

FileTimeIndexCache may open too much file if there is too much time partitions.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/tsfile/timeindex/FileTimeIndexCacheRecorder.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/tsfile/timeindex/FileTimeIndexCacheRecorder.java
@@ -35,7 +35,6 @@ import org.slf4j.LoggerFactory;
 import java.io.File;
 import java.io.IOException;
 import java.nio.ByteBuffer;
-import java.nio.file.Files;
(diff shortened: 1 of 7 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/tsfile/timeindex/FileTimeIndexCacheRecorder.java
```diff
@@ -57,8 +57,7 @@ public class FileTimeIndexCacheRecorder {
 
   private final BlockingQueue<Runnable> taskQueue = new LinkedBlockingQueue<>();
 
-  private final Map<Integer, Map<Long, FileTimeIndexCacheWriter>> writerMap =
-      new ConcurrentHashMap<>();
+  private final Map<Integer, FileTimeIndexCacheWriter> writerMap = new ConcurrentHashMap<>();
```
jt2594838, line 60: Add a comment to explain that the key is the data region id.

jt2594838 (approved):

jt2594838:
One potential improvement: the compaction task clears the cache file at first. If a restart happens at that time, there will be no available cache file for this restart, and the system may suffer from a long restart time.
Maybe it would be better to adopt the write-a-temp-file-and-replace method.

PR #13355, opened 2024-08-30
Rc/1.3.3 flush cli output

chrisdutz:
## Description
Made sure the ouput is flushed, so it appears when being run in IntelliJ


OneSizeFitsQuorum (approved):
LGTM

PR #13359, opened 2024-08-30
Fix WAL Exception catch & support new DataTypes in QueryDataSetUtils

jt2594838:
One Tablet deserialization does not support new types in QueryDataSetUtils, which causes WAL recovery failure.
TsDataType switches are supplemented with all cases to indicate they are CONSIDERED rather than FORGOTTEN.

### DIFF
--- a/integration-test/src/main/java/org/apache/iotdb/db/query/udf/example/UDAFAvg.java
+++ b/integration-test/src/main/java/org/apache/iotdb/db/query/udf/example/UDAFAvg.java
@@ -104,6 +104,8 @@ public class UDAFAvg implements UDAF {
       case STRING:
       case BLOB:
       case BOOLEAN:
+      case TIMESTAMP:
+      case DATE:
       default:
         throw new UnSupportedDataTypeException(
             String.format("Unsupported data type in aggregation AVG : %s", dataType));
--- a/integration-test/src/main/java/org/apache/iotdb/db/query/udf/example/UDAFAvg.java
+++ b/integration-test/src/main/java/org/apache/iotdb/db/query/udf/example/UDAFAvg.java
@@ -101,6 +101,8 @@ public class UDAFAvg implements UDAF {
         addDoubleInput(avgState, columns, bitMap);
         return;
       case TEXT:
+      case STRING:
+      case BLOB:
       case BOOLEAN:
       default:
         throw new UnSupportedDataTypeException(
--- a/integration-test/src/main/java/org/apache/iotdb/db/query/udf/example/UDAFSum.java
+++ b/integration-test/src/main/java/org/apache/iotdb/db/query/udf/example/UDAFSum.java
@@ -104,6 +104,8 @@ public class UDAFSum implements UDAF {
       case STRING:
       case BLOB:
       case BOOLEAN:
+      case TIMESTAMP:
+      case DATE:
       default:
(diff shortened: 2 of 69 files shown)


integration-test/src/main/java/org/apache/iotdb/db/query/udf/example/UDAFAvg.java
```diff
@@ -101,6 +101,8 @@ public void addInput(State state, Column[] columns, BitMap bitMap) {
         addDoubleInput(avgState, columns, bitMap);
         return;
       case TEXT:
+      case STRING:
+      case BLOB:
```
shuwenwei, line 105: Missing data type TIMESTAMP and DATE

integration-test/src/main/java/org/apache/iotdb/db/query/udf/example/UDAFSum.java
```diff
@@ -101,6 +101,8 @@ public void addInput(State state, Column[] columns, BitMap bitMap) {
         addDoubleInput(sumState, columns, bitMap);
         return;
       case TEXT:
+      case STRING:
+      case BLOB:
```
shuwenwei, line 105: Missing data type TIMESTAMP and DATE

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/udf/builtin/UDTFAbs.java
```diff
@@ -89,6 +97,14 @@ public Object transform(Row row) throws IOException {
         return Math.abs(row.getFloat(0));
       case DOUBLE:
         return Math.abs(row.getDouble(0));
+      case VECTOR:
+      case UNKNOWN:
```
shuwenwei, line 101: Should we add VECTOR and UNKNOWN types to other switch case statements?

iotdb-core/node-commons/src/assembly/resources/conf/iotdb-system.properties
```diff
@@ -27,21 +27,21 @@ cluster_name=defaultCluster
 ### Seed ConfigNode
 ####################
 
-cn_seed_config_node=127.0.0.1:10710
+cn_seed_config_node=nel-22:10710
```
shuwenwei, line 30: It also needs to be modified in iotdb-system.properties.template

shuwenwei (approved):

integration-test/src/main/java/org/apache/iotdb/db/query/udf/example/UDAFSum.java
jt2594838: Added.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/udf/builtin/UDTFAbs.java
jt2594838: Removed.

iotdb-core/node-commons/src/assembly/resources/conf/iotdb-system.properties
jt2594838: Reverted.

shuwenwei (approved):

PR #13360, opened 2024-08-31
Pipe: release resources of pipe event that have been GCed but the reference count is not zero by PhantomReference

VGalaxies:
Considering that when the event is garbage collected, the file resources it holds will not automatically end their lifecycle, we adopt a reference counting maintenance + safety net fallback method (releasing resources later is better than never releasing them).


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/resource/ref/PipePhantomReferenceManager.java
```diff
@@ -0,0 +1,137 @@
...
+
+  public PipePhantomReferenceManager() {
+    // Do nothing now.
+  }
+
+  public int getPhantomReferenceCount() {
+    return PIPE_EVENT_PHANTOM_REFERENCES.size();
+  }
+
+  protected void gcHook() {
```
SteveYurongSu, line 52: Use a seperated thread pool (extends PipePeriodicalJobExecutor?) to run this method periodically?

SteveYurongSu: For one thread, what is the max number of events per second can be checked?

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/resource/ref/PipePhantomReferenceManager.java
```diff
@@ -0,0 +1,137 @@
...
+
+    private final String coreReportMessageSnapshot;
+    private PipeEventResource resource;
+
+    private PipeEventPhantomReference(
+        final EnrichedEvent event,
+        final PipeEventResource resource,
+        final ReferenceQueue<? super EnrichedEvent> queue) {
+      super(event, queue);
+      this.coreReportMessageSnapshot = event.coreReportMessage();
```
SteveYurongSu, line 87: Calling coreReportMessage is time consuming ...

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/resource/ref/PipePhantomReferenceManager.java
SteveYurongSu: We have to consider using a multi-threaded pool to run this method if necessary (maybe we can add a conf parma for this?)

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/resource/ref/PipePhantomReferenceManager.java
```diff
@@ -0,0 +1,137 @@
...
+      this.referenceCount = referenceCount;
+    }
+
+    private void clearReferenceCount(final String holderMessage) {
+      if (isReleased.get()) {
+        return;
+      }
+
+      if (referenceCount.get() >= 1) {
+        LOGGER.info("finalize resource for event with holder message {}", holderMessage);
```
SteveYurongSu, line 127:
```suggestion
        LOGGER.error("finalize resource for event with holder message {}", holderMessage);
```
"RESOURCE LEAK DETECTED: {}....", class name...

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/resource/ref/PipePhantomReferenceManager.java
SteveYurongSu, line 52: add a check here: if this run is enabled by conf

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/conf/CommonDescriptor.java
```diff
@@ -604,6 +604,17 @@ private void loadPipeProps(Properties properties) {
...
+
+    config.setPipeEventReferenceTrackingEnabled(
+        Boolean.parseBoolean(
+            properties.getProperty(
+                "pipe_event_reference_tracking_enabled",
+                String.valueOf(config.getPipeEventReferenceTrackingEnabled()))));
+    config.setPipeEventReferenceEliminateIntervalSeconds(
+        Long.parseLong(
+            properties.getProperty(
+                "pipe_event_reference_eliminate_interval_seconds",
```
SteveYurongSu, line 616: Can be shorter by default?

VGalaxies: no other ideas for now...

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/resource/ref/PipePhantomReferenceManager.java
VGalaxies: consider optimizing in the next PR...

SteveYurongSu (approved):

PR #13361, opened 2024-08-31
Decrease default direct buffer size in IT config

shuwenwei:
## Description
Decrease default direct buffer size in IT config


OneSizeFitsQuorum (approved):
LGTM

PR #13362, opened 2024-08-31
[to rc/1.3.3] Decrease default direct buffer size in it config

shuwenwei:
## Description
Decrease default direct buffer size in it config


OneSizeFitsQuorum (approved):
LGTM

PR #13398, opened 2024-09-04
Fix error log when drop database or migrate region

HTHou:
## Description

When execute the following sqls, there is some error log printed.
```
insert into root.sg.d1(time,s1) values(1,2)
flush
drop database root.**
```
image

This pr is fixing this issue. 
image

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/tsfile/timeindex/FileTimeIndexCacheRecorder.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/tsfile/timeindex/FileTimeIndexCacheRecorder.java
@@ -193,7 +193,7 @@ public class FileTimeIndexCacheRecorder {
   }
 
   public void removeFileTimeIndexCache(int dataRegionId) {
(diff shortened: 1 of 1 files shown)


jt2594838 (approved):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/tsfile/timeindex/FileTimeIndexCacheRecorder.java
```diff
@@ -193,10 +193,11 @@ public void close() throws IOException {
   }
 
   public void removeFileTimeIndexCache(int dataRegionId) {
-    for (FileTimeIndexCacheWriter writer : writerMap.values()) {
+    FileTimeIndexCacheWriter writer = writerMap.get(dataRegionId);
```
OneSizeFitsQuorum, line 196: We need to maintain the writerMap when there are region migrations and drop database

HTHou: fixed

PR #13399, opened 2024-09-04
Pipe: Remove and close asynchronous connector manager when all clients are useless

YC27:


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/client/IoTDBDataNodeAsyncClientManager.java
```diff
@@ -94,6 +98,7 @@ public IoTDBDataNodeAsyncClientManager(
       }
     }
     endPoint2Client = ASYNC_PIPE_DATA_TRANSFER_CLIENT_MANAGER_HOLDER.get(receiverAttributes);
+    attributesCount.compute(receiverAttributes, (k, v) -> v == null ? 1 : v + 1);
```
luoluoyuyu, line 101: This step is best placed in a synchronized code block

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/protocol/thrift/async/IoTDBDataRegionAsyncConnector.java
```diff
@@ -539,9 +539,20 @@ public synchronized void close() {
...
     super.close();
   }
 
+  private void decreaseOrRemoveAsyncClient() {
+    clientManager.decreaseAttributesCount();
+
+    if (clientManager.getAttributesCount() == 0) {
+      clientManager.removeAttributesCount();
+      clientManager.closeAndRemoveClientManager();
+    }
```
luoluoyuyu, line 553: This function is best placed in `IoTDBDataNodeAsyncClientManager` because the `put` code is in `IoTDBDataNodeAsyncClientManager`, and the `remove` code is also best placed in `IoTDBDataNodeAsyncClientManager`, which can maintain a better code structure. And the code in this function needs to be synchronized using the `IoTDBDataRegionAsyncConnector` object lock

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/client/IoTDBDataNodeAsyncClientManager.java
luoluoyuyu, line 101: Please use the same lock for put and remove operations, and ensure the atomicity of adding and increasing counts, removing and decreasing counts.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/protocol/thrift/async/IoTDBDataRegionAsyncConnector.java
luoluoyuyu, line 553: Also, it is better to use the `IoTDBDataRegionAsyncConnector.class` lock instead of the `IoTDBDataRegionAsyncConnector` instance lock. The put and remove operations and the corresponding reference counts should be mutually exclusive.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/client/IoTDBDataNodeAsyncClientManager.java
YC27: @luoluoyuyu I tested it and found that using static + Concurrent can ensure thread safety.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/protocol/thrift/async/IoTDBDataRegionAsyncConnector.java
YC27: @luoluoyuyu It adds the synchronized keyword to the entire close method

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/client/IoTDBDataNodeAsyncClientManager.java
luoluoyuyu: This step can only guarantee the atomicity of updating the reference. Here, it is necessary to ensure that the createClient and update reference steps are synchronized.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/protocol/thrift/async/IoTDBDataRegionAsyncConnector.java
luoluoyuyu: CreateClient and removeClient do not use the same lock, so it may cause a thread to successfully update CreateClient but fail to update the reference, causing the Client to be deleted when removing, causing NPE

luoluoyuyu (approved):

SteveYurongSu (approved):

SteveYurongSu (approved):

luoluoyuyu (approved):

PR #13405, opened 2024-09-04
Subscription: reduce the length of the stringified subscription event & do not print warn logs for eagerly pollable events

VGalaxies:
As title.


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/event/batch/SubscriptionPipeTabletEventBatch.java
```diff
@@ -201,15 +201,21 @@ public String toString() {
...
     coreReportMessage.put("size of tablets", String.valueOf(tablets.size()));
     coreReportMessage.put("firstEventProcessingTime", String.valueOf(firstEventProcessingTime));
     coreReportMessage.put("totalBufferSize", String.valueOf(totalBufferSize));
     return coreReportMessage;
   }
+
+  private static String formatEnrichedEvents(
+      final List<EnrichedEvent> enrichedEvents, final int threshold) {
+    List<String> eventMessageList =
+        enrichedEvents.stream().map(EnrichedEvent::coreReportMessage).collect(Collectors.toList());
```
SteveYurongSu, line 214:
```suggestion
        enrichedEvents.stream().limit(threshold).map(EnrichedEvent::coreReportMessage).collect(Collectors.toList());
```

SteveYurongSu (approved):

PR #13418, opened 2024-09-05
Properly handle the case where a single IndexedRequest contains multiple “delete PlanNode” operations on the receiving end of IoTConsensus.

liyuheng55555:
A delete operation involving multiple paths or multiple Regions is split into several PlanNodes for execution and synchronization. On the receiving end of IoTConsensus, these PlanNodes need to be merged back into a single PlanNode for execution.


OneSizeFitsQuorum (approved):
LGTM

PR #13456, opened 2024-09-09
Optimize column batch selection when there are many value columns

shuwenwei:
## Description
Optimize column batch selection when there are many value columns.
Column num is 20w.
## Before
截屏2024-09-09 19 01 23

## After
截屏2024-09-09 18 59 16

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/utils/executor/batch/utils/AlignedSeriesBatchCompactionUtils.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/utils/executor/batch/utils/AlignedSeriesBatchCompactionUtils.java
@@ -120,7 +120,7 @@ public class AlignedSeriesBatchCompactionUtils {
           && currentSchema
               .getMeasurementId()
               .equals(currentValueChunkMetadata.getMeasurementUid())) {
-        newValueChunkMetadataArr[i] = originAlignedChunkMetadata;
+        newValueChunkMetadataArr[i] = currentValueChunkMetadata;
         currentValueChunkMetadataIndex++;
       }
     }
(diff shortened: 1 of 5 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/utils/executor/batch/utils/AlignedSeriesBatchCompactionUtils.java
```diff
@@ -173,4 +172,75 @@ public static ModifiedStatus checkIsModified(
...
+                : largeTypeSortedColumnIndexList;
+        Integer columnIndex = sourceTypeSortedList.removeFirst();
+        this.columnIndexList.add(columnIndex);
+        this.currentSelectedColumnSchemaList.add(this.schemaList.get(columnIndex));
+        selectedColumnNum++;
+        if (this.columnIndexList.size() >= batchSize) {
+          break;
+        }
+      }
+    }
```
jt2594838, line 238: A queue seems to be more suitable for this job than LinkedList. Or you may remove a sublist from the list to reduce iteration.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/utils/executor/batch/utils/AlignedSeriesBatchCompactionUtils.java
```diff
@@ -173,4 +172,75 @@ public static ModifiedStatus checkIsModified(
...
+
+    private static boolean isLargeDataType(byte dataType) {
+      return dataType == TSDataType.TEXT.getType()
+          || dataType == TSDataType.STRING.getType()
+          || dataType == TSDataType.BLOB.getType();
+    }
+  }
```
jt2594838, line 245: TSDataType.isBinary does the same. May use that to reduce redundant code.

jt2594838 (approved):

jt2594838 (approved):

jt2594838 (approved):

PR #13461, opened 2024-09-09
Feat/cli disable history

chrisdutz:
Currently the CLI has an option to write a history file, which has a fixed location in the current processes user's home directory. This is causing some problems in SNAP based deployments as the sandbox doesn't allow accessing files outside of the SNAP's sandbox. 

This PR adds an option to the CLI to set a variable in the context to disable this feature.


OneSizeFitsQuorum (approved):
LGTM

PR #13478, opened 2024-09-11
Remove redundant closing of consensus layer in DataNode shutdown process

liyuheng55555:
Besides, extract and inline some functions to make code looks better


OneSizeFitsQuorum (approved):
LGTM

PR #13487, opened 2024-09-12
Keep all null rows in table model

JackieTien97:
More discussions can be seen in mail list: https://lists.apache.org/thread/xjmbxh6jj513dfldg4x2ngh8nf5o1hvn


HTHou (approved):
lgtm

PR #13496, opened 2024-09-12
[Metrics] Fix the grafana panel display error caused by not initializing trigTime

133tosakarin:
Fix the grafana panel display error caused by not initializing trigTime


OneSizeFitsQuorum (approved):
LGTM

PR #13509, opened 2024-09-13
Fix the logic of catch ServerNotReadyException to enable read retry

HxpSerein:
## Description
- When a ServerNotReadyException occurs, Ratis will throw an IllegalStateException.

img_v3_02em_751b43f9-a823-44d3-a326-eaeba58df0fg

img_v3_02em_b697ef5c-eb89-4b89-8624-99e34ddc6e9g


OneSizeFitsQuorum (approved):
LGTM

PR #13517, opened 2024-09-14
Use CountDownLatch to replace Semaphore in IoTConsensus log dispatcher closing

liyuheng55555:
To avoid unnecessary waiting when close function is called more than one time


OneSizeFitsQuorum (approved):
LGTM~
Learned a lot

PR #13544, opened 2024-09-19
Pipe: make progress index immutable

VGalaxies:
In Java, immutable objects have clear requirements:

1. The state of the object cannot be modified after creation.
2. All member variables of the object are final.
3. The object is correctly created (during the creation of the object, the `this` reference does not escape).

This class is designed to be immutable, meaning that instances of ProgressIndex can be treated as value objects. Immutability ensures thread safety, as no external synchronization is required for concurrent access to instances of this class. It also simplifies reasoning about the state of the object as it cannot change once created.

However, if a ProgressIndex instance holds any mutable objects, like a Map, they must be deeply copied during construction or when exposed through accessors to maintain the immutability contract. This prevents unintended modifications to the underlying mutable state from affecting other parts of the program.


VGalaxies:
merge after #13542

SteveYurongSu (approved):

PR #13547, opened 2024-09-19
Fix AINode status error after ConfigNode Leader restarts

OneSizeFitsQuorum:
image


chrisdutz (approved):
LGTM (Tested it on a Google Compute Cloud installation and it resolved the issues)

PR #13570, opened 2024-09-20
Improve likespeed

linxt20:
This work is mainly part of the work to improve the efficiency of like matching, mainly including the following modifications:
- Modify the call to the template processing function in like, and delete the redundant code for template processing
- Use the new template class and matching class LikePattern and LikeMatcher at the same time, modify the call of the template class and the call of the matching method
- Distinguish the call of regexp and like, and implement likecolumntransformer and liketransformer
- Modify the use call of like in the original test, add the test of escape setting of like in the table model and the test of the functional correctness of like


pom.xml
```diff
@@ -40,6 +40,7 @@
         <module>distribution</module>
         <module>example</module>
         <module>library-udf</module>
+        <module>integration-test</module>
```
JackieTien97, line 43:
```suggestion
```

JackieTien97 (changes requested):

JackieTien97 (approved):

PR #13587, opened 2024-09-23
Feature/dependency check

chrisdutz:
Added a check to the build, that fails the build if the list of dependencies reported by the sbom project doesn't match the list on file in the "dependencies.json" file in the root of the project. This way PRs changing the dependencies also need to update this file, which makes it more obvious which dependencies have been added or removed from the build in a PR.

As the usage of the maven "-pl" and "-am" have a huge impact on the content the sbom plugin will report, it doesn't really make much sense to have this check active per default, so I've implemented the check to only be active by setting the property `dependencyCheck.skip` to true. So by adding `-DdependencyCheck.skip=true" to the commandline.

In order to ensure the check is run for a PR, I've added a new github-actions pipeline, which only has the goal to do this check.

Here's an example on what the build output produces with enabled check if I added one dependency (plc4j-spi) to the build without updating the dependencies.json file.

image


src/main/xslt/sbom-filter.xsl
```diff
@@ -0,0 +1,47 @@
...
+    xmlns:sbom="http://cyclonedx.org/schema/bom/1.4">
+    
+    <xsl:output
+        method="text"
+        indent="no"
+        encoding="utf-8"
+    />
+    
+    <xsl:template match="/">{<!--
+    "project": [<xsl:for-each select="sbom:bom/sbom:components/sbom:component[sbom:group='org.apache.iotdb']">
```
OneSizeFitsQuorum, line 33: Shall we remove this annotated lines?

chrisdutz: I thought I would leave it in there so it could be commented in quickly. But sure ... I'll remove it.

chrisdutz: Removed

OneSizeFitsQuorum (approved):
LGTM

PR #13588, opened 2024-09-23
Make log dispatcher stop in parallel

liyuheng55555:


OneSizeFitsQuorum (approved):
LGTM~
Although not useful for two replicas of most scenes, it is useful for three replicas

PR #13591, opened 2024-09-23
Fix print error log when dir is not exist

133tosakarin:
As described in the title


iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
```diff
@@ -774,6 +774,9 @@ public List<ConsensusGroupId> getAllConsensusGroupIds() {
   @Override
   public List<ConsensusGroupId> getAllConsensusGroupIdsWithoutStarting() {
     List<ConsensusGroupId> consensusGroupIds = new ArrayList<>();
+    if (!storageDir.exists()) {
+      return consensusGroupIds;
```
OneSizeFitsQuorum, line 778:
```suggestion
    if (!storageDir.exists()) {
      return Collections.emptyList();
      }
    List<ConsensusGroupId> consensusGroupIds = new ArrayList<>();

```

OneSizeFitsQuorum:
BTW, Please see whether there are similar problem in schema engine dir or dataengine dir

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
133tosakarin: ok

PR #13592, opened 2024-09-23
Feature/show ainode command

chrisdutz:
Implemented the "show ainodes" command for the table model.


OneSizeFitsQuorum (approved):
LGTM

iotdb-core/relational-grammar/src/main/antlr4/org/apache/iotdb/db/relational/grammar/sql/RelationalSql.g4
```diff
@@ -781,6 +786,7 @@ COMMITTED: 'COMMITTED';
 CONDITION: 'CONDITION';
 CONDITIONAL: 'CONDITIONAL';
 CONFIGNODES: 'CONFIGNODES';
+AINODES: 'AINODES';
```
Caideyipi, line 789: Maybe using lexicographical order here is better?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/execution/config/metadata/relational/ShowAINodesTask.java
```diff
@@ -0,0 +1,78 @@
...
+
+public class ShowAINodesTask implements IConfigTask {
+
+  @Override
+  public ListenableFuture<ConfigTaskResult> execute(IConfigTaskExecutor configTaskExecutor)
+      throws InterruptedException {
+    return configTaskExecutor.showAINodes();
+  }
+
+  public static void buildTSBlock(
```
Caideyipi, line 50: It may seem that there's no need to implement "buildTsBlocks" in relational tasks since they are not called :-)

Caideyipi (commented):
Good codes! But there seems to be a few points to improve ;-)

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/execution/config/metadata/relational/ShowAINodesTask.java
chrisdutz: But all of the other implementations all have this implemented? Is there a reason ShowConfigNodesTask and ShowDataNodesTask have it implemented? Or could it also be removed for them?

Caideyipi: Well, I guess if the method is uncalled it can be removed. Personally speaking, after the refactor, the method is left unused and may better be cleared?

PR #13600, opened 2024-09-24
Pipe: basic support for table model from SQL to IConfigTask

SteveYurongSu:
As title.


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/config/constant/PipeExtractorConstant.java
```diff
@@ -24,6 +24,12 @@ public class PipeExtractorConstant {
   public static final String EXTRACTOR_KEY = "extractor";
   public static final String SOURCE_KEY = "source";
 
+  public static final String EXTRACTOR_DIALECT_KEY = "extractor.dialect";
```
Caideyipi, line 27: Seemingly this is not decided by the model used in creation and is left for the users to fill in?

Caideyipi:
The newly added statements shall be placed in the coordinators to be directed to the configNodes.

SteveYurongSu:
All comments will be solved in next PR. Thanks for reviewing @Caideyipi

PR #13622, opened 2024-09-25
Fix NodeInfo processLoadSnapshot method (avoid compatibility issues caused by ainode serialization)

liyuheng55555:


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/node/NodeInfo.java
```diff
@@ -770,6 +774,17 @@ private void deserializeRegisteredDataNode(InputStream inputStream, TProtocol pr
     }
   }
 
+  private void tryDeserializeRegisteredAINode(
+      BufferedInputStream bufferedInputStream, TProtocol protocol) throws IOException {
+    try {
+      bufferedInputStream.mark(1024);
```
OneSizeFitsQuorum, line 780: Why use magic number 1024?

OneSizeFitsQuorum (approved):
LGTM

PR #13640, opened 2024-09-26
Pipe: refact pipe remaining metrics calculation

VGalaxies:
TBD.


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/broker/SubscriptionBlockingPendingQueue.java
```diff
@@ -40,4 +42,9 @@ public int size() {
   public boolean isEmpty() {
     return inputPendingQueue.isEmpty();
   }
+
+  /** DO NOT FORGET to set eventCounter to new value after invoking this method. */
```
VGalaxies, line 46: @SteveYurongSu what for?

PR #13649, opened 2024-09-27
Feature/move SBOM check into separate profile

chrisdutz:
Moves the sbom check into a dedicated profile, which needs to be enabled explicitly.


CritasWang (approved):
LGTM

PR #13651, opened 2024-09-27
Support global-aggregation in TableModel

Wei-hao-Li:
1. Now only support count, avg aggregation function;
2. ITs are not added now, we will do this work after all aggregation operators are developed.

image


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/source/relational/aggregation/AggregationOperator.java
```diff
@@ -0,0 +1,158 @@
...
+  @Override
+  public boolean hasNext() throws Exception {
+    return !finished;
+  }
+
+  @Override
+  public TsBlock next() throws Exception {
+    // Each call only calculate at most once, no need to check time slice.
+    TsBlock block;
+    if (child.hasNextWithTimer()) {
```
Beyyes, line 93: Add is resultTsBlock.isFull() judgement

JackieTien97 (approved):

PR #13673, opened 2024-10-01
Pipe: omit region id when marking collect invocation count for remaining time calculation

VGalaxies:
fixup #12799


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/subtask/processor/PipeProcessorSubtask.java
```diff
@@ -137,12 +139,14 @@ protected boolean executeOnce() throws Exception {
           pipeProcessor.process((TabletInsertionEvent) event, outputEventCollector);
           PipeProcessorMetrics.getInstance().markTabletEvent(taskID);
           PipeDataNodeRemainingEventAndTimeMetrics.getInstance()
-              .markCollectInvocationCount(taskID, outputEventCollector.getCollectInvocationCount());
+              .markCollectInvocationCount(
+                  pipeName + "_" + creationTime, outputEventCollector.getCollectInvocationCount());
```
SteveYurongSu, line 143: Cache this string for better performance

SteveYurongSu (changes requested):
See the comment

SteveYurongSu (approved):

PR #13674, opened 2024-10-02
Feature/make load active listening pipe dir settable

chrisdutz:
In order to dynamically set the IoTDB directories when running IoTDB in embedded mode, we need to make this property settable (which seems to be the only one that isn't settable right now)


OneSizeFitsQuorum (approved):
LGTM

PR #13678, opened 2024-10-08
Fix the order of stop schema region and clear schema engine

liyuheng55555:
Make schema region able to take snapshot while stopping


OneSizeFitsQuorum (approved):
LGTM

PR #13681, opened 2024-10-08
Fix the failure of update cluster name in metric framework of ConfigNode

SpriCoder:
Fix the failure of update cluster name in metric framework of ConfigNode, more details see: https://jira.infra.timecho.com:8443/browse/TIMECHODB-1004


OneSizeFitsQuorum (approved):
LGTM

PR #13683, opened 2024-10-08
Change database in SessionConnection while using db

JackieTien97:
While using `use XXX` in Session.executeNonQueryStatement, we should also update the database field in SessionConnection, because while retrying, we will reconnect to any node and then redo request which will use the local database field in SessionConnection.

So, we may see the following error msg in client if retrying phase happens.
img_v3_02ff_5789ce5d-b736-4be1-828e-abf01b99f38g


wangchao316 (approved):
LGTM

PR #13705, opened 2024-10-10
Load: Fix tsfile that was failed to load exceptionally deleted by LoadTsFileScheduler

YC27:
as title

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/load/LoadSingleTsFileNode.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/load/LoadSingleTsFileNode.java
@@ -202,16 +202,16 @@ public class LoadSingleTsFileNode extends WritePlanNode {
         + '}';
   }
 
-  public void clean() {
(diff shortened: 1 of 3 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/load/LoadSingleTsFileNode.java
```diff
@@ -202,16 +202,16 @@ public String toString() {
         + '}';
   }
 
-  public void clean() {
+  public void clean(final boolean isLoadSuccess) {
```
Caideyipi, line 205: Since a false "isLoadSuccess" will just skip the whole function, why don't we just judge outside?

YC27: > Since a false "isLoadSuccess" will just skip the whole function, why don't we just judge outside?

Both methods are acceptable. Two if statements need to be added to the external judgment. There are already many if statements in the calling place, so they are done in clean.

SteveYurongSu (approved):

PR #13714, opened 2024-10-10
Pipe: Support TablePattern matching in PipeDataRegionMatcher

SteveYurongSu:
As title.


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/extractor/dataregion/realtime/matcher/CachedSchemaPatternMatcher.java
```diff
@@ -117,54 +133,18 @@ public Set<PipeRealtimeDataRegionExtractor> match(final PipeRealtimeEvent event)
...
-          //
-          // case 1: the pattern can match all measurements of the device.
-          // in this case, the extractor can be matched without checking the measurements.
-          //
-          // case 2: the pattern may match some measurements of the device.
-          // in this case, we can't get all measurements efficiently here,
-          // so we just ASSUME the extractor matches and do more checks later.
-          matchedExtractors.addAll(extractorsFilteredByDevice);
+        if (deviceID instanceof StringArrayDeviceID
+            && !deviceID.getTableName().startsWith(PATH_ROOT + PATH_SEPARATOR)) {
```
luoluoyuyu, line 139: The judgment here seems to be unable to correctly determine whether to use the table model or the tree model to match

PR #13744, opened 2024-10-12
Fix some bugs of aggregation in TableModel

Wei-hao-Li:
1. count(*) process
2. intermediateTypes of count and avg
3. serialize and deserialize of aggregationMap
4. return type of aggregator


Beyyes (approved):
LGTM

PR #13809, opened 2024-10-17
Enhance Remove DataNode Test

HxpSerein:
## Description

Improve remove datanode related UT/IT for complex test scenarios in region migration tests.
- Task management and conflict checking
- Running/ReadOnly nodes can be rejoined as new nodes after being removed
- Nodes and their region statuses that failed to remove are correctly rolled back


integration-test/src/test/java/org/apache/iotdb/confignode/it/removedatanode/IoTDBRemoveDataNodeITFramework.java
```diff
@@ -88,7 +89,8 @@ public void successTest(
       final int configNodeNum,
       final int dataNodeNum,
       final int removeDataNodeNum,
-      final int dataRegionPerDataNode)
+      final int dataRegionPerDataNode,
+      final boolean needRestartRemoveDataNode)
```
OneSizeFitsQuorum, line 93: rejoinRemovedDataNode?

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/load/cache/node/DataNodeHeartbeatCache.java
```diff
@@ -38,10 +38,6 @@ public DataNodeHeartbeatCache(int dataNodeId) {
 
   @Override
   public synchronized void updateCurrentStatistics() {
-    // The Removing status can not be updated
```
OneSizeFitsQuorum, line 41: Why remove this line?Then the remove status maybe updated by Running or Readonly or Unknown?

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/load/cache/region/RegionCache.java
```diff
@@ -65,7 +65,7 @@ public synchronized void cacheHeartbeatSample(
       return;
     }
     RegionStatus lastStatus = ((RegionHeartbeatSample) getLastSample()).getStatus();
-    if (lastStatus.equals(RegionStatus.Adding) || lastStatus.equals(RegionStatus.Removing)) {
```
OneSizeFitsQuorum, line 68: same as above. This change may occur some bugs

integration-test/src/test/java/org/apache/iotdb/confignode/it/removedatanode/IoTDBRemoveDataNodeITFramework.java
HxpSerein: renamed

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/load/LoadManager.java
```diff
@@ -282,7 +282,7 @@ public void forceUpdateNodeCache(
    */
   public void removeNodeCache(int nodeId) {
     loadCache.removeNodeCache(nodeId);
-    loadCache.updateNodeStatistics();
+    loadCache.updateNodeStatistics(false);
```
OneSizeFitsQuorum, line 285: Why set here to false?

OneSizeFitsQuorum: It looks like this is also an active update, not a heartbeat update, maybe set it to true?

iotdb-protocol/thrift-datanode/src/main/thrift/datanode.thrift
```diff
@@ -804,7 +808,7 @@ service IDataNodeRPCService {
   /**
    * Config node will stop the Data node.
    */
-  common.TSStatus stopDataNode()
+  common.TSStatus stopDataNode(TStopDataNodeReq req)
```
OneSizeFitsQuorum, line 811: It appears that currently, this interface is only called when removing a datanode, and its parameter is always set to true. Can we assume this is a fixed behavior and directly proceed to delete the properties file? If, in the future, there’s a need to stop a node without deleting the properties file, we can then extend this interface. Additionally, its semantics now seem to be more than just stopping a node; perhaps it should be called stopAndClearDatanode?

OneSizeFitsQuorum (approved):
LGTM

PR #13813, opened 2024-10-17
Pipe: add seperated thread pool for phantom reference clean job

VGalaxies:
As title.


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/resource/ref/PipePhantomReferenceManager.java
```diff
@@ -55,19 +55,35 @@ protected void gcHook() {
...
+      while (count < maxCount && ((reference = REFERENCE_QUEUE.poll()) != null)) {
         finalizeResource((PipeEventPhantomReference) reference);
+        count++;
       }
     } catch (final Exception e) {
       // Nowhere to really log this.
     }
+
+    if (maxCount != 0 || getPhantomReferenceCount() != 0) {
+      LOGGER.info(
```
SteveYurongSu, line 81: Too frequent!

SteveYurongSu (approved):

PR #13858, opened 2024-10-22
Fix ignored bitmap of attribute in writeTablet

jt2594838:


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/sql/ast/InsertTablet.java
```diff
@@ -82,7 +82,9 @@ public List<Object[]> getAttributeValueList() {
       Object[] attrValues = new Object[attrColumnIndices.size()];
       for (int j = 0; j < attrColumnIndices.size(); j++) {
         final int columnIndex = attrColumnIndices.get(j);
-        attrValues[j] = ((Object[]) insertTabletStatement.getColumns()[columnIndex])[i];
+        if (!insertTabletStatement.isNull(i, j)) {
```
Caideyipi, line 85: isNull(i, columnIndex)?

jt2594838: Fixed

PR #13901, opened 2024-10-24
[region migration] Enable write retry for ratis local write

HxpSerein:
## Description

img_v3_02fv_9636f1cd-b322-4ecf-ab47-25da990c218g

PR #13913, opened 2024-10-25
[To rc/1.3.3][region migration] Enable write retry for ratis local write

HxpSerein:
cp from https://github.com/apache/iotdb/pull/13901


OneSizeFitsQuorum (approved):
LGTM

PR #13921, opened 2024-10-28
[region migration] Retry WRITE_PROCESS_REJECT status code

HxpSerein:
## Description

When the region is inactive, the status code returned by write is `WRITE_PROCESS_REJECT`. To enable retries in this situation, the mpp retry policy should also retry on this status code.

img_v3_02g0_c20fe30d-7392-4005-bc4e-3bfc4886d88g


OneSizeFitsQuorum (approved):
LGTM

PR #13936, opened 2024-10-28
[Region Migration]: Fix IoTV2 addRemotePeer stuck.

Pengzna:
as title.


Pengzna:
@OneSizeFitsQuorum PTAL~

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/pipe/consensuspipe/ConsensusPipeManager.java
```diff
@@ -60,22 +63,65 @@ public ConsensusPipeManager(PipeConsensusConfig.Pipe config, ReplicateMode repli
...
-                String.valueOf(consensusPipeName.getSenderDataNodeId()))
-            .put(
-                EXTRACTOR_CONSENSUS_RECEIVER_DATANODE_ID_KEY,
-                String.valueOf(consensusPipeName.getReceiverDataNodeId()))
-            .put(EXTRACTOR_REALTIME_MODE_KEY, replicateMode.getValue())
-            .build(),
+        consensusPipeName.toString(), params.getLeft(), params.getMiddle(), params.getRight());
+  }
+
+  public void createConsensusPipe(
```
OneSizeFitsQuorum, line 72: Do we necessarily know that an rpc sent to another node is not a coordinator? In this case, there is no need to add any parameters, and the sender and coordinator can be directly processed on the receiving side according to the inequality

OneSizeFitsQuorum: ignore me

OneSizeFitsQuorum (approved):
LGTM~
Awesome work！

PR #13951, opened 2024-10-29
Fix IoTV2 metric NPE due to concurrency

OneSizeFitsQuorum:
just addMetricSet after requestExecutor's initialization


Pengzna (approved):
LGTM！！

PR #13966, opened 2024-10-30
Load: create DB in lower-case & truncate null suffix of device ID

DanielWang2035:


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/analyze/load/LoadTsFileTableSchemaCache.java
```diff
@@ -226,7 +226,7 @@ public List<Object[]> getDeviceIdList() {
             final int realColumnIndex = fileColumn2RealColumn.getValue();
             deviceIdArray[realColumnIndex] = device.getSegments()[fileColumnIndex + 1];
           }
```
luoluoyuyu, line 228: `device.getSegments()[fileColumnIndex + 1]` will cause array out of bounds. If a table is defined with an id, when the id of a row is null, the length of `device.getSegments()` is 1 and the content is only [table], and the null value will not be saved.

luoluoyuyu (approved):

SteveYurongSu (approved):

PR #13983, opened 2024-11-04
[remove datanode] Fixed jvm heap memory in remove datanode script

HxpSerein:
## Description

When using the remove datanode script, a new Java process will be started. Using the default JVM heap memory parameter will cause startup failure. The solution is to set a fixed, smaller value for the heap memory.


iotdb-core/datanode/src/assembly/resources/sbin/remove-datanode.bat
```diff
@@ -125,7 +125,7 @@ goto :eof
 
 rem echo CLASSPATH: %CLASSPATH%
 
-"%JAVA_HOME%\bin\java" %JAVA_OPTS% %IOTDB_HEAP_OPTS% -cp %CLASSPATH% %IOTDB_JMX_OPTS% %MAIN_CLASS% %CONF_PARAMS%
+"%JAVA_HOME%\bin\java" %JAVA_OPTS% %IOTDB_HEAP_OPTS% -Xms2g -Xmx2g -cp %CLASSPATH% %IOTDB_JMX_OPTS% %MAIN_CLASS% %CONF_PARAMS%
```
OneSizeFitsQuorum, line 128: Can we set two new variables in this configuration file to indicate out-of-heap and in-heap memory? In case the 2g memory is not enough in some scenarios, users can further reduce the memory usage manually

iotdb-core/datanode/src/assembly/resources/sbin/remove-datanode.sh
```diff
@@ -73,7 +73,7 @@ launch_service()
 	iotdb_parms="$iotdb_parms -Dname=iotdb\.IoTDB"
 	iotdb_parms="$iotdb_parms -DIOTDB_LOG_DIR=${IOTDB_LOG_DIR}"
 
-	exec "$JAVA" $iotdb_parms $IOTDB_JMX_OPTS -cp "$CLASSPATH" "$class" $PARAMS
+	exec "$JAVA" $iotdb_parms $IOTDB_JMX_OPTS -Xms2g -Xmx2g -cp "$CLASSPATH" "$class" $PARAMS
```
OneSizeFitsQuorum, line 76: same

OneSizeFitsQuorum (approved):
LGTM

PR #13987, opened 2024-11-04
[remove datanode] Identify and display invalid nodes before removing

HxpSerein:
## Description

When the remove script passes multiple DataNode IDs, some of them may be invalid. Before removing, invalid IDs should be filtered out in advance and explained in the log.


OneSizeFitsQuorum (approved):
LGTM

PR #13995, opened 2024-11-05
[region migration] Fix sync lag metric during add region

liyuheng55555:
To avoid this:
img_v3_02gb_59b3708e-2e13-4e2f-87ca-f9441458b2ag

### DIFF
--- a/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/iot/IoTConsensusServerImpl.java
+++ b/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/iot/IoTConsensusServerImpl.java
@@ -844,7 +844,7 @@ public class IoTConsensusServerImpl {
   }
 
   public long getSyncLag() {
-    // make sure the getMinFlushedSyncIndex() be executed first
+    // To avoid a negative result, make sure getMinFlushedSyncIndex() be executed first
     return -getMinFlushedSyncIndex() + getSearchIndex();
   }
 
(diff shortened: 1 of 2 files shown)


iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/iot/IoTConsensusServerImpl.java
```diff
@@ -844,8 +844,7 @@ public long getSearchIndex() {
   }
 
   public long getSyncLag() {
-    long minSyncIndex = getMinSyncIndex();
-    return getSearchIndex() - minSyncIndex;
+    return getSearchIndex() - getMinFlushedSyncIndex();
```
HxpSerein, line 847: It is recommended to use a local variable as before. Because the two get operation is not an atomic operation, subtracting the two function may result in a negative number.

HxpSerein (commented):
PTAL~

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/iot/IoTConsensusServerImpl.java
liyuheng55555: - getMinFlushedSyncIndex() + getSearchIndex() looks better :P

HxpSerein (approved):
LGTM~

OneSizeFitsQuorum:
solved by https://github.com/apache/iotdb/pull/14084

PR #14007, opened 2024-11-07
[remove datanode] Enhance remove message on environment with only ConfigNode

HxpSerein:
## Description

The remove-datanode script only supports removing DataNodes in an environment where DataNode has been successfully started. If the remove script fails in an environment with only ConfigNode, a clear error message must be given.

image


OneSizeFitsQuorum (approved):
LGTM

PR #14008, opened 2024-11-07
Adjustment of setString method in IoTDDBPreparedStatement

CritasWang:
if the sql is insert and the value is not a string literal, add single quotes The table model only supports single quotes, tree model all both，but this method only judges based on single quotes
image

### DIFF
--- a/iotdb-client/jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBPreparedStatement.java
+++ b/iotdb-client/jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBPreparedStatement.java
@@ -894,9 +894,11 @@ public class IoTDBPreparedStatement extends IoTDBStatement implements PreparedSt
 
   @Override
   public void setString(int parameterIndex, String x) {
-    // if the sql is insert and the value is not a string literal, add double quotes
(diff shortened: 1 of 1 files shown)


iotdb-client/jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBPreparedStatement.java
```diff
@@ -894,9 +894,11 @@ public void setShort(int parameterIndex, short x) throws SQLException {
 
   @Override
   public void setString(int parameterIndex, String x) {
-    // if the sql is insert and the value is not a string literal, add double quotes
-    if (sql.trim().toUpperCase().startsWith("INSERT") && !x.startsWith("\"") && !x.endsWith("'")) {
-      this.parameters.put(parameterIndex, "\"" + x + "\"");
+    // if the sql is insert and the value is not a string literal, add single quotes
+    // The table model only supports single quotes, tree model all both， but this method only judges
+    // based on single quotes
```
HTHou, line 899:
```suggestion
    // if the sql is an insert statement and the value is not a string literal, add single quotes
    // The table model only supports single quotes, the tree model sql both single and double quotes. Therefore this method only judges
    // based on single quotes
```

HTHou (approved):

PR #14016, opened 2024-11-07
Fix ConfignNode LoadManager NPE when removing datanodes

OneSizeFitsQuorum:
When Datanodes are currently removed, the deletion of some maps of cn loadmanager and the callback of asynchronous rpc may have some concurrency conflicts. In this case, some robustness needs to be improved


HxpSerein (approved):
LGTM~

liyuheng55555 (approved):
Excellent use of Optional !

PR #14021, opened 2024-11-08
[Region Migration] Fix Create ConsensusPipe On Coordinator Failed

Pengzna:
Currently create consensusPipe on coordinator will failed, because the pre-check detect a consensusPipe from coordinator pointing to coordinator, as this image shows:
image
Fix this bugs caused by minor mistake...


OneSizeFitsQuorum (approved):
LGTM

PR #14047, opened 2024-11-11
Add default value in stop node scripts

shuwenwei:
## Description
Add default value in stop node scripts


OneSizeFitsQuorum (approved):
LGTM

PR #14050, opened 2024-11-11
IoTConsensusV2: Fix Create DataRegion Failed.

Pengzna:
Due to the introduction of a fix for region migration, NPE occurs when a non-region migration creates a pipe.


OneSizeFitsQuorum (approved):
LGTM

PR #14051, opened 2024-11-11
[remove datanode] Log the node information when submitting RegionMigrateProcedure

HxpSerein:
## Description

Log specific node information

### DIFF
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RemoveDataNodeHandler.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RemoveDataNodeHandler.java
@@ -151,10 +151,11 @@ public class RemoveDataNodeHandler {
               NodeType.DataNode, dataNodeId, new NodeHeartbeatSample(currentTime, nodeStatus));
 
       LOGGER.info(
-          "{}, Force update NodeCache: dataNodeId={}, nodeStatus={}",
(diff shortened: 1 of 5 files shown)


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RemoveDataNodeHandler.java
```diff
@@ -130,23 +130,40 @@ public void changeDataNodeStatus(
...
+        LOGGER.error(
+            "{}, Failed to change DataNode status, dataNodeId={}, nodeStatus={}",
+            REMOVE_DATANODE_PROCESS,
+            dataNodeId,
+            nodeStatus);
+        continue;
+      }
+
       // Force updating NodeStatus
+      long currentTime = System.nanoTime();
```
OneSizeFitsQuorum, line 147: print nanotime is useless?

HxpSerein: `updateCurrentStatistics` will choose the latest sample, so I print the time here to compare with the sample in the updateCurrentStatistics to check if it is because of the race of normal hearbeat from DataNode.

OneSizeFitsQuorum: got it

OneSizeFitsQuorum (approved):
LGTM

PR #14060, opened 2024-11-12
Modify default heap dump path of datanode

shuwenwei:
## Description
Modify default heap dump path of datanode


OneSizeFitsQuorum (approved):
LGTM

PR #14062, opened 2024-11-12
[Session]: When the thread is interrupted while retrying a connection, the loop will exit early.

YangYumings:


wangchao316 (approved):
LGTM

OneSizeFitsQuorum (approved):
LGTM

PR #14088, opened 2024-11-14
Modify default heap dump path of confignode

shuwenwei:
## Description
Modify default heap dump path of confignode


OneSizeFitsQuorum (approved):
LGTM

PR #14115, opened 2024-11-18
[remove datanode] Procedure no longer acquires node lock

Pengzna:
as title.


HxpSerein (approved):
LGTM~

wangchao316:
hi , why do no need node lock?   if have some addnodeprocedure perform now, our remove other node,   whether is effect?

OneSizeFitsQuorum:
| hi , why do no need node lock? if have some addnodeprocedure perform now, our remove other node, whether is effect?

Hi, chao, this is because we have do check before executing RemoveDataNode Procedure(you could see the code in https://github.com/apache/iotdb/blob/master/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/ProcedureManager.java#L619-L725).

Currently, only one removeDataNode task is allowed globally, so there is no need for additional concurrency control during procedure execution, otherwise it may cause some mutual blocking with pipe

OneSizeFitsQuorum (approved):
LGTM

Pengzna:
> hi , why do no need node lock? if have some addnodeprocedure perform now, our remove other node, whether is effect?

hi, sorry for replying late, I just saw this message. As @OneSizeFitsQuorum  said, the upper layer of the procedure has been guaranteed to be safe, so the lock here is unnecessary because it may block other procedures.

PR #14127, opened 2024-11-18
Make the root.__system database has a normal number of replicas (no longer forcing the replica to be 1)

liyuheng55555:


OneSizeFitsQuorum (commented):
image
modify here

OneSizeFitsQuorum (approved):
LGTM

PR #14137, opened 2024-11-19
LOGGER.error(e.getMessage()) IS REALLY A BAD IDEA

liyuheng55555:


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/consensus/statemachine/dataregion/DataRegionStateMachine.java
```diff
@@ -349,7 +349,7 @@ public DataSet read(IConsensusRequest request) {
       try {
         fragmentInstance = getFragmentInstance(request);
       } catch (IllegalArgumentException e) {
-        logger.error(e.getMessage());
+        logger.error(e.getMessage(), e);
```
OneSizeFitsQuorum, line 352: If use log.error("xxxx", e) will contains e.message, maybe we do not need to use `logger.error(e.getMessage(), e);`?

liyuheng55555: Indeed, but log4j interface always asks for a String parameter. There are some cases where I can only provide the exception stack and no additional information. In such cases, how should the string be written? Maybe just an empty string?

OneSizeFitsQuorum: Add some context for string?

OneSizeFitsQuorum (approved):
LGTM

PR #14145, opened 2024-11-19
[remove datanode] Refuse to remove when there are any other unknown or readonly DataNodes in the consensus group

HxpSerein:
## Description

1. When there are other unknown or readonly DataNodes in the consensus group that are not the removed node, for security reasons, the current remove peer operation will wait for the unknown node to come online. To optimize this situation, we can check before and refuse to remove.
2. In the future, the two-stage operations of add and remove in region migration can be provided to users as separate instructions. Therefore, for the remove operation, we can consider adding the unsafe instruction to force deletion to avoid stuck situations.

image

### DIFF
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RemoveDataNodeHandler.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RemoveDataNodeHandler.java
@@ -481,4 +481,21 @@ public class RemoveDataNodeHandler {
         .map(TRegionReplicaSet::getRegionId)
         .collect(Collectors.toList());
   }
+
+  /**
+   * Retrieves all DataNodes related to the specified DataNode.
+   *
+   * @param removedDataNode the DataNode to be removed
+   * @return a set of TDataNodeLocation representing the DataNodes associated with the specified
+   *     DataNode
+   */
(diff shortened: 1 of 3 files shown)


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/env/RemoveDataNodeHandler.java
```diff
@@ -481,4 +481,21 @@ public List<TConsensusGroupId> getMigratedDataNodeRegions(TDataNodeLocation remo
...
+   * @param removedDataNode the DataNode to be removed
+   * @return a set of TDataNodeLocation representing the DataNodes associated with the specified
+   *     DataNode
+   */
+  public Set<TDataNodeLocation> getRelatedDataNodeLocations(TDataNodeLocation removedDataNode) {
+    return configManager.getPartitionManager().getAllReplicaSets().stream()
+        .filter(
+            replicaSet ->
+                replicaSet.getDataNodeLocations().contains(removedDataNode)
+                    && replicaSet.regionId.getType() != TConsensusGroupType.ConfigRegion)
```
liyuheng55555, line 497: Unnecessary check? getAllReplicaSets() won't return config region

liyuheng55555 (changes requested):

liyuheng55555 (approved):
LGTM

OneSizeFitsQuorum (approved):
LGTM

PR #14146, opened 2024-11-20
Batch update inserted points metric for insertMultiTablets

HTHou:
## Description

As title.
image


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/DataRegion.java
jt2594838, line 1: I am quite concerned about the readability of the code. 
The length of the array and its content are unclear when someone reads top-level methods like `insertTablet`.
To figure it out, the reader must dig down via the call chain to `updateTsFileProcessorMetric`.
I think sufficient comments or an abstraction with an inner class is needed to avoid such confusion.

jt2594838 (approved):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/DataRegion.java
HTHou: Added some comments.

PR #14157, opened 2024-11-21
Refactor Table session

HTHou:
In this PR, we are introducing ITableSession and ITableSessionPool for Session API of the table model. 
The previous usages of Session with sqlDialect and IPooledSession will be removed.


JackieTien97 (approved):

iotdb-client/session/src/main/java/org/apache/iotdb/session/pool/TableSessionWrapper.java
```diff
@@ -131,67 +107,19 @@ public void executeNonQueryStatement(String sql)
...
+      if (!Objects.equals(session.getDatabase(), sessionPool.database)
+          && sessionPool.database != null) {
+        try {
+          session.executeNonQueryStatement("use " + sessionPool.database);
+        } catch (StatementExecutionException e) {
+          LOGGER.warn(
+              "Failed to change back database by executing: use {}", sessionPool.database, e);
+        }
+      }
+      sessionPool.putBack(session);
```
jt2594838, line 121: If the execution fails, is it still necessary to put the session back?

jt2594838 (approved):

iotdb-client/session/src/main/java/org/apache/iotdb/session/pool/TableSessionWrapper.java
HTHou: OK, fixed.

jt2594838 (approved):

PR #14184, opened 2024-11-22
Pipe: Fix filter issue when both tree extraction and table extraction are enabled and one extraction has non default pattern

SteveYurongSu:


luoluoyuyu (approved):
Wonderful！

PR #14195, opened 2024-11-25
[Region Migration X IoTV2]: Let other peer transfer data ONLY after snapshot completely transferred.

Pengzna:
as title


Pengzna:
expired, refer https://github.com/apache/iotdb/pull/14196

PR #14196, opened 2024-11-25
[IoTConsensusV2 X Region Migration] Start consensus pipe in proper time rather than automatically start it when create it

Pengzna:
as title

### DIFF
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/pipe/task/CreatePipeProcedureV2.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/pipe/task/CreatePipeProcedureV2.java
@@ -203,7 +203,9 @@ public class CreatePipeProcedureV2 extends AbstractOperatePipeProcedureV2 {
     }
 
     pipeRuntimeMeta = new PipeRuntimeMeta(consensusGroupIdToTaskMetaMap);
-    pipeRuntimeMeta.getStatus().set(PipeStatus.RUNNING);
+    if (!createPipeRequest.needManuallyStart) {
+      pipeRuntimeMeta.getStatus().set(PipeStatus.RUNNING);
+    }
   }
 
   @Override
--- a/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/pipe/PipeConsensusServerImpl.java
+++ b/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/pipe/PipeConsensusServerImpl.java
@@ -425,7 +425,7 @@ public class PipeConsensusServerImpl {
     try {
       // This node which acts as coordinator will transfer complete historical snapshot to new
       // target.
-      createConsensusPipeToTargetPeer(targetPeer, thisNode);
+      createConsensusPipeToTargetPeer(targetPeer, thisNode, false);
     } catch (Exception e) {
       LOGGER.warn(
           "{} cannot create consensus pipe to {}, may because target peer is unknown currently, please manually check!",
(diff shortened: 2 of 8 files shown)


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/pipe/task/CreatePipeProcedureV2.java
```diff
@@ -203,7 +203,9 @@ public void executeFromCalculateInfoForTask(ConfigNodeProcedureEnv env) {
     }
 
     pipeRuntimeMeta = new PipeRuntimeMeta(consensusGroupIdToTaskMetaMap);
-    pipeRuntimeMeta.getStatus().set(PipeStatus.RUNNING);
+    if (!createPipeRequest.needManuallyStart) {
```
SteveYurongSu, line 206: Can needManuallyStart be null?

OneSizeFitsQuorum: should check issetxxx? or you should relay on the default value for boolean

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/pipe/PipeConsensusServerImpl.java
```diff
@@ -510,6 +510,24 @@ public synchronized void dropConsensusPipeToTargetPeer(Peer targetPeer)
...
 
+  public void startOtherConsensusPipesToTargetPeer(Peer targetPeer)
+      throws ConsensusGroupModifyPeerException {
+    final List<Peer> otherPeers = peerManager.getOtherPeers(thisNode);
+    for (Peer peer : otherPeers) {
+      if (peer.equals(targetPeer)) {
+        continue;
+      }
+      try {
+        consensusPipeManager.updateConsensusPipe(
```
OneSizeFitsQuorum, line 521: If a switchover leader is generated during the addpeer phase, the pipe state of the new leader has not been started at this time, is there some corner case that needs to be considered？

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/pipe/task/CreatePipeProcedureV2.java
Pengzna: yes, we rely on the default `false` value here

SteveYurongSu: I'm afraid it's of type Boolean.. not boolean.

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/pipe/PipeConsensusServerImpl.java
Pengzna: no, only delay leader replicate without mistakes

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/pipe/task/CreatePipeProcedureV2.java
Pengzna: image
I have tested it, it's a primitive boolean type with default `false` value. And pipe it all passed, which indicates its correctness as well.

OneSizeFitsQuorum: It's boolean, so the logic here is right. @SteveYurongSu 
image

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/pipe/PipeConsensusServerImpl.java
OneSizeFitsQuorum: OK

OneSizeFitsQuorum (approved):
LGTM

SteveYurongSu (approved):

PR #14215, opened 2024-11-27
IoTConsensusV2: Fix reqBuffer reset and receive concurrent bug

Pengzna:
If some reqs find the buffer no longer contains their requestMeta after jumping out from condition.await, it may indicate that during their wait, some reqs with newer pipeTaskStartTimes or rebootTimes came in and refreshed the requestBuffer. In that cases we need to discard these requests.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/receiver/protocol/pipeconsensus/PipeConsensusReceiver.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/receiver/protocol/pipeconsensus/PipeConsensusReceiver.java
@@ -1391,11 +1391,11 @@ public class PipeConsensusReceiver {
         // Judge whether connector has rebooted or not, if the rebootTimes increases compared to
         // connectorRebootTimes, need to reset receiver because connector has been restarted.
         if (tCommitId.getDataNodeRebootTimes() > connectorRebootTimes) {
-          resetWithNewestRebootTime(tCommitId.getDataNodeRebootTimes());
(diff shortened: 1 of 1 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/receiver/protocol/pipeconsensus/PipeConsensusReceiver.java
```diff
@@ -1471,6 +1471,21 @@ private TPipeConsensusTransferResp onRequest(
                   !condition.await(
                       PIPE_CONSENSUS_RECEIVER_MAX_WAITING_TIME_IN_MS, TimeUnit.MILLISECONDS);
 
+              // If some reqs find he buffer no longer contains their requestMeta after jumping out
+              // from condition.await, it may indicate that during their wait, some reqs with newer
+              // pipeTaskStartTimes or rebootTimes came in and refreshed the requestBuffer. In that
+              // cases we need to discard this requests.
+              if (!reqExecutionOrderBuffer.contains(requestMeta)) {
```
OneSizeFitsQuorum, line 1478: Please add `condition.signalAll(); ` after `reqExecutionOrderBuffer.clear()`

Pengzna: fixed

OneSizeFitsQuorum (approved):
LGTM

PR #14217, opened 2024-11-27
Fix query failure if a table deletion involves deviceId with null & Fix status code when deleting illegal time range & support is null in deletion

jt2594838:


JackieTien97 (approved):

integration-test/src/test/java/org/apache/iotdb/relational/it/db/it/IoTDBDeletionTableIT.java
```diff
@@ -570,6 +571,52 @@ public void testMultiDevice() throws SQLException {
...
+        Statement statement = connection.createStatement()) {
+      statement.execute("use test");
+      statement.execute(
+          "create table t" + testNum + " (id1 string id, id2 string id, s1 int32 measurement)");
+      // id1 is null for this record
+      statement.execute("insert into t" + testNum + " (time, id2, s1) values (1, '1', 1)");
+      statement.execute("insert into t" + testNum + " (time, id2, s1) values (2, '', 2)");
+      statement.execute("flush");
+
+      statement.execute("delete from t" + testNum + " where id1 = NULL and time <= 1");
```
JackieTien97, line 587: should be id1 IS NULL

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/analyze/AnalyzeUtils.java
```diff
@@ -540,13 +547,14 @@ private static IDPredicate getIdPredicate(
     String rightHandValue;
     if (right instanceof StringLiteral) {
       rightHandValue = ((StringLiteral) right).getValue();
+    } else if (right instanceof NullLiteral) {
+      rightHandValue = null;
```
JackieTien97, line 551: result of XXX = NULL is NULL, so in such case, we should never delete any rows.

JackieTien97 (changes requested):

integration-test/src/test/java/org/apache/iotdb/relational/it/db/it/IoTDBDeletionTableIT.java
jt2594838: Done

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/analyze/AnalyzeUtils.java
jt2594838: Done

JackieTien97 (approved):

PR #14219, opened 2024-11-27
IoTConsensusV2: Fix syncLag error when using steam mode and 2+ replicas

Pengzna:
as title.

### DIFF
--- a/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/pipe/metric/PipeConsensusSyncLagManager.java
+++ b/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/pipe/metric/PipeConsensusSyncLagManager.java
@@ -25,6 +25,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.locks.ReentrantLock;
 
 /**
  * This class is used to aggregate the write progress of all Connectors to calculate the minimum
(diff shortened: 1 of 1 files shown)


iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/pipe/metric/PipeConsensusSyncLagManager.java
```diff
@@ -33,66 +34,50 @@
...
-      userWriteProgress = 0;
-    }
+  private long getSyncLagForSpecificConsensusPipe(ConsensusPipeConnector consensusPipeConnector) {
+    long userWriteProgress = consensusPipeConnector.getConsensusPipeCommitProgress();
+    long replicateProgress = consensusPipeConnector.getConsensusPipeReplicateProgress();
+    return Math.max(userWriteProgress - replicateProgress, 0);
   }
 
   public void addConsensusPipeConnector(ConsensusPipeConnector consensusPipeConnector) {
     consensusPipeConnectorList.add(consensusPipeConnector);
```
OneSizeFitsQuorum, line 48: lock?

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/pipe/metric/PipeConsensusSyncLagManager.java
```diff
@@ -33,66 +34,50 @@
  * <p>Note: every consensusGroup/dataRegion has and only has 1 instance of this class.
  */
 public class PipeConsensusSyncLagManager {
-  long userWriteProgress = 0;
-  long minReplicateProgress = Long.MAX_VALUE;
+  long syncLag = Long.MIN_VALUE;
+  ReentrantLock lock = new ReentrantLock();
   List<ConsensusPipeConnector> consensusPipeConnectorList = new CopyOnWriteArrayList<>();
```
OneSizeFitsQuorum, line 39: do we need copy write list anymore if we use lock?

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/pipe/metric/PipeConsensusSyncLagManager.java
Pengzna: this method is called and is only called in a locked method. so don't need lock here

Pengzna: fixed

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/pipe/metric/PipeConsensusSyncLagManager.java
Pengzna: fixed

PR #14227, opened 2024-11-27
[IoTConsensusV2 X Pipe Deletion]: Release deletion resource when delete region & Improve for release when buffer is null

Pengzna:
as title.


OneSizeFitsQuorum (approved):
LGTM

PR #14241, opened 2024-11-28
[remove datanode] Do not disable the entire region group for one removing region

HxpSerein:
## Description

With enhancements to region migration and datanode removal processes, it’s increasingly common for regions to enter the "Removing" status. However, a region in this status disables the entire consensus group design, potentially leading to a significant number of consensus groups becoming unavailable. This, in turn, can impact the overall stability of the system.


OneSizeFitsQuorum (approved):
LGTM~

This pr will improve the membership change for the three-copy consensus group, but it has not improved for the one-copy and two-copy consensus groups, but I think the latter work can be completed by @CRZbulabula

PR #14253, opened 2024-11-29
Pipe: add monitoring metrics for toTabletInsertionEvents function

XNX02:
## Description
two metrics : The size of tsfiles transferred and the number of TabletInsertionEvents returned by  toTabletInsertionEvents.


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/event/common/tsfile/parser/TabletInsertionEventIterable.java
VGalaxies, line 1: add license header plz~

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/metric/PipeTsFileToTabletMetrics.java
```diff
@@ -0,0 +1,256 @@
...
+  public void markTsFileSize(final PipeID pipeID, final long size) {
+    if (Objects.isNull(metricService)) {
+      return;
+    }
+    final Rate rate = tsFileSizeMap.get(pipeID);
+    if (rate == null) {
+      LOGGER.info("Failed to mark pipe tsfile size, PipeID({}) does not exist", pipeID);
+      return;
+    }
+    rate.mark(size);
```
Caideyipi, line 163: Is it OK to use "rate"? If we use "rate" we may concentrate on "TsFileSize / s" and "TabletCount / s", rather than "per tsFile"... Seemingly we only need to record the last value? If we want to know the average of the values, or any other staticstics, the Grafana may help us to do this.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/event/common/tsfile/parser/TabletInsertionEventIterable.java
```diff
@@ -0,0 +1,68 @@
...
+      public boolean hasNext() {
+        boolean hasNext = originalIterator.hasNext();
+        if (!hasNext && !isMarked) {
+          isMarked = true;
+          if (sourceEvent != null) {
+            PipeTsFileToTabletMetrics.getInstance()
+                .markTabletCount(
+                    PipeTsFileToTabletMetrics.PipeID.getPipeID(
+                        sourceEvent.getPipeName(), sourceEvent.getCreationTime()),
+                    count);
```
Caideyipi, line 55: Only needs to record "n"? Now it seems as if this will record 0 ~ n-1.....

SteveYurongSu:
https://github.com/apache/iotdb/pull/15202

PR #14266, opened 2024-12-02
[Vulnerability-check] Add weekly Vulnerability-check CI

133tosakarin:
dependency-check plugin


OneSizeFitsQuorum (approved):
LGTM~

If tomorrow goes well, we'll move it to Sunday at 0 a.m.

PR #14306, opened 2024-12-04
IoTV2: Improve create consensus log based on product scenario

Pengzna:
as title.

in some product cases such as TPC, we find that it is necessary to improve consensus pipe's log.


OneSizeFitsQuorum (approved):
LGTM

PR #14332, opened 2024-12-05
IoTV2: Fix addRemotePeer data inconsistency

Pengzna:
as title.

We do not need other peers to follow the progress of the coordinator. Other peers can directly extract the data file from the beginning and automatically filter out the part written by the user and transmit it.


OneSizeFitsQuorum (approved):
LGTM

PR #14333, opened 2024-12-05
[remove datanode] Fix concurrent modification when removing fails and rollback

HxpSerein:
## Description
When removing fails and rollback occurs, both ConfigNode and DataNode may modify the configuration file simultaneously.

img_v3_02h8_2f1b5b22-f434-4281-a53c-e248cb4eb65g


iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/iot/IoTConsensus.java
```diff
@@ -492,8 +493,9 @@ public void resetPeerList(ConsensusGroupId groupId, List<Peer> correctPeers)
       deleteLocalPeer(groupId);
       return;
     }
-    String previousPeerListStr = impl.getConfiguration().toString();
-    for (Peer peer : impl.getConfiguration()) {
+    ImmutableList<Peer> currentMembers = ImmutableList.copyOf(impl.getConfiguration());
```
OneSizeFitsQuorum, line 496: Will this line case concurrent bugs?

OneSizeFitsQuorum (approved):
LGTM

PR #14336, opened 2024-12-06
[remove datanode X IoTV2] Fix concurrent modification when removing fails and rollback

Pengzna:
as title.
refer https://github.com/apache/iotdb/pull/14333


OneSizeFitsQuorum (approved):
LGTM

PR #14348, opened 2024-12-06
Adjust default data region num per node from 5 to to cpu core / 2

OneSizeFitsQuorum:
kind of revert https://github.com/apache/iotdb/pull/9386 to improve blind test scene performance


HxpSerein (approved):
LGTM~

PR #14360, opened 2024-12-09
Restruct procedure waiting strategy to fix NPE bug when completed procedures were cleaned before getting result

OneSizeFitsQuorum:
Currently, there is a bug in the way we use the procedure framework to track task statuses. The current process primarily involves three steps:
* Submit the task to the procedure.
* Retain the procedure ID.
* Use the procedure ID to retrieve the status of the procedure.

The second and third steps in the above process have issues due to the asynchronous nature of procedure completion and cleanup. There are two relevant characteristics here:

* When a procedure is completed, it transitions from the procedures map to the completed map.
A background thread periodically cleans up the completed map.
* As a result, for a procedure that has already been completed, there is a chance that it will be neither in the procedures map nor in the completed map. In this case, the procedure’s status can no longer be retrieved, potentially leading to other exceptions.

The corresponding solution is to use the procedure itself directly to get its status instead of relying on the procID. HBase has adopted a similar approach (refer to [this link](https://github.com/apache/hbase/blob/master/hbase-server/src/main/java/org/apache/hadoop/hbase/master/snapshot/EnabledTableSnapshotHandler.java#L96)).

By using this approach, even if there is a CN Leader switch, threads that depend on the status of the old procedure can still complete their tasks based on the old reference.

```Java
  private boolean waitingProcedureFinished(List<Long> procedureIds, List<TSStatus> statusList) {
    boolean isSucceed = true;
    for (long procedureId : procedureIds) {
      final long startTimeForCurrentProcedure = System.currentTimeMillis();
      while (executor.isRunning()
          && !executor.isFinished(procedureId)
          && System.currentTimeMillis() - startTimeForCurrentProcedure < PROCEDURE_WAIT_TIME_OUT) {
        sleepWithoutInterrupt(PROCEDURE_WAIT_RETRY_TIMEOUT);
      }
      final Procedure<ConfigNodeProcedureEnv> finishedProcedure =
          executor.getResultOrProcedure(procedureId);
      if (!finishedProcedure.isFinished()) {
        // The procedure is still executing
        statusList.add(
            RpcUtils.getStatus(TSStatusCode.OVERLAP_WITH_EXISTING_TASK, PROCEDURE_TIMEOUT_MESSAGE));
        isSucceed = false;
        continue;
      }
      if (finishedProcedure.isSuccess()) {
        if (Objects.nonNull(finishedProcedure.getResult())) {
          statusList.add(
              RpcUtils.getStatus(
                  TSStatusCode.SUCCESS_STATUS, Arrays.toString(finishedProcedure.getResult())));
        } else {
          statusList.add(StatusUtils.OK);
        }
      } else {
        if (finishedProcedure.getException().getCause() instanceof IoTDBException) {
          final IoTDBException e = (IoTDBException) finishedProcedure.getException().getCause();
          if (e instanceof BatchProcessException) {
            statusList.add(
                RpcUtils.getStatus(
                    Arrays.stream(((BatchProcessException) e).getFailingStatus())
                        .collect(Collectors.toList())));
          } else {
            statusList.add(RpcUtils.getStatus(e.getErrorCode(), e.getMessage()));
          }
        } else {
          statusList.add(
              StatusUtils.EXECUTE_STATEMENT_ERROR.setMessage(
                  finishedProcedure.getException().getMessage()));
        }
        isSucceed = false;
      }
    }
    return isSucceed;
  }

  public Procedure<Env> getResultOrProcedure(long procId) {
    CompletedProcedureContainer retainer = completed.get(procId);
    if (retainer == null) {
      return procedures.get(procId);
    } else {
      return retainer.getProcedure();
    }
  }
```

### DIFF
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/ProcedureManager.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/ProcedureManager.java
@@ -145,9 +145,7 @@ import javax.annotation.Nonnull;
 
 import java.io.IOException;
 import java.nio.ByteBuffer;
-import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -234,7 +232,7 @@ public class ProcedureManager {
 
   public TSStatus deleteDatabases(
       final List<TDatabaseSchema> deleteSgSchemaList, final boolean isGeneratedByPipe) {
-    final List<Long> procedureIds = new ArrayList<>();
+    DeleteDatabaseProcedure procedure = null;
     final long startCheckTimeForProcedures = System.currentTimeMillis();
     for (final TDatabaseSchema databaseSchema : deleteSgSchemaList) {
       final String database = databaseSchema.getName();
@@ -248,9 +246,8 @@ public class ProcedureManager {
           hasOverlappedTask = procedureIdDuplicatePair.getRight();
 
           if (Boolean.FALSE.equals(procedureIdDuplicatePair.getRight())) {
-            final DeleteDatabaseProcedure deleteDatabaseProcedure =
-                new DeleteDatabaseProcedure(databaseSchema, isGeneratedByPipe);
-            procedureIds.add(this.executor.submitProcedure(deleteDatabaseProcedure));
+            procedure = new DeleteDatabaseProcedure(databaseSchema, isGeneratedByPipe);
+            this.executor.submitProcedure(procedure);
             break;
           }
           try {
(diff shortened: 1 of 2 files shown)


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/ProcedureManager.java
```diff
@@ -960,17 +904,15 @@ public TSStatus migrateRegion(TMigrateRegionReq migrateRegionReq) {
...
+    CreateRegionGroupsProcedure procedure =
+        new CreateRegionGroupsProcedure(consensusGroupType, createRegionGroupsPlan);
+    executor.submitProcedure(procedure);
+    TSStatus status = waitingProcedureFinished(procedure);
+    if (status.getCode() == TSStatusCode.SUCCESS_STATUS.getStatusCode()) {
+      return status;
     } else {
-      return new TSStatus(TSStatusCode.CREATE_REGION_ERROR.getStatusCode())
-          .setMessage(statusList.get(0).getMessage());
+      return new TSStatus(TSStatusCode.CREATE_TRIGGER_ERROR.getStatusCode())
```
Caideyipi, line 914: Maybe create_region_error?

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/ProcedureManager.java
```diff
@@ -1017,25 +951,15 @@ && new UpdateProcedurePlan(createTriggerProcedure).getSerializedSize() > planSiz
...
   public TSStatus dropTrigger(String triggerName, boolean isGeneratedByPipe) {
-    long procedureId =
-        executor.submitProcedure(new DropTriggerProcedure(triggerName, isGeneratedByPipe));
-    final List<TSStatus> statusList = new ArrayList<>();
-    final boolean isSucceed =
-        waitingProcedureFinished(Collections.singletonList(procedureId), statusList);
-    if (isSucceed) {
-      return RpcUtils.SUCCESS_STATUS;
-    } else {
-      return new TSStatus(TSStatusCode.DROP_TRIGGER_ERROR.getStatusCode())
```
Caideyipi, line 1028: Seemingly in functions like this, the return status for failure may change.. This may affect IT and Cli/workbench.

liyuheng55555 (changes requested):

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/ProcedureManager.java
```diff
@@ -268,33 +265,28 @@ public TSStatus deleteDatabases(
         }
       }
     }
-    final List<TSStatus> procedureStatus = new ArrayList<>();
-    final boolean isSucceed = waitingProcedureFinished(procedureIds, procedureStatus);
+    final TSStatus result = waitingProcedureFinished(procedure);
```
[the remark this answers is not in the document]
OneSizeFitsQuorum: Nice catch! Fixed

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/ProcedureManager.java
OneSizeFitsQuorum: Fixed

Caideyipi (approved):

PR #14390, opened 2024-12-12
When the config node can not find the configuration file, the set configuration command does not update other nodes' configuration

shuwenwei:
## Description
1. When the config node can not find the configuration file, the set configuration command does not update other nodes' configuration
```
IoTDB> set configuration "enable_seq_space_compaction"="true"
Msg: org.apache.iotdb.jdbc.IoTDBSQLException: 301: Unable to find the configuration file. Some modifications are made only in memory.

```
2. Ignore undefined configuration items.
```
IoTDB> set configuration "enable_seq_space_compaction1"="true" on 1
Msg: org.apache.iotdb.jdbc.IoTDBSQLException: 301: ignored config items: [enable_seq_space_compaction1] because they are immutable or undefined.

```

### DIFF
--- a/integration-test/src/test/java/org/apache/iotdb/relational/it/db/it/IoTDBSetConfigurationTableIT.java
+++ b/integration-test/src/test/java/org/apache/iotdb/relational/it/db/it/IoTDBSetConfigurationTableIT.java
@@ -56,28 +56,34 @@ public class IoTDBSetConfigurationTableIT {
 
   @Test
   public void testSetConfigurationWithUndefinedConfigKey() {
+    String expectedExceptionMsg =
+        "301: ignored config items: [a] because they are immutable or undefined.";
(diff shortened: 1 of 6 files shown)


integration-test/src/test/java/org/apache/iotdb/relational/it/db/it/IoTDBSetConfigurationTableIT.java
```diff
@@ -54,6 +54,35 @@ public static void tearDown() throws Exception {
...
+      Assert.fail(e.getMessage());
+    }
+  }
+
+  private void executeAndExpectException(Statement statement, String sql) {
+    try {
+      statement.execute(sql);
+    } catch (Exception ignored) {
+      return;
+    }
```
jt2594838, line 82: Better to check if the error message is expected.

jt2594838 (approved):

jt2594838 (approved):

PR #14399, opened 2024-12-12
IoTV2: Fix consensus pipe operation time out

Pengzna:
1. Use segmented locks distinguished by groupid for peer operations instead of large locks
2. Change the cn retry count to 10


OneSizeFitsQuorum (approved):
LGTM

PR #14406, opened 2024-12-12
Subscription: intro table model subscription session and consumer

VGalaxies:
basic support

TODO:

1. topic isolation
2. db name


iotdb-client/isession/src/main/java/org/apache/iotdb/isession/subscription/model/Subscription.java
VGalaxies, line 1: consider compatibility

SteveYurongSu (approved):

PR #14426, opened 2024-12-14
[IOTDB-6353] replace cglib to byte-buddy

wangchao316:
The cglib has stopped evolution, The cglib community suggests byte-buddy instead, so we'll do

https://github.com/cglib/cglib

IMPORTANT NOTE: cglib is unmaintained and does not work well (or possibly at all?) in newer JDKs, particularly JDK17+. If you need to support newer JDKs, we will accept well-tested well-thought-out patches... but you'll probably have better luck migrating to something like [ByteBuddy](https://bytebuddy.net/).

_*[https://github.com/raphw/byte-buddy.gi](https://github.com/raphw/byte-buddy.git*_)

### DIFF
--- a/dependencies.json
+++ b/dependencies.json
@@ -1,6 +1,6 @@
 {
     "dependencies": [
-        "cglib:cglib",
+        "net.bytebuddy:byte-buddy",
         "ch.qos.logback:logback-classic",
         "ch.qos.logback:logback-core",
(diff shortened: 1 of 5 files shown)


OneSizeFitsQuorum (approved):
LGTM

OneSizeFitsQuorum:
@wangchao316 Thanks for this patch! Learned a lot

PR #14477, opened 2024-12-18
Fix setting illegal default_storage_group_level does not report an error

jt2594838:


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/conf/IoTDBConfig.java
```diff
@@ -2478,7 +2478,19 @@ public int getDefaultStorageGroupLevel() {
...
-  void setDefaultStorageGroupLevel(int defaultStorageGroupLevel) {
+  void setDefaultStorageGroupLevel(int defaultStorageGroupLevel, boolean startUp) {
+    if (defaultStorageGroupLevel < 1) {
+      if (startUp) {
+        logger.warn(
+            "Illegal defaultStorageGroupLevel: {}, should >= 1, use default value 1",
+            defaultStorageGroupLevel);
+        defaultStorageGroupLevel = 1;
+      } else {
+        throw new IllegalArgumentException(
```
shuwenwei, line 2489: If the configuration file contains an value less than 1, it is allowed at startup, but it may cause failure during hot loading. Why not handle it the same way during hot loading as startup?

jt2594838: First, start-up failures should be avoided as much as possible because users tend not to check the logs after start-up failures and directly throw the problems to us.
However, if we simply ignore the illegal value during hot-load, the user may think the statement is successful, and the configuration has been changed, which is also unexpected.

shuwenwei (approved):

HTHou (approved):

PR #14497, opened 2024-12-19
Revert "[IOTDB-6353] replace cglib to byte-buddy (#14426)"

OneSizeFitsQuorum:
image

Performance degradation was observed


HTHou:
Should we try some newer versions before we do this revert? 
https://github.com/raphw/byte-buddy/releases

HTHou:
ok, the latest version of byte-buddy still has the problem. 
image

HTHou (approved):

PR #14527, opened 2024-12-23
[remove datanode] Remove LoadScore sort to fix RegionPriority order

HxpSerein:
## Description

After the introduction of DataNode removing, the region status is no longer bound to the Datanode, and the sorting of loadscore will cause the RegionPriority order to be wrong.


OneSizeFitsQuorum (approved):
LGTM

PR #14538, opened 2024-12-24
Compaction task can not stopped when abort compaction

shuwenwei:
## Description
Compaction task can not stopped when abort compaction

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/performer/impl/FastCompactionPerformer.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/performer/impl/FastCompactionPerformer.java
@@ -286,6 +286,19 @@ public class FastCompactionPerformer
           throw (StopReadTsFileByInterruptException) cause;
         }
         throw new IOException("[Compaction] SubCompactionTask meet errors ", e);
+      } catch (InterruptedException e) {
+        abortAllSubTasks(futures);
+        throw e;
+      }
+    }
+  }
+
+  private void abortAllSubTasks(List<Future<Void>> futures) {
+    for (Future<Void> future : futures) {
+      future.cancel(true);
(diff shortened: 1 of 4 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/performer/impl/FastCompactionPerformer.java
```diff
@@ -286,6 +286,19 @@ private void compactNonAlignedSeries(
...
+  }
+
+  private void abortAllSubTasks(List<Future<Void>> futures) {
+    for (Future<Void> future : futures) {
+      future.cancel(true);
+      try {
+        future.get();
+      } catch (Exception ignored) {
       }
     }
```
jt2594838, line 303: Maybe we can cancel them all first and then wait for them.

iotdb-core/datanode/src/test/java/org/apache/iotdb/db/storageengine/dataregion/compaction/CompactionWorkerTest.java
```diff
@@ -261,4 +266,30 @@ public void testFailedToCheckValidInInnerTask() throws InterruptedException {
...
+    AtomicBoolean isInterrupted = new AtomicBoolean(false);
+    Phaser phaser = new Phaser(2);
+    Thread t =
+        new Thread(
+            () -> {
+              compactionTaskSummary.set(
+                  new CompactionWorker.CompactionTaskFuture(new CompactionTaskSummary()));
+              phaser.arriveAndAwaitAdvance();
+              try {
+                Thread.sleep(10000);
```
jt2594838, line 283: This can be unstable. If the test is hung up for 10s or there is a clock problem in the test machine, this test is likely to fail.

jt2594838 (approved):

jt2594838 (approved):

PR #14544, opened 2024-12-25
Add database name for table model CLI

liyuheng55555:
image

### DIFF
--- a/iotdb-client/cli/src/main/java/org/apache/iotdb/cli/AbstractCli.java
+++ b/iotdb-client/cli/src/main/java/org/apache/iotdb/cli/AbstractCli.java
@@ -20,6 +20,7 @@
 package org.apache.iotdb.cli;
 
 import org.apache.iotdb.cli.utils.CliContext;
+import org.apache.iotdb.common.rpc.thrift.Model;
 import org.apache.iotdb.exception.ArgsErrorException;
 import org.apache.iotdb.jdbc.IoTDBConnection;
 import org.apache.iotdb.jdbc.IoTDBJDBCResultSet;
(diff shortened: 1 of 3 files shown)


iotdb-client/cli/src/main/java/org/apache/iotdb/cli/AbstractCli.java
```diff
@@ -881,4 +883,13 @@ static boolean processCommand(CliContext ctx, String s, IoTDBConnection connecti
     }
     return true;
   }
+
+  private static void updateUsingDatabaseIfNecessary(String database) {
+    if (!Objects.equals(usingDatabase, database)) {
+      usingDatabase = database;
+      if (sqlDialect != null && Model.TABLE.name().equals(sqlDialect.toUpperCase())) {
```
JackieTien97, line 890: what if usingDatabase is null? IOTDB_CLI_PREFIX should be IOTDB instead of "IOTDB:"

JackieTien97 (changes requested):

iotdb-client/cli/src/main/java/org/apache/iotdb/cli/AbstractCli.java
liyuheng55555: OK, I've changed this part

liyuheng55555: But still wondering in which case the "usingDatabase" will changed from not-null to null, seems we don't have "unuse database" statement for now

OneSizeFitsQuorum (approved):
LGTM

JackieTien97 (approved):

PR #14551, opened 2024-12-26
Fix that config changes rejected by DN will be written to CN's config

jt2594838:


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/ConfigManager.java
```diff
@@ -1647,39 +1648,60 @@ public TSStatus setConfiguration(TSetConfigurationReq req) {
...
+          () -> {
+            List<TSStatus> statusListOfOtherNodes = nodeManager.setConfiguration(req);
+            List<TSStatus> statusList = new ArrayList<>(statusListOfOtherNodes.size() + 1);
+            statusList.add(finalTsStatus);
+            statusList.addAll(statusListOfOtherNodes);
+            return RpcUtils.squashResponseStatusList(statusList);
+          });
+    } else {
+      // not for this node, ignore it
+      return tsStatus;
```
shuwenwei, line 1669: When the set configuration target is not this node, should it be forwarded through nodeManager?

jt2594838: fixed

shuwenwei (approved):

PR #14568, opened 2024-12-27
Add High Level Memory Threshold Metrics

SpriCoder:
Add High-Level Memory Threshold Metrics(Total, Storage Engine, Query Engine, Schema Engine, Consensus, Stream Engine, Direct Buffer)


SpriCoder:
Panel:
image

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/metrics/memory/ThresholdMemoryMetrics.java
```diff
@@ -0,0 +1,194 @@
...
+        "1");
+    metricService.remove(
+        MetricType.GAUGE,
+        Metric.THRESHOLD_MEMORY_SIZE.toString(),
+        Tag.NAME.toString(),
+        "DirectBuffer",
+        Tag.TYPE.toString(),
+        "OffHeap",
+        Tag.LEVEL.toString(),
+        "1");
```
OneSizeFitsQuorum, line 181: You could put define some field for these magic level string

OneSizeFitsQuorum, line 181: You could define some field for these magic level string，could do it in next pr

OneSizeFitsQuorum (commented):
LGTM

PR #14570, opened 2024-12-27
Detect singular frame lengths in TElasticFramedTransport

jt2594838:
If someone mistakenly send HTTP requests to the Thrift-RPC port, the error message is now more clear.


CritasWang (approved):
LGTM

PR #14594, opened 2024-12-30
Pipe: fix the problems that unable to start when cannot parse reboot times

XNX02:
## Description
Two problems remained:

1. when reboot_times.txt is deleted mistakenly, rebootTimes will be set to 0 
2. the type of rebootTimes is int while  System.currentTimeMillis() is long. It could be a overflow in 2038

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/runtime/SimpleProgressIndexAssigner.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/runtime/SimpleProgressIndexAssigner.java
@@ -20,7 +20,6 @@
 package org.apache.iotdb.db.pipe.agent.runtime;
 
 import org.apache.iotdb.commons.consensus.index.impl.SimpleProgressIndex;
-import org.apache.iotdb.commons.exception.StartupException;
 import org.apache.iotdb.commons.file.SystemFileFactory;
 import org.apache.iotdb.db.conf.IoTDBConfig;
 import org.apache.iotdb.db.conf.IoTDBDescriptor;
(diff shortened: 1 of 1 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/runtime/SimpleProgressIndexAssigner.java
```diff
@@ -86,15 +89,19 @@ private void parseRebootTimes() {
     try {
       String content = FileUtils.readFileToString(file, StandardCharsets.UTF_8);
       rebootTimes = Integer.parseInt(content);
-    } catch (IOException e) {
+    } catch (Exception e) {
       LOGGER.error("Cannot parse reboot times from file {}", file.getAbsolutePath(), e);
-      rebootTimes = 0;
+      rebootTimes = (int) (System.currentTimeMillis() / 1000);
```
SteveYurongSu, line 94: log rebootTimes if exception occurs

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/runtime/SimpleProgressIndexAssigner.java
```diff
@@ -86,15 +89,19 @@ private void parseRebootTimes() {
...
     }
   }
 
   private void recordRebootTimes() throws IOException {
     File file = SystemFileFactory.INSTANCE.getFile(PIPE_SYSTEM_DIR + REBOOT_TIMES_FILE_NAME);
-    FileUtils.writeStringToFile(file, String.valueOf(rebootTimes + 1), StandardCharsets.UTF_8);
+    try (FileOutputStream fos = new FileOutputStream(file, false);
+        FileChannel channel = fos.getChannel()) {
+      channel.write(ByteBuffer.wrap(String.valueOf(rebootTimes + 1).getBytes()));
```
Caideyipi, line 102: .getBytes(TSFileConfig.STRING_CHARSET)

SteveYurongSu (approved):

PR #14603, opened 2024-12-31
Compaction task memory is incorrect when compaction_max_aligned_series_num_in_one_batch <= 0

shuwenwei:
## Description
Compaction task memory is incorrect when compaction_max_aligned_series_num_in_one_batch <= 0

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/selector/estimator/FastCompactionInnerCompactionEstimator.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/selector/estimator/FastCompactionInnerCompactionEstimator.java
@@ -82,6 +82,9 @@ public class FastCompactionInnerCompactionEstimator extends AbstractInnerSpaceEs
   @Override
   public long roughEstimateInnerCompactionMemory(List<TsFileResource> resources)
       throws IOException {
+    if (config.getCompactionMaxAlignedSeriesNumInOneBatch() <= 0) {
+      return -1L;
+    }
     Optional<MetadataInfo> metadataInfo =
         CompactionEstimateUtils.collectMetadataInfo(
             resources,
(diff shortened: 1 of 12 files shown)


jt2594838 (approved):
Can any tests be added？

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/selector/estimator/FastCompactionInnerCompactionEstimator.java
```diff
@@ -82,6 +82,9 @@ public long calculatingDataMemoryCost(CompactionTaskInfo taskInfo) throws IOExce
   @Override
   public long roughEstimateInnerCompactionMemory(List<TsFileResource> resources)
       throws IOException {
+    if (config.getCompactionMaxAlignedSeriesNumInOneBatch() <= 0) {
+      return -1L;
+    }
```
jt2594838, line 87: If the number < 0, then compaction will be disabled? Why is so?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/selector/estimator/MetadataInfo.java
```diff
@@ -23,17 +23,18 @@
...
         IoTDBDescriptor.getInstance().getConfig().getCompactionMaxAlignedSeriesNumInOneBatch();
     compactionMaxAlignedSeriesNumInOneBatch =
-        Math.min(
-            compactionMaxAlignedSeriesNumInOneBatch <= 0
-                ? Integer.MAX_VALUE
-                : compactionMaxAlignedSeriesNumInOneBatch,
-            maxConcurrentAlignedSeriesNum);
+        compactionMaxAlignedSeriesNumInOneBatch <= 0
+            ? Integer.MAX_VALUE
+            : compactionMaxAlignedSeriesNumInOneBatch;
```
jt2594838, line 37: Maybe we can just avoid setting this config to a non-positive value.

jt2594838 (approved):

jt2594838 (approved):

PR #14613, opened 2025-01-02
IoTConsensusV2: Fix some consensus group missing due to recover failed or blocked

Pengzna:
as title.

### DIFF
--- a/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/pipe/PipeConsensus.java
+++ b/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/pipe/PipeConsensus.java
@@ -72,9 +72,12 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
 import java.util.concurrent.locks.Lock;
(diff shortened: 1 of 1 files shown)


iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/pipe/PipeConsensus.java
```diff
@@ -132,19 +135,28 @@ public synchronized void start() throws IOException {
...
 
-  private void initAndRecover() throws IOException {
+  private Future<Void> initAndRecover() throws IOException {
     if (!storageDir.exists()) {
       // init
       if (!storageDir.mkdirs()) {
         LOGGER.warn("Unable to create consensus dir at {}", storageDir);
         throw new IOException(String.format("Unable to create consensus dir at %s", storageDir));
       }
+      return null;
```
OneSizeFitsQuorum, line 159: Directly return Completed future so that you will not need to judge null?

Pengzna: good idea, fixed

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/pipe/PipeConsensus.java
```diff
@@ -132,19 +134,26 @@ public synchronized void start() throws IOException {
       throw new IOException(e);
     }
 
+    try {
+      recoverFuture.get();
+    } catch (InterruptedException | ExecutionException e) {
+      LOGGER.error("Exception while waiting for recover future completion", e);
```
OneSizeFitsQuorum, line 140: Thread.current.interupt()?

PR #14623, opened 2025-01-03
Implement distinct in AggFunction

Wei-hao-Li:
1. Support all agg-functions include UDAF.
2. This PR only contains the basic implement, the optimizers will be supported in next PR.


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/udf/access/RecordIterator.java
```diff
@@ -34,10 +34,10 @@
...
 public class RecordIterator implements Iterator<Record> {
 
-  private final List<Column> childrenColumns;
-  private final List<org.apache.tsfile.read.common.type.Type> dataTypes;
-  private final int positionCount;
-  private int currentIndex;
+  protected final List<Column> childrenColumns;
+  protected final List<org.apache.tsfile.read.common.type.Type> dataTypes;
+  protected final int positionCount;
+  protected int currentIndex;
```
JackieTien97, line 40:
```suggestion
  private final List<Column> childrenColumns;
  private final List<org.apache.tsfile.read.common.type.Type> dataTypes;
  private final int positionCount;
  private int currentIndex;
```

JackieTien97 (changes requested):

lancelly:
I noticed that In Trino there's a MarkDistinctOperator, does it have something to do this Distinct Function?
截屏2025-01-11 17 25 18

Wei-hao-Li:
> I noticed that In Trino there's a MarkDistinctOperator, does it have something to do with Distinct Function? 截屏2025-01-11 17 25 18

Yes, it is another implement of AggFunc-distinct, it will be introduced to our code in next optimizer PR (You can see design doc of this feature to see more details).

lancelly:
> > I noticed that In Trino there's a MarkDistinctOperator, does it have something to do with Distinct Function? 截屏2025-01-11 17 25 18
> 
> Yes, it is another implement of AggFunc-distinct, it will be introduced to our code in next optimizer PR (You can see design doc of this feature to see more details).

OK👍

PR #14624, opened 2025-01-03
IoTConsensusV2: Use custom commitIndex to indicate synchronization progress instead of pipe's commitIndex

Pengzna:
as title.

Now, the order of **commitIndex(or so called `replicateIndex` in code)** and **client operations** is exactly the same and will not be affected by pipe task restarts

This PR also helps to solve region migration bugs related to ProgressIndex. After this pr merged, I will try to use our custom commitIndex to calculate syncLag in region migration process instead of using progress index which may be inaccurate.


Pengzna:
TODO:
1. for coordinator: wait historical extractor complete transfer snapshot(historical data)
2. for other peers: maintain progressIndex locally for calculate syncLag(pinned ProgressIndex - replicateProgressIndex), if syncLag == 0, return true.

Pengzna:
This pr is ready for review and pass self test, PTAL @OneSizeFitsQuorum

OneSizeFitsQuorum:
We should determine the progressIndex when we decide to flush memtables

Pengzna:
@OneSizeFitsQuorum fixed, PTAL

OneSizeFitsQuorum (approved):
LGTM

PR #14675, opened 2025-01-11
fix author procedure deserialize problem.

ColinLeeo:
In this pr, fix auth plan loss when use simple consensus.
Because of the failure of authorProcedure, simple consensus will truncate some author action.


Caideyipi (approved):
LGTM

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/sync/AuthOperationProcedure.java
```diff
@@ -221,19 +227,19 @@ public void deserialize(ByteBuffer byteBuffer) {
...
     try {
-      ReadWriteIOUtils.readInt(byteBuffer);
+      int cap = ReadWriteIOUtils.readInt(byteBuffer);
+      int position = byteBuffer.position();
       this.plan = (AuthorPlan) ConfigPhysicalPlan.Factory.create(byteBuffer);
+      byteBuffer.position(cap + position);
     } catch (IOException e) {
       LOGGER.error("IO error when deserialize authplan.", e);
     }
-    if (byteBuffer.hasRemaining()) {
```
JackieTien97, line 229: why removing this judgement? if you think this bytebuffer will always has remaining, you should add a else branch and then throw exeception

JackieTien97 (changes requested):

PR #14681, opened 2025-01-13
Renaming the IT Methods of the IoTDB REST Service Table Model

CloudWise-Lukemiao:
Renaming the IT Methods of the IoTDB REST Service Table Model


integration-test/src/test/java/org/apache/iotdb/relational/it/rest/it/IoTDBRestServiceInsertValuesIT.java
```diff
@@ -51,7 +51,7 @@
 
 @RunWith(IoTDBTestRunner.class)
 @Category({LocalStandaloneIT.class, ClusterIT.class, RemoteIT.class})
```
HTHou, line 53: The IT for table should use `@Category({TableLocalStandaloneIT.class, TableClusterIT.class})`

PR #14712, opened 2025-01-16
The time range of device in resource file may be larger than the actual time range in tsfile

shuwenwei:
## Description
The time range of device in resource file may be larger than the actual time range in tsfile


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/repair/RepairDataFileScanUtil.java
```diff
@@ -283,10 +282,27 @@ private void checkNonAlignedDeviceSeries(
...
+      // actual
+      if (!deviceTimeRangeInResource.contains(actualDeviceTimeRange)) {
+        throw new CompactionStatisticsCheckFailedException(
+            device, deviceTimeRangeInResource, actualDeviceTimeRange);
+      }
+    } else {
+      if (!actualDeviceTimeRange.equals(deviceTimeRangeInResource)) {
+        throw new CompactionStatisticsCheckFailedException(
+            device, deviceTimeRangeInResource, actualDeviceTimeRange);
+      }
```
jt2594838, line 305: The two exceptions seem identical. Is the upper level or the log able to distinguish them (the missing of a device or the mismatch of time range)?

jt2594838 (approved):

PR #14737, opened 2025-01-20
Let the DataPartitionTable be automatically cleanable

CRZbulabula:
A data partition is not necessary to exist when all corresponding data are expired given the pre-configurated TTL. Hence, this PR added a thread to automatically clean these expired data partitions, making the cache size of the DataPartitionTable is always acceptable! Specifically, the main updates include:

1. In TTLManager, to retrieve the maximum TTL of a specific database, we implement getDatabaseMaxTTL().
2. In TimePartitionUtils, to locate the time partition slot of the current timestamp, we implement the getCurrentTimePartitionSlot().
3. In PartitionManager, we setup a periodically asynchronous thread, which combines the aforementioned interfaces to construct an "AutoCleanPartitionTablePlan." As a result, for each database, any data partition whose data are all expired will be deleted automatically by this independent thread.

Incidentally, the corresponding IT is available at "IoTDBPartitionTableAutoCleanTest."

### DIFF
--- a/integration-test/src/test/java/org/apache/iotdb/confignode/it/partition/IoTDBPartitionTableAutoCleanTest.java
+++ b/integration-test/src/test/java/org/apache/iotdb/confignode/it/partition/IoTDBPartitionTableAutoCleanTest.java
@@ -1,0 +1,161 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iotdb.confignode.it.partition;
+
+import org.apache.iotdb.common.rpc.thrift.TConsensusGroupId;
+import org.apache.iotdb.common.rpc.thrift.TSeriesPartitionSlot;
+import org.apache.iotdb.common.rpc.thrift.TTimePartitionSlot;
+import org.apache.iotdb.commons.client.sync.SyncConfigNodeIServiceClient;
+import org.apache.iotdb.commons.utils.TimePartitionUtils;
+import org.apache.iotdb.confignode.rpc.thrift.TDataPartitionReq;
+import org.apache.iotdb.confignode.rpc.thrift.TDataPartitionTableResp;
+import org.apache.iotdb.it.env.EnvFactory;
+import org.apache.iotdb.it.framework.IoTDBTestRunner;
+import org.apache.iotdb.itbase.category.ClusterIT;
+import org.apache.iotdb.rpc.TSStatusCode;
+
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+
+import java.sql.Connection;
+import java.sql.Statement;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+import java.util.concurrent.TimeUnit;
+
+@RunWith(IoTDBTestRunner.class)
+@Category({ClusterIT.class})
+public class IoTDBPartitionTableAutoCleanTest {
+
+  private static final int TEST_REPLICATION_FACTOR = 1;
+  private static final long TEST_TIME_PARTITION_INTERVAL = 604800000;
+  private static final long TEST_TTL_CHECK_INTERVAL = 5_000;
+  //  private static final int TEST_SERIES_SLOT_NUM = 1000;
+  //  private static final String TEST_SERIES_EXECUTOR_CLASS =
+  // "org.apache.iotdb.commons.partition.executor.hash.BKDRHashExecutor";
+
+  //  private static final SeriesPartitionExecutor TEST_EXECUTOR =
+  //    SeriesPartitionExecutor.getSeriesPartitionExecutor(TEST_SERIES_EXECUTOR_CLASS,
+  // TEST_SERIES_SLOT_NUM);
+  private static final TTimePartitionSlot TEST_CURRENT_TIME_SLOT =
+      TimePartitionUtils.getCurrentTimePartitionSlot();
+  private static final long TEST_TTL = 7 * TEST_TIME_PARTITION_INTERVAL;
+
+  @Before
+  public void setUp() throws Exception {
+    EnvFactory.getEnv()
+        .getConfig()
+        .getCommonConfig()
+        //      .setSeriesSlotNum(TEST_SERIES_SLOT_NUM)
+        //      .setSeriesPartitionExecutorClass(TEST_SERIES_EXECUTOR_CLASS)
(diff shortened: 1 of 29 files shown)


liyuheng55555 (changes requested):
Haven't finished reading all the code yet, I have some thoughts for discussing: 

If the default execution interval is every 2 hours, then maybe adding a periodically executed Procedure is more lightweight compared to adding a new thread pool?

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/concurrent/ThreadName.java
```diff
@@ -90,6 +90,8 @@ public enum ThreadName {
   CONFIG_NODE_LOAD_PUBLISHER("Cluster-LoadStatistics-Publisher"),
   // -------------------------- ConfigNode-RegionManagement --------------------------
   CONFIG_NODE_REGION_MAINTAINER("IoTDB-Region-Maintainer"),
+  // -------------------------- ConfigNode-PartitionManagement --------------------------
+  CONFIG_NODE_PARTITION_CLEANER("IoTDB-Partition-Cleaner"),
```
OneSizeFitsQuorum, line 94: remove this thread?

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/partition/PartitionManager.java
```diff
@@ -1448,6 +1464,54 @@ public void stopRegionCleaner() {
...
+   * The ConfigNode-leader will periodically invoke this interface to automatically clean expired
+   * partition table.
+   */
+  public void autoCleanDataPartitionTable() {
+    List<String> databases = getClusterSchemaManager().getDatabaseNames(null);
+    Map<String, Long> databaseTTLMap = getClusterSchemaManager().getTTLInfoForUpgrading();
+    for (String database : databases) {
+      long subTreeMaxTTL = getTTLManager().getDatabaseMaxTTL(database);
+      databaseTTLMap.put(
+          database, Math.max(subTreeMaxTTL, databaseTTLMap.getOrDefault(database, -1L)));
```
OneSizeFitsQuorum, line 1477: add some judgement like "isDatabaseExisted(database) && 0 < ttl && ttl < Long.MAX_VALUE" here to remove overhead? 
BTW, If all the databases don't have ttl, we can logically just do this check and find that none of them need to be cleaned up, so there's no need to do a consensus write

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/partition/PartitionManager.java
```diff
@@ -136,21 +142,31 @@ public class PartitionManager {
...
+  private final ScheduledExecutorService partitionCleaner;
+  private Future<?> currentPartitionCleanerFuture;
+
   public PartitionManager(IManager configManager, PartitionInfo partitionInfo) {
     this.configManager = configManager;
     this.partitionInfo = partitionInfo;
     this.regionMaintainer =
         IoTDBThreadPoolFactory.newSingleThreadScheduledExecutor(
             ThreadName.CONFIG_NODE_REGION_MAINTAINER.getName());
+    this.partitionCleaner =
```
OneSizeFitsQuorum, line 167: maybe try to reuse procedure periodic tasks

integration-test/src/test/java/org/apache/iotdb/confignode/it/partition/IoTDBPartitionTableAutoCleanTest.java
```diff
@@ -0,0 +1,161 @@
...
+  private static final TTimePartitionSlot TEST_CURRENT_TIME_SLOT =
+      TimePartitionUtils.getCurrentTimePartitionSlot();
+  private static final long TEST_TTL = 7 * TEST_TIME_PARTITION_INTERVAL;
+
+  @Before
+  public void setUp() throws Exception {
+    EnvFactory.getEnv()
+        .getConfig()
+        .getCommonConfig()
+        //      .setSeriesSlotNum(TEST_SERIES_SLOT_NUM)
```
OneSizeFitsQuorum, line 71: please do not comment code

integration-test/src/test/java/org/apache/iotdb/confignode/it/partition/IoTDBPartitionTableAutoCleanTest.java
```diff
@@ -0,0 +1,161 @@
...
+
+@RunWith(IoTDBTestRunner.class)
+@Category({ClusterIT.class})
+public class IoTDBPartitionTableAutoCleanTest {
+
+  private static final int TEST_REPLICATION_FACTOR = 1;
+  private static final long TEST_TIME_PARTITION_INTERVAL = 604800000;
+  private static final long TEST_TTL_CHECK_INTERVAL = 5_000;
+  //  private static final int TEST_SERIES_SLOT_NUM = 1000;
+  //  private static final String TEST_SERIES_EXECUTOR_CLASS =
```
OneSizeFitsQuorum, line 56: please do not comment code

integration-test/src/test/java/org/apache/iotdb/confignode/it/partition/IoTDBPartitionTableAutoCleanTest.java
```diff
@@ -0,0 +1,161 @@
...
+            }
+            if (!partitionTableAutoCleaned) {
+              break;
+            }
+          }
+        }
+        if (partitionTableAutoCleaned) {
+          return;
+        }
+        TimeUnit.SECONDS.sleep(1);
```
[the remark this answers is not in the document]
CRZbulabula: An awesome suggestion! The corresponding test codes are simplified significantly!

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/partition/PartitionManager.java
CRZbulabula: Thanks for pinpointing this logic enhancement. The judgement is available at PartitionTableAutoCleaner.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/partition/PartitionManager.java
CRZbulabula: Sure. I now employ the PartitionTableAutoCleaner rather than creating an extra thread pool.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/concurrent/ThreadName.java
CRZbulabula: Removed

liyuheng55555 (commented):
Have you considered concurrency issues, such as setTTL and Cleaner running simultaneously? 

(I think one solution could be to clean up partitions only after they’ve exceeded the TTL by one hour, but you might have solved this in another way ：)

CRZbulabula:
> Have you considered concurrency issues, such as setTTL and Cleaner running simultaneously?
> 
> (I think one solution could be to clean up partitions only after they’ve exceeded the TTL by one hour, but you might have solved this in another way ：)

Thanks for raising this critical issue. To address your concern, please refer to the `autoCleanPartitionTable` in the `SeriesPartitionTable`, as outlined below:

```
public void autoCleanPartitionTable(long TTL, TTimePartitionSlot currentTimeSlot) {
    seriesPartitionMap
        .entrySet()
        .removeIf(entry -> entry.getKey().getStartTime() + TTL < currentTimeSlot.getStartTime());
  }
```

Here, the removing condition is `<` rather than `<=`. As a result, a data partition is removed from cache only when it exceeded the TTL by a whole time partitioning interval. Therefore, the PartitionTable always contains a row of "empty" data partitions. I believe this approach can avoid this concurrency problem.

liyuheng55555 (approved):
Good work! Thank you for addressing my feedback patiently!

OneSizeFitsQuorum (approved):
LGTM

PR #14739, opened 2025-01-21
[Table Model Subquery] Support Exists subquery

lancelly:
This PR introduces both correlated and uncorrelated Exists subquery into IoTDB TableModel.
Correlated subquery has such limitations:
1. limit clause with row count greater than 1 is not supported in correlated subquery for now.
2. TopK clause is not supported in correlated subquery for now.
3. Non-Equality comparison is not supported in correlated subquery for now.(Error msg for this is kind of confusing since it's related to join: `For now, FullOuterJoin and LeftJoin only support EquiJoinClauses` )


Beyyes:
image
Fix these TODOs.

Beyyes (approved):

PR #14750, opened 2025-01-21
Fix wrong deviceId after PartialPath.concatAsMeasurementPath

jt2594838:
The deviceId is re-generated after PartialPath.concatAsMeasurementPath, which may be incorrect because it is recalculated by the default rule.

The result is that the wrong deviceId cannot be matched by existing mod entries, causing data to re-appear after compaction.

To fix it, the result of PartialPath.concatAsMeasurementPath will inherit the deviceId from the parameter.


jt2594838 (commented):
Check

shuwenwei (approved):

PR #14766, opened 2025-01-23
Add some tsfile-related tools

jt2594838:
1. TsFileRewriteOverPrecisedI64Scan
Rewrite Int64 chunks that can be converted to Int32 losslessly.

2. TsFileRewriteSmallRangeI64Scan
Rewrite Int64 chunks whose ranges are no larger than Int32.MAX_VALUE by replacing the value with the difference with the first value in the chunk.

3. TsFileStatisticScan
It prints:
a. the number of series of each data type;
b. the number of Int64 chunks that are over-precised (can be converted to Int32 losslessly), just-precised (cannot be converted to Int32 losslessly), small range (range not larger than Int32.MAX_VALUE), or large range (range larger than Int32.MAX_VALUE);
c. the number of points of each data type;
d. the total chunk size of each data type;
e. average distinct binary value num of STRING and TEXT chunks;

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/tools/utils/TsFileRewriteOverPrecisedI64Scan.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/tools/utils/TsFileRewriteOverPrecisedI64Scan.java
@@ -26,8 +26,8 @@ import org.apache.tsfile.file.metadata.statistics.Statistics;
 import org.apache.tsfile.read.common.BatchData;
 import org.apache.tsfile.read.common.Chunk;
 import org.apache.tsfile.read.reader.IPageReader;
-import org.apache.tsfile.read.reader.chunk.AlignedChunkReader;
 import org.apache.tsfile.read.reader.chunk.ChunkReader;
+import org.apache.tsfile.read.reader.chunk.TableChunkReader;
 import org.apache.tsfile.utils.Pair;
 import org.apache.tsfile.utils.TsPrimitiveType;
 import org.apache.tsfile.write.chunk.ChunkWriterImpl;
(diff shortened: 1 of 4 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/tools/utils/TsFileRewriteOverPrecisedI64Scan.java
```diff
@@ -0,0 +1,217 @@
...
+    boolean shouldScan = super.onFileOpen(file);
+    if (shouldScan) {
+      writer = new TsFileIOWriter(target);
+    }
+    return shouldScan;
+  }
+
+  @Override
+  protected void onFileEnd() throws IOException {
+    writer.endFile();
```
shuwenwei, line 78: Does the last chunk group call writer.endChunkGroup()?

jt2594838: Fixed

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/tools/utils/TsFileRewriteOverPrecisedI64Scan.java
```diff
@@ -0,0 +1,218 @@
...
+      }
+      chunkWriter.sealCurrentPage();
+    }
+    chunkWriter.writeToFileWriter(writer);
+    return true;
+  }
+
+  private boolean rewriteInt64ChunkAligned(Chunk chunk) throws IOException {
+    AlignedChunkReader chunkReader =
+        new AlignedChunkReader(currTimeChunk, Collections.singletonList(chunk));
```
shuwenwei, line 168: There are only one column. If the value of this column is null, one row may be skipped.

jt2594838: This is to find how much space can be saved by rewriting Int64 to Int32. If nulls are removed, the comparison will be not that precise.

shuwenwei: Suppose there are (time, s1, s2, s3,) and we want to rewrite chunks of s1. At this time, a TimeValuePair is (1, null, 1, 1). When only (time, s1) is provided to AlignedChunkReader, this row will not be returned because all value is null, and the new chunk written by ValueChunkWriter may not be able to align the value with the original time chunk.

shuwenwei: To retain these null values, TableChunkReader should be used here

jt2594838: Fixed

shuwenwei (approved):

PR #14779, opened 2025-01-26
[RatisConsensus] prevent misuse of addRemotePeer

SzyWilliam:
We should forbid adding new peer having same id/address with existing peers.


OneSizeFitsQuorum (approved):
LGTM

PR #14781, opened 2025-01-26
fix_docker,iotdb-system.properties contains all configuration,stop-standalone.sh will be called anyway

xiaoyekanren:
fix:
1. iotdb-system.properties is no full configuration in iotdb-system.properties. 
2. Not sure stop-standalone.sh will be called anyway.


CritasWang (approved):
LGTM

PR #14783, opened 2025-01-27
[WIP] RTO/RPO: retry framework

SzyWilliam:
WIP


iotdb-client/session/src/main/java/org/apache/iotdb/session/util/retry/RetryDecision.java
```diff
@@ -0,0 +1,9 @@
+package org.apache.iotdb.session.util.retry;
```
wangchao316, line 1: new class need add apache license header.

OneSizeFitsQuorum:
duplicated

PR #14787, opened 2025-02-04
fix auth parser error with grant all on root.** from non-root user.

ColinLeeo:
## Description
Fix bug "Not support Relation statement in tree sql_dialect " when a non-root user grant all on root.** to user other.


Caideyipi (approved):
LGTM

PR #14809, opened 2025-02-10
Support 'set sql_dialect' in Tree and Table Model

Wei-hao-Li:
This PR also supported 'show current_sql_dialect' and 'show current_user' in Tree Model.

image

### DIFF
--- a/iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/RpcUtils.java
+++ b/iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/RpcUtils.java
@@ -375,10 +375,17 @@ public class RpcUtils {
   }
 
   public static boolean isSetSqlDialect(String sql) {
-    sql = sql.toLowerCase();
-    return sql.length() > 15
-        && "set ".equals(sql.substring(0, 4))
-        && sql.substring(4).trim().startsWith("sql_dialect");
+    // check if startWith 'set '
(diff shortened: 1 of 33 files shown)


JackieTien97 (changes requested):
Support show current_sql_dialect for tree model, and add it

JackieTien97 (changes requested):
You should also think about JDBC set sql_dialect

iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/RpcUtils.java
```diff
@@ -375,7 +375,10 @@ public static boolean isUseDatabase(String sql) {
   }
 
   public static boolean isSetSqlDialect(String sql) {
-    return sql.length() > 15 && "set sql_dialect".equalsIgnoreCase(sql.substring(0, 15));
+    sql = sql.toLowerCase();
+    return sql.length() > 15
+        && "set ".equals(sql.substring(0, 4))
+        && sql.substring(4).trim().startsWith("sql_dialect");
```
JackieTien97, line 381:
```suggestion
    return sql.length() > 15
        && "set ".equalsIgnoreCase(sql.substring(0, 4))
        && sql.substring(4).trim().toLowerCase().startsWith("sql_dialect");
```

JackieTien97 (approved):

PR #14812, opened 2025-02-11
Replace ExportTsFile By SubscriptionTsFile

2b3c511:
Implementation of Replacing Export_tsfile Tool with Subscript_tsfile


CritasWang (changes requested):

CritasWang (approved):

iotdb-client/cli/src/main/java/org/apache/iotdb/tool/tsfile/subscription/SubscriptionTableTsFile.java
```diff
@@ -0,0 +1,173 @@
...
+                  .consumerId(Constants.CONSUMER_NAME_PREFIX + i)
+                  .consumerGroupId(groupId)
+                  .autoCommit(Constants.AUTO_COMMIT)
+                  .autoCommitIntervalMs(Constants.AUTO_COMMIT_INTERVAL)
+                  .fileSaveDir(commonParam.getTargetDir())
+                  .buildTablePullConsumer());
+    }
+    for (ISubscriptionTablePullConsumer consumer : commonParam.getPullTableConsumers()) {
+      consumer.open();
+    }
```
luoluoyuyu, line 112: I think there are some details that need to be dealt with here, such as whether the failure of a consumer.open will affect all consumers executing open, or the operation of the entire program.

PR #14814, opened 2025-02-11
【remember to add Hu and Peng to co-arthor and delete this bracket before merge】IoTConsensus and IoTConsensusV2 no longer stores the peer list locally on the DataNode

liyuheng55555:
The new `resetPeerList` interface is super powerful. 

Now, IoTConsensus member management can be fully handled by ConfigNode, and DataNode can simply obtain the peer list from ConfigNode during startup (through `resetPeerList` interface).

### DIFF
--- a/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/iot/IoTConsensus.java
+++ b/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/iot/IoTConsensus.java
@@ -497,7 +497,7 @@ public class IoTConsensus implements IConsensus {
     Peer localPeer = new Peer(groupId, thisNodeId, thisNode);
     if (!correctPeers.contains(localPeer)) {
       logger.info(
-          "[RESET PEER LIST] Local peer is not in the correct configuration, delete local peer {}",
+          "[RESET PEER LIST] {} Local peer is not in the correct configuration, delete it.",
           groupId);
       deleteLocalPeer(groupId);
       return;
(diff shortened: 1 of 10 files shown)


iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/iot/IoTConsensus.java
```diff
@@ -174,7 +175,7 @@ private void initAndRecover() throws IOException {
               new IoTConsensusServerImpl(
                   path.toString(),
                   new Peer(consensusGroupId, thisNodeId, thisNode),
-                  new ArrayList<>(),
+                  new TreeSet<>(),
```
Pengzna, line 178: why uses treeset?

Pengzna (commented):
LGTM

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/iot/IoTConsensus.java
Pengzna, line 178: why use treeset?

liyuheng55555: Previously, duplicates were removed during configuration persistence. Now that the persistence code is gone, a set is used to keep this logic. 

TreeSet is used to return a List<Peer> with a consistent order (for an interface), which I believe is better than the unordered HashSet.

iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/iot/logdispatcher/LogDispatcher.java
```diff
@@ -98,7 +98,13 @@ private void initLogSyncThreadPool() {
 
   public synchronized void start() {
     if (!threads.isEmpty()) {
-      threads.forEach(executorService::submit);
+      threads.stream()
```
OneSizeFitsQuorum, line 101: try to do not start thread before start was called

liyuheng55555: Added an parameter to control whether whether sync log channel start or "just build without start". 

In this way, IoTV1 and IoTV2 are quite similar :)

OneSizeFitsQuorum (approved):
LGTM~
Really Awesome wrok！

PR #14834, opened 2025-02-13
Fix NPE of RelationalDeleteDataNode merge method

liyuheng55555:
Database name might be null, and this kind of delete operation should be allowed to merge with others.


jt2594838 (approved):

OneSizeFitsQuorum (approved):
LGTM

PR #14840, opened 2025-02-13
[IoTConsensusV2] Fix NPE when transfer tsfile mods

Pengzna:
After discussion, we decide to maintain both mods transfer and DAL transfer for DELETION replication.

This pr enhances the support for mods transfer.

As this image shows, when transfer mods may encounter npe.
image

This is caused by 2 bugs:
1. sync a file after close its handler(we should sync before close)
2. `handleTsFileSealWithMods` doesn't support mods transfer well, need some enhancement.


OneSizeFitsQuorum (approved):
LGTM

PR #14853, opened 2025-02-14
Delete remove-confignode.sh and remove-datanode.sh

HxpSerein:
## Description

For the sake of cluster security, delete the remove-confignode/datanode script after adding the remove SQL.


OneSizeFitsQuorum (approved):
LGTM

PR #14856, opened 2025-02-16
Subscription: fix consumer infinite pulling event & fully managed tsfile parsing process &  increase the reference count for subscribed parsed raw tablet event

VGalaxies:
As title.


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/connection/PipeEventCollector.java
```diff
@@ -127,6 +131,11 @@ private void parseAndCollectEvent(final PipeTsFileInsertionEvent sourceEvent) th
       return;
     }
 
+    if (skipParseTsFile) {
+      collectEvent(sourceEvent);
+      return;
+    }
```
VGalaxies, line 137: move up

SteveYurongSu (approved):

PR #14915, opened 2025-02-20
[RTO/RPO] Make ConfigNode aware of DataNode topology

SzyWilliam:
This patch enables DataNode to report connectivity with other datanodes.


OneSizeFitsQuorum:
https://github.com/apache/iotdb/pull/15014

PR #14921, opened 2025-02-21
[to dev/1.3] Initialize WALNode when DataRegion is created for iotv1 and iotv2

shuwenwei:
## Description
Initialize WALNode when DataRegion is created for iotv1 and iotv2


OneSizeFitsQuorum (approved):
LGTM

PR #14922, opened 2025-02-21
Pipe: Implement sink.data-distribution-strategy feature where a pipe can send data to all sink nodes

XNX02:


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/pipe/connector/protocol/IoTDBConfigRegionConnector.java
```diff
@@ -144,45 +146,48 @@ private void doTransferWrapper(
...
+        shouldSendToAllClients
+            ? clientManager.getAllClients()
+            : Collections.singletonList(clientManager.getClient());
+    for (final Pair<IoTDBSyncClient, Boolean> clientAndStatus : clientsAndStatuses) {
+      final TPipeTransferResp resp;
+      try {
+        final TPipeTransferReq req =
+            compressIfNeeded(
+                PipeTransferConfigPlanReq.toTPipeTransferReq(
+                    pipeConfigRegionWritePlanEvent.getConfigPhysicalPlan()));
```
luoluoyuyu, line 159: If the byte[] here can also be reused, the sending performance can be greatly improved. However, it should be noted that the internal ByteBuffer needs to copy the internal array.

XNX02: Could you clarify what you meant by 'internal ByteBuffer needs to copy the internal array'? It sounds that simply place the req outside the loop is not enough in this situation.

luoluoyuyu: image
Because ByteBuffer has capacity, limit, position and mark to record the corresponding progress. But I looked at the code and found that a new ByteBuffer will be created when writing. You can ignore my words.

XNX02: Got it

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/connector/protocol/IoTDBConnector.java
```diff
@@ -331,6 +336,22 @@ public void validate(final PipeParameterValidator validator) throws Exception {
...
+    final String dataDistributionStrategy =
+        parameters
+            .getStringOrDefault(
+                Arrays.asList(
+                    CONNECTOR_DATA_DISTRIBUTION_STRATEGY, SINK_DATA_DISTRIBUTION_STRATEGY),
+                CONNECTOR_DATA_DISTRIBUTION_STRATEGY_DEFAULT_VALUE)
+            .trim()
+            .toLowerCase();
+    validator.validate(
+        arg -> arg.equals("any") || arg.equals("all"),
```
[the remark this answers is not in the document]
XNX02: Makes sense. Done.

PR #14933, opened 2025-02-24
Delete remove node logic in ServerCommandLine

HxpSerein:
## Description

For the sake of cluster security, delete the remove-confignode/datanode logic in ServerCommandLine after adding the remove SQL.

### DIFF
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/service/ConfigNode.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/service/ConfigNode.java
@@ -30,7 +30,6 @@ import org.apache.iotdb.commons.concurrent.ThreadPoolMetrics;
 import org.apache.iotdb.commons.conf.CommonConfig;
 import org.apache.iotdb.commons.conf.CommonDescriptor;
 import org.apache.iotdb.commons.conf.IoTDBConstant;
-import org.apache.iotdb.commons.exception.BadNodeUrlException;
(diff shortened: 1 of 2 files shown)


OneSizeFitsQuorum (approved):
LGTM

liyuheng55555 (changes requested):
When updating, users are likely not to delete the remove-node.sh file. After the update, they will find that the script no longer works. 

I’m wondering if it would be better to temporarily keep some prompts for the -r parameter to guide users to use the CLI for node removal ?

liyuheng55555:
> When updating, users are likely not to delete the remove-node.sh file. After the update, they will find that the script no longer works.
> 
> I’m wondering if it would be better to temporarily keep some prompts for the -r parameter to guide users to use the CLI for node removal ?

By the way, -s option seems uncessary too. We can keep it for now, and print some suggesstion, and remove it in next version.

OneSizeFitsQuorum:
| I’m wondering if it would be better to temporarily keep some prompts for the -r parameter to guide users to use the CLI for node removal ?

sounds better

wangchao316:
we need this script for remove node.  I suggest script and sql support together.

OneSizeFitsQuorum:
| we need this script for remove node. I suggest script and sql support together.

The script does not implement authentication and is recommended to be removed for security reasons. We can discuss it on the mailing list?

OneSizeFitsQuorum (approved):
LGTM

PR #14938, opened 2025-02-24
IoTConsensusV2: persistent isGeneratedByPipeConsensus

Pengzna:
Let us consider one case:

- There is node {3, 4, 5}, 3 is the old leader of region1, 4 is the new leader of region1.

- Focus on node 4, let's say, on which tsfileA is replicated from node3, tsfileB is a local generated SSTable. And tsfileB has already been replicated to node 3 successfully.

- when node 4 **restarts**, consensus pipe will try to flush and extract tsfiles.

- For some cases, if tsfileA and tsfileB are **compacted** to a new tsfileC, then the tsfileC's progressIndex will consists of both node 3(old leader/tsfileA) and node 4(new leader/tsfileB) 's mark.

- Since tsfileC's progressIndex contains local(node 4) mark, consensus pipe will transfer it to follower to prevent data loss, which is **totally unnecessary** because tsfileC's data is **existed** in node 3(tsfileA and replicated tsfileB)

So we need to persistent `isGeneratedByPipeConsensus` to prevent such case. After this pr, high level tsfiles extracted by consensus pipe when dn restarts will be filtered by `isGeneratedByPipeConsensus` and not be transferred to followers   anymore.

Latest update:

In image below, the TsFileC will be unnecessarily transferred.

image


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/analyze/load/LoadTsFileAnalyzer.java
```diff
@@ -458,6 +458,10 @@ private TsFileResource constructTsFileResource(
       tsFileResource.updatePlanIndexes(reader.getMaxPlanIndex());
     } else {
       tsFileResource.deserialize();
+      // Clean up tsfileResource's isGeneratedByPipe mark to prevent deserializing the wrong mark.
+      // If this tsfile is loaded by a pipe receiver, the correct mark will be added in
+      // `listenToTsFile`
+      tsFileResource.setGeneratedByPipe(false);
```
DanielWang2035, line 464: tsFileResource.setGeneratedByPipe(isGeneratedByPipe)?

Pengzna: great point, fixed

DanielWang2035 (approved):
LGTM

OneSizeFitsQuorum (approved):
LGTM

PR #14963, opened 2025-02-26
[To dev/1.3] Adjust default data region num per node from 5 to to cpu core / 2 & IoTConsensusV2: Flush old leader when leader transfer & Delay the execution of invalidateSchemaCache after leader change 

Pengzna:
as title.


OneSizeFitsQuorum (approved):
LGTM

PR #14965, opened 2025-02-26
[IoTConsensus] More accurate statistics on IoTConsensus memory management

HxpSerein:
## Description

The actual memory occupied by IoTConsensus is 2-3 times larger than the statistical memory of memory management, so more detailed statistics of consensus memory are needed.

As shown in the figure, the statistical result of using `getMemorySize` is about twice the size of `buffer`, which makes the result more accurate.

image


OneSizeFitsQuorum (approved):
LGTM

PR #15003, opened 2025-03-03
Load: Support loading table Tsfile by load-tsfile.sh

MiniSho:
## Description
add new configuration `-db` or `--database` to specify the database.

### DIFF
--- a/iotdb-client/cli/src/main/java/org/apache/iotdb/tool/tsfile/ImportTsFileLocally.java
+++ b/iotdb-client/cli/src/main/java/org/apache/iotdb/tool/tsfile/ImportTsFileLocally.java
@@ -22,6 +22,12 @@ package org.apache.iotdb.tool.tsfile;
 import org.apache.iotdb.cli.utils.IoTPrinter;
 import org.apache.iotdb.session.pool.SessionPool;
 
+import java.util.ArrayList;
+import java.util.List;
+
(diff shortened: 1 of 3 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/receiver/protocol/thrift/IoTDBDataNodeReceiver.java
```diff
@@ -545,8 +545,8 @@ protected String getSenderPort() {
   protected TSStatus loadFileV1(final PipeTransferFileSealReqV1 req, final String fileAbsolutePath)
```
Caideyipi, line 545: Seems that "importTsFileRemotely" will also use loadFileV2?

iotdb-client/cli/src/main/java/org/apache/iotdb/tool/tsfile/ImportTsFileLocally.java
```diff
@@ -34,11 +40,9 @@ public void loadTsFile() {
     String filePath;
     try {
       while ((filePath = ImportTsFileScanTool.pollFromQueue()) != null) {
-        final String sql =
-            "load '" + filePath + "' onSuccess=none " + (verify ? "" : "verify=false");
+        String sql = buildLoadSql(filePath);
         try {
           sessionPool.executeNonQueryStatement(sql);
```
Caideyipi, line 45: Is the session sql dialect changed?

PR #15017, opened 2025-03-05
Reject insert without field column

HTHou:
## Description

As title.


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/write/RelationalInsertRowNode.java
jt2594838, line 1: Also modify RelationalInsertTabletNode and add a test where all columns in a Tablet fail.

jt2594838 (approved):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/write/RelationalInsertRowNode.java
HTHou: Fixed.

PR #15023, opened 2025-03-06
Fix frozen IoTConsensus sync lag

jt2594838:
If the IoTConsensus dispatcher is about to wait right after the last WAL flush, it cannot be woken up as long as there is no further insertion.
The result is that the dispatcher cannot sense the WALs of the last WAL flush, and the sync lag is frozen.

To avoid this, the dispatcher should double-check the search index before waiting.


OneSizeFitsQuorum (approved):
LGTM

PR #15030, opened 2025-03-06
Fix some issue in auth model.

ColinLeeo:
In this pr:
1. remove MAINTAIN in relational model.
2. FIx the issue where special characters could not be used when creating users or roles .
3. Fix alter user statement fail.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/sql/parser/AstBuilder.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/sql/parser/AstBuilder.java
@@ -1443,10 +1443,32 @@ public class AstBuilder extends RelationalSqlBaseVisitor<Node> {
     return text;
   }
 
+  private String stripDoubleQuotes(String text) {
+    if (text != null && text.length() >= 2 && text.startsWith("\"") && text.endsWith("\"")) {
+      return text.substring(1, text.length() - 1).replace("''", "'");
+    }
+    return text;
+  }
+
+  private String stripBackQuotes(String quoted) {
+    return quoted.replaceAll("^`|`$", "");
+  }
+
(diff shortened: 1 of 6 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/sql/parser/AstBuilder.java
```diff
@@ -1443,81 +1443,102 @@ private String stripQuotes(String text) {
...
   }
 
+  private String stripDoubleQuotes(String text) {
+    if (text != null && text.length() >= 2 && text.startsWith("\"") && text.endsWith("\"")) {
+      return text.substring(1, text.length() - 1).replace("''", "'");
+    }
+    return text;
+  }
+
+  private String stripBackQuotes(String quoted) {
```
Caideyipi, line 1453: Do we really need to support this?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/sql/parser/AstBuilder.java
```diff
@@ -1443,81 +1443,102 @@ private String stripQuotes(String text) {
...
+      return stripBackQuotes(id.getText());
+    }
+    throw new SemanticException("IdentifierContext not support");
+  }
+
   @Override
   public Node visitCreateUserStatement(RelationalSqlParser.CreateUserStatementContext ctx) {
     RelationalAuthorStatement stmt = new RelationalAuthorStatement(AuthorRType.CREATE_USER);
-    stmt.setUserName(ctx.userName.getText());
+    stmt.setUserName(handleStripQuotes(ctx.userName));
```
Caideyipi, line 1471: ((Identifier)visit(ctx.userName)).getValue()?

PR #15034, opened 2025-03-07
Fix Memory NPE when use IDEA to debug

SpriCoder:
Fix Memory NPE when use IDEA to debug. This is mainly because when there are no properties, init method in DataNodeMemoryConfig will never be called.


SpriCoder:
Before: 
image
After:
image

Caideyipi (approved):

OneSizeFitsQuorum (approved):
LGTM

PR #15036, opened 2025-03-07
[Fix](query) Fix the calc logic of dispatch read time

Beyyes:
## Description

As the title.


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/scheduler/FragmentInstanceDispatcherImpl.java
```diff
@@ -151,11 +153,13 @@ private Future<FragInstanceDispatchResult> dispatchRead(List<FragmentInstance> i
           instance.getFragment().clearTypeProvider();
         }
 
-        long dispatchReadTime = System.nanoTime() - startTime;
-        QUERY_EXECUTION_METRIC_SET.recordExecutionCost(DISPATCH_READ, dispatchReadTime);
-        queryContext.recordDispatchCost(dispatchReadTime);
+        long fragInstanceDispatchReadTime = System.nanoTime() - fragInstanceStartTime;
+        QUERY_EXECUTION_METRIC_SET.recordExecutionCost(DISPATCH_READ, fragInstanceDispatchReadTime);
```
ycycse, line 157: Wasn't the original monitoring metrics also cumulative? Why not use the same approach?

ycycse (approved):
LGTM

JackieTien97 (approved):

PR #15049, opened 2025-03-10
Load: Fix memory size counting on cases when query engine's memory is not sufficient

DanielWang2035:


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/load/memory/LoadTsFileMemoryManager.java
```diff
@@ -88,17 +88,24 @@ public synchronized void releaseToQuery(long sizeInBytes) {
...
+    QUERY_ENGINE_MEMORY_MANAGER.releaseToFreeMemoryForOperators(sizeToRelease);
     this.notifyAll();
   }
 
   public synchronized LoadTsFileAnalyzeSchemaMemoryBlock allocateAnalyzeSchemaMemoryBlock(
       long sizeInBytes) throws LoadRuntimeOutOfMemoryException {
     try {
       forceAllocateFromQuery(sizeInBytes);
+      LOGGER.info(
+          "Load: Allocated AnalyzeSchemaMemoryBlock from query engine, size: {}", sizeInBytes);
```
SteveYurongSu, line 102: is debug enabled -> log.debug

SteveYurongSu (approved):

PR #15122, opened 2025-03-18
RegionMigrateProcedure and RegionReconstructProcedure no longer acquire lock

liyuheng55555:
There are two reasons:
- The concurrency control for region migration and reconstruction has already been handled at a higher level, see ProcedureManager.regionOperationCommonCheck).
- This part of the code conflicts with the unified lock management implemented at the Dec of 2024 (#14250), which may lead to one procedure being executed by multiple threads simultaneously.

### DIFF
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/region/ReconstructRegionProcedure.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/region/ReconstructRegionProcedure.java
@@ -28,7 +28,6 @@ import org.apache.iotdb.confignode.procedure.env.ConfigNodeProcedureEnv;
 import org.apache.iotdb.confignode.procedure.exception.ProcedureException;
 import org.apache.iotdb.confignode.procedure.exception.ProcedureSuspendedException;
 import org.apache.iotdb.confignode.procedure.exception.ProcedureYieldException;
-import org.apache.iotdb.confignode.procedure.state.ProcedureLockState;
 import org.apache.iotdb.confignode.procedure.state.ReconstructRegionState;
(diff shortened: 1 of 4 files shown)


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/region/ReconstructRegionProcedure.java
```diff
@@ -125,38 +124,6 @@ protected void rollbackState(
...
-
-      LOGGER.info("procedureId {} wait for lock.", getProcId());
-      return ProcedureLockState.LOCK_EVENT_WAIT;
-    } finally {
-      configNodeProcedureEnv.getSchedulerLock().unlock();
-    }
-  }
-
-  @Override
-  protected void releaseLock(ConfigNodeProcedureEnv configNodeProcedureEnv) {
```
Pengzna, line 146: maybe still need to override this method?

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/region/ReconstructRegionProcedure.java
```diff
@@ -125,38 +124,6 @@ protected void rollbackState(
       ConfigNodeProcedureEnv configNodeProcedureEnv, ReconstructRegionState reconstructRegionState)
       throws IOException, InterruptedException, ProcedureException {}
 
-  @Override
-  protected ProcedureLockState acquireLock(ConfigNodeProcedureEnv configNodeProcedureEnv) {
```
Pengzna, line 129: ditto. still need to acquire scheduler lock

Pengzna (approved):
LGTM🔥💪🏻

OneSizeFitsQuorum (approved):
LGTM

PR #15129, opened 2025-03-19
IoTConsensusV2: Fix concurrency bug when client traffic is extremely high

Pengzna:
All details are described in comments between code.


OneSizeFitsQuorum (approved):
LGTM

PR #15141, opened 2025-03-19
Fix privilege bug in JDBC

LJW21-02:
# Key Changes
- used double try catch to get tables, columns, primary keys


CritasWang (approved):

CritasWang (approved):
LGTM

PR #15252, opened 2025-04-01
IoTV2: Clean receiver files when dropping consensus pipe & Improve robustness when cleaning some dirs.

Pengzna:
In original method which deletes receiver file, the parameter pipeName and map key are opposites. So we can always not find the correct receiver to delete file.

image


OneSizeFitsQuorum (commented):
LGTM

PR #15256, opened 2025-04-02
Get correct TTL in tree model

CRZbulabula:
We found that the current TTL getting interface in TTLManager cannot return the correct result when the name of database is consists with multiple layers. Thus, we improve the TTL search code to address this bug.

Additionally, we split the PartitionTableAutoCleaner IT into tree and table models, ensuring better robustness in future updates.


OneSizeFitsQuorum (approved):
LGTM

PR #15276, opened 2025-04-05
Pipe: use mem table to batch write tree data into tsfile

VGalaxies:
As title.


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/util/builder/PipeTreeModelTsFileBuilder.java
```diff
@@ -91,178 +105,201 @@ public synchronized void close() {
...
+              tablet.getRowSize());
+
+      // TODO: unused results
+      final TSStatus[] results = new TSStatus[insertTabletNode.getRowCount()];
+      Arrays.fill(results, RpcUtils.SUCCESS_STATUS);
+
+      final int loc = insertTabletNode.checkTTL(results, getTTL(insertTabletNode));
+      final List<Pair<IDeviceID, Integer>> deviceEndOffsetPairs =
+          insertTabletNode.splitByDevice(loc, insertTabletNode.getRowCount());
+      int start = loc;
```
luoluoyuyu, line 178: I think it is not necessary to check TTL here, because we need to ensure that the data can be synchronized. All that can be captured are the data that needs to be synchronized. The start should be 0.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/util/builder/PipeTreeModelTsFileBuilder.java
```diff
@@ -91,178 +105,201 @@ public synchronized void close() {
...
+    }
+
+    return sealedFiles;
+  }
+
+  private void writeTabletsIntoOneFile(final RestorableTsFileIOWriter writer) throws Exception {
+    // TODO: database & region
+    final String database = "test_db";
+    final String dataRegionId = "test_dr";
+    final IMemTable memTable = new PrimitiveMemTable(database, dataRegionId);
```
luoluoyuyu, line 130: ```memTable.memSize()``` can get the memory used by MemTable. Here, the memory used by MemTable needs to be included in Pipe memory management.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/util/builder/PipeTreeModelTsFileBuilder.java
```diff
@@ -91,178 +105,201 @@ public synchronized void close() {
...
+    final boolean isSequence = false;
+    while (loc < endOffset) {
+      final long time = insertTabletNode.getTimes()[loc];
+      final long timePartitionId = TimePartitionUtils.getTimePartitionId(time);
+
+      // judge if we should insert sequence
+      if (timePartitionId != beforeTimePartition) {
+        updateSplitInfo(splitInfo, beforeTimePartition, isSequence, new int[] {before, loc});
+        before = loc;
+        beforeTimePartition = timePartitionId;
```
luoluoyuyu, line 218: We don't need to distinguish between time partitions and whether the Memtable is sequential. We just need to write all the data of each tablet to the Memtable.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/util/builder/PipeTreeModelTsFileBuilder.java
VGalaxies: The code here completely refers to the method call chain of `org.apache.iotdb.db.storageengine.dataregion.DataRegion#insertTablet`. Some pruning has already been done, pending discussion.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/util/builder/PipeTreeModelTsFileBuilder.java
VGalaxies: It seems unnecessary, as memory control has already been implemented in the outer layer using PipeTreeModelTsFileBuilder, such as PipeTabletEventBatch. We should clearly identify whether the caller or callee performs memory control, and here it is the caller.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/util/builder/PipeTreeModelTsFileBuilder.java
luoluoyuyu: The work completed by the code of ```org.apache.iotdb.db.storageengine.dataregion.DataRegion#insertTablet``` is to find the memtable of the corresponding partition and whether the partition is written sequentially. However, for TsFileBuilder, there is no concept of time partition, nor is it necessary to distinguish whether it is written sequentially. It only needs to use the memTable disk flushing mechanism to achieve time series deduplication. Therefore, my suggestion is that this code is not needed. We only need to ensure that the internal time of Tablet is sequential.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/util/builder/PipeTreeModelTsFileBuilder.java
luoluoyuyu: The main reason here is that writing to memtable requires memory replication and requires secondary memory application, so my suggestion is not to temporarily save the tablet and write directly to memtable

luoluoyuyu (approved):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/util/builder/PipeTreeModelTsFileBuilder.java
```diff
@@ -24,6 +24,7 @@
 import org.apache.tsfile.file.metadata.IDeviceID;
 import org.apache.tsfile.read.common.Path;
 import org.apache.tsfile.utils.Pair;
+import org.apache.tsfile.write.TsFileWriter;
```
SteveYurongSu, line 27: rename as V1？ iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/util/builder/PipeTreeModelTsFileBuilderV1.java

SteveYurongSu (changes requested):

SteveYurongSu (approved):

PR #15290, opened 2025-04-07
Get device path from cache

shuwenwei:
## Description
Get device path from cache
## Test
Compaction With 500 devices and 20 measurements per device.
master:
A9zm5wZAI3
current branch:
截屏2025-04-14 14 29 37

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/utils/executor/fast/SeriesCompactionExecutor.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/utils/executor/fast/SeriesCompactionExecutor.java
@@ -24,6 +24,7 @@ import org.apache.iotdb.commons.path.PartialPath;
 import org.apache.iotdb.commons.path.PatternTreeMap;
 import org.apache.iotdb.db.exception.WriteProcessException;
 import org.apache.iotdb.db.storageengine.dataregion.compaction.execute.task.subtask.FastCompactionTaskSummary;
+import org.apache.iotdb.db.storageengine.dataregion.compaction.execute.utils.CompactionPathUtils;
 import org.apache.iotdb.db.storageengine.dataregion.compaction.execute.utils.executor.ModifiedStatus;
 import org.apache.iotdb.db.storageengine.dataregion.compaction.execute.utils.executor.fast.element.ChunkMetadataElement;
 import org.apache.iotdb.db.storageengine.dataregion.compaction.execute.utils.executor.fast.element.FileElement;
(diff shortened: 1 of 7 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/utils/executor/fast/SeriesCompactionExecutor.java
```diff
@@ -475,16 +476,16 @@ protected void removeFile(FileElement fileElement) throws IllegalPathException,
...
   protected List<ModEntry> getModificationsFromCache(
-      TsFileResource tsFileResource, PartialPath path) {
+      TsFileResource tsFileResource, IDeviceID deviceId, String measurement)
+      throws IllegalPathException {
     PatternTreeMap<ModEntry, PatternTreeMapFactory.ModsSerializer> allModifications =
         modificationCacheMap.get(tsFileResource.getTsFile().getName());
     if (allModifications == null) {
       return Collections.emptyList();
     }
+    PartialPath path = CompactionPathUtils.getPath(deviceId, measurement);
```
jt2594838, line 488: Can we avoid using PartialPath here? If so, maybe we can remove CompactionPathUtils.getPath completely and void array copy.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/modification/TreeDeletionEntry.java
```diff
@@ -133,16 +134,17 @@ public boolean affects(IDeviceID deviceID, long startTime, long endTime) {
   @Override
   public boolean affects(IDeviceID deviceID) {
     try {
+      PartialPath deviceIdPath =
+          DataNodeDevicePathCache.getInstance().getPartialPath(deviceID.toString());
```
jt2594838, line 138: Is it possible that we simply use deviceId as the key? toString may be costly for DeviceId.

jt2594838 (approved):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/utils/executor/fast/SeriesCompactionExecutor.java
shuwenwei: If we avoid using PartialPath here, it will be constructed in PatternTreeMap and we can not import DataNodeDevicePathCache class there.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/modification/TreeDeletionEntry.java
shuwenwei: The key of this cache is string and there are a lot of uses to get the value directly through string. Compared to constructing PartialPath directly, the cost of toString may not be obvious

jt2594838:
Let us first look at the results of the performance tests to see whether there is a need of further optimization.

PR #15294, opened 2025-04-08
Add retry for inactivePeer of region migration

liyuheng55555:
In region migration, if the target node restarts before the “addRegionPeer” phase within 1 minutes, the client in the ClientManager will become invalid.

This PR adds 1 retry at this point to ensure that region migration can still proceed correctly in such cases.


OneSizeFitsQuorum (approved):
LGTM

PR #15296, opened 2025-04-08
Remove dn_rpc_min_concurrent_client_num to fix minWorker > maxWorker bugs 

OneSizeFitsQuorum:
Fix bug below
image

In some environments, the number of cores will be less than dn_rpc_max_concurrent_client_num, and starting the datanode will cause an error


HTHou (approved):
lgtm

PR #15307, opened 2025-04-09
Subscription: poll heartbeat event when there are no consumers

VGalaxies:
As title.


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/broker/SubscriptionPrefetchingQueue.java
```diff
@@ -311,6 +313,29 @@ public void prefetchEvent(@NonNull final SubscriptionEvent thisEvent) {
...
+      return;
+    }
+
+    final Event polledEvent = inputPendingQueue.waitedPoll();
+    if (!Objects.equals(peekedEvent, polledEvent)) {
+      LOGGER.warn(
+          "Subscription: inconsistent heartbeat event when {} peeking (broken invariant), expected {}, actual {}, do nothing",
+          this,
+          peekedEvent,
+          polledEvent);
```
SteveYurongSu, line 332: Not enough?

SteveYurongSu (approved):

PR #15308, opened 2025-04-10
Implement ASOF INNER JOIN

Wei-hao-Li:


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/sql/parser/AstBuilder.java
```diff
@@ -2186,6 +2214,10 @@ public Node visitJoinRelation(RelationalSqlParser.JoinRelationContext ctx) {
       joinType = Join.Type.INNER;
     }
 
+    if (criteria instanceof AsofJoinOn && joinType != Join.Type.INNER) {
+      throw new SemanticException("ASOF JOIN is only support INNER type now");
+    }
```
JackieTien97, line 2219: why we need this restriction? left asof join should also support tolerance

JackieTien97 (changes requested):

PR #15340, opened 2025-04-14
Pipe: fix drop pipe stuck

DanielWang2035:


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/agent/task/subtask/PipeAbstractConnectorSubtask.java
```diff
@@ -46,6 +46,9 @@ public abstract class PipeAbstractConnectorSubtask extends PipeReportableSubtask
   // For thread pool to execute callbacks
   protected ExecutorService subtaskCallbackListeningExecutor;
 
+  protected final Object deregisterLock = new Object();
+  protected boolean isDeregistering = false;
```
luoluoyuyu, line 50: Hi @DanielWang2035
I think this method cannot fundamentally solve the problem of slow Drop Pipe. It is necessary to fundamentally optimize the lock structure of AsyncDataRegionConnector.

DanielWang2035: I think the problem is subtask-level: https://timechor.feishu.cn/docx/KvE2dbKSZoGn0Gxl60bcsQT2nXg

PR #15348, opened 2025-04-15
Include thrift genarated python code in python client to support AINode

HTHou:
## Description

As title.

### DIFF
--- a/iotdb-client/client-py/pom.xml
+++ b/iotdb-client/client-py/pom.xml
@@ -43,6 +43,18 @@
             <version>2.0.2-SNAPSHOT</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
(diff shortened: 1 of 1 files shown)


iotdb-client/client-py/pom.xml
```diff
@@ -114,13 +126,16 @@
                             <resources>
                                 <resource>
                                     <directory>${basedir}/../../iotdb-protocol/thrift-datanode/target/generated-sources-python/iotdb/thrift/</directory>
-                                    <includes>
-                                        <include>**/rpc/*</include>
-                                    </includes>
```
ycycse, line 119: Will this cause error in the import code of py-client? It seems like there is no rpc root module any more.

ycycse (approved):

PR #15371, opened 2025-04-19
Pipe: switch to IdentityHashMap backed set to avoid infinite loop in table tsfile builder

VGalaxies:
## Background

In `PipeTableModeTsFileBuilder.tryBestToWriteTabletsIntoOneFile(...)`, we collect per‑device lists of tablets into a `LinkedHashSet<LinkedList<T>>` and then peel off entries by calling `iterator.remove()` when a list becomes empty. However, because `LinkedList`’s `hashCode()`/`equals()` depend on its contents, repeatedly polling elements from the list changes its hash code. Once its hash no longer matches its original bucket, `iterator.remove()` silently fails to expunge the “empty” list, leaving behind a zero‑length `LinkedList` in the set.

## Problem

- Mutating a `LinkedList` stored in a hash‑based set invalidates the set’s internal invariants.
- After polling all elements, `iterator.remove()` cannot locate the entry by its mutated hash code.
- **The set ends up retaining empty lists and the loop never terminates cleanly.**

## Proposed Solution

Switch from a `LinkedHashSet<LinkedList<T>>` (which relies on `equals()`/`hashCode()`) to a reference‑identity set backed by an `IdentityHashMap`. Because identity‑based collections use `==` for comparison, mutating the list’s contents has no impact on its bucket placement. Removal via the iterator then works as expected.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/util/builder/PipeTableModeTsFileBuilder.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/util/builder/PipeTableModeTsFileBuilder.java
@@ -36,11 +36,12 @@ import org.slf4j.LoggerFactory;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.IdentityHashMap;
 import java.util.Iterator;
-import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


luoluoyuyu:
I found that this bug is related to JDK. In fact, this problem will not occur in JDK11, but in JDK8. Great fix, thanks @VGalaxies 

JDK 8
```
 public final void remove() {
 Node<K,V> p = current;
 if(p==null)
 throw new IllegalStateException();
 if (modCount != expectedModCount)
 throw new ConcurrentModificationException();
 current = null;
 K key = p.key;
 removeNode(hash(key), key, null, false, false);
 expectedModCount = modCount;
 }
```

JDK 11
``` public final void remove() {
 Node<K,V> p = current;
 if(p==null)
 throw new IllegalStateException();
 if (modCount != expectedModCount)
 throw new ConcurrentModificationException();
 current = null;
 removeNode(p.hash, p.key, null, false, false);
 expectedModCount = modCount;
 }
```

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/util/builder/PipeTableModeTsFileBuilder.java
```diff
@@ -130,24 +125,29 @@ List<Pair<String, File>> writeTableModelTabletsToTsFiles(
           });
     }
 
+    // Create a Set backed by an IdentityHashMap, so elements are compared by reference (==) rather
+    // than equals()/hashCode()
+    final Set<LinkedList<T>> device2TabletsLinkedList =
+        Collections.newSetFromMap(new IdentityHashMap<>());
```
luoluoyuyu, line 131: For this fix, I suggest using LinkedHashMap, storing the Key in TableName and the Value in List can also solve this problem well.

luoluoyuyu (approved):

SteveYurongSu (approved):

PR #15394, opened 2025-04-22
Add TableFunctionHandle for TVF

Cpaulyz:
## Description

Add TableFunctionHandle for TVF


JackieTien97 (changes requested):
still need to think about how to remove `arguments` from Node

JackieTien97 (approved):

PR #15395, opened 2025-04-22
DAL: Fix thread leak when DN exit.

Pengzna:
We should use `shutdownNow` instead of `shutdown` to send an interruption signal to thread.

Test has shown results as expected:
image


OneSizeFitsQuorum (approved):
LGTM

PR #15410, opened 2025-04-24
IoTV2: Try to fix tsfile corruption & receiver dir clean

Pengzna:
as title.


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/receiver/protocol/pipeconsensus/PipeConsensusReceiver.java
```diff
@@ -117,6 +119,7 @@ public class PipeConsensusReceiver {
   private final PipeConsensusReceiverMetrics pipeConsensusReceiverMetrics;
   private final FolderManager folderManager;
   private final AtomicBoolean isClosed = new AtomicBoolean(false);
+  private final ReadWriteLock tsFilePieceWriteLock = new ReentrantReadWriteLock(true);
```
OneSizeFitsQuorum, line 122: why name a ReentrantReadWriteLock to a WriteLock?

Pengzna: a mistake, fixed

PR #15415, opened 2025-04-26
IoTV2: Fix global client manager close by mistake when running

Pengzna:
as title.

All global client managers should only be closed in `PipeConsensus`
image


OneSizeFitsQuorum (approved):
LGTM

PR #15418, opened 2025-04-27
Fix memory leak in wal compressed buffer

jt2594838:
After calling `org.apache.iotdb.db.storageengine.dataregion.wal.io.WALInputStream#skipToGivenLogicalPosition`, the compressed buffer is cleaned but not nullified.
As a result, the succinct call of `org.apache.iotdb.db.storageengine.dataregion.wal.io.WALInputStream#loadNextSegmentV2` may use the cleaned buffer and cause memory issues.


OneSizeFitsQuorum (approved):
LGTM

PR #15434, opened 2025-04-29
Preallocate array list & update TsFile version

jt2594838:
Some ArrayLists are initialized with sizes that are too small in massive series scenarios, which are now allocated with proper sizes in the context.


OneSizeFitsQuorum (approved):
LGTM

PR #15437, opened 2025-04-29
Append TTL logs for PartitionTableAutoCleaner

CRZbulabula:
In integration-tests, we found the corresponding IT for the PartitionTableAutoCleaner fails frequently. To facilitate bug location, we append some logs for the PartitionTableAutoCleaner.


OneSizeFitsQuorum (approved):
LGTM

PR #15454, opened 2025-05-06
Move built-in TVF from common sub-module into datanode sub-module

JackieTien97:
Built-in TVF should be in datanode package instead of common package. in datanode package, it can share some classes in datanode


Cpaulyz (approved):
LGTM

PR #15495, opened 2025-05-13
[AINode] Fixed abnormal inference results in timerxl

jtmer:
## Description
Fixed the issue of abnormal output when the input length of timerxl is not a multiple of 96

<hr>

This PR has:
- [x] been self-reviewed.

<hr>


DigitalLifeYZQiu (approved):
Nice job!

PR #15499, opened 2025-05-14
[RTO/RPO] Fix NPE error and improve log message

SzyWilliam:
1. Fix npe in topologyService.
2. Add log id in Phi Accural detector and in session retry.

### DIFF
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/load/service/TopologyService.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/load/service/TopologyService.java
@@ -228,7 +228,7 @@ public class TopologyService implements Runnable, IClusterStatusSubscriber {
           && !failureDetector.isAvailable(entry.getKey(), entry.getValue())) {
         LOGGER.debug("Connection from DataNode {} to DataNode {} is broken", fromId, toId);
       } else {
-        latestTopology.get(fromId).add(toId);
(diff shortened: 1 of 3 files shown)


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/load/service/TopologyService.java
```diff
@@ -228,7 +228,7 @@ private synchronized void topologyProbing() {
           && !failureDetector.isAvailable(entry.getKey(), entry.getValue())) {
         LOGGER.debug("Connection from DataNode {} to DataNode {} is broken", fromId, toId);
       } else {
-        latestTopology.get(fromId).add(toId);
+        Optional.ofNullable(latestTopology.get(fromId)).ifPresent(s -> s.add(toId));
```
CRZbulabula, line 231: Why not just computeIfAbsent? Intuitively, I bet the bug is merely due to ur `latestTopology` lost some keys. And the `latestTopology` should cache all nodeIds as the key

CRZbulabula (commented):
PTAL

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/load/service/TopologyService.java
SzyWilliam: Let me change to computeIfAbsent. Some keys are **intentionally** left out because these datanodes are starting and cannot answer to probing RPC requests.

CRZbulabula (approved):
LGTM!

PR #15509, opened 2025-05-15
forbid changing encrypt key again

zhujt20:
Changing encrypt key for tsfile encryption is not permitted after the encrypt function is open

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/conf/IoTDBStartCheck.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/conf/IoTDBStartCheck.java
@@ -29,7 +29,9 @@ import org.apache.iotdb.consensus.ConsensusFactory;
 import org.apache.iotdb.db.storageengine.dataregion.wal.utils.WALMode;
 import org.apache.iotdb.db.storageengine.rescon.disk.DirectoryChecker;
 
+import com.google.common.base.Objects;
 import org.apache.commons.io.FileUtils;
+import org.apache.tsfile.encrypt.EncryptUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/conf/IoTDBStartCheck.java
```diff
@@ -333,4 +347,23 @@ public void generateOrOverwriteSystemPropertiesFile() throws IOException {
...
+    if (encryptMagicString != null) {
+      byte[] magicBytes = EncryptUtils.hexStringToByteArray(encryptMagicString);
+      if (Objects.equal(magicString, new String(magicBytes))) {
+        serializeEncryptMagicString();
+        return;
+      }
+      String newMagicString =
+          new String(EncryptUtils.getEncrypt().getDecryptor().decrypt(magicBytes));
+      if (!Objects.equal(magicString, newMagicString)) {
+        logger.error("encrypt_magic_string is not match");
```
jt2594838, line 363: match -> matched

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/conf/IoTDBStartCheck.java
```diff
@@ -333,4 +347,23 @@ public void generateOrOverwriteSystemPropertiesFile() throws IOException {
...
+    properties = systemPropertiesHandler.read();
+    String encryptMagicString = properties.getProperty("encrypt_magic_string");
+    if (encryptMagicString != null) {
+      byte[] magicBytes = EncryptUtils.hexStringToByteArray(encryptMagicString);
+      if (Objects.equal(magicString, new String(magicBytes))) {
+        serializeEncryptMagicString();
+        return;
+      }
+      String newMagicString =
+          new String(EncryptUtils.getEncrypt().getDecryptor().decrypt(magicBytes));
```
jt2594838, line 361: Will changing the key cause decrypt() to throw an exception?
If it will, it's better to catch the exception and convert it to an understandable one.

jt2594838 (approved):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/conf/IoTDBStartCheck.java
```diff
@@ -333,4 +347,23 @@ public void generateOrOverwriteSystemPropertiesFile() throws IOException {
...
     systemPropertiesHandler.overwrite(properties);
   }
+
+  public void checkEncryptMagicString() throws IOException, ConfigurationException {
+    properties = systemPropertiesHandler.read();
+    String encryptMagicString = properties.getProperty("encrypt_magic_string");
+    if (encryptMagicString != null) {
+      byte[] magicBytes = EncryptUtils.hexStringToByteArray(encryptMagicString);
+      if (Objects.equal(magicString, new String(magicBytes))) {
+        serializeEncryptMagicString();
```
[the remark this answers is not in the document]
zhujt20: This is for permitting the restart with the transition from unencrypted to encrypted. But after the discussion, we decide to forbid it, too.

PR #15523, opened 2025-05-16
Enable the kernel to support passing topics to MQTT custom plugins

wenyanshi-123:
Enable the kernel to support passing topics to MQTT custom plugins.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/mqtt/LinePayloadFormatter.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/mqtt/LinePayloadFormatter.java
@@ -63,6 +63,7 @@ public class LinePayloadFormatter implements PayloadFormatter {
   }
 
   @Override
+  @Deprecated
   public List<Message> format(ByteBuf payload) {
     List<Message> messages = new ArrayList<>();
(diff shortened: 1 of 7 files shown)


HTHou (approved):
lgtm

HTHou:
Please fix the failed UTs.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/mqtt/LinePayloadFormatter.java
```diff
@@ -63,7 +64,7 @@ public LinePayloadFormatter() {
   }
 
   @Override
-  public List<Message> format(ByteBuf payload) {
+  public List<Message> format(String topic, ByteBuf payload) {
```
CritasWang, line 67: Since the topics can be obtained here, I think the logic for setting up the database should be included here

```
        if (useTableInsert) {
          TableMessage tableMessage = (TableMessage) message;
          // '/' previously defined as a database name
          String database =
              !msg.getTopicName().contains("/")
                  ? msg.getTopicName()
                  : msg.getTopicName().substring(0, msg.getTopicName().indexOf("/"));
          tableMessage.setDatabase(database);
```

CritasWang (approved):

PR #15563, opened 2025-05-22
Remove all kinds of ProcedureException's subclass & Fix a security problem

liyuheng55555:
According to the Apache Security Team’s report, there is a security issue in the procedure module of ConfigNode. By crafting specific procedure files, an attack can be executed. 

I believe the root cause lies in the uncontrolled use of Class.forName, so I removed this call and ensured that only the ProcedureException class is used here. **The mainly modification is in Procedure.java.**

Its various subclasses have not been practically used in the past three years, so I deleted them to simplify the code.


Caideyipi (approved):
LGTM

HTHou (approved):

CRZbulabula (approved):
LGTM! Thanks ur great work, which significantly simplifies the Procedure moduler in ConfigNode :)

PR #15532, opened 2025-05-19
Fix null pointer exception casued by measurements values and measureSchemas values are not consistent

zerolbsony:
Fix the problem of null pointer below: 
sql覆盖 IoTV2 3副本 DN
NullPointerException: Cannot invoke "org.apache.tsfile.write.schema.MeasurementSchema.serializeTo(java.io.OutputStream)" because "this.measurementSchemas[i]" is null

`
2025-05-06 13:19:02,468 [pool-35-IoTDB-into-operation-executor-2] ERROR o.a.i.d.u.ErrorHandlingUtils:67 - Status code: EXECUTE_STATEMENT_ERROR(301), operation: insertTablets failed
java.lang.NullPointerException: Cannot invoke "org.apache.tsfile.write.schema.MeasurementSchema.serializeTo(java.io.OutputStream)" because "this.measurementSchemas[i]" is null
        at org.apache.iotdb.db.queryengine.plan.planner.plan.node.write.InsertTabletNode.writeMeasurementsOrSchemas(InsertTabletNode.java:502)
        at org.apache.iotdb.db.queryengine.plan.planner.plan.node.write.InsertTabletNode.subSerialize(InsertTabletNode.java:463)
        at org.apache.iotdb.db.queryengine.plan.planner.plan.node.write.InsertMultiTabletsNode.serializeAttributes(InsertMultiTabletsNode.java:246)
        at org.apache.iotdb.db.queryengine.plan.planner.plan.node.PlanNode.serialize(PlanNode.java:148)
        at org.apache.iotdb.db.queryengine.plan.planner.plan.node.PlanNode.serializeToByteBuffer(PlanNode.java:184)
        at org.apache.iotdb.db.queryengine.plan.scheduler.AsyncPlanNodeSender.<init>(AsyncPlanNodeSender.java:77)
        at org.apache.iotdb.db.queryengine.plan.scheduler.FragmentInstanceDispatcherImpl.dispatchWriteAsync(FragmentInstanceDispatcherImpl.java:220)
        at org.apache.iotdb.db.queryengine.plan.scheduler.FragmentInstanceDispatcherImpl.dispatch(FragmentInstanceDispatcherImpl.java:123)
        at org.apache.iotdb.db.queryengine.plan.scheduler.ClusterScheduler.start(ClusterScheduler.java:117)
        at org.apache.iotdb.db.queryengine.plan.planner.TreeModelPlanner.doSchedule(TreeModelPlanner.java:158)
[...]

PR #15574, opened 2025-05-23
Add a judgment in order to avoid occur NPE when one value of measurementSchemas is null

zerolbsony:
Add a judgment in order to avoid occur NPE when it's value is null for measurementSchemas.

NullPointerException: Cannot invoke "org.apache.iotdb.db.pipe.resource.memory.InsertNodeMemoryEstimator.sizeOfValues(Object[], MeasurementSchema[])" because "measurementSchemas[i]" is null

related pull request: https://github.com/apache/iotdb/pull/15532


HTHou (approved):

HTHou (changes requested):
There is another NPE needs to fix.
image

PR #15577, opened 2025-05-24
Add new udf UDAFMKIdentify

Beasttt:
## Add new udf UDAFMKIdentify

<hr>

This PR has:
- [✅] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


example/udf/src/main/java/org/apache/iotdb/udf/table/MatchingKeyState.java
```diff
@@ -0,0 +1,623 @@
...
+  int columnLength;
+  int totalPairs;
+  int[] distanceMin;
+  int[] distanceMax;
+  List<TupleEntry> tupleList;
+  Set<String> positivePairs;
+  Set<String> negativePairs;
+  Map<Integer, Set<Integer>> distanceMap;
+  Map<String, int[]> distanceCache;
+  Set<String> CandidateCache;
```
JackieTien97: make it private

example/udf/src/main/java/org/apache/iotdb/udf/table/MatchingKeyState.java
```diff
@@ -0,0 +1,623 @@
...
+    columnLength = 0;
+    distanceMap = new HashMap<>();
+    distanceCache = new HashMap<>();
+    CandidateCache = new HashSet<>();
+    labelIndex = 0;
+    min_support = 0.0;
+    min_confidence = 0.0;
+  }
+
+  public byte[] serialize() {
```
JackieTien97, line 122: add UT for serialize and deserialize method, make sure you can deserialize from what you serialize and the deserialized object is same as the previous one

example/udf/src/main/java/org/apache/iotdb/udf/UDAFMKIdentify.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.List;
+import java.util.Set;
+
+public class UDAFMKIdentify implements AggregateFunction {
+  int label;
+  int length;
+  double min_confidence;
+  double min_support;
+  String a;
+  int l = 0;
```
JackieTien97: make them private

example/udf/src/main/java/org/apache/iotdb/udf/UDAFMKIdentify.java
```diff
@@ -0,0 +1,157 @@
...
+
+  private FunctionArguments arguments;
```
JackieTien97: don't store this in your function, `UDAFMKIdentify.analyze` will only be called in query plan stage, and then other methods like `addInput` will be called in query execution stage. It's not the same object in these two stages, so the field you init in analyze won't be accessible to `addInput`

example/udf/src/main/java/org/apache/iotdb/udf/UDAFMKIdentify.java
```diff
@@ -0,0 +1,157 @@
...
+    label = input.getInt(length) - 1;
+    min_support = input.getDouble(length + 1);
+    min_confidence = input.getDouble(length + 2);
+    mkState.init(label, min_support, min_confidence);
+    length = length - 1;
+    mkState.setColumnLength(length);
+    long time = input.getLong(0);
+    String[] fullTuple = new String[length];
+    for (int i = 0; i < length; i++) {
+      Type col0Type = arguments.getDataType(i + 1);
```
JackieTien97: you can use `Type getDataType(int columnIndex);` in Record instead

JackieTien97 (changes requested):
Add IT about this UDAF

example/udf/src/main/java/org/apache/iotdb/udf/UDAFMKIdentify.java
```diff
@@ -0,0 +1,171 @@
...
+   * "PlateText", "VehicleID", "CameraID") values ('2025-01-01T01:56:19.000+08:00', 'B0D3O444s',
+   * '100', '78'); INSERT INTO test(Time, "PlateText", "VehicleID", "CameraID") values
+   * ('2025-01-01T01:56:39.000+08:00', 'B0D3O444s', '100', '78'); INSERT INTO test(Time,
+   * "PlateText", "VehicleID", "CameraID") values ('2025-01-01T03:30:59.000+08:00', 'B0134s00', '2',
+   * '150');
+   *
+   * <p>CREATE FUNCTION mkidentify AS 'org.apache.iotdb.udf.UDAFMK'; SHOW FUNCTIONS;
+   *
+   * <p>SELECT mkidentify(Time, PlateText, VehicleID, CameraID, 2, 0.2, 1.0) FROM root.test;
+   */
```
JackieTien97, line 36: It's an IT case, shouldn't put it in source code, we need to move it to integration-test/src/test/java/org/apache/iotdb/relational/it/udaf/

JackieTien97 (changes requested):
mv the implementation into library-udf sub-module

PR #15631, opened 2025-06-03
Fix tree ttl in table view scan

shuwenwei:
## Description
Fix tree ttl in table view scan


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/source/SeriesScanUtil.java
```diff
@@ -183,7 +183,7 @@ public void initQueryDataSource(QueryDataSource dataSource) {
     // Only the data in the table model needs to retain rows where all value
     // columns are null values, so we can use isIgnoreAllNullRows to
     // differentiate the data of tree model and table model.
```
JackieTien97: change the comment

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/source/SeriesScanUtil.java
```diff
@@ -183,7 +183,7 @@ public void initQueryDataSource(QueryDataSource dataSource) {
...
-    if (context.isIgnoreAllNullRows()) {
+    if (context.isIgnoreAllNullRows() || scanOptions.isTableViewForTreeModel()) {
```
JackieTien97, line 188: we need to pick the min for tree & view

JackieTien97 (changes requested):

PR #15644, opened 2025-06-04
Pipe: Add retry when TsFile parsing failed to avoid race among processor threads (follow up #15624)

DanielWang2035:


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/connection/PipeEventCollector.java
```diff
@@ -144,32 +141,9 @@ private void parseAndCollectEvent(final PipeTsFileInsertionEvent sourceEvent) th
...
-                  sourceEvent.getTsFile(),
-                  retryCount,
-                  e);
-            }
-          }
-        }
-      }
+      sourceEvent.consumeTabletInsertionEventsWithRetry(
+          this::collectParsedRawTableEvent,
+          this.getClass().getSimpleName() + "::parseAndCollectEvent");
```
Caideyipi: Better use a "static final" variable instead of compute this string each time...

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/processor/downsampling/DownSamplingProcessor.java
```diff
@@ -45,6 +48,7 @@
 import static org.apache.iotdb.commons.pipe.config.constant.PipeProcessorConstant.PROCESSOR_DOWN_SAMPLING_SPLIT_FILE_KEY;
 
 public abstract class DownSamplingProcessor implements PipeProcessor {
+  private static final Logger LOGGER = LoggerFactory.getLogger(DownSamplingProcessor.class);
```
Caideyipi: Is this used?

PR #15654, opened 2025-06-05
[AINode] Enable log record by default

CRZbulabula:
This is a trivial function that does not need further description.


ycycse (approved):
LGTM

PR #15656, opened 2025-06-06
Fix arrary index out-of-bounds in agg query with align by device when cross region

Wei-hao-Li:


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/distribution/SourceRewriter.java
```diff
@@ -313,6 +353,21 @@ public List<PlanNode> visitDeviceView(DeviceViewNode node, DistributionPlanConte
...
+  private static List<Expression> getWithDeviceExpressions(
+      FunctionExpression aggExpression, String device) {
+    return aggExpression.getExpressions().stream()
+        .map(
+            // process each argument of FunctionExpression
+            argument -> {
+              checkArgument(
+                  argument instanceof TimeSeriesOperand,
+                  "Argument of AggregationFunction should be TimeSeriesOperand here");
+              return new TimeSeriesOperand(
```
JackieTien97, line 365: add dataType

PR #15675, opened 2025-06-09
Add strong password related configs

jt2594838:
This PR adds the infrastructure that can be used to enable security constraints over password.


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeDescriptor.java
```diff
@@ -755,6 +755,31 @@ public boolean isSeedConfigNode() {
...
+                    String.valueOf(commonDescriptor.getConfig().getPasswordExpirationSeconds()))));
+
+    commonDescriptor
+        .getConfig()
+        .setPasswordReuseIntervalSeconds(
+            Long.parseLong(
+                properties.getProperty(
+                    "password_reuse_interval_seconds",
+                    String.valueOf(
+                        commonDescriptor.getConfig().getPasswordReuseIntervalSeconds()))));
```
HTHou: Consider adding some check for `<= 0` like this?
https://github.com/apache/iotdb/blob/c874b3c9c6c9e682747872f221db6ecd000096ec/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeDescriptor.java#L703-L715

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/DataNodeAuthUtils.java
```diff
@@ -0,0 +1,167 @@
...
+        passwordChangeTime,
+        currentTimeMillis);
+  }
+
+  public static TSStatus recordPassword(String username, String password, String oldPassword) {
+    InsertRowStatement insertRowStatement = new InsertRowStatement();
+    try {
+      insertRowStatement.setDevicePath(
+          new PartialPath(SystemConstant.PREFIX_PASSWORD_HISTORY + ".`" + username + "`"));
+      insertRowStatement.setTime(System.currentTimeMillis());
```
HTHou: Should we consider the system timestamp precision here?

jt2594838: Done

scripts/sbin/start-cli.sh
```diff
@@ -61,7 +61,7 @@ fi
 
 # Added parameters when default parameters are missing
 user_param="-u root"
-passwd_param="-pw root"
+passwd_param="-pw IoTDB@2017"
```
HTHou: Just come up a question. If upgrading from an old version, the password is still `root`. Will using this CLI without input password meet error?

jt2594838: Yes, that cannot be avoided. 
The user should modify the scripts manually or avoid upgrading the scripts.

PR #15680, opened 2025-06-09
User password encryption upgraded from MD5 to SHA256. 

761417898:
User password encryption upgraded from MD5 to SHA256. Existing users' passwords will be upgraded upon next login.


jt2594838 (approved):

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/security/encrypt/MessageDigestEncrypt.java
```diff
@@ -29,16 +29,15 @@
 public class MessageDigestEncrypt implements AsymmetricEncrypt {
   private static final Logger logger = LoggerFactory.getLogger(MessageDigestEncrypt.class);
 
-  private static final String ENCRYPT_ALGORITHM = "MD5";
   private static final String STRING_ENCODING = "utf-8";
```
HTHou, line 32: Why don't we implement another AsymmetricEncrypt such as SHA256MessageDigestEncrypt?

hongzhi-gao: Both SHA - 256 and MD5 use the MessageDigest class, and we just need to pass different algorithm names when calling MessageDigest.getInstance(). Implementing another AsymmetricEncrypt like SHA256MessageDigestEncrypt would be redundant as it's essentially the same logic.

PR #15684, opened 2025-06-09
IoTV2: Remove events that do not retry in leader's transfer buffer & fix concurrency bug of retryQueue & fix retry interval & refine log

Pengzna:
as title.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/protocol/pipeconsensus/PipeConsensusAsyncConnector.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/protocol/pipeconsensus/PipeConsensusAsyncConnector.java
@@ -72,10 +72,10 @@ import java.io.IOException;
 import java.util.Comparator;
 import java.util.Iterator;
 import java.util.Objects;
-import java.util.PriorityQueue;
 import java.util.Queue;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingDeque;
+import java.util.concurrent.PriorityBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -101,7 +101,7 @@ public class PipeConsensusAsyncConnector extends IoTDBConnector implements Conse
   private static final long PIPE_CONSENSUS_EVENT_ENQUEUE_TIMEOUT_IN_MS =
       IOTDB_CONFIG.getConnectionTimeoutInMS() / 6;
   private final Queue<EnrichedEvent> retryEventQueue =
-      new PriorityQueue<>(
+      new PriorityBlockingQueue<>(
           IOTDB_CONFIG.getIotConsensusV2PipelineSize(),
           Comparator.comparingLong(EnrichedEvent::getReplicateIndexForIoTV2));
   // We use enrichedEvent here to make use of EnrichedEvent.equalsInPipeConsensus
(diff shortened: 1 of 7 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/protocol/pipeconsensus/PipeConsensusAsyncConnector.java
```diff
@@ -535,9 +535,10 @@ private void asyncTransferQueuedEventsIfNecessary() {
                 ? peekedEvent.getRetryInterval()
                 : 0L;
         LOGGER.info(
-            "PipeConsensus-ConsensusGroup-{}: retry with interval {} for {}",
+            "PipeConsensus-ConsensusGroup-{}: retry with interval {} for index {} {}",
             consensusGroupId,
             retryInterval,
+            peekedEvent.getReplicateIndexForIoTV2(),
             peekedEvent);
```
jt2594838, line 542: Maybe this kind of event should just include replicateIndex in its toString() method.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/protocol/pipeconsensus/PipeConsensusAsyncConnector.java
```diff
@@ -616,7 +617,7 @@ private void retryTransfer(final PipeDeleteDataNodeEvent deleteDataNodeEvent) {
    * @param event event to retry
    */
   @SuppressWarnings("java:S899")
-  public void addFailureEventToRetryQueue(final EnrichedEvent event) {
+  public synchronized void addFailureEventToRetryQueue(final EnrichedEvent event) {
```
jt2594838, line 620: Why add synchronized?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/protocol/pipeconsensus/PipeConsensusAsyncConnector.java
```diff
@@ -630,11 +631,20 @@ public void addFailureEventToRetryQueue(final EnrichedEvent event) {
...
+          consensusGroupId,
+          event,
+          event.getReplicateIndexForIoTV2());
+    } else {
+      LOGGER.warn(
+          "PipeConsensus-ConsensusGroup-{}: Event {} replicate index {} transfer failed, added to retry queue failed, this event will be ignored.",
+          consensusGroupId,
+          event,
+          event.getReplicateIndexForIoTV2());
+    }
```
jt2594838, line 647: You may, by the way, change "PipeConsensus"  to "IoTV2Consensus" in the logs to avoid someone being confused.

jt2594838 (approved):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/protocol/pipeconsensus/PipeConsensusAsyncConnector.java
Pengzna: yes, but its too long and not convenient for me to check log. I add this mark for quick check

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/protocol/pipeconsensus/PipeConsensusAsyncConnector.java
Pengzna: keep serialization for this method(enqueue) and `asyncTransferQueuedEventsIfNecessary`(dequeue), in case that one thread is infinitely polling events in `asyncTransferQueuedEventsIfNecessary` because there are always enqueue events.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/protocol/pipeconsensus/PipeConsensusAsyncConnector.java
Pengzna: good idea. I will do that in a further pr

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/connector/protocol/pipeconsensus/PipeConsensusAsyncConnector.java
jt2594838: So what. If there is something to be transferred, why can't we transfer it?

PR #15699, opened 2025-06-11
Pipe: Add upper bound check for Pipe request decompression buffer

XNX02:


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/connector/payload/thrift/request/PipeTransferCompressedReq.java
```diff
@@ -144,6 +149,17 @@ public static byte[] toTPipeTransferReqBytes(
     }
   }
 
+  /** This method is used to prevent decompression bomb attacks. */
+  private static void checkDecompressedLength(final int decompressedLength)
+      throws IllegalArgumentException {
+    if (decompressedLength < 0 || decompressedLength > MAX_DECOMPRESSED_LENGTH) {
```
SteveYurongSu:
```suggestion
    final int maxDecompressedLength = PipeConfig.getInstance().getPipeReceiverReqDecompressedMaxLengthInBytes();
    if (decompressedLength < 0 || decompressedLength > maxDecompressedLength) {
```

SteveYurongSu (approved):

PR #15711, opened 2025-06-12
use environment variable and PBKDF to generate main encrypt key

zhujt20:
use environment variable and PBKDF(hmac-sha256) to generate main encrypt key

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/conf/IoTDBStartCheck.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/conf/IoTDBStartCheck.java
@@ -314,7 +314,7 @@ public class IoTDBStartCheck {
       String token = System.getenv("user_encrypt_token");
       if (token == null || token.trim().isEmpty()) {
         throw new EncryptException(
-            "encryptType is not UNENCRYPTED, but user_encrypt_token is not set.");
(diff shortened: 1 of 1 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/conf/IoTDBStartCheck.java
```diff
@@ -306,11 +307,19 @@ public void serializeClusterID(String clusterId) throws IOException {
...
 
   public void serializeEncryptMagicString() throws IOException {
+    if (!Objects.equals(TSFileDescriptor.getInstance().getConfig().getEncryptType(), "UNENCRYPTED")
+        && !Objects.equals(
+            TSFileDescriptor.getInstance().getConfig().getEncryptType(),
+            "org.apache.tsfile.encrypt.UNENCRYPTED")) {
+      String token = System.getenv("user_encrypt_token");
+      if (token == null || token.trim().isEmpty()) {
+        throw new EncryptException(
+            "encryptType is not UNENCRYPTED, but user_encrypt_token is not set.");
```
jt2594838: The user may mistake it for a configuration item in iotdb-system.properties, state it more clearly like:
"encryptType is not UNENCRYPTED, but user_encrypt_token is not set. Please set it in the environment variable."

jt2594838 (approved):

PR #15728, opened 2025-06-16
Enhance "remove region" sql to handle the case of DataNode not exist

liyuheng55555:
In exceptionally rare situations, a node may have been deleted from the node table ( NodeManager )—let’s call it node A—yet regions associated with node A still remain in the partition table ( PartitionManager ). 

To enable their cleanup, the PR refines the `REMOVE REGION` command so it can handle cases where node A is no longer present in the node table.

For this case, running `REMOVE REGION` fixes only the CN-side partition table; once `SHOW REGIONS` no longer displays the anomalous regions, you must restart all DNs to complete the repair on the DN side.


CRZbulabula (approved):
LGTM

PR #15746, opened 2025-06-17
Fix TypeProvider and AggregationDescriptor serde in Templated Query

Wei-hao-Li:


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/OperatorTreeGenerator.java
```diff
@@ -705,6 +706,8 @@ public Operator visitAlignedSeriesAggregationScan(
         scanOrder = scanOrder.reverse();
         aggregationDescriptors = context.getTemplatedInfo().getDescendingDescriptorList();
       }
+      AggregationStep step = node.getAggregationDescriptorList().get(0).getStep();
+      aggregationDescriptors.forEach(aggregationDescriptor -> aggregationDescriptor.setStep(step));
```
JackieTien97: reuse the aggregationDescriptors in plannode

JackieTien97 (changes requested):

JackieTien97 (approved):

PR #15790, opened 2025-06-19
Subscription: intro poll and prefetch v2 for tsfile topic

VGalaxies:
The core idea of v2 is:
1. Polling should not drive prefetch; instead, during the remaining time, it should poll the prefetching queue.
2. Prefetch should handle tsfile events specially by performing stateful, on-demand parsing (with retries if necessary).


VGalaxies:
image

image

image

SteveYurongSu (approved):

PR #15799, opened 2025-06-23
[AINode] Update the class fields of IoTDB dataset

CRZbulabula:
The `seq_len`, `input_token_len` and `output_token_len` are required by each model, hence we clear up the class fields of dataset in this PR.

### DIFF
--- a/iotdb-core/ainode/ainode/core/ingress/iotdb.py
+++ b/iotdb-core/ainode/ainode/core/ingress/iotdb.py
@@ -56,18 +56,19 @@ class IoTDBTreeModelDataset(BasicDatabaseForecastDataset):
     def __init__(
         self,
         model_id: str,
-        input_len: int,
-        out_len: int,
+        seq_len: int,
+        input_token_len: int,
+        output_token_len: int,
         data_schema_list: list,
         ip: str = "127.0.0.1",
         port: int = 6667,
         username: str = "root",
         password: str = "root",
         time_zone: str = "UTC+8",
-        start_split: float = 0,
-        end_split: float = 1,
+        # start_split: float = 0,
+        # end_split: float = 1.0,
     ):
-        super().__init__(ip, port, input_len, out_len)
+        super().__init__(ip, port, seq_len, input_token_len, output_token_len)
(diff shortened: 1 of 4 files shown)


iotdb-core/ainode/ainode/core/ingress/iotdb.py
```diff
@@ -56,18 +56,19 @@ class IoTDBTreeModelDataset(BasicDatabaseForecastDataset):
...
         data_schema_list: list,
         ip: str = "127.0.0.1",
         port: int = 6667,
         username: str = "root",
         password: str = "root",
         time_zone: str = "UTC+8",
-        start_split: float = 0,
-        end_split: float = 1,
+        # start_split: float = 0,
+        # end_split: float = 1.0,
```
ycycse: I think we could just remove this. Data split in the future can be implemented by a more general way(like sampler)

ycycse (approved):

iotdb-core/ainode/ainode/core/ingress/iotdb.py
CRZbulabula: These parameters are inconvinient, thus I refactor the input of this class, PTAL

iotdb-core/ainode/ainode/core/ingress/iotdb.py
```diff
@@ -84,9 +84,9 @@ def __init__(
             zone_id=time_zone,
         )
         self.session.open(False)
+        self.use_rate = use_rate
+        self.offset_rate = offset_rate
```
ycycse, line 88: I think these two names are ambiguous. 
* If we only define the training set by its size, how can we get the validation set, considering the DataLoader shuffles the indices randomly.
* And why we need randow offset in one window, we already get the window randomly.

CRZbulabula: I agree with ur opportunity, yet my implementation is the easiest way, let's endure this bad code for now.

CRZbulabula (commented):
the conversation is resolved

PR #15801, opened 2025-06-23
[to dev/1.3] Prevent the contents of log_datanode_sample_query.log and log_explain_ananlyze.log from being output to log_datanode_all.log

shuwenwei:
## Description
Prevent the contents of log_datanode_sample_query.log and log_explain_ananlyze.log from being output to log_datanode_all.log


iotdb-core/datanode/src/assembly/resources/conf/logback-datanode.xml
```diff
@@ -260,7 +260,7 @@
     <logger level="info" name="SLOW_SQL">
```
JackieTien97:
```suggestion
    <logger level="info" name="SLOW_SQL" additivity="false">
```

PR #15815, opened 2025-06-24
IoTV2: Refactor replicate index so that it is shared at the pipe task level & Add some log for delete local peer

Pengzna:
Refactor replicate index so that it is shared at the **pipe task** level INSTEAD OF **region level** for 2 reasons:
1. prevent index jumps when events pile up and tasks are restarted
2. historicalExtractor is shared at the pipe task level, so When a certain pipetask restarts extracting historical data, it may interfere with the global index.

In this pr:
1. index is shared at the pipe task level
2. index is reset when pipe task is dropped. And can ensure that after index is reset, only when the pipe task restarts can this index increases.
3. index is used for calculating syncLag metric accurately.
4. Add some log for delete local peer


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/subtask/connector/PipeConnectorSubtaskManager.java
```diff
@@ -201,6 +202,9 @@ public synchronized void deregister(
     }
 
     PipeEventCommitManager.getInstance().deregister(pipeName, creationTime, regionId);
+    // Reset IoTV2 replicate index to prevent index jumps. Do this when consensus pipe no longer
+    // replicates data, since extractor and processor is already dropped now.
```
jt2594838: consensus pipe -> a consensus pipe
is already dropped -> are already dropped

Pengzna: fixed

PR #15816, opened 2025-06-24
[AINode] Modify dataset module for table-model training

ycycse:
As title. 
This pr  modelfiies dataset module to implement a new way to execute training sql for table model.


CRZbulabula (approved):
LGTM

PR #15844, opened 2025-06-29
RatisConsensus: extend election timeout threshold

RkGrit:
RatisConsensus: modify election timeout threshold from 10s to 30s

### DIFF
--- a/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
+++ b/iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/RatisConsensus.java
@@ -722,7 +722,7 @@ class RatisConsensus implements IConsensus {
       return client
           .getRaftClient()
           .admin()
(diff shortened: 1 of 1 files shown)


iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/config/RatisConfig.java
```diff
@@ -1130,6 +1139,10 @@ public void setSleepDeviationThresholdMs(int sleepDeviationThresholdMs) {
       public void setCloseThresholdMs(int closeThresholdMs) {
         this.closeThresholdMs = closeThresholdMs;
       }
+
+      public void setTransferLeaderTimeoutMs(int transferLeaderTimeoutMs) {
+        this.transferLeaderTimeoutMs = transferLeaderTimeoutMs;
+      }
```
jt2594838: Return `this` if you wish to use it in chained calls.

PR #15854, opened 2025-07-02
make with-grant-option optinal

zhujt20:
use parameter to make with-grant-option optinal


integration-test/src/test/java/org/apache/iotdb/db/it/auth/IoTDBGrantOptionIT.java
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+package org.apache.iotdb.db.it.auth;
+
+import org.apache.iotdb.it.env.EnvFactory;
+import org.apache.iotdb.it.framework.IoTDBTestRunner;
+import org.apache.iotdb.itbase.category.ClusterIT;
+import org.apache.iotdb.itbase.category.LocalStandaloneIT;
+
+import org.junit.*;
```
CRZbulabula: DO NOT import *

PR #15867, opened 2025-07-03
Pipe: Enable support for transferring TsFiles from object storage in the Pipe framework   

XNX02:


Caideyipi:
Can add a configuration to send "remote file" or "path"

PR #15868, opened 2025-07-03
[AINode] Fix built-in model inference & support user parameters

yunbow30944:
## Description

Fix built-in model inference & support user parameters

+ Fix bug where built-in models could not be invoked via call inference(...).

+ Support user-specified parameters (e.g. k=5, order="(1,1,0)") in inference calls.

+ Improve attribute parsing and error logging.

### DIFF
--- a/iotdb-core/ainode/ainode/core/manager/inference_manager.py
+++ b/iotdb-core/ainode/ainode/core/manager/inference_manager.py
@@ -84,7 +84,7 @@ class SundialStrategy(InferenceStrategy):
 
 
 class BuiltInStrategy(InferenceStrategy):
-    def infer(self, full_data, **_):
+    def infer(self, full_data):
         data = pd.DataFrame(full_data[1]).T
         output = self.model.inference(data)
         df = pd.DataFrame(output)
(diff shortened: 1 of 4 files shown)


iotdb-core/ainode/ainode/core/manager/inference_manager.py
```diff
@@ -156,11 +155,11 @@ def _run(
 
             # load model
             accel = str(inference_attrs.get("acceleration", "")).lower() == "true"
-            model = self.model_manager.load_model(model_id, accel)
+            model = self.model_manager.load_model(model_id, inference_attrs, accel)
```
CRZbulabula, line 158: Seems the parameter `acceleration` is useless currently, plz remove it in ur PR

CRZbulabula (changes requested):
PTAL

iotdb-core/ainode/ainode/core/manager/inference_manager.py
yunbow30944: accel is used in function `load_model` (model_storage.py, line265) to compile the user_defined model.

PR #15994, opened 2025-07-22
Support SSL in cluster internal connections

HTHou:
## Description

 In this PR, IoTDB will support mutual TLS in cluster Internal connections.


jt2594838 (approved):

integration-test/src/test/java/org/apache/iotdb/relational/it/db/it/IoTDBInternalSSLIT.java
jt2594838, line 1: Is it possible that you simply override an existing test(s) and only modify the encryption-related configs?
In that case, it is more conveniet to supplement more tests and new functions are less likely to be ignored.

iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/NettyTNonblockingTransport.java
```diff
@@ -0,0 +1,621 @@
...
+                                  "SSL handshake completed successfully for {}:{}", host, port);
+                            }
+                          } else {
+                            if (logger.isDebugEnabled()) {
+                              logger.debug(
+                                  "SSL handshake failed for {}:{}: {}",
+                                  host,
+                                  port,
+                                  future.cause().getMessage());
+                            }
```
jt2594838, line 175: Maybe we can use warn for failures?

iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/NettyTNonblockingTransport.java
```diff
@@ -0,0 +1,621 @@
...
+  }
+
+  private SslContext createSslContext() throws Exception {
+    SslContextBuilder sslContextBuilder = SslContextBuilder.forClient();
+
+    if (keystorePath != null && keystorePassword != null) {
+      KeyStore keyStore = KeyStore.getInstance("JKS");
+      try (FileInputStream fis = new FileInputStream(keystorePath)) {
+        keyStore.load(fis, keystorePassword.toCharArray());
+      }
```
jt2594838, line 190: Is KeyStore/KeyManagerFactory thread-safe? (I suppose we use it in a read-only manner) If so, we may cache an instance in the config.

iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/NettyTNonblockingTransport.java
```diff
@@ -0,0 +1,621 @@
...
+  private final int connectTimeoutMs;
+  private final long sslHandshakeTimeoutMs;
+  private final EventLoopGroup eventLoopGroup;
+  private final Bootstrap bootstrap;
+  private volatile Channel channel;
+  private final AtomicBoolean connected = new AtomicBoolean(false);
+  private final AtomicBoolean connecting = new AtomicBoolean(false);
+  private ChannelFuture future;
+  private final LinkedBlockingQueue<ByteBuf> readQueue = new LinkedBlockingQueue<>();
+  private final Object lock = new Object();
```
jt2594838: Nit: how about replacing it with `this`.

iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/NettyTNonblockingTransport.java
```diff
@@ -0,0 +1,621 @@
...
+          logger.debug("No data available for ByteBuffer read");
+        }
+        return 0;
+      }
+
+      int available = Math.min(buffer.remaining(), byteBuf.readableBytes());
+      if (available > 0) {
+        byte[] tempArray = new byte[available];
+        byteBuf.readBytes(tempArray);
+        buffer.put(tempArray);
```
jt2594838: May consider using ByteBuf.readBytes(ByteBuffer) to avoid memory copy.
Since the method does not allow providing a length, it is a little tricky to manipulate their positions
image

iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/NettyTNonblockingTransport.java
```diff
@@ -0,0 +1,621 @@
...
+    }
+
+    if (logger.isDebugEnabled()) {
+      logger.debug("Writing {} bytes from ByteBuffer", remaining);
+    }
+
+    synchronized (lock) {
+      ByteBuf byteBuf = Unpooled.buffer(remaining);
+      try {
+        byteBuf.writeBytes(buffer);
```
jt2594838: May use `Unpooled.wrappedBuffer()`.
May need re-positioning the input buffer.

iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/NettyTNonblockingTransport.java
```diff
@@ -0,0 +1,621 @@
...
+          logger.debug("Failed to drain dummy channel", e);
+        }
+        if (channel == null || !channel.isActive()) {
+          throw new TTransportException(TTransportException.END_OF_FILE, "Connection reset");
+        }
+      }
+      // Trigger OP_READ on dummy by writing dummy byte
+      ByteBuffer dummyByte = ByteBuffer.wrap(new byte[1]);
+      dummyServerAccepted.write(dummyByte);
+      // Wakeup selector if needed
```
jt2594838, line 281: Any need to flush?

iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/NettyTNonblockingTransport.java
```diff
@@ -0,0 +1,621 @@
...
+    } catch (Throwable e) {
+      throw new IOException(e);
+    }
+    if (connecting.get()) {
+      try {
+        TimeUnit.MILLISECONDS.sleep(1);
+      } catch (InterruptedException e) {
+        throw new IOException(e);
+      }
+    }
```
jt2594838: What are these for?

iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/NettyTNonblockingTransport.java
```diff
@@ -0,0 +1,621 @@
...
+    synchronized (lock) {
+      return channel != null ? channel.remoteAddress() : null;
+    }
+  }
+
+  public SocketAddress getLocalAddress() {
+    synchronized (lock) {
+      return channel != null ? channel.localAddress() : null;
+    }
+  }
```
jt2594838: Necessary to synchronize?

iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/NettyTNonblockingTransport.java
```diff
@@ -0,0 +1,621 @@
...
+      }
+      synchronized (lock) {
+        if (dummyServerAccepted != null) {
+          dummyServerAccepted.close();
+          dummyServerAccepted = null;
+        }
+        if (channel != null) {
+          channel.close();
+          channel = null;
+        }
```
jt2594838, line 595: How about dummyServer and dummyClient. 
Or `NettyTNonblockingTransport.this.close`?

iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/NettyTNonblockingTransport.java
jt2594838: Or maybe we can move the file content into a temporary file, now that someone has mentioned the possibility of it being modified during runtime.

iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/NettyTNonblockingTransport.java
HTHou: SSL handshake failed may happen when restart some nodes. For normal TNonblockingTransport, all failure logs are all debug level. Therefore we use debug here.

iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/NettyTNonblockingTransport.java
HTHou: Actually this part is similar to the code in https://github.com/apache/thrift/blob/f6fa1794539e68ac294038ac388d6bde40a6c237/lib/java/src/org/apache/thrift/transport/TSSLTransportFactory.java#L182

We cannot pass the KeyManagerFactory to TSSLTransportFactory directly. Therefore I use the same way to create  KeyManagerFactory as TSSLTransportFactory.java.

iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/NettyTNonblockingTransport.java
HTHou: future is `bootstrap.connect(host, port);`
However, when bootstrap.connect(host, port) executed, the `connected` may not be `true` because it be set at Line 465, which is also after  bootstrap.connect(host, port) executed.

sleep(1) to ensure `connected` be set.

iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/NettyTNonblockingTransport.java
HTHou: The toString() method used only in debug. It is ok to remove the lock I think.

iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/NettyTNonblockingTransport.java
HTHou: I didn't see any issue in my tests. I think it is fine without flush.

iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/NettyTNonblockingTransport.java
HTHou: Fixed

integration-test/src/test/java/org/apache/iotdb/relational/it/db/it/IoTDBInternalSSLIT.java
jt2594838: How about putting the test methods into an abstract class?
IoTDBInsertQueryIT and IoTDBInsertQueryWIthSslIT both extend it, and only implement different setUpClass and tearDownClass.

iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/NettyTNonblockingTransport.java
HTHou: Fixed

integration-test/src/test/java/org/apache/iotdb/relational/it/db/it/IoTDBInternalSSLIT.java
HTHou: I find another way to do that.

iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/NettyTNonblockingTransport.java
HTHou: I find another way to change that part.

iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/NettyTNonblockingTransport.java
HTHou: `NettyTNonblockingTransport.close` will be called in `TAsyncClient.onError` method. So I think it is OK to keep the current implement?

iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/NettyTNonblockingTransport.java
HTHou: Fixed

jt2594838 (approved):

PR #15998, opened 2025-07-22
Pipe: Handle existing hardlink in createHardLink method

SteveYurongSu:
Adds logic to check if a hardlink already exists when creating a hardlink. If the existing file matches the source (by MD5), a warning is logged and creation is skipped. If not, the existing file is deleted and creation is retried, with error handling and logging for failures.


Caideyipi (approved):
LGTM

PR #16070, opened 2025-07-31
[AINode] Add callback and release resources

yunbow30944:
## Description

+ shutdown the inference_manager when stop is called
+ optimize the log

### DIFF
--- a/iotdb-core/ainode/ainode/core/ainode.py
+++ b/iotdb-core/ainode/ainode/core/ainode.py
@@ -159,6 +159,7 @@ class AINode:
     def stop(self):
         if not self._stop_event.is_set():
             self._stop_event.set()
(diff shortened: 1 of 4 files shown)


iotdb-core/ainode/ainode/core/ainode.py
```diff
@@ -159,6 +159,7 @@ def _handle_signal(self, signum, frame):
     def stop(self):
         if not self._stop_event.is_set():
             self._stop_event.set()
+            self._rpc_handler._inference_manager.shutdown()
```
CRZbulabula: We'd better shutdown the `rpc_handler` rather than specific service belongs to the `rpc_handler`

CRZbulabula (changes requested):
PTAL

CRZbulabula (approved):
LGTM

PR #16103, opened 2025-08-05
Implement last cache optimize for query like select last(s1), last(s1, time), last(s2), last(s2, time)

Wei-hao-Li:
Last Cache hit condition: 
1. Select items are all made up of `last` or `last_by`;
2. For `last_by`, the first argument category type can be `tag`, `attribute` or `time`, the second can only be `measurement`.

Example:
Schema of table1
image
Hit cases:
`select last(s1),last(device),last(time) from table1`;
`select last(s1),last_by(time,s1),last_by(device,s1) from table1`
Miss cases:
`select last(s1),last_by(s1,time) from table1`;
`select last(s1),last_by(s2,s1) from table1`

**Performance Test**
Data generation：disable out_of_order write，1000 device * 50 sensor，client_number=10，loop=100，batch_size=100，disable all compaction
image

### DIFF
--- a/integration-test/src/test/java/org/apache/iotdb/relational/it/db/it/IoTDBMultiTAGsWithAttributesTableIT.java
+++ b/integration-test/src/test/java/org/apache/iotdb/relational/it/db/it/IoTDBMultiTAGsWithAttributesTableIT.java
@@ -27,7 +27,6 @@ import org.apache.iotdb.rpc.TSStatusCode;
 
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
-import org.junit.ComparisonFailure;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
@@ -2546,8 +2545,7 @@ public class IoTDBMultiTAGsWithAttributesTableIT {
     sql =
         "select level, attr1, device, attr2, last(time),last(device),last(level),last(attr1),last(attr2),"
             + "last_by(time,num),last_by(time,bignum),last_by(time,floatnum) from table0 where time>1971-04-26T17:46:40.000 group by attr1, device, attr2, level order by device,level,attr1,attr2";
-   repeatTest(sql, expectedHeader, retArray, DATABASE_NAME, 3);
-
+    repeatTest(sql, expectedHeader, retArray, DATABASE_NAME, 3);
 
     expectedHeader = new String[] {"_col0", "_col1"};
     retArray = new String[] {"null,null,"};


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/source/relational/aggregation/LastAccumulator.java
```diff
@@ -43,7 +43,7 @@ public class LastAccumulator implements TableAccumulator {
   protected TsPrimitiveType lastValue;
   protected long maxTime = Long.MIN_VALUE;
   protected boolean initResult = false;
-  protected boolean isTimeColumn = false;
+  protected final boolean isTimeColumn;
```
JackieTien97: It seems that it will only be used in sub class? Maybe we should move it into subClass

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/source/relational/aggregation/LastByAccumulator.java
```diff
@@ -63,14 +63,6 @@ public LastByAccumulator(
...
   }
 
-  public boolean xIsTimeColumn() {
-    return xIsTimeColumn;
-  }
-
-  public boolean yIsTimeColumn() {
-    return this.yIsTimeColumn;
-  }
-
```
JackieTien97, line 73: It seems that these two will only be used in sub class? Maybe we should move them into subClass

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/TableOperatorGenerator.java
```diff
@@ -3024,63 +3040,137 @@ private LastQueryAggTableScanOperator constructLastQueryAggTableScanOperator(
...
+          hitCachesIndexes.add(i);
+          lastRowCacheResults.add(lastByResult.get());
+          cachedDeviceEntries.add(node.getDeviceEntries().get(i));
         }
+      }
+    } else {
+      // LAST_VALUES optimize
+      lastValuesCacheResults = new ArrayList<>();
+      int measurementSize = parameter.getMeasurementColumnNames().size();
+      boolean needTime =
```
JackieTien97, line 3114: add some comments about this `needTime` field, when we need to set this field to `true`?

Actually, I'm a little bit confused by `parameter.getTimeColumnName() == null` since in my first impression, if it's null should mean that we don't need time column.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/metadata/fetcher/cache/TableDeviceSchemaCache.java
```diff
@@ -314,8 +314,7 @@ public TimeValuePair getLastEntry(
   }
 
   /**
-   * Get the last {@link TimeValuePair}s of given measurements, the measurements shall never be
-   * "time".
+   * Get the last {@link TimeValuePair}s of given measurements, the measurements shall never be "".
```
JackieTien97: actually, you pass "" empty string into this method.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/TableOperatorGenerator.java
```diff
@@ -3024,63 +3040,137 @@ private LastQueryAggTableScanOperator constructLastQueryAggTableScanOperator(
...
+                    node.getQualifiedObjectName().getDatabaseName(),
+                    node.getDeviceEntries().get(i).getDeviceID(),
+                    targetColumns);
+        boolean allHitCache = true;
+        if (lastResult != null) {
+          for (TimeValuePair timeValuePair : lastResult) {
+            if (timeValuePair == null
+                || timeValuePair.getValue() == null
+                || (updateTimeFilter != null
+                    && !LastQueryUtil.satisfyFilter(updateTimeFilter, timeValuePair))) {
```
JackieTien97: // the process logic should be same with tree model which examine if
              // `isFilterGtOrGe(seriesScanOptions.getGlobalTimeFilter())`, set
              // `lastByResult.get().getRight()[j] = EMPTY_PRIMITIVE_TYPE`,
              // but it should skip in table model

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/source/relational/LastQueryAggTableScanOperator.java
```diff
@@ -285,11 +295,202 @@ private void buildResultUseLastCache() {
     currentHitCacheIndex++;
   }
 
-  private void updateLastCacheIfPossible() {
-    if (!needUpdateCache) {
-      return;
+  private void buildResultUseLastValuesCache() {
+    appendGroupKeysToResult(cachedDeviceEntries, currentHitCacheIndex);
+    TimeValuePair[] currentHitResult = lastValuesCacheResults.get(currentHitCacheIndex);
+    // there is no problem when the cache result doesn't contain time column, because we will not
```
JackieTien97:
```suggestion
    // there is no problem when the cache result doesn't contain time column, because if in such case we will not
```

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/source/relational/LastQueryAggTableScanOperator.java
```diff
@@ -285,11 +295,202 @@ private void buildResultUseLastCache() {
...
+        case TAG:
+          String id =
+              getNthIdColumnValue(
+                  cachedDeviceEntries.get(currentHitCacheIndex), aggColumnsIndexArray[columnIdx]);
+          if (aggregator.getAccumulator() instanceof LastDescAccumulator) {
+            if (id == null) {
+              if (aggregator.getStep().isOutputPartial()) {
+                columnBuilder.writeBinary(
+                    new Binary(
+                        serializeTimeValue(getTSDataType(schema.getType()), lastRowTime, null)));
```
JackieTien97: won't NPE if you pass null?

JackieTien97: always append null is enough?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/source/relational/LastQueryAggTableScanOperator.java
```diff
@@ -285,11 +295,202 @@ private void buildResultUseLastCache() {
...
+            int measurementIdx = aggColumnsIndexArray[aggregatorInputChannels.get(channel)];
+            long lastTime =
+                lastValuesCacheResults.get(currentHitCacheIndex)[measurementIdx].getTimestamp();
+
+            // last_by
+            if (id == null) {
+              if (aggregator.getStep().isOutputPartial()) {
+                columnBuilder.writeBinary(
+                    new Binary(
+                        serializeTimeValue(getTSDataType(schema.getType()), lastTime, true, null)));
```
JackieTien97, line 353: same as `TAG`

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/source/relational/LastQueryAggTableScanOperator.java
```diff
@@ -285,11 +295,202 @@ private void buildResultUseLastCache() {
...
+                          getTSDataType(schema.getType()),
+                          lastTime,
+                          false,
+                          new TsPrimitiveType.TsLong(lastTime))));
+            } else {
+              columnBuilder.writeTsPrimitiveType(new TsPrimitiveType.TsLong(lastTime));
+            }
+          }
+          break;
+        case FIELD:
```
JackieTien97, line 471: there is no difference between `LastDescAccumulator` and `LastBy` for FIELD?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/source/relational/LastQueryAggTableScanOperator.java
```diff
@@ -365,12 +566,84 @@ private void updateLastCacheIfPossible() {
...
+                  new TimeValuePair(
+                      lastAccumulator.getMaxTime(),
+                      new TsPrimitiveType.TsLong(lastAccumulator.getMaxTime())));
+            } else {
+              TABLE_DEVICE_SCHEMA_CACHE.updateLastCacheIfExists(
+                  dbName,
+                  currentDeviceEntry.getDeviceID(),
+                  new String[] {""},
+                  new TimeValuePair[] {EMPTY_TIME_VALUE_PAIR});
+            }
```
JackieTien97: Should last row be same as here?

JackieTien97 (changes requested):
Add some perf experiment results and current last cache hit condition in pr descriptions

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/source/relational/LastQueryAggTableScanOperator.java
JackieTien97: if (tsPrimitiveType == EMPTY_PRIMITIVE_TYPE) always should append null

PR #16110, opened 2025-08-06
Skip TTL check when there is no ttl in the data region and no mods file

shuwenwei:
## Description
Skip TTL check when there is no ttl in the data region and no mods file


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/schema/ttl/TTLCache.java
```diff
@@ -199,6 +199,48 @@ public long getDatabaseMaxTTL(String database) {
     return maxTTL;
   }
 
+  public boolean dataInDatabaseMayHaveTTL(String database) {
+    String[] nodeNames = database.split("\\.");
```
jt2594838: Cannot simply split because the database segment may contain '.'.

jt2594838 (approved):

PR #16131, opened 2025-08-08
[AINode] Decoupling inference manager into request_manager, pool_manager

yunbow30944:
## Description

+ Adding pool state and support state transfer
+ Decoupling inference manager into request_manager, pool_manager, pool_scheduler
+ Adding PoolGroup structure

### DIFF
--- a/iotdb-core/ainode/ainode/core/inference/pool_manager.py
+++ b/iotdb-core/ainode/ainode/core/inference/pool_manager.py
@@ -1,0 +1,160 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+from collections import defaultdict
+import threading
+from ainode.core.inference.inference_request import InferenceRequest
+from ainode.core.log import Logger
+from ainode.core.exception import (
+    InferenceModelInternalError,
+)
+from typing import Dict, Optional
+import torch.multiprocessing as mp
+from ainode.core.inference.inference_request_pool import InferenceRequestPool, PoolState
+import torch
+from ainode.core.inference.inference_request_pool_group import PoolGroup
+
+logger = Logger()
+
+
+class PoolManager:
+    """
+    A manager for handling inference request pools.
+    It handles the registration of pools, adding and removing requests,
+    and managing the state of each pool.
+    """
+
+    DEFAULT_DEVICE = torch.device("cpu")
+    # DEFAULT_DEVICE = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+
+    def __init__(self):
+        # structure: {model_id: {pool_id: set(req_id)}}
+        self.pool_to_reqs: Dict[str, Dict[int, set[str]]] = defaultdict(
+            lambda: defaultdict(set)
+        )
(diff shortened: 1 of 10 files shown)


iotdb-core/ainode/ainode/core/inference/request_manager.py
```diff
@@ -0,0 +1,87 @@
...
+                )
+
+    def process_request(self, req):
+        infer_proxy = InferenceRequestProxy(req.req_id)
+        with self._result_wrapper_lock:
+            self._result_wrapper_map[req.req_id] = infer_proxy
+        # lazy initialization for first request
+        model_id = req.model_id
+        if model_id not in self._pool_manager.get_request_pool_map():
+            self._pool_scheduler._first_req_init(model_id)
```
CRZbulabula: Get the whole pool map for judging the existence is not a good implementation. I suggest to add a new interface in the `pool_manager`, which checks whether the specified pools exist. Subsequently, every requests will enter the synchronized function and wait until the first pool is initialized; otherwise, every requests can be dispatched immediately.

iotdb-core/ainode/ainode/core/inference/pool_manager.py
```diff
@@ -0,0 +1,160 @@
...
+            f"[Inference][Device-{self.DEFAULT_DEVICE}][Pool-{pool_idx}][ID-{req.req_id}] Request is queued for inference"
+        )
+
+    def _get_optimal_pool(self, model_id) -> int:
+        loads = []
+        for pool_idx in self.get_pool_ids(model_id):
+            load_count = self.get_load(model_id, pool_idx)
+            loads.append((pool_idx, load_count))
+        min_idx = min(loads, key=lambda x: x[1])[0]
+        return min_idx
```
CRZbulabula: In our current stage, I suggest we just employ intuitive schedule algorithms, eg. hash, random and round robin. Because probing the load of pools results in transimission between processes, or maintaining the `pool_to_reqs` and `req_to_pools`.

iotdb-core/ainode/ainode/core/inference/pool_manager.py
```diff
@@ -0,0 +1,160 @@
...
+            lambda: defaultdict(set)
+        )
+        # structure: {req_id: pool_id}
+        self.req_to_pool: Dict[str, int] = {}
+        # structure: {model_id: {pool_id: PoolState}}
+        self.pool_states: Dict[str, Dict[int, PoolState]] = defaultdict(dict)
+        # structure: {model_id: {pool_id: threading.RLock}}
+        self.pool_locks: Dict[str, Dict[int, threading.RLock]] = defaultdict(
+            lambda: defaultdict(threading.RLock)
+        )
```
CRZbulabula: Should further distinguish the read and write locks

iotdb-core/ainode/ainode/core/inference/request_manager.py
```diff
@@ -0,0 +1,87 @@
...
+)
+from ainode.core.log import Logger
+import torch.multiprocessing as mp
+from ainode.core.inference.pool_scheduler import PoolScheduler
+from ainode.core.inference.pool_manager import PoolManager
+
+logger = Logger()
+
+
+class RequestManager:
```
CRZbulabula: The term `manager` corresponds to highest abstract under the `RPCService`, use a different name to distinguish between the different levels.

CRZbulabula (changes requested):
PTAL

iotdb-core/ainode/ainode/core/inference/pool_scheduler.py
```diff
@@ -0,0 +1,123 @@
...
+    """
+
+    DEFAULT_DEVICE = torch.device("cpu")
+    # DEFAULT_DEVICE = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+
+    def __init__(self, pool_manager: PoolManager, result_queue: mp.Queue):
+        self._pool_manager = pool_manager
+        self._result_queue = result_queue
+
+    @synchronized(threading.Lock())
```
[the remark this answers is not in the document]
CRZbulabula: Reconsider this comment, fix it or state ur reason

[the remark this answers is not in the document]
yunbow30944: `@synchronized(threading.Lock())` creates a single lock object at class definition time, 
so it is actually shared across all instances rather than creating a new one per method call. 

At the moment there is only one pool_scheduler globally, so the global lock is sufficient.

PR #16132, opened 2025-08-08
Load: Fix LoadPieceNode reading aligned Chunk causing OOM

luoluoyuyu:
## Description
Fix OOM issue by lazy loading ChunkData

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/load/splitter/AlignedChunkData.java
```diff
@@ -291,26 +286,36 @@ public void writeDecodeValuePage(
...
     if (needDecodeChunk) {
-      buildChunkWriter(stream);
+      writeChunkToWriter(stream, writer);
     } else {
-      deserializeEntireChunk(stream);
+      writeEntireChunkToWriter(stream, writer);
+    }
+  }
+
+  protected void deserializeTsFileDataByte(final InputStream stream) throws IOException {
```
shuwenwei, line 298: Maybe we can use the existing buffer instead of the current input stream, which can avoid a memory copy

Caideyipi (approved):

PR #16163, opened 2025-08-13
Use global sink executor pool for IoTV2

Pengzna:
As IoTV2's sink num is equals to data regions num, and data regions num is limited and strictly controlled, there is no need for IoTV2 to use separate executor pool for each sink, which may cause threads booming.

So we use global singleton executor pool for all of IoTV2 sinks. 

Use fixedThreadPool here, and its core size is carefully set to Min(cpu cores num, data region soft limit).

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/consensus/PipeConsensusSubtaskExecutor.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/consensus/PipeConsensusSubtaskExecutor.java
@@ -26,28 +26,34 @@ import org.apache.iotdb.db.storageengine.dataregion.DataRegion;
 public class PipeConsensusSubtaskExecutor extends PipeSinkSubtaskExecutor {
   public PipeConsensusSubtaskExecutor() {
     super(
-        // The number of data regions for a datanode is limited by offHeapMemory. At the same time,
-        // in order to ensure multicore performance, the number of data regions usually does not
-        // exceed the number of cores. To prevent the thread from exploding, we take the min of
-        // both.
-        (int)
-            Math.min(
-                // NOTE1: The number of data regions is limited by the number of cores.
-                Runtime.getRuntime().availableProcessors(),
-                // NOTE2: The number of data regions is also limited by the offHeapMemory. In fact,
-                // schema regions also take up off-heap memory. So in fact, the soft cap of
-                // data region will be smaller than what is calculated here, but that's okay, we can
-                // set the core size quota of the Executor pool a little bit higher slightly
-                Math.max(
-                        IoTDBDescriptor.getInstance()
-                            .getMemoryConfig()
-                            .getDirectBufferMemoryManager()
-                            .getTotalMemorySizeInBytes(),
-                        IoTDBDescriptor.getInstance()
-                            .getMemoryConfig()
-                            .getOffHeapMemoryManager()
-                            .getTotalMemorySizeInBytes())
-                    / DataRegion.getAcquireDirectBufferMemCost()),
+        DataRegion.getAcquireDirectBufferMemCost() == 0
+            ? Runtime.getRuntime().availableProcessors()
+            :
+            // The number of data regions for a datanode is limited by offHeapMemory. At the same
+            // time,
+            // in order to ensure multicore performance, the number of data regions usually does not
+            // exceed the number of cores. To prevent the thread from exploding, we take the min of
+            // both.
+            (int)
+                Math.min(
+                    // NOTE1: The number of data regions is limited by the number of cores.
+                    Runtime.getRuntime().availableProcessors(),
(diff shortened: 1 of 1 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/stage/PipeTaskSinkStage.java
```diff
@@ -59,7 +61,11 @@ protected void registerSubtask() {
     this.connectorSubtaskId =
         PipeSinkSubtaskManager.instance()
             .register(
-                executor,
+                // IoTV2 uses global singleton executor pool.
+                pipeName.startsWith(PipeStaticMeta.CONSENSUS_PIPE_PREFIX)
+                    ? (PipeSinkSubtaskExecutor)
+                        IoTV2GlobalComponentContainer.getInstance().getConsensusExecutor()
+                    : executor.get(),
```
Caideyipi: Do not get here... A normal sink executor has an increasing ID, thus you shall only get when you really need it

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/builder/PipeDataNodeTaskBuilder.java
```diff
@@ -118,9 +118,7 @@ public PipeDataNodeTask build() {
               pipeStaticMeta.getCreationTime(),
               connectorParameters,
               regionId,
-              pipeType.equals(PipeType.USER)
-                  ? PipeSubtaskExecutorManager.getInstance().getConnectorExecutorSupplier()
-                  : PipeSubtaskExecutorManager.getInstance().getConsensusExecutorSupplier());
+              PipeSubtaskExecutorManager.getInstance().getConnectorExecutorSupplier());
```
Caideyipi: Why not write your codes here?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/consensus/PipeConsensusSubtaskExecutor.java
```diff
@@ -19,18 +19,25 @@
...
+        // exceed the number of cores. To prevent the thread from exploding, we take the min of
+        // both.
+        (int)
+            Math.min(
+                Runtime.getRuntime().availableProcessors(),
+                IoTDBDescriptor.getInstance()
+                        .getMemoryConfig()
+                        .getOffHeapMemoryManager()
+                        .getTotalMemorySizeInBytes()
+                    / DataRegion.getAcquireDirectBufferMemCost()),
```
Caideyipi: May this be directly configurable?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/stage/PipeTaskSinkStage.java
Pengzna: fixed.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/builder/PipeDataNodeTaskBuilder.java
Pengzna: fixed

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/consensus/PipeConsensusSubtaskExecutor.java
Pengzna: I think there is no need to configure this.
1. users may not aware of this configuration and I think most users have no idea for configuring this.
2. the soft limit of data region can be approximated, that we just set core size to this soft limit is fine.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/consensus/PipeConsensusSubtaskExecutor.java
```diff
@@ -27,12 +27,17 @@ public class PipeConsensusSubtaskExecutor extends PipeSinkSubtaskExecutor {
   public PipeConsensusSubtaskExecutor() {
     super(
         // The number of data regions for a datanode is limited by offHeapMemory. At the same time,
-        // in order to ensure multi-core performance, the number of data regions usually does not
+        // in order to ensure multi-cores performance, the number of data regions usually does not
```
jt2594838: I suppose `multi-core` is correct.
You call something a multi-player game instead of multi-players game, right?

Pengzna: you are right!

PR #16171, opened 2025-08-13
IoTV2: Fix file corrupt due to RandomAccessFile is lost.

Pengzna:
The problem can be abstracted as:

1. Write file A with RandomFileWriterA, write 0~x bytes.
2. Re-new RandomFileWriterB on file 
3. Use RandomFileWriterB to write file A and write x~end bytes
4. It is expected to directly override file A into x~end bytes

So every time we new RandomAccessFile, we should point its pointer to the end of the file, keeping the append semantics.
In addition, when the follower fails the precheck due to read-only or other reasons, the corresponding RandomAccessFile needs to be recycled.


jt2594838 (approved):

Pengzna:
DO NOT merge this pr currently... I'd like to check sth else

jt2594838 (approved):

PR #16176, opened 2025-08-14
encrypt key generate and destroy

zhujt20:
encrypt key generate and destroy


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/conf/IoTDBDescriptor.java
```diff
@@ -1821,6 +1821,9 @@ private void loadTsFileProps(TrimProperties properties) throws IOException {
     TSFileDescriptor.getInstance()
         .getConfig()
         .setEncryptType(properties.getProperty("encrypt_type", "UNENCRYPTED"));
+    //    TSFileDescriptor.getInstance()
+    //        .getConfig()
+    //        .setEncryptKeyFromToken(System.getenv("user_encrypt_token"));
```
jt2594838: Remove it

PR #16229, opened 2025-08-21
Pipe: fix MQTT Source Session logic & Enhanced error reporting for MQTT port and data path conflicts

XNX02:


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/source/mqtt/MQTTSource.java
```diff
@@ -86,6 +95,45 @@ public void validate(final PipeParameterValidator validator) throws Exception {
             PipeSourceConstant.EXTERNAL_EXTRACTOR_SINGLE_INSTANCE_PER_NODE_DEFAULT_VALUE)) {
       throw new PipeParameterNotValidException("single mode should be true in MQTT extractor");
     }
+
```
XNX02, line 102: Could we just pre-check port using socket.bind and pre-check data path using file.canWrite()?

Caideyipi: Of course, we can judge the data path and socket instead of looking for the other MQTT service, because the two may be occupied for other reasons

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/source/mqtt/MQTTSource.java
```diff
@@ -86,6 +100,72 @@ public void validate(final PipeParameterValidator validator) throws Exception {
...
+    final String formatType =
+        validator
+            .getParameters()
+            .getStringOrDefault(
+                PipeSourceConstant.MQTT_PAYLOAD_FORMATTER_KEY,
+                SystemConstant.SQL_DIALECT_TREE_VALUE.equals(sqlDialect)
+                    ? PipeSourceConstant.MQTT_PAYLOAD_FORMATTER_TREE_DIALECT_VALUE
+                    : PipeSourceConstant.MQTT_PAYLOAD_FORMATTER_TABLE_DIALECT_VALUE);
+    payloadFormat = PayloadFormatManager.getPayloadFormat(formatType);
+    if (!sqlDialect.equals(payloadFormat.getType())) {
```
Caideyipi: Shall not restrict this

PR #16196, opened 2025-08-18
Region operation "extend" and "remove" support multi regions in one SQL

liyuheng55555:
```
IoTDB> extend region 1,2,3 to 3
Msg: org.apache.iotdb.jdbc.IoTDBSQLException: 909: Total regions: 3, successfully submitted: 2, failed to submit: 1
region 1: Successfully submitted
region 2: Successfully submitted
region 3: Target DataNode 3 already contains region TConsensusGroupId(type:DataRegion, id:3)
```

PR #16234, opened 2025-08-22
[to dev/1.3] Region operation "extend" and "remove" support multi regions in one SQL

liyuheng55555:
cp #16196


CRZbulabula (approved):
LGTM!

PR #16250, opened 2025-08-25
Pipe: split full-sync pipe into history and realtime pipes

VGalaxies:
This PR implements syntactic sugar for pipe creation by automatically splitting full-sync pipes into separate history and realtime pipes. When a pipe is configured for full synchronization (both history and realtime enabled, excluding snapshot mode), the system now creates two separate pipes with "_history" and "_realtime" suffixes instead of a single full-sync pipe.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/execution/config/executor/ClusterConfigTaskExecutor.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/execution/config/executor/ClusterConfigTaskExecutor.java
@@ -2022,7 +2022,8 @@ public class ClusterConfigTaskExecutor implements IConfigTaskExecutor {
             new TCreatePipeReq()
                 // Append suffix to the pipeline name for historical data
                 .setPipeName(createPipeStatement.getPipeName() + "_history")
-                .setIfNotExistsCondition(createPipeStatement.hasIfNotExistsCondition())
+                // NOTE: set if not exists always to true to handle partial failure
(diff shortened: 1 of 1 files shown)


VGalaxies:
consider partial failure

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/execution/config/executor/ClusterConfigTaskExecutor.java
```diff
@@ -2007,6 +2009,77 @@ public SettableFuture<ConfigTaskResult> createPipe(
...
+    if (PipeConfig.getInstance().getPipeAutoSplitFullEnabled()
+        && PipeDataNodeAgent.task().isFullSync(extractorPipeParameters)) {
+      try (final ConfigNodeClient configNodeClient =
+          CONFIG_NODE_CLIENT_MANAGER.borrowClient(ConfigNodeInfo.CONFIG_REGION_ID)) {
+        // 1. Send request to create the historical data synchronization pipeline
+        final TCreatePipeReq historyReq =
+            new TCreatePipeReq()
+                // Append suffix to the pipeline name for historical data
+                .setPipeName(createPipeStatement.getPipeName() + "_history")
+                .setIfNotExistsCondition(createPipeStatement.hasIfNotExistsCondition())
```
Caideyipi: I still think we need to handle the situation when the history creation has succeeded but realtime has not. Such as, the "if not exists" shall be always true, because user may retry the sql when realtime failed, and it's bad to return "xx_history already exists" in this situation.

PR #16256, opened 2025-08-25
implement new session constructor

761417898:
as title


jt2594838 (commented):
Please add an example in

PR #16281, opened 2025-08-27
implement client-session node selector

hongzhi-gao:
Here are the changes
- Introduce a new EndpointSelectionStrategy interface with multiple implementations (e.g., random, round-robin).
- Update SessionConnection so that during reconnect, it selects a node based on the configured EndpointSelectionStrategy.
- Add corresponding constructor parameters in Session/SessionBuilder and SessionPool/SessionPoolBuilder to support strategy configuration.

Unit tests have been added for `EndpointSelectionStrategy`.
However, the `reconnect` function was only manually verified, and no related test code has been included.


iotdb-client/isession/src/main/java/org/apache/iotdb/isession/SessionConfig.java
```diff
@@ -57,4 +60,26 @@ public class SessionConfig {
...
+   * @return configured strategy instance
+   * @throws IllegalArgumentException if strategyName is null or empty
+   */
+  public static EndpointSelectionStrategy createSelectionStrategy(String strategyName) {
+    if (ENDPOINT_SELECTION_STRATEGY_SEQUENTIAL.equalsIgnoreCase(strategyName)) {
+      return new SequentialSelectionStrategy();
+    }
+    // Default to random strategy
+    return new RandomSelectionStrategy();
+  }
```
jt2594838: May print a warn log if the provided name is neither.

jt2594838 (approved):

iotdb-client/isession/src/main/java/org/apache/iotdb/isession/SessionConfig.java
hongzhi-gao: fixed

iotdb-client/session/src/main/java/org/apache/iotdb/session/SessionConnection.java
```diff
@@ -121,6 +121,8 @@ public class SessionConnection {
   // ms is 1_000, us is 1_000_000, ns is 1_000_000_000
   private int timeFactor = 1_000;
 
+  private final Set<TEndPoint> triedEndpoints = ConcurrentHashMap.newKeySet();
```
HTHou, line 124: This set seems be used only in `reconnect()` method, can this field be changed to a normal variable?

jt2594838 (approved):

PR #16295, opened 2025-08-28
[Bug Fix] Download config.json for Large TimeSeries Model

RkGrit:
As shown in the following figure, when performing inference tasks using a large time series model, the Cli reports an error. The reason is that only the model.safetensor file was downloaded, but not the config.json file.

json


CRZbulabula (approved):
LGTM!

PR #16306, opened 2025-08-29
Fix the issue where string filtering cannot be used in the DEFINE clause

Young-Leo:
Fix the issue where string & date & blob filtering cannot be used in the DEFINE clause

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/process/rowpattern/expression/Computation.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/process/rowpattern/expression/Computation.java
@@ -25,7 +25,9 @@ import org.apache.iotdb.db.queryengine.plan.relational.planner.rowpattern.Expres
 import org.apache.iotdb.db.queryengine.plan.relational.sql.ast.ArithmeticBinaryExpression;
 import org.apache.iotdb.db.queryengine.plan.relational.sql.ast.BinaryLiteral;
 import org.apache.iotdb.db.queryengine.plan.relational.sql.ast.BooleanLiteral;
+import org.apache.iotdb.db.queryengine.plan.relational.sql.ast.Cast;
 import org.apache.iotdb.db.queryengine.plan.relational.sql.ast.ComparisonExpression;
+import org.apache.iotdb.db.queryengine.plan.relational.sql.ast.DataType;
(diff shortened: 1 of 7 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/process/rowpattern/expression/PatternExpressionComputation.java
```diff
@@ -161,7 +162,7 @@ private Object getValueFromPartition(
       return partition.getFloat(channel, position);
     } else if (type instanceof DoubleType) {
       return partition.getDouble(channel, position);
-    } else if (type instanceof StringType) {
+    } else if (type instanceof StringType || type instanceof BinaryType) {
```
JackieTien97:
```suggestion
    } else if (type instanceof AbstractVarcharType) {
```

JackieTien97: And what about BlobType

JackieTien97 (changes requested):
Add IT about all the support types

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/process/rowpattern/expression/PatternExpressionComputation.java
Young-Leo: Added

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/process/rowpattern/expression/Computation.java
```diff
@@ -129,6 +131,20 @@ private static Computation parse(
...
+        GenericLiteral constExpr = (GenericLiteral) expression;
+        String type = constExpr.getType();
+
+        if ("DATE".equalsIgnoreCase(type)) {
+          String dateStr = constExpr.getValue();
+          int dateInt = Integer.parseInt(dateStr);
+          return new ConstantComputation(dateInt);
+        } else {
+          return new ConstantComputation(constExpr.getValue());
+        }
```
JackieTien97: when this branch will take effect?

PR #16324, opened 2025-09-01
Fix CN pipe procedures restore dead lock

Pengzna:
as title.


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/ProcedureExecutor.java
```diff
@@ -116,9 +120,41 @@ public void init(int numThreads) {
...
+        lockedPipeProcedures.stream()
+            .max(Comparator.comparingLong(AbstractOperatePipeProcedureV2::getLockSeqId))
+            .map(p -> (Procedure<Env>) p);
+
+    if (lockedPipeProcedures.size() > 1) {
+      LOG.warn(
+          "[Procedure restore]Detected multiple locked pipe procedures in procedure executor {}, only keep last one {}",
+          lockedPipeProcedures,
+          maxPipeProcedure.get());
+    }
```
jt2594838, line 148: Other lockedPipeProcedures should be cleaned up (removing their disk files).

jt2594838 (commented):
Try adding an IT like:
1. Set up a 3C1D cluster;
2. Shutdown the two non-leader confignodes;
3. Create two pipes;
4. Restart all confignodes.

PR #16351, opened 2025-09-05
Use non-blocking transport for ConfigNodeRPCProcessor threads

CRZbulabula:
The ConfigNodeRPCProcessor service would block for waiting to receive requests from DataNode, once when the connection is established. However, these threads would not be automatically recycled even there is no further requests. As a result, the number of threads in ConfigNode keep increase.

To recycle these idle threads in time, we use the non-blocking transport in this PR.


OneSizeFitsQuorum:
Seems uncorrelated for the code changes and the pr title/description

CRZbulabula:
> Seems uncorrelated for the code changes and the pr title/description

Thx 4 ur kind concern. This PR is still W.I.P. Since we haven't found the best implementation according to the ConfigNodeRPCProcessor, I mark this PR to draft now.

PR #16375, opened 2025-09-09
[AINode] Fix pom.xml of ainode

Alchuang22-dev:
## Description
We found that the `ainode/conf` is not moved to correct location during previous compile link update. 
Thus, we wove config files into target package of ainode, while fixing the bug that ainode cluster info were shown as "UNKNOWN" in this PR.


Alchuang22-dev:
Here is the execute screeenshot using this pull request.
截屏2025-09-11 下午4 36 50

PR #16432, opened 2025-09-18
Append user index field for User

wenyanshi-123:
A reconstruction PR.

### DIFF
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/schema/CNPhysicalPlanGenerator.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/schema/CNPhysicalPlanGenerator.java
@@ -103,14 +103,15 @@ public class CNPhysicalPlanGenerator
   private Exception latestException = null;
   private String userName;
 
-  public CNPhysicalPlanGenerator(final Path snapshotFilePath, final CNSnapshotFileType fileType)
+  public CNPhysicalPlanGenerator(
+      final Path snapshotFilePath, final CNSnapshotFileType fileType, final String userName)
       throws IOException {
     if (fileType == CNSnapshotFileType.SCHEMA) {
       logger.warn("schema_template need two files");
       return;
     }
     if (fileType == CNSnapshotFileType.USER_ROLE) {
-      userName = snapshotFilePath.getFileName().toString().split("_role.profile")[0];
+      this.userName = userName;
     }
     snapshotFileType = fileType;
     inputStream = Files.newInputStream(snapshotFilePath);
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/schema/CNPhysicalPlanGenerator.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/schema/CNPhysicalPlanGenerator.java
@@ -103,7 +103,6 @@ public class CNPhysicalPlanGenerator
   private Exception latestException = null;
   private String userName;
 
-
   public CNPhysicalPlanGenerator(
       final Path snapshotFilePath, final CNSnapshotFileType fileType, final String userName)
       throws IOException {
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/schema/CNPhysicalPlanGenerator.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/schema/CNPhysicalPlanGenerator.java
@@ -200,12 +200,12 @@ public class CNPhysicalPlanGenerator
       int tag = dataInputStream.readInt();
       boolean fromOldVersion = tag < 0;
       String user;
-      if (tag<0) {
+      if (tag < 0) {
         user = readString(dataInputStream, STRING_ENCODING, strBufferLocal, -1 * tag);
-      } else if (tag==1) {
+      } else if (tag == 1) {
         user = readString(dataInputStream, STRING_ENCODING, strBufferLocal);
-      }else{
-        dataInputStream.readLong();//userId
+      } else {
+        dataInputStream.readLong(); // userId
         user = readString(dataInputStream, STRING_ENCODING, strBufferLocal);
       }
 
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/schema/CNPhysicalPlanGenerator.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/schema/CNPhysicalPlanGenerator.java
@@ -200,9 +200,12 @@ public class CNPhysicalPlanGenerator
       int tag = dataInputStream.readInt();
       boolean fromOldVersion = tag < 0;
       String user;
-      if (fromOldVersion) {
+      if (tag<0) {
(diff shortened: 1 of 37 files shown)


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/schema/CNPhysicalPlanGenerator.java
```diff
@@ -200,9 +200,12 @@ private void generateUserRolePhysicalPlan(final boolean isUser) {
       int tag = dataInputStream.readInt();
       boolean fromOldVersion = tag < 0;
       String user;
-      if (fromOldVersion) {
+      if (tag < 0) {
         user = readString(dataInputStream, STRING_ENCODING, strBufferLocal, -1 * tag);
+      } else if (tag == 1) {
+        user = readString(dataInputStream, STRING_ENCODING, strBufferLocal);
       } else {
+        dataInputStream.readLong(); // userId
```
jt2594838: Why is not UserId processed in the branch where isUser = true?

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/auth/entity/User.java
```diff
@@ -56,6 +58,13 @@ public User(String name, String password) {
     this.roleSet = new HashSet<>();
   }
 
+  public User(String name, String password, long userId) {
+    super(name);
+    this.password = password;
+    this.userId = userId;
+    this.roleSet = new HashSet<>();
+  }
+
```
jt2594838: Why not remove the constructor where the userId is not provided?
Is any case legal that creates a user without an ID?

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/auth/entity/User.java
```diff
@@ -73,7 +82,15 @@ public void addRole(String roleName) {
     roleSet.add(roleName);
   }
 
+  public void setUserId(long userId) {
+    this.userId = userId;
+  }
```
jt2594838, line 91: Why provide this method? UserId should be immutable.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/auth/role/LocalFileRoleAccessor.java
```diff
@@ -223,6 +223,23 @@ public Role loadEntity(String entityName) throws IOException {
...
+    }
+    FileInputStream inputStream = new FileInputStream(userIdFile);
+    try (DataInputStream dataInputStream =
+        new DataInputStream(new BufferedInputStream(inputStream))) {
+      return dataInputStream.readLong();
+    } catch (Exception e) {
+      throw new IOException(e);
+    } finally {
+      strBufferLocal.remove();
+    }
```
jt2594838: How about just storing the number in the file name, which, in my opinion, would make the logic a little simpler.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/auth/role/LocalFileRoleAccessor.java
```diff
@@ -301,6 +318,7 @@ public static List<String> getEntityStrings(String[] names) {
       }
       retList.addAll(set);
     }
+    retList.remove("user_id");
```
jt2594838: image
I think it has already been filtered.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/auth/user/BasicUserManager.java
```diff
@@ -102,6 +106,20 @@ private void initAdmin() throws AuthException {
...
+  private void initUserId() {
+    try {
+      long maxUserId = this.accessor.loadUserId();
+      if (maxUserId == -1 || maxUserId < 10000) {
+        nextUserId = 10000;
+      } else {
+        nextUserId = maxUserId;
+      }
+    } catch (IOException e) {
+      LOGGER.warn("meet error in load max userId.");
```
jt2594838: Give the exception

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/auth/user/BasicUserManager.java
```diff
@@ -128,7 +146,15 @@ public boolean createUser(
     }
     lock.writeLock(username);
     try {
-      user = new User(username, enableEncrypt ? AuthUtils.encryptPassword(password) : password);
+      long userid = 0;
+      if (username.equals("root")) {
```
jt2594838: Use a constant.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/schema/CNPhysicalPlanGenerator.java
CRZbulabula: We've fixed this. Plz check it.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/auth/entity/User.java
CRZbulabula: It is illegal for the case without userId. To address ur concern, we've added a `TestOnly` annotation to this function.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/auth/entity/User.java
CRZbulabula: Cuz we encounter some compatitive issues. Specifically, when loading the old version of authority snapshot, where there are no userId records, we cannot set userId accordingly. To set the userId correctly, we have to wait until all authority entities are loaded completely. Thus we left this setter here.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/auth/role/LocalFileRoleAccessor.java
CRZbulabula: Unfortunately not >_<, cuz this new file also has the `.profile` suffix.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/auth/user/BasicUserManager.java
CRZbulabula:
```suggestion
      LOGGER.warn("meet error in load max userId.", e);
```

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/auth/role/BasicRoleManager.java
```diff
@@ -229,4 +236,20 @@ public List<String> listAllEntities() {
     rtlist.sort(null);
     return rtlist;
   }
+
+  public List<TListUserInfo> listAllEntitiesInfo() {
+
+    List<TListUserInfo> rtlist = new ArrayList<>();
+    for (Role r : entityMap.values()) {
+      // System.out.println(r.getRoleId());
```
CRZbulabula: Remove this.

CRZbulabula (changes requested):
PTAL.

CRZbulabula (approved):
LGTM

PR #16542, opened 2025-10-09
Use userId to record password history

jt2594838:


hongzhi-gao:
rename user test
image

PR #16544, opened 2025-10-09
No filtering audit DBs in some procdure & update idle time after logging in

jt2594838:


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/ConfigManager.java
```diff
@@ -2643,7 +2643,7 @@ private Map<TConsensusGroupId, TRegionReplicaSet> getRelatedSchemaRegionGroup(
   public Map<TConsensusGroupId, TRegionReplicaSet> getRelatedDataRegionGroup(
       final PathPatternTree patternTree) {
     return getRelatedDataRegionGroup(
-        getSchemaPartition(patternTree, false).getSchemaPartitionTable());
+        getSchemaPartition(patternTree, true).getSchemaPartitionTable());
```
JackieTien97: Normal user can also execute delete timeseries root.** or delete from root.** and won't be rejected by Auth Check. And these users expect to delete all timeseries under root.** which they have WRITE_DATA(except for root.__audit)

jt2594838: Fixed

PR #16551, opened 2025-10-10
encrypt audit log

zhujt20:
add encrypt param passing logic for encrypting audit log

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/performer/constant/CrossCompactionPerformer.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/performer/constant/CrossCompactionPerformer.java
@@ -19,6 +19,7 @@
 
 package org.apache.iotdb.db.storageengine.dataregion.compaction.execute.performer.constant;
 
+import org.apache.iotdb.commons.utils.TestOnly;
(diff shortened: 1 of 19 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/EncryptDBUtils.java
```diff
@@ -0,0 +1,72 @@
...
+  private static final Logger LOGGER = LoggerFactory.getLogger(EncryptDBUtils.class);
+
+  public static EncryptParameter getFirstEncryptParamFromDatabase(String database) {
+    if (database == null || database.isEmpty()) {
+      return new EncryptParameter(tsFileConfig.getEncryptType(), tsFileConfig.getEncryptKey());
+    }
+    for (Map.Entry<String, EncryptParameter> entry : conf.getTSFileDBToEncryptMap().entrySet()) {
+      if (database.startsWith(entry.getKey())) {
+        return entry.getValue();
+      }
```
jt2594838: Why use startWith?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/EncryptDBUtils.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+    return new EncryptParameter(tsFileConfig.getEncryptType(), tsFileConfig.getEncryptKey());
```
jt2594838: Cache the result?

PR #16560, opened 2025-10-11
Fix author type conversion

shuwenwei:
## Description
Fix author type conversion


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/statement/AuthorType.java
```diff
@@ -35,101 +42,14 @@ public enum AuthorType {
...
   LIST_ROLE_PRIVILEGE,
+
   ACCOUNT_UNLOCK,
+
   // Remind to renew the convert codes in ConfigNodeRPCServiceProcessor
+  // If the config node plan is involved, the type defined in following lines needs to manually add
+  // conversion logic
   RENAME_USER,
+  UPDATE_USER_MAX_SESSION,
+  UPDATE_USER_MIN_SESSION,
```
CRZbulabula, line 53: No need append them in AuthorRType?

PR #16563, opened 2025-10-12
Patch for AUDIT privilege authentication on table model

CRZbulabula:
Visit the __audit database only requires AUDIT privilege.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/security/ITableAuthCheckerImpl.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/security/ITableAuthCheckerImpl.java
@@ -75,7 +75,7 @@ public class ITableAuthCheckerImpl implements ITableAuthChecker {
               .setPrivilegeType(PrivilegeType.AUDIT)
               .setResult(hasAuditPrivilege),
           () -> databaseName);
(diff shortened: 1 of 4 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/security/ITableAuthCheckerImpl.java
```diff
@@ -66,23 +66,19 @@ public void checkDatabaseVisibility(
...
+      // The audit database only requires audit privilege
+      boolean hasAuditPrivilege =
+          AuthorityChecker.checkSystemPermission(userName, PrivilegeType.AUDIT);
+      recordAuditLog(
+          auditEntity
+              .setAuditLogOperation(AuditLogOperation.QUERY)
+              .setPrivilegeType(PrivilegeType.AUDIT)
+              .setResult(hasAuditPrivilege),
+          () -> databaseName);
       if (AuthorityChecker.checkSystemPermission(userName, PrivilegeType.AUDIT)) {
```
JackieTien97:
```suggestion
      if (hasAuditPrivilege)) {
```

CRZbulabula: Fixed.

PR #16567, opened 2025-10-13
Fix aggregation query when device cross region in TreeModel

Wei-hao-Li:


integration-test/src/test/java/org/apache/iotdb/db/it/alignbydevice/IoTDBAlignByDeviceWithTemplateIT.java
```diff
@@ -970,12 +970,12 @@ public void emptyResultTest() {
     String[] expectedHeader = new String[] {"Time,Device,s3,s1,s2"};
     String[] retArray = new String[] {};
     resultSetEqualTest(
-        "SELECT * FROM root.sg1.** where time>=now()-1d and time<=now() "
+        "SELECT * FROM root.sg1.** where time>=now()-1ms and time<=now() "
```
JackieTien97, line 976: why do this change?

PR #16576, opened 2025-10-13
Optimize memtable scan

shuwenwei:
## Description
Optimize memtable scan
1. unpackOneFakeMemChunkMetaData exits early after exceeding filter timeranges
2. Non-aligned series push offset and limit down to the nextBatch method during the construction of TsBlock
3. Aligned series handles pushDownFilter, offset, and limit when the rebuildTsBlock method needs to be called
4. Skip unnecessary timestamps by binary search TVList
5. Added a LazyBitmap to avoid constructing a too large Bitmap when calling nextBatch of AlignedTVListIterator

## Optimization effect
100w rows of a non aligned series in memtable
Querying a single point with predicate 'where time = xxx'
image
After this optimization
image

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/fragment/FragmentInstanceContext.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/fragment/FragmentInstanceContext.java
@@ -524,11 +524,17 @@ public class FragmentInstanceContext extends QueryContext {
     if (globalTimeFilter == null) {
       return Collections.singletonList(new TimeRange(Long.MIN_VALUE, Long.MAX_VALUE));
     }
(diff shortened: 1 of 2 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/fragment/FragmentInstanceContext.java
```diff
@@ -517,6 +520,17 @@ public Filter getGlobalTimeFilter() {
     return globalTimeFilter;
   }
 
+  public List<TimeRange> getGlobalTimeFilterTimeRanges() {
```
JackieTien97, line 523: should think about concurrently call

JackieTien97 (changes requested):

PR #16580, opened 2025-10-14
fix compatibility issues when loading snapshot in pipe.

wenyanshi-123:
## Description
Fix compatibility issues when loading snapshot in pipe.When reading old user files, pipe-loadsnapshpt needs to be compatible with reading both username and userid formats.

### Content3 ...

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/auth/user/UserId.java
```diff
@@ -0,0 +1,27 @@
+package org.apache.iotdb.commons.auth.user;
+
+public class UserId {
```
CRZbulabula: Why u need a specified class to process user_id?

PR #16602, opened 2025-10-16
Add disk & network metrics for windows

jt2594838:


Caideyipi (commented):
This PR appears to have been superseded by #17290, with follow-up fixes in #17423 and #17747. Since it is now conflicting with `master`, I recommend closing it as superseded rather than rebasing this implementation.

For completeness, I reviewed the current head (`c196e0c`) and found the following correctness issues:

1. **[P1] Process I/O counters never refresh on Windows**

   [`WindowsDiskMetricsManager.java:110-117`](https://github.com/apache/iotdb/blob/c196e0c9f8ab55a042b96853989c61f70d33650b/iotdb-core/metrics/interface/src/main/java/org/apache/iotdb/metrics/metricsets/disk/WindowsDiskMetricsManager.java#L110-L117)

   `OSProcess` is obtained only once in the constructor. OSHI 6.4.0 treats it as a snapshot, so `getBytesRead()` and `getBytesWritten()` remain at their construction-time values unless `updateAttributes()` is called or a fresh process object is fetched. I verified this on Windows: after writing 16 MiB, the value stayed at `235008` without a refresh and became `17012224` immediately after `updateAttributes()`. Consequently, the new `process_io_size{actual_read/actual_write}` gauges are effectively frozen.

2. **[P1] Average disk I/O size is calculated from the wrong counters and units**

   [`WindowsDiskMetricsManager.java:158-166`](https://github.com/apache/iotdb/blob/c196e0c9f8ab55a042b96853989c61f70d33650b/iotdb-core/metrics/interface/src/main/java/org/apache/iotdb/metrics/metricsets/disk/WindowsDiskMetricsManager.java#L158-L166)

   The update path records only transfer time and read/write size; it never populates `incrementReadOperationCountForDisk` or `incrementWriteOperationCountForDisk`. Both average-size getters therefore divide by the fallback value `1`. In addition, the sampled byte counts have already been converted to KiB and are then truncated through `longValue()`. The result is a truncated KiB delta rather than bytes per operation. Raw `readBytes`/`writeBytes` and `reads`/`writes` should be sampled together and differenced.

3. **[P2] The Windows file-count gauge is never unbound**

   [`SystemRelatedFileMetrics.java:104-113`](https://github.com/apache/iotdb/blob/c196e0c9f8ab55a042b96853989c61f70d33650b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/metrics/file/SystemRelatedFileMetrics.java#L104-L113)

   `bindTo()` was extended to Windows, but `unbindFrom()` still handles only Linux and macOS. Metric service reload/unbind therefore leaves the Windows auto-gauge registered.

4. **[P2] Network interfaces are keyed by a non-unique display name**

   [`WindowsNetMetricManager.java:90-95`](https://github.com/apache/iotdb/blob/c196e0c9f8ab55a042b96853989c61f70d33650b/iotdb-core/metrics/interface/src/main/java/org/apache/iotdb/metrics/metricsets/net/WindowsNetMetricManager.java#L90-L95)

   `NetworkIF.getDisplayName()` is not guaranteed to be unique. Machines with multiple identical adapters can collapse entries in both the `Set` and metric maps, losing or mixing interface statistics. A stable unique identifier such as `getName()` or the interface index should be used consistently.

5. **[P2] The binary-license dependency coordinate is incomplete**

   [`LICENSE-binary:284`](https://github.com/apache/iotdb/blob/c196e0c9f8ab55a042b96853989c61f70d33650b/LICENSE-binary#L284)

   `com.github.oshi:6.4.0` should be `com.github.oshi:oshi-core:6.4.0` so the bundled MIT-licensed artifact is identified unambiguously.

Verification: `mvn -pl iotdb-core/metrics/interface -DskipTests compile` and `git diff --check` both pass at the PR head. However, no tests were added for the new Windows implementations, and Codecov reports 0% patch coverage.

PR #16604, opened 2025-10-17
Support a new read_consistency_level

JackieTien97:
This pull request introduces support for a new read consistency level, `follower_read`, across the IoTDB codebase. The changes ensure that configuration, logic, and documentation now recognize and correctly handle this new option alongside the existing `strong` and `weak` levels.

**Read Consistency Level Enhancements:**

* Added `FOLLOWER_READ` to the `ReadConsistencyLevel` enum in `ReadConsistencyLevel.java`, allowing the system to use follower replicas for read operations.
* Updated the configuration logic in `ConfigNodeDescriptor.java` and `IoTDBConfig.java` to accept and validate the new `follower_read` value, throwing exceptions for unknown values. [[1]](diffhunk://#diff-3ed021394fbf9a40c9ad48c24dfb3ba90ebb20aba5bdcf9a7763ebe0cca5e70dL362-R364) [[2]](diffhunk://#diff-30d3dc10c256a0068e3b87dd128011b653b4a8ecd0f0bf1149979022b5d92ea7L3345-R3351)
* Enhanced the planner logic in `AbstractFragmentParallelPlanner.java` to select follower replicas when `FOLLOWER_READ` is specified, ensuring the leader is skipped and a follower is chosen based on session ID.
* Updated the configuration template in `iotdb-system.properties.template` to document the new `follower_read` option for users.

**Refactoring and Logic Separation:**

* Refactored the data node selection logic in `AbstractFragmentParallelPlanner.java` by extracting target index determination into a separate method, improving readability and maintainability.

Supposing we have a 3C3D cluster and data replica number is set to 3, we execute the following sqls to prepare data:

```
create database test;
use test;
create table t1(device_id STRING TAG, s1 float);
insert into t1 values(now(),'d1', 1.0); 
```

Before this pr, if we do `select * from t1;` query, the query will always be routed to leader:
f374fca129e9c9da5bb8f5e86e53384f

After this pr, if we set `read_consistency_level=follower_read` and then restart the cluster, then the above query will always be routed to any follower:
edcf2ad9bdf44cb525d6ef196c171704

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/AbstractFragmentParallelPlanner.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/AbstractFragmentParallelPlanner.java
@@ -137,6 +137,11 @@ public abstract class AbstractFragmentParallelPlanner implements IFragmentParall
   }
 
   private int getTargetIndex(List<TDataNodeLocation> availableDataNodes) {
+    // if only one node is available, just return 0
+    if (availableDataNodes.size() == 1) {
+      return 0;
+    }
+
     int targetIndex;
     if (ReadConsistencyLevel.STRONG == this.readConsistencyLevel
         || queryContext.getSession() == null) {
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/AbstractFragmentParallelPlanner.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/AbstractFragmentParallelPlanner.java
@@ -116,7 +116,6 @@ public abstract class AbstractFragmentParallelPlanner implements IFragmentParall
       throw new IllegalArgumentException(
           String.format("regionReplicaSet is invalid: %s", regionReplicaSet));
     }
-    boolean selectRandomDataNode = ReadConsistencyLevel.WEAK == this.readConsistencyLevel;
 
     // When planning fragment onto specific DataNode, the DataNode whose endPoint is in
     // black list won't be considered because it may have connection issue now.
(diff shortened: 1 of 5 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/AbstractFragmentParallelPlanner.java
```diff
@@ -133,13 +132,29 @@ protected TDataNodeLocation selectTargetDataNode(TRegionReplicaSet regionReplica
...
       targetIndex = 0;
-    } else {
+    } else if (ReadConsistencyLevel.WEAK == this.readConsistencyLevel) {
       targetIndex = (int) (queryContext.getSession().getSessionId() % availableDataNodes.size());
+    } else if (ReadConsistencyLevel.FOLLOWER_READ == this.readConsistencyLevel) {
+      // The first available data node is always leader which is guaranteed by ConfigNode and
+      // PartitionFetcher in DataNode
+      // We only need to randomly choose any one from [1, availableDataNodes.size()).
+      // SessionId is a unchanged long value for each connection, so we can use that as random seed
+      targetIndex =
```
OneSizeFitsQuorum, line 156: When we only have one replica, maybe we should return 0 even user set consistency_level to follower_read, return a number >= 1 will cause an error?

JackieTien97: Good catch

JackieTien97: fixed

OneSizeFitsQuorum (approved):
LGTM for the code~

As for the name， strong, weak, and follower_read don't seem very consistent... but I can't think of better names either.

PR #16606, opened 2025-10-17
Cache non-exist timeseries metadata to reduce IO

jt2594838:
When repeatedly querying a timeseries that is not in a TsFile and the bloomfilter fails to filter it, each query will result in a necessary IO.

Since TsFile is immutable, we can cache a placeholder to indicate the non-existence.

In a test, 10000 series were written into a TsFile, then 20000 series (only 10000 exist) were queried from the file.
The result shows a significant reduction in query time.
```
warming up
Do not cache non-exist series
time cost with outer k: 460ms
time cost with inner k: 199ms
Cache non-exist series
time cost with outer k: 200ms
time cost with inner k: 151ms
actual test
Do not cache non-exist series
time cost with outer k: 189ms
time cost with inner k: 187ms
Cache non-exist series
time cost with outer k: 156ms
time cost with inner k: 152ms
```

For 0.25M timeseries
```
warming up
Do not cache non-exist series
time cost with outer k: 15580ms
time cost with inner k: 13881ms
Cache non-exist series
time cost with outer k: 11843ms
time cost with inner k: 10173ms
actual test
Do not cache non-exist series
time cost with outer k: 14940ms
time cost with inner k: 13383ms
Cache non-exist series
time cost with outer k: 11451ms
time cost with inner k: 10177ms
```

For 1M timeseries
```
warming up
Do not cache non-exist series
time cost with outer k: 67343ms
time cost with inner k: 55918ms
Cache non-exist series
time cost with outer k: 65831ms
time cost with inner k: 43465ms
actual test
Do not cache non-exist series
time cost with outer k: 67438ms
time cost with inner k: 56995ms
Cache non-exist series
time cost with outer k: 56674
time cost with inner k: 45369ms
```


Caideyipi (approved):
LGTM

Caideyipi:
I found a few correctness risks around the negative metadata cache. The optimization itself makes sense, but the cache
  entry currently loses some important context and clear() changes the cache invalidation semantics.

  Inline comments

  1. On TimeSeriesMetadataCache#get, around the placeholder insertion:

  This negative cache entry does not preserve the `ignoreNotExists` semantics.

  If one read uses `ignoreNotExists=true` and the device is missing, this code may cache the placeholder. A later read
  for the same key with `ignoreNotExists=false` will hit the placeholder and return `null`, while the original
  `reader.readTimeseriesMetadata(..., false, ...)` path should throw an `IOException` for a missing device.

  Can we only cache the placeholder when we are sure the device exists and only the measurement is missing, or avoid
  caching negative results produced under `ignoreNotExists=true`?

  2. On TimeSeriesMetadataCache#clear:

  This changes the semantics of `clear()` from invalidating the existing Caffeine cache to replacing the `lruCache`
  reference.

  `clear()` is used by runtime paths such as cache clearing and snapshot/data-region replacement. With the new
  implementation, concurrent readers may still be operating on the old cache instance and can continue to read stale
  metadata after `clear()` returns. Also, `lruCache` is not `volatile`, so the replacement does not have a clear
  visibility guarantee.

  I think we should keep `lruCache` final and continue using `invalidateAll()` + `cleanUp()` here.

  3. On the placeholder put after the synchronized block:

  The negative-cache `put` happens outside the per-device/file synchronized block.

  For an existing metadata entry, the cache is populated inside the lock, so concurrent requests for the same missing
  key avoid duplicate IO after the first load. For the placeholder path, waiting threads can enter the lock before the
  first thread inserts the placeholder, miss the cache again, and repeat the same disk read.

  Can we insert the placeholder inside the same critical section that performs the disk lookup?

  4. On DataNodeMemoryConfig default value:

  Should this fallback use `isMayCacheNonExistSeries()` instead of `isMetaDataCacheEnable()`?

  The field default is controlled by `mayCacheNonExistSeries`, so using `metaDataCacheEnable` here makes the new option
  implicitly depend on a different setting when the property is absent.

  Test coverage comment

  The added test is ignored as a performance test, so CI will not cover the new behavior. It would be useful to add a
  small non-ignored unit test for the negative cache semantics, especially the `ignoreNotExists=true` followed by
  `ignoreNotExists=false` case.

PR #16607, opened 2025-10-17
Avoid NPE in the Procedure framework

CRZbulabula:


OneSizeFitsQuorum (approved):
LGTM

Use optional maybe better

PR #16632, opened 2025-10-21
Pipe: support path exclusion under tree model

VGalaxies:
This PR adds support for path exclusion patterns in the tree model for pipe data filtering.


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/datastructure/pattern/ExclusionTreePattern.java
```diff
@@ -0,0 +1,107 @@
...
+import org.apache.tsfile.file.metadata.IDeviceID;
+
+/**
+ * Represents a pattern that includes data matched by an inclusion pattern, except for data matched
+ * by an exclusion pattern.
+ *
+ * <p>The logic implemented in the methods is: "Matches the inclusion pattern AND NOT the exclusion
+ * pattern."
+ */
+public class ExclusionTreePattern extends TreePattern {
```
Caideyipi: Since the inclusion is in it, may be "TreePatternWithExclusion" is better

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/datastructure/pattern/ExclusionTreePattern.java
```diff
@@ -0,0 +1,107 @@
...
+        + inclusionPattern.getPattern()
+        + "), EXCLUSION("
+        + exclusionPattern.getPattern()
+        + ")";
+  }
+
+  @Override
+  public boolean isRoot() {
+    // Matches "all" only if inclusion is "all" and exclusion is not "all"
+    return inclusionPattern.isRoot() && !exclusionPattern.isRoot();
```
Caideyipi: If the inclusion is "all" and exclusion is empty?

Caideyipi (commented):
PTAL

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/datastructure/pattern/ExclusionTreePattern.java
VGalaxies: PixPin_2025-11-05_15-29-41

The constructor of ExclusionIoTDBTreePattern guarantees that exclusionPattern is not empty.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/datastructure/pattern/ExclusionTreePattern.java
VGalaxies: I'm wondering whether we need to keep the naming consistent with UnionIoTDBTreePattern and UnionTreePattern (i.e., place the qualifier before "TreePattern" as a prefix)?

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/datastructure/pattern/WithExclusionIoTDBTreePattern.java
```diff
@@ -0,0 +1,208 @@
...
+    this.inclusionPattern = inclusionPattern;
+    this.exclusionPattern = exclusionPattern;
+  }
+
+  public ExclusionIoTDBTreePattern(
+      final IoTDBPatternOperations inclusionPattern,
+      final IoTDBPatternOperations exclusionPattern) {
+    super(true);
+    this.inclusionPattern = inclusionPattern;
+    this.exclusionPattern = exclusionPattern;
```
Caideyipi: I think it's better to throw some exception if the exclusion cover the inclusion

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/datastructure/pattern/ExclusionTreePattern.java
Caideyipi: Then maybe WithExclusionTreePattern?

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/datastructure/pattern/ExclusionTreePattern.java
Caideyipi: Then since the exclusion is not empty, the whole pattern is always not root because it may more or less filter some data.

iotdb-core/datanode/src/test/java/org/apache/iotdb/db/pipe/sink/PipeStatementTreePatternParseVisitorTest.java
```diff
@@ -75,140 +93,234 @@ public void testCreateTimeSeries() throws IllegalPathException {
...
+    Assert.assertFalse(
+        new PipeStatementTreePatternParseVisitor()
+            .visitCreateTimeseries(createTimeSeriesStatementToFilter, exclusionPattern)
+            .isPresent());
   }
 
   @Test
   public void testAlterTimeSeries() throws IllegalPathException {
-    final AlterTimeSeriesStatement alterTimeSeriesStatement = new AlterTimeSeriesStatement(true);
+    final CreateAlignedTimeSeriesStatement expectedS1Only = new CreateAlignedTimeSeriesStatement();
```
Caideyipi, line 136: May check the mismatch of content and title

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/datastructure/pattern/WithExclusionIoTDBTreePattern.java
Caideyipi: Leave a warn log if some of the inclusions are covered by the exclusions but some are not, and it's better to clear the covered ones. The users may not know whether an inclusion is covered when the number is large, so it's may not be wise to refuse the creation, however we can handle it in our codes.

jt2594838 (approved):

PR #16642, opened 2025-10-23
[AINode] Make DataNode Directly Connect to AINode

Alchuang22-dev:
## Description

- Currently Updating

- Should be reviewed by @CRZbulabula 

This PR will refactor the AI model management and inference workflow to remove unnecessary dependencies on the ConfigNode for AI-related operations.

Previously, all AI tasks (such as model loading, inference, and training) were routed through the ConfigNodeClient before reaching the AINode, causing redundant RPC hops and tight coupling between DataNode and ConfigNode.

With this change:

- DataNode now communicates directly with AINode for all AI-related requests, including model loading/unloading, training, inference, and device queries.

- `ModelFetcher` no longer retrieves AINode endpoints via `ConfigNode.getModelInfo`; instead, it resolves AINode addresses locally from configuration or environment variables.

- `ClusterConfigTaskExecutor` now calls `AINodeClient` directly instead of routing through `ConfigNodeClient`.

- AI-related methods in `ConfigNodeClient` are marked as deprecated and throw exceptions to prevent accidental use.

- Database-related operations that require ConfigNode coordination (e.g., `createModel`, `dropModel`, `updateModelInfo`) remain unchanged.

This update effectively turns the DataNode into an independent AI infrastructure layer, improving performance and simplifying the AI service architecture while keeping core database management logic intact.

<hr>

This PR has:
- [x] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

**The PR will change**:

- `ClusterConfigTaskExecutor.java`
- `ModelManager.java`
- `ModelFetcher.java`
- Other critical files

### DIFF
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/partition/PartitionManager.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/partition/PartitionManager.java
@@ -1003,14 +1003,10 @@ public class PartitionManager {
   }
 
   public Optional<TConsensusGroupId> generateTConsensusGroupIdByRegionId(final int regionId) {
-    if (configManager
-        .getPartitionManager()
-        .isRegionGroupExists(new TConsensusGroupId(TConsensusGroupType.SchemaRegion, regionId))) {
+    if (isRegionGroupExists(new TConsensusGroupId(TConsensusGroupType.SchemaRegion, regionId))) {
       return Optional.of(new TConsensusGroupId(TConsensusGroupType.SchemaRegion, regionId));
     }
-    if (configManager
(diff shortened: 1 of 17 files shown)


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/ModelManager.java
```diff
@@ -54,7 +54,9 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.*;
```
CRZbulabula: Do not import *, plz reset ur IDEA config

PR #16686, opened 2025-11-01
[AINode] Refactor table forecast function

CRZbulabula:
To prepare for supporting multi/co-variable models in AINode, we refactor the table forecast function in this PR, main ideas and updates include:

1. When invoking table value forecast function, all specified input columns should be delivered to AINode as the model's input channels.
2. When invoking table value forecast function, the users are suggested to specify the `PREDICATED_COLUMNS`. In this way, the IoTDB can realize the count and data types of the forecast result.


RkGrit:
LGTM!
But why not also support multi/co-variable models for the tree value forecast function?

PR #16691, opened 2025-11-03
SeriesScanUtil throws exception when using filters that could not match any time range

shuwenwei:
## Description
```
IoTDB:test_db> explain analyze verbose select date_bin(5ms, time) as time, region_id, int32, int64, avg(float) as avg_value from table_1 where time >= 2 and time < 1 GROUP BY 1,2,3,4;
Msg: org.apache.iotdb.jdbc.IoTDBSQLException: 301: min:9223372036854775807 should not be larger than max: -9223372036854775808
```

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/source/SeriesScanUtil.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/source/SeriesScanUtil.java
@@ -219,10 +219,7 @@ public class SeriesScanUtil implements Accountable {
     orderUtils.setCurSeqFileIndex(dataSource);
     curUnseqFileIndex = 0;
 
-    List<TsFileResource> seqResources = dataSource.getSeqResources();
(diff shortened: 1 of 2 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/source/SeriesScanUtil.java
```diff
@@ -219,6 +219,14 @@ public void initQueryDataSource(QueryDataSource dataSource) {
...
     curUnseqFileIndex = 0;
 
+    List<TsFileResource> seqResources = dataSource.getSeqResources();
+    List<TsFileResource> unseqResources = dataSource.getUnseqResources();
+    if ((seqResources == null || seqResources.isEmpty())
+        && (unseqResources == null || unseqResources.isEmpty())) {
+      // no satisfied resources
+      return;
+    }
+
```
JackieTien97: put these into `QueryDataSource` as a public method.

PR #16707, opened 2025-11-07
[AINode] Package AINode via PyInstaller

CRZbulabula:
Recently, we heard lots of our users complaint about the entry level for registering our AINode into their deployed IoTDB clusters. Practically, installing AINode requires the user config the correct python version, which is usually accomplished via pyenv or conda etc., and maintain smooth network connection to download python dependencies and model weight files. Nevertheless, as an application program like both the ConfigNode and the DataNode of IoTDB, the AINode is naturally required to be easy to use, aka can be started through a simple line of instruction.

As a result, we realized the former solution--packaging AINode as a python whl--is wrong! To bridge the gap between the Apache IoTDB community and our users, in this PR, we refactored the package pipeline from a whl to an application program! With this new feature, we are now proudly announce that, everyone could easily enjoy the cutting-edge research achievements of the time series models since our next official release by the following command,

`./sbin/start-ainode.<sh/bat>`

without any strenuously environmental setup process!


RkGrit:
LTGM~

PR #16717, opened 2025-11-07
chore!: Bind real config address for Thrift and Raft services

CritasWang:
The listening addresses for Thrift and Raft services are now explicitly bound to a non-loopback interface. This may affect service discovery and communication between nodes in a cluster.
﻿
BREAKING CHANGE: The Thrift and Raft services now bind to the physical interface address (e.g., 192.168.1.100) instead of 0.0.0.0. Similarly, the ExternalRpc Address, which defaults to 0.0.0.0, will now bind to the explicitly configured IP address. Ensure your cluster configuration and firewall rules are updated to allow communication on the real internal network and the configured ExternalRpc address.
﻿

image

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/ExternalRPCService.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/ExternalRPCService.java
@@ -71,7 +71,7 @@ public class ExternalRPCService extends ThriftService implements ExternalRPCServ
                   processor,
                   getID().getName(),
                   ThreadName.CLIENT_RPC_PROCESSOR.getName(),
-                  null,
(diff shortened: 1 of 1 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/service/ExternalRPCService.java
```diff
@@ -71,7 +71,7 @@ public void initThriftServiceThread() throws IllegalAccessException {
                   processor,
                   getID().getName(),
                   ThreadName.CLIENT_RPC_PROCESSOR.getName(),
-                  getBindIP(),
+                  null,
```
HTHou:
```suggestion
                  getBindIP(),
```

PR #16720, opened 2025-11-07
Added name check to avoid directory attack

Caideyipi:
## Description
As the title said.

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/table/DataNodeTableCache.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/table/DataNodeTableCache.java
@@ -107,7 +107,6 @@ public class DataNodeTableCache implements ITableCache {
                               Function.identity(),
                               (v1, v2) -> v2,
                               ConcurrentHashMap::new))));
-
       preCreateMap.forEach(
           (key, value) ->
               preUpdateTableMap.put(
--- a/iotdb-protocol/thrift-datanode/src/main/thrift/datanode.thrift
+++ b/iotdb-protocol/thrift-datanode/src/main/thrift/datanode.thrift
@@ -1211,6 +1211,7 @@ service IDataNodeRPCService {
    */
   common.TSStatus deleteColumnData(TDeleteColumnDataReq req)
 
+
   /**
    * Construct table device black list
    */


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/schema/table/AddTableColumnProcedure.java
```diff
@@ -77,6 +84,10 @@ protected Flow executeFromState(final ConfigNodeProcedureEnv env, final AddTable
         case PRE_RELEASE:
           LOGGER.info("Pre release info of table {}.{} when adding column", database, tableName);
           preRelease(env);
+          if (table.setNeedCheck4Object()) {
+            checkObject(env, database, tableName);
+          }
```
JackieTien97: Actually, there is no need to do this check, if the table has already owned another object type column before.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/execution/config/executor/ClusterConfigTaskExecutor.java
```diff
@@ -4273,6 +4273,12 @@ public SettableFuture<ConfigTaskResult> alterTableAddColumn(
           || (TSStatusCode.COLUMN_ALREADY_EXISTS.getStatusCode() == tsStatus.getCode()
               && columnIfExists)) {
         future.set(new ConfigTaskResult(TSStatusCode.SUCCESS_STATUS));
+      } else if (tsStatus.getMessage().contains("When there are object fields")) {
+        future.setException(
+            new SemanticException(
+                tsStatus
+                    .getMessage()
+                    .replace("org.apache.iotdb.commons.exception.IoTDBRuntimeException: ", "")));
```
JackieTien97: don't use error msg to do this check, use the specific status code to do this check

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/table/DataNodeTableCache.java
```diff
@@ -99,6 +100,11 @@ public void init(final byte[] tableInitializationBytes) {
               databaseTableMap.put(
                   PathUtils.unQualifyDatabaseName(key),
                   value.stream()
+                      .filter(
```
JackieTien97: use filter to do for-each style is not a good habit. You can do another for-each out of the databaseTableMap.put

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/table/DataNodeTableCache.java
```diff
@@ -110,6 +116,18 @@ public void init(final byte[] tableInitializationBytes) {
               preUpdateTableMap.put(
                   PathUtils.unQualifyDatabaseName(key),
                   value.stream()
+                      .filter(
```
JackieTien97: same as above. If you really want to this, you at least add some comments about this.

integration-test/src/test/java/org/apache/iotdb/relational/it/schema/IoTDBTableIT.java
```diff
@@ -761,6 +761,74 @@ public void testConcurrentAutoCreateAndDropColumn() throws Exception {
...
+      statement.execute("create table \".\" ()");
+
+      try {
+        statement.execute("alter table \".\" add column a object");
+        fail();
+      } catch (final SQLException e) {
+        Assert.assertEquals(
+            "701: When there are object fields, the tableName . shall not be '.', '..' or contain './', '.\\'",
+            e.getMessage());
+      }
```
JackieTien97, line 809: also need alter table statement whose table name is correct but column name is not correct.

integration-test/src/test/java/org/apache/iotdb/relational/it/schema/IoTDBTableIT.java
```diff
@@ -761,6 +761,74 @@ public void testConcurrentAutoCreateAndDropColumn() throws Exception {
     }
   }
 
+  @Test
+  public void testTableObjectCheck() throws Exception {
```
JackieTien97, line 784: you need to construct a list which contains all illegal paths and then use for-loop to format each sql

JackieTien97 (changes requested):

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/schema/table/TsTable.java
```diff
@@ -362,6 +386,33 @@ public void setProps(Map<String, String> props) {
...
+          TSStatusCode.SEMANTIC_ERROR.getStatusCode());
+    }
+    for (final TsTableColumnSchema schema : columnSchemaMap.values()) {
+      if (schema.getDataType().equals(TSDataType.OBJECT)
+          && isInvalid4ObjectType(schema.getColumnName())) {
+        throw new MetadataException(
+            getObjectStringError("objectName", schema.getColumnName()),
+            TSStatusCode.SEMANTIC_ERROR.getStatusCode());
+      }
+    }
```
jt2594838, line 430: Why check the column name

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/schema/table/AddTableColumnProcedure.java
```diff
@@ -123,7 +139,52 @@ protected void columnCheck(final ConfigNodeProcedureEnv env) {
...
+              ((tConsensusGroupId, tDataNodeLocations, failureMap) -> {
+                final String message = parseStatus(failureMap.values());
+                // Shall not be SUCCESS here
+                return Objects.nonNull(message)
+                    ? new IoTDBException(message, TSStatusCode.SEMANTIC_ERROR.getStatusCode())
+                    : null;
+              }))
+          .execute();
+    }
+  }
```
jt2594838: Why check while adding a column？

Caideyipi: When adding a object column, we need to ensure that the directory parts does not contain illegal paths

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/schema/table/TsTable.java
Caideyipi: Because the object field name will occur in object's path

jt2594838: Column name shall not occur, only column value will

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/schema/table/AddTableColumnProcedure.java
jt2594838: You can do this during schema validation of insertion. No need to do so in adding columns.
If an existing device will not be inserted with object data, you do not have to fail the column addition.

jt2594838 (approved):

PR #16734, opened 2025-11-11
Support the JDBC client to prevent SQL injection

Young-Leo:
Support the JDBC client to prevent SQL injection

### DIFF
--- a/iotdb-client/jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBPreparedStatement.java
+++ b/iotdb-client/jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBPreparedStatement.java
@@ -917,9 +917,6 @@ public class IoTDBPreparedStatement extends IoTDBStatement implements PreparedSt
   }
 
   private String escapeSingleQuotes(String value) {
-    if (value == null) {
-      return null;
-    }
     // Escape single quotes with double single quotes
     return value.replace("'", "''");
   }


iotdb-client/jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBPreparedStatement.java
```diff
@@ -916,12 +916,38 @@ public void setString(int parameterIndex, String x) {
             && !((x.startsWith("'") && x.endsWith("'"))
                 || ((x.startsWith("\"") && x.endsWith("\""))
                     && "tree".equals(getSqlDialect())))))) {
-      this.parameters.put(parameterIndex, "'" + x + "'");
+      // Escape single quotes to prevent SQL injection: ' -> \'
+      String escapedValue = escapeString(x);
+      this.parameters.put(parameterIndex, "'" + escapedValue + "'");
```
JackieTien97: in tree model, constant string can use both `'` and `"`

iotdb-client/jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBPreparedStatement.java
```diff
@@ -916,12 +916,38 @@ public void setString(int parameterIndex, String x) {
...
+   *
+   * @param value the string to escape
+   * @return the escaped string
+   */
+  private String escapeString(String value) {
+    if (value == null) {
+      return null;
+    }
+    // Escape single quotes to prevent SQL injection
+    return value.replace("'", "\\'");
```
JackieTien97:
```suggestion
    return value.replace("'", "''");
```
using `''` to be the escaper?

iotdb-client/jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBPreparedStatement.java
```diff
@@ -909,19 +909,21 @@ public void setShort(int parameterIndex, short x) throws SQLException {
...
+      this.parameters.put(parameterIndex, null);
     } else {
-      this.parameters.put(parameterIndex, x);
+      this.parameters.put(parameterIndex, "'" + escapeSingleQuotes(x) + "'");
     }
   }
 
+  private String escapeSingleQuotes(String value) {
+    if (value == null) {
+      return null;
```
JackieTien97: throw Exception? You've already do null check in caller.

JackieTien97 (changes requested):

PR #16741, opened 2025-11-12
Adjust the exception message cause by the WHERE clause uses a faulty time range

zerolbsony:
Adjust the exception message caused by the WHERE clause, which uses range comparisons on the same field, especially when the left value of the range is greater than the right value.
img_v3_02rv_5e5d9a3e-9d06-4bac-ab88-6daf460eb7cg

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/parser/ASTVisitor.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/parser/ASTVisitor.java
@@ -344,6 +344,9 @@ public class ASTVisitor extends IoTDBSqlParserBaseVisitor<Statement> {
   private static final String LIMIT_CONFIGURATION_ENABLED_ERROR_MSG =
       "Limit configuration is not enabled, please enable it first.";
 
+  public static final String DELETE_RANGE_COMPARISON_ERROR_MSG =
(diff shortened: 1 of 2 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/parser/ASTVisitor.java
```diff
@@ -344,6 +344,9 @@ public class ASTVisitor extends IoTDBSqlParserBaseVisitor<Statement> {
   private static final String LIMIT_CONFIGURATION_ENABLED_ERROR_MSG =
       "Limit configuration is not enabled, please enable it first.";
 
+  public static final String DELETE_RANGE_COMPARISON_ERROR_MSG =
+      "For delete statement, where clause use a range comparison on the same field, the left value of the range cannot be greater than the right value of the range, it must be written like this : time > 5 and time < 10";
```
jt2594838, line 348: No need to be so wordy,
"Invalid deletion range: [%d, %d]"
should be enough

zerolbsony: > No need to be so wordy, "Invalid deletion range: [%d, %d]" should be enough

Actually I originally wish so, but change my mind after I saw some similar error message template exist.

PR #16766, opened 2025-11-17
[to dev/1.3] Fix slow query threshold & fix a bug for lastQuery & optimize encodeBatch for multi tvlist scene

shuwenwei:
## Description
https://github.com/apache/iotdb/pull/16765
### Default slowQueryThreshold should be 10s
截屏2025-11-17 11 49 57

#### Scenario

Rows: 0–99999 (100,000 rows)
Duplicate timestamps occur at only two positions:

index 123

index 90732

#### Using regular BitMap

Must allocate 100,000 bits immediately.

Even though only 2 bits will be marked.

#### Using LazyBitMap (blockSize = 10,000)

LazyBitMap creates blocks only when needed:

Block for index 123 → block index = 0

Block for index 90732 → block index = 90

Total allocated blocks: 2 blocks
Each block has only 10,000 bits, so memory is:

2 blocks × 10,000 bits = 20,000 bits total

#### Savings

Regular BitMap: 100,000 bits
LazyBitMap: 20,000 bits

➡ 80% memory reduction
➡ Matches the sparse nature of duplicated timestamps


Caideyipi (approved):
LGTM

PR #16799, opened 2025-11-25
Fix that settle compaction may lgnore some newly incoming deletions

jt2594838:
Due to cascading mods files are not properly set before performing a settle compaction, concurrent deletion with a settle compaction may be lost.


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/task/SettleCompactionTask.java
```diff
@@ -136,7 +137,11 @@ protected boolean doCompaction() {
         new File(
             allSourceFiles.get(0).getTsFile().getAbsolutePath()
                 + CompactionLogger.SETTLE_COMPACTION_LOG_NAME_SUFFIX);
+
     try (SimpleCompactionLogger compactionLogger = new SimpleCompactionLogger(logFile)) {
+      CompactionUtils.prepareCompactionModFiles(
+          filesView.targetFilesInPerformer, filesView.sourceFilesInCompactionPerformer);
+
       calculateSourceFilesAndTargetFiles();
```
shuwenwei: targetFilesInPerformer are calculated after this line

jt2594838: fixed

shuwenwei (approved):

PR #16806, opened 2025-11-25
[AINode] Add table value function : classify

yunbow30944:
## Description

+ Add table value function classify, which will be used for time series classification in the future.
+ Extract interface ResultColumnAppender to simplify the code.
+ Limit the input length of udtf forecast.


RkGrit:
LGTM~

PR #16829, opened 2025-11-27
fix object functions & add tests for object type & fix some bugs

shuwenwei:
## Description
fix object functions & add tests for object type & fix some bugs
### Test Coverage Summary
✔ Added in this release

READ_OBJECT
Added tests for reading object data with offset/length, validating correct binary retrieval

✔ Previously existing, newly verified for OBJECT type

LENGTH
Verified that LENGTH returns INT64 for OBJECT columns instead of INT32.

CAST / TRY_CAST
Verified that casting OBJECT → STRING works, and casting to OBJECT is not permitted.

COALESCE
Verified correct fallback behavior with multiple OBJECT values.

IS [NOT] NULL
Verified null evaluation correctness on OBJECT columns.

✔ Aggregation functions supporting OBJECT, validated

COUNT, FIRST, LAST, FIRST_BY, LAST_BY
Verified OBJECT compatibility and correctness of returned values.

✔ Window functions, verified with OBJECT input

FIRST_VALUE, LAST_VALUE, NTH_VALUE, LEAD, LAG
Confirmed support and correctness when applied to OBJECT data.

✔ Row pattern recognition functions, verified

RPR_FIRST, RPR_LAST, PREV, NEXT
Verified that the first argument may be OBJECT.

✔ SQL clause & syntax support, verified

FROM, WHERE, SELECT, HAVING
Confirmed these clauses can read and process OBJECT fields.

✔ Constraints validated

GROUP BY, ORDER BY, PARTITION BY / ORDER BY in MATCH_RECOGNIZE and WINDOW, OFFSET, LIMIT
Verified that these do not support OBJECT as sorting/partitioning keys.

✔ FILL behavior validated

Confirmed that only FILL PREVIOUS is supported for OBJECT values.

✔ TTL behavior validated
Verified that the expired object file can be deleted by compaction.

### DIFF
--- a/iotdb-core/datanode/src/test/java/org/apache/iotdb/db/queryengine/transformation/dag/column/unary/scalar/BlobLengthColumnTransformerTest.java
+++ b/iotdb-core/datanode/src/test/java/org/apache/iotdb/db/queryengine/transformation/dag/column/unary/scalar/BlobLengthColumnTransformerTest.java
@@ -75,7 +75,7 @@ public class BlobLengthColumnTransformerTest {
     Column result = blobLengthColumnTransformer.getColumn();
 
     int expectedLength = input.getBytes(StandardCharsets.UTF_8).length;
-    Assert.assertEquals(expectedLength, result.getInt(0));
+    Assert.assertEquals(expectedLength, result.getLong(0));
   }
 
   @Test
@@ -93,7 +93,7 @@ public class BlobLengthColumnTransformerTest {
     Column result = blobLengthColumnTransformer.getColumn();
 
     int expectedLength = inputBytes.length;
-    Assert.assertEquals(expectedLength, result.getInt(0));
+    Assert.assertEquals(expectedLength, result.getLong(0));
   }
 
   @Test
@@ -113,9 +113,9 @@ public class BlobLengthColumnTransformerTest {
     blobLengthColumnTransformer.addReferenceCount();
     blobLengthColumnTransformer.evaluate();
     Column result = blobLengthColumnTransformer.getColumn();
-    Assert.assertEquals(inputBytes1.length, result.getInt(0));
+    Assert.assertEquals(inputBytes1.length, result.getLong(0));
     Assert.assertTrue(result.isNull(1));
-    Assert.assertEquals(inputBytes2.length, result.getInt(2));
+    Assert.assertEquals(inputBytes2.length, result.getLong(2));
   }
 
   @Test
@@ -142,7 +142,7 @@ public class BlobLengthColumnTransformerTest {
     int expectedValue3 = bytes3.length;
 
     Assert.assertTrue(result.isNull(0));
-    Assert.assertEquals(expectedValue2, result.getInt(1));
-    Assert.assertEquals(expectedValue3, result.getInt(2));
+    Assert.assertEquals(expectedValue2, result.getLong(1));
+    Assert.assertEquals(expectedValue3, result.getLong(2));
   }
 }


iotdb-core/confignode/src/test/resources/confignode1conf/iotdb-system.properties
```diff
@@ -34,7 +34,7 @@ timestamp_precision=ms
 data_region_consensus_protocol_class=org.apache.iotdb.consensus.iot.IoTConsensus
 schema_region_consensus_protocol_class=org.apache.iotdb.consensus.ratis.RatisConsensus
 schema_replication_factor=3
-data_replication_factor=3
+data_replication_factor=1
```
JackieTien97: change it back

integration-test/src/test/java/org/apache/iotdb/relational/it/query/recent/IoTDBObjectTypeQueryIT.java
```diff
@@ -0,0 +1,295 @@
...
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+
+import java.sql.Connection;
+import java.sql.Statement;
+import java.time.LocalDate;
+
+@RunWith(IoTDBTestRunner.class)
+@Category({TableLocalStandaloneIT.class, TableClusterIT.class})
+public class IoTDBObjectTypeQueryIT {
```
JackieTien97: move this class to integration-test/src/test/java/org/apache/iotdb/relational/it/query/object package

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/transformation/dag/column/unary/scalar/ReadObjectColumnTransformer.java
```diff
@@ -107,35 +100,14 @@ private void transform(Column column, ColumnBuilder columnBuilder, int i) {
   }
 
   private Binary readObject(Binary binary) {
-    File file = ObjectTypeUtils.getObjectPathFromBinary(binary);
-    long actualReadSize = getActualReadSize(file);
+    Pair<Long, String> ObjectLengthPathPair = ObjectTypeUtils.parseObjectBinary(binary);
```
JackieTien97:
```suggestion
    Pair<Long, String> objectLengthPathPair = ObjectTypeUtils.parseObjectBinary(binary);
```

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/ObjectTypeUtils.java
```diff
@@ -40,6 +58,117 @@ public class ObjectTypeUtils {
...
+              .borrowClient(dataNodeLocation.getInternalEndPoint())) {
+        while (toReadSizeInCurrentDataNode > 0) {
+          req.setOffset(offset + buffer.position());
+          req.setSize(Math.min(toReadSizeInCurrentDataNode, batchSize));
+          toReadSizeInCurrentDataNode -= req.getSize();
+          ByteBuffer partial = client.readObject(req);
+          buffer.put(partial);
+        }
+      } catch (Exception e) {
+        logger.error("Failed to read object from datanode: {}", dataNodeLocation, e);
```
JackieTien97:
```suggestion
        logger.warn("Failed to read object from datanode: {}", dataNodeLocation, e);
```

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/ObjectTypeUtils.java
```diff
@@ -40,6 +58,117 @@ public class ObjectTypeUtils {
...
+        if (i == regionReplicaSet.getDataNodeLocations().size() - 1) {
+          throw new IoTDBRuntimeException(e, TSStatusCode.INTERNAL_SERVER_ERROR.getStatusCode());
```
JackieTien97: We've already record the real thread exception in warn log, here we only need a new error code and a error msg to notify users that this problem.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/ObjectTypeUtils.java
```diff
@@ -40,6 +58,117 @@ public class ObjectTypeUtils {
...
+        }
+        buffer.clear();
+        req.setOffset(offset);
```
JackieTien97: We can continue from the last break point?

JackieTien97: if you want to read from start, req.setOffset(offset) is also needless.

JackieTien97 (changes requested):

PR #16877, opened 2025-12-08
Pipe: Fixed the bug that reused plugins may not get loader and visibility

Caideyipi:
## Description
When a table model pipeplugin is reusing another jar, the visibility logic will be skipped and it will only be shown at tree model. This PR has fixed this problem.

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/pipe/PipePluginInfo.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/pipe/PipePluginInfo.java
@@ -233,7 +233,7 @@ public class PipePluginInfo implements SnapshotProcessor {
         } else {
           throw new PipeException(
               String.format(
-                  "The %s's creation has not passed in jarName, which does not exists in other pipePlugins. Please check",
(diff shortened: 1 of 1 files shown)


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/pipe/PipePluginInfo.java
```diff
@@ -212,34 +212,29 @@ public TSStatus createPipePlugin(final CreatePipePluginPlan createPipePluginPlan
...
+        computeFromPluginClass(pluginName, className);
+      } else {
+        final String existed = pipePluginMetaKeeper.getPluginNameByJarName(jarName);
+        if (Objects.nonNull(existed)) {
+          pipePluginExecutableManager.linkExistedPlugin(existed, pluginName, jarName);
+          computeFromPluginClass(pluginName, className);
+        } else {
+          throw new PipeException(
+              String.format(
+                  "The %s's creation has not passed in jarName, which does not exists in other pipePlugins. Please check",
```
jt2594838: does not exist

PR #16888, opened 2025-12-09
Avoid CN OOM by Pulling User/Roles to DN When Cache Misses.

wenyanshi-123:
When there is no User Cache on the DN side, the entire time series will be handed over to CN for authentication, which ultimately leads to CN OOM in some PIPE scenarios with a large number of time series. To solve this problem, it is necessary to prohibit passing all time series to CN during authentication, and instead pull the user and it's corresponding roles to dn when the user cache and role caches fail.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/auth/ClusterAuthorityFetcher.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/auth/ClusterAuthorityFetcher.java
@@ -186,7 +186,7 @@ public class ClusterAuthorityFetcher implements IAuthorityFetcher {
             checkRoleFromConfigNode(username, rolename);
             cachedRole = iAuthorCache.getRoleCache(rolename);
           }
-          if (cachedRole.checkPathPrivilege(path, permission)) {
+          if (cachedRole != null && cachedRole.checkPathPrivilege(path, permission)) {
             checkFromRole = true;
             break;
           }


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/auth/ClusterAuthorityFetcher.java
```diff
@@ -169,35 +169,31 @@ public List<Integer> checkUserPathPrivileges(
...
           }
         }
-        pos++;
+        if (!checkFromRole) {
+          posList.add(pos);
+        }
       }
-      return posList;
-    } else {
-      return checkPathFromConfigNode(username, allPath, permission);
```
CRZbulabula, line 199: Remove `checkPathFromConfigNode`, avoid other developers to invoke it.

CRZbulabula (changes requested):
PTAL.

CRZbulabula (approved):
LGTM!

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/auth/ClusterAuthorityFetcher.java
```diff
@@ -169,35 +169,31 @@ public List<Integer> checkUserPathPrivileges(
...
+        for (String rolename : user.getRoleSet()) {
+          Role cachedRole = iAuthorCache.getRoleCache(rolename);
+          if (cachedRole == null) {
+            checkRoleFromConfigNode(username, rolename);
           }
-          if (!checkFromRole) {
-            posList.add(pos);
+          if (cachedRole.checkPathPrivilege(path, permission)) {
+            checkFromRole = true;
+            break;
```
jt2594838, line 191: If cachedRole is null in line 182, it will still be null in line 185.

jt2594838:
And, how is this PR related to PIPE?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/auth/ClusterAuthorityFetcher.java
jt2594838: `cachedRole` is a local variable, how could it be modified in line 183?

wenyanshi-123: I’ve fixed the issue.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/auth/ClusterAuthorityFetcher.java
```diff
@@ -169,35 +169,32 @@ public List<Integer> checkUserPathPrivileges(
...
+    int pos = 0;
+    for (PartialPath path : allPath) {
+      if (!user.checkPathPrivilege(path, permission)) {
+        boolean checkFromRole = false;
+        for (String rolename : user.getRoleSet()) {
+          Role cachedRole = iAuthorCache.getRoleCache(rolename);
+          if (cachedRole == null) {
+            checkRoleFromConfigNode(username, rolename);
+            cachedRole = iAuthorCache.getRoleCache(rolename);
           }
```
jt2594838, line 188: There is still a chance that the role is evicted between line 183 and 184, therefore, cachedRold can still be null.
One possible solution is to make `checkRoleFromConfigNode()` return the Role it just fetched.
And may rename `checkRoleFromConfigNode()` to `fetchRoleFromConfigNode()` which is more precise.

jt2594838:
Avoid using force-push, which will mix your new changes with the old ones.
This forces me to go through the whole changes again, wasting some time.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/auth/ClusterAuthorityFetcher.java
jt2594838: And what if the role is deleted concurrently? You can still not fetch it.

wenyanshi-123: I have resolved the possible NPE issue, but the concurrency problem requires a more systematic design.

PR #16889, opened 2025-12-09
Remove the code check port is occupied and resolve the problem that can't rename file successfully

zerolbsony:
When starting datanode on windows os, the error will occur :
`Failed to replace formal system properties file, you may manually rename it: system.properties.tmp -> system.properties`

Because here use the FileUtils related functions from the ratis package, instead of similar function from node-commons.

### DIFF
--- a/iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/file/SystemPropertiesHandler.java
+++ b/iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/file/SystemPropertiesHandler.java
@@ -184,15 +184,8 @@ public abstract class SystemPropertiesHandler {
               "Delete formal system properties file fail: %s", formalFile.getAbsoluteFile());
       throw new IOException(msg);
     }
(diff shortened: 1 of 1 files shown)


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/file/SystemPropertiesHandler.java
```diff
@@ -185,8 +185,8 @@ private void replaceFormalFile() throws IOException {
       throw new IOException(msg);
     }
     try {
-      FileUtils.move(tmpFile.toPath(), formalFile.toPath());
-    } catch (IOException e) {
+      FileUtils.moveFileSafe(tmpFile, formalFile);
+    } catch (Exception e) {
```
CRZbulabula: Should catch the corresponding exception rather than `Exception`.

CRZbulabula (changes requested):
PTAL.

PR #16900, opened 2025-12-11
Hotfix concurrent connect error

zerolbsony:


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/session/IClientSession.java
```diff
@@ -37,27 +38,27 @@
...
+  private volatile boolean login = false;
 
   private long logInTime;
 
   private SqlDialect sqlDialect = SqlDialect.TREE;
 
   @Nullable private String databaseName;
 
-  private long lastActiveTime;
+  private volatile long lastActiveTime = CommonDateTimeUtils.currentTime();
```
JackieTien97: No need to add these volatile? As discussed last night, ConcurrentHashMap has already made sure that getAllConnections can see the changes made by other threads.

CRZbulabula (approved):
LGTM!

PR #16927, opened 2025-12-18
prevent auto_map gets covered

RkGrit:
prevent auto_map gets covered and rename chronos's model.py


RkGrit:
@CRZbulabula Please review this PR

PR #16928, opened 2025-12-19
prevent auto_map gets covered and add model_list for IT

RkGrit:
prevent auto_map gets covered and add model_list for AINodeConcurrentForecastIT


RkGrit:
@CRZbulabula 
Plz review this pr.

PR #16957, opened 2025-12-26
Pipe: Enabled config audit logger & Fixed some checks of config privilege & keep the tree model show pipes aligned with table model's & Fixed the bug that password with "-" prefix cannot be directly entered in Cli & Fixed some pipe ITs & Added the missing source privilege check for PipeAlterEncodingCompressorPlan

Caideyipi:
## Description
As the title said.

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/security/TreeAccessCheckVisitor.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/security/TreeAccessCheckVisitor.java
@@ -1996,7 +1996,7 @@ public class TreeAccessCheckVisitor extends StatementVisitor<TSStatus, TreeAcces
         };
     auditEntity.setPrivilegeTypes(privilegeList);
     if (AuthorityChecker.SUPER_USER.equals(auditEntity.getUsername())) {
-      recordObjectAuthenticationAuditLog(auditEntity.setResult(true), supplier);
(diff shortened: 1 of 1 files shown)


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/audit/AbstractAuditLogger.java
```diff
@@ -47,4 +48,16 @@ public abstract class AbstractAuditLogger {
   public boolean noNeedInsertAuditLog(IAuditEntity auditLogFields) {
     return true;
   }
+
+  public void recordAuditLog(final IAuditEntity auditEntity, final Supplier<String> auditObject) {
```
CRZbulabula: I don't think it is a good idea that roughly moving the `recordAuditLog` function from AuthCheckImpl to AbstractAuditLogger, because the current implementation is only for `OBJECT_AUTHENTICATION` audit. If u indeed require this refactor, I know current change is for pipe audit, at least rename this interface to `recordObjectAuthenticationAuditLog`.

CRZbulabula (changes requested):
PTAL.

CRZbulabula (approved):
LGTM!

jt2594838 (approved):

PR #16998, opened 2026-01-08
[AINode] Integrate device manager framework

CRZbulabula:
To support AINode managing models under different process units, we introduce device manager in this PR. We provide unified backend usage interfaces, which can be found at `iotdb/ainode/core/manager/device_manager.py`.


yunbow30944:
LGTM

RkGrit:
LGTM~

PR #17015, opened 2026-01-12
Don't allow IntColumn DataType is DATE when query column data of Date type used by Python client

zerolbsony:
Fix this scene that query data used by Python client:
img_v3_02tm_774cd932-7684-4c2f-a5be-6ad0230afa3g
img_v3_02tm_c904f397-1d77-4d19-a996-a2a6db04b24g
In addition, resovle the problem involved can't fetch result from the table view:
only set timeseries metadata to null and don't effect whole line data can be returned when column type is not compatible
image

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -173,7 +173,7 @@
         <thrift.version>0.14.1</thrift.version>
         <xz.version>1.9</xz.version>
         <zstd-jni.version>1.5.6-3</zstd-jni.version>
-        <tsfile.version>2.2.1-260114-SNAPSHOT</tsfile.version>
+        <tsfile.version>2.2.1-260115-SNAPSHOT</tsfile.version>
     </properties>
     <!--
     if we claim dependencies in dependencyManagement, then we do not claim


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/source/SeriesScanUtil.java
```diff
@@ -1230,14 +1233,19 @@ private TsBlock getTransferedDataTypeTsBlock(TsBlock tsBlock) {
...
-            newValueColumns[i] = valueColumns[i];
+            newValueColumns[i] =
+                new IntColumn(
+                    positionCount, Optional.of(new boolean[positionCount]), new int[positionCount]);
+            for (int j = 0; j < valueColumns[i].getPositionCount(); j++) {
+              newValueColumns[i].isNull()[j] = valueColumns[i].isNull()[j];
+              if (!valueColumns[i].isNull()[j]) {
+                newValueColumns[i].getInts()[j] = valueColumns[i].getInts()[j];
+              }
+            }
```
jt2594838: Avoid copying the whole column

jt2594838 (approved):

jt2594838 (approved):

PR #17039, opened 2026-01-19
fix session cpp node-supplier

hongzhi-gao:
## Description

1. Use "SHOW AVAILABLE URLS" instead of "SHOW DATANODES"  
2. The builder of the newly added SessionBuilder will detect the issue of setting both nodeUrls and host:port simultaneously.

### Content3 ...

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/iotdb-client/client-py/iotdb/tsfile/utils/tsblock_serde.py
+++ b/iotdb-client/client-py/iotdb/tsfile/utils/tsblock_serde.py
@@ -18,7 +18,6 @@
 
 import numpy as np
 
-
 # Serialized tsBlock:
 #    +-------------+---------------+---------+------------+-----------+----------+
 #    | val col cnt | val col types | pos cnt | encodings  | time col  | val col  |


iotdb-client/client-cpp/src/main/SessionBuilder.h
```diff
@@ -80,6 +80,11 @@ class SessionBuilder : public AbstractSessionBuilder {
     }
 
    std::shared_ptr<Session> build() {
+        if (!AbstractSessionBuilder::nodeUrls.empty() &&
+            (AbstractSessionBuilder::host != DEFAULT_HOST ||
+                AbstractSessionBuilder::rpcPort != DEFAULT_RPC_PORT)) {
+            throw IoTDBException("Session builder does not support node urls");
```
jt2594838: The exception message is not precise, maybe: 
Session builder does not support setting node urls and host/rpcPort at the same time.

PR #17040, opened 2026-01-19
Repair the problem that accumulate index used chunkMetadataList

zerolbsony:
Repair the problem that accumulate index used chunkMetadataList;
Rewrite statistics for mem table.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/utils/ResourceByPathUtils.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/utils/ResourceByPathUtils.java
@@ -72,7 +72,6 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
-import java.util.stream.IntStream;
 
 import static org.apache.iotdb.commons.path.AlignedPath.VECTOR_PLACEHOLDER;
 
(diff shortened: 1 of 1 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/utils/ResourceByPathUtils.java
```diff
@@ -351,6 +347,34 @@ public AbstractAlignedTimeSeriesMetadata generateTimeSeriesMetadata(
...
+      List<TSDataType> schemaTypes) {
+    int index = 0;
+    for (IChunkMetadata valueChunkMetadata : alignedChunkMetadata.getValueChunkMetadataList()) {
+      if (valueChunkMetadata == null) {
+        index++;
+        continue;
+      }
+
+      String measurement = valueChunkMetadata.getMeasurementUid();
+      int targetIndex = measurementList.indexOf(measurement);
```
shuwenwei: It is better to use a map

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/utils/ResourceByPathUtils.java
```diff
@@ -351,6 +351,34 @@ public AbstractAlignedTimeSeriesMetadata generateTimeSeriesMetadata(
         : new AlignedTimeSeriesMetadata(timeTimeSeriesMetadata, valueTimeSeriesMetadataList);
   }
 
+  private static void rewriteStatistics(
+      AbstractAlignedChunkMetadata alignedChunkMetadata,
+      Map<String, Integer> measurementMap,
+      List<TSDataType> schemaTypes) {
```
jt2594838: Why not just use Map<String, TSDataType>?

jt2594838 (approved):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/utils/ResourceByPathUtils.java
zerolbsony: > Why not just use Map<String, TSDataType>?

Good idea.

shuwenwei (approved):

PR #17053, opened 2026-01-20
Fix NULL predicate error in table model without TAG columns

Young-Leo:
This fixes a table-model issue where NULL-related predicates fail only
when the table has no TAG columns, while the same predicates work when
TAG columns exist.

### Table without TAG (used to fail)
```sql
CREATE TABLE test_table_no_tag(
  id INT32 FIELD,
  name STRING FIELD,
  value DOUBLE FIELD
);

SELECT * FROM test_table_no_tag WHERE name = null;
```

### Table with TAG (used to work)
```sql
CREATE TABLE test_table_with_tag(
  id STRING TAG,
  name STRING FIELD,
  value DOUBLE FIELD
);

SELECT * FROM test_table_with_tag WHERE name = null;
```

## Fix
Handle casted NULL predicates across predicate visitors so the no-TAG
path no longer fails.


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/analyzer/predicate/ConvertPredicateToFilterVisitor.java
```diff
@@ -371,6 +373,14 @@ protected Filter visitIsNullPredicate(IsNullPredicate node, Context context) {
     throw new IllegalArgumentException("IS NULL cannot be pushed down");
   }
 
+  @Override
+  protected Filter visitCast(Cast node, Context context) {
+    if (node.getExpression() instanceof NullLiteral) {
+      return new FalseLiteralFilter();
+    }
+    return process(node.getExpression(), context);
+  }
```
JackieTien97, line 382: not right, what if I write CAST(NULL AS BOOLEAM) IS NULL ?

JackieTien97 (changes requested):

PR #17065, opened 2026-01-22
feat: show timeseries [order by timseries] clause

xiangmy21:
# feat: show timeseries [order by timseries] clause

Support `ORDER BY TIMESERIES` in `SHOW TIMESERIES` and Optimize `OFFSET` with Subtree Measurement Statistics

---

## Brief

This PR introduces ordering support for `SHOW TIMESERIES` by measurement name (lexicographical) and optimizes the performance of the `OFFSET` clause using subtree measurement statistics.

Currently, `SHOW TIMESERIES` supports the `ORDER BY TIMESERIES` clause.
When the query scope is within a single database, `OFFSET X` no longer requires traversing the first X measurements one by one. Instead, it skips subtrees in the metadata tree based on subtree measurement counts.

Example:

```sql
IoTDB> show timeseries root.dbtest.** order by timeseries desc limit 5 offset 5001
+--------------------------------+-----+-----------+--------+--------+-----------+----+----------+--------+------------------+--------+
|                      Timeseries|Alias|   Database|DataType|Encoding|Compression|Tags|Attributes|Deadband|DeadbandParameters|ViewType|
+--------------------------------+-----+-----------+--------+--------+-----------+----+----------+--------+------------------+--------+
|root.dbtest.plant9.device84.m188| null|root.dbtest|   INT32|     RLE|     SNAPPY|null|      null|    null|              null|    BASE|
|root.dbtest.plant9.device84.m187| null|root.dbtest|   INT32|     RLE|     SNAPPY|null|      null|    null|              null|    BASE|
|root.dbtest.plant9.device84.m186| null|root.dbtest|   INT32|     RLE|     SNAPPY|null|      null|    null|              null|    BASE|
|root.dbtest.plant9.device84.m185| null|root.dbtest|   INT32|     RLE|     SNAPPY|null|      null|    null|              null|    BASE|
|root.dbtest.plant9.device84.m184| null|root.dbtest|   INT32|     RLE|     SNAPPY|null|      null|    null|              null|    BASE|
+--------------------------------+-----+-----------+--------+--------+-----------+----+----------+--------+------------------+--------+
```

* `DESC`: descending order
* `ASC`: ascending order (default if not specified)

---

### 1. Syntax Extension

* Updated the grammar of `SHOW TIMESERIES` to support `orderByTimeseriesClause`.
* Added ordering-related fields to `ShowTimeSeriesStatement` to record sorting requirements.

### 2. Logical Plan Construction

* Introduced a `SortNode` at the top of the logical plan to enable global ordering across partitions.
* For single-partition queries, `LIMIT` and `OFFSET` are pushed down to the metadata tree traversal stage in `LogicalPlanVisitor`.

### 3. Limit/Offset Handling in Multi-Partition Queries

* For multi-partition queries, `OFFSET` cannot be pushed down.
  Instead, `limit' = limit + offset` is used.
* Fixed a bug in the previous implementation:
  when `limit = 0` (which is treated as unlimited in IoTDB), `limit'` was incorrectly pushed down, causing unexpected truncation of results.

### 4. Ordered Traversal in Metadata Tree

* In `MTreeBelowSGMemoryImpl`, the iteration strategy of child nodes in `schemaReader` is overridden under ordering mode.
* Child nodes are traversed in lexicographical order by name.

### 5. Subtree Measurement Count

* Implemented subtree measurement statistics (only for in-memory metadata tree):

  * Added `subtreeMeasurementCount` to `IMemMNode` and related classes.
  * Maintained `subtreeMeasurementCount` along ancestor paths during measurement insertion and deletion.
  * This value is not persisted; it is initialized via a DFS during metadata tree loading.
* During schemaReader construction, `acceptFullMatchedNode` is overridden to skip subtree traversal when `subtreeMeasurementCount < offset`.

---

### Functional Tests
* Added `IoTDBShowTimeseriesOrderByTimeseriesIT` to verify correctness of ordering and offset behavior.
### Performance Evaluation

* Evaluated performance on a single-partition dataset with a typical three-level hierarchy:
  ```
  root.plantA.deviceB.measurementC
  ```
  with approximately 900,000 measurements.
* Query pattern:
  ```sql
  SHOW TIMESERIES root.dbtest.** order by timeseries LIMIT 100 OFFSET X;
  ```
  where `X` ranges from 0 to 900,000 with a step of 5,000.
* Compared the original implementation and the optimized version.
* Results are shown in the figure below: 
image

* Detailed benchmark methodology and results are available at: [[link].](https://apache-iotdb-project.feishu.cn/wiki/P0RcwbUleipgDjkKwhqcOkkynJd?from=from_copylink)

---

## Impact

* Enables deterministic ordering for `SHOW TIMESERIES`.
* Significantly reduces traversal cost for large `OFFSET` values in single-partition scenarios $O(limit+offset)\to O(limit+TreeHeight)$.
* Improves correctness of limit/offset handling in multi-partition queries.
* Introduces minimal overhead to metadata maintenance (in-memory only).

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/impl/mem/MTreeBelowSGMemoryImpl.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/impl/mem/MTreeBelowSGMemoryImpl.java
@@ -1651,17 +1651,10 @@ public class MTreeBelowSGMemoryImpl {
             }
 
             List<IMemMNode> children = new ArrayList<>();
-            if (baseIterator instanceof IMNodeIterator) {
-              IMNodeIterator<IMemMNode> it = (IMNodeIterator<IMemMNode>) baseIterator;
-              while (it.hasNext()) {
-                children.add(it.next());
-              }
-              it.close();
-            } else {
-              while (baseIterator.hasNext()) {
-                children.add(baseIterator.next());
-              }
+            while (baseIterator.hasNext()) {
+              children.add(baseIterator.next());
             }
+            releaseNodeIterator(baseIterator);
 
             children.sort(
                 (a, b) -> {


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/impl/mem/MTreeBelowSGMemoryImpl.java
```diff
@@ -1435,6 +1479,46 @@ public ISchemaReader<ITimeSeriesSchemaInfo> getTimeSeriesReader(
...
+            if (!node.isMeasurement()) {
+              return false;
+            }
+            if (remainingOffset > 0) {
+              // skip this measurement
+              remainingOffset--;
+              return false;
+            }
+            return true;
+          }
```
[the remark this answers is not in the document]
xiangmy21: Exactly! I fixed this issue in #[0177c66](https://github.com/apache/iotdb/pull/17065/commits/0177c6699654a66f38fb92809c28c65551a3419a)
 by disabling pushdown when schemaFilter is present, and by refining the subtree-skipping logic: a subtree is skipped only when its nodes are fully covered by the path pattern.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/impl/mem/MTreeBelowSGMemoryImpl.java
```diff
@@ -192,6 +193,22 @@ public synchronized boolean createSnapshot(final File snapshotDir) {
...
+    IMemMNode current = startNode;
+    while (current != null) {
+      current.setSubtreeMeasurementCount(current.getSubtreeMeasurementCount() + delta);
+      current = current.getParent();
+    }
+  }
+
+  private long getTemplateMeasurementCount(final int templateId) {
+    final Template template = ClusterTemplateManager.getInstance().getTemplate(templateId);
+    return template == null ? 0L : template.getMeasurementNumber();
```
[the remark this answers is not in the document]
xiangmy21: Great catch! I fixed this issue in commit [4027f97](https://github.com/apache/iotdb/pull/17065/commits/4027f9724ad417771b64eff82a53c90180a47297). When a template is modified, the update is broadcast to the corresponding SchemaRegion, and the subtree statistics are updated upward from the nodes activated by the template in the MTree.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/impl/mem/MTreeBelowSGMemoryImpl.java
```diff
@@ -1123,6 +1150,23 @@ public long countPathsUsingTemplate(final PartialPath pathPattern, final int tem
...
 
+  public void rebuildSubtreeMeasurementCount() {
+    rebuildSubtreeMeasurementCountFromNode(rootNode);
+  }
+
+  private long rebuildSubtreeMeasurementCountFromNode(final IMemMNode node) {
+    long count = node.isMeasurement() ? 1L : 0L;
+    final IMNodeIterator<IMemMNode> iterator = store.getChildrenIterator(node);
+    while (iterator.hasNext()) {
+      count += rebuildSubtreeMeasurementCountFromNode(iterator.next());
```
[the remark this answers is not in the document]
xiangmy21: The intermediate logic here is unlikely to throw exceptions, and the iterator will be automatically closed when the function returns. Therefore, we do not modify it for now.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/statement/metadata/ShowStatement.java
```diff
@@ -59,6 +59,13 @@ public void setOffset(long offset) {
     this.offset = offset;
   }
 
+  public long getLimitWithOffset() {
+    if (limit <= 0) {
+      return limit;
+    }
```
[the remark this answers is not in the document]
xiangmy21: Both limit and offset are of type long, so overflow is unlikely in practice.

integration-test/src/test/java/org/apache/iotdb/db/it/schema/IoTDBShowTimeseriesOrderByTimeseriesIT.java
```diff
@@ -0,0 +1,221 @@
...
+    expected.remove("root.db1.devB.x");
+    expected.add("root.db1.devC.m0");
+    expected.add("root.db4.devZ.z");
+    Collections.sort(expected);
+    expected = expected.subList(5, 10); // offset 5 limit 5
+
+    List<String> actual =
+        queryTimeseries("show timeseries root.db*.** order by timeseries offset 5 limit 5");
+    assertEquals(expected, actual);
+  }
```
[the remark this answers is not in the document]
xiangmy21: Good point! I’ve added test cases to cover the interaction between the WHERE clause and OFFSET, as well as runtime template modification scenarios, in commit [c56e155](https://github.com/apache/iotdb/pull/17065/commits/c56e155754074d42eed1396b4304929b1bb0bce5).

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/thrift/impl/DataNodeInternalRPCServiceImpl.java
```diff
@@ -2571,7 +2573,19 @@ public TSStatus updateTemplate(final TUpdateTemplateReq req) {
         ClusterTemplateManager.getInstance().commitTemplatePreSetInfo(req.getTemplateInfo());
         break;
       case UPDATE_TEMPLATE_INFO:
+        Template newTemplate =
+            TemplateInternalRPCUtil.parseUpdateTemplateInfoBytes(
+                ByteBuffer.wrap(req.getTemplateInfo()));
+        Template oldTemplate =
+            ClusterTemplateManager.getInstance().getTemplate(newTemplate.getId());
         ClusterTemplateManager.getInstance().updateTemplateInfo(req.getTemplateInfo());
```
JackieTien97: pass `newTemplate` directly into ClusterTemplateManager.getInstance().updateTemplateInfo instead of parsing it again.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/schema/source/TimeSeriesSchemaSource.java
```diff
@@ -54,6 +54,8 @@ public class TimeSeriesSchemaSource implements ISchemaSource<ITimeSeriesSchemaIn
   private final SchemaFilter schemaFilter;
   private final Map<Integer, Template> templateMap;
   private final boolean needViewDetail;
+  private final boolean orderByTimeseries;
+  private final boolean orderByTimeseriesDesc;
```
JackieTien97: just one `private final Boolean orderByTimeseries` field is enough? and then some comments, like: null for no order by, Boolean.TRUE for asc, Boolean.FALSE for desc.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/metadata/read/TimeSeriesSchemaScanNode.java
```diff
@@ -48,6 +48,11 @@ public class TimeSeriesSchemaScanNode extends SchemaQueryScanNode {
   // if is true, the result will be sorted according to the inserting frequency of the timeseries
   private final boolean orderByHeat;
 
+  // Whether to order result by timeseries full path in this region.
+  private final boolean orderByTimeseries;
+  // When orderByTimeseries is true, whether the ordering is descending.
+  private final boolean orderByTimeseriesDesc;
+
```
JackieTien97: same as above.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/process/ActiveRegionScanMergeNode.java
```diff
@@ -70,6 +70,9 @@ public PlanNode clone() {
 
   @Override
   public List<String> getOutputColumnNames() {
+    if (!children.isEmpty()) {
+      return children.get(0).getOutputColumnNames();
+    }
```
JackieTien97, line 75: why we need to do this change? The origin codes should still take effect.
```
ColumnHeaderConstant.showDevicesColumnHeaders.stream()
            .map(ColumnHeader::getColumnName)
            .collect(Collectors.toList())
```

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/statement/metadata/ShowTimeSeriesStatement.java
```diff
@@ -42,6 +43,9 @@ public class ShowTimeSeriesStatement extends ShowStatement {
   // if is true, the result will be sorted according to the inserting frequency of the time series
   private final boolean orderByHeat;
   private WhereCondition timeCondition;
+  // order by timeseries name
+  private boolean orderByTimeseries;
+  private Ordering nameOrdering = Ordering.ASC;
```
JackieTien97: why it's called nameOrdering? just order is enough?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/read/req/impl/ShowTimeSeriesPlanImpl.java
```diff
@@ -37,6 +37,10 @@ public class ShowTimeSeriesPlanImpl extends AbstractShowSchemaPlanImpl
   private final SchemaFilter schemaFilter;
   private final boolean needViewDetail;
 
+  // order-by-timeseries pushdown flags inside a single SchemaRegion
+  private final boolean orderByTimeseries;
+  private final boolean orderByTimeseriesDesc;
+
```
JackieTien97: same as above

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/thrift/impl/DataNodeInternalRPCServiceImpl.java
xiangmy21: solved in commit [42e1528](https://github.com/apache/iotdb/pull/17065/commits/42e1528e35fd1dab3d2db97abc2dd5de3c710720)

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/schema/source/TimeSeriesSchemaSource.java
xiangmy21: To clarify its function, I'll use an Ordering enum to replace these two boolean. See commit [2121c9a](https://github.com/apache/iotdb/pull/17065/commits/2121c9aec2c5f994f1ed6d2f63c8f9fc5abdfb25)

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/process/ActiveRegionScanMergeNode.java
xiangmy21: The showDevicesColumnHeaders does not contains the 'Timeseries' type we need for sort operation. It contains 'Device' instead of 'Timeseries'. Without this modification the execution of `show timeseries root.sg4.** where time > 0 order by timeseries` will raise error: org.apache.iotdb.jdbc.IoTDBSQLException: 301: Cannot invoke "org.apache.tsfile.enums.TSDataType.ordinal()" because the return value of "java.util.List.get(int)" is null

PR #17069, opened 2026-01-23
Optimize TsTable with Copy-on-Write pattern for thread-safe concurrent access

luoluoyuyu:
## Description
- Change columnSchemaMap, tagColumnIndexMap, and idColumnIndexMap from final to volatile
- Implement Copy-on-Write in executeWrite method:
  1. Create local copies of maps before modification
  2. Execute write operations on local copies
  3. Atomically update class fields after write completes
- Add WriteOperation functional interface to pass map copies to write operations
- This ensures readers see either complete old data or complete new data, avoiding ConcurrentModificationException

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/schema/table/TsTable.java
+++ b/iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/schema/table/TsTable.java
@@ -183,7 +183,6 @@ public class TsTable {
    * the map during copy (e.g., for rename operations) in a single pass.
    *
    * @param columnSchemaMapTransformer transforms columnSchemaMap entries during copy
-   * @param writeOperation the write operation to execute on the new map copies
    */
   private void executeWriteWithTransform(ColumnSchemaMapTransformer columnSchemaMapTransformer) {
     readWriteLock.writeLock().lock();
--- a/iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/schema/table/TsTable.java
+++ b/iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/schema/table/TsTable.java
@@ -183,6 +183,7 @@ public class TsTable {
    * the map during copy (e.g., for rename operations) in a single pass.
    *
    * @param columnSchemaMapTransformer transforms columnSchemaMap entries during copy
+   * @param writeOperation the write operation to execute on the new map copies
    */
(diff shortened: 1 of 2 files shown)


jt2594838 (approved):

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/schema/table/TsTable.java
```diff
@@ -149,23 +150,78 @@ public TsTableColumnSchema getColumnSchema(final String columnName) {
...
+      Map<String, Integer> newTagColumnIndexMap = new HashMap<>(tagColumnIndexMap);
+      Map<String, Integer> newIdColumnIndexMap = new HashMap<>(idColumnIndexMap);
+
+      // Execute write operation on local copies
+      writeOperation.execute(newColumnSchemaMap, newTagColumnIndexMap, newIdColumnIndexMap);
+
+      // After write completes, atomically update the class fields
+      columnSchemaMap = newColumnSchemaMap;
+      tagColumnIndexMap = newTagColumnIndexMap;
+      idColumnIndexMap = newIdColumnIndexMap;
```
jt2594838: There may still introduce inconsistency

jt2594838: And, may only retain tagColumnIndexMap since `id` is a deprecated name

luoluoyuyu: Fixed the issue of Tag column updates during rename.

luoluoyuyu: The  ```private final transient AtomicLong instanceVersion = new AtomicLong(0L);``` and ```private final transient AtomicBoolean isNotWrite = new AtomicBoolean(true);``` define two variables to ensure optimistic read operation; data being written will be downgraded to acquiring data through a read lock.

luoluoyuyu: This PR primarily aims to prevent the HashMap from being modified during the read process, ensuring that unpredictable behavior does not occur during reads.

jt2594838 (approved):

Caideyipi (approved):
LGTM

PR #17109, opened 2026-01-29
Fix timeseries alias display in last for TreeModel

Wei-hao-Li:


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/process/last/AlignedUpdateViewPathLastCacheOperator.java
```diff
@@ -28,11 +28,14 @@
...
 import org.apache.tsfile.utils.TsPrimitiveType;
 
+import java.util.List;
+
 import static com.google.common.base.Preconditions.checkArgument;
 
 public class AlignedUpdateViewPathLastCacheOperator extends AlignedUpdateLastCacheOperator {
 
-  private final String outputViewPath;
+  private final List<String> outputPaths;
```
JackieTien97, line 38: add some comments about this field.

JackieTien97 (changes requested):

PR #17116, opened 2026-01-29
fix: Fix ShowTimeseries estimated offset without limit scenario

luoluoyuyu:
## Description
When ShowTimeseries has offset but no limit, the limit calculation was incorrectly adding offset to 0, causing incorrect behavior. This commit fixes:
1. LogicalPlanVisitor: When limit is 0, keep it as 0 instead of adding offset
2. SchemaReaderLimitOffsetWrapper: When limit is 0, allow all results to pass through (limit == 0 means no limit)

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/integration-test/src/test/java/org/apache/iotdb/db/it/IoTDBSimpleQueryIT.java
+++ b/integration-test/src/test/java/org/apache/iotdb/db/it/IoTDBSimpleQueryIT.java
@@ -827,6 +827,122 @@ public class IoTDBSimpleQueryIT {
   }
 
   @Test
+  public void testShowDevicesWithOffsetOnly() throws SQLException {
+    try (Connection connection = EnvFactory.getEnv().getConnection();
+        Statement statement = connection.createStatement()) {
+
+      statement.execute("CREATE DATABASE root.db1");
+      statement.execute("INSERT INTO root.db1.d0(timestamp, s1) VALUES (1, 1)");
+      statement.execute("INSERT INTO root.db1.d0(timestamp, s2) VALUES (1, 2)");
+      statement.execute("INSERT INTO root.db1.d1(timestamp, s3) VALUES (1, 3)");
+      statement.execute("flush");
+
+      List<String> sg1AllDevices = new ArrayList<>();
+      try (ResultSet resultSet = statement.executeQuery("show devices root.db1.**")) {
+        while (resultSet.next()) {
+          sg1AllDevices.add(resultSet.getString(1));
+        }
+      }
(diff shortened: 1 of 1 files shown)


integration-test/src/test/java/org/apache/iotdb/db/it/IoTDBSimpleQueryIT.java
```diff
@@ -698,6 +698,160 @@ public void testShowTimeseriesWithLimitOffset() throws SQLException {
...
+      // Test Case 1: Single Region scenario (single database)
+      // Create first database and write data, then verify immediately to ensure single Region
+      statement.execute("CREATE DATABASE root.sg1");
+      statement.execute("INSERT INTO root.sg1.d0(timestamp, s1) VALUES (1, 1)");
+      statement.execute("INSERT INTO root.sg1.d0(timestamp, s2) VALUES (1, 2)");
+      statement.execute("INSERT INTO root.sg1.d0(timestamp, s3) VALUES (1, 3)");
+      statement.execute("INSERT INTO root.sg1.d1(timestamp, s1) VALUES (2, 4)");
+      statement.execute("INSERT INTO root.sg1.d1(timestamp, s2) VALUES (2, 5)");
+      statement.execute("INSERT INTO root.sg1.d2(timestamp, s1) VALUES (3, 6)");
+      statement.execute("flush");
```
jt2594838: There is more than one device. Why are you sure that there is only one region?

integration-test/src/test/java/org/apache/iotdb/db/it/IoTDBSimpleQueryIT.java
```diff
@@ -698,6 +698,160 @@ public void testShowTimeseriesWithLimitOffset() throws SQLException {
...
+
+      // Verify single Region scenario - offset only
+      List<String> sg1AllPaths = new ArrayList<>();
+      try (ResultSet resultSet = statement.executeQuery("show timeseries root.sg1.**")) {
+        while (resultSet.next()) {
+          sg1AllPaths.add(resultSet.getString(1));
+        }
+      }
+      Assert.assertTrue("Should have at least 6 timeseries in root.sg1", sg1AllPaths.size() >= 6);
```
jt2594838: Why use >=?

jt2594838:
May also check the offset-limit in normal query?

Caideyipi (approved):
LGTM

jt2594838 (approved):

PR #17127, opened 2026-01-31
Fix blob accumulator query can't skip read blobStatistics caused by isModified don't change to true in the metadata of chunk and pageReader 

zerolbsony:
img_v3_02uf_23cb79bd-4841-4728-9c65-8001ce7dbf6g

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/SchemaUtils.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/SchemaUtils.java
@@ -327,7 +327,7 @@ public class SchemaUtils {
         dataTypeColumnStatisticsClassMap.get(dataType));
   }
 
-  public static boolean canUseStatistics(TSDataType dataType) {
+  public static boolean canUseStatisticsAfterAlter(TSDataType dataType) {
     return !canNotUseStatisticAfterAlterClassSet.contains(dataType);
   }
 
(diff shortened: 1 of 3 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/SchemaUtils.java
```diff
@@ -446,15 +478,15 @@ public static Statistics<?> getNewStatistics(
       case FLOAT:
       case DOUBLE:
       case BOOLEAN:
-        if (targetDataType == TSDataType.STRING) {
+        if (targetDataType == org.apache.tsfile.enums.TSDataType.STRING) {
```
jt2594838: ？

jt2594838 (commented):
Add an associated IT

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/SchemaUtils.java
```diff
@@ -46,12 +47,16 @@
...
+import java.util.Set;
 
 import static org.apache.iotdb.db.queryengine.execution.operator.AggregationUtil.addPartialSuffix;
 
 public class SchemaUtils {
 
   private static final Map<TSDataType, Class> dataTypeColumnClassMap;
+  private static final Map<TSDataType, Class> dataTypeColumnStatisticsClassMap;
+  private static final Set<TSDataType> canNotUseStatisticsClassSet =
+      ImmutableSet.of(TSDataType.STRING, TSDataType.TEXT, TSDataType.BLOB);
```
jt2594838: It is weird to call STRING canNotUseStatistic, because it has all the statistics.
Maybe you should say canNotUseStatisticAfterAlter.

jt2594838 (approved):

PR #17148, opened 2026-02-03
 Fix sort branch release logic of SORT operator

Wei-hao-Li:


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/process/AbstractSortOperator.java
```diff
@@ -240,7 +241,9 @@ private void mergeSort() throws IoTDBException {
         mergeSortHeap.push(mergeSortKey);
       } else {
         noMoreData[readerIndex] = true;
-        sortBufferManager.releaseOneSortBranch();
+        if (sortReaders.get(readerIndex) instanceof FileSpillerReader) {
+          sortBufferManager.releaseOneSortBranch();
+        }
```
JackieTien97: add a method in `SortReader`, let `FileSpillerReader`  and `MemoryReader` implement in different ways, and then add comments for their difference.

PR #17170, opened 2026-02-05
Pipe: preserve exclusions for complex wildcards

VGalaxies:
Preserve exclusion pruning when inclusion uses complex wildcards by falling back to PartialPath overlap/include checks in PatternTrie.
Add regression coverage for inclusion/exclusion pruning in unit and integration tests.

---

**This PR was primarily authored with Codex using gpt-5.2-codex high and then hand-reviewed by me. I AM responsible for every change made in this PR. I aimed to keep it aligned with our goals, though I may have missed minor issues. Please flag anything that feels off, I'll fix it quickly.**

### DIFF
--- a/iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/datastructure/pattern/TreePattern.java
+++ b/iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/datastructure/pattern/TreePattern.java
@@ -890,15 +890,16 @@ public abstract class TreePattern {
 
     /** Adds a path to the Trie. */
     public void add(final PartialPath path) {
-      if (containsComplexWildcard(path)) {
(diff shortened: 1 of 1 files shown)


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/datastructure/pattern/TreePattern.java
```diff
@@ -888,6 +890,11 @@ private static class TrieNode {
 
     /** Adds a path to the Trie. */
     public void add(final PartialPath path) {
+      if (containsComplexWildcard(path)) {
+        complexPatterns.add(path);
+        return;
+      }
+
```
jt2594838: Could you place the judge in the loop below? Otherwise, a path without complex wildcard will be traversed twice.

PR #17180, opened 2026-02-07
[IOTDB-17108] Support inclusive end time syntax [start, end] in GROUP BY clause

### LINKED ISSUE #17108: [Bug] Support inclusive end time syntax ] in GROUP BY clause IoTDB 2.0.5
### Search before asking

### Version

OS: Ubuntu 24.04 LTS
IoTDB: 2.0.5

### Describe the bug and provide the minimal reproduce step

Currently, the `GROUP BY` clause in IoTDB strictly uses the left-closed, right-open `[start, end)` time interval format. Users often need to include the data point exactly at the end time (e.g., `[start, end])`, but the current syntax does not support using `]` to specify an inclusive end time.

When querying time-series data, it is common to want a precise inclusive range (e.g., from `12:00` to `13:00` inclusive). To achieve this now, users must manually extend the end_time by a small fraction (e.g., `+1ms`) or by a full time interval unit, which is unintuitive and can lead to errors or off-by-one queries.

### What did you expect to see?

Support the standard interval syntax `]` to denote an inclusive end time in the `GROUP BY` clause.

Example:
Allow this syntax:
```SQL
SELECT count(s1) FROM root.sg.d1 
GROUP BY ([2026-01-27T12:00:00, 2026-01-27T13:00:00], 1m)
```

### What did you see instead?

Current syntex doesn't allow `]` for end time. I have to use the below syntex:

```SQL
SELECT count(s1) FROM root.sg.d1 
GROUP BY ([2026-01-27T12:00:00, 2026-01-27T13:00:00), 1m)
```
but it excludes the value at 13:00:00

### Anything else?

Current Workaround:
Users currently have to modify the timestamp manually:
```SQL
-- Extended by 1ms to capture the last point
GROUP BY ([2026-01-27T12:00:00, 2026-01-27T13:00:00.001), 1m)
```

PDGGK:
## Summary

This PR adds support for double-closed interval `[start, end]` in GROUP BY time clause, complementing the existing half-open interval `[start, end)`.

### New Syntax
```sql
-- Inclusive end time (NEW)
SELECT count(s1) FROM root.test GROUP BY ([0, 100], 10ms)  -- includes timestamp 100

-- Exclusive end time (existing)
SELECT count(s1) FROM root.test GROUP BY ([0, 100), 10ms)  -- excludes timestamp 100
```

## Changes

| File | Description |
|------|-------------|
| `IoTDBSqlParser.g4` | Add grammar rule for `]` right bracket |
| `GroupByTimeComponent.java` | Add `rightClosed` field |
| `GroupByTimeParameter.java` | Add serialization support for `rightClosed` |
| `ASTVisitor.java` | Parse right bracket type |
| `AggregationUtil.java` | Pass `rightClosed` to factory |
| `TimeRangeIteratorFactory.java` | Update signature (9 args) |
| `AggrWindowIterator.java` | Override `getFinalTimeRange()` for inclusive end |
| `PreAggrWindowIterator.java` | Override `getFinalTimeRange()` for inclusive end |
| `PreAggrWindowWithNaturalMonthIterator.java` | Constructor update |
| `GroupByTimeExpression.java` | Add fields and serialization |
| `ExpressionFactory.java` | Normalize `[start, end]` to `[start, end+1)` |
| `TimeRangeIteratorTest.java` | Fix test signatures for new factory |

## Test Plan

- [x] Unit tests pass (`TimeRangeIteratorTest` - 4 tests)
- [x] Code formatting verified (`mvn spotless:check`)
- [ ] Integration tests (to be added if needed)

## Compatibility

This is a **backward-compatible** change. The existing `[start, end)` syntax continues to work as before. The new `[start, end]` syntax is opt-in.

Fixes #17108

### DIFF
--- a/iotdb-core/antlr/src/main/antlr4/org/apache/iotdb/db/qp/sql/IoTDBSqlParser.g4
+++ b/iotdb-core/antlr/src/main/antlr4/org/apache/iotdb/db/qp/sql/IoTDBSqlParser.g4
@@ -978,6 +978,7 @@ number
 timeRange
     : LS_BRACKET startTime=timeValue COMMA endTime=timeValue RR_BRACKET
     | LR_BRACKET startTime=timeValue COMMA endTime=timeValue RS_BRACKET
+    | LS_BRACKET startTime=timeValue COMMA endTime=timeValue RS_BRACKET
     ;
 
 // ---- Having Clause
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/aggregation/timerangeiterator/AggrWindowIterator.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/aggregation/timerangeiterator/AggrWindowIterator.java
@@ -42,6 +42,7 @@ public class AggrWindowIterator implements ITimeRangeIterator {
 
   private final boolean isAscending;
   private final boolean leftCRightO;
+  private final boolean rightClosed;
 
   private TimeRange curTimeRange;
   private boolean hasCachedTimeRange;
@@ -58,6 +59,7 @@ public class AggrWindowIterator implements ITimeRangeIterator {
       TimeDuration slidingStep,
       boolean isAscending,
       boolean leftCRightO,
+      boolean rightClosed,
       ZoneId zoneId) {
     this.startTime = startTime;
     this.endTime = endTime;
@@ -65,6 +67,7 @@ public class AggrWindowIterator implements ITimeRangeIterator {
     this.slidingStep = slidingStep;
     this.isAscending = isAscending;
     this.leftCRightO = leftCRightO;
+    this.rightClosed = rightClosed;
     this.timeRangeCount = 0;
     this.zoneId = zoneId;
   }
@@ -157,7 +160,7 @@ public class AggrWindowIterator implements ITimeRangeIterator {
       } else {
         retStartTime = curStartTime + slidingStep.nonMonthDuration;
       }
-      // This is an open interval , [0-100)
+      // This is an open interval, [0-100)
       if (retStartTime >= endTime) {
         return false;
       }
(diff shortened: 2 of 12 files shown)


henjoe:
@JackieTien97  Hi I wonder if this is already merged? This feature is actually useful

PDGGK:
Thanks for the interest, @henjoe — this isn't merged yet, it's still awaiting review.

@JackieTien97 the branch is up to date and mergeable whenever you have a chance to look. Quick recap of the change: it adds the inclusive interval syntax `[start, end]` to `GROUP BY` time alongside the existing half-open `[start, end)` form, so the right boundary can be included (e.g. `GROUP BY ([0, 100], 10ms)` produces a final window that includes timestamp 100). Implemented via a `rightClosed` flag threaded through the grammar, `GroupByTimeComponent`/`GroupByTimeParameter`, and the time-range iterators, with `ExpressionFactory` normalizing `[start, end]` to `[start, end+1)` (guarding against `Long.MAX_VALUE` overflow). Tests are in `TimeRangeIteratorTest`. Happy to rebase or address any feedback.

PDGGK:
@henjoe — following up with a more useful answer than the one I gave in June.

**Not merged.** As of master `75eaa40` (2026-07-31) the syntax still isn't there: `timeRange` in `IoTDBSqlParser.g4:1018-1021` has exactly two alternatives, `[start, end)` and `(start, end]`. `[start, end]` doesn't parse. Issue #17108 is still open and untriaged.

**You probably don't need to wait, though.** If the goal is to make sure the final timestamp lands in a bucket, the existing left-closed form already does it for integer timestamps: `GROUP BY ([start, end+1), interval)`. For example `GROUP BY ([0, 101), 10ms)` includes `t=100`, and the Time column stays left-aligned the way you'd expect.

One caveat if you go looking at the other existing form: `(start, end]` also includes the right endpoint, but it is *not* a drop-in replacement, because right-closed windows are reported by their **max** timestamp rather than their min (`ITimeRangeIterator.currentOutputTime()` — "minTime if leftCloseRightOpen, else maxTime"). The repo's own test shows it: `group by((0, 9], 2ms)` returns timestamps `2, 4, 6, 8, 9`, not `0, 2, 4, 6, 8`. So `[start, end+1)` is the closer substitute.

**On the PR**, rather than leaving you waiting on it: it's 642 commits behind master, real CI has never run on the current head (only the greeting check), and — correcting what I wrote in June — `TimeRangeIteratorTest` does not actually cover the new syntax; every call site there passes the new flag as `false`, so `[start, end]` has no test at all. That same comment also said `ExpressionFactory` normalizes `[start, end]` to `[start, end+1)`; that's only true of the pushed-down predicate. The window iterator keeps the un-bumped end time and widens the final window instead, which is a semantic choice nobody has reviewed.

So before asking anyone to spend time on it I'm going to rebase, add an integration test in the style of `IoTDBLeftORightCIT`, and put the design question on `dev@`: given that both `(start, end]` and `[start, end+1)` already exist, is `[start, end]` wanted as sugar, and if so should the last window widen or should an extra window be emitted. Either way you'll get a definite answer instead of an open PR — thanks for pushing on it.

---

henjoe:
Hopefully this one shouldget merged soon.

PR #17181, opened 2026-02-07
[IOTDB-17179] CLI: support automatic reconnection when connection is lost

### LINKED ISSUE #17179: [Feature request] CLI should support automatic reconnection when connection is lost during interactive session
### Search before asking

### Motivation

I am always frustrated when the IoTDB CLI exits as soon as the connection to the server is lost (e.g. server restart, network blip, or idle timeout). I have to reconnect manually, re-enter connection parameters, and lose my session context.

Other IoTDB clients already support reconnection: the Session API (Java), JDBC (via callWithRetryAndReconnect in IoTDBStatement), and the C++/Python clients all retry and reconnect on connection failure. The CLI is the only major client that does not—it holds a single connection for the entire interactive session and exits on any connection-related error. Adding automatic reconnection to the CLI would align its behaviour with the rest of the client stack and improve the experience for long-lived interactive sessions (e.g. over SSH or after server restarts).

### Solution

**Proposed behaviour**

1. **Detect connection loss**  
   When a command fails with a connection-related error (e.g. SQLException caused by connection refused, timeout, or closed connection), do not exit immediately.

2. **Attempt reconnection**  
   Try to establish a new connection using the same parameters (host, port, user, password, and options) that were used at login. Reuse the same logic as the initial connection (e.g. DriverManager.getConnection with the same URL and properties).

3. **Retry the command**  
   After a successful reconnection, retry the same user command (the current line/statement that failed) with the new connection, with a bounded number of retries (e.g. 3) and optional backoff.

4. **User feedback**  
   On successful reconnection, print a short message such as: "Connection lost. Reconnected. Retrying command." and continue the interactive loop with the new connection.

5. **Failure handling**  
   If reconnection fails after all retries, exit with a clear error message, e.g. "Could not reconnect after N attempts. Please check that the server is running and try again."

**Relevant code**

- `iotdb-client/cli/src/main/java/org/apache/iotdb/cli/Cli.java`:  
  - `receiveCommands()` opens one connection and runs a `while (true)` loop; any SQLException is caught at the top level and the process exits (lines 215–217).  
  - Change: catch connection-related failures inside the loop (or around `processCommand`), attempt reconnection, then retry the command or continue with the new connection.

- `iotdb-client/jdbc`:  
  - `IoTDBConnection.reconnect()` and `IoTDBStatement.callWithRetryAndReconnect()` can be used as reference for retry/reconnect behaviour.

**Optional**

- Make retry count and interval configurable (e.g. CLI options or environment variables) for consistency with Session/JDBC.

### Alternatives

1. **No retry, only reconnect**  
   On connection loss, attempt reconnection and, on success, continue the session without retrying the failed command (user re-enters it). Simpler to implement but less convenient than retrying the last command.

2. **Prompt user**  
   On connection loss, prompt: "Connection lost. Reconnect? (y/n)". If yes, reconnect and optionally retry the command. Gives control but adds friction for automated or scripted use.

3. **Leave as-is**  
   Keep current behaviour (exit on connection loss). This preserves simplicity but leaves the CLI as the only client without reconnection support and worsens UX for long sessions and unstable networks.

- [x] I'm willing to submit a PR!

miantalha45:
## Description

Add automatic reconnection to the IoTDB CLI when the connection to the server is lost during an interactive session (e.g. server restart, network blip, or idle timeout). The CLI no longer exits immediately on connection-related errors; it attempts to reconnect with the same parameters and retries the failed command, aligning behavior with the Session API, JDBC, and C++/Python clients.

### Content1 — Detection and reconnection flow

- **Detection**: Connection loss is detected when a command fails with a connection-related `SQLException`. We treat an exception as connection-related if its message (or cause message, lowercased) contains any of: `connection`, `refused`, `timeout`, `closed`, `reset`, `network`, `broken pipe`. This logic lives in `AbstractCli.isConnectionRelated(SQLException)` and `matchesConnectionFailure(String)` so it can be shared and reused.
- **Reconnection**: On such a failure, the CLI closes the current connection and opens a new one using the same parameters (host, port, user, password, and options) via `DriverManager.getConnection` and the existing `info` properties. Helper methods `openConnection()`, `setupConnection()`, and `closeConnectionQuietly()` in `Cli` encapsulate open/setup/close so the main loop stays clear.
- **Retry**: After a successful reconnection, the same user command (the current line that failed) is retried with the new connection. We retry reconnection up to **3** times with a **1 s** delay between attempts (no delay before the first attempt). Constants `RECONNECT_RETRY_NUM` and `RECONNECT_RETRY_INTERVAL_MS` in `Cli` control this; they are not yet user-configurable.
- **Feedback**: On successful reconnection we print: `Connection lost. Reconnected. Retrying command.` If all reconnection attempts fail we print: `IoTDB: Could not reconnect after 3 attempts. Please check that the server is running and try again.` and exit with error code.

### Content2 — Class and method organization

- **AbstractCli**: Added `isConnectionRelated(SQLException)` (package-private static) and `matchesConnectionFailure(String)` (private static) for shared detection. In `executeQuery`, `setTimeZone`, and `showTimeZone`, we catch `SQLException` (or `Exception` where the API does not throw `SQLException`) and rethrow when `isConnectionRelated(e)`; otherwise we keep the existing "print error and return error code" behavior. `handleInputCmd` and `processCommand` now declare `throws SQLException` so connection failures propagate to the CLI loop instead of being swallowed.
- **Cli**: Introduced `ReadLineResult` (inner class with `stop`, `failedCommand`) and factory methods `continueLoop()`, `stopLoop()`, `reconnectAndRetry(String)` so the read-eval loop can signal "continue", "exit", or "reconnect and retry this command". `receiveCommands()` no longer uses try-with-resources for the connection; it holds the connection in a variable, and when `readerReadLine()` returns a result with `failedCommand != null`, it runs the reconnect loop (close → retry open/setup → print message → retry command). `readerReadLine()` wraps `processCommand()` in a try-catch; on connection-related `SQLException` it returns `reconnectAndRetry(s)` with the current line; on other `SQLException` it prints and returns `stopLoop()`.
- **AbstractCliTest**: `testHandleInputInputCmd()` now declares `throws SQLException` and imports `java.sql.SQLException` so it compiles with the updated `handleInputCmd` signature.

### Content3 — Corner cases and alternatives

- **Corner cases**: If reconnection succeeds but the retried command fails again with a connection-related error, the outer loop will see another `reconnectAndRetry` and run the same reconnect/retry flow again (each time with up to 3 reconnect attempts). Non-connection `SQLException`s still print the error and stop the loop (exit) as before. Interrupt and EOF handling in `readerReadLine()` are unchanged.
- **Session/statement errors after reconnect**: If the server returns an error such as "StatementId doesn't exist in this session" (e.g. after a server stop/start), the retried command or the next user command can fail with that instead of a connection error. We treat such exceptions as session/statement state errors via `isSessionOrStatementError()` in AbstractCli and show: *"Reconnected, but the previous command could not be completed. Please run your command again."* so the user is not shown the raw exception. This handling is applied both in the reconnect-retry path in Cli and in AbstractCli.executeQuery for the normal command path.
- **Alternatives considered**: (1) Reconnect without retrying the failed command—simpler but worse UX. (2) Prompt "Reconnect? (y/n)"—gives control but adds friction and is less script-friendly. (3) Leave current behavior—rejected to align CLI with other clients and improve long-lived session UX.

<hr>

This PR has:
- [x] been self-reviewed.
    - [x] concurrent read
    - [x] concurrent write

<hr>

- `org.apache.iotdb.cli.AbstractCli` — `isConnectionRelated`, `matchesConnectionFailure`, `isSessionOrStatementError`, `matchesSessionOrStatementFailure`; rethrow connection-related `SQLException` in `executeQuery`, `setTimeZone`, `showTimeZone`; in `executeQuery` show friendly message for session/statement errors; `handleInputCmd`, `processCommand` now `throws SQLException`
- `org.apache.iotdb.cli.Cli` — `ReadLineResult`, `openConnection()`, `setupConnection()`, `closeConnectionQuietly()`; refactored `receiveCommands()` and `readerReadLine()` for reconnect-and-retry flow; handle session/statement error in reconnect-retry path
- `org.apache.iotdb.cli.AbstractCliTest` — `testHandleInputInputCmd()` updated for `throws SQLException`

Closes #17179


CritasWang:
@miantalha45 Copilot has identified some interesting issues, and there are conflicts with the current PR.

miantalha45:
Sure. I am on it

miantalha45:
@CritasWang Please have a look on this PR now.

miantalha45:
@CritasWang Any update on this PR?

PR #17193, opened 2026-02-10
Fixed concurrency issues caused by write and flush sorting during query execution

shuwenwei:
## Description
When writing and flushing data, the internal TVList sorting mechanism could conflict with active queries. 
✅ What’s Changed
✔️ Concurrency control enhancements
	•	The write and flush sort logic has been revised to safely handle concurrent queries by introducing synchronization and controlled cloning of TVLists under contention.
	•	The implementation ensures that sorting operations do not corrupt or disrupt query access on the same data structures.

✔️ Memtable & TVList adjustments
	•	Modified internal MemChunk and TVList management to better isolate sort state changes from active queries.
	•	New clone-on-flush behaviour ensures that a working TVList can be safely sorted even when there are concurrent readers.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/memtable/AbstractWritableMemChunk.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/memtable/AbstractWritableMemChunk.java
@@ -236,6 +236,7 @@ public abstract class AbstractWritableMemChunk implements IWritableMemChunk {
     workingListForFlush.sort();
   }
 
+  @Override
   public void releaseTemporaryTvListForFlush() {
     workingListForFlush = null;
   }


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/memtable/AbstractWritableMemChunk.java
```diff
@@ -198,7 +200,41 @@ public abstract void writeAlignedTablet(
   public abstract IMeasurementSchema getSchema();
 
   @Override
-  public abstract void sortTvListForFlush();
+  public synchronized void sortTvListForFlush() {
```
JackieTien97: why we need `synchronized` here? Will sort for one memtable be parallel?

JackieTien97: including the encode method

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/memtable/AlignedWritableMemChunk.java
```diff
@@ -557,6 +550,8 @@ public void encodeWorkingAlignedTVList(
       long maxNumberOfPointsInChunk,
       int maxNumberOfPointsInPage) {
     BitMap allValueColDeletedMap;
+    AlignedTVList list = (AlignedTVList) listForFlushSort;
```
JackieTien97: Change all used places to listForFlushSort instead of hiding the `list` field.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/datastructure/AlignedTVList.java
```diff
@@ -165,6 +165,18 @@ public TVList getTvListByColumnIndex(
     return alignedTvList;
   }
 
+  @Override
+  public AlignedTVList cloneForFlushSort() {
```
JackieTien97:
```suggestion
  public synchronized AlignedTVList cloneForFlushSort() {
```

JackieTien97 (changes requested):

PR #17205, opened 2026-02-13
Support select locally for data consistency check

jt2594838:
Add a query option `LOCALLY` that may allow only selecting data from the currently connected node.

This is designed for the data consistency check for developers and maintainers mainly.

Currently, `LOCALLY` only takes effect on raw data query.

Refer to the newly added ITs for examples.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/analyzer/StatementAnalyzer.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/analyzer/StatementAnalyzer.java
@@ -860,6 +860,8 @@ public class StatementAnalyzer {
     @Override
     protected Scope visitQuery(Query node, Optional<Scope> context) {
       analysis.setQuery(true);
+      analysis.setLocalQuery(node.getQueryBody().isLocalQuery());
+
       Scope withScope = analyzeWith(node, context);
       hasFillInParentScope = node.getFill().isPresent() || hasFillInParentScope;
       Scope queryBodyScope = process(node.getQueryBody(), withScope);
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/planner/distribute/TableDistributedPlanGenerator.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/planner/distribute/TableDistributedPlanGenerator.java
@@ -29,6 +29,7 @@ import org.apache.iotdb.commons.partition.SchemaPartition;
 import org.apache.iotdb.commons.schema.table.TsTable;
 import org.apache.iotdb.commons.schema.table.column.TsTableColumnCategory;
 import org.apache.iotdb.commons.utils.TimePartitionUtils;
+import org.apache.iotdb.db.conf.IoTDBDescriptor;
 import org.apache.iotdb.db.exception.sql.SemanticException;
 import org.apache.iotdb.db.queryengine.common.MPPQueryContext;
 import org.apache.iotdb.db.queryengine.common.QueryId;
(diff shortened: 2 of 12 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/planner/distribute/TableDistributedPlanGenerator.java
```diff
@@ -810,6 +811,19 @@ private List<PlanNode> constructDeviceTableScanByRegionReplicaSet(
...
+        if (analysis.isLocalQuery()) {
+          // only query this node in local query mode
+          int dataNodeId = IoTDBDescriptor.getInstance().getConfig().getDataNodeId();
+          boolean containsThisNode =
+              regionReplicaSet.dataNodeLocations.stream()
+                  .anyMatch(dn -> dn.getDataNodeId() == dataNodeId);
+          if (!containsThisNode) {
+            continue;
+          } else {
+            regionReplicaSet.dataNodeLocations.removeIf(dn -> dn.getDataNodeId() != dataNodeId);
```
Caideyipi, line 823: This mutates the `TRegionReplicaSet` returned from `DataPartition`/`PartitionCache`. Those objects are shared by the query analysis/route cache, so after one local query the replica set can remain truncated for later normal queries. Please build a new `TRegionReplicaSet(regionReplicaSet.getRegionId(), localLocations)` and set/use that instead of changing `dataNodeLocations` in place.

Caideyipi (commented):
Correctness issues found in the SELECT LOCALLY implementation.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/analyzer/StatementAnalyzer.java
```diff
@@ -860,6 +860,8 @@ protected Scope visitExplainAnalyze(ExplainAnalyze node, Optional<Scope> context
     @Override
     protected Scope visitQuery(Query node, Optional<Scope> context) {
       analysis.setQuery(true);
+      analysis.setLocalQuery(node.getQueryBody().isLocalQuery());
```
Caideyipi, line 863: `Analysis` is shared across the whole statement, but `visitQuery` is also called for subqueries/CTEs. A nested SELECT will overwrite this flag, so `SELECT LOCALLY ... FROM (SELECT ...)` can lose local mode, and `SELECT ... FROM (SELECT LOCALLY ...)` can make outer scans local. The local marker needs to be scoped to the specific query/table scan or saved/restored when analyzing nested queries.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/planner/distribute/TableDistributedPlanGenerator.java
```diff
@@ -810,6 +811,19 @@ private List<PlanNode> constructDeviceTableScanByRegionReplicaSet(
         context.deviceCrossRegion = true;
       }
       for (final TRegionReplicaSet regionReplicaSet : regionReplicaSets) {
+        if (analysis.isLocalQuery()) {
```
Caideyipi, line 814: This only runs in `constructDeviceTableScanByRegionReplicaSet`. When `visitDeviceTableScan` takes the `context.isPushDownGrouping()` branch, `constructDeviceTableScanByTags` never applies the local filter, so local queries with grouping/tag pushdown can still schedule remote replicas. The local restriction should be applied before both split paths or the unsupported shapes should be rejected.

iotdb-core/relational-grammar/src/main/antlr4/org/apache/iotdb/db/relational/grammar/sql/RelationalSql.g4
```diff
@@ -1669,6 +1669,7 @@ LISTAGG: 'LISTAGG';
 LOAD: 'LOAD';
 LOADED: 'LOADED';
```
Caideyipi, line 1670: Adding `LOCALLY` as a lexer token makes unquoted `locally` stop parsing as an identifier unless it is also included in `nonReserved`. That breaks existing table/column names such as `SELECT locally FROM t` or `CREATE TABLE locally (...)`. Please keep it non-reserved if this is only a SELECT modifier.

PR #17229, opened 2026-02-28
Add project helpful info readme for AI Coding model

JackieTien97:
This pull request adds documentation files to clarify project guidelines, coding standards, and build/test procedures for contributors. The main changes are the introduction of `AGENTS.md` with comprehensive rules and `CLAUDE.md` as a pointer to those rules.

Documentation additions:

* Added `AGENTS.md`, which provides a detailed overview of the project, Java code style guide, naming conventions, code structure requirements, license header templates, verification and build commands, and instructions for running unit and integration tests.
* Added `CLAUDE.md`, which includes the standard license header and directs readers to `AGENTS.md` for all project guidelines and instructions.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -140,7 +140,7 @@ mvn checkstyle:check -P with-integration-tests
 # Full build (produces distribution in distribution/target/)
 mvn clean package -pl distribution -am -DskipTests
 
-# Build server only
+# Build all modules (skip tests)
 mvn clean package -DskipTests
 
 # Build CLI only
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -158,13 +158,13 @@ mvn spotless:apply -P with-integration-tests
 
 ```bash
 # Unit tests for a specific module
-mvn test -pl iotdb-core/datanode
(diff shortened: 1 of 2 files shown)


chrisdutz:
Please double-check if formally supporting agentic frameworks is not in contrast to current Apache policy.

HTHou:
Actually, we are not the first Apache project to add these files.

https://github.com/apache/spark
https://github.com/apache/shardingsphere

chrisdutz:
Yeah ... so having had a look at the current policy, it should be ok ... it really depends on how AI is used by the committers/pmc members. 

But I'd like to point out: Just because other projects do somethings is not always proof that something's ok. Usually especially with new things arising, some projects just go ahead and then other projects treat that as proof, that it's ok and it's sometimes hard to catch this from a foundations perspective.

If really adopting and embracing AI, possibly thinking about adopting SpecKit would be a good idea? (https://github.com/github/spec-kit) ... because the things you are telling the AI actually are also interesing for humans. So going a bit down the path of spec-driven development would have benefits for human contributors as well as AI ones.

PR #17257, opened 2026-03-04
Add 3C3D IoTConsensusV2 integration tests

Pengzna:
Add IT for IoTConsensusV2 on a 3 ConfigNode + 3 DataNode cluster.

Tests:

1. test3C3DWriteFlushAndQuery: Start 3C3D cluster, write data, flush, and verify data via query.
2. testReplicaConsistencyAfterLeaderStop: With replication factor 2, stop the leader DataNode and verify the former follower (new leader) serves the same data after failover.

Implementation details:

1. Config: IoTConsensusV2 batch/stream mode, DataReplicationFactor=2, SchemaReplicationFactor=3.
2. Uses getConnection(followerNode) to query the specific follower instead of a random available node.
3. Extends IoTDBRegionOperationReliabilityITFramework for getDataRegionMapWithLeader and cluster setup/teardown.
4. Uses getString() in verifyDataConsistency for IoTDB JDBC compatibility.

### DIFF
--- a/integration-test/src/test/java/org/apache/iotdb/db/it/iotconsensusv2/IoTDBIoTConsensusV23C3DBasicITBase.java
+++ b/integration-test/src/test/java/org/apache/iotdb/db/it/iotconsensusv2/IoTDBIoTConsensusV23C3DBasicITBase.java
@@ -240,10 +240,7 @@ public abstract class IoTDBIoTConsensusV23C3DBasicITBase
                 double syncLag = Double.parseDouble(matcher.group(1));
                 LOGGER.debug("Found syncLag metric value: {}", syncLag);
                 Assert.assertEquals(
-                    "Consensus pipe syncLag should be 0.0 but was " + syncLag,
-                    0.0,
-                    syncLag,
-                    0.001);
+                    "Consensus pipe syncLag should be 0.0 but was " + syncLag, 0.0, syncLag, 0.001);
               }
               Assert.assertTrue(
                   "No pipe_consensus syncLag metric found in leader DataNode metrics", found);


Pengzna:
TODO: move @DailyIT to batch and stream files instead of base file

integration-test/src/test/java/org/apache/iotdb/db/it/iotconsensusv2/IoTDBIoTConsensusV23C3DBasicITBase.java
```diff
@@ -0,0 +1,252 @@
...
+      LOGGER.info("Writing data to 3C3D cluster (mode: {})...", getIoTConsensusV2Mode());
+      statement.execute(INSERTION1);
+      statement.execute(INSERTION2);
+      statement.execute(INSERTION3);
+      statement.execute(FLUSH_COMMAND);
+
+      verifyDataConsistency(statement);
+
+      LOGGER.info("Sleeping 2 seconds to wait replicate ...");
+      Thread.sleep(1000 * 2);
```
jt2594838: This is very unstable in the test environment; it is better to check the data directory of each node to ensure that two nodes have the associated TsFiles.

integration-test/src/test/java/org/apache/iotdb/db/it/iotconsensusv2/IoTDBIoTConsensusV23C3DBasicITBase.java
```diff
@@ -0,0 +1,252 @@
...
+          EnvFactory.getEnv()
+              .dataNodeIdToWrapper(leaderDataNodeId)
+              .orElseThrow(() -> new AssertionError("DataNode not found in cluster"));
+      LOGGER.info(
+          "Stopping leader DataNode {} (region {}) for replica consistency test",
+          leaderDataNodeId,
+          targetRegionId);
+
+      leaderNode.stopForcibly();
+      Assert.assertFalse("Leader should be stopped", leaderNode.isAlive());
```
jt2594838, line 181: Killing the node may result in losing some unflushed data.
Although you may have called FLUSH previously, the data may not have been synchronized then, so the FLUSH command may not work on the follower.
You should first wait until the unsealed TsFile appears on the follower, and then FLUSH.

Pengzna: Good Catch, now I check the syncLag(progress of replicate) to ensure the leader has finished the replication.

PR #17292, opened 2026-03-12
Add statistical type aggregate functions, including autocorrelation, skewness, and linear regression

CoollZzz:
This pull request adds integration tests for new statistical aggregation functions in IoTDB, specifically focusing on correlation, covariance, and related statistical measures. It introduces comprehensive test coverage for these functions, updates utility constants, and expands the list of recognized aggregation functions.

**New aggregation function support and test coverage:**

* Added new aggregation functions to `BuiltinAggregationFunctionEnum`: `CORR`, `COVAR_POP`, `COVAR_SAMP`, `REGR_SLOPE`, `REGR_INTERCEPT`, `SKEWNESS`, and `KURTOSIS`.
* Added corresponding string helper methods for these functions in `TestConstant` (e.g., `corr`, `covarPop`, `covarSamp`, etc.), enabling easier test query construction.

**Integration tests for statistical functions:**

* Added `IoTDBCorrelationIT` integration test class, which tests the `corr` (correlation) function across various scenarios: type checking, different data types, alignment by device, use in HAVING clauses, multi-device queries, group by level, and sliding window queries.
* Added `IoTDBCovarianceIT` integration test class, which tests `covar_pop` and `covar_samp` (covariance) functions, including type and arity validation, cross-type queries, alignment by device, HAVING clause usage, group by level, and sliding window queries.This pull request adds support for several advanced statistical aggregation functions to the IoTDB query engine, including correlation, covariance, regression, skewness, and kurtosis. It introduces new accumulator classes to implement these functions and updates the relevant factory and enum classes to register and handle them appropriately.

**New statistical aggregation functions:**

* Added new aggregation function types: `CORR`, `COVAR_POP`, `COVAR_SAMP`, `REGR_SLOPE`, `REGR_INTERCEPT`, `SKEWNESS`, and `KURTOSIS` to `BuiltinAggregationFunctionEnum` for recognition in the system.
* Updated `AccumulatorFactory` to recognize these new functions as multi-input or single-input aggregations and to instantiate the appropriate new accumulator classes when requested. [[1]](diffhunk://#diff-97211aaf9f3d32085a6e45b00415b62ef8c8ce3ea0513a7ff42fc0560608f43cR72-R76) [[2]](diffhunk://#diff-97211aaf9f3d32085a6e45b00415b62ef8c8ce3ea0513a7ff42fc0560608f43cR92-R116) [[3]](diffhunk://#diff-97211aaf9f3d32085a6e45b00415b62ef8c8ce3ea0513a7ff42fc0560608f43cR173-R178)

**Implementation of new accumulator classes:**

* Introduced `CorrelationAccumulator` for correlation and covariance calculations, supporting both population and sample variants.
* Introduced `RegressionAccumulator` for regression slope and intercept calculations.
* Introduced `CentralMomentAccumulator` for skewness and kurtosis calculations.

**Integration with sliding window aggregators:**

* Updated `SlidingWindowAggregatorFactory` to support the new statistical functions within sliding window queries.This pull request introduces support for several advanced statistical aggregation functions in the IoTDB query engine, including correlation, covariance, regression, skewness, and kurtosis. It adds new accumulator implementations for these functions and integrates them into the aggregation and sliding window frameworks.

The most important changes are:

**New statistical accumulator implementations:**

* Added `CentralMomentAccumulator` for computing skewness and kurtosis. (`iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/aggregation/CentralMomentAccumulator.java`)
* Added `CorrelationAccumulator` for correlation and covariance calculations. (`iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/aggregation/CorrelationAccumulator.java`)
* Added `RegressionAccumulator` for regression slope and intercept calculations. (`iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/aggregation/RegressionAccumulator.java`)

**Integration with aggregation framework:**

* Updated `AccumulatorFactory` to support the new aggregation functions, including logic for multi-input and single-input accumulators, and to instantiate the new accumulator classes as appropriate. (`iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/aggregation/AccumulatorFactory.java`) [[1]](diffhunk://#diff-97211aaf9f3d32085a6e45b00415b62ef8c8ce3ea0513a7ff42fc0560608f43cR72-R76) [[2]](diffhunk://#diff-97211aaf9f3d32085a6e45b00415b62ef8c8ce3ea0513a7ff42fc0560608f43cR92-R116) [[3]](diffhunk://#diff-97211aaf9f3d32085a6e45b00415b62ef8c8ce3ea0513a7ff42fc0560608f43cR173-R178)

**Sliding window support:**

* Modified `SlidingWindowAggregatorFactory` to support the new statistical functions within sliding window aggregations. (`iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/aggregation/slidingwindow/SlidingWindowAggregatorFactory.java`)


integration-test/src/test/java/org/apache/iotdb/relational/it/query/recent/IoTDBTableAggregationIT.java
```diff
@@ -110,6 +110,28 @@ public class IoTDBTableAggregationIT {
         "INSERT INTO table1(time,province,city,region,device_id,color,type,s3,s5,s7,s9,s10) values (2024-09-24T06:15:30.000+00:00,'beijing','beijing','haidian','d16','yellow','BBBBBBBBBBBBBBBB',30.0,true,'beijing_haidian_yellow_B_d16_30',2024-09-24T06:15:30.000+00:00,'2024-09-24')",
         "INSERT INTO table1(time,province,city,region,device_id,color,type,s2,s9) values (2024-09-24T06:15:40.000+00:00,'beijing','beijing','haidian','d16','yellow','BBBBBBBBBBBBBBBB',40000,2024-09-24T06:15:40.000+00:00)",
         "INSERT INTO table1(time,province,city,region,device_id,color,type,s1,s4,s6,s8,s9) values (2024-09-24T06:15:55.000+00:00,'beijing','beijing','haidian','d16','yellow','BBBBBBBBBBBBBBBB',55,55.0,'beijing_haidian_yellow_B_d16_55',X'cafebabe55',2024-09-24T06:15:55.000+00:00)",
+        // stat_table for statistical aggregation function tests (CORR, COVAR_POP, COVAR_SAMP,
+        // REGR_SLOPE, REGR_INTERCEPT, KURTOSIS, SKEWNESS)
+        "CREATE TABLE stat_table(device_id STRING TAG, s1 INT32 FIELD, s2 INT64 FIELD, s3 FLOAT FIELD, s4 DOUBLE FIELD, s5 BOOLEAN FIELD, s6 TEXT FIELD)",
```
JackieTien97: include timestamp type

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/TypeInferenceUtils.java
```diff
@@ -190,7 +197,22 @@ private static void verifyIsAggregationDataTypeMatched(String aggrFuncName, TSDa
...
+          return;
+        }
+        throw new SemanticException(
+            "Aggregate functions [SKEWNESS, KURTOSIS] only support "
+                + "numeric data types [INT32, INT64, FLOAT, DOUBLE, TIMESTAMP]");
+      case SqlConstant.CORR:
+      case SqlConstant.COVAR_POP:
+      case SqlConstant.COVAR_SAMP:
+      case SqlConstant.REGR_SLOPE:
+      case SqlConstant.REGR_INTERCEPT:
```
JackieTien97, line 220: these functions should be same as `SKEWNESS` and `KURTOSIS`, only support [INT32, INT64, FLOAT, DOUBLE, TIMESTAMP]?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/aggregation/CentralMomentAccumulator.java
```diff
@@ -0,0 +1,294 @@
...
+  }
+
+  @Override
+  public void outputIntermediate(ColumnBuilder[] columnBuilders) {
+    checkArgument(columnBuilders.length == 1, "partialResult should be 1");
+    if (count == 0) {
+      columnBuilders[0].appendNull();
+    } else {
+
+      byte[] bytes = new byte[40];
```
JackieTien97: use Long.BYTES + 4 * Double.BYTES instead of magic number 40.


checking all the related places.

JackieTien97 (changes requested):

integration-test/src/test/java/org/apache/iotdb/db/it/aggregation/IoTDBCorrelationIT.java
```diff
@@ -0,0 +1,188 @@
...
+import static org.apache.iotdb.db.it.utils.TestUtils.prepareData;
+import static org.apache.iotdb.db.it.utils.TestUtils.resultSetEqualTest;
+import static org.apache.iotdb.itbase.constant.TestConstant.DEVICE;
+import static org.apache.iotdb.itbase.constant.TestConstant.TIMESTAMP_STR;
+import static org.apache.iotdb.itbase.constant.TestConstant.corr;
+import static org.junit.Assert.fail;
+
+@RunWith(IoTDBTestRunner.class)
+@Category({LocalStandaloneIT.class, ClusterIT.class})
+public class IoTDBCorrelationIT {
```
JackieTien97: merge all these test IT class into one IT class to avoid start and stop cluster many times, you can clean up data after each method finished and prepare data at the very beginning of each method.

integration-test/src/test/java/org/apache/iotdb/db/it/aggregation/IoTDBCorrelationIT.java
```diff
@@ -0,0 +1,188 @@
...
+
+import static org.apache.iotdb.db.it.utils.TestUtils.prepareData;
+import static org.apache.iotdb.db.it.utils.TestUtils.resultSetEqualTest;
+import static org.apache.iotdb.itbase.constant.TestConstant.DEVICE;
+import static org.apache.iotdb.itbase.constant.TestConstant.TIMESTAMP_STR;
+import static org.apache.iotdb.itbase.constant.TestConstant.corr;
+import static org.junit.Assert.fail;
+
+@RunWith(IoTDBTestRunner.class)
+@Category({LocalStandaloneIT.class, ClusterIT.class})
```
JackieTien97:
```suggestion
@Category({TableLocalStandaloneIT.class, TableClusterIT.class})
```

JackieTien97 (changes requested):

integration-test/src/test/java/org/apache/iotdb/db/it/aggregation/IoTDBCorrelationIT.java
CoollZzz: IoTDBCorrelationIT is a tree-model integration test, so it should keep @Category({LocalStandaloneIT.class, ClusterIT.class})

iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/execution/aggregation/CentralMomentAccumulator.java
```diff
@@ -0,0 +1,298 @@
...
+      columnBuilder.writeDouble(result);
+    } else {
+      if (count < 4) {
+        columnBuilder.appendNull();
+      } else {
+
+        double variance = m2 / (count - 1);
+        double term1 =
+            (count * (count + 1) * m4)
+                / ((count - 1) * (count - 2) * (count - 3) * variance * variance);
```
JackieTien97: **Likely `long` overflow in the kurtosis denominator.**

`(count - 1) * (count - 2) * (count - 3)` is evaluated as `long` (all three operands are `long`) *before* it is promoted to `double`. For `count` greater than ~2.1M this product exceeds `Long.MAX_VALUE`, and around `count` ~2.5M it even **flips sign**, so kurtosis comes out wrong (often negative-signed) instead of correct. Million-row groups are routine in a time-series DB, so this is reachable in practice.

The numerator `count * (count + 1)` and `term2`'s `(count - 2) * (count - 3)` have the same problem at larger thresholds.

Suggest forcing `double` early:
```java
double n = count;
double term1 = (n * (n + 1) * m4) / ((n - 1) * (n - 2) * (n - 3) * variance * variance);
double term2 = (3 * (n - 1) * (n - 1)) / ((n - 2) * (n - 3));
```
The same fix is needed in `TableCentralMomentAccumulator` and `GroupedCentralMomentAccumulator`. A test over ~3M non-constant rows would lock in the regression.

iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/execution/operator/source/relational/aggregation/TableCentralMomentAccumulator.java
```diff
@@ -0,0 +1,298 @@
...
+      double result = Math.sqrt((double) count) * m3 / Math.pow(m2, 1.5);
+      columnBuilder.writeDouble(result);
+    } else {
+      if (count < 4) {
+        columnBuilder.appendNull();
+      } else {
+        double variance = m2 / (count - 1);
+        double term1 =
+            (count * (count + 1) * m4)
+                / ((count - 1) * (count - 2) * (count - 3) * variance * variance);
```
JackieTien97: Same `long`-overflow in the kurtosis denominator as in `CentralMomentAccumulator` (see that comment): `(count - 1) * (count - 2) * (count - 3)` overflows `long` for `count` greater than ~2.1M. Please compute the factorial-style terms in `double`.

iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/execution/operator/source/relational/aggregation/grouped/GroupedCentralMomentAccumulator.java
```diff
@@ -0,0 +1,269 @@
...
+      columnBuilder.writeDouble(result);
+    } else {
+      if (count < 4) {
+        columnBuilder.appendNull();
+      } else {
+        double m4 = m4s.get(groupId);
+        double variance = m2 / (count - 1);
+        double term1 =
+            (count * (count + 1) * m4)
+                / ((count - 1) * (count - 2) * (count - 3) * variance * variance);
```
JackieTien97: Same `long`-overflow in the kurtosis denominator here too — `(count - 1) * (count - 2) * (count - 3)` is a `long` product that overflows for `count` greater than ~2.1M. Compute in `double`.

iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/execution/aggregation/CentralMomentAccumulator.java
```diff
@@ -0,0 +1,298 @@
...
+      columnBuilder.appendNull();
+      return;
+    }
+
+    if (momentType == MomentType.SKEWNESS) {
+      if (count < 3) {
+        columnBuilder.appendNull();
+        return;
+      }
+      double result = Math.sqrt((double) count) * m3 / Math.pow(m2, 1.5);
```
JackieTien97: **Skewness and kurtosis use inconsistent estimators.**

This is the *population / biased* skewness g1 (`sqrt(n) * m3 / m2^1.5`), while `KURTOSIS` below uses the *sample / bias-corrected* excess kurtosis (the Excel `KURT` formula with the `n(n+1)/((n-1)(n-2)(n-3))` factor). Mixing biased skewness with unbiased kurtosis is unusual — most engines are internally consistent (Spark: both biased; Excel/SAS: both sample-corrected).

Could you confirm which semantics are intended and make the two consistent? Either way it would help to document the exact definition (and how it compares to other DBs) so users aren't surprised.

iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/execution/aggregation/CentralMomentAccumulator.java
```diff
@@ -0,0 +1,298 @@
...
+      buffer.putDouble(m2);
+      buffer.putDouble(m3);
+      buffer.putDouble(m4);
+      columnBuilders[0].writeBinary(new Binary(bytes));
+    }
+  }
+
+  @Override
+  public void outputFinal(ColumnBuilder columnBuilder) {
+    if (count == 0 || m2 == 0) {
```
JackieTien97: The exact `m2 == 0` test is fragile here. Unlike the table variants (`TableCentralMomentAccumulator` / `TableCorrelationAccumulator`), which apply `normalizeZero` with `EPSILON = 1e-12` after `removeInput`, this tree-model accumulator subtracts partial state in `removeIntermediate` without clamping. After removal in a sliding window over constant data, `m2` can land at a tiny residual (e.g. `1e-13`) rather than exactly `0`, so this guard is skipped and `skewness = sqrt(n) * m3 / m2^1.5` blows up to a spurious huge value instead of returning null. Consider the same epsilon clamp for consistency.

iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/execution/aggregation/CorrelationAccumulator.java
```diff
@@ -0,0 +1,271 @@
...
+
+import java.nio.ByteBuffer;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+public class CorrelationAccumulator implements Accumulator {
+
+  private static final int INTERMEDIATE_SIZE = Long.BYTES + 5 * Double.BYTES;
+
+  public enum CorrelationType {
```
JackieTien97: `CorrelationType` has a single value `CORR`, and `outputFinal` then does `if (correlationType != CorrelationType.CORR) throw ...`, a branch that can never fire. Since covariance lives in its own `CovarianceAccumulator`, this enum + the constructor parameter + the dead check are unnecessary indirection and can be removed. (The same single-value-enum pattern is repeated in `TableCorrelationAccumulator` and `GroupedCorrelationAccumulator`.)

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/TypeInferenceUtils.java
```diff
@@ -190,7 +197,27 @@ private static void verifyIsAggregationDataTypeMatched(String aggrFuncName, TSDa
...
+                + "numeric data types [INT32, INT64, FLOAT, DOUBLE]");
+      case SqlConstant.SKEWNESS:
+      case SqlConstant.KURTOSIS:
+        if (dataType.isNumeric() || TSDataType.TIMESTAMP.equals(dataType)) {
+          return;
+        }
+        throw new SemanticException(
+            "Aggregate functions [SKEWNESS, KURTOSIS] only support "
+                + "numeric data types [INT32, INT64, FLOAT, DOUBLE, TIMESTAMP]");
+      /**
```
JackieTien97: Use a normal comment (`//` or `/* ... */`) instead of a Javadoc `/** ... */` block here. Javadoc attached to a `case` label inside a method body is misplaced — it documents nothing and some linters / the `javadoc` tool will warn about it.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/analyze/ExpressionTypeAnalyzer.java
```diff
@@ -361,6 +361,7 @@ public TSDataType visitFunctionExpression(
       }
 
       if (functionExpression.isBuiltInAggregationFunctionExpression()) {
+
```
JackieTien97: Stray blank line — this is the only change in this file and appears accidental. Please remove it to keep the diff clean.

JackieTien97 (changes requested):
Thanks for the substantial work — the coverage is thorough (tree + table + grouped accumulators, sliding-window / window-frame support, type validation, and a good spread of ITs including multi-region merge).

I have one correctness issue I'd like resolved before merge, one semantics question, and a few minor cleanups (all left as inline comments):

**Correctness**
- **`long` overflow in the KURTOSIS denominator** (`CentralMomentAccumulator`, `TableCentralMomentAccumulator`, `GroupedCentralMomentAccumulator`): `(count-1)*(count-2)*(count-3)` is computed in `long` and overflows for `count` above ~2.1M (sign-flips near ~2.5M), producing wrong kurtosis. Million-row groups are common here. Fix by computing in `double`.

**Semantics**
- **SKEWNESS (population / biased) vs KURTOSIS (sample / bias-corrected) use different estimator conventions.** Please confirm intent, make them consistent, and document the exact definitions.

**Minor**
- Tree `CentralMomentAccumulator` lacks the `EPSILON` zero-clamp that the table variants apply after element removal — can yield spurious non-null skew/kurtosis in sliding windows over constant data.
- `CorrelationType` is a single-value enum with an unreachable guard branch (dead code; repeated in the table/grouped variants).
- `TypeInferenceUtils`: Javadoc `/** */` used as an inline comment inside a `switch`.
- `ExpressionTypeAnalyzer`: a stray blank line is the only change in the file.

A couple of non-blocking notes not tied to a specific line:
- No user-facing documentation update for the 7 new SQL functions — is a docs PR planned?
- `getDoubleValue(...)` is duplicated nearly verbatim across ~9 new classes; it could be a shared helper.

Overall direction looks good; mainly the kurtosis overflow needs addressing.

iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/execution/aggregation/CentralMomentAccumulator.java
CoollZzz: Fixed.

I forced the kurtosis arithmetic onto double by introducing a local `double n = count` and using `n` consistently in both `term1` and `term2`. This avoids the intermediate long multiplication overflow in expressions like `(count - 1) * (count - 2) * (count - 3)`. The same fix was applied to `TableCentralMomentAccumulator` and `GroupedCentralMomentAccumulator`.

iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/execution/operator/source/relational/aggregation/TableCentralMomentAccumulator.java
CoollZzz: Fixed.

iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/execution/operator/source/relational/aggregation/grouped/GroupedCentralMomentAccumulator.java
CoollZzz: Fixed.

iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/execution/aggregation/CentralMomentAccumulator.java
CoollZzz: Confirmed and fixed.
Previously `SKEWNESS` used the uncorrected Fisher moment coefficient, while `KURTOSIS` used the sample-corrected excess kurtosis formula. I changed `SKEWNESS` to the sample-corrected form as well, so the two functions now follow the same sample-corrected convention.

iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/execution/aggregation/CentralMomentAccumulator.java
CoollZzz: Fixed.
I added `EPSILON = 1e-12` handling to the tree-model `CentralMomentAccumulator`:
- `removeIntermediate()` now normalizes `m2`, `m3`, and `m4` back to zero when the residual is within epsilon
- `outputFinal()` now uses `Math.abs(m2) < EPSILON` instead of exact `m2 == 0`
This keeps the tree-model sliding-window behavior consistent with the table variants. and avoids spurious non-null skewness/kurtosis on constant windows.

iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/execution/aggregation/CorrelationAccumulator.java
CoollZzz: Cleaned up.

`CorrelationType` only had a single value and the corresponding guard branch was indeed unreachable. I removed that enum-based indirection from: `CorrelationAccumulator` 、`TableCorrelationAccumulator`、
 `GroupedCorrelationAccumulator` and simplified the related factory calls as well.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/TypeInferenceUtils.java
CoollZzz: Fixed.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/analyze/ExpressionTypeAnalyzer.java
CoollZzz: Fixed.

PR #17293, opened 2026-03-12
Subscription: support consumer timeout and idle disconnect

VGalaxies:
## Summary
- Expose `connectionTimeoutInMs` for all subscription consumers and pass it through builder, session open, and handshake.
- Add a lightweight server-side idle disconnect check based on subscription heartbeat inactivity without changing the receiver threading model.
- Reuse the existing server-side `closeConsumer(...)` flow so timeout-triggered disconnect also unsubscribes subscribed topics before dropping the consumer.

## Verification
- `mvn -T 8 spotless:apply -P with-integration-tests`
- `MAVEN_OPTS='-XX:ReservedCodeCacheSize=512m' mvn -T 8 clean package -P with-integration-tests -DskipUTs -pl integration-test,distribution -DfailIfNoTests=false -am -U`

---

**This PR was primarily authored with Codex using gpt-5.4 xhigh and then hand-reviewed by me. I AM responsible for every change made in this PR. I aimed to keep it aligned with our goals, though I may have missed minor issues. Please flag anything that feels off, I'll fix it quickly.**


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/receiver/SubscriptionReceiverV1.java
jt2594838, line 1: May add a unit test for this.

jt2594838 (approved):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/receiver/SubscriptionReceiverV1.java
VGalaxies: Added a lightweight UT in `SubscriptionReceiverV1Test` under datanode. It covers the timeout threshold calculation and the non-timeout branches for recently active consumers / in-flight requests. Verified with `mvn -pl iotdb-core/datanode -am -DskipITs -DskipIntegrationTests -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false -Dtest=SubscriptionReceiverV1Test test`.

PR #17321, opened 2026-03-19
Remove extra user caching for REST services and migrate to the unified user check cache.

wenyanshi-123:
Remove extra user caching for REST services and migrate to the unified user check cache.

### DIFF
--- a/external-service-impl/rest/pom.xml
+++ b/external-service-impl/rest/pom.xml
@@ -133,18 +133,6 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
-            <groupId>com.github.ben-manes.caffeine</groupId>
-            <artifactId>caffeine</artifactId>
-            <scope>provided</scope>
-            <exclusions>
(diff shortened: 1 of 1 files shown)


iotdb-core/node-commons/src/assembly/resources/conf/iotdb-system.properties.template
```diff
@@ -584,21 +584,6 @@ enable_swagger=false
...
-# maximum number of users can be stored in the user login cache.
-# effectiveMode: restart
-# Datatype: int
-cache_max_num=100
-
-# init capacity of users can be stored in the user login cache.
-# effectiveMode: restart
-# Datatype: int
-cache_init_num=10
-
```
CRZbulabula, line 601: Why u delete them?

HTHou: They are all no longer in use.

PR #17347, opened 2026-03-24
Session:  session-c interface

hongzhi-gao:
### What
Adds a C API (SessionC.h / SessionC.cpp) around the existing C++ Session / TableSession, for C/FFI callers without exposing C++ types.

### Design

Opaque handles (CSession, CTableSession, CTablet, CSessionDataSet, CRowRecord); public symbols prefixed with ts_*.
TsStatus return codes + thread-local last error via ts_get_last_error().
Thin wrappers: convert C args → C++ types, map exceptions to status codes; reuse existing client logic (no duplicate Thrift layer).
### Scope
Tree + table model: connect, metadata, inserts (record/tablet), queries, result iteration, deletes—aligned with current Session behavior.

### This PR has:

self-reviewed
tests added or updated

### Key files
client-cpp/src/main/SessionC.h, SessionC.cpp
client-cpp/src/test/main_c*.cpp, sessionC*IT.cpp

### DIFF
--- a/example/client-c-example/src/table_example.c
+++ b/example/client-c-example/src/table_example.c
@@ -1,0 +1,151 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*
+ * Table model: CREATE DATABASE/TABLE, insert rows via Tablet, SELECT, DROP DATABASE.
+ * Requires IoTDB table-SQL support. Edit HOST / PORT / credentials below.
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "SessionC.h"
+
+#define HOST "127.0.0.1"
+#define PORT 6667
+#define USER "root"
+#define PASS "root"
+
+#define DB_NAME "cdemo_db"
+#define TABLE_NAME "cdemo_t0"
+
+static void fail(const char* ctx, CTableSession* s) {
+    fprintf(stderr, "[table_example] %s failed: %s\n", ctx, ts_get_last_error());
+    if (s) {
+        ts_table_session_close(s);
+        ts_table_session_destroy(s);
+    }
+    exit(1);
+}
+
+int main(void) {
+    CTableSession* session = ts_table_session_new(HOST, PORT, USER, PASS, "");
+    if (!session) {
+        fprintf(stderr, "[table_example] ts_table_session_new returned NULL: %s\n", ts_get_last_error());
+        return 1;
+    }
+    if (ts_table_session_open(session) != TS_OK) {
+        fail("ts_table_session_open", session);
+    }
+
+    char sql[512];
+    snprintf(sql, sizeof(sql), "DROP DATABASE IF EXISTS %s", DB_NAME);
+    (void)ts_table_session_execute_non_query(session, sql);
+
+    snprintf(sql, sizeof(sql), "CREATE DATABASE %s", DB_NAME);
+    if (ts_table_session_execute_non_query(session, sql) != TS_OK) {
+        fail("CREATE DATABASE", session);
+    }
+
+    snprintf(sql, sizeof(sql), "USE \"%s\"", DB_NAME);
+    if (ts_table_session_execute_non_query(session, sql) != TS_OK) {
+        fail("USE DATABASE", session);
+    }
+
+    const char* ddl =
+        "CREATE TABLE " TABLE_NAME " ("
+        "tag1 string tag,"
+        "attr1 string attribute,"
+        "m1 double field)";
+    if (ts_table_session_execute_non_query(session, ddl) != TS_OK) {
+        fail("CREATE TABLE", session);
+    }
+
+    const char* columnNames[] = {"tag1", "attr1", "m1"};
+    TSDataType_C dataTypes[] = {TS_TYPE_STRING, TS_TYPE_STRING, TS_TYPE_DOUBLE};
+    TSColumnCategory_C colCategories[] = {TS_COL_TAG, TS_COL_ATTRIBUTE, TS_COL_FIELD};
+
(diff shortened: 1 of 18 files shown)


iotdb-client/client-cpp/src/main/SessionC.cpp
```diff
@@ -0,0 +1,1421 @@
...
+    try {
+        auto* cs = new CSession_();
+        cs->cpp = std::make_shared<Session>(
+            std::string(host), rpcPort,
+            std::string(username), std::string(password));
+        return cs;
+    } catch (const std::exception& e) {
+        handleException(e);
+        return nullptr;
+    }
```
jt2594838, line 233: Will cs be freed when an exception occurs?
The same below.

iotdb-client/client-cpp/src/main/SessionC.cpp
```diff
@@ -0,0 +1,1421 @@
...
+        return cts;
+    } catch (const std::exception& e) {
+        handleException(e);
+        return nullptr;
+    }
+}
+
+void ts_table_session_destroy(CTableSession* session) {
+    delete session;
+}
```
jt2594838, line 358: Check if it is nullptr?

iotdb-client/client-cpp/src/main/SessionC.cpp
```diff
@@ -0,0 +1,1421 @@
...
+                                          int attrsCount,
+                                          const char* const* attrKeys,
+                                          const char* const* attrValues,
+                                          const char* measurementAlias) {
+    clearError();
+    if (!session) return setError(TS_ERR_NULL_PTR, "session is null");
+    try {
+        std::map<std::string, std::string> props = propsCount > 0 ? toStringMap(propsCount, propKeys, propValues) : std::map<std::string, std::string>();
+        std::map<std::string, std::string> tags = tagsCount > 0 ? toStringMap(tagsCount, tagKeys, tagValues) : std::map<std::string, std::string>();
+        std::map<std::string, std::string> attrs = attrsCount > 0 ? toStringMap(attrsCount, attrKeys, attrValues) : std::map<std::string, std::string>();
```
jt2594838, line 491: Will this result in unnecessary copying?

iotdb-client/client-cpp/src/main/SessionC.cpp
```diff
@@ -0,0 +1,1421 @@
...
+}
+
+bool ts_dataset_has_next(CSessionDataSet* dataSet) {
+    if (!dataSet || !dataSet->cpp) return false;
+    try {
+        return dataSet->cpp->hasNext();
+    } catch (...) {
+        return false;
+    }
+}
```
jt2594838, line 1281: global error?
The same below.

iotdb-client/client-cpp/src/main/SessionC.cpp
```diff
@@ -0,0 +1,1421 @@
...
+        case TSDataType::TIMESTAMP: return !f.longV.is_initialized();
+        case TSDataType::FLOAT:     return !f.floatV.is_initialized();
+        case TSDataType::DOUBLE:    return !f.doubleV.is_initialized();
+        case TSDataType::TEXT:
+        case TSDataType::STRING:
+        case TSDataType::BLOB:      return !f.stringV.is_initialized();
+        case TSDataType::DATE:      return !f.dateV.is_initialized();
+        default:                    return true;
+    }
+}
```
jt2594838, line 1359: No such a method in Cpp?

iotdb-client/client-cpp/src/main/SessionC.cpp
```diff
@@ -0,0 +1,1421 @@
...
+    if (f.stringV.is_initialized()) {
+        g_stringBuf = f.stringV.value();
+        return g_stringBuf.c_str();
+    }
+    return "";
+}
+
+TSDataType_C ts_row_record_get_data_type(CRowRecord* record, int index) {
+    if (!record || !record->cpp) return TS_TYPE_TEXT;
+    if (index < 0 || index >= (int)record->cpp->fields.size()) return TS_TYPE_TEXT;
```
jt2594838: TS_TYPE_INVALID?

iotdb-client/client-cpp/src/test/main.cpp
```diff
@@ -23,30 +23,35 @@
...
+            SessionBuilder builder;
+            session = builder.host("127.0.0.1")
+                          ->rpcPort(6667)
+                          ->username("root")
+                          ->password("root")
+                          ->useSSL(false)
+                          ->build();
+        } else {
+            session->open(false);
+        }
```
jt2594838, line 43: Session is not open in the first test?

iotdb-client/client-cpp/src/main/SessionC.cpp
hongzhi-gao: If `new CSession_()` succeeds but the `make_shared/Session` construction throws an exception, the old code’s `catch` block would not delete `cs`, resulting in a memory leak. **This has now been fixed** by first completing all potentially failing C++ constructions, and only then allocating the C-side wrapper (similar patterns for datasets/row records and other cases have already been aligned using stack-based temporary objects).

iotdb-client/client-cpp/src/main/SessionC.cpp
hongzhi-gao: delete applied to a nullptr has no effect and is well-defined behavior according to the C++ standard.

iotdb-client/client-cpp/src/main/SessionC.cpp
hongzhi-gao: When initializing a `std::map` from a return value, it will use move semantics.

iotdb-client/client-cpp/src/main/SessionC.cpp
hongzhi-gao: Updated dataset iteration APIs to populate ts_get_last_error() on errors; added clearError() + null-handle errors for consistency with other bindings.

iotdb-client/client-cpp/src/main/SessionC.cpp
hongzhi-gao: Added Field::isNull() in Common.h (same logic as the previous C helper), ts_row_record_is_null now delegates to it. RowRecord::toString uses isNull() where it matches;

iotdb-client/client-cpp/src/main/SessionC.cpp
hongzhi-gao: Done. Added TS_TYPE_INVALID (255) to TSDataType_C

iotdb-client/client-cpp/src/test/main.cpp
hongzhi-gao: SessionBuilder::build() already calls open(false) before returning the Session (see SessionBuilder.h), so the first test case starts with an open session. Subsequent cases rely on the else branch after testCaseEnded closes the session.

example/client-c-example/src/table_example.c
```diff
@@ -0,0 +1,151 @@
...
+    fprintf(stderr, "[table_example] %s failed: %s\n", ctx, ts_get_last_error());
+    if (s) {
+        ts_table_session_close(s);
+        ts_table_session_destroy(s);
+    }
+    exit(1);
+}
+
+int main(void) {
+    CTableSession* session = ts_table_session_new(HOST, PORT, USER, PASS, "");
```
jt2594838, line 50: Add a comment to explain what the last "" is.

PR #17364, opened 2026-03-26
Fixed the issue where prepareStatement did not support the Blob type.When the ResultSet is empty, an SQL exception should be thrown

Adam-chengzhuo:
## Description
The PreparedStatement.setBlob() method in IoTDB JDBC directly throws an exception, making it impossible to set Blob-type values.
ResultSet.findColumn() handles tree models and table models inconsistently. According to the JDBC specification, when the ResultSet is empty, an SQL exception should be thrown.

### DIFF
--- a/iotdb-client/jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBJDBCResultSet.java
+++ b/iotdb-client/jdbc/src/main/java/org/apache/iotdb/jdbc/IoTDBJDBCResultSet.java
@@ -231,8 +231,15 @@ public class IoTDBJDBCResultSet implements ResultSet {
   }
 
   @Override
-  public int findColumn(String columnName) {
-    return ioTDBRpcDataSet.findColumn(columnName);
+  public int findColumn(String columnName) throws SQLException {
+    if (isClosed()) {
+      throw new SQLException("ResultSet is closed");
+    }
+    try {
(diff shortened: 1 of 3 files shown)


HTHou:
Hi, you can subscribe to the development mailing list and mention this PR there. This way, other developers can help you review the PR more quickly.

https://iotdb.apache.org/Community/Communication-Channels.html#_1-mailing-lists

PDGGK:
Hi @Adam-chengzhuo, thanks for fixing this -- I ran into the same findColumn issue recently.

One thing that might be useful: the unknown-column NPE actually originates one layer down, in `IoTDBRpcDataSet.findColumn` (which this method delegates to) -- it does `return columnOrdinalMap.get(columnName)`, so `get()` returning null on an unknown column is what auto-unboxes into the NPE. The same pattern is also in `IoTDBJDBCDataSet.findColumn`, and `SessionDataSet.DataIterator.findColumn` reaches it too (a non-JDBC path).

Guarding the null at that source -- the way the sibling `findColumnNameByIndex` already does -- would cover all of those callers, not just the JDBC ResultSet path, and the boundary could then just convert the resulting exception to SQLException. It'd be a small addition (the rpc-layer guard plus the two extra call sites, with a test). Happy to help with that here or as a follow-up if you think it's worth covering.

Thanks again!

Issue #17337, opened 2026-03-23
[Feature request] Support `GROUP BY ALL` syntax for improved query ergonomics (Table Model only)


JackieTien97:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/iotdb/issues) and found nothing similar.

### Motivation

When performing complex aggregation queries, standard SQL requires users to explicitly repeat all non-aggregated columns from the `SELECT` list inside the `GROUP BY` clause. In real-world development and data analysis scenarios, this introduces several pain points:

1. **Tedious and Error-Prone Queries:** When grouping by multiple dimensional attributes or tags (e.g., `region`, `plant_id`, `device_model`, `status`), users have to write these fields twice. If a user updates the `SELECT` list during iteration but forgets to update the `GROUP BY` list, it results in a syntax error.
2. **High Friction in Exploratory Data Analysis (EDA):** Data analysts frequently add or remove dimension columns to observe metrics at different granularities. Synchronizing the `SELECT` and `GROUP BY` clauses for every minor adjustment breaks the analytical flow and reduces efficiency.
3. **Alignment with Modern Analytical Databases:** Mainstream modern OLAP and analytical databases (such as DuckDB, ClickHouse, etc.) already support the `GROUP BY ALL` syntax. It automatically infers the grouping columns, significantly simplifying SQL writing.

With the introduction of IoTDB's Table Model, the relational semantics are highly geared towards data analysts. Introducing `GROUP BY ALL` will further elevate the ease of use and the overall developer experience within the Table Model.

### Solution

We propose introducing the `GROUP BY ALL` syntax into the IoTDB SQL dialect. 
**Note: To maintain architectural clarity and backward compatibility, this feature will ONLY be implemented within the Table Model. The existing Tree Model requires no changes.**

The implementation will primarily focus on the frontend parsing and logical planning phases of the Table Model:

1. **Update Antlr4 Grammar (Parser):** Modify the table-model-specific grammar file (`iotdb-core/relational-grammar/src/main/antlr4/org/apache/iotdb/db/relational/grammar/sql/RelationalSql.g4`) to support the `ALL` keyword within the `GROUP BY` clause rules.
2. **Rewrite Analyzer Logic (Analyzer / Logical Planner):** Intercept the `GROUP BY ALL` syntax during the semantic analysis phase. If detected, the analyzer should iterate through the current `SELECT` clause:
   * Automatically extract all scalar expressions or column references that **do not contain aggregate functions**.
[...]

PR #17380, opened 2026-03-27
Support GROUP BY ALL in relational queries

Eliaaazzz:
Addresses #17337

## Description

This PR adds support for `GROUP BY ALL` in Table Model relational queries.

`GROUP BY ALL` infers grouping keys from non-aggregate and non-window expressions in the `SELECT` list, which improves query ergonomics and reduces duplicated grouping expressions in common aggregation queries.

This PR also keeps `date_bin_gapfill(...)` behavior consistent under `GROUP BY ALL`, including existing validation for multiple gapfill expressions.

<hr>

This PR has:
- [x] been self-reviewed.

<hr>

- `iotdb-core/relational-grammar/src/main/antlr4/org/apache/iotdb/db/relational/grammar/sql/RelationalSql.g4`
- `iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/sql/parser/AstBuilder.java`
- `iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/sql/ast/GroupBy.java`
- `iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/analyzer/StatementAnalyzer.java`
- `iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/sql/util/SqlFormatter.java`
- `iotdb-core/datanode/src/test/java/org/apache/iotdb/db/queryengine/plan/relational/analyzer/GroupByAllTest.java`

Tested with:
- `./mvnw.cmd -pl iotdb-core/datanode -Dtest=GroupByAllTest test`

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/analyzer/StatementAnalyzer.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/analyzer/StatementAnalyzer.java
@@ -2595,6 +2595,12 @@ public class StatementAnalyzer {
     private Analysis.GroupingSetAnalysis analyzeGroupBy(
         QuerySpecification node, Scope scope, List<Expression> outputExpressions) {
       if (node.getGroupBy().isPresent()) {
+
+        // Handle GROUP BY ALL: infer grouping columns from SELECT expressions
+        if (node.getGroupBy().get().isAll()) {
+          return analyzeGroupByAll(node, scope, outputExpressions);
+        }
+
         ImmutableList.Builder<List<Set<FieldId>>> cubes = ImmutableList.builder();
         ImmutableList.Builder<List<Set<FieldId>>> rollups = ImmutableList.builder();
         ImmutableList.Builder<List<Set<FieldId>>> sets = ImmutableList.builder();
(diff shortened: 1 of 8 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/analyzer/StatementAnalyzer.java
```diff
@@ -2718,6 +2724,69 @@ private Analysis.GroupingSetAnalysis analyzeGroupBy(
...
+      ImmutableList.Builder<Expression> gapFillGroupingExpressions = ImmutableList.builder();
+
+      for (Expression outputExpression : outputExpressions) {
+        List<FunctionCall> aggregates =
+            extractAggregateFunctions(ImmutableList.of(outputExpression));
+        List<FunctionCall> windowFunctions =
+            extractWindowFunctions(ImmutableList.of(outputExpression));
+        if (!aggregates.isEmpty() || !windowFunctions.isEmpty()) {
+          continue;
+        }
```
JackieTien97, line 2759: what about `select s1+avg(s2) from t1 group by all`? You can try what will duckdb deal with this sql, and IMO, `group by all` should equal to `group by s1` which means group by all should include all non-agg expression in select including sub-expression inside one select item.

JackieTien97 (changes requested):
Add IT in integration-test module

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/analyzer/StatementAnalyzer.java
Eliaaazzz: Fixed in ed29a53. Now for expressions containing both aggregate and non-aggregate parts, extract column references outside aggregate/window function boundaries instead of skipping the whole expression.

For example, `SELECT s1 + avg(s2) FROM t1 GROUP BY ALL` is now equivalent to `GROUP BY s1`.

Implementation: added `extractNonAggregateColumnReferences` in `ExpressionTreeUtils` — it walks the expression tree but stops recursing into aggregate/window function subtrees, collecting only the `Identifier`/`DereferenceExpression` nodes outside those boundaries.

Also added corresponding unit tests and integration tests to cover this case.

PR #17411, opened 2026-03-31
Fix last_by returning null row after deleting all table data (#16985)

### LINKED ISSUE #16985: [Bug] 2.0.5版本当数据删除之后再使用last_by函数查询会有数据返回，且这条数据为null
### Search before asking

### Version

2.0.5 window版本

### Describe the bug and provide the minimal reproduce step

表中对应的标签没有数据的情况下last_by查询是没有数据

Image

给这个对应的标签添加一条数据

Image

**问题出现**：删除这条数据之后，使用last_by 方法查询时会返回一条数据，这个数据全部返回的字段为空

Image

### What did you expect to see?

数据删除之后使用last_by查询应该不返回任何数据

### What did you see instead?

数据删除之后使用last_by查询应该返回一条数据，而且这条数据全部字段都返回了null

### Anything else?

_No response_

PDGGK:
## Description

Fixes #16985

In table model, after deleting all data of a device, `last_by` queries still returned one row with null values instead of returning no rows:

```sql
-- After deleting all data:
select last_by(s0, time) from table where deviceId = 'd0'
-- Expected: 0 rows
-- Actual: 1 row with null
```

### Root Cause

Three related issues:

1. **`LastQueryAggTableScanOperator`** could still emit a result row for non-grouped `last_by` scans even when every `LastByAccumulator` had no initialized result, causing `evaluateFinal()` to produce a null row.

2. **`FileLoaderUtils`** read aligned time-column modifications using the concrete measurement ID instead of `AlignedFullPath.VECTOR_PLACEHOLDER`, inconsistent with aligned deletion semantics used in other paths.

3. **`ModificationUtils`** did not set `modified = true` when an aligned value chunk was fully deleted, allowing statistics-based shortcuts to use stale metadata.

### Changes

| File | Change |
|------|--------|
| `LastQueryAggTableScanOperator.java` | Skip emitting result row when all aggregators are `LastByAccumulator` and none has initialized result. Applied to both normal aggregation and cache-hit paths. |
| `FileLoaderUtils.java` | Use `AlignedFullPath.VECTOR_PLACEHOLDER` for time-column modification lookups |
| `ModificationUtils.java` | Set `modified = true` when value chunk is fully removed |
| `IoTDBDeletionTableIT.java` | Add regression tests for 2-arg and 3-arg `last_by` after delete |

### Semantic Note

This changes non-grouped `last_by` on empty results from returning one null row to returning zero rows, consistent with grouped-query behavior (which already returns zero rows for empty devices). Other aggregations like `count`/`max`/`last` are not affected by this change. Please confirm this is the intended semantics.

### Verification

Integration tests pass locally:
```
mvn -P with-integration-tests,TableSimpleIT -pl integration-test -am \
  -DskipUTs -Dspotless.skip=true -Dcheckstyle.skip=true -Drat.skip=true \
  -Dit.test=IoTDBDeletionTableIT#testLastByWithTimeShouldReturnNoRowsAfterDeletingAllData+testThreeArgLastByShouldReturnNoRowsAfterDeletingAllData \
  verify
```
Result: 2 tests run, 0 failures, 0 errors.

### DIFF
--- a/integration-test/src/test/java/org/apache/iotdb/relational/it/db/it/IoTDBDeletionTableIT.java
+++ b/integration-test/src/test/java/org/apache/iotdb/relational/it/db/it/IoTDBDeletionTableIT.java
@@ -1049,6 +1049,84 @@ public class IoTDBDeletionTableIT {
   }
 
   @Test
+  public void testLastByWithTimeShouldReturnNoRowsAfterDeletingAllData() throws SQLException {
+    try (Connection connection = EnvFactory.getEnv().getConnection(BaseEnv.TABLE_SQL_DIALECT);
+        Statement statement = connection.createStatement()) {
+      statement.execute("use test");
+      statement.execute("create table last_by_delete_0(deviceId string tag, s0 int32 field)");
+      statement.execute("insert into last_by_delete_0(time, deviceId, s0) values (1, 'd0', 1)");
+      statement.execute("flush");
+
+      try (ResultSet resultSet =
+          statement.executeQuery(
+              "select last_by(s0, time) from last_by_delete_0 where deviceId = 'd0'")) {
+        assertTrue(resultSet.next());
+      }
+
+      statement.execute("delete from last_by_delete_0 where deviceId = 'd0'");
+
+      try (ResultSet resultSet =
+          statement.executeQuery(
+              "select last_by(s0, time) from last_by_delete_0 where deviceId = 'd0'")) {
+        assertResultSetEmpty(resultSet);
+      }
+
+      try (ResultSet resultSet =
+          statement.executeQuery(
+              "select last_by(s0, time) from last_by_delete_0 where deviceId = 'd0'")) {
+        assertResultSetEmpty(resultSet);
+      }
+    }
+  }
+
+  @Test
+  public void testThreeArgLastByShouldReturnNoRowsAfterDeletingAllData() throws SQLException {
+    try (Connection connection = EnvFactory.getEnv().getConnection(BaseEnv.TABLE_SQL_DIALECT);
+        Statement statement = connection.createStatement()) {
+      statement.execute("use test");
+      statement.execute("create table last_by_delete_1(deviceId string tag, s0 int32 field)");
+      statement.execute("insert into last_by_delete_1(time, deviceId, s0) values (1, 'd0', 1)");
+      statement.execute("flush");
+
+      try (ResultSet resultSet =
+          statement.executeQuery(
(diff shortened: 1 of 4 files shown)


PDGGK:
Rebased onto the latest master. The only conflict was in `LastQueryAggTableScanOperator`, where the table last-row optimization was refactored since this PR was opened (the accumulator classes moved to the `calc-commons` module and the `EMPTY_PRIMITIVE_TYPE` placeholder was renamed to `PLACEHOLDER_NO_VALUE`); I reconciled the fix against those changes, and the `last_by` empty-result handling (`shouldSkipEmptyLastByResult`) is preserved. The two regression tests in `IoTDBDeletionTableIT` are unchanged.

@Caideyipi would you have a chance to take a look? You touched this operator most recently (#17677), so your eyes here would be very helpful. Thanks!

Wei-hao-Li:
Hello, the correct performance of target SQL I have commented in https://github.com/apache/iotdb/issues/16985#issuecomment-5018814871. You can try to re-fix this issue.

PDGGK:
Thanks @Wei-hao-Li — you're right, and I looked into this more carefully.

For the deleted-data case in this issue, master already returns the correct single NULL row: after a device's data is deleted the device still exists, so `last_by` over it yields one row with a NULL value. My change in this PR added a guard that suppressed that row — the wrong direction — so I'm closing this PR.

While tracing it I did notice a separate case that may be the empty result you're pointing at: querying a device that was **never** written (`where deviceId = 'dX'` with no such device) appears to return 0 rows rather than a single NULL row. If a no-GROUP-BY aggregate over a non-existent device should also return one NULL row per the "always return NULL when there's no input data" rule, I'd be glad to take that on as a separate, properly-scoped fix — could you confirm that's the intended behavior?

Thanks again for the correction.

Wei-hao-Li:
> Thanks @Wei-hao-Li — you're right, and I looked into this more carefully.
> 
> For the deleted-data case in this issue, master already returns the correct single NULL row: after a device's data is deleted the device still exists, so `last_by` over it yields one row with a NULL value. My change in this PR added a guard that suppressed that row — the wrong direction — so I'm closing this PR.
> 
> While tracing it I did notice a separate case that may be the empty result you're pointing at: querying a device that was **never** written (`where deviceId = 'dX'` with no such device) appears to return 0 rows rather than a single NULL row. If a no-GROUP-BY aggregate over a non-existent device should also return one NULL row per the "always return NULL when there's no input data" rule, I'd be glad to take that on as a separate, properly-scoped fix — could you confirm that's the intended behavior?
> 
> Thanks again for the correction.

Yes, we should keep performance consistent with other query engine which meets SQL standard, like Trino.
image

PR #17425, opened 2026-04-02
Support some configurations about data partition base on database level

zerolbsony:


CRZbulabula (commented):
Please do not merge the time-partition ALTER path in its current form. Time partition origin/interval define the physical partition key. Once data exists, changing either value requires a coordinated migration of DataPartitionTable entries, region assignments, TsFile partition IDs, recovery metadata, and every DataNode cache; this PR does not perform such a migration, and the current local-only invalidation can leave different DataNodes computing different partition IDs.

The supported lifecycle should be: (1) time_partition_origin/time_partition_interval may be specified only by CREATE DATABASE; (2) if either property is omitted at CREATE time, resolve it from the global configuration and persist the resolved value in the database schema; (3) ALTER DATABASE must reject these properties, including the general ALTER request and dedicated setter paths. Please add tests proving ALTER is rejected and omitted CREATE properties remain consistent across a 1C3D cluster and restart/recovery.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/service/thrift/ConfigNodeRPCServiceProcessor.java
```diff
@@ -537,6 +523,12 @@ public TSStatus setTimePartitionInterval(final TSetTimePartitionIntervalReq req)
         new SetTimePartitionIntervalPlan(req.getDatabase(), req.getTimePartitionInterval()));
   }
 
+  @Override
+  public TSStatus setTimePartitionOrigin(final TSetTimePartitionOriginReq req) throws TException {
```
CRZbulabula: Please do not expose a mutable time-partition setter through the RPC layer. `time_partition_origin` and `time_partition_interval` define the physical partition key; changing either one remaps timestamps to different `TTimePartitionSlot` / partition IDs, while existing `DataPartitionTable` entries and TsFiles are not migrated. These properties should be accepted only by `CREATE DATABASE`, and the general `ALTER DATABASE` request should reject them.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/schema/ClusterSchemaInfo.java
```diff
@@ -280,11 +284,21 @@ public TSStatus alterDatabase(final DatabaseSchemaPlan plan) {
             currentSchema.isNeedLastCache());
       }
 
+      if (alterSchema.isSetTimePartitionOrigin()) {
+        currentSchema.setTimePartitionOrigin(alterSchema.getTimePartitionOrigin());
```
CRZbulabula: Please remove this ALTER mutation. Updating only the schema and `TimePartitionUtils` changes the partition function for an existing database without migrating its data-partition table, region assignments, TsFile partition IDs, or recovery metadata. The CREATE path should persist either the explicit value or the resolved global default, and these values must remain immutable afterwards.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/schema/ClusterSchemaManager.java
```diff
@@ -485,6 +498,41 @@ public TSStatus setTimePartitionInterval(
...
+  public TSStatus setTimePartitionOrigin(SetTimePartitionOriginPlan setTimePartitionOriginPlan) {
+    final TSStatus validationResult =
+        validateTimePartitionConfig(
+            new TDatabaseSchema(setTimePartitionOriginPlan.getDatabase())
+                .setTimePartitionOrigin(setTimePartitionOriginPlan.getTimePartitionOrigin()));
+    if (validationResult != null) {
+      return validationResult;
+    }
+    try {
+      return getConsensusManager().write(setTimePartitionOriginPlan);
```
CRZbulabula: Please remove this setter path instead of adding DataNode notification. Because the contract is CREATE-only, origin/interval must not have consensus setter APIs that can mutate an existing database. Rejecting these fields at the ALTER boundary also avoids a distributed window where different DataNodes calculate different partition IDs.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/execution/config/executor/ClusterConfigTaskExecutor.java
```diff
@@ -524,9 +553,15 @@ public SettableFuture<ConfigTaskResult> alterDatabase(
         }
         future.setException(new IoTDBException(tsStatus));
       } else {
+        if (databaseSchema.isSetTimePartitionOrigin()
+            || databaseSchema.isSetTimePartitionInterval()
+            || databaseSchema.isSetNeedLastCache()) {
+          ClusterPartitionFetcher.getInstance().invalidAllCache();
```
CRZbulabula: Local cache invalidation cannot make this ALTER safe: it only refreshes the DataNode executing the statement, while other DataNodes can keep the old database configuration and route the same timestamp to a different partition ID. Please remove the ALTER handling end-to-end; cache population should happen only after CREATE, using explicit values or resolved global defaults.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/service/thrift/ConfigNodeRPCServiceProcessor.java
zerolbsony: Resolved

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/schema/ClusterSchemaInfo.java
zerolbsony: Resolved

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/schema/ClusterSchemaManager.java
zerolbsony: Resolved

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/execution/config/executor/ClusterConfigTaskExecutor.java
zerolbsony: Resolved

PR #17430, opened 2026-04-03
Support EXPLAIN FORMAT JSON for Table Model

JackieTien97:
## Summary

- Add `FORMAT` option to `EXPLAIN` and `EXPLAIN ANALYZE` statements in the Table Model, allowing users to choose the output format (JSON, GRAPHVIZ, or TEXT)
- JSON format outputs a single-row, machine-parseable JSON document instead of multi-line text, making it easy for tools and scripts to programmatically analyze query plans
- **CLI output optimization**: when using `FORMAT JSON`, the CLI now prints JSON content without `|` table borders, so users can directly copy the JSON for visualization. The column header retains its table border formatting.
- Backward compatible: default output formats remain unchanged (GRAPHVIZ for EXPLAIN, TEXT for EXPLAIN ANALYZE)

## Syntax

```sql
-- EXPLAIN with format option
EXPLAIN (FORMAT <format>) <query>

-- EXPLAIN ANALYZE with format option
EXPLAIN ANALYZE [VERBOSE] (FORMAT <format>) <query>
```

**Supported formats:**

| Format | EXPLAIN | EXPLAIN ANALYZE | Description |
|--------|---------|-----------------|-------------|
| `GRAPHVIZ` | default | N/A | Box-drawing plan visualization |
| `TEXT` | N/A | default | Text-based output |
| `JSON` | supported | supported | Structured JSON output |

## CLI Output Format (FORMAT JSON)

Before (hard to copy JSON due to `|` borders):
```
+------------------------------------------+
|                          distribution plan|
+------------------------------------------+
|{                                         |
|  "name": "OutputNode-4",                 |
|  ...                                     |
|}                                         |
+------------------------------------------+
```

After (JSON content can be directly copied):
```
+-----------------+
|distribution plan|
+-----------------+
{
  "name": "OutputNode-4",
  "id": "4",
  "properties": { ... },
  "children": [ ... ]
}
+-----------------+
Total line number = 1
```

### 1. EXPLAIN (FORMAT JSON)

```sql
EXPLAIN (FORMAT JSON) SELECT * FROM testtb
```

### 2. EXPLAIN ANALYZE (FORMAT JSON)

```sql
EXPLAIN ANALYZE (FORMAT JSON) SELECT * FROM testtb
```

### 3. Invalid format produces clear error

```sql
EXPLAIN (FORMAT XML) SELECT * FROM testtb
-- Error: Unsupported EXPLAIN output format: XML
```

## Key Design Decisions

- **Single-row output**: JSON format returns all data in a single row, simplifying client-side parsing.
- **CLI border-free JSON**: JSON content lines in CLI output have no `|` borders, while the column header retains its table formatting.
- **Backward compatible**: No syntax change for existing queries. The `(FORMAT ...)` clause is entirely optional.

## Test Plan

- [x] Unit tests for `PlanGraphJsonPrinter` and `ExplainFormatTest`
- [x] Integration tests: JSON/GRAPHVIZ/TEXT formats, CTE, scalar subquery, multi-partition, error handling
- [x] CLI output format IT (`ExplainJsonCliOutputIT`): verifies JSON content has no `|` borders while header retains table formatting


Caideyipi:
I found a few functional issues in the new JSON explain path:

1. `EXPLAIN (FORMAT JSON)` with materialized CTEs can fail. CTE subqueries only inherit `ExplainType`, not the requested `ExplainOutputFormat` (`CteMaterializer` -> `Coordinator.executeForTableModel`). The inner CTE explain plan is still captured via `PlanGraphPrinter` text output in `TableDistributedPlanner`, but `TableModelStatementMemorySourceVisitor.mergeExplainResultsJson()` later parses that CTE result with `JsonParser.parseString(...)`. A query like this should be covered:

```sql
EXPLAIN (FORMAT JSON)
WITH cte1 AS MATERIALIZED (SELECT * FROM t)
SELECT * FROM cte1;
```

2. `EXPLAIN ... EXECUTE` and `EXPLAIN ... EXECUTE IMMEDIATE` lose the requested format when the prepared/immediate SQL is unwrapped in `Coordinator`. The code recreates `new Explain(resolvedSql)` / `new ExplainAnalyze(resolvedSql, verbose)`, which resets the format to `GRAPHVIZ` / `TEXT`, so `EXPLAIN (FORMAT JSON) EXECUTE stmt` and `EXPLAIN ANALYZE (FORMAT JSON) EXECUTE stmt` will not return JSON.

3. In JSON analyze mode, `ExplainAnalyzeOperator` sets `fragmentInstanceStatisticsDrawer` to `null`, but the scheduled intermediate logger still calls `buildFragmentInstanceStatistics()`, which dereferences that text drawer. Long-running `EXPLAIN ANALYZE (FORMAT JSON)` can therefore log repeated errors instead of intermediate analyze output. It should either skip the intermediate logger in JSON mode or render through the JSON drawer.

PR #17435, opened 2026-04-07
[To del/1.3] Fix Percentile, Quantile UDFs and update Cluster UDF

suyx1999:
# Update 3 UDFs: Percentile, Quantile and Cluster
### Percentile
- Fix out-of-bounds index issue
- Add according unit tests

### Quantile
- Fix incorrect type conversion
- Add according unit tests

## Update Cluster UDF

- This function takes a **single input time series**, splits it into **non-overlapping** contiguous subsequences (windows) of fixed length `l`, and clusters those subsequences to discover local patterns or segment structure.

### Input series

- **One** input series only.
- Types: `INT32` / `INT64` / `FLOAT` / `DOUBLE`.
- Points are read in time order; trailing samples that do not fill a full window are **dropped** (only `⌊n/l⌋` windows are used, where `n` is the number of valid points).

### Parameters

| Name | Meaning | Default | Notes |
|------|---------|---------|--------|
| `l` | Subsequence (window) length | (required) | Positive integer; each window has `l` consecutive samples. |
| `k` | Number of clusters | (required) | Integer ≥ 2. |
| `method` | Clustering algorithm | `kmeans` | Optional: `kmeans`, `kshape`, `medoidshape` (case-insensitive). Defaults to k-means if omitted. |
| `norm` | Z-score normalize each subsequence | `true` | Boolean; if `true`, each subsequence is standardized before clustering. |
| `maxiter` | Maximum iterations | `200` | Positive integer. |
| `output` | Output mode | `label` | `label`: one cluster id per window; `centroid`: concatenate the `k` centroid vectors in cluster order. |
| `sample_rate` | Greedy sampling rate | `0.3` | Used only when **`method` = `medoidshape`**; must be in `(0, 1]`. |

#### `method` details

- **kmeans**: k-means in Euclidean space (optionally after per-window normalization).
- **kshape**: Assign by shape-based distance (SBD from normalized cross-correlation, NCC); centroids updated via **SVD** on the cluster matrix.
- **medoidshape**: Coarse **k-means** with **`min(2k, number of windows)`** clusters, then greedy selection of `k` representative subsequences; `sample_rate` controls how many candidates are sampled each round.

### Output series

Controlled by `output`:

#### `output` = `label` (default)

- **One** output series, type **`INT32`**.
- Number of points = number of full windows, `⌊n/l⌋`.
- Timestamp of each point = **time of the first sample** in that window; value = cluster id **0 … k−1**.

#### `output` = `centroid`

- **One** output series, type **`DOUBLE`**.
- Number of points = **`k × l`**: for clusters **0 → k−1**, emit the `l` components of each centroid in order (concatenated).
- Timestamps are `0, 1, 2, …` (placeholders only, no physical time meaning).

### DIFF
--- a/library-udf/pom.xml
+++ b/library-udf/pom.xml
@@ -24,7 +24,7 @@
     <parent>
         <groupId>org.apache.iotdb</groupId>
         <artifactId>iotdb-parent</artifactId>
-        <version>2.0.7-SNAPSHOT</version>
+        <version>1.3.7-SNAPSHOT</version>
     </parent>
     <artifactId>library-udf</artifactId>
     <name>IoTDB: UDF</name>
@@ -41,7 +41,7 @@
         <dependency>
             <groupId>org.apache.iotdb</groupId>
             <artifactId>udf-api</artifactId>
-            <version>2.0.7-SNAPSHOT</version>
+            <version>1.3.7-SNAPSHOT</version>
         </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>
@@ -58,12 +58,10 @@
         <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-math3</artifactId>
-            <version>${commons-math3.version}</version>
         </dependency>
         <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
-            <version>${commons-lang3.version}</version>
         </dependency>
         <!-- FFT, IFFT, LowPass, HighPass -->
         <dependency>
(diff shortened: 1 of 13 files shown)


library-udf/pom.xml
waynextz, line 1: the version of dev/1.3 should be 1.3.x?

library-udf/src/assembly/tools/register-UDF.bat
waynextz, line 1: The bat files are not in sbin/winodws folder in dev/1.3 branch.

waynextz (commented):
two little issues

library-udf/pom.xml
suyx1999: Have restored the pom file.

library-udf/src/assembly/tools/register-UDF.bat
suyx1999: Have restored the bat file and added cluster registration.

PR #17439, opened 2026-04-07
[IOTDB-17433] Fix timezone offset bug in DateTimeUtils

Lexert19:
## Description

Refactored DateTimeUtils.toZoneOffset to calculate the timezone offset based on the provided string instead of the current system time (Instant.now()).
Added comprehensive unit tests in DateTimeUtilsTest to cover DST transitions.
Fixes #17433

<hr>

This PR has:
- [x] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/receiver/transform/converter/ValueConverter.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/receiver/transform/converter/ValueConverter.java
@@ -834,8 +834,7 @@ public class ValueConverter {
   private static int parseDateTimeToDate(final String value) {
     try {
       return DateUtils.parseDateExpressionToInt(
-          Instant.ofEpochMilli(
-                  DateTimeUtils.convertDatetimeStrToLong(value, ZoneOffset.UTC, 0, "ms"))
+          Instant.ofEpochMilli(DateTimeUtils.convertDatetimeStrToLong(value, ZoneOffset.UTC, "ms"))
               .atZone(ZoneOffset.UTC)
               .toLocalDate());
     } catch (final Exception e) {
--- a/iotdb-core/datanode/src/test/java/org/apache/iotdb/db/utils/DateTimeUtilsTest.java
+++ b/iotdb-core/datanode/src/test/java/org/apache/iotdb/db/utils/DateTimeUtilsTest.java
@@ -452,4 +452,18 @@ public class DateTimeUtilsTest {
               DateTimeUtils.toZoneOffset("2024-12-31 10", zoneId);
             });
   }
+
+  @Test
+  public void testToZoneOffsetHistoricalLMT() {
+    ZoneId zoneId = ZoneId.of("Europe/Warsaw");
+    ZoneOffset offset = DateTimeUtils.toZoneOffset("0001-01-01 00:00:00", zoneId);
+    Assert.assertEquals(ZoneOffset.ofHoursMinutes(1, 24), offset);
+
+    ZoneId shanghaiId = ZoneId.of("Asia/Shanghai");
+    ZoneOffset shanghaiOffset = DateTimeUtils.toZoneOffset("0001-01-01 00:00:00", shanghaiId);
+    Assert.assertEquals(ZoneOffset.ofHoursMinutes(8, 5), shanghaiOffset);
+
+    ZoneId utcId = ZoneId.of("UTC");
+    Assert.assertEquals(ZoneOffset.UTC, DateTimeUtils.toZoneOffset("0001-01-01 00:00:00", utcId));
+  }
 }
--- a/iotdb-core/datanode/src/test/java/org/apache/iotdb/db/utils/DateTimeUtilsTest.java
+++ b/iotdb-core/datanode/src/test/java/org/apache/iotdb/db/utils/DateTimeUtilsTest.java
@@ -28,12 +28,14 @@ import org.junit.Assert;
 import org.junit.Ignore;
 import org.junit.Test;
 
+import java.time.DateTimeException;
+import java.time.Instant;
 import java.time.ZoneId;
 import java.time.ZoneOffset;
-import java.time.ZonedDateTime;
 import java.util.TimeZone;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThrows;
 import static org.junit.Assert.fail;
 
 public class DateTimeUtilsTest {
@@ -48,7 +50,7 @@ public class DateTimeUtilsTest {
   /** Test convertDatetimeStrToLong() method with different time precision. */
   @Test
   public void convertDatetimeStrToLongTest1() {
-    zoneOffset = ZonedDateTime.now().getOffset();
+    zoneOffset = Instant.ofEpochMilli(timestamp).atZone(ZoneId.systemDefault()).getOffset();
     zoneId = ZoneId.systemDefault();
     if (zoneOffset.toString().equals("Z")) {
       delta = 8 * 3600000;
@@ -372,5 +374,82 @@ public class DateTimeUtilsTest {
 
     timeDuration = DataNodeDateTimeUtils.constructTimeDuration("10000000000ms");
     Assert.assertEquals(10000000000L, timeDuration.nonMonthDuration);
+  }
+
+  @Test
+  public void testToZoneOffsetForWinterTime() {
+    ZoneId zoneId = ZoneId.of("Europe/Warsaw");
+    ZoneOffset offset = DateTimeUtils.toZoneOffset("2024-01-15 12:00:00", zoneId);
+    Assert.assertEquals(ZoneOffset.ofHours(1), offset);
+  }
+
+  @Test
+  public void testToZoneOffsetForSummerTime() {
+    ZoneId zoneId = ZoneId.of("Europe/Warsaw");
+    ZoneOffset offset = DateTimeUtils.toZoneOffset("2024-06-15 12:00:00", zoneId);
+    Assert.assertEquals(ZoneOffset.ofHours(2), offset);
+  }
+
+  @Test
+  public void testToZoneOffsetJustBeforeSpringDST() {
+    ZoneId zoneId = ZoneId.of("Europe/Warsaw");
+    ZoneOffset offset = DateTimeUtils.toZoneOffset("2024-03-31 02:00:00", zoneId);
+    Assert.assertEquals(ZoneOffset.ofHours(1), offset);
+  }
+
+  @Test
+  public void testToZoneOffsetJustAfterSpringDST() {
+    ZoneId zoneId = ZoneId.of("Europe/Warsaw");
+    ZoneOffset offset = DateTimeUtils.toZoneOffset("2024-03-31 03:00:00", zoneId);
+    Assert.assertEquals(ZoneOffset.ofHours(2), offset);
+  }
+
+  @Test
+  public void testToZoneOffsetDuringSpringDSTGap() {
+    ZoneId zoneId = ZoneId.of("Europe/Warsaw");
+    ZoneOffset offset = DateTimeUtils.toZoneOffset("2024-03-31 02:30:00", zoneId);
+    Assert.assertEquals(ZoneOffset.ofHours(1), offset);
+  }
+
+  @Test
+  public void testToZoneOffsetDuringAutumnDSTTransition() {
+    ZoneId zoneId = ZoneId.of("Europe/Warsaw");
+    ZoneOffset offset = DateTimeUtils.toZoneOffset("2024-10-27 02:30:00", zoneId);
+    Assert.assertEquals(ZoneOffset.ofHours(2), offset);
+  }
+
+  @Test
+  public void testToZoneOffsetAfterAutumnDST() {
+    ZoneId zoneId = ZoneId.of("Europe/Warsaw");
+    ZoneOffset offset = DateTimeUtils.toZoneOffset("2024-10-27 03:00:00", zoneId);
+    Assert.assertEquals(ZoneOffset.ofHours(1), offset);
+  }
+
+  @Test
+  public void testToZoneOffsetWithExplicitOffsetInString() {
+    ZoneId zoneId = ZoneId.of("Europe/Warsaw");
+    ZoneOffset offset = DateTimeUtils.toZoneOffset("2024-06-15 12:00:00+02:00", zoneId);
+    Assert.assertEquals(ZoneOffset.ofHours(2), offset);
+    offset = DateTimeUtils.toZoneOffset("2024-06-15 12:00:00Z", zoneId);
+    Assert.assertEquals(ZoneOffset.UTC, offset);
+    offset = DateTimeUtils.toZoneOffset("2024-06-15 12:00:00-08:00", zoneId);
+    Assert.assertEquals(ZoneOffset.ofHours(-8), offset);
+  }
+
+  @Test
+  public void testToZoneOffsetWithUTCZoneId() {
+    ZoneId utc = ZoneId.of("UTC");
+    Assert.assertEquals(ZoneOffset.UTC, DateTimeUtils.toZoneOffset("2024-06-15 12:00:00", utc));
+  }
+
+  @Test
+  public void testToZoneOffsetWithBrokenDate() {
+    ZoneId zoneId = ZoneId.of("Europe/Warsaw");
+    DateTimeException exception =
+        assertThrows(
+            DateTimeException.class,
+            () -> {
+              DateTimeUtils.toZoneOffset("2024-12-31 10", zoneId);
+            });
   }
 }
--- a/iotdb-core/datanode/src/test/java/org/apache/iotdb/db/utils/DateTimeUtilsTest.java
+++ b/iotdb-core/datanode/src/test/java/org/apache/iotdb/db/utils/DateTimeUtilsTest.java
@@ -30,7 +30,6 @@ import org.junit.Test;
 
 import java.time.DateTimeException;
 import java.time.Instant;
-import java.time.LocalDateTime;
 import java.time.ZoneId;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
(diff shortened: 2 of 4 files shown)


Lexert19:
Changed solution: dropped toZoneOffset(String, ZoneId) to avoid double parsing. Now using parseBest + atZone(zoneId) to derive offset from input datetime. Tests updated.

Lexert19:
Rebased onto latest master to resolve conflicts. The `depth` parameter in `convertDatetimeStrToLong()` looks like an internal recursion guard and is never used externally — should we deprecate this overload and keep only the public version without `depth`?

PDGGK (commented):
I've been working on time-bucketed aggregation in the table model, and `DateTimeUtils`
sits directly underneath that layer, so I went through this one carefully. It has not
had a review yet, so this goes into some detail.

## The direction is right

`toZoneOffset()` flattening a `ZoneId` into `getRules().getOffset(Instant.now())` is a
real bug, not a style nit — it resolves the offset at *query time* instead of at the
instant being parsed. Reproduced against current master (`e23500e1a`) with the server
currently in CEST:

```
convertDatetimeStrToLong("2024-01-15 12:00:00", ZoneId.of("Europe/Warsaw"), "ms")
  master   1705312800000  ->  2024-01-15T11:00 local   (off by one hour)
  this PR  1705316400000  ->  2024-01-15T12:00 local
```

And the pre-standardisation case, which I didn't expect to also fall out correctly but
does:

```
convertDatetimeStrToLong("1900-01-01 00:00:00", ZoneId.of("Asia/Shanghai"), "ms")
  master   -2209017600000   (+08:00)
  this PR  -2209017943000   (+08:05:43, LMT)
```

This matters downstream of where I work: `date_bin`'s `origin` argument is folded into a
literal through exactly this path, so on master the same
`date_bin(1mo, time, 2024-01-01 00:00:00)` under `time_zone='Europe/Warsaw'` produces a
bucket start of `2024-05-01 00:00` when run in winter and `2024-04-30 23:00` when run in
summer. Same SQL, same data, different answer depending on the season the query is
issued. Your change makes it stable. Thanks for tracking this down.

## Things I checked that came back clean

Listing these so you know where I looked and didn't find anything, not just where I did:

- **No parse regressions on the common formats.** I extracted the formatter block plus
  the conversion logic from both `e23500e1a` and this branch into standalone classes and
  diffed them over a large input matrix (multiple zones × ms/us/ns × all six
  date/separator families × fractional widths 0–9). Every string both versions accept
  converts to the identical `long`, apart from the intended timezone correction.
- **Explicit offsets are untouched.** `parseBest(OffsetDateTime::from, LocalDateTime::from)`
  keeps in-string offsets winning over the session zone; `2024-06-15 12:00:00Z` and
  `...+05:00` are byte-identical before and after.
- **Optional ordering is safe.** The six new local optionals are appended after all
  eighteen offset optionals, so they can't steal a prefix, and trailing garbage
  (`2024-06-15T12:00:00 x`, `...+01`) is still rejected — `parse` requires full
  consumption.
- **Static init order is fine.** The new static block at `:490` runs after
  `ISO_LOCAL_DATE_WIDTH_1_2` / `_WITH_SLASH` / `_WITH_DOT` / `ISO_LOCAL_TIME_WITH_NS` and
  before `formatter` at `:534`. No NPE at class load — the harness compiles and runs.
- **Only the NS variant for the local formatters is correct**, not a coverage gap:
  `appendFraction(NANO_OF_SECOND, 0, 9, true)` is a strict superset of the ms(0–3) and
  us(0–6) variants.
- **Overload resolution doesn't break anything.** `ZoneOffset` is more specific than
  `ZoneId`, so existing `ZoneOffset` call sites bind to your new deprecated overloads, and
  `(ZoneId) offset` is semantically identical to the old `toZoneOffset(zoneOffset)` for a
  fixed offset. The root `pom.xml` has no `-Werror` / `-Xlint:deprecation` /
  `showDeprecation`, so the new `@Deprecated` annotations can't turn the build red.
- **Recursion still terminates without the guard.** The `Z` branch strips from the first
  `Z` onward so the result can't contain `Z`; the `length()==10` branch yields length 19.
  Max two frames.
- **No rebase drift.** Your merge base is `7563ac8f` (2026-05-20) and master hasn't touched
  either file since, so `git diff` on both is empty against current master.

## The one thing I'd want a second opinion on: the new tests don't fail on master

This is the finding I'd most want you to look at, and it's about the tests rather than the
fix.

Eight of the nine new tests assert on the offset the zone has *at the
resulting instant*, rather than on the instant itself. Six use this shape
literally:

```java
assertEquals(ZoneOffset.ofHours(1), zoneId.getRules().getOffset(Instant.ofEpochMilli(winter)));
```

`springGapShouldShiftForwardInWarsaw` uses it alongside a second assertion, and
`historicalDateBeforeStandardizedOffsetShouldUseLMT` does the same thing via
`offset.getTotalSeconds()`. Only `explicitOffsetInStringOverridesZoneId` asserts
something else — it compares two conversions to each other.

`getRules().getOffset(instant)` is a property of the *zone* — it only reports which DST
segment the result landed in. Master's error is exactly one hour, and DST segments are
months long, so the wrong instant lands in the same segment and the assertion still holds.

The point does not depend on running anything: it is visible in what the
assertions say. Every one of them asks the JDK's tz database which offset
applies to the instant that was computed. None of them compares that instant
to an expected timestamp. So an assertion of this shape cannot separate a
correct result from one an hour off, because both land inside the same
months-long bracket.

`convertWinterTimeShouldUseUtcPlus1` is the clearest case. On master the
conversion yields `2024-01-15T10:00:00Z` where the correct answer is
`11:00:00Z` — exactly the bug this PR fixes — but `10:00Z` is still `+01:00`
in Warsaw, so the assertion holds and the test goes green.
`historicalDateBeforeStandardizedOffsetShouldUseLMT` never discriminates at
all: both `-2209017600000` and `-2209017943000` sit inside Shanghai's LMT
segment.

As corroboration — and I should be clear this is an extraction rather than a
run against real master — I lifted master's conversion logic into a harness
and pinned `toZoneOffset(Instant.now())` to each of the two values it can take for Warsaw:

```
master logic, now = +02:00   ->  8 PASS / 1 FAIL   (only springGapShouldShiftForwardInWarsaw)
master logic, now = +01:00   ->  7 PASS / 2 FAIL   (autumnOverlap, explicitOffsetInStringOverridesZoneId)
this PR                      ->  9 PASS / 0 FAIL
```

If you want to confirm that on real master rather than on my extraction:
reverting just the `DateTimeUtils` change and re-running the new tests should
leave most of them passing. That is a two-minute check and it settles the
question better than my harness does.

So at no point in the year would this suite fail on more than two of the nine, and *which*
two flips with the calendar. The clearest example is `convertWinterTimeShouldUseUtcPlus1`:
master returns `2024-01-15T10:00:00Z` where the right answer is `11:00:00Z` — exactly the
bug you're fixing — but `10:00Z` is still `+01:00` in Warsaw, so it goes green.
`historicalDateBeforeStandardizedOffsetShouldUseLMT` is the one that never discriminates at
all: both `-2209017600000` and `-2209017943000` sit inside Shanghai's LMT segment.

Asserting the epoch directly makes every one of them discriminating, and stays correct for
the gap/overlap cases too, since `ZonedDateTime.of` uses the same resolution rules your
implementation does:

```java
assertEquals(
    ZonedDateTime.of(LocalDateTime.parse("2024-01-15T12:00:00"), zoneId).toInstant().toEpochMilli(),
    winter);
```

Is the offset-window form deliberate — e.g. you wanted the assertions to stay readable as
"this instant is in winter time" — or would switching to epochs be fine with you?

## The accepted-literal syntax shifted in three places

Removing the `str + offset` fallback means unqualified strings now go through the six new
local formatters instead of being rewritten into offset-bearing ones. That changed which
literals are accepted, in both directions. Everything below is measured, session zone
`Europe/Warsaw` for the ms values:

| input | master | this PR |
|---|---|---|
| `2024-6-5 12:00:00` | rejected | `1717581600000` |
| `2024-6-5 12:00:00+02:00` | rejected | rejected |
| `2024-6-5T12:00:00` / `2024/6/5 12:00:00` / `2024.6.5 12:00:00` | `1717581600000` | `1717581600000` |
| `2024-06-15 12:00:00z` (lowercase) | rejected | `1718452800000` (UTC) |
| `2024-06-15 12:00:00` | `1718445600000` | `1718445600000` |
| `-0004-06-15 12:00:00` | parses | **rejected** |
| `-0004-06-15 12:00:00+01:00` | parses | parses |
| `+12024-06-15 12:00:00` | parses | **rejected** |
| `12024-06-15 12:00:00` | rejected | parses |

Three separate things going on:

**1. One-or-two-digit month/day with a space separator is now accepted.** This actually
makes the class *more* consistent, not less: master already accepted `2024-6-5T12:00:00`,
`2024/6/5 12:00:00` and `2024.6.5 12:00:00`. Hyphen-plus-space was the odd one out, because
`ISO_OFFSET_DATE_TIME_WITH_SPACE{,_US,_NS}` (`:361-395`) are the only formatters in the file
that use the JDK's `DateTimeFormatter.ISO_LOCAL_DATE` instead of the local
`ISO_LOCAL_DATE_WIDTH_1_2`. Your new `ISO_LOCAL_DATE_TIME_WITH_NS_SPACE` (`:512-517`) uses
`WIDTH_1_2`, which lines it up with the majority. The residual oddity is that
`2024-6-5 12:00:00` now works but `2024-6-5 12:00:00+02:00` still doesn't — same date shape,
different answer depending on whether an offset is attached.

**2. Signed and >4-digit years with a space separator stopped parsing.** Same one-line cause
seen from the other side: `WIDTH_1_2` is `appendValue(YEAR, 4, 19, SignStyle.NEVER)`, so it
won't take a leading `-` or `+`, while `DateTimeFormatter.ISO_LOCAL_DATE` is
`SignStyle.EXCEEDS_PAD` and will. So `-0004-06-15 12:00:00` parsed on master and now throws,
while `-0004-06-15 12:00:00+01:00` still parses on this branch — the local path is narrower
than the offset path for the same date shape. To be clear about how exotic this is: master's
own `T` form (`-0004-06-15T12:00:00`) already rejected it, so master's acceptance was itself
an accident of that one formatter family, and BC dates in a time-series DB are not a
realistic workload. I'd call it a consistency wart rather than a regression. If you wanted
both shapes, `appendOptional`-ing the two date variants into
`ISO_LOCAL_DATE_TIME_WITH_NS_SPACE` covers it; swapping to `ISO_LOCAL_DATE` outright would
also work but would give back the one-digit month/day case from (1).

**3. Lowercase `z` is now accepted as UTC.** `2024-06-15 12:00:00z` was rejected on master
and now returns `1718452800000` — i.e. UTC, ignoring the session zone, two hours away from
how the same string without the `z` would be read. This one isn't from the new formatters;
it's because a bare trailing `z` used to be pushed through the `str + offset` rewrite and
now reaches `appendOffsetId()`'s `noOffsetText` under `parseCaseInsensitive`. Arguably an
improvement (lowercase `t` was always accepted), just worth knowing it happened.

None of these are in the PR description and none have test coverage. Were they intended, or
incidental to removing the fallback?

## `depth` and `DATETIME_CONVERT_FAILED` — I think this answers your May question

You asked back in May whether the `depth` parameter should be deprecated and nobody replied.
I think the answer after this PR is stronger than "should": it's now unreachable. The deleted
branch was the only `depth + 1` in the file — on this branch, `git grep depth` on
`DateTimeUtils.java` returns `:623`, `:624` (the deprecated overload forwarding it), `:676`,
`:677` (the guard), `:683` and `:685` (both recursive calls passing it through unchanged).
Nothing increments it, and every public entry point hardcodes `0`, so `if (depth >= 2)` at
`:677` can't be true for any string input.

That makes the message at `:679` unreachable too. `DATETIME_CONVERT_FAILED` has exactly three
hits repo-wide: the `en` and `zh` definitions at `QueryMessages.java:230` and that one throw.
Worth noting it *was* reachable on master — any 4-character input works, e.g. `abcd` →
`abcd+08:00` (length 10) → `abcd+08:00T00:00:00` → second increment → the localised
`"Failed to convert %s to millisecond, zone offset is %s, please input like
2011-12-03T10:15:30..."`. On this branch the same input surfaces a raw
`DateTimeParseException` instead. Both are errors and arguably the new one is more precise,
but two localised strings quietly become orphans in the process. Was dropping that message
deliberate?

## Two small ones

- `toZoneOffset` at `:747` is marked `@Deprecated()` but kept `public static`, and it now has
  zero callers anywhere in the repo — it's the sole grep hit. Since it's the exact trap this
  issue is about, a `@deprecated` javadoc line saying "returns the offset at the *current*
  instant, don't use it to convert other timestamps" would stop the next person rediscovering
  it. (Also `@Deprecated()` → `@Deprecated`, and the other three deprecations in this diff
  carry a `Use {@link ...} instead`, this one doesn't.)
- The six new formatters don't call `.parseCaseInsensitive()`, unlike all eighteen offset
  formatters. They still parse lowercase `t` today, but only because `caseSensitive` lives on
  `DateTimeParseContext` and `endOptional()` doesn't roll it back — the first offset optional's
  settings parser leaks into them. Verified: pulled out on its own,
  `ISO_LOCAL_DATE_TIME_WITH_NS_T` rejects `2024-06-15t12:00:00`; inside the composite it
  accepts it; move the local optionals ahead of the offset ones and it breaks again. Purely a
  robustness thing, no user-visible effect today.

## Overall

The fix itself I couldn't fault — I went looking for a numeric regression across the format
matrix and didn't find one, and the `date_bin` origin case genuinely gets better. My only
real reservation is that the test suite doesn't currently hold the fix in place, which is a
shame given how easy the bug would be to reintroduce.

For what it's worth: I'm not a committer, so this is one contributor's read, not an approval.
Two procedural notes for whoever picks it up — CI has never run on this PR (fork PRs need a
committer to approve the workflow, `statusCheckRollup` is empty), and the only existing review
is from the Copilot bot against the pre-rebase version. The branch is still `MERGEABLE` and
neither touched file has moved on master since your base commit, so nothing has rotted.
```

---

**Comment by Lexert19** [comment]
Hi @PDGGK,
thank you for the detailed review! I have addressed all your points: I updated the tests to assert on exact epoch milliseconds, unified the formatters using `ISO_LOCAL_DATE_WIDTH_1_2 `so parsing behavior is consistent, added tests to codify the new edge-case syntax, cleaned up the unused depth logic and `DATETIME_CONVERT_FAILED `message, completely deleted the problematic `toZoneOffset `method, and added .parseCaseInsensitive() to the new formatters.

Hi @Wei-hao-Li, since you were requested as a reviewer on this PR, could you please approve running the CI workflow and give this PR a review when you get a chance?

**Review by PDGGK — COMMENTED** [review]
Thanks for turning this around so quickly. Read against `2b3ff81b`; this is
source reading rather than a run, since CI has not executed on this PR yet.

The new approach is the right one. Deleting `toZoneOffset` removes the actual
defect rather than working around it: `zoneId.getRules().getOffset(Instant.now())`
resolved the offset for *now* and then applied it to whatever instant was being
parsed, which is why the result depended on when the code ran rather than on the
timestamp. `formatter.parseBest(str, OffsetDateTime::from, LocalDateTime::from)`
followed by `atZone(zoneId)` resolves the offset for the instant in question,
which is the thing that was missing. Dropping the `depth` recursion and the
`str + offset` re-parse falls out of that naturally.

The tests now compare against expected epoch milliseconds instead of asking the
tz database which offset applies, which is the change that lets them fail.

Checked and clear: whether the format coverage narrowed
-------------------------------------------------------

The removed branch appended the offset to *any* string that lacked one, so it
implicitly covered every local-form input the formatter understood. The six new
formatters cover a fixed list instead, so the question is whether anything that
used to parse now does not.

It does not. `ISO_LOCAL_TIME_WITH_NS` is built with
`appendFraction(NANO_OF_SECOND, 0, 9, true)` inside an optional section, so it
accepts zero through nine fractional digits -- it subsumes the MS and US
variants rather than sitting alongside them. And the six formatters are
{`ISO_LOCAL_DATE_WIDTH_1_2`, `ISO_LOCAL_DATE_WITH_SLASH`,
`ISO_LOCAL_DATE_WITH_DOT`} x {`T`, space}, which is the same date-style by
separator matrix the offset-bearing formatters use. So the local forms and the
offset forms accept the same shapes. Recording it because it is the kind of
narrowing that would not show up in the new tests.

Two things on the tests
-----------------------

1. `historicalDateBeforeStandardizedOffsetShouldUseLMT` still computes

   ```java
   ZoneOffset offset = shanghaiId.getRules().getOffset(Instant.ofEpochMilli(oldTime));
   ```

   and never uses it -- `expected` comes from `ZonedDateTime.of(...)` and the
   assertion is against that. It is a leftover from the previous version of the
   test, and it is specifically the `getRules().getOffset(...)` call the rewrite
   was meant to stop asserting through, so it reads as if it were still load
   bearing. Worth deleting.

2. `convertWinterTimeShouldUseUtcPlus1` and `convertSummerTimeShouldUseUtcPlus2`
   look redundant and are not. Which of the two catches a regression depends on
   the time of year the suite runs, because the old code took its offset from
   `Instant.now()`:

   - run in July, Warsaw is +02:00 now, so the winter case converts
     `2024-01-15 12:00` as +02:00 and lands an hour off -- winter fails, summer
     passes.
   - run in January, Warsaw is +01:00 now, so the summer case is the one that
     lands an hour off -- summer fails, winter passes.

   Exactly one of the pair fails at any given time of year, and which one
   changes. The same holds for `convertJustBeforeSpringDstShouldKeepWinterOffset`
   and `convertJustAfterSpringDstShouldUseSummerOffset`, for the same reason.
   That is reasoning from the deleted implementation rather than something I
   ran, with the offset-suffix arithmetic on that branch and the Warsaw offsets
   both checked. Neither half of either pair can be dropped as a duplicate
   without losing detection for part of the year. A comment saying so
   would probably save them from a future cleanup.

On the deprecated overload
--------------------------

The four-argument `convertDatetimeStrToLong(String, ZoneOffset, int, String)` is
kept as `@Deprecated` and ignores `depth`. After this PR the test file no longer
calls it at all -- all 38 call sites there are on the three-argument form -- so
in the current tree it has exactly one caller left:

`ValueConverter.java:838`, inside `parseDateTimeToDate` in the pipe receiver:

```java
DateTimeUtils.convertDatetimeStrToLong(value, ZoneOffset.UTC, 0, "ms")
```

That one passes `depth = 0`, which is the value the parameter being ignored
cannot affect, so this PR does not change its behaviour. The part worth deciding
is what happens to anyone who passes a non-zero `depth`: today that is a
`DateTimeException` on the second pass, and afterwards it is silently accepted
and ignored. If keeping the overload is for out-of-tree compatibility that seems
reasonable, but it might be worth making the parameter throw rather than be
inert, or updating that single production call site and removing the overload
outright.

Lexert19:
Hi @PDGGK, thanks for the review!

Addressed your latest points and pushed the fixes:
- Removed the unused line in the LMT test and added comments to the seasonal DST test pairs.
- Updated `ValueConverter.java` and removed the deprecated overload entirely.
- Fixed bare date parsing (e.g. `2014-4-4`) by replacing `str.length() == 10` with `!str.contains(":")`, along with new unit tests.

PR #17469, opened 2026-04-14
Fix: send set configuration only to target nodes and harden compaction schedule interruption handling

shuwenwei:
## Summary
- Fix `set configuration ... on <nodeId>` so ConfigNode only forwards the request to the targeted DataNode instead of all registered DataNodes
- Harden compaction/TTL schedule workers so user-triggered stop interrupts exit cleanly while unexpected interruptions are logged and tolerated
- Isolate object-file TTL check failures and extend integration coverage for targeted DataNode configuration updates
## Testing
- Updated `IoTDBSetConfigurationTableIT` to cover DataNode-scoped configuration changes

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/schedule/CompactionScheduleTaskWorker.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/schedule/CompactionScheduleTaskWorker.java
@@ -81,7 +81,7 @@ public class CompactionScheduleTaskWorker implements Callable<Void> {
         if (isStoppedByUser) {
           return null;
         }
-      } catch (Exception e) {
+      } catch (Throwable e) {
         logger.error(
             "[CompactionScheduleTaskWorker-{}] Failed to execute compaction schedule task",
             workerId,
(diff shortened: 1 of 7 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/schedule/CompactionScheduleTaskWorker.java
```diff
@@ -72,9 +72,20 @@ public Void call() {
...
         logger.info(
-            "[CompactionScheduleTaskWorker-{}] compaction schedule is interrupted", workerId);
-        return null;
+            "[CompactionScheduleTaskWorker-{}] compaction schedule is interrupted, isStopByUser: {}",
+            workerId,
+            isStoppedByUser);
+        if (isStoppedByUser) {
+          return null;
+        }
+      } catch (Exception e) {
```
JackieTien97, line 84:
```suggestion
      } catch (Throwable e) {
```

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/schedule/TTLScheduleTask.java
```diff
@@ -78,8 +78,17 @@ public Void call() throws Exception {
...
+        boolean isStoppedByUser =
+            CompactionScheduleTaskManager.getInstance().isStoppingAllScheduleTask();
+        logger.info(
+            "[TTLCheckTask-{}] TTL checker is interrupted, isStoppedByUser: {}",
+            workerId,
+            isStoppedByUser);
+        if (isStoppedByUser) {
+          return null;
+        }
+      } catch (Exception e) {
```
JackieTien97, line 90:
```suggestion
      } catch (Throwable e) {
```

JackieTien97 (changes requested):

JackieTien97 (approved):

PR #17481, opened 2026-04-15
Add REST write request size and batch limits

HTHou:
## Summary
- add configurable REST request limits for body size, row count, column count, and total values
- reject oversized request bodies early with a pre-matching filter and return HTTP 413 for per-request body limit violations
- add a REST in-flight request body memory quota and return HTTP 503 when concurrent request bodies exceed the quota
- align the REST concurrent request body budget with Thrift `AutoResizingBuffer` memory control: `0` uses the same auto budget calculation, while a negative value disables the total quota
- enforce write batch limits across REST v1, v2, and table APIs during request validation
- update REST exception handling to preserve 413 responses instead of always returning HTTP 200
- document the new REST configuration properties in system property templates and test resources

## Testing
- `mvn spotless:apply -pl external-service-impl/rest`
- `mvn test -pl iotdb-core/datanode -am -Dtest=DataNodeMemoryConfigTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false`
- `mvn test -pl external-service-impl/rest -am -Dtest=RequestSizeLimitFilterTest,RequestValidationLimitTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false`
- `mvn test-compile -P with-zh-locale -pl external-service-impl/rest -am -DskipTests`


external-service-impl/rest/src/test/java/org/apache/iotdb/rest/protocol/handler/RequestValidationLimitTest.java
```diff
@@ -0,0 +1,108 @@
...
+    request.setDeviceId("root.sg.d1");
+    request.setIsAligned(false);
+    request.setMeasurements(Collections.singletonList("s1"));
+    request.setDataTypes(Collections.singletonList("INT64"));
+    request.setTimestamps(Arrays.asList(1L, 2L, 3L));
+    request.setValues(Collections.singletonList(Arrays.asList(1L, 2L, 3L)));
+
+    org.apache.iotdb.rest.protocol.v1.handler.RequestValidationHandler.validateInsertTabletRequest(
+        request);
+  }
```
jt2594838, line 68: Better to use short names.

external-service-impl/rest/src/main/java/org/apache/iotdb/rest/protocol/filter/RequestSizeLimitFilter.java
```diff
@@ -0,0 +1,98 @@
...
+      return;
+    }
+
+    requestContext.setEntityStream(
+        new LimitedInputStream(requestContext.getEntityStream(), maxBodySize));
+  }
+
+  private static Response buildPayloadTooLargeResponse(long maxBodySize) {
+    return Response.status(413)
+        .type(MediaType.TEXT_PLAIN_TYPE)
```
Caideyipi: `RequestSizeLimitFilter` returns this 413 as `text/plain`, but the filter runs before the resource methods and the per-version `ExceptionHandler`. The new row/column/value limit path returns a JSON `ExecutionStatus` with code 413, and the REST OpenAPI/clients/tests consume REST responses as JSON `ExecutionStatus`.

This makes oversized request bodies the one limit-failure path that can break clients which parse REST errors as JSON. Could we return `application/json` here with an `ExecutionStatus` entity, e.g. `{code: 413, message: ...}`, and add coverage for both the `Content-Length` abort path and the streamed/chunked overflow path?

external-service-impl/rest/src/main/i18n/en/org/apache/iotdb/rest/i18n/RestMessages.java
```diff
@@ -0,0 +1,126 @@
...
+  public static final String ROW_VALUES_SIZE_MISMATCH =
+      "The number of values in the %dth row is not equal to the data_types size";
+  public static final String ERROR_MESSAGE_SEPARATOR = ",";
+
+  // --- RequestSizeLimitFilter ---
+  public static final String MESSAGE_REST_REQUEST_BODY_EXCEEDS_LIMIT_ARG_BYTES_D9F9B412 =
+      "REST request body exceeds limit %d bytes";
+  public static final String
+      MESSAGE_REST_REQUEST_BODY_MEMORY_QUOTA_EXCEEDS_LIMIT_ARG_BYTES_7F2994D9 =
+          "REST request body memory quota exceeds limit %d bytes";
```
jt2594838: Indicate which configuration should be modified to avoid this.
Even better to give the `SET CONFIGURATION` statement directly.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/conf/rest/IoTDBRestServiceDescriptor.java
```diff
@@ -96,6 +99,17 @@ private void loadProps(TrimProperties trimProperties) {
...
+        DataNodeMemoryConfig.calculateAutoResizingBufferMemorySizeInBytes(trimProperties);
+    long maxTotalConcurrentRequestBodySizeInBytes =
+        Long.parseLong(
+            trimProperties.getProperty(
+                "rest_max_total_concurrent_request_body_size_in_bytes",
+                Long.toString(defaultMaxTotalConcurrentRequestBodySizeInBytes)));
+    conf.setRestMaxTotalConcurrentRequestBodySizeInBytes(
+        maxTotalConcurrentRequestBodySizeInBytes == 0
+            ? defaultMaxTotalConcurrentRequestBodySizeInBytes
+            : maxTotalConcurrentRequestBodySizeInBytes);
```
jt2594838: Better to support hot-reload.

jt2594838 (approved):

PR #17491, opened 2026-04-15
Fix can not use currentGeneratorFuture and currentGenerator to get current progress if DN is restarted before the data partition generation complete

zerolbsony:
Avoid to submit duplicate procedue;
Resolve the problem that occurs "Pool not open" again when stoping ConfigNode;
also, don't submit duplicate instances when a procedure isn't yet finished;
Fix can not use currentGeneratorFuture and currentGenerator to get current progress if DN is restarted before the data partition generation complete.

image


CRZbulabula (approved):
LGTM

PR #17511, opened 2026-04-17
Fix Integer overflow when converting 604800000000000ms to -3942554432415203328ns

zerolbsony:
timePartitionInterval is inited use 604800000 * 1000000 by ns in the DataNode, and 604800000 * 1000000 again after loadGlobalConfig executed.


CRZbulabula (approved):
LGTM

PR #17524, opened 2026-04-20
[To dev/1.3] Pipe: Rewrote the OPC UA subscription logic to avoid the bug of third-party subscription model

Caideyipi:
## Description
Now the original subscriptionModel has some problems:
1. The subscription logic for node is async, timely pull, and won't pull the inital value directly. This may cause plenty time, if a node's update frequency is low, the latency may be large to pull the first value.
2. In subscription model, one sample failure will cause permanent halt to the whole task.
3. And may be other problems.
Thus, this PR rewrote the subscription logic, and changed the subscription value on push (with some performance optimization) to avoid "badWaitingForInitialData" after long term running.

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/sink/protocol/opcua/server/OpcUaNameSpace.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/sink/protocol/opcua/server/OpcUaNameSpace.java
@@ -72,6 +72,8 @@ import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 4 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/sink/protocol/opcua/server/OpcUaNameSpace.java
```diff
@@ -451,24 +468,131 @@ public static NodeId convertToOpcDataType(final TSDataType type) {
...
+                      // into the thread pool, not blocking your data update thread
+                      for (DataItem item : subscribedItems) {
+                        try {
+                          item.setValue(newValue);
+                        } catch (Exception e) {
+                          // Single client push failure does not affect other clients
+                          LOGGER.warn(
+                              "Failed to push value change to client, nodeId={}", nodeId, e);
+                        }
+                      }
```
jt2594838, line 518: How about adding some retry?

jt2594838 (approved):

PR #17551, opened 2026-04-27
Clean up dead RPC thread config and use node-specific selectorNum

JackieTien97:
## Summary

This PR cleans up dead/misconfigured RPC thread parameters, ensures ConfigNode and DataNode each use their own `selectorNumOfClientManager` value when constructing async client pools, and adds `maxIdleClientNumForEachNode` support for client pool eviction control.

### 1. Fix config naming mismatch

- Rename `dn_selector_thread_count_of_client_manager` → `dn_selector_thread_nums_of_client_manager` in both the properties template and `IoTDBDescriptor` to match the name already used in `CommonDescriptor`.

### 2. Remove dead config parameters

- **`dn_rpc_selector_thread_count`**: Parsed and stored in `IoTDBConfig.rpcSelectorThreadCount` but never actually consumed by any server component. Removed the field, getter, setter, template entry, and loading code in `IoTDBDescriptor`.
- **`coordinator_write_executor_size`**: Created a `writeOperationExecutor` thread pool in `Coordinator` that was passed to `TreeModelPlanner` but never referenced in any of its methods — completely dead code. Removed the config field, the thread pool, the `TreeModelPlanner` constructor parameter, and the `MPP_COORDINATOR_WRITE_EXECUTOR` thread name.

### 3. Fix `ClientPoolProperty.Builder` hardcoded default

- The `Builder.maxClientNumForEachNode` was initialized to the compile-time constant `DefaultProperty.MAX_CLIENT_NUM_FOR_EACH_NODE` (1000), ignoring any runtime config override from `dn_max_client_count_for_each_node_in_client_manager` or `cn_max_client_count_for_each_node_in_client_manager`. Changed to read from `CommonDescriptor.getInstance().getConfig().getMaxClientNumForEachNode()` so callers that don't explicitly call `setMaxClientNumForEachNode()` still respect the configured value.

### 4. Use node-specific `selectorNum` for async client pools

Previously, `AsyncDataNodeInternalServiceClientPoolFactory` always read `selectorNumOfClientManager` from `CommonConfig`, which is shared. ConfigNode and DataNode may configure different values via `cn_selector_thread_nums_of_client_manager` and `dn_selector_thread_nums_of_client_manager` respectively.

Changes:
- **`AsyncDataNodeInternalServiceClientPoolFactory`**: Removed the no-arg constructor; the only constructor now requires an explicit `selectorNumOfAsyncClientManager` int.
- **`Coordinator`**: Passes `IoTDBConfig.getSelectorNumOfClientManager()` when constructing the factory.
- **`DataNodeInternalServiceRequestManager`**: Added a `final int selectorNumOfAsyncClientManager` field and a `protected` constructor so each subclass supplies its own value.
- **`DnToDnInternalServiceAsyncRequestManager`**: Reads from `IoTDBConfig.getSelectorNumOfClientManager()`.
- **`CnToDnInternalServiceAsyncRequestManager`**: Reads from `ConfigNodeConfig.getSelectorNumOfClientManager()`.
- **`ConfigNodeConfig`**: Added `selectorNumOfClientManager` field with getter/setter.
- **`ConfigNodeDescriptor`**: Loads `cn_selector_thread_nums_of_client_manager` into `ConfigNodeConfig`.

### 5. Auto-compute selector threads and add idle client eviction control

- **`selector_thread_nums` default changed to 0 (auto-compute)**: Both `cn_selector_thread_nums_of_client_manager` and `dn_selector_thread_nums_of_client_manager` now default to `0` in the properties template. When `<= 0`, the system auto-computes the value as `max(1, CPU core number / 4)`. The descriptor files (`CommonDescriptor`, `ConfigNodeDescriptor`, `IoTDBDescriptor`) only update the config when the parsed value is `> 0`.
- **`ThriftClientProperty` default updated**: Changed from hardcoded `1` to `max(1, availableProcessors() / 4)` to match the auto-compute logic.
- **`ConfigNodeConfig` / `IoTDBConfig` defaults updated**: Both now use `max(1, availableProcessors() / 4)` as the default `selectorNumOfClientManager`.
- **`maxIdleClientNumForEachNode` support added**: New config parameter `cn_max_idle_client_count_for_each_node_in_client_manager` / `dn_max_idle_client_count_for_each_node_in_client_manager` (default `0`, meaning no idle eviction limit). Added the field and getter/setter to `CommonConfig`, `ConfigNodeConfig`, and `IoTDBConfig`. `ClientPoolProperty.Builder` now reads this value from `CommonDescriptor` and uses it to set `maxIdlePerKey` (previously hardcoded to `maxClientNumForEachNode`). Negative values in the properties file are ignored.

## Test plan

- [ ] Verify the cluster starts successfully with default config (1C1D)
- [ ] Verify custom `dn_selector_thread_nums_of_client_manager` / `cn_selector_thread_nums_of_client_manager` values take effect (only when > 0)
- [ ] Verify `dn_max_client_count_for_each_node_in_client_manager` is respected by client pools that don't explicitly set it
- [ ] Verify `dn_max_idle_client_count_for_each_node_in_client_manager` / `cn_max_idle_client_count_for_each_node_in_client_manager` controls idle client eviction (0 = no limit, negative values ignored)
- [ ] Verify selector thread count defaults to `max(1, CPU/4)` when set to 0


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/client/DataNodeClientPoolFactory.java
```diff
@@ -63,34 +63,6 @@ public GenericKeyedObjectPool<ConfigRegionId, ConfigNodeClient> createClientPool
...
-
-    @Override
-    public GenericKeyedObjectPool<ConfigRegionId, ConfigNodeClient> createClientPool(
-        ClientManager<ConfigRegionId, ConfigNodeClient> manager) {
-      GenericKeyedObjectPool<ConfigRegionId, ConfigNodeClient> clientPool =
-          new GenericKeyedObjectPool<>(
-              new ConfigNodeClient.Factory(
-                  manager,
-                  new ThriftClientProperty.Builder()
-                      .setConnectionTimeoutMs(CONF.getConnectionTimeoutInMS() * 10)
```
CRZbulabula, line 77: This factory set `ConnectionTimeoutMs` to **10x** the normal value (`CONF.getConnectionTimeoutInMS() * 10`, i.e. ~600s vs 60s) for schema-modification RPCs (`deactivateSchemaTemplate`, `deleteTimeSeries`, `dropTable`, `deleteDevice`, `alterLogicalView`, etc.).

All ~15 call sites in `ClusterConfigTaskExecutor` now use the standard `CONFIG_NODE_CLIENT_MANAGER` with a 60s timeout. This could cause timeout failures for long-running schema operations on large clusters.

If the intent is to remove the separate pool, the base connection timeout should be validated as sufficient for these operations, or the timeout should be extended on these specific RPC calls.

iotdb-core/node-commons/src/assembly/resources/conf/iotdb-system.properties.template
```diff
@@ -505,6 +506,14 @@ cn_selector_thread_nums_of_client_manager=1
 # Datatype: int
 cn_max_client_count_for_each_node_in_client_manager=1000
 
+# The maximum number of idle clients that can be retained for a node in a clientManager.
+# When the number of idle clients to a single node exceeds this number, excess idle clients will be evicted.
+# Idle clients are determined by a time threshold (default 1 minute of inactivity).
+# 0 means no idle clients will be retained, connections are destroyed immediately upon return.
```
CRZbulabula, line 512: The PR description says:
> default `0`, meaning no idle eviction limit

But the template defaults to `1000` (see the commented-out value below), and `GenericKeyedObjectPool.setMaxIdlePerKey(0)` actually means **destroy all idle connections immediately** — which is the opposite of "no limit."

The template comment here is correct; the PR description should be fixed to avoid confusion.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/client/property/ClientPoolProperty.java
```diff
@@ -49,7 +51,11 @@ public static class Builder<V> {
      * the maximum number of clients that can be allocated for a node. When some clients are idle
      * for more than {@code maxIdleTimeForClient}, they will be cleaned up.
      */
-    private int maxClientNumForEachNode = DefaultProperty.MAX_CLIENT_NUM_FOR_EACH_NODE;
+    private int maxClientNumForEachNode =
+        CommonDescriptor.getInstance().getConfig().getMaxClientNumForEachNode();
```
CRZbulabula, line 55: Reading from `CommonDescriptor.getInstance()` at field-initialization time couples this Builder to the singleton's lifecycle. If a `Builder` is ever constructed before `CommonDescriptor` has loaded its properties (e.g., during static init of an early-loaded class), it silently falls back to the field's Java default (1000) — which happens to match the old hardcoded constant, so it works today.

Consider keeping the compile-time constant as the field default and relying on callers or a factory method to inject the configured value, which avoids the hidden ordering dependency.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/conf/IoTDBDescriptor.java
```diff
@@ -291,11 +291,23 @@ public void loadProperties(TrimProperties properties) throws BadNodeUrlException
...
             properties.getProperty(
-                "dn_selector_thread_count_of_client_manager",
-                String.valueOf(conf.getSelectorNumOfClientManager()))));
+                "dn_max_idle_client_count_for_each_node_in_client_manager",
+                String.valueOf(conf.getMaxIdleClientNumForEachNode())));
+    if (dnMaxIdleClientNumForEachNode >= 0) {
+      conf.setMaxIdleClientNumForEachNode(dnMaxIdleClientNumForEachNode);
+    }
+
+    int dnSelectorNumOfClientManager =
```
CRZbulabula, line 303: Minor: `CommonConfig` and `ConfigNodeConfig` both change their `selectorNumOfClientManager` field default to `max(1, CPU/4)`, but `IoTDBConfig.selectorNumOfClientManager` doesn't appear to be updated in this PR (the field itself is not in the diff).

The behavior is correct because the `> 0` guard here preserves the existing default when the template value is 0. But the field initialization in `IoTDBConfig` is inconsistent with the other two config classes — worth aligning to `max(1, CPU/4)` for clarity.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/client/property/ClientPoolProperty.java
```diff
@@ -87,7 +98,7 @@ public Builder<V> setTimeBetweenEvictionRuns(long timeBetweenEvictionRuns) {
     public ClientPoolProperty<V> build() {
       GenericKeyedObjectPoolConfig<V> poolConfig = new GenericKeyedObjectPoolConfig<>();
       poolConfig.setMaxTotalPerKey(maxClientNumForEachNode);
```
CRZbulabula, line 100: Previously `maxIdlePerKey` was set equal to `maxTotalPerKey` (= `maxClientNumForEachNode`). Now they can diverge, which is the intended behavior.

Worth noting: if `maxIdleClientNumForEachNode > maxClientNumForEachNode`, the pool silently clamps idle to total — `GenericKeyedObjectPool` never keeps more idle objects than the total per key. Consider adding a validation or comment to prevent users from configuring a misleading idle > max combination.

PR #17567, opened 2026-04-28
[ISSUE-xxxx] Fix IPv6 address parsing across server and client components

xiaoyekanren:
## Problem

Multiple components in IoTDB cannot correctly handle IPv6 addresses when configured with IPv6 endpoints like `[::1]:10710`. Using `split(":")` to parse IPv6 addresses causes:

- `NumberFormatException` when parsing empty string as port
- ConfigNode/DataNode startup failure
- JDBC URL parsing failure
- Python client `init_from_node_urls` failure

For IPv6 address `[::1]:10710`, `split(":")` produces:
- `items[0]` = `[`
- `items[1]` = `""` (empty string) ← **Integer.parseInt throws exception**
- `items[2]` = `1]`
- `items[3]` = `10710`

## Solution

This PR fixes IPv6 address parsing in 3 components:

### 1. Ratis Consensus Utils (Server-side)
- **File**: `iotdb-core/consensus/src/main/java/org/apache/iotdb/consensus/ratis/utils/Utils.java`
- **Methods**:
  - `hostAddress()` - Wrap IPv6 addresses in brackets `[ip]:port` format
  - `fromRaftPeerAddressToTEndPoint()` - Parse IPv6 format `[ipv6]:port` correctly
  - `fromRaftPeerProtoToTEndPoint()` - Parse IPv6 format correctly

### 2. JDBC Client
- **File**: `iotdb-client/jdbc/src/main/java/org/apache/iotdb/jdbc/Utils.java`
- **Method**: `parseUrl()` - Handle JDBC URLs like `jdbc:iotdb://[::1]:6667/` by detecting `[ipv6]` format

### 3. Python Client
- **File**: `iotdb-client/client-py/iotdb/Session.py`
- **Method**: `init_from_node_urls()` - Parse IPv6 node URLs like `[::1]:6667` correctly

## Affected Components

| Component | IPv6 Support Before | IPv6 Support After |
|-----------|--------------------|--------------------|
| ConfigNode/DataNode | ❌ Startup fails | ✅ Works |
| JDBC URL | ❌ Parse fails | ✅ `jdbc:iotdb://[::1]:6667/` |
| Python Session | ❌ `init_from_node_urls` fails | ✅ `[::1]:6667` |
| Java Session | ✅ Already works | ✅ No change needed |
| C++ Client | ✅ Already works | ✅ No change needed |

## Test Plan

- Manual testing with IPv6 configuration:
  - `cn_internal_address=::1`
  - `dn_rpc_address=::1`
  - JDBC URL: `jdbc:iotdb://[::1]:6667/`
  - Python: `Session.init_from_node_urls(["[::1]:6667"])`
- Verify ConfigNode and DataNode can start successfully with IPv6 addresses
- Verify JDBC and Python clients can connect to IPv6 endpoints

### DIFF
--- a/iotdb-client/client-py/iotdb/Session.py
+++ b/iotdb-client/client-py/iotdb/Session.py
@@ -150,9 +150,22 @@ class Session(object):
         session.__hosts = []
         session.__ports = []
         for node_url in node_urls:
-            split = node_url.split(":")
-            session.__hosts.append(split[0])
-            session.__ports.append(int(split[1]))
+            # Handle IPv6 address format [ipv6]:port
+            if node_url.startswith("["):
+                bracket_end = node_url.find("]")
+                if bracket_end > 0:
+                    host = node_url[1:bracket_end]
+                    # Port comes after "]:"
+                    port_str = node_url[bracket_end + 2:]
+                    session.__hosts.append(host)
+                    session.__ports.append(int(port_str))
+                else:
+                    raise RuntimeError(f"Invalid IPv6 address format: {node_url}")
+            else:
+                # IPv4 format: host:port
+                split = node_url.split(":")
+                session.__hosts.append(split[0])
+                session.__ports.append(int(split[1]))
         session.__host = session.__hosts[0]
         session.__port = session.__ports[0]
         session.__default_endpoint = TEndPoint(session.__host, session.__port)
--- a/iotdb-client/jdbc/src/main/java/org/apache/iotdb/jdbc/Utils.java
+++ b/iotdb-client/jdbc/src/main/java/org/apache/iotdb/jdbc/Utils.java
@@ -42,8 +42,8 @@ public class Utils {
   static final String RPC_COMPRESS = "rpc_compress";
 
   /**
-   * Parse JDBC connection URL The only supported format of the URL is:
-   * jdbc:iotdb://localhost:6667/.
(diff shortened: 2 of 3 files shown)


Caideyipi:
Thanks for working on the IPv6 parsing fix. I found two issues that look worth addressing before merge:

1. Ratis still cannot read legacy unbracketed IPv6 peer addresses. `hostAddress()` now writes IPv6 as `[ip]:port`, but `fromRaftPeerAddressToTEndPoint()` and `fromRaftPeerProtoToTEndPoint()` still parse non-bracket addresses with `split(:)`. If a node has existing/persisted peer metadata from the previous format, e.g. `::1:10710`, recovery/redirect parsing will still hit the old `NumberFormatException`. It would be safer to parse the non-bracket path with the last colon as the port separator, or explicitly normalize legacy addresses during recovery.

2. The bracketed IPv6 branches do not verify that `]` is followed by `:`. Inputs such as `[::1]6667` are currently accepted and parsed as port `667` because the code blindly starts at `bracketEnd + 2`. This affects the Ratis utils, JDBC URL parser, and Python `Session.init_from_node_urls()`. Please validate the `]:` separator and reject empty/non-numeric ports.

Could you also add regression tests for `[::1]:port` plus malformed bracket inputs like `[::1]6667`, `[::1]:`, and missing `]`? JDBC already has URL parser tests, and Ratis `UtilsTest` would be a good place to lock down the server-side behavior.

PR #17581, opened 2026-04-30
Fix allocate of MemoryBlock

Wei-hao-Li:
This PR fixes a memory concurrent allocate bug in AtomicLongMemoryBlock:
AtomicLongMemoryBlock.allocate() used AtomicLong.updateAndGet() with a lambda that had an external side effect (result.set(true)).

updateAndGet() is CAS-retry based: when CAS fails, the lambda can be executed multiple times. Under contention, this sequence could happen:

1. First lambda run sees enough memory, sets result=true, computes next.
2. CAS fails because another thread changed usedMemoryInBytes.
3. Lambda runs again with a new value and may decide allocation should fail (no increment).
4. Method still returns result=true from the earlier side effect.

So the caller may observe “allocation succeeded” while usedMemoryInBytes was not actually increased (phantom allocation success).
After enough such mismatches, release paths can try to release more than recorded memory, triggering warnings like “release is larger than memory cost.”

### DIFF
--- a/iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/memory/AtomicLongMemoryBlock.java
+++ b/iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/memory/AtomicLongMemoryBlock.java
@@ -23,7 +23,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 
 public class AtomicLongMemoryBlock extends IMemoryBlock {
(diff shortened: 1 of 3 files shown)


JackieTien97 (approved):

JackieTien97 (commented):
## Review Summary

The core fix is correct — replacing `updateAndGet` + `AtomicBoolean` side-effect with a hand-written CAS loop eliminates the phantom allocation bug. The `ramBytesUsed()` fix for the missing `queryId` field is also correct and relevant since `RamUsageEstimator.sizeOfObject()` delegates to `ramBytesUsed()` for `Accountable` objects.

Two minor suggestions as inline comments.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/memory/AtomicLongMemoryBlock.java
```diff
@@ -58,30 +57,30 @@ public long forceAllocateWithoutLimitation(long sizeInByte) {
...
   @Override
   public boolean allocateIfSufficient(final long sizeInByte, final double maxRatio) {
-    AtomicBoolean result = new AtomicBoolean(false);
-    usedMemoryInBytes.updateAndGet(
-        memCost -> {
-          if (memCost + sizeInByte > totalMemorySizeInBytes * maxRatio) {
-            return memCost;
-          }
-          result.set(true);
-          return memCost + sizeInByte;
```
JackieTien97, line 82: `release()` 方法中的 `LOGGER.warn()` 同样是 `updateAndGet` lambda 内的副作用，CAS 重试时可能导致：

1. **虚假告警**：第一次 lambda 执行时 `sizeInByte > memCost` 触发 WARN，CAS 失败后重试时 memCost 已变大（其他线程 allocate 了），release 实际成功 — 但 WARN 已经打出
2. **重复告警**：同一次 `release()` 调用打出多条 WARN

不影响正确性，但建议也改为手写 CAS 循环，与 `allocate()` 风格保持一致：
```java
public long release(long sizeInByte) {
    long prev, next;
    do {
        prev = usedMemoryInBytes.get();
        if (sizeInByte > prev) {
            LOGGER.warn("The memory cost to be released is larger than the memory cost of memory block {}", this);
            next = 0;
        } else {
            next = prev - sizeInByte;
        }
    } while (!usedMemoryInBytes.compareAndSet(prev, next));
    return next;
}
```

JackieTien97:
补充一个小建议：

`release()` 方法中的 `LOGGER.warn()` 同样是 `updateAndGet` lambda 内的副作用。虽然不影响正确性，但 CAS 重试时可能出现虚假告警（第一次 lambda 执行 `sizeInByte > memCost` 触发 WARN，CAS 失败重试时 memCost 已变大导致 release 实际成功，但 WARN 已打出）或同一次 release 调用打出多条 WARN。建议也改为手写 CAS 循环，与 `allocate()` 风格一致：

```java
public long release(long sizeInByte) {
    long prev, next;
    do {
        prev = usedMemoryInBytes.get();
        if (sizeInByte > prev) {
            LOGGER.warn("The memory cost to be released is larger than the memory cost of memory block {}", this);
            next = 0;
        } else {
            next = prev - sizeInByte;
        }
    } while (!usedMemoryInBytes.compareAndSet(prev, next));
    return next;
}
```

PR #17628, opened 2026-05-09
Fix that WALBuffer waits for flush instead of file-roll

jt2594838:
## Summary

- **Bug**: `WALNode.ReqIterator` used `waitForFlush` (signaled on buffer sync) to wait for new WAL data, but data only becomes readable after a WAL **file roll** — not a buffer flush. This caused the iterator to wake up prematurely (on flush) and then loop/timeout, or miss data entirely.
- **Fix**: Introduced a dedicated `rollLogWriterCondition` in `WALBuffer` that is signaled only when `rollLogWriter` creates a new WAL file. Renamed `waitForFlush` → `waitForRollFile` across `IWALBuffer`/`WALBuffer`/`WALNode` to reflect the corrected semantics.
- **Additional fix**: In `waitForNextReady(time, unit)`, changed `timeout || !hasNext()` to `timeout && !hasNext()` — previously a successful (non-timeout) wake-up could still throw `TimeoutException` if `hasNext()` happened to return false during a race.
- Added `tryToCollectInsertNodeAndBumpIndex.run()` before updating file index so the iterator processes entries from the current file before advancing.

## Changed files

- `IWALBuffer.java` — renamed interface methods `waitForFlush` → `waitForRollFile`
- `WALBuffer.java` — added `rollLogWriterCondition`, signal it in `rollLogWriter()`, use it in the renamed `waitForRollFile` methods
- `WALNode.java` — updated all call sites, fixed `timeout || !hasNext()` → `timeout && !hasNext()`, reordered collect call

## Test plan

- [x] Added `WALNodeWaitForRollFileTest` with 5 tests covering:
  - Timeout when no data is available
  - Flush without roll does **not** wake the iterator
  - Explicit roll wakes the iterator
  - Concurrent roll wakes a waiting iterator
  - Auto-roll on WAL file size threshold
  - Auto-triggered roll after timeout expiry


.idea/icon.png
JackieTien97, line 1: don't remove

JackieTien97 (changes requested):

JackieTien97 (approved):

PR #17631, opened 2026-05-10
Fix transport leak in ConfigNodeClient during leader redirect

CRZbulabula:
## Summary

- Fix a TCP connection leak in `ConfigNodeClient.connect()` on the DataNode side: when a DataNode is redirected from a Follower ConfigNode to the Leader, the old transport is not closed before being overwritten, causing server-side `TThreadPoolServer` worker threads to block indefinitely on abandoned sockets
- This was observed in a 5C20D cluster during ConfigNode failover testing — after two ConfigNodes were restarted, the former leader accumulated 55 leaked `ConfigNodeRPC-Processor` threads (vs. 1 for a normal follower), all stuck in `SocketDispatcher.read0` waiting on connections that would never receive data

## Root Cause

In `ConfigNodeClient.tryToConnect()`, when `configLeader` is set (redirect scenario), the code calls `connect(configLeader)` which directly overwrites `this.transport` with a new connection. The old transport to the previous ConfigNode is never closed. The server-side `TThreadPoolServer` keeps the corresponding worker thread alive, blocking on the now-orphaned socket.

The leak path:
1. DataNode borrows `ConfigNodeClient` from pool, already connected to ConfigNode A (Follower)
2. RPC returns `REDIRECTION_RECOMMEND` → `configLeader` is set to ConfigNode B (Leader)
3. Retry loop calls `connectAndSync()` → `tryToConnect()` → `connect(B)`
4. `connect()` overwrites `transport` with a new connection to B; old transport to A is leaked
5. ConfigNode A's `TThreadPoolServer` worker thread blocks forever on the abandoned socket

## Fix

Close the existing transport at the beginning of `connect()` before creating a new one. This covers all call paths (redirect branch and round-robin fallback loop in `tryToConnect()`).

## Test Plan

- [x] `mvn compile -pl iotdb-core/datanode` passes
- [ ] Verify in a multi-ConfigNode cluster: after leader failover and redirect, the old Follower's `ConfigNodeRPC-Processor` thread count stays at normal levels (1 for idle followers) instead of accumulating leaked threads

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/client/ConfigNodeClient.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/client/ConfigNodeClient.java
@@ -265,10 +265,7 @@ public class ConfigNodeClient implements IConfigNodeRPCService.Iface, ThriftClie
   }
 
   public void connect(TEndPoint endpoint, int timeoutMs) throws TException {
-    // Close old transport to avoid leaking TCP connections on the server side.
-    // Without this, redirect scenarios (Follower -> Leader) overwrite the transport
-    // field and leave the old connection open, causing server-side RPC threads to
-    // block indefinitely on the abandoned socket.
+    // Close existing transport before reassigning to prevent connection leaks.
     if (transport != null) {
       transport.close();
     }
@@ -332,10 +329,6 @@ public class ConfigNodeClient implements IConfigNodeRPCService.Iface, ThriftClie
         Thread.currentThread().interrupt();
         logger.warn("Unexpected interruption when waiting to try to connect to ConfigNode");
       }
-    }
-
-    if (transport != null) {
-      transport.close();
     }
 
     for (int tryHostNum = 0; tryHostNum < configNodes.size(); tryHostNum++) {
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/client/ConfigNodeClient.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/client/ConfigNodeClient.java
@@ -265,6 +265,13 @@ public class ConfigNodeClient implements IConfigNodeRPCService.Iface, ThriftClie
   }
 
   public void connect(TEndPoint endpoint, int timeoutMs) throws TException {
+    // Close old transport to avoid leaking TCP connections on the server side.
+    // Without this, redirect scenarios (Follower -> Leader) overwrite the transport
+    // field and leave the old connection open, causing server-side RPC threads to
+    // block indefinitely on the abandoned socket.
+    if (transport != null) {
+      transport.close();
+    }
     transport =
         commonConfig.isEnableInternalSSL()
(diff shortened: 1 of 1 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/client/ConfigNodeClient.java
```diff
@@ -265,6 +265,13 @@ public ConfigNodeClient(
   }
 
   public void connect(TEndPoint endpoint, int timeoutMs) throws TException {
+    // Close old transport to avoid leaking TCP connections on the server side.
+    // Without this, redirect scenarios (Follower -> Leader) overwrite the transport
+    // field and leave the old connection open, causing server-side RPC threads to
+    // block indefinitely on the abandoned socket.
```
JackieTien97: Nit: This 4-line comment is somewhat verbose for what is essentially a defensive `close()` before reassignment. The full context (redirect scenario, server-side thread leak) is already well-documented in the commit message and PR description. Consider trimming to a single line, e.g.:

```java
// Close existing transport before reassigning to prevent connection leaks.
```

The code itself (`if (transport != null) { transport.close(); }` before a field reassignment) is a well-known resource-management pattern that most Java developers will recognize on sight.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/client/ConfigNodeClient.java
```diff
@@ -265,6 +265,13 @@ public ConfigNodeClient(
...
+    if (transport != null) {
```
JackieTien97, line 269: Now that `connect()` always closes the old transport before creating a new one, the existing `transport.close()` in `tryToConnect()` (lines 337-339 on master, lines 344-346 in this PR) becomes redundant — when the round-robin loop calls `connect(tryEndpoint, timeoutMs)`, the old transport will be closed at entry.

The duplicate close is *harmless* (`TTransport.close()` is idempotent), but having two call sites that both claim responsibility for transport cleanup is a maintenance hazard. It's ambiguous whether `tryToConnect()` or `connect()` owns the lifecycle.

Suggestion: remove the `if (transport != null) { transport.close(); }` block in `tryToConnect()` so that `connect()` is the single owner of transport teardown. This makes the contract clearer: **every** `connect()` call manages its own cleanup.

JackieTien97 (commented):
## Review Summary

**Overall**: The fix is correct and well-motivated. The root cause analysis in the PR description is thorough and the change is minimal — exactly the right scope for a resource-leak fix.

### What's Good
- Correctly identifies the single point of failure: `connect()` overwrites `transport` without closing the old one
- Placing the close inside `connect()` covers all call paths (redirect branch and round-robin fallback), which is better than patching individual callers
- The null guard handles the first-connection case cleanly

### Suggestions
See inline comments for two minor items:
1. The 4-line comment could be trimmed — the commit message already documents the full context
2. The existing `transport.close()` in `tryToConnect()` (line 337-339 on master) is now redundant and could be removed to make ownership clearer

Neither blocks merge.

JackieTien97 (approved):

PR #17652, opened 2026-05-13
Fix multi InnerTimeJoinNode with different TimePartitions in one FI

Wei-hao-Li:


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/distribution/ExchangeNodeAdder.java
```diff
@@ -237,6 +241,10 @@ public PlanNode visitSingleDeviceView(SingleDeviceViewNode node, NodeGroupContex
 
   @Override
   public PlanNode visitMergeSort(MergeSortNode node, NodeGroupContext context) {
+    // Force Exchange if any child subtree contains InnerTimeJoin.
+    if (hasInnerTimeJoinInSubtree(node.getChildren())) {
```
JackieTien97: melt this judgement into previous DFS visit process

JackieTien97 (changes requested):

JackieTien97 (approved):

PR #17664, opened 2026-05-14
Optimized the overall performance of IoTDB & Fixed the NPE in LimitOperatorTest

Caideyipi:
## Description
1. Removed the unnecessary copies in QueryDataSetUtils & PartialPath.
2. Cached the deletion ranges to speedup the mod checks
3. Removed some unnecessary copies for partial paths
4. Removed some boxings on hot paths
5. Optimized some data structures (List::contains -> Set::contains, LinkedList -> ArrayList)
6. Fixed the write-lock leak in TsFileManager.removeAll.
7. Fixed wildcard tree last-cache invalidation for device paths.

<hr>

This PR has:
- [x] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
- [ ] added comments explaining the why and the intent of the code wherever would not be obvious 
  for an unfamiliar reader.
  for code coverage.

<hr>


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/read/filescan/impl/ClosedFileScanHandleImpl.java
```diff
@@ -87,10 +89,22 @@ public boolean isDeviceTimeDeleted(IDeviceID deviceID, long timestamp)
...
-        modifications.stream().map(ModEntry::getTimeRange).collect(Collectors.toList());
-    return ModificationUtils.isPointDeletedWithoutOrderedRange(timestamp, timeRangeList);
+    List<TimeRange> timeRangeList = deviceToDeletionRanges.get(deviceID);
+    if (timeRangeList == null) {
+      timeRangeList =
+          getMergedTimeRanges(queryContext.getPathModifications(curFileModEntries, deviceID));
+      deviceToDeletionRanges.put(deviceID, timeRangeList);
+    }
+    return ModificationUtils.isPointDeleted(timestamp, timeRangeList);
+  }
```
jt2594838, line 102: ComputeIfAbsent?
Is the additional memory footprint risky?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/StorageEngine.java
```diff
@@ -1086,9 +1086,11 @@ private void stopTimedServiceAndThrow(ScheduledExecutorService pool, String pool
...
   public void getDiskSizeByDataRegion(
       Map<Integer, Long> dataRegionDisk, List<Integer> dataRegionIds) {
+    final java.util.Collection<Integer> targetDataRegionIds =
+        dataRegionIds.size() > 1 ? new java.util.HashSet<>(dataRegionIds) : dataRegionIds;
     dataRegionMap.forEach(
         (dataRegionId, dataRegion) -> {
-          if (dataRegionIds.contains(dataRegionId.getId())) {
+          if (targetDataRegionIds.contains(dataRegionId.getId())) {
             dataRegionDisk.put(dataRegionId.getId(), dataRegion.countRegionDiskSize());
           }
```
jt2594838: for id : dataRegionIds
   dataRegionMap.computeIfPresent?

iotdb-client/client-cpp/src/main/SessionDataSet.cpp
```diff
@@ -246,38 +246,39 @@ const std::vector<std::string>& SessionDataSet::DataIterator::getColumnTypeList(
 
 shared_ptr<RowRecord> SessionDataSet::constructRowRecordFromValueArray() {
   std::vector<Field> outFields;
-  for (int i = iotdbRpcDataSet_->getValueColumnStartIndex(); i < iotdbRpcDataSet_->getColumnSize();
-       i++) {
+  const int32_t valueColumnStartIndex = iotdbRpcDataSet_->getValueColumnStartIndex();
+  const int32_t columnSize = iotdbRpcDataSet_->getColumnSize();
+  outFields.reserve(columnSize - valueColumnStartIndex);
+  for (int32_t columnIndex = valueColumnStartIndex + 1; columnIndex <= columnSize; ++columnIndex) {
```
JackieTien97, line 252: The loop changed from 0-based name-list indexing (`[valueColumnStartIndex, columnSize)`) to 1-based index access (`[valueColumnStartIndex + 1, columnSize]`). This is only correct if `isNullByIndex` / `getXxxByIndex` are 1-based such that `columnIndex == oldNameListIndex + 1`. Please double-check this against `IoTDBRpcDataSet` — an off-by-one here silently returns the wrong column's value with no error.

Also note `getBooleanByIndex(...).value()` will throw `bad_optional_access` if the optional is empty. It is guarded by `isNullByIndex` here, so it should be fine, but the `isNull` guard and the optional must agree on null semantics.

iotdb-client/isession/src/main/java/org/apache/iotdb/isession/SessionDataSet.java
```diff
@@ -153,45 +153,42 @@ public boolean hasNext() throws StatementExecutionException, IoTDBConnectionExce
...
 
   private RowRecord constructRowRecordFromValueArray() throws StatementExecutionException {
-    List<Field> outFields = new ArrayList<>();
-    for (int i = ioTDBRpcDataSet.getValueColumnStartIndex();
-        i < ioTDBRpcDataSet.getColumnSize();
-        i++) {
+    int valueColumnStartIndex = ioTDBRpcDataSet.getValueColumnStartIndex();
+    int columnSize = ioTDBRpcDataSet.getColumnSize();
+    List<Field> outFields = new ArrayList<>(columnSize - valueColumnStartIndex);
+    for (int columnIndex = valueColumnStartIndex + 1; columnIndex <= columnSize; columnIndex++) {
```
JackieTien97, line 159: Same index-base change as the C++ client: the loop is now `valueColumnStartIndex + 1 .. columnSize` (inclusive) feeding 1-based `isNull(columnIndex)` / `getDataType(columnIndex)` / `getXxx(columnIndex)`. Please confirm these index-based getters are 1-based relative to the column-name list (i.e. `columnIndex == oldNameListIndex + 1`). An off-by-one is silent here — it reads an adjacent column rather than failing. A unit test exercising a multi-column row would lock this down.

iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/IoTDBJDBCDataSet.java
```diff
@@ -321,6 +311,31 @@ public IoTDBJDBCDataSet(
     this.emptyResultSet = (queryDataSet == null || !queryDataSet.time.hasRemaining());
   }
 
+  private static int getDeduplicatedColumnSize(Map<String, Integer> columnNameIndex) {
```
JackieTien97, line 314: `getDeduplicatedColumnSize` now returns `max(index) + 1` instead of the previous `distinct().count()`. These are equal only when the dedup indices are contiguous from 0 — which the old `columnTypeDeduplicatedList.set(index, ...)` already required (a gap would have thrown `IndexOutOfBoundsException`), so in practice this is equivalent and more robust. Just confirm no downstream consumer of `columnTypeDeduplicatedList` relies on `size()` being exactly the distinct count, or assumes no `null` entries.

iotdb-client/session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -1817,10 +1765,12 @@ private boolean filterNullValueAndMeasurement(
       List<String> measurementsList,
       List<TSDataType> types,
       List<Object> valuesList) {
-    Map<String, Object> nullMap = new HashMap<>();
+    Map<String, Object> nullMap = logger.isInfoEnabled() ? new HashMap<>() : null;
```
JackieTien97, line 1768: `nullMap` is now `null` whenever info logging is disabled. The loop body is correctly guarded, but the code after this loop (not in the diff) presumably consumes `nullMap` for logging — e.g. `if (!nullMap.isEmpty()) logger.info(...)`. If that consumer dereferences `nullMap` without a null check (or without short-circuiting on `logger.isInfoEnabled()` first), this throws an NPE when info logging is off. Please verify every subsequent use of `nullMap` in this method is null-safe.

iotdb-client/session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -1865,10 +1815,12 @@ private void filterNullValueAndMeasurementWithStringType(
    */
   private boolean filterNullValueAndMeasurementWithStringType(
       List<String> valuesList, String deviceId, List<String> measurementsList) {
-    Map<String, Object> nullMap = new HashMap<>();
+    Map<String, Object> nullMap = logger.isInfoEnabled() ? new HashMap<>() : null;
```
JackieTien97, line 1820: Same as the other `nullMap` change above — `nullMap` is `null` when info logging is disabled. Confirm the code after this loop guards every `nullMap` access (null check, or `logger.isInfoEnabled()` short-circuit before `nullMap.isEmpty()`).

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/metadata/fetcher/cache/TableDeviceSchemaCache.java
```diff
@@ -482,7 +482,7 @@ void invalidateLastCache(final PartialPath devicePath, final String measurement)
           },
           cachedDeviceID -> {
             try {
-              return new PartialPath(cachedDeviceID).matchFullPath(devicePath);
+              return devicePath.matchFullPath(cachedDeviceID);
```
JackieTien97, line 485: This swaps the `matchFullPath` direction: previously `new PartialPath(cachedDeviceID).matchFullPath(devicePath)` (cached entry as the pattern), now `devicePath.matchFullPath(cachedDeviceID)` (devicePath as the pattern). The new direction is correct and consistent with `invalidateCache` below.

Note this also changes behavior when `devicePath` contains wildcards: the old form would have failed to match (a concrete path cannot match a wildcard path when used as the pattern), so entries that should have been invalidated were left stale. If that is an intended bug fix, great — worth calling out in the PR description and covering with a test.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/SchemaEngine.java
```diff
@@ -371,11 +371,13 @@ public int getSchemaRegionNumber() {
 
   public Map<Integer, Long> countDeviceNumBySchemaRegion(final List<Integer> schemaIds) {
     final Map<Integer, Long> deviceNum = new HashMap<>();
+    final java.util.Collection<Integer> targetSchemaIds =
```
JackieTien97: Use regular imports for `java.util.Collection` / `java.util.HashSet` rather than inline fully-qualified names — the rest of the file uses imports and this is inconsistent (and may trip checkstyle). The same pattern appears in `StorageEngine.getDiskSizeByDataRegion`. The `size() > 1` guard to avoid allocating a `Set` for the trivial case is a reasonable micro-optimization.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/StorageEngine.java
```diff
@@ -1116,12 +1116,14 @@ private void stopTimedServiceAndThrow(ScheduledExecutorService pool, String pool
 
   public void getDiskSizeByDataRegion(
       Map<Integer, Long> dataRegionDisk, List<Integer> dataRegionIds) {
-    dataRegionMap.forEach(
-        (dataRegionId, dataRegion) -> {
-          if (dataRegionIds.contains(dataRegionId.getId())) {
-            dataRegionDisk.put(dataRegionId.getId(), dataRegion.countRegionDiskSize());
-          }
-        });
+    final java.util.Collection<Integer> targetDataRegionIds =
```
JackieTien97: Same fully-qualified-name nit as `SchemaEngine` — please import `Collection` / `HashSet`. Logic-wise the rewrite (iterate the requested ids and look each up) is fine and equivalent: duplicates in `dataRegionIds` are handled by the `Set` when `size > 1`, and `dataRegionDisk.put` is idempotent regardless. Switching from a full-map scan to direct `get` is a good improvement.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/read/filescan/impl/ClosedFileScanHandleImpl.java
```diff
@@ -87,10 +87,18 @@ public boolean isDeviceTimeDeleted(IDeviceID deviceID, long timestamp)
         curFileModEntries != null
             ? curFileModEntries
             : queryContext.loadAllModificationsFromDisk(tsFileResource);
-    List<ModEntry> modifications = queryContext.getPathModifications(curFileModEntries, deviceID);
     List<TimeRange> timeRangeList =
-        modifications.stream().map(ModEntry::getTimeRange).collect(Collectors.toList());
-    return ModificationUtils.isPointDeletedWithoutOrderedRange(timestamp, timeRangeList);
+        getMergedTimeRanges(queryContext.getPathModifications(curFileModEntries, deviceID));
```
JackieTien97: `isDeviceTimeDeleted` recomputes `getPathModifications` and now also runs `sortAndMerge` on every call, with no caching — whereas `isTimeSeriesTimeDeleted` just below caches via `deviceToModifications`, and `UnclosedFileScanHandleImpl` caches both device- and series-level ranges. If `isDeviceTimeDeleted` is called per-timestamp, this is a small regression versus the old `isPointDeletedWithoutOrderedRange` (an extra sort on each call). Consider caching the device-level merged ranges here too, for consistency.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/read/filescan/impl/UnclosedFileScanHandleImpl.java
```diff
@@ -48,6 +51,8 @@ public class UnclosedFileScanHandleImpl implements IFileScanHandle {
   private final TsFileResource tsFileResource;
   private final Map<IDeviceID, Map<String, List<IChunkMetadata>>> deviceToChunkMetadataMap;
   private final Map<IDeviceID, Map<String, List<IChunkHandle>>> deviceToMemChunkHandleMap;
+  private final Map<IDeviceID, List<TimeRange>> deviceToDeletionRanges;
+  private final Map<IDeviceID, Map<String, List<TimeRange>>> deviceToTimeSeriesDeletionRanges;
```
JackieTien97, line 55: These two new caches are plain `HashMap`s, mutated lazily by `isDeviceTimeDeleted` / `isTimeSeriesTimeDeleted` (`put` and `computeIfAbsent`). If a single `UnclosedFileScanHandleImpl` can be queried from multiple threads concurrently — common in the scan/query path — these unsynchronized mutations can corrupt the map or lose updates. The PR's "concurrent read" checkbox is unchecked. Please confirm single-threaded access, or switch to `ConcurrentHashMap`.

Separately: `appendDeletionRanges` adds `TimeRange` references straight from `chunkMetadata.getDeleteIntervalList()` and then `sortAndMerge` runs over them. Please confirm `sortAndMerge` does not mutate the shared `TimeRange` instances in place, otherwise it could corrupt the chunk metadata's delete intervals.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/tsfile/TsFileManager.java
```diff
@@ -223,10 +216,18 @@ public void removeAll(List<TsFileResource> tsFileResourceList, boolean sequence)
     writeLock("removeAll");
     try {
       for (TsFileResource resource : tsFileResourceList) {
-        remove(resource, sequence);
+        removeFromPartitionFileList(resource, sequence);
       }
     } finally {
-      writeLock("removeAll");
+      writeUnlock();
+    }
```
JackieTien97, line 223: Good catch — the original `removeAll` `finally` block called `writeLock("removeAll")` instead of `writeUnlock()`, which acquired the write lock a second time and never released it (a write-lock count leak on every `removeAll`). This fix is correct and important; worth highlighting explicitly in the PR description since it is a real bug fix, not just an optimization.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/tsfile/TsFileManager.java
```diff
@@ -223,10 +216,18 @@ public void removeAll(List<TsFileResource> tsFileResourceList, boolean sequence)
...
+  }
+
+  private void removeFromPartitionFileList(TsFileResource tsFileResource, boolean sequence) {
+    Map<Long, TsFileResourceList> selectedMap = sequence ? sequenceFiles : unsequenceFiles;
+    TsFileResourceList tsFileResources = selectedMap.get(tsFileResource.getTimePartition());
+    if (tsFileResources != null && tsFileResources.remove(tsFileResource)) {
```
JackieTien97, line 229: `removeFromPartitionFileList` now looks up only the partition `tsFileResource.getTimePartition()` instead of scanning all partitions — a good O(1) improvement. It assumes a resource is always stored in the list keyed by its own `getTimePartition()`. That should hold as an invariant; just confirm there is no path where a resource's time partition can change after insertion, otherwise `remove` would silently fail to remove it (the old full scan would still have found it).

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/ModificationUtils.java
```diff
@@ -82,7 +82,11 @@ public static void modifyChunkMetaData(
               if (range.contains(metaData.getStartTime(), metaData.getEndTime())) {
                 return true;
               } else {
-                if (range.overlaps(new TimeRange(metaData.getStartTime(), metaData.getEndTime()))) {
+                if (overlap(
```
JackieTien97, line 85: Replacing `range.overlaps(new TimeRange(start, end))` with `overlap(start, end, range.getMin(), range.getMax())` avoids the allocation — good. Please make sure `overlap(...)` uses exactly the same boundary semantics as `TimeRange.overlaps` (inclusive endpoints). A subtle inclusive/exclusive mismatch here changes whether a deletion is considered to touch a chunk, which directly affects query correctness.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/QueryDataSetUtils.java
```diff
@@ -572,40 +558,30 @@ private static void fillRemainingBitMap(
...
 
   private static void fillTimeColumn(
-      int rowCount, ByteArrayOutputStream[] byteArrayOutputStreams, TSQueryDataSet tsQueryDataSet) {
-    // calculate the time buffer size
-    int timeOccupation = rowCount * 8;
-    ByteBuffer timeBuffer = ByteBuffer.allocate(timeOccupation);
-    timeBuffer.put(byteArrayOutputStreams[0].toByteArray());
-    timeBuffer.flip();
-    tsQueryDataSet.setTime(timeBuffer);
+      int rowCount, PublicBAOS[] byteArrayOutputStreams, TSQueryDataSet tsQueryDataSet) {
```
JackieTien97: After the refactor `rowCount` is no longer used in `fillTimeColumn` (the old `rowCount * 8` allocation is gone). The same applies to `rowCount` in `fillValueColumnsAndBitMaps` (line 566), where `valueOccupation` is now only read for `.length`. Please drop the now-unused parameters — they will otherwise trip checkstyle/IDE warnings and are misleading to readers. `columnNum` can be derived as `byteArrayOutputStreams.length / 2` if you want to remove `valueOccupation` entirely.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/QueryDataSetUtils.java
```diff
@@ -572,40 +558,30 @@ private static void fillRemainingBitMap(
...
-      bitmapBuffer.flip();
-      bitmapList.add(bitmapBuffer);
+      valueList.add(wrapBuffer(byteArrayOutputStreams[i]));
+      bitmapList.add(wrapBuffer(byteArrayOutputStreams[i + 1]));
     }
     tsQueryDataSet.setBitmapList(bitmapList);
     tsQueryDataSet.setValueList(valueList);
   }
 
+  private static ByteBuffer wrapBuffer(PublicBAOS outputStream) {
```
JackieTien97, line 510: Nice zero-copy win — wrapping the `PublicBAOS` internal buffer avoids the `toByteArray()` + `allocate` + `put` double copy. One invariant to keep: the returned `ByteBuffer` aliases the stream's internal array, so nothing must write to these `PublicBAOS` / `DataOutputStream` instances after `wrapBuffer` is called. That holds in the current flow (wrapping happens last, right before return), just keep it in mind if this code is extended later.

JackieTien97 (commented):
Thanks for the performance work — the changes are mostly solid and several are clear wins: the zero-copy `wrapBuffer` rewrite in `QueryDataSetUtils`, the `bytesToHex` lookup-table rewrite, factoring the device/time filter out of the per-measurement loop in `ModificationUtils`, the `removeAll` write-lock leak fix in `TsFileManager`, and the new `matchFullPath(IDeviceID)` / `matchPrefixPath(IDeviceID)` overloads in `PartialPath` (no more throwaway `PartialPath` allocations on every cache check).

A few items below are worth confirming before merge, grouped by priority:

**Should confirm (potential correctness)**
- `SessionDataSet` (Java + C++): the loop switched to 1-based index access (`valueColumnStartIndex + 1 .. columnSize`). Please confirm the index-based getters are 1-based relative to the column-name list — an off-by-one here silently reads adjacent columns with no error.
- `Session.java`: `nullMap` is now `null` when info logging is disabled. Every subsequent dereference in those methods must be guarded.
- `UnclosedFileScanHandleImpl`: the new lazy `HashMap` caches are not thread-safe; please confirm single-threaded access or switch to `ConcurrentHashMap`.
- `TableDeviceSchemaCache.invalidateLastCache`: the `matchFullPath` direction was swapped — the new direction looks correct (and likely fixes a latent wildcard bug), worth calling out and covering with a test.
- `ModificationUtils`: confirm the new `overlap(...)` helper has the same inclusive/exclusive boundary semantics as `TimeRange.overlaps`.

**Minor / style / cleanup**
- Inline fully-qualified `java.util.Collection` / `java.util.HashSet` in `SchemaEngine` and `StorageEngine` — please use imports.
- `QueryDataSetUtils`: `rowCount` and `valueOccupation` parameters are unused after the refactor — please drop them.
- `IoTDBJDBCDataSet`: `getDeduplicatedColumnSize` now returns `max(index)+1` instead of distinct-count — equivalent for contiguous indices, just confirm no downstream code depends on the old size.

Details inline.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/read/filescan/impl/ClosedFileScanHandleImpl.java
Caideyipi: Updated. I switched the caches to ConcurrentHashMap and use putIfAbsent here. I kept the merge outside computeIfAbsent so the relatively expensive mod collection and merge is not executed while holding a CHM bin lock; duplicate concurrent computation is harmless. The added memory is per scan handle and only for queried devices, storing merged ranges instead of raw mods, and is released with the handle.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/StorageEngine.java
Caideyipi: Done. getDiskSizeByDataRegion now iterates the requested ids and does a direct dataRegionMap.get(new DataRegionId(id)). I avoided computeIfPresent since this is read-only and should not mutate the map.

iotdb-client/client-cpp/src/main/SessionDataSet.cpp
Caideyipi: Confirmed against C++ IoTDBRpcDataSet: getTsBlockColumnIndexForColumnIndex subtracts 1, and findColumnNameByIndex rejects index <= 0, so these getters are 1-based. The null guard uses the same column-index path as the typed getters.

iotdb-client/isession/src/main/java/org/apache/iotdb/isession/SessionDataSet.java
Caideyipi: Confirmed against Java IoTDBRpcDataSet: findColumnNameByIndex and getTsBlockColumnIndexForColumnIndex are 1-based and access columnNameList.get(columnIndex - 1). Existing multi-column client coverage also uses 1-based indexes.

iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/IoTDBJDBCDataSet.java
Caideyipi: Confirmed. Downstream uses columnTypeDeduplicatedList.size() to allocate arrays indexed by the deduplicated ordinal, so max(index) + 1 is the right size. Valid input has contiguous ordinals and no null entries; with a gap, the previous set(index, ...) path would already have failed.

iotdb-client/session/src/main/java/org/apache/iotdb/session/Session.java
Caideyipi: Fixed in latest push. The post-loop logging in filterNullValueAndMeasurement now checks nullMap != null before logging, so the info-disabled path cannot dereference it.

iotdb-client/session/src/main/java/org/apache/iotdb/session/Session.java
Caideyipi: Fixed in latest push. The string-type variant now has the same null-safe post-loop logging guard: nullMap != null.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/read/filescan/impl/UnclosedFileScanHandleImpl.java
Caideyipi: Updated. Both lazy caches are now ConcurrentHashMaps and use computeIfAbsent. appendDeletionRanges now copies each TimeRange before merge, so sortAndMerge cannot mutate intervals owned by chunk metadata.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/tsfile/TsFileManager.java
Caideyipi: Updated the PR description to call out the removeAll write-lock leak fix explicitly.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/tsfile/TsFileManager.java
Caideyipi: Confirmed. TsFileManager inserts resources into partition maps using the same resource.getTimePartition() key; managed resources are not moved to another partition in place. Replacement and rename flows build or update the resource before adding or replacing it in the manager.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/ModificationUtils.java
Caideyipi: Confirmed. ModificationUtils.overlap uses closed-interval semantics: endB >= startA && startB <= endA, matching TimeRange.overlaps inclusive endpoints.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/QueryDataSetUtils.java
Caideyipi: Done. Dropped the unused rowCount parameters and removed valueOccupation entirely; columnNum now derives from byteArrayOutputStreams.length / 2.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/QueryDataSetUtils.java
Caideyipi: Confirmed. wrapBuffer is only called after serialization is complete and immediately before populating TSQueryDataSet; no writes happen to those PublicBAOS/DataOutputStream instances after wrapping.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/metadata/fetcher/cache/TableDeviceSchemaCache.java
Caideyipi: Done. Added a PR description bullet for the wildcard tree last-cache invalidation fix and added TreeDeviceSchemaCacheManagerTest#testInvalidateLastCacheByWildcardDevicePath to cover the wildcard device-path case.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/SchemaEngine.java
Caideyipi: Done in latest push. SchemaEngine now imports Collection and HashSet instead of using inline fully-qualified names.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/StorageEngine.java
Caideyipi: Done in latest push. StorageEngine now imports Collection and HashSet instead of using inline fully-qualified names.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/read/filescan/impl/ClosedFileScanHandleImpl.java
Caideyipi: Done. ClosedFileScanHandleImpl now has a deviceToDeletionRanges cache for merged device-level ranges, using putIfAbsent for concurrent-safe lazy publication.

jt2594838 (approved):

luoluoyuyu (commented):
## Review 总结

性能优化覆盖面广；TsFileManager.removeAll 写锁泄漏修复是关键正确性修复，建议优先 cherry-pick。

合入前：修 CI（cpp macOS、Windows、部分 IT）；客户端列索引变更需全语言一致性验证（见行内）。

建议考虑拆 PR 降低风险。

iotdb-client/session/src/main/java/org/apache/iotdb/session/Session.java
```diff
@@ -1865,18 +1817,22 @@ private void filterNullValueAndMeasurementWithStringType(
...
```
luoluoyuyu, line 1820: nullMap is null when info logging is disabled; all uses in this method are guarded. Please verify every overload of filterNullValueAndMeasurement changed in this PR uses the same pattern.

luoluoyuyu (commented):
## Review summary

Performance improvements across client, schema cache, storage scan, and TsFileManager. The removeAll finally block now calls writeUnlock instead of acquiring the write lock again, fixing a serious lock leak.

Please fix CI failures in this PR. See inline notes.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/tsfile/TsFileManager.java
```diff
@@ -223,10 +216,18 @@ public void removeAll(List<TsFileResource> tsFileResourceList, boolean sequence)
     writeLock("removeAll");
     try {
       for (TsFileResource resource : tsFileResourceList) {
-        remove(resource, sequence);
+        removeFromPartitionFileList(resource, sequence);
       }
     } finally {
-      writeLock("removeAll");
+      writeUnlock();
```
luoluoyuyu, line 222: Critical fix: finally now calls writeUnlock() instead of writeLock again in removeAll, which previously leaked the write lock on every call.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/read/filescan/impl/ClosedFileScanHandleImpl.java
```diff
@@ -87,10 +90,24 @@ public boolean isDeviceTimeDeleted(IDeviceID deviceID, long timestamp)
         curFileModEntries != null
             ? curFileModEntries
             : queryContext.loadAllModificationsFromDisk(tsFileResource);
-    List<ModEntry> modifications = queryContext.getPathModifications(curFileModEntries, deviceID);
-    List<TimeRange> timeRangeList =
-        modifications.stream().map(ModEntry::getTimeRange).collect(Collectors.toList());
-    return ModificationUtils.isPointDeletedWithoutOrderedRange(timestamp, timeRangeList);
+    List<TimeRange> timeRangeList = deviceToDeletionRanges.get(deviceID);
```
luoluoyuyu, line 93: deviceToDeletionRanges caches merged deletion ranges per device via putIfAbsent, avoiding repeated sortAndMerge in isDeviceTimeDeleted. If this handle can be used from multiple threads, document single-thread use or use a concurrent map.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/tsfile/TsFileManager.java
Caideyipi: Confirmed, this fix is included and has been called out in the PR description as a correctness fix, not just a performance optimization.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/read/filescan/impl/ClosedFileScanHandleImpl.java
Caideyipi: Verified. ClosedFileScanHandleImpl now uses ConcurrentHashMap for deviceToDeletionRanges and deviceToModifications, publishes device-level ranges with putIfAbsent, and creates per-device time-series maps as ConcurrentHashMap instances as well.

iotdb-client/session/src/main/java/org/apache/iotdb/session/Session.java
Caideyipi: Verified. The two filterNullValueAndMeasurement overloads changed in this PR guard every nullMap access with nullMap != null, and the wrapper overloads only delegate to them without dereferencing the map.

Caideyipi:
Thanks for the review. I rechecked the latest head and replied inline. The Windows and IT failures mentioned earlier are now green. The remaining C++ macOS failure was a Maven assembly OOM (Java heap space) while building iotdb-cli, so I reran the failed Multi-Language Client workflow; the macOS rerun is pending now.

PR #17672, opened 2026-05-14
Added flags to mark whether a device has device descendants to optimize query like select xx from xxx.**

Caideyipi:
## Description
As the title said.

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/impl/mem/MTreeBelowSGMemoryImpl.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/impl/mem/MTreeBelowSGMemoryImpl.java
@@ -225,8 +225,7 @@ public class MTreeBelowSGMemoryImpl {
   }
 
   private boolean hasDeviceDescendantInChildren(final IMemMNode node) {
-    final IMNodeIterator<IMemMNode> iterator = store.getChildrenIterator(node);
-    try {
+    try (final IMNodeIterator<IMemMNode> iterator = store.getChildrenIterator(node)) {
       while (iterator.hasNext()) {
         final IMemMNode child = iterator.next();
         if (child.isDevice() || child.hasDeviceDescendant()) {
@@ -234,8 +233,6 @@ public class MTreeBelowSGMemoryImpl {
         }
       }
       return false;
-    } finally {
-      iterator.close();
     }
   }
 
@@ -697,14 +694,15 @@ public class MTreeBelowSGMemoryImpl {
       boolean hasMeasurement = false;
       boolean hasNonViewMeasurement = false;
       IMemMNode child;
-      IMNodeIterator<IMemMNode> iterator = store.getChildrenIterator(curNode);
-      while (iterator.hasNext()) {
-        child = iterator.next();
-        if (child.isMeasurement()) {
-          hasMeasurement = true;
-          if (!child.getAsMeasurementMNode().isLogicalView()) {
-            hasNonViewMeasurement = true;
-            break;
+      try (final IMNodeIterator<IMemMNode> iterator = store.getChildrenIterator(curNode)) {
+        while (iterator.hasNext()) {
+          child = iterator.next();
+          if (child.isMeasurement()) {
+            hasMeasurement = true;
+            if (!child.getAsMeasurementMNode().isLogicalView()) {
+              hasNonViewMeasurement = true;
+              break;
+            }
           }
         }
       }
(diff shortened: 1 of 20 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/impl/mem/MTreeBelowSGMemoryImpl.java
```diff
@@ -208,6 +207,46 @@ private void applySubtreeMeasurementDelta(IMemMNode startNode, final long delta)
...
+    IMemMNode current = deviceNode.getParent();
+    while (current != null && !current.hasDeviceDescendant()) {
+      current.setHasDeviceDescendant(true);
+      current = current.getParent();
+    }
+  }
+
+  private boolean hasDeviceDescendantInChildren(final IMemMNode node) {
+    final IMNodeIterator<IMemMNode> iterator = store.getChildrenIterator(node);
+    try {
```
jt2594838: try-with-resource?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/impl/mem/MTreeBelowSGMemoryImpl.java
```diff
@@ -315,7 +355,7 @@ public IMeasurementMNode<IMemMNode> createTimeSeries(
       if (device.isDevice()) {
         entityMNode = device.getAsDeviceMNode();
       } else {
-        entityMNode = store.setToEntity(device);
+        entityMNode = setToEntityAndUpdateFlags(device);
```
luoluoyuyu, line 355: setToEntityAndUpdateFlags marks ancestors when a node becomes a device. Please confirm refreshAncestorsHavingDeviceDescendant on deleteEmptyInternalMNode keeps flags correct when the last device under a prefix is removed.

luoluoyuyu (commented):
## Review summary

Adds hasDeviceDescendant flags on schema nodes and maintains them when creating devices and when deleting empty internal nodes. This should speed up queries that need to know if a prefix has device descendants.

Please fix CI failures and add tests for flag updates on delete/move paths touched in this PR.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/impl/mem/MTreeBelowSGMemoryImpl.java
Caideyipi: Done. `IMNodeIterator` now extends `AutoCloseable`, and the newly added child-iterator scans use try-with-resources in memory, cached, and traverser paths.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/impl/mem/MTreeBelowSGMemoryImpl.java
Caideyipi: Confirmed and covered with tests. The delete refresh recomputes ancestors from children, so when the last device under one prefix is removed, higher ancestors stay true if another sibling device subtree still exists and become false only after the last remaining device subtree is removed. I extended both memory and cached MTree tests for that case, and added memory-mode table-device drop coverage as well.

Caideyipi:
Addressed the review comments in commit 2a68c3842fa:

- Made IMNodeIterator AutoCloseable and converted the new iterator scans to try-with-resources.
- Reduced the Sonar duplication from repeated flag delegate methods by moving the shared accessors to default methods on IMemMNode / ICachedMNode.
- Added coverage for delete refresh correctness when one device subtree is removed while another sibling device subtree still exists, plus table-device drop coverage in memory mode.

Verification:
- mvn -pl iotdb-core/datanode,iotdb-core/node-commons spotless:apply
- MAVEN_OPTS=-Xmx768m mvn -Ddevelocity.off=true -Dcheckstyle.skip=true -Dspotless.check.skip=true -pl iotdb-core/node-commons,iotdb-core/datanode -Dtest=MTreeBelowSGMemoryImplTest,MTreeBelowSGCachedImplTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false test

For current CI: the latest failing Simple (17) job failed during ctions/checkout with a GitHub TLS fetch error before build/test. The dual-tree-auto-basic failure is in IoTDBPipeLifeCycleIT.testLifeCycleLogMode, unrelated to this schema flag change.

PR #17674, opened 2026-05-14
Load: Optimized the downgraded logic for tsFile to insert more data when tsFile corrupted

Caideyipi:
## Description
As the title said.

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/load/converter/LoadTreeStatementDataTypeConvertExecutionVisitor.java
```diff
@@ -89,16 +87,9 @@ public Optional<TSStatus> visitLoadFile(
...
-            new TsFileInsertionEventScanParser(
-                file,
-                new IoTDBTreePattern(null),
-                Long.MIN_VALUE,
-                Long.MAX_VALUE,
-                null,
-                null,
-                true)) {
-          for (final Pair<Tablet, Boolean> tabletWithIsAligned : parser.toTabletWithIsAligneds()) {
+        try (final LoadTreeTsFileTabletIterator tabletIterator =
```
luoluoyuyu, line 91: Replacing TsFileInsertionEventScanParser with LoadTreeTsFileTabletIterator centralizes corrupted-file handling for load. Please document what data may still be skipped versus inserted when a file is partially corrupted.

luoluoyuyu (commented):
## Review summary

Introduces LoadTreeTsFileTabletIterator and uses it in LoadTreeStatementDataTypeConvertExecutionVisitor so corrupted TsFiles can yield more insertable tablets on downgrade. Adds unit tests for the converter path.

Please fix failing datanode unit tests before merge.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/load/converter/LoadTreeTsFileTabletIterator.java
```diff
@@ -0,0 +1,558 @@
...
+        final IDeviceID device = deviceIterator.next().getLeft();
+        deviceMeasurementsMap.put(
+            device,
+            reader.readDeviceMetadata(device).values().stream()
+                .map(TimeseriesMetadata::getMeasurementId)
+                .collect(Collectors.toList()));
+      }
+    }
+    return deviceMeasurementsMap;
+  }
```
jt2594838, line 279: May use org.apache.tsfile.read.TsFileSequenceReader#getAllTimeseriesMetadata

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/load/converter/LoadTreeTsFileTabletIterator.java
```diff
@@ -0,0 +1,558 @@
...
+
+    final IDeviceID device = IDeviceID.Factory.DEFAULT_FACTORY.create(tablet.getDeviceId());
+    final List<String> measurements = extractMeasurementNames(tablet);
+
+    if (Objects.isNull(activeQueryParser)) {
+      recordScanProgress(device, measurements);
+    }
+
+    lastEmittedDevice = device;
+    lastEmittedMeasurements = measurements;
```
jt2594838, line 380: Why not directly use the measurements array from the tablet?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/load/converter/LoadTreeTsFileTabletIterator.java
Caideyipi: Updated in a3e00d5. I switched this to the TsFileSequenceReader metadata API via iterAllTimeseriesMetadata(false, false). I used the iterator instead of getAllTimeseriesMetadata(false) because the latter builds a HashMap in the current TsFile version, while this fallback needs to preserve device traversal order when resuming after scan parser failure.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/load/converter/LoadTreeTsFileTabletIterator.java
Caideyipi: Updated in a3e00d5. The current Tablet API does not expose a public measurements array/getter; the public source of measurement names is the schema list. I replaced the manual schema loop with IMeasurementSchema.getMeasurementNameList(tablet.getSchemas()) so this now uses the shared helper while keeping the same schema-order semantics.

jt2594838 (approved):

PR #17680, opened 2026-05-15
Support container-aware Docker startup.

wenyanshi-123:
Support a container-aware Docker startup mechanism, enabling IoTDB to correctly detect container-allocated resources such as CPU cores, memory, and disks.

### DIFF
--- a/docker/src/main/DockerCompose/iotdb-docker-env.sh
+++ b/docker/src/main/DockerCompose/iotdb-docker-env.sh
@@ -1,0 +1,334 @@
+#!/bin/bash
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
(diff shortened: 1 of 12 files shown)


docker/src/main/DockerCompose/iotdb-docker-env.sh
```diff
@@ -0,0 +1,334 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+start_what="$1"
```
CRZbulabula, line 35: Append annotations to explain the logic of this file

CRZbulabula (changes requested):
PTAL.

PR #17687, opened 2026-05-16
Speed up AINode CI by consolidating tests and caching PyInstaller output

JackieTien97:
## Description

Speed up the **Cluster IT - 1C1D1A** CI pipeline (~52min → ~15-20min) via two optimizations:

### 1. Test consolidation (saves ~20min)

Merge 5 AINode IT test classes into a single `AINodeSharedClusterIT` that shares one 1C1D1A cluster:
- `AINodeDeviceManageIT`
- `AINodeModelManageIT`
- `AINodeCallInferenceIT`
- `AINodeForecastIT`
- `AINodeInstanceManagementIT`

This reduces cluster startups from **8 to 3** (SharedCluster + ClusterConfig + ConcurrentForecast).

Also converts `AINodeClusterConfigIT` from `@Before/@After` to `@BeforeClass/@AfterClass`, eliminating one redundant cluster restart by merging both dialect tests into a single method.

The 1C1D1A cluster startup takes ~3.5min each time (ConfigNode + DataNode + AINode), so eliminating 5 restarts saves ~17.5min. The actual test execution time is only ~9min out of the original 52min total.

### 2. PyInstaller dist caching (saves ~11min)

Added hash-based caching to `build_binary.py`:
- Computes SHA256 of all AINode source files, `pyproject.toml`, `poetry.lock`, and `ainode.spec`
- Caches the `dist/` output at `~/.cache/iotdb-ainode-build/dist-cache/` (outside project dir, survives `mvn clean`)
- On cache hit, restores `dist/` directly and skips the entire PyInstaller analysis + packaging phase

The PyInstaller phase scans thousands of hidden imports from torch/transformers/numpy and takes ~11min. When AINode source hasn't changed, this is entirely redundant.

### Testing

All existing test logic is preserved — the tests are reorganized, not changed. `AINodeConcurrentForecastIT` remains separate (different data setup and LOAD/UNLOAD side effects).


JackieTien97 (commented):
## Code Review

### Verdict: LGTM

The test logic is faithfully preserved — this is a reorganization, not a behavioral change. The caching implementation is sound.

---

### Observations

**Test Consolidation (`AINodeSharedClusterIT.java`)**

- All test methods from the original 5 classes are present and structurally identical.
- `@BeforeClass` correctly calls both `prepareDataInTree()` and `prepareDataInTable()` to satisfy all tests.
- `AINodeConcurrentForecastIT` correctly left separate (different data setup, LOAD/UNLOAD side effects).
- `AINodeClusterConfigIT` correctly left separate (it does `REMOVE AINODE` which would break the shared cluster).

**`AINodeClusterConfigIT` refactoring**

- Conversion from `@Before/@After` to `@BeforeClass/@AfterClass` eliminates one redundant cluster restart.
- The merged test now verifies both dialects in a single method with proper sequencing. Cleaner than the original.

**PyInstaller caching (`build_binary.py`)**

- `compute_source_hash` correctly excludes `build/`, `dist/`, `__pycache__`; includes all relevant sources + `poetry.lock`.
- `shutil.copytree(..., symlinks=True)` is the right call for PyInstaller output with symlinked shared libs.
- Hashes both relative path and file content — prevents collisions from file renames.

---

### One Suggestion (non-blocking)

Consider adding Python version to the cache hash to avoid stale cache if the interpreter gets upgraded on CI:

```python
hasher.update(sys.version.encode())
```

Since `poetry.lock` is already hashed, PyInstaller version changes are covered. But Python interpreter version is not reflected in any of the hashed files.

PR #17688, opened 2026-05-16
Fix wrong metric name in NetMetrics.unbindFrom for transmitted bytes

HxpSerein:
## Description

unbindFrom used TRANSMIT ("transmit") as the metric name instead of SystemMetric.TRANSMITTED_BYTES, causing the gauge to not be removed.


SpriCoder (approved):
LGTM！

PR #17694, opened 2026-05-17
Shard subscription-tree-regression-consumer to speed up Multi-Cluster IT

JackieTien97:
## Summary

- Splits the `subscription-tree-regression-consumer` job in `pipe-it.yml` into 3 parallel matrix shards. Expected: ~30–45 min → ~12–18 min per shard, removing it as the Multi-Cluster IT pipeline's long pole.
- Reuses the `\$RUNNER_TEMP/it-shard.txt` + `-Dfailsafe.includesFile` pattern from PR #17692 (the 1C1D Windows sharding work).
- Emits paths relative to `src/test/java/` (e.g. `org/apache/iotdb/.../IoTDBFooIT.java`) rather than bare class names — needed because this suite has 6 pairs of duplicate simple names across `pushconsumer/multi/` and `pullconsumer/multi/` that would otherwise run twice across shards.

## Why this job specifically

The Multi-Cluster IT pipeline currently runs 11 parallel jobs. All others finish in 10–20 min; `subscription-tree-regression-consumer` runs 72 IT classes serially in a single `forkCount=1` JVM, each restarting a 2-cluster environment. That single job dictates the whole pipeline's wall clock on every PR.

Other subscription/dual-cluster jobs in `pipe-it.yml` are not sharded:
- `subscription-tree-regression-misc` (13 classes) — borderline; defer to follow-up
- arch-verification jobs (1–4 classes each) — sharding overhead would exceed savings
- dual-tree/dual-table jobs (9–13 classes) — already under the new shard wall clock

## Local verification

```
$ grep -rlE --include='*IT.java' '\bMultiClusterIT2SubscriptionTreeRegressionConsumer\b' integration-test/src/test/java | wc -l
72
$ for s in 0 1 2; do
    grep -rlE --include='*IT.java' '\bMultiClusterIT2SubscriptionTreeRegressionConsumer\b' integration-test/src/test/java \
      | sed 's|.*/src/test/java/||' | sort | awk -v s=\$s -v t=3 'NR%t==s' | wc -l
  done
24
24
24
$ # And unique-paths == total (i.e. no collisions after disambiguation):
$ grep -rlE --include='*IT.java' '\bMultiClusterIT2SubscriptionTreeRegressionConsumer\b' integration-test/src/test/java | sed 's|.*/src/test/java/||' | sort -u | wc -l
72
```

## Test plan

- [ ] CI: 3 parallel `subscription-tree-regression-consumer (…, 0/1/2)` jobs appear under `Multi-Cluster IT` and all go green.
- [ ] Each shard finishes in ~12–18 min (down from ~30–45 min for the single un-sharded job).
- [ ] No \`Files with unapproved licenses\` warning referencing \`it-shard.txt\` in any shard's log.
- [ ] Union of executed test classes across the 3 shards == 72 (no class missing, no class run twice — verify via surefire-reports artifacts).

## Tracker

This is item #2 from `Remaining bottlenecks` in the CI optimization status doc. The other two (`AINode cold build` and broader Subscription/daily-it sharding) will be addressed separately.

### DIFF
--- a/.github/workflows/pipe-it.yml
+++ b/.github/workflows/pipe-it.yml
@@ -548,6 +548,8 @@ jobs:
           name: cluster-log-subscription-table-arch-verification-java${{ matrix.java }}-${{ runner.os }}-${{ matrix.cluster1 }}-${{ matrix.cluster2 }}
           path: integration-test/target/cluster-logs
           retention-days: 30
+  # 72 IT classes split across 3 parallel shards to cut the longest-pole job
+  # from ~30-45 min to ~12-18 min. See cluster-it-1c1d.yml for the prior art.
   subscription-tree-regression-consumer:
     strategy:
       fail-fast: false
@@ -558,6 +560,7 @@ jobs:
         cluster1: [ScalableSingleNodeMode]
         cluster2: [ScalableSingleNodeMode]
         os: [ubuntu-latest]
+        shard: [0, 1, 2]
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v5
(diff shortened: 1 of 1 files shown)


JackieTien97:
Closing — this PR optimized the wrong target.

After looking at actual durations from 3 recent successful Multi-Cluster IT runs, `subscription-tree-regression-consumer` was already finishing in ~5 minutes unsharded; my pre-PR estimate of 30-45 min was wrong. The real long poles are the dual-* jobs:

| Job                                       | Duration | Classes |
| ----------------------------------------- | -------- | ------- |
| dual-table-manual-basic                   | ~63 min  | 13      |
| dual-table-manual-enhanced                | ~62 min  | 11      |
| dual-tree-auto-enhanced                   | ~51 min  | 9       |
| dual-tree-auto-basic                      | ~42 min  | 12      |
| dual-tree-manual                          | ~27 min  | 11      |
| subscription-tree-regression-consumer     | ~5 min   | 72      |

Sharding subscription added ~10 runner-min per CI run for zero wall-clock benefit. A follow-up PR will shard the dual-* jobs above instead — that should cut Multi-Cluster IT wall clock from ~63 min to ~22 min.

PR #17711, opened 2026-05-18
Support OBJECT data in IoTConsensusV2

Pengzna:
## Description

This PR adds OBJECT data support to IoTConsensusV2.

- Transfers object file pieces before the corresponding InsertNode or TsFile payload reaches followers.
- Applies object file pieces on the receiver side through DataRegion object writes, then writes the final row metadata.
- Preserves OBJECT paths across region-id remapping and table validation paths.
- Decodes and parses OBJECT values in tablet and TsFile parsing flows, including a lightweight TsFile object descriptor scan for IoTConsensusV2 replication.
- Adds focused Stream and Batch IoTConsensusV2 IT coverage for OBJECT replica consistency and follower query after leader stop.

## Tests

- mvn compile -pl iotdb-core/datanode -am -P with-integration-tests -DskipTests
- mvn clean verify -Drat.skip=true -DskipUTs -Dit.test=IoTDBIoTConsensusV2Stream3C3DBasicIT#testObjectReplicaConsistency -DfailIfNoTests=false -Dfailsafe.failIfNoSpecifiedTests=false -pl integration-test -am -PDailyIT -P with-integration-tests
- mvn verify -Drat.skip=true -DskipUTs -Dit.test=IoTDBIoTConsensusV2Batch3C3DBasicIT#testObjectReplicaConsistency -DfailIfNoTests=false -Dfailsafe.failIfNoSpecifiedTests=false -pl integration-test -am -PDailyIT -P with-integration-tests


Caideyipi (commented):
I reviewed the IoTConsensusV2 OBJECT replication path and ran `mvn -Ddevelocity.off=true -pl iotdb-core/calc-commons,iotdb-core/node-commons,iotdb-core/datanode -DskipTests compile` successfully. I left a few correctness comments below.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/write/RelationalInsertTabletNode.java
```diff
@@ -404,6 +405,11 @@ protected List<WritePlanNode> doSplit(Map<TRegionReplicaSet, List<Integer>> spli
         // Avoid using system arraycopy when there is no need to split
         setRange(entry.getValue());
         setDataRegionReplicaSet(entry.getKey());
+        for (int i = 0; i < columns.length; i++) {
+          if (dataTypes[i] == TSDataType.OBJECT) {
+            handleObjectValue(i, 0, times.length, entry, result);
```
Caideyipi: `times.length` can be the tablet capacity, not the number of valid rows in this insert. If a reused tablet has stale OBJECT values after `rowCount`, this loop will generate ObjectNodes for rows that are not part of the current write and may create extra object rows on the follower. Please bound this by `rowCount` (or the active split range) and add coverage where the backing arrays are larger than the current row count.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/write/RelationalInsertTabletNode.java
```diff
@@ -455,6 +471,44 @@ private List<WritePlanNode> generateOneSplitList(
...
   }
 
+  private void handleObjectValue(
+      int column,
+      int startRow,
+      int endRow,
+      Map.Entry<TRegionReplicaSet, List<Integer>> entry,
+      List<WritePlanNode> result) {
+    for (int row = startRow; row < endRow; row++) {
+      if (((Binary[]) columns[column])[row] == null) {
```
Caideyipi, line 486: This should honor the null bitmap before reading the `Binary` slot. For OBJECT columns a null row can still leave a stale value in a reused backing array; this code will transfer that object and only later mark the row null. Please mirror the descriptor collection logic and skip rows where `bitMaps != null && bitMaps[column] != null && bitMaps[column].isMarked(row)`.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/sink/protocol/iotconsensusv2/payload/request/IoTConsensusV2ObjectFilePieceReq.java
```diff
@@ -0,0 +1,107 @@
...
+      final int thisDataNodeId) {
+    final IoTConsensusV2ObjectFilePieceReq req = new IoTConsensusV2ObjectFilePieceReq();
+
+    req.objectNode = objectNode;
+    req.commitId = commitId;
+    req.consensusGroupId = consensusGroupId;
+    req.dataNodeId = thisDataNodeId;
+    req.version = IoTConsensusV2RequestVersion.VERSION_1.getVersion();
+    req.type = IoTConsensusV2RequestType.TRANSFER_OBJECT_FILE_PIECE.getType();
+    req.body = objectNode.serialize();
```
Caideyipi, line 76: `objectNode.serialize()` currently treats source-file read failures as a successful transfer: `ObjectNode.serialize()` logs only at debug, writes `readSuccess && isEOF`, and still sends the zero-filled `contents` buffer. It also uses `RandomAccessFile.read(byte[])` without validating the byte count. With this new IoTConsensusV2 path, a removed/truncated object file or short read can be accepted by the follower as corrupted content. Please fail fast here (and use `readFully` or validate the bytes read) so the event is retried instead of committed with bad object data.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/write/RelationalInsertTabletNode.java
Pengzna: 已修复，handleObjectValue() 会先检查 null bitmap，跳过 marked row 后才读取 Binary；补了 stale backing value + null bitmap 的覆盖。

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/write/RelationalInsertTabletNode.java
Pengzna: 已修复，OBJECT no-split 处理和 redirect 构造都改为按 rowCount 遍历；补了 backing array 长于 rowCount 时不会生成额外 ObjectNode 的单测。

iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/utils/ObjectTypeUtils.java
```diff
@@ -109,7 +111,23 @@ public static Binary replaceRegionIdForObjectBinary(int newRegionId, Binary orig
...
+      final IObjectPath newObjectPath;
+      if (objectPath instanceof PlainObjectPath) {
+        newObjectPath =
+            new PlainObjectPath(objectPath.toString().replaceFirst(regionId + "", newRegionId + ""));
+      } else {
+        final String[] subPath = new String[path.getNameCount() - 1];
+        for (int i = 1; i < path.getNameCount(); i++) {
+          subPath[i - 1] = path.getName(i).toString();
+        }
+        newObjectPath = new Base32ObjectPath(Paths.get(newRegionId + "", subPath));
```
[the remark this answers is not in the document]
Pengzna: 已修复，replaceRegionIdForObjectBinary() 现在从 Path segments 重建对象路径，只替换第 0 个 regionId segment，避免字符串 replaceFirst 误替换路径中其他数字。

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/sink/protocol/iotconsensusv2/payload/request/IoTConsensusV2ObjectFilePieceReq.java
Pengzna: 已修复，ObjectNode.serialize() 现在在 object/tmp 文件均读不到或短读时直接抛 SerializationRunTimeException，读取内容改为 readFully，避免提交零填充/截断内容；补了缺失文件 serialize 失败测试。

PR #17723, opened 2026-05-20
Fix column-prune of FillNode

Wei-hao-Li:


JackieTien97 (approved):

anukalp2804:
Hi @JackieTien97 ,

I noticed two issues in this PR, so I created a separate PR to address them with cleaner and more consistent code.
Could you please review it and let me know if any changes are needed from my side?
Address: #17772

PR #17732, opened 2026-05-21
Fixed UDF jar metadata handling in `UDFInfo` when multiple UDFs share the same jar

Caideyipi:
## Description

  This PR fixes UDF jar metadata handling in `UDFInfo` when multiple UDFs share the same jar.

  ### Background

  `UDFInfo` previously tracked uploaded jars only with `jarName -> md5`. When a UDF was dropped, the jar metadata was
  removed immediately. This breaks the case where multiple UDFs reference the same jar:
  - dropping one UDF can make ConfigNode think the shared jar no longer exists
  - later validation may no longer detect conflicting MD5 values for the same jar name
  - after snapshot load, jar metadata is restored but shared-jar reference state is not rebuilt, so the same issue can
  reappear after restart

  ### Changes

  This PR introduces reference counting for shared UDF jars in `UDFInfo`.
  - add `existedJarToReferenceCount` to track how many UDFs are using each jar
  - update UDF creation flow to increase jar reference count instead of only recording MD5
  - update UDF drop flow to remove jar metadata only when the last reference is removed
  - rebuild jar metadata and reference counts from the UDF table after loading a snapshot

  This keeps shared jar metadata consistent across normal create/drop operations and snapshot recovery.

  ### Tests

  Updated `UDFInfoTest` to cover the shared-jar cases:
  - dropping one UDF does not remove jar metadata if another UDF still references the same jar
  - validation still rejects the same jar name with a different MD5 after one reference is dropped
  - snapshot load rebuilds shared-jar metadata correctly, and subsequent drop behavior remains correct

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/UDFInfo.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/UDFInfo.java
@@ -68,6 +68,7 @@ public class UDFInfo implements SnapshotProcessor {
 
   private final UDFTable udfTable;
   private final Map<String, String> existedJarToMD5;
+  private final Map<String, Integer> existedJarToReferenceCount;
 
   private final UDFExecutableManager udfExecutableManager;
 
@@ -78,6 +79,7 @@ public class UDFInfo implements SnapshotProcessor {
   public UDFInfo() throws IOException {
     udfTable = new UDFTable();
     existedJarToMD5 = new HashMap<>();
+    existedJarToReferenceCount = new HashMap<>();
     udfExecutableManager =
         UDFExecutableManager.setupAndGetInstance(
             CONFIG_NODE_CONF.getUdfTemporaryLibDir(), CONFIG_NODE_CONF.getUdfDir());
@@ -135,7 +137,7 @@ public class UDFInfo implements SnapshotProcessor {
       final UDFInformation udfInformation = physicalPlan.getUdfInformation();
       udfTable.addUDFInformation(udfInformation.getFunctionName(), udfInformation);
       if (udfInformation.isUsingURI()) {
-        existedJarToMD5.put(udfInformation.getJarName(), udfInformation.getJarMD5());
+        addJarReference(udfInformation.getJarName(), udfInformation.getJarMD5());
         if (physicalPlan.getJarFile() != null) {
           udfExecutableManager.saveToInstallDir(
               ByteBuffer.wrap(physicalPlan.getJarFile().getValues()), udfInformation.getJarName());
@@ -185,7 +187,10 @@ public class UDFInfo implements SnapshotProcessor {
 
   public TSStatus dropFunction(Model model, String functionName) {
     if (udfTable.containsUDF(model, functionName)) {
-      existedJarToMD5.remove(udfTable.getUDFInformation(model, functionName).getJarName());
+      final UDFInformation udfInformation = udfTable.getUDFInformation(model, functionName);
+      if (udfInformation.isUsingURI()) {
+        removeJarReference(udfInformation.getJarName());
+      }
       udfTable.removeUDFInformation(model, functionName);
     }
     return new TSStatus(TSStatusCode.SUCCESS_STATUS.getStatusCode());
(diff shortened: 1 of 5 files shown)


luoluoyuyu (commented):
## Review 总结

整体方向正确：用引用计数修复「多 UDF 共享同一 jar 时 drop 一个误删 jar 元数据」的问题，`rebuildJarMetadataFromUDFTable()` 也能在 snapshot 恢复后重建一致性。UT 覆盖了共享 jar 与 snapshot 场景，质量不错。

建议在合入前处理下面 2 点（见行内评论），其中 MD5 一致性建议在 `addJarReference` 内防御性校验。

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/UDFInfo.java
```diff
@@ -272,6 +278,32 @@ public void deserializeExistedJarToMD5(InputStream inputStream) throws IOExcepti
 
   public void clear() {
     existedJarToMD5.clear();
+    existedJarToReferenceCount.clear();
     udfTable.clear();
   }
+
+  private void addJarReference(String jarName, String jarMD5) {
+    existedJarToMD5.putIfAbsent(jarName, jarMD5);
```
luoluoyuyu: putIfAbsent keeps the first MD5 for a jar name but always increments the reference count. That is fine when validate() runs before addUDFInTable, but if addJarReference is ever called without that check, ref count and MD5 could diverge.

Consider rejecting a conflicting MD5 inside addJarReference (same check as in validate() at lines 107-115) so this helper is safe on its own.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/UDFInfo.java
```diff
@@ -248,6 +253,7 @@ public void processLoadSnapshot(File snapshotDir) throws IOException {
       deserializeExistedJarToMD5(fileInputStream);
 
       udfTable.deserializeUDFTable(fileInputStream);
+      rebuildJarMetadataFromUDFTable();
```
luoluoyuyu, line 254: After deserializeExistedJarToMD5, rebuildJarMetadataFromUDFTable() clears both maps and rebuilds from udfTable. That makes udfTable the source of truth on load.

A short comment here explaining that deserialized existedJarToMD5 is intentionally discarded would help future readers.

luoluoyuyu (commented):
## Review summary

Reference counting for shared JARs fixes incorrect removal of existedJarToMD5 when only one of several UDFs is dropped. rebuildJarMetadataFromUDFTable() keeps jar metadata consistent after snapshot load. New unit tests cover the shared-jar and snapshot paths.

Two small suggestions below; otherwise this looks good to merge.

jt2594838 (approved):
May unify the jar management of UDF and PIPE (and even more).

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/executable/ReferenceCountedJarMetaKeeper.java
```diff
@@ -0,0 +1,121 @@
...
+  public synchronized void serializeJarNameToMd5AndReferenceCount(final OutputStream outputStream)
+      throws IOException {
+    ReadWriteIOUtils.write(jarNameToMd5Map.size(), outputStream);
+    for (final Map.Entry<String, String> entry : jarNameToMd5Map.entrySet()) {
+      final String jarName = entry.getKey();
+      ReadWriteIOUtils.write(jarName, outputStream);
+      ReadWriteIOUtils.write(entry.getValue(), outputStream);
+      ReadWriteIOUtils.write(jarNameToReferenceCountMap.getOrDefault(jarName, 0), outputStream);
+    }
+  }
```
jt2594838, line 119: Is it possible or necessary to serialize an entry whose count is zero?

Caideyipi: Good point. A zero reference count should not be serialized: under the normal add/remove path, removeReference removes both maps when the count reaches zero. I updated the snapshot logic to serialize only entries with a positive reference count and matching md5 metadata, and to skip non-positive counts when loading snapshots as a defensive cleanup. Added a regression test for the zero-count snapshot case as well.

jt2594838 (approved):

PR #17734, opened 2026-05-21
Fix CQ recovery gap and stale callback contamination

Caideyipi:
## Description
This PR fixes two CQ issues in ConfigNode:

  - Recovering an ACTIVE CQ now resubmits the schedule task even if ConfigNode restarts between metadata activation and
    task submission.
  - CQ schedule tokens are now unique per create instance, so stale async callbacks from a dropped CQ cannot affect a
    recreated CQ with the same cqId.

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


luoluoyuyu (commented):
## Review 总结

`locallyScheduledCQs` + `markCQLocallyScheduled` 能有效防止 CN 重启或 procedure 重入时同一 CQ 被重复 submit，drop 时清理 map 也正确。`CreateCQProcedure` 侧改动需结合完整 diff 再确认 token 生成与 recovery 路径。

整体可合入，建议处理行内关于 `markCQLocallyScheduled` 边界条件的说明/测试。

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/cq/CQManager.java
```diff
@@ -176,11 +195,30 @@ public void stopCQScheduler() {
...
       previous.shutdown();
     }
   }
+
+  public boolean markCQLocallyScheduled(String cqId, String md5) {
+    AtomicBoolean shouldSchedule = new AtomicBoolean(false);
+    locallyScheduledCQs.compute(
+        cqId,
+        (ignored, previousMd5) -> {
+          if (!md5.equals(previousMd5)) {
```
luoluoyuyu: When previousMd5 is null, !md5.equals(previousMd5) schedules the CQ as expected. When previousMd5 equals md5, scheduling is skipped, which avoids duplicates.

If the same cqId is updated with a new md5, this returns true and schedules again. Please confirm in this PR that any previously running CQScheduleTask for the old md5 is stopped or superseded; otherwise two tasks could run briefly.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/cq/CQManager.java
```diff
@@ -176,11 +195,30 @@ public void stopCQScheduler() {
...
+            shouldSchedule.set(true);
+            return md5;
+          }
+          return previousMd5;
+        });
+    return shouldSchedule.get();
+  }
+
+  public void unmarkCQLocallyScheduled(String cqId, String md5) {
+    locallyScheduledCQs.remove(cqId, md5);
```
luoluoyuyu: Using remove(cqId, md5) ensures we only unmark the entry we own after a failed submitSelf. Good defensive pattern.

luoluoyuyu (commented):
## Review summary

locallyScheduledCQs prevents duplicate CQScheduleTask submission on scheduler restart or procedure retry. dropCQ clears the map on success; submitSelf failure rolls back via unmarkCQLocallyScheduled.

Please address the inline note on markCQLocallyScheduled when md5 changes. Otherwise the approach looks sound.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/cq/CQManager.java
```diff
@@ -57,11 +60,14 @@ public class CQManager {
 
   private final ReadWriteLock lock;
 
+  private final ConcurrentMap<String, String> locallyScheduledCQs;
```
jt2594838: Explain the key and the value.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/cq/CQManager.java
```diff
@@ -79,7 +85,11 @@ public TSStatus createCQ(TCreateCQReq req) {
 
   public TSStatus dropCQ(TDropCQReq req) {
     try {
-      return configManager.getConsensusManager().write(new DropCQPlan(req.cqId));
+      TSStatus status = configManager.getConsensusManager().write(new DropCQPlan(req.cqId));
+      if (status.getCode() == TSStatusCode.SUCCESS_STATUS.getStatusCode()) {
+        locallyScheduledCQs.remove(req.cqId);
+      }
+      return status;
```
jt2594838, line 94: Dropping a CQ does not abort it？

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/cq/CQManager.java
```diff
@@ -155,8 +166,16 @@ public void startCQScheduler() {
...
+              continue;
+            }
             CQScheduleTask cqScheduleTask = new CQScheduleTask(entry, executor, configManager);
-            cqScheduleTask.submitSelf();
+            try {
+              cqScheduleTask.submitSelf();
+            } catch (RuntimeException e) {
+              unmarkCQLocallyScheduled(entry.getCqId(), entry.getMd5());
+              throw e;
+            }
```
jt2594838: Add the task as part of the value in locallyScheduledCQs?
So that you may abort it when dropping.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/cq/CreateCQProcedure.java
```diff
@@ -75,7 +81,7 @@ public CreateCQProcedure(ScheduledExecutorService executor) {
   public CreateCQProcedure(TCreateCQReq req, ScheduledExecutorService executor) {
     super();
     this.req = req;
-    this.md5 = DigestUtils.md2Hex(req.cqId);
+    this.md5 = generateCQToken(req.cqId);
```
jt2594838: Improper to call it md5 now.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/cq/CreateCQProcedure.java
```diff
@@ -168,6 +177,42 @@ private void activeCQ(ConfigNodeProcedureEnv env) {
...
+    submitScheduleTask(env, new CQScheduleTask(cqEntry.get(), executor, env.getConfigManager()));
+  }
+
+  Optional<CQInfo.CQEntry> getCurrentCQEntry(ConfigNodeProcedureEnv env) throws ConsensusException {
+    ShowCQResp response =
+        (ShowCQResp) env.getConfigManager().getConsensusManager().read(new ShowCQPlan());
+    return response.getCqList().stream()
+        .filter(entry -> req.cqId.equals(entry.getCqId()) && md5.equals(entry.getMd5()))
+        .findFirst();
+  }
```
jt2594838, line 197: Is it possible to moderate ShowCQPlan to only return the desired CQs instead of filtering at the upper level?

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/cq/CreateCQProcedure.java
```diff
@@ -168,6 +177,42 @@ private void activeCQ(ConfigNodeProcedureEnv env) {
...
+
+  private static String generateCQToken(String cqId) {
+    return DigestUtils.md2Hex(cqId + "-" + UUID.randomUUID());
+  }
```
jt2594838: Now that you use UUID, it is no longer necessary to use md5.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/cq/CreateCQProcedure.java
Caideyipi: Done in f09c3b0: renamed the CQ identifier from md5 to cqToken across the CQ plans, metadata, procedure and scheduler code.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/cq/CQManager.java
Caideyipi: Done in f09c3b0: the local schedule registry now stores the metadata token together with the CQScheduleTask. If the same CQ id is registered with a different token, the old task is cancelled before the new one is scheduled. Drop/start/stop also cancel local tasks, and cancelled callbacks no longer resubmit.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/cq/CQManager.java
Caideyipi: Done in f09c3b0: dropCQ now coordinates with scheduler start/stop and cancels the local CQScheduleTask after the metadata drop succeeds, so queued future executions are aborted.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/cq/CreateCQProcedure.java
Caideyipi: Done in f09c3b0: generateCQToken now uses UUID.randomUUID().toString() directly and the DigestUtils/md2Hex dependency was removed from CreateCQProcedure.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/cq/CQManager.java
Caideyipi: Done in f09c3b0: locallyScheduledCQs now stores a LocallyScheduledCQ value containing the token and CQScheduleTask. CQScheduleTask owns its latest ScheduledFuture and exposes cancel(), so drop and token replacement can cancel it.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/cq/CQManager.java
Caideyipi: Done in f09c3b0: added an inline explanation. The key is CQ id, and the value is the local task plus the metadata token it owns.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/cq/CreateCQProcedure.java
Caideyipi: Done in f09c3b0: ShowCQPlan now accepts an optional CQ id, and CQInfo filters by CQ id before returning the response. The recovery path now reads only the target CQ and then checks the token.

PR #17735, opened 2026-05-21
Fixed TTL problems

Caideyipi:
## Description

  This PR fixes two correctness issues in ConfigNode TTL handling.

  1. TTL rule capacity checking incorrectly treated updates to existing TTL rules as new rules.
  2. `SetTTLProcedure` did not rollback TTL state when ConfigNode metadata had been updated but DataNode TTL cache
  update failed.

  ## Details

  ### 1. Fix TTL rule capacity accounting

  `TTLInfo` now checks the number of newly introduced TTL rules instead of simply checking the current total rule count.

  This also fixes the database TTL case: setting TTL on a database effectively writes both the database path and its
  `**` wildcard path, so capacity validation now accounts for both entries.

  As a result:
  - updating an existing TTL rule no longer fails when the capacity limit is already reached
  - database-level TTL no longer bypasses the real capacity requirement

  ### 2. Add rollback for `SetTTLProcedure`

  `SetTTLProcedure` now captures the previous TTL state before writing the new value to ConfigNode.

  If the procedure fails while updating DataNode TTL cache, it will rollback:
  - the TTL metadata on ConfigNode
  - the TTL cache on DataNodes

  For database TTL, the rollback also restores the wildcard TTL entry (`db.**`) separately.

  The rollback state is serialized with the procedure so that recovery/replay can still restore the previous TTL state
  correctly.

  ## Tests

  Added/updated tests for:
  - updating existing TTL when capacity is reached
  - updating existing TTL when current state is already oversize
  - database TTL capacity accounting
  - rollback when previous TTL does not exist
  - rollback of database wildcard TTL
  - procedure serialization with captured rollback state

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/TTLInfo.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/TTLInfo.java
@@ -175,6 +175,15 @@ public class TTLInfo implements SnapshotProcessor {
     }
   }
 
+  public long getTTL(final String[] pathPattern) {
+    lock.readLock().lock();
+    try {
+      return ttlCache.getLastNodeTTL(pathPattern);
+    } finally {
+      lock.readLock().unlock();
+    }
+  }
+
   /**
    * Get the maximum ttl of the corresponding database level.
    *
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/TTLInfo.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/TTLInfo.java
@@ -71,13 +71,14 @@ public class TTLInfo implements SnapshotProcessor {
       // check ttl rule capacity
       final int tTlRuleCapacity = CommonDescriptor.getInstance().getConfig().getTTlRuleCapacity();
       final int newTTLRuleCount = calculateNewTTLRuleCount(plan);
-      if (newTTLRuleCount > 0 && ttlCache.getTtlCount() + newTTLRuleCount > tTlRuleCapacity) {
+      final int requestedTTLRuleCount = ttlCache.getTtlCount() + newTTLRuleCount;
+      if (newTTLRuleCount > 0 && requestedTTLRuleCount > tTlRuleCapacity) {
         TSStatus errorStatus = new TSStatus(TSStatusCode.OVERSIZE_TTL.getStatusCode());
         errorStatus.setMessage(
             String.format(
-                "The number of TTL rules has reached the limit (%d). Please delete "
-                    + "some existing rules first.",
-                tTlRuleCapacity));
+                "The number of TTL rules has reached the limit "
+                    + "(capacity: %d, requested total: %d). Please delete some existing rules first.",
+                tTlRuleCapacity, requestedTTLRuleCount));
         return errorStatus;
       }
       ttlCache.setTTL(plan.getPathPattern(), plan.getTTL());
@@ -94,17 +95,17 @@ public class TTLInfo implements SnapshotProcessor {
   }
 
   private int calculateNewTTLRuleCount(SetTTLPlan plan) {
-    int newTTLRuleCount = getNewTTLRuleCount(plan.getPathPattern());
+    int newTTLRuleCount = isNewTTLRule(plan.getPathPattern()) ? 1 : 0;
     if (plan.isDataBase()) {
       String[] pathNodes = Arrays.copyOf(plan.getPathPattern(), plan.getPathPattern().length + 1);
       pathNodes[pathNodes.length - 1] = IoTDBConstant.MULTI_LEVEL_PATH_WILDCARD;
-      newTTLRuleCount += getNewTTLRuleCount(pathNodes);
+      newTTLRuleCount += isNewTTLRule(pathNodes) ? 1 : 0;
     }
     return newTTLRuleCount;
   }
 
-  private int getNewTTLRuleCount(String[] pathNodes) {
-    return ttlCache.getLastNodeTTL(pathNodes) == TTLCache.NULL_TTL ? 1 : 0;
+  private boolean isNewTTLRule(String[] pathNodes) {
+    return ttlCache.getLastNodeTTL(pathNodes) == TTLCache.NULL_TTL;
   }
 
   /** Only used for upgrading from database level ttl to device level ttl. */
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/TTLInfo.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/TTLInfo.java
@@ -70,7 +70,8 @@ public class TTLInfo implements SnapshotProcessor {
     try {
       // check ttl rule capacity
       final int tTlRuleCapacity = CommonDescriptor.getInstance().getConfig().getTTlRuleCapacity();
-      if (getTTLCount() >= tTlRuleCapacity) {
+      final int newTTLRuleCount = calculateNewTTLRuleCount(plan);
+      if (newTTLRuleCount > 0 && ttlCache.getTtlCount() + newTTLRuleCount > tTlRuleCapacity) {
         TSStatus errorStatus = new TSStatus(TSStatusCode.OVERSIZE_TTL.getStatusCode());
         errorStatus.setMessage(
             String.format(
@@ -90,6 +91,20 @@ public class TTLInfo implements SnapshotProcessor {
       lock.writeLock().unlock();
     }
     return RpcUtils.getStatus(TSStatusCode.SUCCESS_STATUS);
+  }
+
+  private int calculateNewTTLRuleCount(SetTTLPlan plan) {
+    int newTTLRuleCount = getNewTTLRuleCount(plan.getPathPattern());
+    if (plan.isDataBase()) {
+      String[] pathNodes = Arrays.copyOf(plan.getPathPattern(), plan.getPathPattern().length + 1);
+      pathNodes[pathNodes.length - 1] = IoTDBConstant.MULTI_LEVEL_PATH_WILDCARD;
+      newTTLRuleCount += getNewTTLRuleCount(pathNodes);
+    }
+    return newTTLRuleCount;
+  }
+
+  private int getNewTTLRuleCount(String[] pathNodes) {
+    return ttlCache.getLastNodeTTL(pathNodes) == TTLCache.NULL_TTL ? 1 : 0;
   }
 
   /** Only used for upgrading from database level ttl to device level ttl. */
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/schema/SetTTLProcedure.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/schema/SetTTLProcedure.java
@@ -56,6 +56,7 @@ import java.util.Objects;
 
 public class SetTTLProcedure extends StateMachineProcedure<ConfigNodeProcedureEnv, SetTTLState> {
   private static final Logger LOGGER = LoggerFactory.getLogger(SetTTLProcedure.class);
+  // Distinguishes no previous TTL from TTLCache.NULL_TTL, the explicit unset marker for rollback.
   private static final long TTL_NOT_EXIST = Long.MIN_VALUE;
 
   private SetTTLPlan plan;
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/schema/SetTTLProcedure.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/schema/SetTTLProcedure.java
@@ -302,8 +302,10 @@ public class SetTTLProcedure extends StateMachineProcedure<ConfigNodeProcedureEn
   public void deserialize(ByteBuffer byteBuffer) {
     super.deserialize(byteBuffer);
     try {
-      ReadWriteIOUtils.readInt(byteBuffer);
+      final int length = ReadWriteIOUtils.readInt(byteBuffer);
+      final int position = byteBuffer.position();
       this.plan = (SetTTLPlan) ConfigPhysicalPlan.Factory.create(byteBuffer);
+      byteBuffer.position(position + length);
       if (byteBuffer.remaining() >= 17) {
         this.previousTTLStateCaptured = byteBuffer.get() != 0;
         this.previousTTL = byteBuffer.getLong();
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/schema/SetTTLProcedure.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/schema/SetTTLProcedure.java
@@ -58,6 +58,7 @@ public class SetTTLProcedure extends StateMachineProcedure<ConfigNodeProcedureEn
   private static final Logger LOGGER = LoggerFactory.getLogger(SetTTLProcedure.class);
   // Distinguishes no previous TTL from TTLCache.NULL_TTL, the explicit unset marker for rollback.
   private static final long TTL_NOT_EXIST = Long.MIN_VALUE;
+  private static final int ROLLBACK_STATE_BYTES = Byte.BYTES + Long.BYTES * 2;
 
   private SetTTLPlan plan;
   private long previousTTL = TTL_NOT_EXIST;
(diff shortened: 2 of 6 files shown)


CRZbulabula (commented):
(Review superseded by the English version below.)

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/TTLInfo.java
```diff
@@ -70,7 +70,8 @@ public TSStatus setTTL(SetTTLPlan plan) {
     try {
       // check ttl rule capacity
       final int tTlRuleCapacity = CommonDescriptor.getInstance().getConfig().getTTlRuleCapacity();
-      if (getTTLCount() >= tTlRuleCapacity) {
+      final int newTTLRuleCount = calculateNewTTLRuleCount(plan);
+      if (newTTLRuleCount > 0 && ttlCache.getTtlCount() + newTTLRuleCount > tTlRuleCapacity) {
```
CRZbulabula: **Capacity check looks correct.** Short-circuiting on `newTTLRuleCount > 0` is what lets `testUpdateExistingTTLWhenCurrentStateIsAlreadyOversize` pass — an update of an existing rule no longer trips the limit even when the current count is already over capacity.

One diagnostic suggestion: when the limit is hit, include both `tTlRuleCapacity` and `ttlCache.getTtlCount() + newTTLRuleCount` in the error message. It makes it much easier to tell at a glance whether the cluster genuinely filled the table, or whether the capacity was tightened by a config change.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/TTLInfo.java
```diff
@@ -92,6 +93,20 @@ public TSStatus setTTL(SetTTLPlan plan) {
...
+    if (plan.isDataBase()) {
+      String[] pathNodes = Arrays.copyOf(plan.getPathPattern(), plan.getPathPattern().length + 1);
+      pathNodes[pathNodes.length - 1] = IoTDBConstant.MULTI_LEVEL_PATH_WILDCARD;
+      newTTLRuleCount += getNewTTLRuleCount(pathNodes);
+    }
+    return newTTLRuleCount;
+  }
+
+  private int getNewTTLRuleCount(String[] pathNodes) {
+    return ttlCache.getLastNodeTTL(pathNodes) == TTLCache.NULL_TTL ? 1 : 0;
```
CRZbulabula: `getNewTTLRuleCount` is slightly mis-named — it returns a 0/1 indicator, not a count. Consider renaming to `isNewTTLRule` (return `boolean`) and accumulating `boolean ? 1 : 0` in `calculateNewTTLRuleCount`. Reads more naturally at the call site.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/schema/SetTTLProcedure.java
```diff
@@ -47,14 +49,19 @@
...
 import java.util.Collections;
 import java.util.Map;
 import java.util.Objects;
 
 public class SetTTLProcedure extends StateMachineProcedure<ConfigNodeProcedureEnv, SetTTLState> {
   private static final Logger LOGGER = LoggerFactory.getLogger(SetTTLProcedure.class);
+  private static final long TTL_NOT_EXIST = Long.MIN_VALUE;
 
   private SetTTLPlan plan;
+  private long previousTTL = TTL_NOT_EXIST;
```
CRZbulabula, line 64: Worth a one-line comment explaining why this sentinel is `Long.MIN_VALUE` rather than `TTLCache.NULL_TTL` (`-1`). They have different meanings on the rollback path: `TTL_NOT_EXIST` means "no TTL was set before this procedure", while `NULL_TTL` is the explicit "unset" marker that `ConfigPlanExecutor` interprets to route to `unsetTTL`. Conflating the two would corrupt rollback behavior.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/schema/SetTTLProcedure.java
```diff
@@ -86,18 +93,9 @@ protected Flow executeFromState(ConfigNodeProcedureEnv env, SetTTLState state)
...
-              .getConsensusManager()
-              .write(isGeneratedByPipe ? new PipeEnrichedPlan(this.plan) : this.plan);
-    } catch (ConsensusException e) {
-      LOGGER.warn(ConfigNodeMessages.FAILED_IN_THE_WRITE_API_EXECUTING_THE_CONSENSUS_LAYER_DUE, e);
-      res = new TSStatus(TSStatusCode.EXECUTE_STATEMENT_ERROR.getStatusCode());
-      res.setMessage(e.getMessage());
-    }
+  protected void setConfigNodeTTL(final ConfigNodeProcedureEnv env) {
+    capturePreviousTTLState(env);
+    final TSStatus res = writeConfigNodePlan(env, plan);
```
CRZbulabula, line 108: **(medium) `previousTTL` can be overwritten with the new value after a crash.**

`capturePreviousTTLState` and `writeConfigNodePlan` both run inside one `executeFromState(SET_CONFIGNODE_TTL)` invocation, but the procedure framework only persists `previousTTLStateCaptured` / `previousTTL` on state transitions. If the ConfigNode crashes after the consensus write succeeds but before `setNextState(UPDATE_DATANODE_CACHE)` is durable, recovery replays `SET_CONFIGNODE_TTL`: `previousTTLStateCaptured` is still `false`, so we re-run capture — and `getAllTTL()` now returns the **new** value, which gets recorded as the "previous" TTL. If the DataNode update then fails, the rollback restores the new value over itself, losing the original.

Suggestion: split capture into its own pre-state (e.g. `CAPTURE_PREVIOUS_TTL` → `SET_CONFIGNODE_TTL` → `UPDATE_DATANODE_CACHE`) so the captured snapshot is persisted with procedure state *before* the consensus write.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/schema/SetTTLProcedure.java
```diff
@@ -86,18 +93,9 @@ protected Flow executeFromState(ConfigNodeProcedureEnv env, SetTTLState state)
...
-      res =
-          env.getConfigManager()
-              .getConsensusManager()
-              .write(isGeneratedByPipe ? new PipeEnrichedPlan(this.plan) : this.plan);
-    } catch (ConsensusException e) {
-      LOGGER.warn(ConfigNodeMessages.FAILED_IN_THE_WRITE_API_EXECUTING_THE_CONSENSUS_LAYER_DUE, e);
-      res = new TSStatus(TSStatusCode.EXECUTE_STATEMENT_ERROR.getStatusCode());
-      res.setMessage(e.getMessage());
-    }
+  protected void setConfigNodeTTL(final ConfigNodeProcedureEnv env) {
```
CRZbulabula: These methods (`setConfigNodeTTL`, `updateDataNodeTTL`, `writeConfigNodePlan`, `sendTTLRequest`) are made `protected` only so that `TestingSetTTLProcedure` in the test file can override them — no production subclass needs them. Consider making them package-private and putting the test class in the same package; that keeps the public/protected surface narrower.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/schema/SetTTLProcedure.java
```diff
@@ -106,35 +104,171 @@ private void setConfigNodeTTL(ConfigNodeProcedureEnv env) {
...
+          new ProcedureException(
+              new MetadataException(ProcedureMessages.UPDATE_DATANODE_TTL_CACHE_FAILED)));
+    }
+  }
+
+  private void capturePreviousTTLState(final ConfigNodeProcedureEnv env) {
+    if (previousTTLStateCaptured) {
+      return;
+    }
+    final Map<String, Long> ttlMap = env.getConfigManager().getTTLManager().getAllTTL();
```
CRZbulabula: **Performance: a full consensus read for one or two entries.**

`TTLManager.getAllTTL()` goes through `getConsensusManager().read(showTTLPlan)`, which reads the entire TTL table across consensus and serializes the DFS result into a `Map`. Every `SetTTLProcedure` now pays this cost, which becomes noticeable on clusters approaching `tTlRuleCapacity`.

Consider adding a targeted `getTTL(String[] pathPattern)` to `TTLManager` / `TTLInfo` (e.g. wrapping `TTLCache.getLastNodeTTL`) and calling it twice when `isDataBase()` is true.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/schema/SetTTLProcedure.java
```diff
@@ -106,35 +104,171 @@ private void setConfigNodeTTL(ConfigNodeProcedureEnv env) {
...
+      restoreTTLOnConfigNode(
+          env, getDatabaseWildcardPathPattern(plan.getPathPattern()), previousDatabaseWildcardTTL);
+    }
+  }
+
+  private void restoreTTLOnConfigNode(
+      final ConfigNodeProcedureEnv env, final String[] pathPattern, final long ttl)
+      throws ProcedureException {
+    final SetTTLPlan rollbackPlan =
+        new SetTTLPlan(pathPattern, ttl == TTL_NOT_EXIST ? TTLCache.NULL_TTL : ttl);
```
CRZbulabula, line 208: The `ttl == TTL_NOT_EXIST ? TTLCache.NULL_TTL : ttl` translation deserves an inline comment. It encodes a cross-file convention: a `SetTTLPlan` whose `ttl == NULL_TTL` is interpreted by `ConfigPlanExecutor.SetTTL` (file `ConfigPlanExecutor.java:430`) as a signal to call `ttlInfo.unsetTTL` rather than `setTTL`, and similarly `StorageEngine.setTTL` (`StorageEngine.java:973`) routes to `unsetTTLForTree`. Reading the rollback code in isolation, it's not obvious that writing `-1` is what undoes the previous SET — a one-line note here would save a lot of code-spelunking.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/schema/SetTTLProcedure.java
```diff
@@ -106,35 +104,171 @@ private void setConfigNodeTTL(ConfigNodeProcedureEnv env) {
...
+    rollbackPlan.setDataBase(false);
```
CRZbulabula, line 210: The explicit `setDataBase(false)` here is correct but non-obvious — please add a comment. The reason: in the database case, `rollbackConfigNodeTTL` already issues two separate `restoreTTLOnConfigNode` calls (one for `db`, one for `db.**`). If this inner `SetTTLPlan` carried `isDataBase=true`, `TTLInfo.setTTL` would auto-expand the wildcard *again*, overwriting the just-restored `db.**` value with the value being applied to `db`. Forcing `isDataBase=false` keeps each path independent.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/schema/SetTTLProcedure.java
```diff
@@ -106,35 +104,171 @@ private void setConfigNodeTTL(ConfigNodeProcedureEnv env) {
...
+    }
+    ProcedureException rollbackFailure = null;
+    try {
+      rollbackConfigNodeTTL(env);
+    } catch (ProcedureException e) {
+      LOGGER.error("Failed to rollback ConfigNode ttl state.", e);
+      rollbackFailure = e;
+    }
+    try {
+      rollbackDataNodeTTL(env);
```
CRZbulabula, line 270: Two suggestions on the rollback error-aggregation strategy:

1. Only the *first* exception survives; the second is logged but lost. Use `rollbackFailure.addSuppressed(e)` so a postmortem can still see both.
2. Add a method-level comment clarifying the policy: "best-effort — if one side fails, the other is still attempted; the earliest exception is what propagates." Otherwise readers might expect fail-fast semantics.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/schema/SetTTLProcedure.java
```diff
@@ -159,14 +293,24 @@ public void serialize(DataOutputStream stream) throws IOException {
...
 
   @Override
   public void deserialize(ByteBuffer byteBuffer) {
     super.deserialize(byteBuffer);
     try {
-      ReadWriteIOUtils.readInt(byteBuffer);
+      final int length = ReadWriteIOUtils.readInt(byteBuffer);
+      final int position = byteBuffer.position();
       this.plan = (SetTTLPlan) ConfigPhysicalPlan.Factory.create(byteBuffer);
+      byteBuffer.position(position + length);
```
CRZbulabula, line 325: This `byteBuffer.position(position + length)` is necessary (not redundant) — worth a comment so a future reader doesn't "clean it up".

Reason: `serializeToByteBuffer()` returns `ByteBuffer.wrap(publicBAOS.getBuf(), 0, size())`. `ReadWriteIOUtils.write(ByteBuffer, OutputStream)` then writes `byteBuffer.capacity()` (the *full backing array length*, which can be greater than the actual data size) followed by the whole array, padding included. Without forcing the position back to `position + length`, the new `boolean + 2*long` trailing fields would be read from those padding zero bytes.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/schema/SetTTLProcedure.java
```diff
@@ -159,14 +293,24 @@ public void serialize(DataOutputStream stream) throws IOException {
...
+      if (byteBuffer.remaining() >= 17) {
```
CRZbulabula: `17` is a magic number (`1 byte boolean + 2 * 8 byte long`). Extract a constant:

```java
private static final int ROLLBACK_STATE_BYTES = Byte.BYTES + Long.BYTES * 2;
...
if (byteBuffer.remaining() >= ROLLBACK_STATE_BYTES) { ... }
```

Makes it survive any future addition of rollback fields without forgetting to update the guard.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/schema/SetTTLProcedure.java
```diff
@@ -180,12 +324,21 @@ public boolean equals(Object o) {
     if (o == null || getClass() != o.getClass()) {
       return false;
     }
-    return this.plan.equals(((SetTTLProcedure) o).plan)
-        && this.isGeneratedByPipe == (((SetTTLProcedure) o).isGeneratedByPipe);
+    final SetTTLProcedure that = (SetTTLProcedure) o;
+    return this.isGeneratedByPipe == that.isGeneratedByPipe
+        && this.previousTTL == that.previousTTL
+        && this.previousDatabaseWildcardTTL == that.previousDatabaseWildcardTTL
+        && this.previousTTLStateCaptured == that.previousTTLStateCaptured
```
CRZbulabula, line 348: Including `previousTTL` / `previousDatabaseWildcardTTL` / `previousTTLStateCaptured` in `equals`/`hashCode` is consistent with the serialization change, but please confirm no caller dedups procedures by `equals` on the wait queue. If something does, two procedures with the same `plan` but different captured state would now be considered distinct, which could change idempotency behavior. If there's no such caller, ignore this comment.

CRZbulabula (commented):
Both the capacity-accounting and the rollback-on-DataNode-failure fixes look correct. Comments below are organized by code location and focus on the timing of state capture in `SetTTLProcedure`, the cost of the consensus read used to capture it, and a handful of readability/test improvements.

iotdb-core/confignode/src/test/java/org/apache/iotdb/confignode/procedure/impl/schema/SetTTLProcedureTest.java
```diff
@@ -65,4 +83,234 @@ public void serializeDeserializeTest() throws IOException, IllegalPathException
...
+    procedure.executeFromState(mockProcedureEnv(ttlMap), SetTTLState.SET_CONFIGNODE_TTL);
+
+    procedure.serialize(outputStream);
+    final ByteBuffer buffer =
+        ByteBuffer.wrap(byteArrayOutputStream.getBuf(), 0, byteArrayOutputStream.size());
+    final SetTTLProcedure deserializedProcedure =
+        (SetTTLProcedure) ProcedureFactory.getInstance().create(buffer);
+    assertSerializedProcedure(
+        deserializedProcedure, "root.db", 2000L, true, true, 500L, 600L, false);
+  }
```
CRZbulabula, line 116: Please add a backward-compatibility deserialization test: construct an "old format" byte stream by hand — `[procedureType:short][super state...][length:int][plan bytes]` with no trailing 17 bytes — feed it through `deserialize`, and assert `previousTTLStateCaptured == false`, `previousTTL == Long.MIN_VALUE`, `previousDatabaseWildcardTTL == Long.MIN_VALUE`. The current round-trip test only exercises new-write/new-read; without an old-format fixture, a rolling upgrade where an in-flight pre-PR procedure is persisted and then resumed on a post-PR ConfigNode is not covered.

CRZbulabula (approved):
LGTM

PR #17737, opened 2026-05-21
Pipe: Mask sensitive attributes in sink subtask display strings

luoluoyuyu:
## Description
Use masked PipeParameters display string for logs, metrics and subtask names while keeping unmasked sorted string for internal lifecycle map keys. Also treat scp.password as a sensitive parameter.

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/iotdb-api/pipe-api/src/main/java/org/apache/iotdb/pipe/api/customizer/parameter/PipeParameters.java
+++ b/iotdb-api/pipe-api/src/main/java/org/apache/iotdb/pipe/api/customizer/parameter/PipeParameters.java
@@ -429,6 +429,7 @@ public class PipeParameters {
 
     static {
       KEYS.add("ssl.trust-store-pwd");
+      KEYS.add("scp.password");
       KEYS.add("password");
     }
 
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/subtask/sink/PipeSinkSubtaskManager.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/subtask/sink/PipeSinkSubtaskManager.java
@@ -91,6 +91,7 @@ public class PipeSinkSubtaskManager {
     final int sinkNum;
     boolean realTimeFirst = false;
     String attributeSortedString = generateAttributeSortedString(pipeSinkParameters);
+    final String attributeDisplayString = generateAttributeDisplayString(pipeSinkParameters);
     if (isDataRegionSink) {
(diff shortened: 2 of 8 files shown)


Caideyipi:
Medium: iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/subtask/sink/PipeSinkSubtaskManager.java:123 now writes
    the masked display string into the sink runtime environment and subtask ID/tag path, but iotdb-core/datanode/src/main/java/org/
    apache/iotdb/db/pipe/metric/sink/PipeDataRegionSinkMetrics.java:200 still keys compressionTimerMap only by
    sink.getAttributeSortedString(), and the sinks look timers up only by that same string (iotdb-core/datanode/src/main/java/org/apache/
    iotdb/db/pipe/sink/protocol/thrift/async/IoTDBDataRegionAsyncSink.java:512, iotdb-core/datanode/src/main/java/org/apache/iotdb/db/
    pipe/sink/protocol/thrift/sync/IoTDBDataRegionSyncSink.java:596, iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/sink/
    protocol/airgap/IoTDBDataRegionAirGapSink.java:604). That means two active sinks that differ only in masked fields like password,
    scp.password, or ssl.trust-store-pwd will create separate subtasks but share one compression-timer entry; when either subtask is
    removed, iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/metric/sink/PipeDataRegionSinkMetrics.java:389 removes that
    shared key. Result: PIPE_COMPRESSION_TIME attribution becomes wrong, and a live sink can lose its timer mapping.
  - Low: iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/subtask/sink/PipeSinkSubtaskManager.java:200 still returns
    the raw canonical attribute string, and its failure paths still append that raw string directly into exceptions (iotdb-core/datanode/
    src/main/java/org/apache/iotdb/db/pipe/agent/task/subtask/sink/PipeSinkSubtaskManager.java:209, iotdb-core/datanode/src/main/java/
    org/apache/iotdb/db/pipe/agent/task/subtask/sink/PipeSinkSubtaskManager.java:237, iotdb-core/datanode/src/main/java/org/apache/iotdb/
    db/pipe/agent/task/subtask/sink/PipeSinkSubtaskManager.java:261). So if those exceptions are logged, the secret is still exposed on
    error paths.

PR #17738, opened 2026-05-21
Add region group stats to table model database details

zerolbsony:
Expose min/max schema and data RegionGroup numbers in table-model SHOW DATABASES DETAILS and information_schema.databases, keeping the metadata view aligned with tree-model database detail output.
image


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/schema/table/InformationSchema.java
```diff
@@ -106,9 +106,21 @@ public class InformationSchema {
     databaseTable.addColumnSchema(
         new AttributeColumnSchema(
             ColumnHeaderConstant.SCHEMA_REGION_GROUP_NUM_TABLE_MODEL, TSDataType.INT32));
+    databaseTable.addColumnSchema(
+        new AttributeColumnSchema(
+            ColumnHeaderConstant.MIN_SCHEMA_REGION_GROUP_NUM_TABLE_MODEL, TSDataType.INT32));
```
CRZbulabula: **Breaking change risk:** Inserting `min_schema_region_group_num` here (and the three other new columns at lines 114, 120, 123) shifts the ordinal positions of `data_region_group_num` and any column after it. Any client doing positional access on `SELECT * FROM information_schema.databases` (e.g. `resultSet.getInt(7)` to read what was `data_region_group_num`) will silently read `min_schema_region_group_num` instead.

If positional stability is important, consider appending the four new columns at the end of the table definition. Otherwise, please flag this schema change in the PR description / release notes so downstream consumers can adjust.

integration-test/src/test/java/org/apache/iotdb/relational/it/schema/IoTDBDatabaseIT.java
```diff
@@ -138,7 +138,9 @@ public void testManageDatabase() {
       }
 
       final int[] schemaRegionGroupNum = new int[] {0};
+      final int[] minSchemaRegionGroupNum = new int[] {1};
```
CRZbulabula: The literal `1` here (and `2` on line 143) come from `ConfigNodeConfig.defaultSchemaRegionGroupNumPerDatabase` / `defaultDataRegionGroupNumPerDatabase`, but that's not obvious from the test. Please add a brief comment like `// matches the cluster defaults from ConfigNodeConfig (1 / 2)` so future maintainers don't have to chase down where these magic numbers come from. Without this, the test would silently break if anyone tweaks the defaults via `MppCommonConfig`.

integration-test/src/test/java/org/apache/iotdb/relational/it/schema/IoTDBDatabaseIT.java
```diff
@@ -636,13 +646,37 @@ public void testInformationSchema() throws SQLException {
...
-      TestUtils.assertResultSetEqual(
-          statement.executeQuery("select * from databases"),
-          "database,ttl(ms),schema_replication_factor,data_replication_factor,time_partition_interval,schema_region_group_num,data_region_group_num,",
-          new HashSet<>(
-              Arrays.asList(
-                  "information_schema,INF,null,null,null,null,null,",
-                  "test,INF,1,1,604800000,0,0,")));
+      try (final ResultSet resultSet = statement.executeQuery("select * from databases")) {
+        final ResultSetMetaData metaData = resultSet.getMetaData();
+        assertEquals(11, metaData.getColumnCount());
```
CRZbulabula: This block only spot-checks the names of 4 new columns (at indexes 7, 8, 10, 11). If a future PR reorders any of the unchanged columns (1-6, 9), this test will not catch it. Consider iterating over `getSchemaTables().get(DATABASES).getColumnList()` (similar to how `testManageDatabase` at lines 147-150 verifies the full header via `showDBDetailsColumnHeaders`) to cover all columns.

integration-test/src/test/java/org/apache/iotdb/relational/it/schema/IoTDBDatabaseIT.java
```diff
@@ -636,13 +646,37 @@ public void testInformationSchema() throws SQLException {
...
+        assertEquals("min_schema_region_group_num", metaData.getColumnName(7));
+        assertEquals("max_schema_region_group_num", metaData.getColumnName(8));
+        assertEquals("min_data_region_group_num", metaData.getColumnName(10));
+        assertEquals("max_data_region_group_num", metaData.getColumnName(11));
+
+        int cnt = 0;
+        while (resultSet.next()) {
+          if ("information_schema".equals(resultSet.getString(1))) {
+            for (int columnIndex = 3; columnIndex <= 11; columnIndex++) {
+              Assert.assertNull(resultSet.getObject(columnIndex));
```
CRZbulabula: Style nit: this file already statically imports `assertEquals`, `assertTrue`, `assertFalse`, and `fail` (lines 48-51). For consistency, please add `import static org.junit.Assert.assertNull;` and use `assertNull(resultSet.getObject(columnIndex))` directly instead of the qualified `Assert.assertNull(...)`.

CRZbulabula (commented):
Functional changes look correct — the column ordering is consistent across the three writers (`ShowDBTask`, `InformationSchemaContentSupplierFactory`, schema definition in `InformationSchema`), and the underlying getters on `TDatabaseInfo` already exist (used by tree-model `ShowDatabaseStatement`). Left a few inline comments — the main one to consider is the implicit breaking change on `information_schema.databases` column ordering.

CRZbulabula (approved):
LGTM!

PR #17748, opened 2026-05-22
Fix pipe drop event discard with restart-aware committer keys

Caideyipi:
## Description
This PR fixes pipe event discard logic when dropping a pipe task by using `CommitterKey` instead of only `(pipeName,
  creationTime, regionId)`.

  Previously, queued/retry/batched events were matched only by pipe name, creation time, and region id, which could incorrectly
  discard events from a restarted pipe task. This change propagates the full committer key through pending queues, sink
  subtasks, batch builders, and sink implementations so discard checks can distinguish task restart generations.

  ## Changes

  - Use `CommitterKey` for dropped pipe task tracking and event matching.
  - Propagate restart-aware discard APIs across pipe sink subtasks and connectors.
  - Update async, sync, airgap, and websocket sinks to discard only matching pipe task events.
  - Keep backward-compatible discard methods with wildcard restart time.

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/connection/PipeEventCollector.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/connection/PipeEventCollector.java
@@ -238,7 +238,7 @@ public class PipeEventCollector implements EventCollector {
       enrichedEvent.setRebootTimes(PipeDataNodeAgent.runtime().getRebootTimes());
 
       if (enrichedEvent.getPipeName() != null
-          && pendingQueue.isPipeDropped(enrichedEvent.getPipeName(), creationTime, regionId)) {
+          && pendingQueue.isEventFromDroppedPipe(enrichedEvent)) {
         enrichedEvent.clearReferenceCount(PipeEventCollector.class.getName());
         return;
       }
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/connection/PipeEventCollector.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/connection/PipeEventCollector.java
@@ -238,7 +238,10 @@ public class PipeEventCollector implements EventCollector {
       enrichedEvent.setRebootTimes(PipeDataNodeAgent.runtime().getRebootTimes());
 
       if (enrichedEvent.getPipeName() != null
-          && pendingQueue.isEventFromDroppedPipe(enrichedEvent)) {
+          && (pendingQueue.isEventFromDroppedPipe(enrichedEvent)
+              || (enrichedEvent.getCommitterKey() == null
+                  && pendingQueue.isPipeDropped(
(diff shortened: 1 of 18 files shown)


jt2594838 (approved):

luoluoyuyu (commented):
## Review 总结

用 `CommitterKey`（含 `restartTimes`）精确匹配 drop 的 pipe 实例，修复「DROP 后同名 pipe 重启/重建时误丢新 pipe 事件」的问题，设计合理。

建议合入前确认行内 2 点语义并补 IT。

luoluoyuyu (commented):
## Review summary

Uses CommitterKey (including restartTimes) when discarding events on pipe drop so a new pipe instance with the same name is not affected. discardEventsOfPipe and isEventFromPipe are updated consistently across collector, queues, and sink subtask.

Please clarify the null committerKey fallback inline. Otherwise looks good.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/connection/PipeEventCollector.java
```diff
@@ -238,7 +238,10 @@ private void collectEvent(final Event event) {
       enrichedEvent.setRebootTimes(PipeDataNodeAgent.runtime().getRebootTimes());
 
       if (enrichedEvent.getPipeName() != null
-          && pendingQueue.isPipeDropped(enrichedEvent.getPipeName(), creationTime, regionId)) {
+          && (pendingQueue.isEventFromDroppedPipe(enrichedEvent)
```
luoluoyuyu, line 241: Drop logic uses isEventFromDroppedPipe when possible, and falls back to isPipeDropped only when committerKey is null. Please document when committerKey can still be null at collection time; if that window is wide, the fallback could discard events for a recreated pipe with the same name and creationTime.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/subtask/sink/PipeSinkSubtask.java
```diff
@@ -252,11 +248,18 @@ public void discardEventsOfPipe(
...
+      ((PipeConnectorWithEventDiscard) outputPipeSink).discardEventsOfPipe(committerKey);
     }
   }
 
+  private static boolean isEventFromPipe(
+      final EnrichedEvent event, final CommitterKey committerKey) {
+    return committerKey.getPipeName().equals(event.getPipeName())
+        && committerKey.getCreationTime() == event.getCreationTime()
+        && committerKey.getRegionId() == event.getRegionId()
+        && (committerKey.getRestartTimes() < 0 || committerKey.equals(event.getCommitterKey()));
```
luoluoyuyu, line 260: restartTimes < 0 skips restart comparison for drop-all-restarts behavior. Consider a named constant (e.g. DROP_ALL_RESTARTS) instead of a magic -1, and add a test: drop pipe, recreate same name, verify new events are not discarded.

PR #17649, opened 2026-05-12
Add explicit read-only permissions to CI workflows

arpitjain099:
## Summary
- Add explicit workflow `permissions` blocks to 16 CI workflows that previously relied on default token scopes.
- Set least-privilege `contents: read` for these test/build/check workflows.

## Why
These workflows are CI validation jobs and only need repository read access for checkout and execution. Explicit permissions harden `GITHUB_TOKEN` usage and make required scope clear.

PR #17753, opened 2026-05-25
ci: declare workflow-level `contents: read` on compile-check and todos-check

arpitjain099:
Both workflows are PR-time checks (compile sanity, todo comment validation). No GitHub API writes from the workflows. Workflow-level `contents: read` is the right cap.

Same post-CVE-2025-30066 (`tj-actions/changed-files`) hardening pattern. YAML validated locally.


arpitjain099:
This duplicates #17649 from earlier in the batch. Closing in favor of that one, sorry for the dup.

PR #17756, opened 2026-05-25
Fix flaky pipe auto split IT

Caideyipi:
### Motivation

IoTDBPipeAutoSplitIT assumes that auto split always leaves both history and realtime pipes visible immediately after creation. In practice, the history snapshot pipe can finish and be auto-dropped very quickly when there is no historical data, so SHOW PIPE may only return the realtime pipe and the test fails with expected 2 but was 1.

### Modifications

- Relax the auto split assertion to require the realtime pipe and reject the original unsplit pipe.
- Allow the history pipe to be absent after automatic completion/drop.
- Keep validating the history pipe's sink option when that history pipe is still visible.

### Tests

- Passed: git diff --check -- integration-test/src/test/java/org/apache/iotdb/pipe/it/dual/treemodel/auto/basic/IoTDBPipeAutoSplitIT.java
- Not run successfully locally: IoTDBPipeAutoSplitIT#testSingleEnv. Maven/JVM failed during compilation on this machine due Windows page file/native memory exhaustion.

### DIFF
--- a/integration-test/src/test/java/org/apache/iotdb/pipe/it/dual/treemodel/auto/basic/IoTDBPipeAutoSplitIT.java
+++ b/integration-test/src/test/java/org/apache/iotdb/pipe/it/dual/treemodel/auto/basic/IoTDBPipeAutoSplitIT.java
@@ -99,12 +99,7 @@ public class IoTDBPipeAutoSplitIT extends AbstractPipeDualTreeModelAutoIT {
       final List<TShowPipeInfo> showPipeResult =
           client.showPipe(new TShowPipeReq().setUserName(SessionConfig.DEFAULT_USER)).pipeInfoList;
       showPipeResult.removeIf(i -> i.getId().startsWith("__consensus"));
-      Assert.assertEquals(2, showPipeResult.size());
-      Assert.assertTrue(
-          (Objects.equals(showPipeResult.get(0).id, "a2b_history")
-                  && Objects.equals(showPipeResult.get(1).id, "a2b_realtime"))
-              || (Objects.equals(showPipeResult.get(1).id, "a2b_history")
-                  && Objects.equals(showPipeResult.get(0).id, "a2b_realtime")));
+      assertAutoSplitResult(showPipeResult, "a2b");
     }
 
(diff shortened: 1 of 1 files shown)


luoluoyuyu (approved):
## Review 总结

IT 修复思路正确：history pipe 在 snapshot 传完后可能被 auto-drop，原先用固定 `size==2` + 顺序断言过于脆弱。`assertAutoSplitResult` 改为只要求存在 `_realtime`、不存在未拆分原名、且列表中仅有 `_history`/`_realtime` 子 pipe，更贴近真实行为。

**建议合入**（见行内 1 条 nit）。

integration-test/src/test/java/org/apache/iotdb/pipe/it/dual/treemodel/auto/basic/IoTDBPipeAutoSplitIT.java
```diff
@@ -161,4 +157,18 @@ public void testSingleEnv() throws Exception {
         "Time,root.test.device.field,",
         Collections.singleton("1,2.0,"));
   }
+
+  private void assertAutoSplitResult(
+      final List<TShowPipeInfo> showPipeResult, final String pipeName) {
+    // The history pipe may have already been auto-dropped after snapshot transfer completes.
+    Assert.assertTrue(
+        showPipeResult.stream().anyMatch(i -> Objects.equals(i.id, pipeName + "_realtime")));
+    Assert.assertFalse(showPipeResult.stream().anyMatch(i -> Objects.equals(i.id, pipeName)));
```
luoluoyuyu, line 166: nit: Assert.assertFalse(showPipeResult.stream().anyMatch(...)) can be written as assertTrue(showPipeResult.stream().noneMatch(i -> Objects.equals(i.id, pipeName))) for readability. Not blocking.

luoluoyuyu (approved):
## Review summary

Replaces brittle assertions (exactly two pipes in a fixed order) with assertAutoSplitResult, which matches auto-split behavior when the history pipe may already be dropped after snapshot transfer.

Looks good to merge.

PR #17765, opened 2026-05-25
Respect pre-set env vars and add cgroup memory detection in env scripts

MileaRobertStefan:
## Description

Adds cgroup v1/v2 memory detection.

This PR has:
- [X] been self-reviewed.
    - [X] concurrent read
    - [X] concurrent write
    - [X] concurrent read and write 
- [X] added documentation for new or modified features or behaviors.
- [X] added Javadocs for most classes and all non-trivial methods. 
- [X] added comments explaining the "why" and the intent of the code wherever would not be obvious 
  for an unfamiliar reader.
- [X] added unit tests or modified existing tests to cover new code paths, ensuring the threshold 
  for code coverage.
- [X] been tested in a test IoTDB cluster.

<hr>

### DIFF
--- a/scripts/conf/confignode-env.sh
+++ b/scripts/conf/confignode-env.sh
@@ -19,7 +19,8 @@
 #
 
 # You can set ConfigNode memory size, example '2G' or '2048M'
-MEMORY_SIZE=
+# If the MEMORY_SIZE environment variable is already set, its value will be used.
+MEMORY_SIZE=${MEMORY_SIZE:-}
 
 # You can put your env variable here
 # export JAVA_HOME=$JAVA_HOME
@@ -68,6 +69,24 @@ calculate_memory_sizes()
     case "`uname`" in
         Linux)
             system_memory_in_mb=`free -m| sed -n '2p' | awk '{print $2}'`
+            # When running in a container/pod, use cgroup memory limit instead of host memory
+            if [ -f /sys/fs/cgroup/memory.max ]; then
+                # cgroup v2
+                cgroup_mem=`cat /sys/fs/cgroup/memory.max`
+                if [ "$cgroup_mem" != "max" ]; then
+                    cgroup_mem_in_mb=`expr $cgroup_mem / 1024 / 1024`
+                    if [ "$cgroup_mem_in_mb" -lt "$system_memory_in_mb" ]; then
+                        system_memory_in_mb=$cgroup_mem_in_mb
+                    fi
+                fi
+            elif [ -f /sys/fs/cgroup/memory/memory.limit_in_bytes ]; then
+                # cgroup v1
+                cgroup_mem=`cat /sys/fs/cgroup/memory/memory.limit_in_bytes`
+                cgroup_mem_in_mb=`expr $cgroup_mem / 1024 / 1024`
+                if [ "$cgroup_mem_in_mb" -lt "$system_memory_in_mb" ]; then
+                    system_memory_in_mb=$cgroup_mem_in_mb
+                fi
+            fi
             system_cpu_cores=`egrep -c 'processor([[:space:]]+):.*' /proc/cpuinfo`
         ;;
         FreeBSD)
--- a/scripts/conf/confignode-env.sh
+++ b/scripts/conf/confignode-env.sh
@@ -64,109 +64,6 @@ esac
 # whether we allow enable heap dump files
 IOTDB_ALLOW_HEAP_DUMP="true"
 
-calculate_memory_sizes()
-{
-    case "`uname`" in
-        Linux)
-            system_memory_in_mb=`free -m| sed -n '2p' | awk '{print $2}'`
-            # When running in a container/pod, use cgroup memory limit instead of host memory
-            if [ -f /sys/fs/cgroup/memory.max ]; then
-                # cgroup v2
-                cgroup_mem=`cat /sys/fs/cgroup/memory.max`
-                if [ "$cgroup_mem" != "max" ]; then
-                    cgroup_mem_in_mb=`expr $cgroup_mem / 1024 / 1024`
-                    if [ "$cgroup_mem_in_mb" -lt "$system_memory_in_mb" ]; then
-                        system_memory_in_mb=$cgroup_mem_in_mb
-                    fi
-                fi
-            elif [ -f /sys/fs/cgroup/memory/memory.limit_in_bytes ]; then
-                # cgroup v1
-                cgroup_mem=`cat /sys/fs/cgroup/memory/memory.limit_in_bytes`
-                cgroup_mem_in_mb=`expr $cgroup_mem / 1024 / 1024`
-                if [ "$cgroup_mem_in_mb" -lt "$system_memory_in_mb" ]; then
-                    system_memory_in_mb=$cgroup_mem_in_mb
-                fi
-            fi
-            system_cpu_cores=`egrep -c 'processor([[:space:]]+):.*' /proc/cpuinfo`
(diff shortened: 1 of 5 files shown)


scripts/conf/confignode-env.sh
```diff
@@ -68,6 +69,24 @@ calculate_memory_sizes()
     case "`uname`" in
         Linux)
             system_memory_in_mb=`free -m| sed -n '2p' | awk '{print $2}'`
+            # When running in a container/pod, use cgroup memory limit instead of host memory
+            if [ -f /sys/fs/cgroup/memory.max ]; then
```
CRZbulabula: Scope clarification — suggest expanding this code comment. The check reads the *root* cgroup, which is what containers see (Docker / Kubernetes bind-mount the container's own cgroup at `/sys/fs/cgroup/...`). For a cgroup-restricted process running on a bare-metal host (i.e. not in a container, but in a sub-cgroup with a memory limit), this won't detect the limit because `/proc/self/cgroup` is not consulted. That's a reasonable trade-off, but please make it explicit so future readers don't expect it to work on a host. For example:

```sh
# When running in a container/pod, use cgroup memory limit instead of host memory.
# This reads the root cgroup, which is what containers see; it does NOT walk
# /proc/self/cgroup, so cgroup-restricted processes on bare metal are not covered.
```

scripts/conf/confignode-env.sh
```diff
@@ -68,6 +69,24 @@ calculate_memory_sizes()
...
+                if [ "$cgroup_mem" != "max" ]; then
+                    cgroup_mem_in_mb=`expr $cgroup_mem / 1024 / 1024`
+                    if [ "$cgroup_mem_in_mb" -lt "$system_memory_in_mb" ]; then
+                        system_memory_in_mb=$cgroup_mem_in_mb
+                    fi
+                fi
+            elif [ -f /sys/fs/cgroup/memory/memory.limit_in_bytes ]; then
+                # cgroup v1
+                cgroup_mem=`cat /sys/fs/cgroup/memory/memory.limit_in_bytes`
+                cgroup_mem_in_mb=`expr $cgroup_mem / 1024 / 1024`
```
CRZbulabula: Unlike cgroup v2 (which has the explicit `"max"` string sentinel handled on line 77), cgroup v1 represents *no limit* with a kernel sentinel integer (typically `PAGE_COUNTER_MAX`, e.g. `9223372036854771712`). This branch doesn't filter it; instead the downstream `-lt` comparison on line 87 rejects it because the converted value is enormously larger than `system_memory_in_mb`. The logic is correct under bash, but the reasoning is non-obvious. Suggest a one-line code comment, e.g.:

```sh
# v1 has no "max" sentinel — when unset, this file holds a huge integer (~PAGE_COUNTER_MAX);
# the -lt check below naturally rejects it, so no explicit guard is needed.
```

Side note: `expr` overflow on a 32-bit-only `expr` could be a theoretical concern, but the shebang is `#!/bin/bash` so this is fine in practice.

scripts/conf/datanode-env.sh
```diff
@@ -69,6 +70,24 @@ calculate_memory_sizes()
     case "`uname`" in
         Linux)
             system_memory_in_mb=`free -m| sed -n '2p' | awk '{print $2}'`
+            # When running in a container/pod, use cgroup memory limit instead of host memory
+            if [ -f /sys/fs/cgroup/memory.max ]; then
```
CRZbulabula: Mirror of the suggestion on `confignode-env.sh` — please add the same scope-clarification comment here (root cgroup only; doesn't walk `/proc/self/cgroup`) so the two files stay in sync. Ideally these would share a helper in `scripts/conf/iotdb-common.sh`.

scripts/conf/datanode-env.sh
```diff
@@ -69,6 +70,24 @@ calculate_memory_sizes()
...
+                if [ "$cgroup_mem" != "max" ]; then
+                    cgroup_mem_in_mb=`expr $cgroup_mem / 1024 / 1024`
+                    if [ "$cgroup_mem_in_mb" -lt "$system_memory_in_mb" ]; then
+                        system_memory_in_mb=$cgroup_mem_in_mb
+                    fi
+                fi
+            elif [ -f /sys/fs/cgroup/memory/memory.limit_in_bytes ]; then
+                # cgroup v1
+                cgroup_mem=`cat /sys/fs/cgroup/memory/memory.limit_in_bytes`
+                cgroup_mem_in_mb=`expr $cgroup_mem / 1024 / 1024`
```
CRZbulabula: Mirror of the suggestion on `confignode-env.sh` line 85 — please add a code comment noting that the cgroup v1 "no limit" sentinel (`PAGE_COUNTER_MAX`) is rejected by the `-lt` check rather than being filtered explicitly. Otherwise it's not obvious why there's no parallel to the v2 `!= "max"` guard.

scripts/conf/windows/confignode-env.bat
```diff
@@ -20,7 +20,8 @@
 @echo off
 
 @REM You can set datanode memory size, example '2G' or '2048M'
```
CRZbulabula: Drive-by nit (pre-existing, but you're touching the next line anyway): the comment says `datanode memory size` but this is the **confignode** env file. Free win to fix while here.

CRZbulabula (commented):
Thanks for tackling this — JVM auto-sizing in containers reading host memory is a real and well-known footgun, and the cgroup v1/v2 dual handling here is the right shape. A few inline suggestions below; nothing blocking. Two file-level notes:

1. **No automated tests.** Cgroup-gated shell behavior is awkward to unit-test, but a small harness that parameterizes the cgroup mount point (default `/sys/fs/cgroup`) and runs the function against fixture files would catch regressions cheaply. Acceptable to skip given the PR description confirms manual cluster verification.
2. **Duplication.** The same ~17-line block is copy-pasted across `confignode-env.sh` and `datanode-env.sh`. The existing `calculate_memory_sizes()` is already duplicated in both files, so this is consistent with the current style — but a follow-up that sources a shared helper from `scripts/conf/iotdb-common.sh` would reduce drift.

CRZbulabula:
**The two failing `unit-test ... datanode` jobs are a stale-base false positive, not caused by this PR. A rebase will clear them.**

This PR only touches shell/bat scripts (no Java), but the two failures are in a Java pipe test — `TsFileInsertionEventParserTest#testScanParserSkipsUnnecessaryBitMaps` and `#testTableParserSkipsUnnecessaryBitMaps` — so the change cannot be the cause.

Root cause: the failing workflow run (`26503672419`) was **created on 2026-05-27**. Those two tests were introduced by #17770 and were broken on `master` at that time; they were only fixed afterwards by #17779 (2026-05-28), #17807 (2026-06-03), and #17844 (2026-06-04) — all of which landed *after* this run's merge base. Clicking "Re-run failed jobs" reuses the original 2026-05-27 merge commit and does **not** pull in newer `master`, so re-running will keep failing with the identical result no matter how many times it's retried. Recent unrelated PRs (e.g. #17937 / #17935 / #17926) pass these same datanode unit-test jobs on current `master`, confirming the tests are green now.

**Fix:** rebase this branch onto the latest `master` and force-push (or use the "Update branch" button). That generates a fresh merge base containing the fixes above, and the datanode unit-test jobs should go green. A plain "Re-run" will not work.

MileaRobertStefan:
@CRZbulabula
this was an adventure!

Can we rerun the pipeline now?

CRZbulabula (approved):
LGTM!

PR #17769, opened 2026-05-26
Optimize aligned TVList with lazy column allocation and runtime null bitmap.

luoluoyuyu:
## Description

Allocate value arrays on write only, use per-chunk BitMaps for explicit nulls, and treat unmaterialized columns as null. Adapt query iterators, flush encoding, and memtable memory estimation for sparse aligned writes.

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/memtable/AlignedWritableMemChunk.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/memtable/AlignedWritableMemChunk.java
@@ -939,8 +939,7 @@ public class AlignedWritableMemChunk extends AbstractWritableMemChunk {
     return memCostIncrement;
   }
 
-  private static boolean columnHasNonNullInRange(
-      Object column, BitMap bitMap, int start, int end) {
+  private static boolean columnHasNonNullInRange(Object column, BitMap bitMap, int start, int end) {
     if (bitMap == null) {
       return true;
     }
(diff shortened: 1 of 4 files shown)


Caideyipi:
1. AlignedTVList.java:1430

  (https://github.com/apache/iotdb/blob/ff1190cd414115ebf695a9f1ec11913db22c5dcc/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/datastructure/AlignedTVList.java#L1430)
     serializedSize() still calls getBinaryByValueIndex() unconditionally for TEXT/BLOB/STRING/OBJECT columns. After this PR, a whole value array can be null to
     represent an unwritten sparse aligned column. In that case WAL size calculation can hit an NPE before serializeToWAL() gets to its valueArray == null
     handling.
  2. TsFileProcessor.java:817

  (https://github.com/apache/iotdb/blob/ff1190cd414115ebf695a9f1ec11913db22c5dcc/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/memtable/TsFileProcessor.java#L817)
     / AlignedWritableMemChunk.java:905

  (https://github.com/apache/iotdb/blob/ff1190cd414115ebf695a9f1ec11913db22c5dcc/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/memtable/AlignedWritableMemChunk.java#L905)
     For an existing aligned memchunk, when an insert expands the TVList into a new chunk, the memory estimate adds only alignedTvListArrayMemCost(). The extra
     value-array estimate checks only whether the current last chunk is unallocated. If the previous last chunk already had a value array, but the new chunk
     will allocate value arrays for this insert, those value arrays are not counted. This can underestimate memtable memory usage and weaken write memory
     control.

PR #17770, opened 2026-05-26
[Pipe] Optimize memory usage

Caideyipi:
## Description
- Compact empty Pipe tablet bitmaps and avoid carrying bitmap arrays when no null values are marked.
- Reuse local string intern pools while deserializing receiver-side raw/batch tablet requests, reducing duplicate device/table/measurement/database strings.
- Reuse the same local string pool while splitting TsFile events into tablets, especially for repeated devices and measurement names.
- Add coverage for bitmap compaction and repeated measurement-name deduplication.

## Tests
- git diff --check
- mvn -pl iotdb-core/datanode -DskipTests spotless:apply
- mvn -pl iotdb-core/datanode -DskipTests -Dcheckstyle.skip=true compile (fails locally on unrelated generated/parser sources, e.g. CountDatabasesStatementContext and generated fill/mode classes)

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/event/common/tablet/PipeTabletUtils.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/event/common/tablet/PipeTabletUtils.java
@@ -195,6 +195,17 @@ public final class PipeTabletUtils {
       default:
         throw new UnSupportedDataTypeException(DataNodePipeMessages.UNSUPPORTED + dataType);
     }
+    unmarkNullValue(tablet, rowIndex, columnIndex);
+  }
+
+  private static void unmarkNullValue(
+      final Tablet tablet, final int rowIndex, final int columnIndex) {
+    final BitMap[] bitMaps = tablet.getBitMaps();
+    if (Objects.nonNull(bitMaps)
+        && columnIndex < bitMaps.length
+        && Objects.nonNull(bitMaps[columnIndex])) {
+      bitMaps[columnIndex].unmark(rowIndex);
+    }
   }
 
   private static BitMap[] ensureBitMaps(final Tablet tablet, final int minColumnCount) {
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/event/common/tablet/PipeTabletUtils.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/event/common/tablet/PipeTabletUtils.java
@@ -130,15 +130,6 @@ public final class PipeTabletUtils {
     return Objects.nonNull(tabletStringInternPool) ? tabletStringInternPool.intern(tablet) : tablet;
   }
 
-  /**
-   * Initializes the bitmap array shell without allocating per-column bitmaps. Null values must be
-   * recorded through {@link #markNullValue(Tablet, int, int)} so the column bitmap is allocated
-   * lazily.
-   */
-  public static void initEmptyBitMaps(final Tablet tablet) {
-    tablet.setBitMaps(new BitMap[getColumnCount(tablet)]);
-  }
-
   public static void compactBitMaps(final Tablet tablet) {
     if (Objects.isNull(tablet)) {
       return;
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/event/common/tablet/PipeTabletUtils.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/event/common/tablet/PipeTabletUtils.java
@@ -1,0 +1,248 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.iotdb.db.pipe.event.common.tablet;
+
+import org.apache.iotdb.db.i18n.DataNodePipeMessages;
+
+import org.apache.tsfile.common.conf.TSFileConfig;
+import org.apache.tsfile.enums.TSDataType;
+import org.apache.tsfile.utils.Binary;
+import org.apache.tsfile.utils.BitMap;
+import org.apache.tsfile.write.UnSupportedDataTypeException;
+import org.apache.tsfile.write.record.Tablet;
+import org.apache.tsfile.write.schema.IMeasurementSchema;
+import org.apache.tsfile.write.schema.MeasurementSchema;
+
+import java.time.LocalDate;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+public final class PipeTabletUtils {
+
+  private PipeTabletUtils() {}
+
+  public static final class TabletStringInternPool {
+
+    private final Map<String, String> internedStrings = new HashMap<>();
+
+    public String intern(final String value) {
+      if (Objects.isNull(value)) {
+        return null;
+      }
+
+      final String internedValue = internedStrings.get(value);
+      if (Objects.nonNull(internedValue)) {
+        return internedValue;
+      }
+
+      internedStrings.put(value, value);
+      return value;
+    }
+
+    public void intern(final String[] values) {
+      if (Objects.isNull(values)) {
+        return;
+      }
+
+      for (int i = 0; i < values.length; ++i) {
+        values[i] = intern(values[i]);
+      }
+    }
+
+    public void intern(final List<String> values) {
+      if (Objects.isNull(values)) {
+        return;
+      }
+
+      for (int i = 0; i < values.size(); ++i) {
+        values.set(i, intern(values.get(i)));
+      }
+    }
+
+    public Tablet intern(final Tablet tablet) {
+      if (Objects.isNull(tablet)) {
+        return null;
+      }
+
+      tablet.setInsertTargetName(intern(tablet.getDeviceId()));
+      internMeasurementSchemas(tablet.getSchemas());
+      return tablet;
+    }
+
+    public void internMeasurementSchemas(final List<IMeasurementSchema> schemas) {
+      if (Objects.isNull(schemas)) {
+        return;
+      }
+
+      for (final IMeasurementSchema schema : schemas) {
+        if (schema instanceof MeasurementSchema) {
+          intern((MeasurementSchema) schema);
+        }
+      }
+    }
+
+    public MeasurementSchema intern(final MeasurementSchema schema) {
+      if (Objects.isNull(schema)) {
+        return null;
+      }
+
+      schema.setMeasurementName(intern(schema.getMeasurementName()));
+      schema.setProps(intern(schema.getProps()));
+      return schema;
+    }
+
+    private Map<String, String> intern(final Map<String, String> props) {
+      if (Objects.isNull(props) || props.isEmpty()) {
+        return props;
+      }
+
+      final Map<String, String> internedProps = new HashMap<>(props.size());
+      for (final Map.Entry<String, String> entry : props.entrySet()) {
+        internedProps.put(intern(entry.getKey()), intern(entry.getValue()));
+      }
+      return internedProps;
+    }
+  }
+
+  public static Tablet internTablet(
+      final Tablet tablet, final TabletStringInternPool tabletStringInternPool) {
+    return Objects.nonNull(tabletStringInternPool) ? tabletStringInternPool.intern(tablet) : tablet;
+  }
+
+  public static void initEmptyBitMaps(final Tablet tablet) {
+    tablet.setBitMaps(new BitMap[getColumnCount(tablet)]);
+  }
+
+  public static void compactBitMaps(final Tablet tablet) {
+    if (Objects.isNull(tablet)) {
+      return;
+    }
+    tablet.setBitMaps(compactBitMaps(tablet.getBitMaps(), tablet.getRowSize()));
+  }
+
+  public static BitMap[] compactBitMaps(final BitMap[] bitMaps, final int rowCount) {
+    if (Objects.isNull(bitMaps)) {
+      return null;
+    }
+
+    boolean hasMarkedBitMap = false;
+    for (int i = 0; i < bitMaps.length; ++i) {
+      if (Objects.nonNull(bitMaps[i])
+          && bitMaps[i].isAllUnmarked(Math.min(rowCount, bitMaps[i].getSize()))) {
+        bitMaps[i] = null;
+      }
+      if (Objects.nonNull(bitMaps[i])) {
+        hasMarkedBitMap = true;
+      }
+    }
+
+    return hasMarkedBitMap ? bitMaps : null;
+  }
+
+  public static void markNullValue(final Tablet tablet, final int rowIndex, final int columnIndex) {
+    final BitMap[] bitMaps = ensureBitMaps(tablet, columnIndex + 1);
+    if (Objects.isNull(bitMaps[columnIndex])) {
+      bitMaps[columnIndex] = new BitMap(tablet.getMaxRowNumber());
+    }
+    bitMaps[columnIndex].mark(rowIndex);
+  }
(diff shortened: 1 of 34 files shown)


luoluoyuyu (commented):
## 总体

Pipe 路径上 BitMap 懒分配 + compact、字符串 intern、以及 InsertTablet 反序列化/split 侧的一致性改动方向正确，单测（TsFile 三 parser、split、Thrift）覆盖到位。

## 非阻塞建议

1. `PipeTabletUtils.compactBitMaps` 与 `InsertTabletNode.compactBitMaps` 逻辑重复，建议合并或文档说明 intentionally 双份。
2. `initEmptyBitMaps` 建议在 `PipeTabletUtils` 上补充 JavaDoc：写入 null 必须走 `markNullValue`，勿直接访问 `getBitMaps()[i]`。
3. `TabletStringInternPool` 在 `fromTPipeTransferReq` 每次 new，batch 场景可考虑请求级复用。
4. 确认集群滚动升级时旧节点对 `bitMaps==null` 的兼容性（本 PR 内 WritePlanNodeSplitTest 已覆盖 split）。
5. 两个 `fix` commit 与 PR 描述可再写清优化场景/收益。

LGTM with minor comments.

luoluoyuyu (commented):
**PipeTabletUtils / InsertTabletNode**：`compactBitMaps` 逻辑重复，建议抽到一处复用，避免后续行为不一致。

**PipeTabletUtils**：`initEmptyBitMaps` 后列 bitmap 可能为 null，写入 null 须走 `markNullValue`；请确认所有读路径已用 `tablet.isNull()` / `getBitMap()`，勿直接 `getBitMaps()[i]`。

**TabletStatementConverter**：反序列化无 null 时 `bitMaps` 由空数组改为 `null`，请确认与 WAL/写入/split 及滚动升级兼容。

**PipeTransferTabletRawReq**：`fromTPipeTransferReq` 每次 `new TabletStringInternPool()`，batch 内多 tablet 时 intern 收益有限，可考虑请求级复用。

**TabletInsertionEventParser**：`InsertRowNode` 分支 `filterValueColumnsByRowIndexList` 的 origin bitmap 改为 `null`，与旧逻辑等价但建议保留/补一条 InsertRow 单测。

luoluoyuyu (commented):
### Duplicate `compactBitMaps`

`PipeTabletUtils.compactBitMaps` (`iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/event/common/tablet/PipeTabletUtils.java` L143-161) and `InsertTabletNode.compactBitMaps` (`InsertTabletNode.java` L466-482) are the same loop. Consider reusing one implementation so split (`getSplit` L385) and pipe parsing stay in sync.

### Lazy bitmap contract

`initEmptyBitMaps` (L132-134) only allocates a `BitMap[]` of nulls; nulls are created in `markNullValue` (L168-175). Call sites that still assume `getBitMaps()[i]` is non-null need to use `tablet.isNull()` — e.g. `OpcUaNameSpace.java` was fixed; please double-check any remaining direct index access in this PR’s diff.

### Deserialize: `bitMaps == null`

In `TabletStatementConverter.deserializeStatementFromTabletFormat` (`TabletStatementConverter.java` L180-188), when the bitmap section is absent, `bitMaps` is now `null` instead of `new BitMap[schemaSize]`. `readBitMapsFromBufferWithMemory` (L336-368) also drops all-unmarked columns. This matches `InsertTabletNode.initBitmapsForSplit` (L448-464) but is a wire/WAL contract change — worth a quick audit of non-pipe readers of `InsertTabletNode.getBitMaps()`.

### String intern scope

`PipeTransferTabletRawReq.fromTPipeTransferReq` (`PipeTransferTabletRawReq.java` L160-162) creates a new `TabletStringInternPool` per request. For batch payloads, `PipeTransferTabletBatchReq` (L133) already holds one pool per batch — consider the same pattern for raw req if multiple tablets share device/measurement names.

### `TabletInsertionEventParser` InsertRow path

In `parse(InsertRowNode)` (`TabletInsertionEventParser.java` L171-173), `filterValueColumnsByRowIndexList(..., /*origin*/ null, bitMap)` replaces passing `bitMap` as both origin and output. InsertRow has no source bitmaps, so this should be equivalent; a small unit test on this constructor path would lock the behavior.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/event/common/tsfile/parser/scan/TsFileInsertionEventScanParser.java
```diff
@@ -304,8 +308,9 @@ private Tablet getNextTablet() {
       Tablet tablet = null;
 
       if (!data.hasCurrent()) {
-        tablet = new Tablet(currentDevice.toString(), currentMeasurements, 1);
-        tablet.initBitMaps();
+        tablet = new Tablet(currentDeviceString, currentMeasurements, 1);
+        PipeTabletUtils.initEmptyBitMaps(tablet);
+        PipeTabletUtils.compactBitMaps(tablet);
         return tablet;
```
jt2594838, line 312: What is the point

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/event/common/tsfile/parser/scan/TsFileInsertionEventScanParser.java
```diff
@@ -319,8 +324,8 @@ private Tablet getNextTablet() {
                 PipeMemoryWeightUtil.calculateTabletRowCountAndMemory(data);
             tablet =
                 new Tablet(
-                    currentDevice.toString(), currentMeasurements, rowCountAndMemorySize.getLeft());
-            tablet.initBitMaps();
+                    currentDeviceString, currentMeasurements, rowCountAndMemorySize.getLeft());
+            PipeTabletUtils.initEmptyBitMaps(tablet);
```
jt2594838: Why is an empty bitmap needed here, there, or anywhere,

luoluoyuyu (commented):
Additional code review (follow-up):

### 1. `PipeRowCollector` not migrated — likely NPE / inconsistent with lazy bitmaps

`PipeRowCollector.collectRow` still calls `tablet.initBitMaps()` (L80) and `tablet.getBitMaps()[i].mark(rowIndex)` (L97). Other parsers in this PR use `PipeTabletUtils.initEmptyBitMaps` + `markNullValue`. If `PipeRow` carries compacted/null `bitMaps` and this path is ever switched to lazy init without updating L97, it will NPE. Recommend aligning this class with the new utils (and `compactBitMaps` before emit).

### 2. `PipeRow.isNull` — missing bounds check

`PipeRow.java` L154-157 guards `bitMaps` and `bitMaps[columnIndex]` but not `columnIndex < bitMaps.length`. With sparse/null-padded `bitMaps` arrays from `TabletInsertionEventParser`, out-of-range column access can still throw.

### 3. `compactBitMaps` duplicated in 4 places

Same loop now exists in `PipeTabletUtils` (L143-161), `InsertTabletNode` (L466-482), and **`InsertTabletStatement`** (L903-920, also L393 on split). Drift risk is higher than noted earlier — worth centralizing.

### 4. `PipeTreeModelTsFileBuilder` aggregation may re-allocate empty bitmap slots

`tryBestToAggregateTablets` uses `PipeTabletUtils.copyBitMapsOrCreateEmpty(tablet)` (L234). When `tablet.getBitMaps()` is `null`, this allocates a `BitMap[columnCount]` of nulls per aggregated tablet. The merged `Tablet` constructor (L255-260) does not call `compactBitMaps`, so aggregation can retain O(columns × tablets) null bitmap shells and partially undo the memory win on sink TsFile build paths. Consider `compactBitMaps` on the aggregated tablet (V1/V2 builders share this).

### 5. `TabletStringInternPool.intern(MeasurementSchema)` mutates schema props

`PipeTabletUtils.java` L112-120 replaces `schema.setProps(intern(...))` with a **new** `HashMap` each time. Safe for parser-owned schemas, but if a shared `MeasurementSchema` instance is reused across events, props could be unexpectedly replaced. Worth a quick comment or copy-on-intern only when needed.

### 6. `TabletInsertionEventTableParserTabletIterator.internMeasurementName`

L478-483 calls `tabletStringInternPool.intern((MeasurementSchema) schema)` then `intern(schema.getMeasurementName())` — redundant for `MeasurementSchema`; minor, but the double call is confusing.

### 7. Tests gap: `PipeRowCollector` / TsFile builder aggregation

New tests cover parsers, split, and batch intern (`PipeDataNodeThriftRequestTest`), but not `PipeRowCollector` or multi-tablet TsFile builder aggregation with all-null columns. A small UT would catch (1) and (4).

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/event/common/tsfile/parser/scan/TsFileInsertionEventScanParser.java
```diff
@@ -412,37 +418,68 @@ private boolean putValueToColumns(final BatchData data, final Tablet tablet, fin
...
         isNeedFillTime = true;
         switch (tablet.getSchemas().get(i).getType()) {
           case BOOLEAN:
-            tablet.addValue(rowIndex, i, primitiveType.getBoolean());
+            PipeTabletUtils.putValue(
+                tablet,
+                rowIndex,
+                i,
+                tablet.getSchemas().get(i).getType(),
+                primitiveType.getBoolean());
```
jt2594838, line 432: What is the point?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/event/common/tablet/PipeTabletUtils.java
```diff
@@ -0,0 +1,256 @@
...
+    tablet.setRowSize(Math.max(tablet.getRowSize(), rowIndex + 1));
+  }
+
+  public static void putValue(
+      final Tablet tablet,
+      final int rowIndex,
+      final int columnIndex,
+      final TSDataType dataType,
+      final Object value) {
+    switch (dataType) {
```
jt2594838, line 170: What is the point of the two methods?

luoluoyuyu (commented):
### Potential bug (regression): `DefaultOperationQuota` vs `bitMaps == null`

This PR sets `InsertTabletStatement.bitMaps` to `null` when there are no nulls (`TabletStatementConverter.java` L180-188). That is fine for memtable / `InsertTabletNode` (they already null-check), but **`DefaultOperationQuota` does not**:

```java
// DefaultOperationQuota.java L120-121
for (BitMap bitMap : insertTabletStatement.getBitMaps()) {
  avgSize += bitMap.getSize();
}
```

- If `getBitMaps()` is `null` → **NPE** on the enhanced-for.
- If `getBitMaps()` is a sparse array with `null` slots (after `readBitMapsFromBufferWithMemory` / `compactBitMaps`) → **NPE** on `bitMap.getSize()`.

Same pattern at L155-157 for `InsertMultiTabletsStatement`.

**Repro path:** Pipe sink receives / deserializes a tablet batch with no null columns → statement keeps `bitMaps == null` → any insert that goes through quota estimation.

**Suggested fix:** null-check before iterating; skip `null` elements (same as `AbstractMemTable.computeTabletNullPointsNumber`).

---

Other areas reviewed (PipeRow, parsers, split, TsFile builders, `PipeRowCollector`) look OK under current code — no additional definite bugs found beyond this quota path.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/write/InsertTabletNode.java
```diff
@@ -441,6 +445,42 @@ protected BitMap[] initBitmaps(int columnSize, int rowSize) {
...
+      if (bitMaps[i] != null
+          && bitMaps[i].isAllUnmarked(Math.min(rowCount, bitMaps[i].getSize()))) {
+        bitMaps[i] = null;
+      }
+      if (bitMaps[i] != null) {
+        hasBitMap = true;
+      }
+    }
+    return hasBitMap ? bitMaps : null;
+  }
```
jt2594838: Duplicated?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/statement/crud/InsertTabletStatement.java
```diff
@@ -900,6 +900,24 @@ private Object convertColumnToTablet(
...
+          && bitMaps[i].isAllUnmarked(Math.min(rowCount, bitMaps[i].getSize()))) {
+        bitMaps[i] = null;
+      }
+      if (bitMaps[i] != null) {
+        hasBitMap = true;
+      }
+    }
+    return hasBitMap ? bitMaps : null;
+  }
+
```
jt2594838: Here too?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/event/common/tsfile/parser/scan/TsFileInsertionEventScanParser.java
Caideyipi: Removed this redundant empty-bitmap initialization in 4879c3a. Empty tablets now keep bitMaps == null; a bitmap shell is only created lazily when markNullValue actually records a null.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/event/common/tsfile/parser/scan/TsFileInsertionEventScanParser.java
Caideyipi: Agreed. The empty bitmap shell is not needed now because markNullValue creates the array and the per-column BitMap lazily. Removed these eager init calls from the TsFile parser paths in 4879c3a.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/event/common/tablet/PipeTabletUtils.java
Caideyipi: These helpers keep row filling independent from Tablet.addValue: putTimestamp is the only place that advances rowSize after a row has at least one non-null value, and putValue writes by column index while preserving DATE/Binary conversions. I also removed the obsolete initEmptyBitMaps helper and made bitmap allocation entirely lazy through markNullValue.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/event/common/tsfile/parser/scan/TsFileInsertionEventScanParser.java
Caideyipi: The point is to avoid Tablet.addValue plus eager tablet.initBitMaps. putValue writes directly to the preallocated value arrays by column index, and markNullValue allocates only the column bitmap that actually has nulls. PipeRowCollector was migrated to the same path as well.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/write/InsertTabletNode.java
Caideyipi: Fixed in c16fa23. The duplicated loop is now in BitMapUtils.compactBitMaps, and InsertTabletNode delegates to it.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/statement/crud/InsertTabletStatement.java
Caideyipi: Fixed in c16fa23 as well. InsertTabletStatement now uses the shared BitMapUtils.compactBitMaps implementation.

Caideyipi:
Handled the review items in c16fa23 and 4879c3a:

- Fixed DefaultOperationQuota for bitMaps == null and sparse bitmap arrays, including InsertMultiTabletsStatement. Added DefaultOperationQuotaTest for both cases.
- Centralized compactBitMaps into BitMapUtils and made InsertTabletNode, RelationalInsertTabletNode, InsertTabletStatement, and PipeTabletUtils delegate to the same implementation.
- Removed redundant empty bitmap initialization from TsFile parser tablet builders. markNullValue now creates the bitmap array and per-column bitmap lazily only when a null is recorded.
- Migrated PipeRowCollector to the same lazy bitmap path and compact before emitting the collected tablet.
- Added a bounds guard in PipeRow.isNull for sparse bitmap arrays.

For the string-intern scope comment: raw req carries a single tablet, so a per-raw-request pool is enough there; batch req already reuses one pool across all tablets in the batch. For rolling compatibility: null bitMaps are still represented by the existing hasBitMaps=false branch, and the quota path now skips both null arrays and null slots.

Verified:
- git diff --check
- mvn -pl iotdb-core/datanode -DskipTests spotless:apply
- mvn -pl iotdb-core/datanode -Dtest=DefaultOperationQuotaTest test
- mvn -pl iotdb-core/datanode -Dtest=WritePlanNodeSplitTest test

PR #17774, opened 2026-05-27
Fix NOT expression in where/having for TreeModel

Wei-hao-Li:


Caideyipi:
I found two functional gaps in this fix:

1. The template-optimized TreeModel path still keeps `NOT` in the WHERE predicate. The new calls to `PredicateUtils.predicateRemoveNot(...)` are added in `AnalyzeVisitor`, but multi-device `ALIGN BY DEVICE` queries using one schema template return early through `TemplatedAnalyze.canBuildPlanUseTemplate(...)`. In that path, `TemplatedAnalyze.analyzeDeviceToWhere()` still starts from `queryStatement.getWhereCondition().getPredicate()` and only simplifies it, so a query such as `SELECT s3 FROM root.sg2.** WHERE NOT (s3 > 1 AND s3 < 10) ALIGN BY DEVICE` can still carry a `LogicNotExpression` into predicate pushdown / scan filter handling and hit the same `CONTAIN_NOT_ERR_MSG` class of failure. This path needs the same `predicateRemoveNot` rewrite, and a multi-device template `ALIGN BY DEVICE` test should be added.

2. `PredicateUtils.predicateRemoveNot(...)` does not actually remove all `NOT`s for nested odd-depth cases. For example, `NOT NOT NOT (s1 > 1)` reaches `reversePredicate(child)`, and `ReversePredicateVisitor.visitLogicNotExpression()` just returns `logicNotExpression.getExpression()` without recursively removing/reversing the remaining `LogicNotExpression`. The resulting predicate can still contain `NOT`, which violates the new assumption in `PredicatePushIntoScanChecker` that `NOT` has already been removed in analyze stage. Please add coverage for nested `NOT` such as `not(not(not(...)))` and make the rewrite recursively eliminate it.

PR #17780, opened 2026-05-28
Subscription: add topic owner epoch fencing

VGalaxies:
## Summary

Add owner-epoch fencing for subscription topics so that, after an owner
failover/transfer, stale owners can no longer mutate a topic's consumption
progress.

- **Owner identity + epoch metadata.** A topic may carry `owner-id` and
  `owner-epoch`. Consumers propagate their `owner-id`/`owner-epoch`, and the
  DataNode fences stale owners on heartbeat, subscribe, poll, commit, and
  `seek` (before any position-mutating operation).
- **Globally monotonic, never-reused epochs.** A new `max-owner-epoch`
  attribute records the highest epoch ever granted and survives `clearOwner`,
  so an owner change can never reuse an epoch `<=` any epoch previously
  granted, even across a clear -> re-enable cycle.
- **Owner lease as a liveness bound.** The lease is pushed by the ConfigNode
  over a dedicated, fine-grained `SubscriptionOwnerLeaseSyncer` (independent of
  the node heartbeat) carrying a *relative remaining duration*. The absolute
  expiry is derived and held **DataNode-local at runtime only** -- it is never a
  topic attribute, never serialized, and not part of `equals`/`hashCode`. This
  avoids comparing absolute timestamps across ConfigNode/DataNode clocks. The
  ConfigNode persists only `owner-id`, `owner-epoch`, `max-owner-epoch`, and
  `owner-lease-duration-ms`.
- **Transfer drain via ConfigNode admission-wait.** On an owner transfer the
  ConfigNode stops renewing the previous owner's lease and waits
  `blockSince + lease-duration + heartbeat-interval` on its own clock before
  installing the new owner. Epoch fencing on reachable DataNodes guarantees
  correctness during the wait.

## Notable design points

- The lease is a liveness mechanism layered on top of epoch fencing, not the
  primary fence: if no lease has been established yet (e.g. right after a
  meta push), owner-id/epoch matching still fences correctly.
- No absolute lease timestamp is ever propagated between nodes.

## Tests

Unit tests (pass):

```
mvn test -pl iotdb-core/confignode -am \
  -Dtest=TopicDeSerTest,SubscriptionInfoTest \
  -Dsurefire.failIfNoSpecifiedTests=false
```

`TopicDeSerTest` covers serialization, owner transfer, cross-clear epoch-reuse
rejection, and the DataNode-local-only lease expiry behavior;
`SubscriptionInfoTest` covers epoch-rollback rejection, owner transfer via
updated attributes, the read-only relative-duration lease collector, and
`SHOW TOPICS` output.

Compilation / build validation:

```
mvn compile -pl iotdb-core/datanode,iotdb-core/confignode -am
mvn test-compile -pl integration-test -am -P with-integration-tests
```

Integration tests added but not run under MiniCluster in this change:
`IoTDBSubscriptionTopicOwnerIT` (relative-lease drain) and
`IoTDBSubscriptionTopicOwnerPartitionIT` (multi-DataNode fencing, ConfigNode
failover epoch persistence, DataNode restart).


Caideyipi:
Findings

  - P1: TopicMeta 的新尾部字段破坏旧格式兼容。deserialize(InputStream) 用 inputStream.available() 判断是否还有 owner
    字段，但 TopicMeta 在 snapshot 里是连续写入的；旧版本数据没有 owner flag，多个 topic 时这里会把下一个 topic
    的首字节当成 owner flag 消费，导致后续反序列化错位。ByteBuffer.hasRemaining() 在旧 procedure/plan
    列表里也有同类问题。见 TopicMeta.java#L260

  (https://github.com/apache/iotdb/blob/76f2a882411ba790f7db290740f0232a1e5c4022/iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/subscription/meta/topic/TopicMeta.java#L260)
    和 TopicMetaKeeper.java#L112

  (https://github.com/apache/iotdb/blob/76f2a882411ba790f7db290740f0232a1e5c4022/iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/subscription/meta/topic/TopicMetaKeeper.java#L112)。建议改成版本化或长度前缀格式，或只从
    config 属性恢复 owner，不在非自描述对象尾部追加可选字段。
  - P1: owner epoch 单调性没有在真实元数据替换路径上保证。transferOwner() 只在同一个对象上检查递增，但 AlterTopicPlan /
    metadata replace 路径直接 remove + add 新 TopicMeta，新对象从 -1 初始化，所以较小 epoch 也能覆盖当前 epoch，使旧
    owner 重新合法。见 TopicMeta.java#L135

  (https://github.com/apache/iotdb/blob/76f2a882411ba790f7db290740f0232a1e5c4022/iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/subscription/meta/topic/TopicMeta.java#L135)
    和 SubscriptionInfo.java#L340

  (https://github.com/apache/iotdb/blob/76f2a882411ba790f7db290740f0232a1e5c4022/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/subscription/SubscriptionInfo.java#L340)。建议在
    ConfigNode alter/handle meta change 时对比 existing owner epoch 并拒绝回退。
  - P2: 新增的 owner status code 没有接入客户端错误分类。1913-1916 会落到 default，变成 generic critical
    exception；heartbeat 还会把 provider 标成 unavailable，导致 stale owner 场景反复重连/重试且错误信息不稳定。见
    TSStatusCode.java#L319

  (https://github.com/apache/iotdb/blob/76f2a882411ba790f7db290740f0232a1e5c4022/iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/TSStatusCode.java#L319)、AbstractSubscriptionProvider.java#L445

  (https://github.com/apache/iotdb/blob/76f2a882411ba790f7db290740f0232a1e5c4022/iotdb-client/subscription/src/main/java/org/apache/iotdb/session/subscription/consumer/base/AbstractSubscriptionProvider.java#L445)。建议显式处理
    owner fenced/required/lease expired，映射成明确的不可重试业务异常。

VGalaxies:
Thanks @Caideyipi, addressed the three findings in b6086b2f4e.

- P1 serialization compatibility: removed the optional owner fields from the non-self-describing TopicMeta serialization tail. Owner state is now restored from topic attributes after deserialization, so sequential TopicMeta snapshot/procedure streams are not consumed out of boundary. Added a sequential deserialization regression test.
- P1 epoch rollback: added TopicMeta.validateOwnerProgression(...) and applied it on both ConfigNode alter-topic metadata replacement and DataNode topic-meta update handling, rejecting owner clears/rollbacks/stale same-epoch owner changes. Added ConfigNode rollback coverage and extended the network-partition old-SN test to verify stale topic meta cannot make the old owner valid again.
- P2 client classification: mapped 1913-1917 explicitly to SubscriptionOwnerFencedException, a SubscriptionRuntimeNonCriticalException subclass, so stale-owner/business fencing errors no longer fall into the generic critical default path.

Local verification passed:
- mvn spotless:apply -pl iotdb-client/service-rpc,iotdb-client/subscription,iotdb-core/node-commons,iotdb-core/datanode,iotdb-core/confignode -DskipTests
- mvn test -pl iotdb-client/service-rpc,iotdb-client/subscription,iotdb-core/node-commons -Dtest=TopicDeSerTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false
- mvn test -pl iotdb-client/service-rpc,iotdb-client/subscription,iotdb-core/node-commons,iotdb-core/confignode -Dtest=SubscriptionInfoTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false
- mvn test -pl iotdb-core/relational-grammar,iotdb-client/service-rpc,iotdb-client/subscription,iotdb-core/node-commons,iotdb-core/datanode -Dtest=SubscriptionReceiverV1Test -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false
- mvn test -pl iotdb-client/service-rpc,iotdb-client/subscription -Dtest=TSStatusCodeTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false
- git diff --check

The Sonar duplication check is queued again on the new commit; I will follow up if the rerun still fails.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/agent/SubscriptionTopicAgent.java
```diff
@@ -188,4 +195,73 @@ public Map<String, TopicConfig> getTopicConfigs(final Set<String> topicNames) {
...
+    acquireReadLock();
+    try {
+      if (!topicMetaKeeper.containsTopicMeta(topicName)) {
+        return RpcUtils.SUCCESS_STATUS;
+      }
+
+      final TopicMeta topicMeta = topicMetaKeeper.getTopicMeta(topicName);
+      if (!topicMeta.isOwnerFencingEnabled()) {
+        return RpcUtils.SUCCESS_STATUS;
+      }
```
jt2594838: Is it possible to merge contains and get?

jt2594838 (commented):
Add an IT to show a complete use case.

VGalaxies:
Addressed the latest review comments in 25c34c2bfe.

- Added IoTDBSubscriptionTopicOwnerIT as a complete local subscription use case: a stale owner (sn1/epoch 5) is fenced after the topic is owned by sn2/epoch 6, while the current owner can subscribe, poll data, commit, and unsubscribe.
- Kept the new IT aligned with the current subscription IT convention: it is @Ignore because SubscriptionConfig#getSubscriptionEnabled() is currently hard-coded false in this codebase, but the class compiles and is discovered under the with-integration-tests profile.
- Preserved SubscriptionOwnerFencedException through subscribe redirection so the stale-owner path remains a business fencing error.
- Merged the SubscriptionTopicAgent contains/get check into one getTopicMeta(...) null check.

Local verification:
- mvn spotless:apply -pl integration-test -P with-integration-tests -DskipTests
- mvn verify -DskipUTs -Dit.test=IoTDBSubscriptionTopicOwnerIT -DfailIfNoTests=false -Dfailsafe.failIfNoSpecifiedTests=false -pl integration-test -am -P with-integration-tests
- mvn test -pl iotdb-client/service-rpc,iotdb-client/subscription -Dtest=TSStatusCodeTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false
- mvn test -pl iotdb-core/relational-grammar,iotdb-client/service-rpc,iotdb-client/subscription,iotdb-core/node-commons,iotdb-core/datanode -Dtest=SubscriptionReceiverV1Test -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false
- git diff --check

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/agent/SubscriptionTopicAgent.java
VGalaxies: Done. I merged the contains/get path into a single getTopicMeta(...) call and handle the missing-topic case with a null check before owner-fencing validation.

jt2594838 (commented):
Need one more test to show how to update the ownership and how to show the ownership.

iotdb-client/subscription/src/main/java/org/apache/iotdb/session/subscription/ISubscriptionTableSession.java
```diff
@@ -99,6 +99,49 @@ void createTopic(final String topicName, final Properties properties)
...
+   * Alters a topic with the specified properties.
+   *
+   * @param topicName If the topic name contains single quotes, the passed parameter needs to be
+   *     enclosed in backticks.
+   * @param properties A {@link Properties} object containing the topic properties to alter.
+   * @throws IoTDBConnectionException If a connection issue occurs.
+   * @throws StatementExecutionException If the SQL statement execution fails.
+   */
+  void alterTopic(final String topicName, final Properties properties)
+      throws IoTDBConnectionException, StatementExecutionException;
```
jt2594838, line 112: Is the TopicOwner allowed to be modified by this interface?
Or can it only be modified through `alterTopicOwner`?

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/subscription/SubscriptionCoordinator.java
```diff
@@ -149,6 +151,25 @@ public TSStatus createTopic(TCreateTopicReq req) {
...
+    }
+    return status;
+  }
+
+  public TopicMeta buildAlteredTopicMeta(TAlterTopicReq req) {
+    final TopicMeta existedTopicMeta = subscriptionInfo.deepCopyTopicMeta(req.getTopicName());
+    return existedTopicMeta == null
+        ? null
+        : existedTopicMeta.deepCopyWithUpdatedAttributes(req.getTopicAttributes());
+  }
```
jt2594838, line 242: Why deep copy twice?

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/ProcedureManager.java
```diff
@@ -1733,6 +1736,35 @@ public TSStatus createTopic(TCreateTopicReq req) {
...
+          configManager
+              .getSubscriptionManager()
+              .getSubscriptionCoordinator()
+              .buildAlteredTopicMeta(req);
+      if (updatedTopicMeta == null) {
+        return new TSStatus(TSStatusCode.ALTER_TOPIC_ERROR.getStatusCode())
+            .setMessage(
+                String.format(
+                    "Failed to alter topic %s, the topic is not existed", req.getTopicName()));
+      }
```
jt2594838, line 1778: i18n

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/parser/ASTVisitor.java
```diff
@@ -4425,6 +4426,23 @@ public Statement visitCreateTopic(IoTDBSqlParser.CreateTopicContext ctx) {
...
+  @Override
+  public Statement visitAlterTopic(IoTDBSqlParser.AlterTopicContext ctx) {
+    final AlterTopicStatement alterTopicStatement = new AlterTopicStatement();
+
+    if (ctx.topicName != null) {
+      alterTopicStatement.setTopicName(parseIdentifier(ctx.topicName.getText()));
+    } else {
+      throw new SemanticException(
+          "Not support for this sql in ALTER TOPIC, please enter topicName.");
+    }
```
jt2594838, line 4448: i18n

jt2594838 (commented):
Have you waited for all leases to expire before transferring the ownership?

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/ProcedureManager.java
VGalaxies: Addressed in 82049d7b46. Moved this hard-coded ALTER TOPIC error message into `ManagerMessages` for both en and zh locale sources.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/parser/ASTVisitor.java
VGalaxies: Addressed in 82049d7b46. Moved this parser error into `DataNodeQueryMessages` for both en and zh locale sources.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/subscription/SubscriptionCoordinator.java
VGalaxies: Addressed in 82049d7b46. I added `SubscriptionInfo#deepCopyTopicMetaWithUpdatedAttributes(...)`, so the read-lock protected lookup and attribute merge now create one copied `TopicMeta` instead of deep-copying twice in `SubscriptionCoordinator`.

iotdb-client/subscription/src/main/java/org/apache/iotdb/session/subscription/ISubscriptionTableSession.java
VGalaxies: Addressed in 82049d7b46. The public session `alterTopic(topicName, properties)` now rejects `owner-id`, `owner-epoch`, and `owner-lease-expire-time-ms`; owner fields are only allowed through the internal path used by `alterTopicOwner(...)`.

VGalaxies:
Addressed the lease-transfer review comment in 82049d7b46.

Topic ownership transfer now checks the current topic owner lease before replacing owner metadata: if the current owner has an unexpired `owner-lease-expire-time-ms` and the incoming owner id is different, the transfer is rejected. Once the lease has expired, the transfer can proceed. Added `SubscriptionInfoTest` coverage for both paths.

Also addressed the inline comments in the same commit:
- public session `alterTopic(...)` rejects owner attributes; `alterTopicOwner(...)` is the session API entry for owner updates.
- removed the extra deep copy from the alter-topic metadata build path.
- moved the two hard-coded messages to en/zh i18n constants.

Local verification passed:
- `mvn spotless:apply -pl iotdb-client/subscription,iotdb-core/node-commons,iotdb-core/confignode,iotdb-core/datanode -DskipTests`
- `mvn test -pl iotdb-protocol/thrift-confignode,iotdb-client/service-rpc,iotdb-client/subscription,iotdb-core/node-commons,iotdb-core/confignode -Dtest=SubscriptionInfoTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false`
- `mvn test -pl iotdb-protocol/thrift-confignode,iotdb-core/relational-grammar,iotdb-client/service-rpc,iotdb-client/subscription,iotdb-core/node-commons,iotdb-core/datanode -Dtest=StatementGeneratorTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false`
- `mvn compile -pl iotdb-protocol/thrift-confignode,iotdb-client/service-rpc,iotdb-client/subscription,iotdb-core/relational-grammar,iotdb-core/node-commons,iotdb-core/confignode,iotdb-core/datanode -P with-zh-locale -DskipTests`
- `git diff --check`

jt2594838 (approved):

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/subscription/meta/topic/TopicMeta.java
```diff
@@ -168,14 +187,29 @@ public void transferOwner(
                   + " current owner-epoch: %s, current owner-lease-expire-time-ms: %s",
               topicName, this.ownerId, this.ownerEpoch, this.ownerLeaseExpireTimeMs));
     }
+    if (Objects.nonNull(ownerLeaseDurationMs) && ownerLeaseDurationMs <= 0) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Subscription topic owner lease duration should be positive, topic: %s,"
+                  + " owner-lease-duration-ms: %s",
+              topicName, ownerLeaseDurationMs));
```
jt2594838: i18n

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/subscription/SubscriptionCoordinator.java
```diff
@@ -163,11 +169,81 @@ public TSStatus alterTopic(TAlterTopicReq req) {
...
   }
 
+  public TSStatus renewTopicOwnerLease(TRenewTopicOwnerLeaseReq req) {
+    if (blockedOwnerLeaseRenewalTopics.contains(req.getTopicName())) {
+      return RpcUtils.getStatus(
+          TSStatusCode.SUBSCRIPTION_OWNER_FENCED,
+          String.format(
+              "Subscription: topic %s owner lease renewal is blocked by a pending owner transfer.",
+              req.getTopicName()));
+    }
```
jt2594838: i18n

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/subscription/meta/topic/TopicMeta.java
VGalaxies: Addressed in 6787927100. Moved the TopicMeta owner-fencing exception messages (owner id/epoch validation, monotonic/never-reused epoch, stale-clear, rollback, and lease-duration checks) into `SubscriptionMessages` with both en and zh locale sources. node-commons already depends on the subscription client module, so it reuses that i18n class.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/subscription/SubscriptionCoordinator.java
VGalaxies: Addressed in 6787927100. The originally flagged `renewTopicOwnerLease` message was removed during the owner-lease rework (CN-side renewal was replaced by a DataNode-local lease heartbeat). The remaining PR-introduced hard-coded message in this file, the alter-topic warn, now uses `ManagerMessages.FAILED_TO_ALTER_TOPIC_WITH_ATTRIBUTES_RESULT_STATUS` (en/zh).

integration-test/src/test/java/org/apache/iotdb/subscription/it/local/IoTDBSubscriptionTopicOwnerIT.java
```diff
@@ -150,49 +142,24 @@ public void testHeartbeatRenewsOwnerLeaseAndAlterOwnerWaitsForLeaseExpiration()
...
-        final long ownerLeaseRemainingTimeMs =
-            renewedOwnerLeaseExpireTimeMs.get() - alterStartTimeMs;
-        if (ownerLeaseRemainingTimeMs > 500L) {
-          Assert.assertTrue(alterElapsedTimeMs >= ownerLeaseRemainingTimeMs - 250L);
-        }
+        Assert.assertTrue(
+            "alterTopicOwner should block at least the lease drain duration, but only took "
+                + alterElapsedTimeMs
+                + "ms",
+            alterElapsedTimeMs >= ownerLeaseDurationMs);
```
jt2594838, line 181: The waiting time also depends on the renewal interval (heartbeat interval).

For example,
lease renewed at t0
owner altered at t1
lease expired at t2

then the waiting time will be t2 - t1 <= t2 - t0 = leaseDuration

jt2594838 (approved):

integration-test/src/test/java/org/apache/iotdb/subscription/it/local/IoTDBSubscriptionTopicOwnerIT.java
VGalaxies: Good catch — addressed in 55b95686c6 (comment clarified). You're right that the heartbeat interval matters. The ConfigNode does not wait until a DataNode's absolute lease-expire time (that would couple the two clocks); instead it stops renewal at `blockSince` and waits `leaseDuration + one heartbeat interval (H)` on its own clock. The extra H is needed because a renewal may have been in flight to a DataNode right before renewal was stopped, so that DataNode's lease could expire as late as `blockSince + H + leaseDuration`. Hence the actual wait is `L + H`, which is why `alterElapsedTimeMs >= leaseDuration` holds as a (loose) lower bound. I've updated the test comment to state the `L + H` relationship explicitly.

jt2594838 (approved):

PR #17788, opened 2026-05-29
Improve query modification loading memory control

shuwenwei:
## Summary
- Add quota pre-checking before loading modification files into the query mods cache.
- Reserve memory incrementally while building modification pattern trees.
- Fall back to path-matched scanning when cache quota or memory reservation is insufficient.
- Release partial tree quota and memory before returning fallback matched modifications.
- Extend unit coverage for cache loading, fallback scanning, final reservation failure, and sort/merge reservation adjustment.

### DIFF
--- a/iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/plan/planner/memory/MemoryReservationManager.java
+++ b/iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/plan/planner/memory/MemoryReservationManager.java
@@ -34,6 +34,14 @@ public interface MemoryReservationManager {
   void reserveMemoryImmediately();
 
   /**
+   * Reserve memory for the given size immediately without changing the accumulated pending
+   * reservation size maintained by this manager.
+   *
+   * @param size the size of memory to reserve immediately
+   */
+  void reserveMemoryImmediately(final long size);
+
+  /**
    * Release memory for the given size.
    *
    * @param size the size of memory to release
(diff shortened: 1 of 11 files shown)


Caideyipi:
Review notes:

1. `QueryModificationLoader.java:115` uses `TsFileResource.getTotalModSizeInByte()` as the quota pre-check before trying to cache mods. However, `TsFileResource.getTotalModSizeInByte()` currently counts the full shared mods file length (`TsFileResource.java:453-455`), while the actual iterator only reads shared mods from `sharedModFileOffset` (`TsFileResource.java:1515-1516`). For TsFiles using shared mods, this can make the pre-check fail just because the shared file has grown globally, even when the unread portion for this TsFile is small. That means cache loading will be bypassed and the query will repeatedly fall back to scan. Please consider using the actual readable mods size here, e.g. `exclusiveModFileLength + max(0, sharedModFileLength - sharedModFileOffset)`.

2. `FragmentInstanceContext.java:413` uses `modification -> modification.affects(deviceID)` as the fallback matcher for the device-only overload. This is broader than the cache path (`mods -> getPathModifications(mods, deviceID)`), which matches `deviceID + AlignedPath.VECTOR_PLACEHOLDER`. For example, a deletion on `root.sg.d1.s1` affects device `root.sg.d1`, but it does not match the time-column/vector-placeholder path. So cache success and fallback scan can return different results for the same API. Even if this overload is not on the hot path today, the fallback predicate should be equivalent to the pattern-tree matcher.

JackieTien97 (approved):

PR #17800, opened 2026-05-31
C++ client: add thread-safe SessionPool, enable RPC compression, and harden buffers

SpriCoder:
## Summary

Improvements to the C++ client (`iotdb-client/client-cpp`) along two axes: concurrency and safety/correctness.

### 1. Thread-safe `SessionPool` (new feature)
A single `Session` is not safe to use from multiple threads concurrently (it owns one Thrift transport/client). Previously the C++ client had no pooling, forcing callers to either serialize all requests through one session or hand-roll connection management.

- New `SessionPool` + `SessionPoolBuilder` lend each `Session` to exactly one borrower at a time and reclaim it afterwards, so many application threads can share a bounded set of connections without external locking.
- `getSession()` returns an RAII `PooledSession` handle that returns the session to the pool on scope exit and works with any `Session` method; a generic `execute([](Session&){...})` plus convenience wrappers (`insertTablet`, `insertRecord`, `executeQueryStatement`, …) cover common operations.
- New sessions are constructed **outside** the pool lock so a handshake never blocks other borrowers; `getSession()` blocks up to a configurable timeout when the pool is exhausted, and `maxSize == 0` is rejected at construction.
- Query results are returned as a `PooledSessionDataSet` that keeps the session leased until the result set is fully read, since `SessionDataSet` lazily fetches further blocks over the same connection. Sessions that raise `IoTDBConnectionException` are evicted instead of recycled.
- Integration tests cover basic borrow/insert/query, concurrent writers, and exhaustion-timeout behavior.

### 2. Safety / correctness hardening
- `Tablet::addValue` formatted out-of-range diagnostics with `sprintf` into a fixed 100-byte stack buffer; switched to `snprintf` bounded by `sizeof(buffer)` to remove the overflow risk.
- `MyStringBuffer::putOrderedByte` used `str.assign` on big-endian hosts, which overwrote the whole buffer on every numeric write and corrupted previously serialized content; changed to `str.append` to match the little-endian path.

### Note on RPC compression
An earlier commit wired the (previously ignored) `enableRPCCompression` flag through to `SessionConnection`. Honoring it makes the client negotiate the compact Thrift protocol, which the binary-only IoTDB server used by the C++ integration tests cannot speak — so it broke the pre-existing `ts_session_open_with_compression` smoke test (which had only passed because the flag was a no-op). That change is **reverted here** and will be reintroduced in a dedicated PR with matching compact-protocol test-server support. `SessionPool` keeps its compression option for forward compatibility (currently a no-op, as the rest of the client has always been).

## Test plan
- [x] All changed/new sources compile cleanly against the project's Thrift/Boost headers (`g++ -std=c++11 -fsyntax-only`, 0 errors); new code conforms to the repo `.clang-format` (2-space, 100-col).
- [ ] C++ client integration tests green in CI. The previous run failed on two issues, both addressed in this branch: the new `[sessionPool]` cleanup threw `508` on a fresh DB (now tolerated), and the reverted compression change broke `ts_session_open_with_compression` (protocol mismatch). Awaiting the re-run.

### DIFF
--- a/iotdb-client/client-cpp/src/main/SessionPool.cpp
+++ b/iotdb-client/client-cpp/src/main/SessionPool.cpp
@@ -170,9 +170,9 @@ std::shared_ptr<Session> SessionPool::acquire(int64_t timeoutMs) {
     } else {
       if (cv_.wait_until(lock, deadline) == std::cv_status::timeout && idleQueue_.empty() &&
           size_ >= maxSize_ && !closed_) {
-        throw IoTDBException("Wait to get session timeout in SessionPool, maxSize=" +
-                             std::to_string(maxSize_) + ", waitTimeoutMs=" +
-                             std::to_string(effectiveTimeout) + ".");
+        throw IoTDBException(
+            "Wait to get session timeout in SessionPool, maxSize=" + std::to_string(maxSize_) +
+            ", waitTimeoutMs=" + std::to_string(effectiveTimeout) + ".");
       }
     }
   }
--- a/iotdb-client/client-cpp/src/main/SessionPool.cpp
+++ b/iotdb-client/client-cpp/src/main/SessionPool.cpp
@@ -1,0 +1,266 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "SessionPool.h"
+
+void PooledSession::reset() {
+  if (session_ && pool_ != nullptr) {
+    pool_->putBack(session_, broken_);
+  }
+  pool_ = nullptr;
+  session_ = nullptr;
+  broken_ = false;
+}
+
+SessionPool::SessionPool(std::string host, int rpcPort, std::string username, std::string password,
+                         size_t maxSize)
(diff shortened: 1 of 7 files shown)


iotdb-client/client-cpp/src/main/SessionPool.cpp
```diff
@@ -0,0 +1,266 @@
...
+  }
+  pool_ = nullptr;
+  session_ = nullptr;
+  broken_ = false;
+}
+
+SessionPool::SessionPool(std::string host, int rpcPort, std::string username, std::string password,
+                         size_t maxSize)
+    : host_(std::move(host)), rpcPort_(rpcPort), username_(std::move(username)),
+      password_(std::move(password)), maxSize_(maxSize == 0 ? 1 : maxSize) {}
```
jt2594838: maxSize <= 0?

SpriCoder: Good catch. Since `maxSize` is `size_t` (unsigned), `<= 0` reduces to `== 0` and would trigger a tautological-comparison warning under `-Wall`. Rather than silently clamping an invalid `0` to `1`, I changed the constructors to fail fast and throw `IoTDBException` when `maxSize == 0`, so the misuse surfaces at construction time. Done in 8ca5a42.

iotdb-client/client-cpp/src/main/SessionPool.cpp
```diff
@@ -0,0 +1,273 @@
...
+      return session;
+    }
+    if (size_ < maxSize_) {
+      // Reserve a slot, then build the connection outside the lock so other
+      // borrowers are not blocked by network/handshake latency.
+      ++size_;
+      lock.unlock();
+      std::shared_ptr<Session> session;
+      try {
+        session = constructNewSession();
```
hongzhi-gao, line 157: If another thread calls close() while that connection is being built, closed_ becomes true, but the waiter still hands out a brand-new session after open() completes.

hongzhi-gao (approved):

iotdb-client/client-cpp/src/main/SessionPool.cpp
SpriCoder: Good catch, that was a real race. Fixed in 7d5fa7e: after `constructNewSession()` returns I now re-acquire the lock and re-check `closed_`. If the pool was closed while the connection was being built, the slot is released (`--size_`), the session is torn down outside the lock, and `acquire()` throws `IoTDBException("SessionPool is closed.")` instead of handing it out. The reserved slot is also accounted for correctly: `close()` only subtracts the idle count, and this in-flight slot is decremented on the closed path.

PR #17810, opened 2026-06-01
[IOTDB-17798] Implement table model NEXT fill

### LINKED ISSUE #17798: [Feature][Table Model] Support `NEXT` (backward) value fill in the `FILL` clause
### Search before asking

## Summary

The table model (relational SQL dialect) currently supports three fill methods in the `FILL` clause: `PREVIOUS`, `LINEAR`, and `CONSTANT`. It does **not** support **next/backward value fill** — i.e. filling a null with the nearest *later* non-null value along the time axis.

This issue proposes adding a `NEXT` fill method to the table model, mirroring the existing `PREVIOUS` fill in syntax, and reusing the forward look-ahead machinery that `LINEAR` fill already implements.

This is a good issue to pick up if you want to get familiar with the IoTDB table-model query engine (grammar → analyzer → planner → operator). Comment below to claim it; happy to help with pointers and review.

## Motivation

`PREVIOUS` fill carries the last known value *forward* in time. The symmetric and equally common need is to carry the next known value *backward* in time — useful when a reading is only valid up to the moment the next sample arrives, or when you want to back-fill leading gaps that `PREVIOUS` cannot fill (there is no earlier value to carry).

`NEXT` fill closes this gap and makes the table model's fill capabilities symmetric.

## Current behavior

Only `PREVIOUS` / `LINEAR` / `CONSTANT` are accepted. From the table-model grammar
`iotdb-core/relational-grammar/src/main/antlr4/org/apache/iotdb/db/relational/grammar/sql/RelationalSql.g4`:

```antlr
fillClause
    : FILL METHOD fillMethod
    ;

fillMethod
    : LINEAR timeColumnClause? fillGroupClause?                      #linearFill
    | PREVIOUS timeBoundClause? timeColumnClause? fillGroupClause?   #previousFill
    | CONSTANT literalExpression                                     #valueFill
    ;
```

Note: the `NEXT` keyword token **already exists** in this grammar (`NEXT: 'NEXT';`) and is already in the non-reserved keyword list, so no new reserved word is introduced.

## Proposed syntax

Mirror `PREVIOUS` exactly — `TIME_BOUND`, `TIME_COLUMN`, and `FILL_GROUP` are all optional and carry the same meaning:

```sql
-- basic
SELECT time, device_id, temperature FROM tbl
FILL METHOD NEXT;

-- only fill when the next non-null value is within 2ms of the current row
SELECT time, device_id, temperature FROM tbl
FILL METHOD NEXT TIME_BOUND 2ms;

-- explicit time column (1-based ordinal) and fill-group columns
SELECT time, device_id, temperature FROM tbl
FILL METHOD NEXT TIME_COLUMN 1 FILL_GROUP 2;
```

New grammar alternative to add:

```antlr
| NEXT timeBoundClause? timeColumnClause? fillGroupClause?           #nextFill
```

## Semantics

For each null cell, `NEXT` fill replaces it with the value of the nearest **later** (greater timestamp) non-null cell, within the same fill group, ordered by the time column.

- **Tail behavior:** rows after the last non-null value have no "next" value and remain null.
- **`TIME_BOUND`:** if the next non-null value's timestamp is farther than the bound from the current row, leave the cell null.
- **`TIME_COLUMN`:** which timestamp column defines ordering (defaults to the single `TIMESTAMP` column; error if it cannot be inferred — same rule as `PREVIOUS`/`LINEAR`).
- **`FILL_GROUP`:** fill does not cross group boundaries.
- **Data types:** all types are supported (boolean / text / blob / numeric), because `NEXT` simply copies a value — unlike `LINEAR`, which is numeric-only.

### Example

Input:

| time | temperature |
|-----:|------------:|
| 1 | null |
| 2 | 10.0 |
| 3 | null |
| 4 | null |
| 5 | 20.0 |
| 6 | null |

Result under each method:

| time | PREVIOUS | **NEXT** | LINEAR |
|-----:|---------:|---------:|-------:|
| 1 | null | **10.0** | null |
| 2 | 10.0 | **10.0** | 10.0 |
| 3 | 10.0 | **20.0** | 13.33 |
| 4 | 10.0 | **20.0** | 16.67 |
| 5 | 20.0 | **20.0** | 20.0 |
| 6 | 20.0 | **null** | null |

## Why build on top of `LINEAR` fill

`PREVIOUS` fill only needs to remember the last seen value, so it streams block-by-block trivially. `NEXT` fill must look **ahead** to a value it hasn't read yet, possibly in a later `TsBlock`. That forward look-ahead / cross-block buffering is exactly what `LINEAR` fill already implements (it needs both the previous and the next non-null value to interpolate).

So the recommended split is:
- **Syntax + data-type coverage + node shape** → copy from `PREVIOUS` fill.
- **Operator mechanics (look-ahead, cross-`TsBlock` buffering)** → copy from `LINEAR` fill, but instead of interpolating, just take the next non-null value (and apply `TIME_BOUND` if present).

The reusable look-ahead lives in
`iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/execution/operator/process/fill/linear/LinearFill.java`
(`fill(Column timeColumn, Column valueColumn, long startRowIndex)` plus the `nextRowIndex` / `nextRowIndexInCurrentColumn` forward scan).

## Implementation guide (table model only)

All paths are relative to the repo root and verified against current `master`.

| Layer | File | Change |
|-------|------|--------|
| Grammar | `iotdb-core/relational-grammar/.../sql/RelationalSql.g4` | Add `#nextFill` alternative to `fillMethod` (`NEXT` token already exists) |
| Fill method enum | `iotdb-core/node-commons/.../plan/statement/component/FillPolicy.java` | Add `NEXT((byte) 3)` |
| AST | `iotdb-core/node-commons/.../relational/sql/ast/Fill.java` + the `AstBuilder` that builds it | Handle the `nextFill` rule (same fields as `previousFill`: `timeBound`, `timeColumnIndex`, `groupingElements`) |
| Logical/physical node | `iotdb-core/node-commons/.../relational/planner/node/` (`FillNode.java`, `PreviousFillNode.java`) | Add `NextFillNode extends FillNode` mirroring `PreviousFillNode` (`timeBound`, `helperColumn`, `groupingKeys`); register its serialization type alongside the existing `TABLE_PREVIOUS_FILL_NODE` / `TABLE_LINEAR_FILL_NODE` in `PlanNodeType`, and add the `PlanVisitor.visitNextFill` hook |
| Analyzer | `iotdb-core/datanode/.../relational/analyzer/StatementAnalyzer.java` (`analyzeFill`, `analyzeFillGroup`) | Add a `FillPolicy.NEXT` branch + a `NextFillAnalysis` (mirror `PreviousFillAnalysis`: `TIME_COLUMN` inferred like `PREVIOUS`/`LINEAR`, `TIME_BOUND` optional) |
| Logical planner | `iotdb-core/datanode/.../relational/planner/QueryPlanner.java` (`visitFill`) | Build a `NextFillNode` from `NextFillAnalysis` |
| Operator generator | `iotdb-core/calc-commons/.../plan/planner/TableOperatorGenerator.java` (`visitPreviousFill`, `visitLinearFill`, `visitValueFill`) | Add `visitNextFill`; wire to the new operator (use the linear-fill operators as the template) |
| Operators | `iotdb-core/calc-commons/.../execution/operator/process/` (`AbstractLinearFillOperator.java`, `TableLinearFillOperator.java`, `TableLinearFillWithGroupOperator.java`) | Add `Table[…]NextFillOperator` + grouped variant, modeled on the linear-fill operators (they already do look-ahead / cross-block buffering) |
| Per-datatype fill (codegen) | `iotdb-core/calc-commons/src/main/codegen/templates/` (`previousFill.ftl`, `previousFillWithTimeDuration.ftl`, `linearFill.ftl`) | Add `nextFill.ftl` (and `nextFillWithTimeDuration.ftl` for `TIME_BOUND`) generating `IntNextFill`, `LongNextFill`, `FloatNextFill`, `DoubleNextFill`, `BooleanNextFill`, `BinaryNextFill` — all types, like `previousFill`; register the new template in `codegen/config.fmpp` |

### Suggested step-by-step

1. Grammar: add `#nextFill`; regenerate parser; confirm `FILL METHOD NEXT ...` parses.
2. `FillPolicy.NEXT`; AST + `AstBuilder` handling.
3. `NextFillNode` (+ `PlanNodeType` registration, serde, `PlanVisitor` hook).
4. `StatementAnalyzer` `NEXT` branch + `NextFillAnalysis`.
5. `QueryPlanner.visitFill` → `NextFillNode`.
6. `nextFill.ftl` / `nextFillWithTimeDuration.ftl` per-datatype classes (mirror `LinearFill`'s look-ahead, copy the next value instead of interpolating).
7. `TableNextFillOperator` (+ grouped variant) modeled on the linear-fill operators.
8. `TableOperatorGenerator.visitNextFill` wiring.
9. Tests + docs.

## Tests

- **Operator unit tests:** mirror the existing previous/linear fill operator tests (multi-`TsBlock` inputs, group boundaries, `TIME_BOUND`, the tail-stays-null case, all data types).
- **Analyzer tests:** `NEXT` accepted; `TIME_COLUMN` inference / error path; `FILL_GROUP` validation.
- **Integration tests:** extend the table-model fill IT (e.g. `IoTDBFillTableIT` / the relational fill IT) with `NEXT` cases including leading gap, trailing gap, grouping, and `TIME_BOUND`.

## Acceptance criteria

- [ ] `FILL METHOD NEXT [TIME_BOUND ...] [TIME_COLUMN ...] [FILL_GROUP ...]` parses and executes in the table model.
- [ ] Semantics match the example table above, including tail-null and `TIME_BOUND` behavior.
- [ ] Works for all data types and across multiple `TsBlock`s.
- [ ] `FILL_GROUP` does not leak values across groups.
- [ ] Plan node serialization round-trips (distributed plan).
- [ ] Unit + integration tests added; docs updated.

## Out of scope

- Tree-model `NEXT` fill (the tree model also lacks it; can be a follow-up).
- Changes to `LINEAR` / `PREVIOUS` / `CONSTANT` behavior.

## References

- Grammar: `iotdb-core/relational-grammar/.../sql/RelationalSql.g4` (`fillMethod`)
- Linear-fill look-ahead to reuse: `iotdb-core/calc-commons/.../execution/operator/process/fill/linear/LinearFill.java`
- Previous-fill operator/templates to mirror for syntax & data-type coverage: `previousFill.ftl`, `previousFillWithTimeDuration.ftl`, `PreviousFillWithGroupOperator.java`

### Solution

_No response_

### Alternatives

_No response_

DaZuiZui:
### Add table-model NEXT fill syntax and planning

This PR implements `FILL METHOD NEXT` for the table model.

`NEXT` fills a null value with the nearest non-null value that appears later in the current operator input order. It directly copies the future value, does not interpolate, and does not search by timestamp order. If no next value exists, the value remains null.

The implementation supports:

- plain `FILL METHOD NEXT`
- `TIME_BOUND`
- `TIME_COLUMN`
- `FILL_GROUP`
- numeric, boolean, binary/text/string/blob/object, date, and timestamp columns

Plain `NEXT` does not require a helper/time column. A helper/time column is only required or inferred when `TIME_BOUND` or `FILL_GROUP` needs it.

`FillPolicy.NEXT` is appended after existing enum values to keep existing byte/ordinal compatibility, and `TABLE_NEXT_FILL_NODE` uses a new plan node type id.

### Add NEXT fill plan node and optimizer support

This PR adds `NextFillNode` and wires it through serde, plan visitors, plan graph printing, pruning, symbol rewrite, analyzer, and planner logic.

For `TIME_BOUND` or `FILL_GROUP`, the analyzer resolves the required helper/time column similarly to `PREVIOUS`. For grouped fill, the planner reuses the existing fill-group sorting and grouping-symbol path.

`PushLimitOffsetIntoTableScan` treats `NEXT` like `LINEAR` and avoids pushing down `LIMIT/OFFSET`, because NEXT fill may need future rows that would otherwise be removed before filling.

### Add NEXT fill execution path

This PR adds table NEXT fill operators and the generated fill implementation.

The execution layer reuses `AbstractLinearFillOperator` for cross-`TsBlock` look-ahead caching and grouped slicing. The NEXT fill algorithm scans the current value column from right to left, uses the nearest non-null candidate in the current block when available, and otherwise uses the first non-null candidate found in later cached blocks.

For grouped fill, group boundaries reset the next-value state so values are not filled across groups.

### Tests

This PR adds and updates tests for:

- parsing `FILL METHOD NEXT`
- `NextFillNode` serde
- plain NEXT fill
- NEXT with `TIME_BOUND`
- NEXT with `TIME_COLUMN`
- NEXT with `FILL_GROUP`
- NEXT with `TIME_BOUND + TIME_COLUMN + FILL_GROUP`
- cross-`TsBlock` fill behavior
- group boundary isolation
- trailing null values
- helper/time column null handling
- boolean and binary/text-like values
- current input order semantics, including descending order and sorted subqueries
- `LIMIT` behavior without losing future next values

Closes #17798

<hr>

This PR has:
- [x] been self-reviewed.

<hr>

- `RelationalSql.g4`
- `FillPolicy`
- `Fill`
- `AstBuilder`
- `StatementAnalyzer`
- `Analysis.NextFillAnalysis`
- `QueryPlanner`
- `NextFillNode`
- `PlanNodeType`
- `CommonPlanNodeDeserializer`
- `ICoreQueryPlanVisitor`
- `PlanGraphPrinter`
- `PruneFillColumns`
- `PushLimitOffsetIntoTableScan`
- `UnaliasSymbolReferences`
- `CommonOperatorUtils`
- `TableOperatorGenerator`
- `TableNextFillOperator`
- `TableNextFillWithGroupOperator`
- `fill/next/NextFill`
- `nextFill.ftl`
- `FillStatementTest`
- `NextFillNodeSerdeTest`
- `NextFillTest`
- `IoTDBFillTableIT`


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/queryengine/plan/relational/planner/node/NextFillNode.java
```diff
@@ -0,0 +1,195 @@
...
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    if (!super.equals(o)) {
+      return false;
+    }
+    NextFillNode that = (NextFillNode) o;
+    return Objects.equals(timeBound, that.timeBound)
+        && Objects.equals(helperColumn, that.helperColumn)
+        && Objects.equals(groupingKeys, that.groupingKeys);
```
JackieTien97, line 183: `equals`/`hashCode` here correctly include `groupingKeys`. Worth noting the sibling `PreviousFillNode` does **not** — its `equals`/`hashCode` only use `timeBound` + `helperColumn`, so two PreviousFill nodes that differ only by `FILL_GROUP` keys compare equal. Not a problem for this PR (NextFill is the correct one), but since the two classes are otherwise identical you may want to align `PreviousFillNode` for consistency.

iotdb-core/datanode/src/test/java/org/apache/iotdb/db/queryengine/plan/relational/planner/NextFillNodeSerdeTest.java
```diff
@@ -0,0 +1,52 @@
...
+import org.junit.Test;
+
+import java.nio.ByteBuffer;
+
+import static org.junit.Assert.assertEquals;
+
+public class NextFillNodeSerdeTest {
+
+  @Test
+  public void testNextFillNodeSerde() throws Exception {
```
JackieTien97, line 43: This only exercises the all-fields-present path. `serializeAttributes`/`deserialize` each have three `false` branches (null `timeBound` / `helperColumn` / `groupingKeys`), and the all-null case is actually the most common one — plain `FILL METHOD NEXT`. Could you add a case with all three null, and ideally one partial combination, so those null branches are covered?

iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/execution/operator/process/TableNextFillWithGroupOperator.java
```diff
@@ -0,0 +1,167 @@
...
+  boolean noMoreTsBlockForCurrentGroup() {
+    return noMoreTsBlock || noMoreTsBlockForCurrentGroup.get(0);
+  }
+
+  @Override
+  void resetFill() {
+    boolean isNewGroup = Boolean.TRUE.equals(groupSplitter.remove(0));
+    boolean isNoMoreTsBlockForCurrentGroup =
+        Boolean.TRUE.equals(noMoreTsBlockForCurrentGroup.remove(0));
+    if (isNewGroup || isNoMoreTsBlockForCurrentGroup) {
```
JackieTien97: This reset condition diverges from `TableLinearFillWithGroupOperator#resetFill`, which resets only on `isNewGroup`. The extra `|| isNoMoreTsBlockForCurrentGroup` reset is what discards a cross-group candidate that Step 1 of `AbstractLinearFillOperator#next` may have prepared (via `prepareForNext`) *before* `resetFill` runs — `testNextFillWithGroupDoesNotUseNextGroupAfterContinuedGroup` covers that case, good.

Because `resetFill` runs after `prepareForNext`, could you add the symmetric **same-group** test: the first (or continued) block of a group ends in a trailing null whose fill value lives in a *later* block of the *same* group, so the look-ahead candidate must survive this reset. The existing unit test's first block has no trailing null, so this path is currently untested and it's the one most sensitive to this condition.

iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/execution/operator/process/TableNextFillOperator.java
```diff
@@ -0,0 +1,76 @@
...
+    this.hasTimeBound = hasTimeBound;
+  }
+
+  @Override
+  protected Column getHelperColumn(TsBlock tsBlock) {
+    return helperColumnIndex == -1 ? tsBlock.getTimeColumn() : tsBlock.getColumn(helperColumnIndex);
+  }
+
+  @Override
+  protected Integer getLastRowIndexForNonNullHelperColumn(TsBlock tsBlock) {
```
JackieTien97, line 58: Minor / defensive: when `hasTimeBound` is true this calls `getHelperColumn`, which does `tsBlock.getColumn(helperColumnIndex)` and would throw for `helperColumnIndex == -1`. This relies on the invariant "TIME_BOUND present ⇒ analyzer always resolves a helper column" (so the index is never -1 here), which holds today. A one-line comment — or a `checkArgument` in the constructor — documenting that invariant would prevent a confusing `IndexOutOfBounds` if a future caller ever builds this operator with `hasTimeBound=true, helperColumnIndex=-1`.

JackieTien97 (changes requested):
Thanks for the thorough work here — this is a clean, well-structured PR. It faithfully mirrors the existing PREVIOUS/LINEAR fill machinery end to end (grammar → AST/`AstBuilder` → `StatementAnalyzer`/`Analysis` → `QueryPlanner` → optimizer rules → `NextFillNode` + serde → operators + freemarker codegen), and the visitor coverage looks complete: the generic `visitFill` fallbacks (`SortElimination` keeps the sort for non-ValueFill, `TableDistributedPlanGenerator`/`TableModelTypeProviderExtractor`) all handle `NextFillNode` correctly through inheritance, and every rule that special-cases fill nodes (`PruneFillColumns`, `PushLimitOffsetIntoTableScan`, `UnaliasSymbolReferences`, `PlanGraphPrinter`, `TableOperatorGenerator`) was updated. The `1043` node-type id is free and the serde order matches the siblings. Test coverage (parser / serde / fill execution / IT) is solid.

I've left a few minor, non-blocking comments inline — mostly test-coverage hardening plus one defensive note. The only item I'd like confirmation on before merge is the grouped cross-`TsBlock` reset path in `TableNextFillWithGroupOperator#resetFill` (see inline comment): it's correct for the cross-group case that's already tested, but the symmetric same-group look-ahead case isn't covered yet and is the one most sensitive to that reset condition.

DaZuiZui:
Thanks for the review. The issue has been fixed. @JackieTien97

JackieTien97 (approved):

DaZuiZui:
@JackieTien97  This failure looks like a self-hosted runner environment issue: `mvn` was not found before the integration tests started. Could you please rerun the failing check `Table Cluster IT - 1C3D / Simple (17)`?

DaZuiZui:
Thanks for the review!

I have addressed the latest Copilot comments in this PR:

1. Added an explicit `requireNonNull(fillMethod, "fillMethod is null")` check in the `Fill` constructor before comparing the fill policy.

2. Removed the redundant `groupSplitter` state from `TableNextFillWithGroupOperator`, including its field, initialization, and add/remove bookkeeping. The grouped NEXT fill reset logic still relies on `noMoreTsBlockForCurrentGroup`, so the existing behavior is preserved.

I also verified the related test with:

`mvn test -pl iotdb-core/datanode -am -Dtest=NextFillTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false -DskipITs`

Result: `Tests run: 5, Failures: 0, Errors: 0, Skipped: 0`, `BUILD SUCCESS`.

PR #17815, opened 2026-06-02
Fix pipe tsfile receiver database handling

Caideyipi:
## Description

This patch fixes pipe tsfile receiver behavior when the sender provides a tree database name such as 
oot.test.sg_0:

- derive the tree database level from the provided database name for sync tsfile load, so 
oot.test.sg_0 uses database level 2 instead of the receiver default
- apply the same derived database level to async active-load attributes so async and sync pipe tsfile load behave consistently
- classify IoTDBRuntimeException(DATABASE_NOT_EXIST) as PIPE_RECEIVER_PARALLEL_OR_USER_CONFLICT_EXCEPTION, matching the TSStatus visitor path instead of returning 305
- reduce repeated receiver statement exception WARN stack logs via the existing PipePeriodicalLogReducer style

## Tests

- git diff --check
- mvn '-Ddevelocity.off=true' '-Dcheckstyle.skip=true' '-DargLine=-Xmx512m' test -pl iotdb-core/datanode '-Dtest=IoTDBDataNodeReceiverTest,PipeStatementTsStatusVisitorTest,LoadTsFileStatementTest'
  - blocked before test execution by existing generated-source compilation errors in 	arget/generated-sources/freemarker/.../*ModeAccumulator.java (ComparatorChain missing and IoTDBConfig.getModeMapSizeThreshold() missing)

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/receiver/protocol/thrift/IoTDBDataNodeReceiver.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/receiver/protocol/thrift/IoTDBDataNodeReceiver.java
@@ -59,6 +59,7 @@ import org.apache.iotdb.db.pipe.receiver.visitor.PipeTableStatementDataTypeConve
 import org.apache.iotdb.db.pipe.receiver.visitor.PipeTreeStatementDataTypeConvertExecutionVisitor;
 import org.apache.iotdb.db.pipe.receiver.visitor.PipeTreeStatementToBatchVisitor;
 import org.apache.iotdb.db.pipe.resource.PipeDataNodeResourceManager;
+import org.apache.iotdb.db.pipe.resource.log.PipePeriodicalLogReducer;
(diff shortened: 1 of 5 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/receiver/protocol/thrift/IoTDBDataNodeReceiver.java
```diff
@@ -860,6 +875,29 @@ private TSStatus executeStatementAndClassifyExceptions(
...
+          Objects.isNull(statement) ? null : statement.getPipeLoggingString());
+    }
+  }
+
+  static boolean shouldLogStatementException(
+      final long receiverId, final Statement statement, final Exception e) {
+    // Use the reducer cache as a gate. The actual stack trace is logged only when it passes.
+    return PipePeriodicalLogReducer.log(
+        message -> {},
+        "Receiver id = %s, statement = %s, exception = %s, message = %s",
```
jt2594838, line 894: i18n?

PR #17821, opened 2026-06-02
Improve ConfigNode leader warm-up before serving

CRZbulabula:
### Summary
- Gate ConfigNode leader confirmation on LoadCache warm-up after consensus leader-ready.
- Track first heartbeat coverage for Nodes, Regions, RegionGroups, and ConsensusGroups before serving requests.
- Return CONFIG_NODE_LEADER_WARMING_UP during warm-up so DataNodes wait and retry the current ConfigNode instead of treating it as redirection.

### Tests
- mvn spotless:apply -pl iotdb-core/confignode,iotdb-core/datanode,iotdb-client/service-rpc
- mvn compile -pl iotdb-client/service-rpc,iotdb-core/confignode
- mvn test -pl iotdb-core/confignode -Dtest=LoadManagerTest
- mvn compile -pl iotdb-client/service-rpc,iotdb-core/datanode (fails in unrelated existing sources: ArrayDeviceTimeIndex.java and TableDeviceSchemaCache.java still pass IDeviceID to PartialPath.matchFullPath)


Caideyipi (commented):
I reviewed the warm-up changes on `a57680d2542`. I think there are a few issues that should be fixed before merge:

1. **AINode treats the new warm-up status as a hard failure.** `ConfigManager.registerAINode()` now returns `CONFIG_NODE_LEADER_WARMING_UP` while `confirmLeader()` is warming up, but the Python AINode client only treats `REDIRECTION_RECOMMEND` as retryable in `_update_config_node_leader()`. `node_register()` / `node_restart()` then call `verify_success()` and raise on status `1014`, so an AINode can fail startup if it hits the leader during warm-up. Please add the new code to the AINode constants and retry handling paths.

2. **Non-seed ConfigNode registration has the same gap.** `registerConfigNode()` can now return `CONFIG_NODE_LEADER_WARMING_UP`, but `ConfigNode.sendRegisterConfigNodeRequest()` only retries success/redirection/internal-retry statuses and throws `StartupException` for anything else. A ConfigNode joining during leader warm-up can fail immediately instead of waiting and retrying.

3. **The async leader-service startup has a stepdown race.** `notifyLeaderReady()` now submits `startLeaderServicesAfterLoadReady()` asynchronously. That task checks `isLeaderReady()` only once before starting leader-only services and setting `leaderServicesReady=true`. If `notifyNotLeader()` runs after that check but before/during service startup, the old task can re-enable services after cleanup. Please guard this with a leader epoch/cancellation token, and re-check before setting `leaderServicesReady`.

4. **The DataNode register retry budget is too tight for the 30s warm-up tolerance.** On `CONFIG_NODE_LEADER_WARMING_UP`, `updateConfigNodeLeader()` sleeps 2s and returns retryable, while `registerDataNode()` has 15 attempts. The final request can still happen before the 30s tolerance expires, then sleep and exit without one post-tolerance attempt. A deadline-based retry or a larger retry budget would avoid this edge case.

CRZbulabula:
@Caideyipi Thanks for the detailed review. Fixed in the latest commits, especially f54b4847f0.

1. AINode registration now treats CONFIG_NODE_LEADER_WARMING_UP as retryable instead of a hard failure.
2. Non-seed ConfigNode registration now waits and retries when the leader returns CONFIG_NODE_LEADER_WARMING_UP.
3. ConfigRegionStateMachine now uses a leader-services epoch guard, serializes startup and cleanup with leaderServicesLock, and re-checks the epoch before marking leader services ready.
4. DataNode registration now uses a 60s warm-up retry deadline, so it has requests after the 30s first-heartbeat tolerance.

I also cleaned up the follow-up warm-up sampling concerns: removed the unreported DataNode Region heartbeat chain, removed the extra DataNodeHeartbeatHandler region-group argument, and kept consensus sampling to only cache leader samples when the DataNode reports leader=true with a consensus logical timestamp.

Caideyipi (commented):
I still see one remaining stepdown race that I think should be fixed before merge.

`ConfigRegionStateMachine.submitIfLeaderServicesEpochCurrent()` checks the epoch only before invoking `task.run()`. If an async task passes that check, then `notifyNotLeader()` runs and finishes cleanup, the old task can resume and start leader-only services such as `startCQScheduler()`, `startPipeMetaSync()`, `startPipeHeartbeat()`, or `startSubscriptionMetaSync()` after the node is no longer leader.

The main startup path is guarded by `leaderServicesLock`, but these submitted tasks are not serialized with cleanup. Please either run the task under `leaderServicesLock` and re-check `isCurrentLeaderServicesEpoch(epoch)` inside the lock, or pass a cancellation/epoch guard into the individual service start paths.

CRZbulabula:
@Caideyipi Good catch — thanks. Fixed in e07433767a by reworking the leader-services lifecycle so this race can no longer happen.

The root cause was that `submitIfLeaderServicesEpochCurrent()` only checked the epoch *before* `task.run()`, and those submitted tasks were not serialized against `notifyNotLeader()`'s cleanup. I removed that helper entirely. The new design:

1. **All transitions are serialized on a single-thread executor.** `notifyLeaderReady` (become-leader), `notifyNotLeader` / `notifyLeaderChanged` (resign) all submit to one single-thread `leaderServicesTransitionExecutor`. Because it has exactly one worker, a become-leader orchestration and a resign cleanup can never run concurrently — one runs to completion before the other starts. So `startCQScheduler()` / `startPipeMetaSync()` / `startPipeHeartbeat()` / `startSubscriptionMetaSync()` can no longer interleave with cleanup.

2. **The epoch is bumped eagerly on resign, before cleanup is even queued.** `notifyNotLeader` calls `invalidateLeaderServices()` synchronously on the consensus thread, so the epoch advances the instant we lose leadership. An in-flight `becomeLeader` re-checks `isCurrentLeaderServicesEpoch(epoch)` after the parallel startups join and again before it sets `leaderServicesReady = true`, so a stale epoch bails out and never re-enables services after cleanup.

3. **`leaderServicesReady` is only set inside `leaderServicesLock` with the epoch re-checked**, so the "set ready" step is atomic with respect to the epoch.

Within a single become-leader epoch, load services still start first (for warm-up), then the remaining independent services start in parallel on a cached pool and are joined before the epoch is marked ready. So the check-then-run gap you pointed out is closed both by the single-thread serialization and by the epoch re-check inside the lock.

PR #17831, opened 2026-06-03
Make Tier-A metadata operations HA when a DataNode is down (metadata lease + self-fencing)

JackieTien97:
## Problem

Many cluster metadata operations broadcast a cache-invalidation to **every** registered DataNode and treat *any* unreachable DataNode as a hard failure. As a result, with multiple replicas configured, a **single down DataNode** still makes these operations fail — contradicting the HA goal. The original report was table-model DDL (`CREATE TABLE`, `ALTER TABLE ...`), but the same pattern affects ~all Tier-A metadata ops (tree-model schema, templates, views, TTL, `DROP DATABASE`, permissions).

The reason the broadcast cannot simply "skip" an unreachable DataNode is **correctness**: a DataNode caches ConfigNode-pushed metadata (table/tree schema, permissions, TTL, ...). If it misses an invalidation during a network partition and keeps serving the stale cache, it can produce dirty data / stale reads / stale authorization.

## Approach — metadata lease + self-fencing + a broadcast verdict

1. **DataNode self-fencing (DN side, fail-closed).** The ConfigNode already heartbeats every DataNode. A DataNode records the last ConfigNode-heartbeat time on its own monotonic clock; if none arrives within `metadata_lease_fence_ms` (`T_fence`, default 20s, aligned with the failure detector), it considers itself **fenced** and stops trusting ConfigNode-pushed caches. On the next heartbeat (recovery) it resyncs.
2. **ConfigNode broadcast verdict (CN side, liveness).** Instead of "any unreachable DataNode fails the op", the ConfigNode proceeds once every un-acked DataNode is **provably self-fenced** — out of contact for at least `T_proceed = T_fence + margin`, *and* known to support fencing. Such a DataNode is fail-closed and will resync, so it cannot serve dirty data. A DataNode that is reachable-but-unacked (still possibly serving) blocks the op (wait, then fail).

This gives the desired CP behavior: a healthy majority makes progress; a partitioned minority fails closed.

## What this PR contains

**DataNode side (fail-closed when fenced):**
- `MetadataLeaseManager` — lease tracking, lazy fence check, recovery listeners (monotonic clock, injectable for tests)
- Table schema cache (`DataNodeTableCache`) — throws retryable; tree schema cache (`TreeDeviceSchemaCacheManager`) — forces re-fetch from the quorum-backed SchemaRegion; permission cache (`ClusterAuthorityFetcher`) — drops cache → re-fetch (deny while partitioned); TTL — compaction uses an infinite TTL when fenced (never deletes by a stale TTL)
- Heartbeat records contact + reports a `supportsMetadataLeaseFencing` capability bit; a `metadata_lease_heartbeat_age_ms` metric

**ConfigNode side (proceed past provably-fenced DataNodes):**
- `MetadataBroadcastVerdict` (pure decision logic) + `DataNodeContactTracker` (sound per-DataNode last-successful-response signal, capability bit) + `ClusterCachePropagator` (broadcast → verdict → PROCEED/WAIT/FAIL with retry)
- Leadership/removal lifecycle hooks for the contact tracker
- All Tier-A procedures wired through the propagator: table DDL (`CreateTable` + all 8 alter/drop procedures, forward + rollback), tree-model schema (`DeleteTimeSeries`, `AlterTimeSeriesDataType`, `AlterEncodingCompressor`, logical views, `DeactivateTemplate`), templates (`Set`/`Unset`), `SetTTL`, and `DeleteDatabase` (sync path). Region-task / physical-deletion broadcasts are deliberately left on region consensus.
- `AuthOperationProcedure`: fixed a silent permission-staleness hole (it used to drop an unacked DataNode after a timeout, leaving a live DataNode serving a just-revoked permission); now proceeds only past provably-fenced DataNodes.

## Testing

- Unit tests for every new component and each fail-closed injection (`MetadataLeaseManager`, `DataNodeTableCache`, tree schema, permission, TTL-compaction, `MetadataBroadcastVerdict`, `DataNodeContactTracker`, `ClusterCachePropagator`), plus no-regression on the affected procedures.
- **1C3D integration test** (`IoTDBTableDDLHAIT`): with one DataNode stopped, `CREATE TABLE` **succeeds** after ~`T_proceed` (instead of failing immediately). A new `setMetadataLeaseFenceMs` IT-config setter keeps the wait short.

## Configuration & compatibility

- New knob `metadata_lease_fence_ms` (default 20000). `T_proceed = T_fence + internal margin (~5s)`.
- Takes effect on upgrade, no switch. Rolling-upgrade safe via the capability bit: an old DataNode (no fencing) is never treated as fenced, so the ConfigNode falls back to the strict (current) semantics for it; correctness is preserved DataNode-by-DataNode.

## Not in this PR (follow-up)

Tier-B resource operations (DROP/CREATE `FUNCTION`/`TRIGGER`/`PIPE PLUGIN`, `SET SYSTEM STATUS=ReadOnly`) and the quota re-pull. These need DataNode-side fail-closed on resource **execution** (not a read-through cache) plus a resource recovery-resync mechanism, which is a separate, self-contained effort.


Caideyipi:
I reviewed this implementation and the design note. A few safety-semantics points still look worth tightening or proving explicitly:

1. **There is a window during lease recovery before recovery listeners finish.**
   `MetadataLeaseManager.recordConfigNodeHeartbeat()` updates `lastConfigNodeHeartbeatNanos` before running recovery listeners. During listener execution, and also if a listener throws and the exception is swallowed, `isFenced()` already returns false, so stale caches may be trusted again. That does not match the safety condition that caches are invalidated/resynced before the DataNode leaves the fenced state.
   I suggest making the lease state explicit, e.g. `FENCED -> RECOVERING -> ACTIVE`: only switch to ACTIVE after critical recovery listeners/resync complete successfully; otherwise remain fenced.
   Code:
   https://github.com/apache/iotdb/blob/cc0cbcc1432c804ac4939a3c536279e4994f5d57/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/lease/MetadataLeaseManager.java#L85-L94

2. **Leader-change safety still needs an epoch/token-level proof.**
   A DataNode currently renews the metadata lease on any `getDataNodeHeartBeat` request, and the heartbeat does not carry a leader term or lease token. Resetting `lastResp` on the new leader avoids using stale history, but it does not prevent an old leader in a minority partition from continuing to renew some DataNodes. If the new leader cannot receive responses from those DataNodes, it may classify them as FENCED_SAFE after `T_proceed`, while they are actually being renewed by the old leader and have not self-fenced.
   This needs either a strict proof that an old leader cannot keep sending effective lease-renewing heartbeats after the new leader is ready, or a heartbeat leader epoch/token so a DataNode only renews from the currently valid token.
   Code:
   https://github.com/apache/iotdb/blob/cc0cbcc1432c804ac4939a3c536279e4994f5d57/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/thrift/impl/DataNodeInternalRPCServiceImpl.java#L2227-L2232

3. **`AuthOperationProcedure` still has a post-consensus failure semantics issue.**
   Removing the old 180s silent drop is a good improvement. However, `writePlan` commits to the ConfigNode consensus first, and then `ClusterCachePropagator.propagate()` may return false and call `setFailure`. At that point the permission change is already effective on the ConfigNode, while a heartbeat-fresh DataNode whose invalidate RPC keeps failing may still serve with the old permission cache. For security-sensitive operations such as REVOKE / DROP USER / DROP ROLE, failing the procedure after the write does not restore the safety semantics.
   I suggest keeping this step pending/retrying after a successful write until every DataNode has acked or is fenced, or actively fencing the unsafe DataNode, or introducing an auth version/token so the DataNode fails closed when its permission version is not known to be current.
   Code:
   https://github.com/apache/iotdb/blob/cc0cbcc1432c804ac4939a3c536279e4994f5d57/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/sync/AuthOperationProcedure.java#L100-L118
   https://github.com/apache/iotdb/blob/cc0cbcc1432c804ac4939a3c536279e4994f5d57/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/impl/sync/AuthOperationProcedure.java#L155-L165

4. **Please add a few negative/race tests for the safety proof.**
   The current IT covers `CREATE TABLE` succeeding with one stopped DataNode, but the more important safety cases are still uncovered:
   - heartbeat succeeds while the metadata broadcast RPC keeps failing;
   - leader change with stale/continued heartbeats from the old leader;
   - recovery listener delay/failure must not make the DataNode leave fenced state early;
   - after REVOKE or another permission change, a partitioned or half-broken DataNode must not continue authorizing with the old cache.

These points do not block the availability goal for a stopped DataNode, but they do affect whether the fencing scheme is fully safe under the failure modes it is meant to cover.

PR #17841, opened 2026-06-04
Fix table delete with renamed time column

jt2594838:
## Description

This PR fixes table-model DELETE predicate parsing when the TIME column has been renamed. The deletion analyzer now resolves the actual TIME column name from table schema instead of assuming the default time column name.

It also adds i18n messages for the newly added semantic error and covers the renamed-time-column path in unit tests.

## Tests

- mvn spotless:apply -pl integration-test -P with-integration-tests
- mvn test -pl iotdb-core/datanode -Dtest=AnalyzeUtilsTest,PredicateUtilsTest

Note: Running IoTDBDeletionTableIT#testDeleteWithRenamedTimeColumn was attempted, but the test class failed in its shared setup/teardown before the case body due to CREATE/DROP DATABASE IF EXISTS test being parsed by the tree-model parser in this local environment.

### DIFF
--- a/integration-test/src/test/java/org/apache/iotdb/relational/it/db/it/IoTDBDeletionTableIT.java
+++ b/integration-test/src/test/java/org/apache/iotdb/relational/it/db/it/IoTDBDeletionTableIT.java
@@ -415,6 +415,27 @@ public class IoTDBDeletionTableIT {
   }
 
   @Test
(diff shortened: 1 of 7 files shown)


Caideyipi (approved):
LGTM. I did not find any functional issues in the production change.

Issue #17797, opened 2026-05-30
[Feature] Allow referencing SELECT column aliases in GROUP BY / ORDER BY (and support lateral column alias in SELECT)


JackieTien97:
### Motivation

In the table model, IoTDB already supports column aliases via `AS`, but those aliases **cannot** be reused in `GROUP BY` / `ORDER BY`, nor by later items in the same `SELECT` list. Users have to repeat the full (often long) expression or fall back to positional references like `GROUP BY 1`. This hurts readability and is error-prone for downsampling/aggregation queries, which are extremely common in time-series workloads.

**Today users must write:**

```sql
SELECT date_bin(1h, time) AS hour_time, AVG(temperature) AS avg_temp
FROM table1
GROUP BY date_bin(1h, time)     -- must repeat the whole expression
ORDER BY date_bin(1h, time);    -- ...and repeat it again
```

**What we'd like to support:**

```sql
SELECT date_bin(1h, time) AS hour_time, AVG(temperature) AS avg_temp
FROM table1
GROUP BY hour_time
ORDER BY hour_time;
```

### Proposed Feature

**Part 1 — Use SELECT aliases in `GROUP BY` and `ORDER BY`.**
Allow an alias defined in the `SELECT` list to be referenced as a grouping/ordering key. Existing positional references (`GROUP BY 1`) and full-expression references should keep working.

**Part 2 — Lateral Column Alias (LCA) in the `SELECT` list.**
Allow a later select item to reference an alias defined by an earlier select item in the same `SELECT`:

```sql
SELECT
  date_bin(1h, time)  AS hour_time,
  AVG(temperature)    AS avg_temp,
  avg_temp - 273.15   AS avg_temp_celsius   -- references the earlier alias
FROM table1
GROUP BY hour_time;
```

### Prior Art (how other databases behave)

| Database | Alias in `GROUP BY` | Alias in `ORDER BY` | Lateral alias in `SELECT` |
|---|:---:|:---:|:---:|
| SQL standard | ❌ | ✅ | ❌ |
| PostgreSQL | ✅ (extension) | ✅ | ❌ |
| MySQL | ✅ | ✅ | ❌ |
| SQL Server | ❌ | ✅ | ❌ |
| Oracle | ✅ 23c+ / ❌ before | ✅ | ❌ |
| BigQuery | ✅ | ✅ | ❌ |
| ClickHouse | ✅ | ✅ | ✅ |
| DuckDB | ✅ | ✅ | ✅ |
[...]

PR #17843, opened 2026-06-04
Table: Support SELECT aliases in GROUP BY and ORDER BY

DaZuiZui:
## Description

This PR implements Part 1 of #17797 for the table model SQL analyzer.

It allows explicit SELECT aliases to be referenced in `GROUP BY` and `ORDER BY`.

For example:

```sql
SELECT date_bin(1h, time) AS hour_time, AVG(s1) AS avg_s1
FROM table1
GROUP BY hour_time
ORDER BY hour_time;
```

The alias is resolved during analysis, so existing semantic checks still apply after alias resolution.

### Alias precedence rules

This PR documents and implements the name resolution rules discussed in #17797:

- `GROUP BY` prefers current-query input columns over SELECT aliases. If an unqualified name does not resolve to a local input column, it may resolve to a matching SELECT alias.
- `ORDER BY` prefers SELECT output aliases over input columns. If no SELECT alias matches, it falls back to the existing `ORDER BY` name resolution behavior.
- `ORDER BY` alias resolution also applies to `SELECT DISTINCT`, for example `SELECT DISTINCT s1 AS x FROM table1 ORDER BY x`.
- Duplicate matching SELECT aliases are rejected with an ambiguity error.

### Scope

This PR only handles Part 1 of #17797:

- SELECT alias in `GROUP BY`
- SELECT alias in `ORDER BY`

The following items are intentionally left out of scope for a follow-up PR:

- Lateral column alias references in the SELECT list
- Alias references in `WHERE`
- Alias references in `HAVING`

Refs #17797

<hr>

This PR has:

- [x] been self-reviewed.

<hr>

- `StatementAnalyzer`
- `SelectAliasReuseTest`
- `TestMetadata`

### Test

```bash
./mvnw test -pl iotdb-core/datanode -am -Dtest=SelectAliasReuseTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false -DskipITs
```

Result:

```text
Tests run: 12, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/analyzer/StatementAnalyzer.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/analyzer/StatementAnalyzer.java
@@ -411,7 +411,10 @@ public class StatementAnalyzer {
   }
 
   private static boolean resolvesToInputColumn(Scope scope, Identifier identifier) {
-    return scope.tryResolveField(identifier, QualifiedName.of(identifier.getValue())).isPresent();
+    return scope
+        .tryResolveField(identifier, QualifiedName.of(identifier.getValue()))
+        .filter(ResolvedField::isLocal)
+        .isPresent();
   }
 
   private static Optional<SelectAlias> resolveSelectAlias(
@@ -1652,11 +1655,12 @@ public class StatementAnalyzer {
           ImmutableList.builder();
       ImmutableList.Builder<SelectAlias> selectAliasBuilder = ImmutableList.builder();
 
-      int selectItemPosition = 1;
+      int outputPosition = 1;
       for (SelectItem item : node.getSelect().getSelectItems()) {
         if (item instanceof AllColumns) {
-          analyzeSelectAllColumns(
-              (AllColumns) item, node, scope, outputExpressionBuilder, selectExpressionBuilder);
+          outputPosition +=
+              analyzeSelectAllColumns(
+                  (AllColumns) item, node, scope, outputExpressionBuilder, selectExpressionBuilder);
         } else if (item instanceof SingleColumn) {
           SingleColumn singleColumn = (SingleColumn) item;
           Expression selectExpression = singleColumn.getExpression();
@@ -1673,6 +1677,7 @@ public class StatementAnalyzer {
             for (Expression expression : expandedExpressions) {
               analyzeSelectSingleColumn(
                   expression, node, scope, outputExpressionBuilder, selectExpressionBuilder);
+              outputPosition++;
             }
           } else {
             analyzeSelectSingleColumn(
@@ -1680,14 +1685,14 @@ public class StatementAnalyzer {
             if (singleColumn.getAlias().isPresent()) {
               Identifier alias = singleColumn.getAlias().get();
               selectAliasBuilder.add(
-                  new SelectAlias(alias.getCanonicalValue(), selectExpression, selectItemPosition));
+                  new SelectAlias(alias.getCanonicalValue(), selectExpression, outputPosition));
             }
+            outputPosition++;
           }
         } else {
           throw new IllegalArgumentException(
               "Unsupported SelectItem type: " + item.getClass().getName());
         }
-        selectItemPosition++;
       }
       analysis.setSelectExpressions(node, selectExpressionBuilder.build());
 
@@ -2497,7 +2502,7 @@ public class StatementAnalyzer {
       }
     }
 
-    private void analyzeSelectAllColumns(
+    private int analyzeSelectAllColumns(
         AllColumns allColumns,
         QuerySpecification node,
         Scope scope,
@@ -2543,7 +2548,7 @@ public class StatementAnalyzer {
                             () ->
                                 new NoSuchElementException(
                                     DataNodeQueryMessages.NO_VALUE_PRESENT)));
-            analyzeAllColumnsFromTable(
+            return analyzeAllColumnsFromTable(
                 fields,
                 allColumns,
                 node,
@@ -2552,7 +2557,6 @@ public class StatementAnalyzer {
                 selectExpressionBuilder,
                 relationType,
                 local);
-            return;
           }
         }
         // identifierChainBasis.get().getBasisType == FIELD or target expression isn't a
@@ -2582,7 +2586,7 @@ public class StatementAnalyzer {
               DataNodeQueryMessages.SELECT_NOT_ALLOWED_FROM_RELATION_THAT_HAS_NO);
         }
 
-        analyzeAllColumnsFromTable(
+        return analyzeAllColumnsFromTable(
             fields,
             allColumns,
             node,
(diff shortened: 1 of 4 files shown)


Caideyipi:
I found two correctness issues in the alias resolution path:

1. `GROUP BY` input-column precedence currently treats outer-scope columns as input columns. `resolvesToInputColumn()` calls `scope.tryResolveField(...).isPresent()`, but that can resolve through the query boundary into a correlated outer scope. As a result, a correlated subquery such as:

```sql
SELECT x,
       (SELECT COUNT(*) FROM table1 GROUP BY x)
FROM table_with_x
```

or an inner query with `SELECT expr AS x ... GROUP BY x` can have `GROUP BY x` blocked from resolving to the inner SELECT alias just because the outer query has a column named `x`. The stated rule is that `GROUP BY` prefers input columns, which should mean the current query source scope, not outer query columns. This should check the resolved field is local to the current source scope, e.g. via `ResolvedField.isLocal()` / relation id, before suppressing alias resolution.

2. `ORDER BY` aliases that point to window functions are re-analyzed as window functions in the ORDER BY phase. For example:

```sql
SELECT row_number() OVER (ORDER BY s1) AS rn
FROM table1
ORDER BY rn
```

`ORDER BY rn` is rewritten to the `row_number() OVER (...)` expression, so the function is collected both in `analysis.getWindowFunctions(node)` and `analysis.getOrderByWindowFunctions(orderBy)`. `QueryPlanner` then plans SELECT window functions first and ORDER BY window functions again after switching to the ORDER BY scope, producing duplicate window planning instead of ordering by the SELECT output alias. For ORDER BY alias references, the planner should reuse the SELECT output symbol / field reference rather than treating the alias target as a fresh ORDER BY window expression.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/analyzer/StatementAnalyzer.java
```diff
@@ -2706,6 +2799,7 @@ private Analysis.GroupingSetAnalysis analyzeGroupBy(
                 column = outputExpressions.get(toIntExact(ordinal - 1));
                 verifyNoAggregateWindowOrGroupingFunctions(column, "GROUP BY clause");
               } else {
+                column = resolveGroupBySelectAlias(column, scope, selectAliases);
```
JackieTien97: Could we apply this same alias-resolution step to the `GroupingSets` branch below as well? Right now only `SimpleGroupBy` rewrites SELECT aliases, so queries such as `SELECT s1 AS x, COUNT(*) FROM table1 GROUP BY ROLLUP(x)` still reach `analyzeExpression(column, scope)` with `x` unresolved and fail with `Column 'x' cannot be resolved`. Since `ROLLUP`, `CUBE`, and `GROUPING SETS` are still `GROUP BY` grouping elements, they should follow the same input-column precedence and SELECT-alias fallback rule.

JackieTien97 (changes requested):
Found one correctness gap in the GROUP BY alias resolution path.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/analyzer/StatementAnalyzer.java
```diff
@@ -4102,11 +4212,16 @@ private FieldReference getFieldReferenceForFillGroup(
     }
 
     private List<Expression> analyzeOrderBy(
-        Node node, List<SortItem> sortItems, Scope orderByScope) {
+        Node node,
+        List<SortItem> sortItems,
+        Scope sourceScope,
+        Scope orderByScope,
```
JackieTien97: why we need another `sourceScope`?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/analyzer/StatementAnalyzer.java
```diff
@@ -367,6 +367,74 @@ private enum UpdateKind {
...
+    }
+
+    private List<SelectAlias> getAliases() {
+      return aliases;
+    }
+  }
+
+  private static final class SelectAlias {
+    private final String canonicalName;
+    private final Expression expression;
```
JackieTien97: no need to store this? For `resolveGroupBySelectAlias`, we only need `position` too. With `position`, you can just pick expression from `outputExpressions` just like previous `LongLiteral`

JackieTien97 (changes requested):
Add all related test cases in an existing IT class, make sure that includes all the normal and corner cases. All new functionalities need ITs besides UTs.

DaZuiZui:
Thanks @Caideyipi , @JackieTien97  for the careful review! 

I’ve addressed the comments in the latest commits:
- Fixed SELECT alias resolution for GROUP BY ROLLUP / CUBE / GROUPING SETS, including resolved-expression mapping for grouping sets.
- Restored ORDER BY analysis to use orderByScope by default, while preserving explicit SELECT alias precedence.
- Simplified SelectAlias to store only the canonical alias name and output position.
- Added/updated analyzer UT coverage in SelectAliasReuseTest, including the ORDER BY window alias regression case.
- Added table-model IT coverage in IoTDBTableAggregationIT.

I also verified the fix locally with:
- SelectAliasReuseTest
- IoTDBTableAggregationIT#selectAliasInGroupByAndOrderByTest
- spotless for datanode and integration-test

DaZuiZui:
Updated the test result summary for the latest PR head.

Verified locally:

**UT**

SelectAliasReuseTest

- Tests run: 17
- Failures: 0
- Errors: 0
- Skipped: 0
- Result: BUILD SUCCESS

**IT**

IoTDBTableAggregationIT#selectAliasInGroupByAndOrderByTest

- Tests run: 1
- Failures: 0
- Errors: 0
- Skipped: 0
- Result: BUILD SUCCESS

IT command:

./mvnw -o verify -DskipUTs -Dit.test=IoTDBTableAggregationIT#selectAliasInGroupByAndOrderByTest -DfailIfNoTests=false -Dfailsafe.failIfNoSpecifiedTests=false -pl integration-test -am -PTableSimpleIT -P with-integration-tests

Note: the PR description still mentions 12 UTs, while the latest SelectAliasReuseTest now runs 17 tests.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/analyzer/StatementAnalyzer.java
```diff
@@ -4102,11 +4210,12 @@ private FieldReference getFieldReferenceForFillGroup(
     }
 
     private List<Expression> analyzeOrderBy(
-        Node node, List<SortItem> sortItems, Scope orderByScope) {
+        Node node, List<SortItem> sortItems, Scope orderByScope, List<SelectAlias> selectAliases) {
       ImmutableList.Builder<Expression> orderByFieldsBuilder = ImmutableList.builder();
 
       for (SortItem item : sortItems) {
         Expression expression = item.getSortKey();
+        Scope expressionScope = orderByScope;
```
JackieTien97: I don't understand why we still need this? It seems that it will always be `orderByScope`

JackieTien97 (changes requested):

JackieTien97 (approved):

PR #17854, opened 2026-06-05
Add generic SSL/TLS configuration support

HTHou:
## Description

This PR generalizes IoTDB SSL/TLS handling around the existing thrift SSL switches and wires the selected SSL protocol through the client, service, REST, and consensus paths.

Changes include:
- Add a shared `RpcSslUtils` helper for TLS protocol normalization, thrift SSL transport parameter creation, `SSLContext`/key manager/trust manager creation, keystore/truststore type detection, and certificate validation.
- Add the shared `ssl_protocol` configuration with default `TLS`, load it from common and REST configs, and apply it to thrift SSL services, internal SSL clients, REST HTTPS, and Ratis certificate loading.
- Extend Java client entry points to carry SSL protocol settings through Session, TableSession, SessionPool, TableSessionPool, JDBC URL/properties parsing, CLI, and data/schema import/export tools.
- Keep cipher suites out of user-facing configuration so transports rely on the defaults selected by the configured SSL/TLS protocol and the active JSSE implementation.
- Reuse the shared SSL utility in Ratis and thrift service setup, update the Ratis dependency metadata, and clean up related REST dependency exclusions.
- Make the integration-test environment SSL-aware so EnvFactory-created JDBC connections, Sessions, TableSessions, SessionPools, and TableSessionPools work when thrift client SSL is enabled.
- Add `IoTDBClientSSLIT` to verify that non-SSL clients cannot connect to an SSL thrift port and that SSL Session, TableSession, and JDBC clients can connect and execute basic read/write flows.

## Validation

- `./mvnw spotless:apply -pl iotdb-client/service-rpc,iotdb-core/node-commons`
- `./mvnw -pl iotdb-client/service-rpc,iotdb-client/isession,iotdb-client/session,iotdb-client/jdbc,iotdb-client/cli,iotdb-core/node-commons -DskipTests compile`
- `./mvnw -pl iotdb-client/jdbc -Dtest=UtilsTest#testParseSslConfig test`
- `mvn verify -DskipUTs -Dit.test=IoTDBClientSSLIT -DfailIfNoTests=false -Dfailsafe.failIfNoSpecifiedTests=false -pl integration-test -am -P with-integration-tests`
- `git diff --check`


Caideyipi:
I think there are two correctness issues in this PR:

1. `ssl_protocol=TLS` is used as both the SSLContext algorithm and Jetty enabled protocol for REST HTTPS. In `RestService.configureSSL`, the value is passed to `sslContextFactory.setIncludeProtocols(protocol)`. The default template value is `TLS`, but JSSE enabled protocol names are normally `TLSv1.3`, `TLSv1.2`, etc., not `TLS`. So enabling REST HTTPS with the default config can leave Jetty with no matching enabled protocol / fail TLS handshakes. The context algorithm and enabled protocol list should be separate, or REST should not call `setIncludeProtocols` for the generic `TLS` value.

2. The legacy SSL overloads in `BaseRpcTransportFactory` now explicitly call the new overload with `null, null`, so internal synchronous SSL clients still do not use `ssl_protocol` / `ssl_provider_class`. For example `SyncDataNodeInternalServiceClient`, `SyncConfigNodeIServiceClient`, and IoT consensus sync clients call the old overload while `enable_internal_ssl` is true. If a custom JSSE provider is required, the server side may register it through `CommonDescriptor.configureRpcSsl()`, but these clients create their SSL transport without that provider and can fail to connect. These internal clients should pass `commonConfig.getSslProtocol()` and `commonConfig.getSslProviderClass()` to the new overload.

iotdb-client/cli/src/main/java/org/apache/iotdb/cli/AbstractCli.java
```diff
@@ -156,6 +160,8 @@ static void init() {
     keywordSet.add("-" + USE_SSL_ARGS);
     keywordSet.add("-" + TRUST_STORE_ARGS);
     keywordSet.add("-" + TRUST_STORE_PWD_ARGS);
+    keywordSet.add("-" + SSL_PROTOCOL_ARGS);
+    keywordSet.add("--" + SSL_PROTOCOL_ARGS);
```
HTHou: why using “--”

Caideyipi (approved):
I reviewed the latest head (a55227f). I do not see a functional issue in the implemented ssl_protocol support, so I am approving the implemented functionality.

Non-blocking follow-ups:
1. The PR description still mentions ssl_provider_class and optional JSSE provider registration, but the current code only implements ssl_protocol. Please either remove that claim from the PR text or add provider support.
2. The integration-test EnvFactory SSL helpers do not propagate ssl_protocol into client connection properties/builders, so ITs built through AbstractEnv would still use the default TLS protocol even when a non-default ssl_protocol is configured. This does not appear to affect production code, but it would be worth fixing to make future protocol-specific ITs meaningful.

I also rechecked the previously discussed REST default-protocol concern and the sync internal-client propagation path. Those look addressed in the current version.

HTHou:
Thanks for the review. I addressed both follow-ups:

- Updated the PR description to remove the stale `ssl_provider_class` / provider-registration wording and describe the current full PR scope.
- Added `ssl_protocol` propagation in the integration-test env helpers. When thrift client SSL is enabled, EnvFactory-created JDBC connections, Sessions, TableSessions, SessionPools, and TableSessionPools now receive the configured protocol as well.

Validation:
- `mvn verify -DskipUTs -Dit.test=IoTDBClientSSLIT -DfailIfNoTests=false -Dfailsafe.failIfNoSpecifiedTests=false -pl integration-test -am -P with-integration-tests`
- `git diff --check`

JackieTien97 (approved):

PR #17859, opened 2026-06-08
Upgrade minimum JDK to 17 and migrate JavaX to Jakarta

HTHou:
## Summary
- Raise the minimum Java version and build/runtime checks to JDK 17.
- Upgrade JDK/Jakarta-locked dependencies, including Jetty 12 EE10, Jersey 3, Logback 1.5, ANTLR 4.13, OpenAPI Generator 7, Swagger Jakarta artifacts, and the Jackson Jakarta RS provider.
- Migrate the REST service/OpenAPI generated surface and Jakarta EE runtime APIs to Jakarta namespaces.
- Suppress Thrift generated `@Generated` annotations because the current bundled Thrift 0.14.1 generator does not support `jakarta.annotation.Generated`.

## Notes
- Apache HttpClient usage in REST examples and integration tests is intentionally left unchanged.
- JSR-305 annotations such as `javax.annotation.Nonnull`, `javax.annotation.Nullable`, and `javax.annotation.concurrent` remain where used because they are static annotations and do not block the Jakarta runtime migration.
- The TsFile snapshot bump also adapts the pipe TsFile page readers to use the decoder accessors exposed by the upgraded TsFile reader APIs. This is not part of the Jakarta namespace migration and should be reviewed as a behavior-sensitive TsFile parsing change.

## Validation
- `mvn clean package -pl distribution -am -DskipTests`
- `mvn compile -pl external-service-impl/rest-openapi,external-service-impl/rest -DskipTests`
- `mvn clean compile -pl iotdb-protocol/thrift-commons,iotdb-protocol/thrift-datanode,iotdb-protocol/thrift-confignode,iotdb-protocol/thrift-ainode,iotdb-protocol/thrift-consensus -am -DskipTests`
- `mvn compile -pl iotdb-core/node-commons,iotdb-core/calc-commons,iotdb-core/confignode -DskipTests`
- `mvn spotless:apply -pl iotdb-core/calc-commons,iotdb-core/node-commons,iotdb-core/confignode,iotdb-core/datanode,external-service-impl/rest -DskipTests`
- `mvn spotless:apply -pl integration-test -P with-integration-tests -DskipTests`


Caideyipi:
I think this PR needs to regenerate dependencies.json after the REST/Jakarta dependency changes. dependencies.json is used by -Penable-sbom-check via src/main/groovy/checkDependencies.groovy, and the new 
est-openapi dependency tree now contains coordinates that are not in the reference list, for example:

- com.fasterxml.jackson.module:jackson-module-jakarta-xmlbind-annotations (while the old jackson-module-jaxb-annotations entry remains)
- io.github.classgraph:classgraph
- io.swagger.core.v3:swagger-integration-jakarta
- jakarta.inject:jakarta.inject-api
- com.sun.activation:jakarta.activation

Because the check fails on any dependency added by the current build but missing from dependencies.json, release/SBOM validation can fail even if normal compilation passes. Please regenerate dependencies.json from the transformed SBOM and re-check LICENSE-binary for the newly bundled REST/Jakarta dependencies as well.

CritasWang (commented):
This PR is described as a JDK 17 + JavaX→Jakarta migration, but it also carries a functional behavior change to TsFile time-column decoding that is unrelated to the migration:

- iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/event/common/tsfile/parser/scan/AlignedSinglePageWholeChunkReader.java:184
defaultTimeDecoder → getTimeDecoder(timeChunkHeader.getEncodingType())
- iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/event/common/tsfile/parser/scan/SinglePageWholeChunkReader.java:90
defaultTimeDecoder → getConfiguredTimeDecoder()

Both getTimeDecoder(...) and getConfiguredTimeDecoder() are new APIs that do not exist in the current IoTDB tree — they come from the upgraded TsFile parent class (AbstractChunkReader), which is why this change is coupled to the TsFile version bump in pom.xml:

- <tsfile.version>2.3.x-SNAPSHOT</tsfile.version>
+ <tsfile.version>2.3.2-260608-SNAPSHOT</tsfile.version>

This looks like a genuine fix (the time decoder was previously hard-coded to the default instead of honoring the actual chunk encoding type), but it has real runtime behavior implications for the pipe TsFile-parsing path and deserves its own review, commit message, and test coverage rather than being hidden inside a 262-file migration diff.

Request: please either
1. if it must stay, call it out explicitly in the PR description and add/point to tests that exercise the changed decoding path.

HTHou:
Thanks @CritasWang, good catch. You are right that this is not a pure namespace migration change.

This adaptation is required by the TsFile snapshot bump that is part of the JDK 17 upgrade path: without updating these call sites to the decoder accessors exposed by the upgraded TsFile reader APIs, this branch cannot compile/pass CI after the TsFile upgrade. Splitting it out into a standalone PR would also be difficult because that standalone change depends on the same TsFile/JDK17 upgrade context.

I have updated the PR description to call this out explicitly. I will keep this change in this PR, but will make sure it is reviewed as a behavior-sensitive TsFile parsing adaptation and add/point to dedicated coverage for the affected time-decoding path.

external-service-impl/rest/src/main/java/org/apache/iotdb/rest/RestService.java
```diff
@@ -96,7 +96,7 @@ private ServletContextHandler constructServletContextHandler() {
     holder.setInitOrder(1);
     holder.setInitParameter(
         "jersey.config.server.provider.packages",
-        "io.swagger.jaxrs.listing, io.swagger.sample.resource, org.apache.iotdb.rest.protocol");
+        "io.swagger.v3.jaxrs2.integration.resources, org.apache.iotdb.rest.protocol");
```
JackieTien97, line 99: This switches the registered Swagger package to Swagger Core 2.x, whose resources expose `/openapi` and `/openapi.{json|yaml}` rather than the old `/swagger.json` path. The REST authorization filter still only special-cases `swagger.json`, so `enable_swagger=true` no longer makes the generated spec reachable without credentials, while `enable_swagger=false` also does not hide the actual `/openapi.json` endpoint from authenticated callers. Please update the filter/config/docs to control the new OpenAPI paths, or keep a compatible `/swagger.json` route if that path is still expected by users.

LICENSE-binary
```diff
@@ -224,12 +224,12 @@ com.fasterxml.jackson.core:jackson-databind:2.16.2
...
-org.eclipse.jetty:jetty-io:9.4.58.v20250814
-org.eclipse.jetty:jetty-security:9.4.58.v20250814
-org.eclipse.jetty:jetty-server:9.4.58.v20250814
-org.eclipse.jetty:jetty-servlet:9.4.58.v20250814
-org.eclipse.jetty:jetty-util:9.4.58.v20250814
+org.eclipse.jetty:jetty-http:12.0.36
+org.eclipse.jetty:jetty-io:12.0.36
+org.eclipse.jetty:jetty-security:12.0.36
+org.eclipse.jetty:jetty-server:12.0.36
+org.eclipse.jetty.ee10:jetty-ee10-servlet:12.0.36
```
JackieTien97, line 240: The REST fat jar now bundles additional Jakarta/OpenAPI dependencies that are not reflected here. For example, `mvn package -pl external-service-impl/rest-openapi,external-service-impl/rest -DskipTests` produces `rest-...-jar-with-dependencies.jar` containing `org.eclipse.jetty:jetty-session:12.0.36`, the `io.swagger.core.v3:swagger-*-jakarta:2.2.50` artifacts, `io.github.classgraph:classgraph:4.8.184`, `com.fasterxml.jackson.jakarta.rs:*`, `jakarta.inject:jakarta.inject-api:2.0.1`, and `jakarta.xml.bind:jakarta.xml.bind-api:3.0.1`. `dependencies.json` already lists some of these, such as `jetty-session`, but `LICENSE-binary` still omits them and still lists the old `jakarta.inject:jakarta.inject:2.6.1` coordinate. Please regenerate/recheck `LICENSE-binary` against the packaged external-service-impl artifact so the binary release license metadata matches what is actually bundled.

JackieTien97 (changes requested):
Thanks for the JDK 17/Jakarta migration work. I found two release/runtime issues that should be addressed before merging.

scripts/sbin/windows/start-datanode.bat
```diff
@@ -48,15 +48,11 @@ for /f tokens^=2-5^ delims^=.-_+^" %%j in ('java -fullversion 2^>^&1') do (
...
-)
-IF "%JAVA_VERSION%" == "7" (
-	echo IoTDB only supports jdk >= 8, please check your java version.
-	goto finally
-)
+	@REM IoTDB requires JDK 17 or later.
+	IF %JAVA_VERSION% LSS 17 (
+		echo IoTDB only supports jdk >= 17, please check your java version.
+		goto finally
+	)
```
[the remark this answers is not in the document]
HTHou: Addressed in be178c3bc9. The DataNode Windows startup script now checks whether `JAVA_VERSION` is empty before the numeric `LSS` comparison and exits through `finally` with a clearer error message.

scripts/sbin/windows/start-confignode.bat
```diff
@@ -46,15 +46,11 @@ for /f tokens^=2-5^ delims^=.-_+^" %%j in ('java -fullversion 2^>^&1') do (
...
-)
-IF "%JAVA_VERSION%" == "7" (
-	echo IoTDB only supports jdk >= 8, please check your java version.
-	goto finally
-)
+	@REM IoTDB requires JDK 17 or later.
+	IF %JAVA_VERSION% LSS 17 (
+		echo IoTDB only supports jdk >= 17, please check your java version.
+		goto finally
+	)
```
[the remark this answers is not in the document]
HTHou: Addressed in be178c3bc9. The ConfigNode Windows startup script now has the same empty `JAVA_VERSION` guard before the `LSS` comparison, so it fails gracefully if the Java version cannot be parsed.

LICENSE-binary
```diff
@@ -224,12 +224,12 @@ com.fasterxml.jackson.core:jackson-databind:2.16.2
...
+org.eclipse.jetty:jetty-util:12.0.36
```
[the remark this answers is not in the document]
HTHou: Addressed in be178c3bc9. `org.eclipse.jetty:jetty-session:12.0.36` has been added to `LICENSE-binary` alongside the other Jetty 12 artifacts.

external-service-impl/rest/src/main/java/org/apache/iotdb/rest/RestService.java
HTHou: Addressed in be178c3bc9. I updated `AuthorizationFilter` to control the Swagger Core 2.x OpenAPI resources (`/openapi`, `/openapi.json`, and `/openapi.yaml`) with `enable_swagger`, and updated the config comments/docs to point to `/openapi.json` instead of the old `/swagger.json` path.

LICENSE-binary
HTHou: Addressed in be178c3bc9. `LICENSE-binary` now includes the REST fat-jar additions: `jetty-session`, the Swagger Jakarta artifacts, Jackson Jakarta RS/XMLBind modules, `classgraph`, `jakarta.inject-api`, `jakarta.activation-api`, and `jakarta.xml.bind-api`; it also replaces the stale `jakarta.inject:jakarta.inject:2.6.1` coordinate with `jakarta.inject:jakarta.inject-api:2.0.1`.

HTHou:
@Caideyipi Addressed in be178c3bc9. I removed the duplicate `jakarta.xml.bind:jakarta.xml.bind-api` entry from `dependencies.json` and rechecked the REST/Jakarta coordinates raised here: the Swagger/Jackson/ClassGraph/Jakarta Inject entries are represented, while `com.sun.activation:jakarta.activation` is excluded from the REST fat jar to avoid duplicating Jakarta Activation classes. `LICENSE-binary` has also been updated for the newly bundled REST/OpenAPI/Jakarta dependencies.

JackieTien97 (approved):

PR #17879, opened 2026-06-09
Fix partial insert handling for null measurements

Caideyipi:
Summary:
- Skip failed/null measurements and null tablet columns consistently in insert serialization, split, WAL, memtable, memory estimation, and pipe tablet parsing/conversion paths.
- Keep null row values during pipe row type transfer and avoid composing last-cache values for null retained row/tablet values.
- Add regression coverage for partial insert, null column/value, table non-FIELD columns, pipe parser/converter, and memory estimation paths.

Tests:
- mvn -pl iotdb-core/datanode spotless:apply
- git diff --check
- Narrow datanode test lifecycle could not complete locally after recreating the branch on master: single-module compile first hit missing generated upstream classes, then -am compile rebuilt upstream modules but failed on Windows pagefile/native memory while compiling thrift-datanode. Before recreating the branch, the same narrow Surefire regression suite passed with 99 tests.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/write/InsertNode.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/write/InsertNode.java
@@ -174,6 +174,7 @@ public abstract class InsertNode extends SearchNode {
 
   public void setMeasurementSchemas(MeasurementSchema[] measurementSchemas) {
     this.measurementSchemas = measurementSchemas;
+    measurementColumnCnt = -1;
   }
 
   public String[] getMeasurements() {
@@ -189,13 +190,16 @@ public abstract class InsertNode extends SearchNode {
   }
 
   public boolean isValidMeasurement(int i) {
-    return measurementSchemas != null
+    return measurements != null
+        && measurements[i] != null
+        && measurementSchemas != null
         && measurementSchemas[i] != null
         && (columnCategories == null || columnCategories[i] == TsTableColumnCategory.FIELD);
   }
 
   public void setMeasurements(String[] measurements) {
     this.measurements = measurements;
+    measurementColumnCnt = -1;
   }
 
   public TSDataType[] getDataTypes() {
@@ -327,8 +331,12 @@ public abstract class InsertNode extends SearchNode {
   }
 
   public boolean hasValidMeasurements() {
-    for (Object o : measurements) {
-      if (o != null) {
+    if (measurements == null) {
+      return false;
+    }
+    for (int i = 0; i < measurements.length; i++) {
+      if (measurements[i] != null
+          && (columnCategories == null || columnCategories[i] == TsTableColumnCategory.FIELD)) {
         return true;
       }
     }
@@ -354,15 +362,16 @@ public abstract class InsertNode extends SearchNode {
   }
 
   public boolean isMeasurementFailed(int index) {
-    return measurements[index] == null;
+    return measurements == null || measurements[index] == null;
+  }
+
+  protected boolean isWritableFieldMeasurement(int index) {
+    return !isMeasurementFailed(index)
+        && (columnCategories == null || columnCategories[index] == TsTableColumnCategory.FIELD);
   }
 
   public boolean allMeasurementFailed() {
-    if (measurements != null) {
-      return failedMeasurementNumber
-          >= measurements.length - (tagColumnIndices == null ? 0 : tagColumnIndices.size());
-    }
-    return true;
+    return measurements == null || !hasValidMeasurements();
   }
 
   // endregion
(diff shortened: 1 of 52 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/write/InsertTabletNode.java
```diff
@@ -438,6 +438,16 @@ protected Object[] initTabletValues(int columnSize, int rowSize, TSDataType[] da
...
 
+  protected Object[] initTabletValuesForSplit(int columnSize, int rowSize, TSDataType[] dataTypes) {
+    Object[] values = initTabletValues(columnSize, rowSize, dataTypes);
+    for (int i = 0; i < values.length; i++) {
+      if (!hasColumnForSplit(i)) {
+        values[i] = null;
+      }
+    }
+    return values;
+  }
```
jt2594838, line 451: May init nulls directly in `initTabletValues`, avoid creating unnecessary objects

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/write/InsertNode.java
```diff
@@ -189,13 +190,16 @@ public int measureColumnCnt() {
   }
 
   public boolean isValidMeasurement(int i) {
-    return measurementSchemas != null
+    return measurements != null
```
jt2594838, line 197: How could measurements be null?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/memtable/AbstractMemTable.java
```diff
@@ -341,6 +337,38 @@ private static int computeTabletNullPointsNumber(
...
+  private static boolean isValidMeasurement(
+      InsertNode insertNode, int index, boolean countFieldOnly) {
+    return insertNode.getMeasurements()[index] != null
+        && (!countFieldOnly || isFieldMeasurement(insertNode, index));
+  }
+
+  private static boolean isFieldMeasurement(InsertNode insertNode, int index) {
+    return insertNode.getColumnCategories() == null
+        || insertNode.getColumnCategories()[index] == TsTableColumnCategory.FIELD;
+  }
```
jt2594838: May encapsulate these methods into InsertNode to reduce redundancy and improve clarity.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/MemUtils.java
```diff
@@ -67,14 +67,18 @@ public static long getRecordSize(TSDataType dataType, Object value, boolean addi
...
-      if (value[i] == null) {
-        emptyRecordCount++;
+      if (value[i] == null || !isFieldColumn(columnCategories, i)) {
         continue;
       }
-      memSize += getRecordSize(dataTypes.get(i - emptyRecordCount), value[i], false);
+      memSize += getRecordSize(dataTypes.get(dataTypeIndex++), value[i], false);
     }
     return memSize;
   }
```
jt2594838, line 88: dataTypes is null-free?
If you enforce some restrictions on it now, you should add some comment to mention it.

jt2594838 (commented):
Use a unified entry point to make null-related judgments instead of scattering them around.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/memtable/AbstractMemTable.java
Caideyipi: Done. Moved isValidMeasurement, isFieldMeasurement, and valid measurement counting into InsertNode/InsertRowNode/InsertTabletNode. AbstractMemTable now delegates to those helpers.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/write/InsertNode.java
Caideyipi: Good point. Normal insert paths should provide measurements; the null guard here is defensive for partially rebuilt/deserialized nodes and existing null-safe behavior when all measurements are cleared. I kept the check defensive, and centralized the real validity rules in InsertNode.isValidMeasurement(index, countFieldOnly).

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/node/write/InsertTabletNode.java
Caideyipi: Done. I changed the split path so initTabletValues delegates to a helper with a split flag and only allocates arrays for columns that pass hasColumnForSplit.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/MemUtils.java
Caideyipi: Done. Added comments in MemUtils clarifying that dataTypes contains only non-null FIELD types and is ordered with the non-null FIELD values.

jt2594838 (approved):

jt2594838 (approved):

PR #17893, opened 2026-06-10
Harden timestamp arithmetic and boundary handling across IoTDB

Caideyipi:
### Description

This PR hardens timestamp arithmetic and boundary handling across storage, query, pipe, UDF, and table-function paths.

The scope includes:

- Origin-aware, overflow-safe time-partition ID and boundary derivation, plus TsFile splitter ordering at `Long.MIN_VALUE`/`Long.MAX_VALUE`.
- Primitive `long` fast paths with saturating arithmetic for `date_bin`, `TUMBLE`/`HOP`/`CUMULATE`/`SESSION` window TVFs, sliding windows, aggregation iterators, TTL/progress calculations, and `LIMIT`/`OFFSET` push-down.
- Exact integer timestamp differences for linear fill, SDT/downsampling, and outlier sampling.
- Exclusive real-row gap-fill boundaries and forecast validation for single-row input/output timestamps.
- Localized (English/Chinese) `SemanticException` messages for duration, date, range, and inference errors.
- Related auth login-lock, pipe processor/realtime-source, downsampling, UDF/TVF, AI inference, planner, and storage fixes.
- TsFile filename version validation and adjacent time-range normalization.

### Tests

Passed:

- `mvn -pl iotdb-core/node-commons '-Dtest=TimePartitionUtilsTest,CommonDateTimeUtilsTest,ForecastTimeUtilsTest' '-Dsurefire.failIfNoSpecifiedTests=false' test` (24 tests).
- `mvn -pl iotdb-core/node-commons '-Dtest=TimeWindowTableFunctionBoundaryTest' '-Dsurefire.failIfNoSpecifiedTests=false' test` (5 tests).
- `mvn -pl iotdb-core/datanode '-Dtest=DateTimeUtilsTest,QueryTimePartitionTest,TsFileResourceTest,SlidingTimeWindowEvaluationHandlerTest' '-Dsurefire.failIfNoSpecifiedTests=false' test` (57 tests, 1 skipped).
- `mvn -pl iotdb-core/datanode '-Dtest=TumblingWindowingProcessorTest,DataRegionWatermarkInjectorTest,InferenceOperatorTest,TimeRangeIteratorTest,AnalyzeUtilsTest,TsFileSplitterTest' '-Dsurefire.failIfNoSpecifiedTests=false' test` (19 tests).
- English and Chinese `test-compile` for node-commons, calc-commons, and datanode.
- `git diff --check`.

The full reactor check is blocked in this local environment before the changed modules by generated Thrift source `TaskType.java` requiring `javax.annotation.Generated`; the module-level English/Chinese compiles and tests above complete successfully.


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/utils/TimePartitionUtils.java
```diff
@@ -142,40 +173,47 @@ public static boolean satisfyPartitionStartTime(Filter timeFilter, long partitio
...
+
+  private static long getTimePartitionStartTime(long partitionId) {
+    BigInteger partitionStartTime =
+        BigInteger.valueOf(partitionId)
+            .multiply(BigInteger.valueOf(timePartitionInterval))
+            .add(BigInteger.valueOf(timePartitionOrigin));
+    if (partitionStartTime.compareTo(BigInteger.valueOf(Long.MIN_VALUE)) < 0) {
+      return Long.MIN_VALUE;
+    }
+    if (partitionStartTime.compareTo(BigInteger.valueOf(Long.MAX_VALUE)) > 0) {
```
JackieTien97: [P1] Keep the partition id exact until the boundary slot is resolved

`getTimePartitionIdWithoutOverflow()` still narrows the `BigInteger` with `longValue()`. With the valid configuration `time_partition_origin=1` and `time_partition_interval=1`, the exact partition id for `Long.MIN_VALUE` is `-9223372036854775809`; it wraps to `Long.MAX_VALUE`, reaches this branch, and `getTimePartitionSlot(Long.MIN_VALUE)` returns a slot starting at `Long.MAX_VALUE`. I reproduced this with a JUnit assertion: expected `Long.MIN_VALUE`, actual `Long.MAX_VALUE`. Please handle or clamp the `BigInteger` id before narrowing and add this configuration as a regression test.

iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/execution/operator/process/fill/linear/LinearFill.java
```diff
@@ -157,9 +157,8 @@ private boolean fill(
   }
 
   private double getFactor(long currentTime) {
-    return nextTimeInCurrentColumn - previousTime == 0
-        ? 0.0
-        : ((double) (currentTime - previousTime)) / (nextTimeInCurrentColumn - previousTime);
+    double timeRange = (double) nextTimeInCurrentColumn - (double) previousTime;
```
JackieTien97: [P1] Subtract before converting to `double`

Casting each timestamp separately loses small deltas at normal nanosecond-epoch magnitudes. For `previous=2^60`, `current=previous+50`, and `next=previous+100`, both differences become `0.0`, so this returns factor `0.0` instead of `0.5`. The same pattern was added in `DownSamplingTimeUtils.timeDifferenceAsDouble()` and `UDTFEqualSizeBucketOutlierSample`, where it can also produce zero denominators or invalid slopes. Please compute the exact integer difference first, using `Math.subtractExact` with a `BigInteger` fallback for the overflow case, and then convert the delta to `double`.

iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/execution/operator/process/gapfill/AbstractGapFillOperator.java
```diff
@@ -127,7 +128,9 @@ public TsBlock next() throws Exception {
       // -1 because we should not include current row, current row will be appended in
       // writeCurrentRow
       long currentEndTime =
-          timeColumn.isNull(i) ? endTime : block.getColumn(timeColumnIndex).getLong(i) - 1;
+          timeColumn.isNull(i)
+              ? endTime
+              : saturatingAdd(block.getColumn(timeColumnIndex).getLong(i), -1);
```
JackieTien97: [P2] Do not saturate an exclusive lower-bound conversion

When both the gap-fill start and the first real row are `Long.MIN_VALUE`, this produces `currentEndTime == Long.MIN_VALUE`. `fillGaps()` then emits a synthetic row at that timestamp, `writeCurrentRow()` emits the real row at the same timestamp, and `nextTime()` is advanced twice, skipping the next bucket. Please express this as an exclusive comparison (`currentTime < rowTime`) rather than converting it to an inclusive bound with `rowTime - 1`.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/queryengine/plan/udf/ForecastTimeUtils.java
```diff
@@ -0,0 +1,74 @@
...
+    // Utility class.
+  }
+
+  public static long calculateForecastInterval(
+      long inputStartTime, long inputEndTime, int inputSize, long outputInterval) {
+    if (outputInterval > 0) {
+      return outputInterval;
+    }
+    if (inputSize <= 1) {
+      return 0;
```
JackieTien97: [P2] Reject an uninferable interval for a single input row

For the tree-model `UDTFForecast`, a single input row with the default interval now gets `interval == 0`; the default start time equals the input end time and every generated output timestamp repeats that input timestamp. The table-model forecast explicitly rejects `outputTime <= inputEndTime`, but the tree UDTF has no equivalent check, so the two APIs diverge and this path silently returns non-future timestamps. Please require an explicit positive interval for a single-row input or share the table-model validation.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/queryengine/plan/udf/ForecastTimeUtils.java
```diff
@@ -0,0 +1,74 @@
...
+
+import org.apache.iotdb.commons.exception.SemanticException;
+
+import java.math.BigInteger;
+
+public final class ForecastTimeUtils {
+
+  private static final BigInteger BIG_LONG_MAX = BigInteger.valueOf(Long.MAX_VALUE);
+  private static final String FORECAST_OUTPUT_TIME_OUT_OF_RANGE =
+      "Forecast output time is out of range.";
```
JackieTien97: [P2] Route new user-visible messages through the i18n message classes

This exception text remains English under `-P with-zh-locale`. The same issue appears in the newly added messages in `InferenceOperator`, `AnalyzeUtils`, `ASTVisitor`, `AstBuilder`, `ExpressionFactory`, and `TsFileOverlapValidationAndRepairTool`. Please add the full templates to the appropriate `*Messages` classes in both `src/main/i18n/en` and `src/main/i18n/zh` and reference those constants here.

JackieTien97 (changes requested):
I found correctness regressions in boundary partition routing, nanosecond arithmetic, gap-fill boundary handling, and single-row forecast timestamp generation. I also found several new user-visible messages that bypass the repository's i18n message classes. The existing targeted tests pass, but the partition-routing regression is reproducible with a focused assertion. Please address the inline comments before merging.

iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/transformation/dag/column/unary/scalar/DateBinFunctionColumnTransformer.java
```diff
@@ -201,7 +210,33 @@ public static long nextDateBin(int monthDuration, ZoneId zoneId, long currentTim
...
   public static long nextDateBin(long nonMonthDuration, long currentTime) {
-    return currentTime + nonMonthDuration;
+    return saturatingAdd(currentTime, nonMonthDuration);
+  }
+
+  public static long saturatingAdd(long left, long right) {
+    return saturateToLong(BigInteger.valueOf(left).add(BigInteger.valueOf(right)));
+  }
+
+  private static BigInteger getNonMonthDateBinStart(
```
JackieTien97: [P1] Keep a `long` fast path here; this is a per-row hot path

`dateBin()` is invoked from `doTransform()` for every input row of `date_bin()` and every row of a table-model `GROUP BY date_bin(...)`, and `saturatingAdd()` above is also called per row from `AbstractGapFillOperator`. Every call now allocates four to six `BigInteger` instances plus a `divideAndRemainder` even though overflow can only happen when `source - origin` or `origin + n * duration` leaves the `long` range. Please keep the original `long` arithmetic and only fall back to `BigInteger` when the subtraction or the final addition actually overflows (Guava `LongMath.checkedSubtract` / `checkedAdd` in a try/catch, or an explicit sign-bit check), and make `saturatingAdd()` a plain `LongMath.saturatedAdd()` call.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/udf/builtin/relational/tvf/WindowTVFUtils.java
```diff
@@ -92,6 +95,31 @@ public static int findColumnIndex(
             expectedFieldName));
   }
 
+  static BigInteger getWindowStart(long source, long origin, long interval) {
```
JackieTien97: [P1] Same hot-path concern for the window TVFs

`getWindowStart()` is called once per input row by `TUMBLE`, `HOP` and `CUMULATE`, and `HOPTableFunction` / `CumulateTableFunction` now also run their inner loops on `BigInteger`. `SessionTableFunction` allocates two `BigInteger`s per row for a single comparison. All inputs here are already `>= origin`, so `timeValue - origin` cannot overflow; only `origin + stepCount * interval` and the `+ size` for the window end can, and those can be handled with `LongMath.saturatedAdd` / `saturatedMultiply` without leaving `long`. Please keep the row loop on primitives and saturate only the two additions that can actually overflow.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/utils/TimePartitionUtils.java
```diff
@@ -99,21 +101,50 @@ public static long getTimePartitionUpperBound(long time) {
...
+    }
+    return time == Long.MAX_VALUE && getTimePartitionLowerBound(time) != timePartitionStartTime;
+  }
+
+  public static boolean isTimePartitionStartTime(long time) {
+    return getTimePartitionLowerBound(time) == time;
   }
 
   public static long getTimePartitionId(long time) {
+    if (originMayCauseOverflow) {
```
JackieTien97: [P1] This puts every write with a non-zero origin onto the `BigInteger` path

`getTimePartitionId()` is called per row on the insert path (`DataRegion.insert`, `insertTablet`, `insertRows`, `getOrCreateTsFileProcessor`, and the `ArrayDeviceTimeIndex` / `FileTimeIndex` partition checks). With `time_partition_origin != 0` this now unconditionally allocates three `BigInteger`s and does a `remainder` plus `divide` per call, which is a measurable write-throughput regression for any deployment that sets an origin. The overflow only occurs when `time - timePartitionOrigin` leaves the `long` range, which is cheap to detect with `LongMath.checkedSubtract` or `((time ^ origin) & (time ^ diff)) < 0`. Please keep the `long` path as the default for both the zero-origin and non-zero-origin cases and fall back to `getTimePartitionIdWithoutOverflow()` only when that subtraction overflows.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/windowing/handler/SlidingTimeWindowEvaluationHandler.java
```diff
@@ -25,54 +25,56 @@
...
 
+import java.math.BigInteger;
 import java.util.LinkedList;
 import java.util.Queue;
 
 public class SlidingTimeWindowEvaluationHandler extends SlidingWindowEvaluationHandler {
 
-  private final long timeInterval;
-  private final long slidingStep;
+  private final BigInteger timeInterval;
```
JackieTien97: [P2] Avoid promoting the whole handler state to `BigInteger`

The handler is fed one point at a time, so this now does a `BigInteger.valueOf(timestamp)` plus one or two `compareTo` / `add` allocations per collected point just to guard the two `+ timeInterval` / `+ slidingStep` additions. Since the window boundaries only ever move forward, `LongMath.saturatedAdd` on `long` fields gives the same behaviour: once `nextWindowBeginTime` saturates at `Long.MAX_VALUE` the `while (nextWindowBeginTime <= timestamp)` loop terminates exactly as it does with the `BigInteger` version. Please keep the fields as `long`.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/aggregation/timerangeiterator/ITimeRangeIterator.java
```diff
@@ -47,8 +49,60 @@ public interface ITimeRangeIterator {
 
   default TimeRange getFinalTimeRange(TimeRange timeRange, boolean leftCRightO) {
     return leftCRightO
-        ? new TimeRange(timeRange.getMin(), timeRange.getMax() - 1)
-        : new TimeRange(timeRange.getMin() + 1, timeRange.getMax());
+        ? new TimeRange(timeRange.getMin(), saturatingAdd(timeRange.getMax(), -1))
+        : new TimeRange(saturatingAdd(timeRange.getMin(), 1), timeRange.getMax());
+  }
+
+  static long saturatingAdd(long left, long right) {
```
JackieTien97: [P2] Consolidate the saturating helpers instead of adding a copy per class

This PR adds a private `saturatingAdd` / `saturateToLong` (and in `UDTFTimeDifference` a `saturatingSubtract`) to `DateBinFunctionColumnTransformer`, `QueryTimeoutRuntimeException`, `TumblingWindowingProcessor`, `DataRegionWatermarkInjector`, `ITimeRangeIterator`, `IntermediateLayer`, `LimitOffsetPushDown`, `SeriesPartitionTable`, `WindowTVFUtils` and `UDTFTimeDifference`, using three different techniques (sign-bit XOR, `Long.MAX_VALUE - right` comparison, and `BigInteger` round-trip). `calc-commons`, `node-commons` and `datanode` all already depend on Guava 32, so `com.google.common.math.LongMath.saturatedAdd` / `saturatedSubtract` / `saturatedMultiply` covers every one of these call sites. If a project-owned helper is preferred, please put a single one in `node-commons` and have the other modules import it.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/DataNodeDateTimeUtils.java
```diff
@@ -66,39 +66,39 @@ public static long convertDurationStrToLong(
     switch (durationUnit) {
       case y:
       case year:
-        res *= 365 * 86_400_000L;
+        res = Math.multiplyExact(value, 365 * 86_400_000L);
```
JackieTien97, line 81: [P2] `ArithmeticException` now escapes to the client as INTERNAL_SERVER_ERROR on most SQL paths

Only the two date-expression call sites in `ASTVisitor` / `AstBuilder` wrap these `Math.*Exact` calls in a try/catch. The remaining callers, `GROUP BY ([...], <interval>[, <step>])` via `constructTimeDuration`, `FILL(... , <interval>)`, `GROUP BY SESSION`, CQ `EVERY` / boundary offsets, `IoTDBDescriptor` and `ValueConverter`, do not, so an oversized literal such as `group by ([1, 10), 178956971y)` now surfaces as a raw `ArithmeticException` that `ErrorHandlingUtils` maps to `INTERNAL_SERVER_ERROR` with an error-level stack trace, instead of the `SemanticException` a user would expect for bad input. Please catch the overflow inside `DataNodeDateTimeUtils` and rethrow a `SemanticException` (through the i18n message classes), then update `DateTimeUtilsTest.convertDurationStrToLongOverflowTest` / `testConstructTimeDurationOverflow`, which currently assert `ArithmeticException`.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/analyze/AnalyzeVisitor.java
```diff
@@ -2325,13 +2345,29 @@ public static Pair<List<TTimePartitionSlot>, Pair<Boolean, Boolean>> getTimePart
...
+        || isRangeCoveringRightUnclosedPartitions(minTime, maxTime, needRightAll)) {
       return;
     }
     long size = TimePartitionUtils.getEstimateTimePartitionSize(minTime, maxTime);
-    context.reserveMemoryForFrontEnd(
-        RamUsageEstimator.shallowSizeOfInstance(TTimePartitionSlot.class) * size);
+    context.reserveMemoryForFrontEnd(estimateTimePartitionSlotMemory(size));
+  }
+
+  private static boolean isRangeCoveringRightUnclosedPartitions(
```
JackieTien97: [P2] Replace the `Long.MAX_VALUE - 1` special case with a general adjacency merge

I traced why this only triggers for `maxTime == Long.MAX_VALUE - 1`: the new `getGroupByTimePredicate()` produces `groupByTime(...) OR time = Long.MAX_VALUE`, and tsfile's `Or.getTimeRanges()` only merges overlapping ranges, not adjacent ones, so the analyzer receives `[start, MAX - 1]` followed by `[MAX, MAX]`. The resulting slot list and `needRightAll` flag are correct for that shape (I checked against `SeriesPartitionTable.getDataPartition`), but the helper is keyed on one magic value and will silently stop applying if the predicate shape changes or if another `OR` produces `[a, b], [b + 1, c]` pairs. Please normalise `timeRangeList` before the loop by merging adjacent ranges (`next.getMin() == cur.getMax() + 1`, guarding the `Long.MAX_VALUE` case), which removes the special case and also fixes the `reserveMemoryForTimePartitionSlot` over-estimation for such inputs. At minimum, a comment explaining where the `MAX - 1` range comes from would help the next reader.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/tsfile/generator/TsFileNameGenerator.java
```diff
@@ -415,7 +415,7 @@ public static TsFileResource getSettleCompactionTargetFileResources(
   }
 
   public static class TsFileName {
-    private static final String FILE_NAME_PATTERN = "(\\d+)-(\\d+)-(\\d+)-(\\d+).tsfile$";
+    private static final String FILE_NAME_PATTERN = "(-?\\d+)-(-?\\d+)-(\\d+)-(\\d+)\\.tsfile$";
```
JackieTien97: [P2] Do not accept a negative version in the file-name pattern

The `version` component is a monotonically increasing counter that is never negative in a real file name, so the second `-?` only exists to let the synthetic `generateNewTsFileName(0, Long.MIN_VALUE, 0, 0)` names in `TsFileResourceTest` parse. Loosening the on-disk naming contract for a test fixture is not worth it; please keep `version` as `(\d+)` and test `checkAndCompareFileName` / `compareFileCreationOrderByDesc` with large positive versions instead. The `-?` on `time` is defensible if negative timestamps are actually expected in file names, but it would be good to say so in a comment.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/parser/ASTVisitor.java
```diff
@@ -3188,10 +3189,16 @@ private TimeRange parseTimeRangeForDeleteTimeRange(
     long time = Long.parseLong(((ConstantOperand) valueExpression).getValueString());
     switch (expressionType) {
       case LESS_THAN:
+        if (time == Long.MIN_VALUE) {
+          throw new SemanticException(DELETE_RANGE_COMPARISON_ERROR_MSG);
```
JackieTien97: [P3] `DELETE_RANGE_COMPARISON_ERROR_MSG` does not describe this failure

That message tells the user the left bound of a range comparison must not exceed the right bound and to write `time > 5 and time < 10`. Here there is a single comparison (`time < Long.MIN_VALUE` / `time > Long.MAX_VALUE`) that simply selects nothing. The tree-model `SHOW REGIONID` branch and `AnalyzeUtils` in this PR already use a dedicated "does not select any time range" message for the identical situation; please reuse that constant here (via the i18n classes) so the two delete paths report the same error.

JackieTien97 (changes requested):
Follow-up review, complementary to the earlier one (the i18n point is already covered there, so it is not repeated). The `TimePartitionUtils` boundary derivation itself checks out: `timePartitionUpperBoundWithoutOverflow` is always the start of the last slot, `getTimePartitionEndTime` / `isAfterOrEqualToTimePartitionUpperBound` distinguish a slot whose upper bound is exactly `Long.MAX_VALUE` from the overflowing last slot, and the fast path and the `BigInteger` path agree whenever no overflow occurs, so existing data keeps its partition assignment. The `LinkedHashMap` change in `TsFileSplitter` is also a real fix since `AlignedChunkData` relies on identity hashing.

Remaining concerns are inline: (1) several per-row hot paths were rewritten on top of `BigInteger` even though overflow is only possible at extreme timestamps, (2) about ten private `saturatingAdd` / `saturateToLong` copies with three different implementations were added while Guava `LongMath` is already on the classpath of every touched module, (3) `Math.*Exact` in `DataNodeDateTimeUtils` leaks `ArithmeticException` to users on most SQL paths, plus a few smaller points.

The PR description is also out of date: the Tests section still says the module could not be compiled locally, and neither the title nor the summary mentions the auth, pipe processor, UDF/TVF, AI inference, and LIMIT/OFFSET push-down changes, which make up most of the 98 files. Please refresh it (or split the PR) so reviewers can see the actual scope.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/utils/TimePartitionUtils.java
Caideyipi: Fixed in 63608c37842. The exact partition id now stays as BigInteger while the boundary state is derived, and it is clamped before any long narrowing. The clamped boundary IDs are also handled explicitly when their mathematical start is outside the timestamp range. Added a regression test for `origin = 1`, `interval = 1`, and `Long.MIN_VALUE`.

iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/execution/operator/process/fill/linear/LinearFill.java
Caideyipi: Fixed in 63608c37842. Added the shared `CommonDateTimeUtils.timeDifferenceAsDouble()`, which uses `Math.subtractExact` and falls back to BigInteger only when the integer subtraction overflows. Linear fill, SDT/downsampling, and equal-size-bucket outlier sampling now use it. Added coverage for small deltas around `2^60` and for the full long-range overflow case.

iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/execution/operator/process/gapfill/AbstractGapFillOperator.java
Caideyipi: Fixed in 63608c37842. Real-row timestamps are now passed to gap filling as an exclusive bound, so there is no `rowTime - 1` conversion. Synthetic/end-of-range filling remains inclusive, and the loop also stops if advancing the time iterator no longer makes progress.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/queryengine/plan/udf/ForecastTimeUtils.java
Caideyipi: Fixed in 63608c37842. The shared forecast interval helper now rejects a single-row input unless an explicit positive output interval is supplied, so tree-model and table-model forecast paths no longer generate repeated non-future timestamps. Added focused regression tests.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/queryengine/plan/udf/ForecastTimeUtils.java
Caideyipi: Fixed in 63608c37842. The newly introduced forecast, inference, analyzer/parser, expression, duration/date, and TsFile repair messages are now routed through the appropriate message classes in both English and Chinese. The touched modules compile successfully with `-Pwith-zh-locale`.

iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/transformation/dag/column/unary/scalar/DateBinFunctionColumnTransformer.java
Caideyipi: Fixed in 63608c37842. Non-month `date_bin` now stays on primitive long arithmetic (`subtractExact`, `floorDiv`, `multiplyExact`, and `addExact`) and only falls back to BigInteger on an actual overflow. Boundary additions use Guava `LongMath.saturatedAdd`.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/udf/builtin/relational/tvf/WindowTVFUtils.java
Caideyipi: Fixed in 63608c37842. TUMBLE, HOP, CUMULATE, and SESSION row processing is back on primitive long values. Window ends and loop arithmetic use Guava saturated operations; the window-start helper only uses BigInteger as an exceptional overflow fallback. The boundary TVF tests pass.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/utils/TimePartitionUtils.java
Caideyipi: Fixed in 63608c37842. `getTimePartitionId()` now uses a primitive `Math.subtractExact(time, origin)` plus `Math.floorDiv` fast path for both zero and non-zero origins, and enters the BigInteger path only when the subtraction overflows.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/windowing/handler/SlidingTimeWindowEvaluationHandler.java
Caideyipi: Fixed in 63608c37842. The handler state is long again. Boundary advancement uses `LongMath.saturatedAdd`, with overflow flags preserving the prior beyond-Long.MAX_VALUE behavior without per-point BigInteger allocation. `SlidingTimeWindowEvaluationHandlerTest` passes.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/aggregation/timerangeiterator/ITimeRangeIterator.java
Caideyipi: Fixed in 63608c37842. The duplicated private saturating helpers were removed and the call sites now use Guava `LongMath.saturatedAdd`, `saturatedSubtract`, and `saturatedMultiply`. A single shared `saturateToLong(BigInteger)` remains in node-commons for the few genuine BigInteger narrowing paths.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/utils/DataNodeDateTimeUtils.java
Caideyipi: Fixed in 63608c37842. Public duration conversion and `constructTimeDuration` paths now translate arithmetic overflow into an i18n `SemanticException` inside `DataNodeDateTimeUtils`. The overflow assertions in `DateTimeUtilsTest` were updated and pass in the targeted test run.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/analyze/AnalyzeVisitor.java
Caideyipi: Fixed in 63608c37842. The analyzer now normalizes ordered time ranges by merging both overlapping and adjacent ranges, with a Long.MAX_VALUE guard. The MAX-1 special case was removed, and memory reservation now works from the normalized ranges.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/tsfile/generator/TsFileNameGenerator.java
Caideyipi: Fixed in 63608c37842. The filename pattern now permits a negative timestamp but keeps the version component non-negative, with a comment documenting the distinction. The tests now compare large positive versions (`Long.MAX_VALUE - 1` and `Long.MAX_VALUE`).

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/parser/ASTVisitor.java
Caideyipi: Fixed in 63608c37842. These empty single-comparison ranges now use the same localized time-predicate-selects-no-range message as the analyzer and SHOW REGIONID paths, rather than the range-comparison message.

Caideyipi:
Addressed the review feedback in 63608c37842 and refreshed the PR title/description.

Highlights:
- kept exact partition math through boundary resolution and restored primitive fast paths for partition routing, date_bin, window TVFs, and sliding windows;
- fixed timestamp-difference precision, gap-fill exclusivity, single-row forecast validation, duration overflow handling, adjacent range normalization, and filename version validation;
- consolidated saturating arithmetic on Guava LongMath and moved new user-visible messages to matching English/Chinese i18n classes.

Validation completed:
- node-commons targeted tests: 23 + 4 passed;
- datanode targeted tests: 57 passed / 1 skipped, plus 19 passed;
- datanode English compile and Chinese test-compile passed;
- node-commons/calc-commons English and Chinese test-compile passed;
- git diff --check passed.

The full reactor still stops before the changed modules in thrift-commons because generated TaskType.java requires javax.annotation.Generated in this local environment. All touched module-level builds and targeted tests above complete successfully.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/udf/builtin/relational/tvf/HOPTableFunction.java
```diff
@@ -158,12 +161,15 @@ public void process(
       // n*slide + size
       long timeValue = input.getLong(0);
       if (timeValue >= origin) {
-        long window_start = origin + (timeValue - origin - size + slide) / slide * slide;
-        while (window_start <= timeValue && window_start + size > timeValue) {
-          properColumnBuilders.get(0).writeLong(window_start);
-          properColumnBuilders.get(1).writeLong(window_start + size);
+        long windowStart = getWindowStart(timeValue, origin, slide, slide - size);
```
JackieTien97: [P1] Preserve HOP windows when the initial start underflows

`getWindowStart()` saturates the exact start before this loop. With `SIZE=2`, `SLIDE=1`, `ORIGIN=Long.MIN_VALUE`, and an input row at `Long.MIN_VALUE`, the two matching mathematical window starts are `Long.MIN_VALUE - 1` and `Long.MIN_VALUE`. Under the PR's endpoint-saturation policy, they should be emitted as `[Long.MIN_VALUE, Long.MIN_VALUE + 1)` and `[Long.MIN_VALUE, Long.MIN_VALUE + 2)`. The current code clamps the first start to `Long.MIN_VALUE` before iteration and therefore emits only the second window.

I reproduced this with the same focused test on both revisions: it passes at `87e6564` with 2 outputs, but fails at `63608c3` because `ColumnBuilder.writeLong` is called only once. Please preserve the exact initial start, or otherwise preserve the number of pre-boundary windows, through iteration and saturate only the emitted endpoints. The normal path can remain primitive; a `BigInteger` fallback is only needed after actual overflow.

JackieTien97 (changes requested):
The allocation concern is addressed on normal window-TVF paths, but the overflow fallback now changes HOP cardinality at the lower timestamp boundary. The focused reproduction is included inline.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/udf/builtin/relational/tvf/HOPTableFunction.java
Caideyipi: Fixed in 442c302e09a. HOP now keeps the exact mathematical initial start when the checked long calculation overflows: the normal path stays on primitive long arithmetic, and only that initial overflow uses BigInteger iteration. Window endpoints are saturated only when written, so the Long.MIN_VALUE case preserves both windows. Added testHopWindowPreservesStartsBeforeLongMin; TimeWindowTableFunctionBoundaryTest passes 5/5. I left this thread unresolved.

JackieTien97 (approved):

PR #17895, opened 2026-06-10
Bound Ratis reconfiguration retries and add region migration ITs

CRZbulabula:
## Problem

When a DataNode that hosts a newly ADDING peer is killed during a Ratis region migration, the Ratis configuration change can stay in the staging state. The previous reconfiguration client retried these failures forever, so the DataNode add-peer task could remain PROCESSING indefinitely and the ConfigNode migration procedure could not fail and roll back.

## Changes

- Bound Ratis reconfiguration retries with a new `reconfigurationMaxRetryAttempts` client setting. Reconfiguration requests still use a fixed 2s retry interval, but now stop after the configured attempt count and propagate the last failure to the caller.
- Renamed the reconfiguration retry client factory/policy from the old endless-retry naming to `ReconfigurationRetryFactory` and `RatisReconfigurationRetryPolicy`.
- Added per-scope retry knobs:
  - `config_node_ratis_reconfiguration_max_retry_attempts`
  - `schema_region_ratis_reconfiguration_max_retry_attempts`
  - `data_region_ratis_reconfiguration_max_retry_attempts`
- Set the default retry count to `15`, which is roughly 30 seconds with the fixed 2s interval. This keeps the timeout close to the default ConfigNode Unknown detection window while still leaving a small buffer for transient delays.
- Propagated schema/data reconfiguration retry settings from ConfigNode to DataNodes through optional `TRatisConfig` fields. The fields are optional so rolling upgrades can fall back to local DataNode defaults when an old ConfigNode does not send them.
- Wired the new retry setting through ConfigNode, DataNode, and the Ratis consensus config builders for ConfigNode, schema-region, and data-region consensus.
- Added integration-test config setters for the new Ratis reconfiguration retry knobs.
- Extended the region migration reliability IT framework with a failure-and-rollback path that waits for kill points, verifies the region replica set returns to the original DataNode set, and verifies the target peer is cleared.
- Added a Ratis DataNode kill point during destination local peer creation so the ADDING-peer crash path can be reproduced.

## Ratis IT Coverage

The branch adds Ratis region migration ITs to match the IoTV1 coverage pattern and cover Ratis-specific consensus-pipe states:

- DataNode crash while creating the destination local peer, expecting migration failure and rollback.
- ConfigNode crash at add-region states:
  - `CREATE_NEW_REGION_PEER`
  - `CREATE_CONSENSUS_PIPES`
  - `DO_ADD_REGION_PEER`
  - `UPDATE_REGION_LOCATION_CACHE`
- ConfigNode crash at remove-region states:
  - `TRANSFER_REGION_LEADER`
  - `REMOVE_REGION_PEER`
  - `DELETE_OLD_REGION_PEER`
  - `DROP_CONSENSUS_PIPES`
  - `REMOVE_REGION_LOCATION_CACHE`
- ConfigNode crash with all add/remove migration kill points enabled in one migration.
- Cluster crash at the IoTV1-covered migration states, plus Ratis `CREATE_CONSENSUS_PIPES` and `DROP_CONSENSUS_PIPES`.

The Ratis IT setup uses reconfiguration retry count `10` to keep failure-path tests bounded. The new Ratis IT classes are now tagged with `DailyIT` after passing CI.

## Verification

- `mvn test-compile -pl integration-test -P with-integration-tests -DskipTests`
- `mvn verify -DskipUTs -Dit.test=IoTDBRegionMigrateAddingPeerCrashForRatisIT,IoTDBRegionMigrateConfigNodeCrashForRatisIT,IoTDBRegionMigrateClusterCrashForRatisIT -DintegrationTest.includedGroups= -DintegrationTest.excludedGroups= -DfailIfNoTests=false -Dfailsafe.failIfNoSpecifiedTests=false -pl integration-test -P with-integration-tests`
  - 19 tests passed.
- `mvn spotless:check -pl integration-test -P with-integration-tests`
- `git diff --check`

### DIFF
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeConfig.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeConfig.java
@@ -318,6 +318,16 @@ public class ConfigNodeConfig {
   private long schemaRegionRatisInitialSleepTimeMs = 100;
   private long schemaRegionRatisMaxSleepTimeMs = 10000;
 
+  /**
+   * RatisConsensus protocol, max retry attempts for a configuration change (add/remove peer). Uses
+   * a fixed 2s retry interval; bounding the attempts stops a killed ADDING peer from blocking the
+   * reconfiguration -- and hence a region migration -- forever.
+   */
+  private int configNodeRatisReconfigurationMaxRetryAttempts = 600;
+
+  private int dataRegionRatisReconfigurationMaxRetryAttempts = 600;
+  private int schemaRegionRatisReconfigurationMaxRetryAttempts = 600;
+
   private long configNodeRatisPreserveLogsWhenPurge = 1000;
   private long schemaRegionRatisPreserveLogsWhenPurge = 1000;
   private long dataRegionRatisPreserveLogsWhenPurge = 1000;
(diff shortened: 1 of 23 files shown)


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeConfig.java
```diff
@@ -318,6 +318,16 @@ public class ConfigNodeConfig {
...
+  /**
+   * RatisConsensus protocol, max retry attempts for a configuration change (add/remove peer). Uses
+   * a fixed 2s retry interval; bounding the attempts stops a killed ADDING peer from blocking the
+   * reconfiguration -- and hence a region migration -- forever.
+   */
+  private int configNodeRatisReconfigurationMaxRetryAttempts = 600;
+
+  private int dataRegionRatisReconfigurationMaxRetryAttempts = 600;
+  private int schemaRegionRatisReconfigurationMaxRetryAttempts = 600;
+
```
jt2594838, line 330: Will 20 minutes be too long? May use a similar interval, referring to how CN determines whether a node is unknown.

jt2594838 (approved):

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeConfig.java
CRZbulabula: Agreed. 20 minutes is too long for this failure path. I changed the default reconfiguration retry attempts from 600 to 15, which is about 30 seconds with the fixed 2s retry interval. This keeps it close to the ConfigNode Unknown detection window while leaving a small buffer for transient delays. The newly added Ratis region migration ITs have also been moved into DailyIT after the CI passed.

PR #17896, opened 2026-06-10
Only send table deletion to related region groups

jt2594838:
## Description
- Query ConfigNode for DataRegion groups by delete database and deletion time range instead of reading the full region route map.
- Add coverage for fetching delete replica sets from the target database/time ranges and avoiding getLatestRegionRouteMap.

## Verification
- git diff --check
- mvn test -pl iotdb-core/datanode -Dtest=AnalyzeUtilsTest (blocked during main compile by existing missing generated/dependency classes such as SubscriptionWalRetentionPolicy, RegionProgress, WriterId, WriterProgress, TGetCommitProgressReq, and RelationalSqlParser.NextFillContext)


hongzhi-gao (approved):

Caideyipi (approved):
LGTM. I reviewed the change and did not find any blocking correctness issues. Routing table deletions through getRegionGroupsByTime by database and deletion time range matches the ConfigNode partition lookup semantics, and an empty region result is handled as a no-op by the write dispatch path. Optional follow-up: add coverage for non-success and empty/unset regionReplicaSets responses from ConfigNode.

PR #17899, opened 2026-06-10
Add cache for single measurement validation

jt2594838:
## Description

This PR adds a configurable cache for `PathUtils#checkAndReturnSingleMeasurement` so repeated measurement validation can reuse previous check results. The cache stores both valid and invalid validation results and can be disabled by setting the cache size to `0`.

It also adds configuration plumbing for integration tests and an ignored performance IT that can be run manually to compare end-to-end write performance with the cache disabled and enabled.

## Test Report

Local verification:

```text
mvn spotless:apply -pl integration-test -P with-integration-tests
mvn spotless:apply -pl iotdb-core/datanode
mvn test-compile -DskipUTs -Drat.skip=true -pl integration-test -am -P with-integration-tests
```

Performance averages, 3 repeats per group, 80,000 rows per experiment:

| measurementCount | cacheSize | Relation | Cache Disabled | Cache Enabled | Enabled/Disabled |
|---:|---:|---|---:|---:|---:|
| 1000 | 500 | cacheSize < measurementCount | 4794 ms | 4384 ms | 0.914 |
| 2000 | 2000 | cacheSize = measurementCount | 9676 ms | 8255 ms | 0.853 |
| 4000 | 8000 | cacheSize > measurementCount | 19980 ms | 17223 ms | 0.862 |

### DIFF
--- a/integration-test/src/main/java/org/apache/iotdb/it/env/cluster/config/MppCommonConfig.java
+++ b/integration-test/src/main/java/org/apache/iotdb/it/env/cluster/config/MppCommonConfig.java
@@ -503,6 +503,13 @@ public class MppCommonConfig extends MppBaseConfig implements CommonConfig {
   }
 
   @Override
+  public CommonConfig setSingleMeasurementCheckCacheSize(int singleMeasurementCheckCacheSize) {
+    setProperty(
+        "single_measurement_check_cache_size", String.valueOf(singleMeasurementCheckCacheSize));
+    return this;
+  }
+
+  @Override
   public CommonConfig setDnConnectionTimeoutMs(int connectionTimeoutMs) {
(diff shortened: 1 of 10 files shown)


Caideyipi:
I think this part has a correctness issue: checkTsFileResourceExists currently warns for every .tsfile because it only checks the filename suffix and never checks whether the corresponding resource file exists. For a normal snapshot containing both xxx.tsfile and xxx.tsfile.resource, this still logs The associated resource file ... is not found in the snapshot, which is a false warning. Please check 
ew File(file.getAbsolutePath() + TsFileResource.RESOURCE_SUFFIX).exists() before warning, or remove this unrelated snapshot change from this PR.

PR #17902, opened 2026-06-10
Fix duplicate scheduling in procedure execution

Caideyipi:
### Prevent duplicate procedure execution
- Add a per-procedure execution guard so a duplicated scheduler entry cannot run the same Procedure instance concurrently.
- Release the guard, active executor count, and worker state in the worker finally path.
- Release locks when a scheduled procedure is no longer RUNNABLE.

### Avoid duplicate lock wakeups
- De-duplicate LockQueue waiters by procedure id before adding a waiting procedure.
- Keep PipeTaskCoordinatorLock waiting until a permit is actually acquired, even if the thread is already interrupted.

### Tests
- Add coverage for duplicate procedure scheduling, duplicate LockQueue wakeups, interrupted PipeTaskCoordinatorLock acquisition, and completed PipeHandleLeaderChangeProcedure equality.

<hr>

This PR has:
- [x] been self-reviewed.
    - [x] concurrent read
    - [x] concurrent write

<hr>

- org.apache.iotdb.confignode.procedure.Procedure
- org.apache.iotdb.confignode.procedure.ProcedureExecutor
- org.apache.iotdb.confignode.procedure.scheduler.LockQueue
- org.apache.iotdb.confignode.manager.pipe.coordinator.task.PipeTaskCoordinatorLock

##### Local verification
- PASS: mvn -pl iotdb-core/confignode spotless:apply
- FAIL: mvn -pl iotdb-core/confignode -Dtest=TestLockRegime,TestProcedureExecutor,PipeHandleLeaderChangeProcedureTest,PipeTaskCoordinatorLockTest test failed during compile before test execution because local generated/snapshot dependencies were missing classes such as TGetCommitProgressReq, TPullCommitProgressReq, CommitProgressKeeper, and RegionProgress.


iotdb-core/confignode/src/test/java/org/apache/iotdb/confignode/procedure/TestProcedureExecutor.java
```diff
@@ -96,6 +99,37 @@ public void testWorkerThreadStuck() throws InterruptedException {
...
+
+    Assert.assertTrue(blockingProcedure.awaitExecution(30, TimeUnit.SECONDS));
+
+    procExecutor.getScheduler().addFront(blockingProcedure);
+    boolean duplicated = blockingProcedure.awaitExecution(3, TimeUnit.SECONDS);
+
+    blockingProcedure.releaseExecutions(duplicated ? 2 : 1);
+    ProcedureTestUtil.waitForProcedure(procExecutor, procId);
+
+    Assert.assertFalse(duplicated);
```
jt2594838, line 115: line 112 is weird together with line 115.

PR #17904, opened 2026-06-10
Fix WAL roll file test flush wait

Caideyipi:
Wait for the WALFlushListener returned by WAL writes instead of polling WAL buffer internal consumed state. Tests: mvn -Ddevelocity.off=true spotless:check -pl iotdb-core/datanode


Caideyipi:
Updated this PR onto the latest master (dc47a3baa0a). The reported timeout occurs while polling WALBuffer.isAllWALEntriesConsumed() before the actual waitForNextReady assertion. The test now waits on the WALFlushListener returned by each write, which synchronizes on the completed flush without depending on the buffer's transient internal state. Spotless check passes locally; CI has been re-triggered.

PR #17907, opened 2026-06-11
Fix procedure lock wait scheduling

Caideyipi:
This PR fixes flaky procedure lock wait scheduling by moving LOCK_EVENT_WAIT enqueueing into a procedure hook instead of hard-casting ProcedureExecutor to ConfigNodeProcedureEnv. It also updates the test scheduler to enqueue waiting procedures only after the current execution attempt completes, preserves the exact procedure that should wait/yield during rollback stack handling, and makes root-stack/worker execution cleanup exception-safe. Tests: mvn -pl iotdb-core/confignode -am -Dtest=TestLockRegime#testAcquireLock -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false test; mvn -pl iotdb-core/confignode -am -Dtest=TestLockRegime#testAcquireLock,STMProcedureTest#testRolledBackProcedure -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false test

### DIFF
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/Procedure.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/procedure/Procedure.java
@@ -236,6 +236,16 @@ public abstract class Procedure<Env> implements Comparable<Procedure<Env>> {
   }
 
   /**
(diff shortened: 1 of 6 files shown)


Caideyipi:
Closed because the fix has been integrated into #17902 via cherry-pick commit 7b073bddc056af470ea5a99869d16fafc311392d.

PR #17909, opened 2026-06-11
Fix pipe runtime meta push blocking user operations

Caideyipi:
Summary:
- Add a shared best-effort short-timeout runtime metadata request path for background procedures.
- Use it for pipe leader/meta change pushes and subscription leader-change commit progress/topic/consumer meta pushes so optional DataNode RPCs do not hold coordinator locks for the full sync interval.
- Split subscription runtime state push during leader change: readable DataNodes remain strict, while unreadable DataNodes are updated best-effort with a short timeout.
- Keep user-facing operations and periodic sync procedures on existing strict paths.

Fixes/targets:
- IoTDBPipeClusterIT.testPipeAfterDataRegionLeaderStop flaky timeout on the second createPipe after full cluster restart: the client saw SocketTimeoutException: Read timed out while createPipe(p2) was waiting behind background pipe runtime metadata work.
- In the 2026-06-11 failure logs, PipeHandleLeaderChangeProcedure was pushing PIPE_PUSH_ALL_META to restarting DataNodes, repeatedly hit Connection refused, held PipeTaskCoordinatorLock, and caused Pipe-Runtime-Heartbeat to skip rounds. This PR bounds those background runtime metadata pushes so they do not block user-facing pipe operations for the full sync interval.

Tests:
- git diff --check
- mvn -DskipTests spotless:apply -pl iotdb-core/confignode
- mvn '-Ddevelocity.off=true' -DskipTests compile -pl iotdb-core/confignode -am
- Attempted targeted IoTDBPipeClusterIT#testPipeAfterRegisterNewDataNode; local JVM native OOM during ConfigNode startup before pipe stage

### DIFF
--- a/iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/RpcUtils.java
+++ b/iotdb-client/service-rpc/src/main/java/org/apache/iotdb/rpc/RpcUtils.java
@@ -151,8 +151,7 @@ public class RpcUtils {
   }
 
   public static void verifySuccess(List<TSStatus> statuses) throws BatchExecutionException {
-    StringBuilder errMsgs =
-        new StringBuilder().append(TSStatusCode.MULTIPLE_ERROR.getStatusCode()).append(": ");
+    StringBuilder errMsgs = new StringBuilder();
     for (TSStatus status : statuses) {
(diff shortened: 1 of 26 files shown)


jt2594838 (approved):

CRZbulabula (approved):
LGTM

PR #17913, opened 2026-06-11
Pipe: stop historical supply promptly on DROP PIPE

luoluoyuyu:
## Summary
- Drop pipe tasks in sink → processor → source order so outbound transfer stops before historical source cleanup begins.
- Track the processor thread currently inside `supply()` and interrupt it from `close()` via `compareAndSet`, avoiding stale-thread interrupts.
- Let `start()` abort promptly on close/interrupt and unpin only the TsFile resources that were pinned but not yet supplied.

## Why
On DROP PIPE, `close()` could block behind a long-running `supply()/start()` while processor and sink kept forwarding historical TsFile events even though the pipe had already disappeared from `SHOW PIPES`.

## Test plan
- [x] `mvn spotless:apply -pl iotdb-core/datanode`
- [ ] Reproduce pure historical pipe DROP and verify receiver stops receiving data immediately
- [ ] Capture thread dump and confirm drop worker no longer waits long behind historical extraction

Made with [Cursor](https://cursor.com)

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/PipeDataNodeTask.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/PipeDataNodeTask.java
@@ -22,6 +22,7 @@ package org.apache.iotdb.db.pipe.agent.task;
 import org.apache.iotdb.commons.pipe.agent.task.PipeTask;
 import org.apache.iotdb.commons.pipe.agent.task.stage.PipeTaskStage;
 import org.apache.iotdb.db.i18n.DataNodePipeMessages;
+import org.apache.iotdb.db.pipe.agent.task.stage.PipeTaskSourceStage;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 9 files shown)


Caideyipi (commented):
Thanks for the fix. I found a few correctness issues that look worth addressing before merge:

1. `PipeDataNodeTask.drop()` now drops `sinkStage` before `processorStage` and `sourceStage`. However `PipeSinkSubtaskManager.deregister()` also deregisters the committer and resets the IoTV2 replicate index, and the existing comment there explicitly assumes extractor and processor have already been dropped. With the new order, processor/source may still be running and the historical source may still assign IoTV2 replicate indexes after the reset, which breaks the intended "reset after the consensus pipe no longer replicates data" invariant. I think stopping outbound transfer and deregistering/resetting should be separated, or the reset should stay after processor/source are fully stopped.

2. `PipeHistoricalDataRegionTsFileAndDeletionSource.supply()` can now discard a `PipeTerminateEvent` through `discardEventIfCloseRequested(supplyTerminateEvent())`. `supplyTerminateEvent()` sets `isTerminateSignalSent = true` before returning, so if the close request arrives in that window, the terminate event is cleared and returned as `null`; then `close()` skips `PipeTerminateEvent.clearHistoricalTransferSummary(...)` because `isTerminateSignalSent` is already true. That leaves the historical transfer summary behind, and it also violates the `PipeTerminateEvent` contract that it should never be discarded. The terminate path probably needs special handling, or the summary/state should be cleaned when the terminate event is intentionally dropped.

3. `close()` reads `activeSupplyThread` and interrupts it without claiming the reference. There is still a stale-thread race: `close()` can read the thread, `supply()` can finish and clear `activeSupplyThread` in `finally`, and then `close()` interrupts the same worker after it has left `supply()`. The PR summary says stale-thread interrupts are avoided via compare-and-set, but the CAS is only used on the supply side. `close()` should probably interrupt only if it wins `activeSupplyThread.compareAndSet(supplyThread, null)`.

luoluoyuyu:
Superseded by a minimal fix: mark in-flight event on processor close and skip linking further objects in collector.

PR #17924, opened 2026-06-12
Fix flaky region-migration/cluster ITs and enable IoTV2 daily migration tests

CRZbulabula:
## What this PR does

Fixes three flaky integration tests that surfaced in the nightly **Daily IT**, and re-enables / corrects the IoTV2 region-migration daily tests. All region-migration daily tests stay in the `DailyIT` category — they were only **temporarily** run at PR granularity to pre-validate them (see "Pre-validation" below).

## Flaky-test fixes

All three are *"a result becomes visible before the state the test depends on is actually ready"* timing issues — test bugs, not product bugs:

1. **`IoTDBIoTConsensusV23C3DBasicITBase.testDeleteTimeSeriesReplicaConsistency`** (stream & batch)
   In Step 7 the restarted DataNode was only awaited via `isAlive()` (OS process up), not until it could serve queries. The next loop iteration connected to it and intermittently hit `Connection refused` (RPC port not yet open / not re-registered). Now we wait until the node is actually queryable.

2. **`IoTDBCustomizedClusterIT.testRepeatedlyRestartWholeClusterWithWrite`**
   The `SELECT last s1 FROM root.**` is fanned out to every DataNode and compared positionally across replicas. Right after a full-cluster restart the last cache is reloaded lazily, so the row order could differ across coordinators → `InconsistentDataException`. Added `ORDER BY TIMESERIES` to make the order deterministic (the root cause) and wrapped the comparison in a retry to tolerate the brief convergence window — without masking a genuine, persistent inconsistency.

3. **Region-migration kill-point framework** (`IoTDBRegionOperationReliabilityITFramework`)
   Kill points are detected by a background thread tailing the node log. In the success path `checkKillPointsAllTriggered` could run before that thread processed the kill-point line of the last phase (e.g. `RemoveRegionLocationCache`), failing spuriously with *"Some kill points was not triggered"*. Added a best-effort `graceWaitForKillPointsTriggered` (bounded wait, swallows the timeout) before the authoritative `checkKillPointsAllTriggered`. It deliberately does **not** throw, so the `badKillPoint` test (which expects an `AssertionError` when a kill point never triggers) still passes.

## IoTV2 region-migration daily tests

- **Re-enable** the ConfigNode / Cluster / DataNode crash tests that were commented out pending discussion. The `@Ignore`d `cnCrashDuringPreCheckTest` is left ignored.
- **`originalCrashDuringAddPeerDone`** (batch & stream): `failTest` → `successTest`. Once the add-peer phase is done the new peer already holds the data, so region migration is **designed** to tolerate the original (source) DataNode crashing afterwards — it completes successfully and only leaves the region files on the dead node for later cleanup (`DELETE_OLD_REGION_PEER` logs *"procedure will continue. You should manually delete region file"*). Only the coordinator / destination crash scenarios should fail, and those remain `failTest`.

## Pre-validation

To gain confidence before merging, the **entire region-migration daily suite** (IoTV2 batch/stream, IoTV1, Ratis) was **temporarily** moved from `DailyIT` to the per-PR `ClusterIT` category and run through the full PR pipeline. All region-migration tests passed; the only failures observed were pre-existing, unrelated flaky tests (`IoTDBUDTFAlignByTimeQueryIT.queryWithValueFilter5`, `IoTDBPipePermissionIT.testIllegalPassword`) that also pass on re-run. After validation, the category was restored to `DailyIT`, so this PR's net diff contains only the substantive fixes above.

### DIFF
--- a/integration-test/src/test/java/org/apache/iotdb/confignode/it/regionmigration/IoTDBRegionOperationReliabilityITFramework.java
+++ b/integration-test/src/test/java/org/apache/iotdb/confignode/it/regionmigration/IoTDBRegionOperationReliabilityITFramework.java
@@ -355,6 +355,17 @@ public class IoTDBRegionOperationReliabilityITFramework {
         Assert.fail();
       }
 
+      // The kill point is detected by a background thread tailing the node log, so the migration
+      // result (observed by awaitUntilSuccess above) can become visible before that thread has read
+      // and processed the kill-point line of the last migration phase (e.g.
+      // RemoveRegionLocationCache). Give that thread a short grace period to catch up before
+      // asserting, otherwise checkKillPointsAllTriggered may fail spuriously with "Some kill points
+      // was not triggered". This is best-effort: the authoritative assertion remains
+      // checkKillPointsAllTriggered, which still fails the test if a kill point genuinely never
+      // triggers (e.g. the badKillPoint test).
+      graceWaitForKillPointsTriggered(configNodeKeywords);
+      graceWaitForKillPointsTriggered(dataNodeKeywords);
+
       // make sure all kill points have been triggered
       checkKillPointsAllTriggered(configNodeKeywords);
       checkKillPointsAllTriggered(dataNodeKeywords);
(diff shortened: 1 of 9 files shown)


Caideyipi (commented):
I think the retry here doesn't actually cover the flaky failure described in the PR.

Awaitility.untilAsserted in Awaitility 4.2.0 only treats AssertionError as a retryable assertion failure; other Throwables are rethrown immediately. The transient failure this change is trying to tolerate is InconsistentDataException, which is a RuntimeException thrown from ClusterTestResultSet.next()/getString() via RequestDelegate.requestAllAndCompare(). That means the first cross-DataNode mismatch will still fail the test immediately instead of being retried for up to 60 seconds.

Could you explicitly make InconsistentDataException retryable here, for example with ignoreExceptionsMatching(e -> e instanceof InconsistentDataException), or catch only InconsistentDataException inside the lambda and convert it to an assertion failure/false retry?

CRZbulabula:
@Caideyipi Good catch — you're exactly right, thanks!

`InconsistentDataException` is a `RuntimeException` (thrown from `getString()` via `RequestDelegate.requestAllAndCompare()`), and `Awaitility.untilAsserted` only treats `AssertionError` as retryable by default, so the retry was effectively a no-op for this failure — only the `ORDER BY TIMESERIES` was actually fixing it.

I've pushed a fix following your suggestion (483a4ce0de): added `.ignoreExceptionsMatching(e -> e instanceof InconsistentDataException)` so the retry genuinely tolerates the transient cross-replica inconsistency during the post-restart convergence window, while a genuine error (e.g. a real `SQLException`) still fails fast rather than being retried for 60s.

I kept `ORDER BY TIMESERIES` as the primary fix: the observed mismatch was purely cross-coordinator row **ordering** (timing-independent — different coordinators iterate the last cache in different orders), which retrying an unordered query alone would never converge on. The retry now covers the orthogonal case where a replica is still catching up after the restart.

Caideyipi (approved):
Re-checked after the latest fix. The InconsistentDataException retry is now scoped correctly, and I didn't find any remaining blocker.

Pengzna:
Great work!!!

PR #17926, opened 2026-06-12
Subscription: Drain TsFile batches before termination

VGalaxies:
### Motivation
When export-tsfile uses subscription/pipe to export TsFiles, a PipeTerminateEvent could mark the snapshot topic completed before the final internal TsFile batch was emitted and consumed. The client could then receive termination, clean up the topic, and miss the last batch of data.

### Modifications
- Hold PipeTerminateEvent in the subscription prefetching queue until pending batches are emitted.
- Add a forced batch emit path for termination draining.
- Commit the terminate event only after the prefetched queue and in-flight subscription events are empty.

### Tests
- `git diff --check`
- `mvn compile -pl iotdb-core/datanode` (fails in unrelated existing/generated dependency areas: missing `PipePeriodicalLogReducer`, relational grammar `identifier()`, `CalcMessages`, and `IndexedBlockingReserveQueue` API mismatches; modified files are not reported in the compilation errors)

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/broker/SubscriptionPrefetchingQueue.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/broker/SubscriptionPrefetchingQueue.java
@@ -117,6 +117,7 @@ public abstract class SubscriptionPrefetchingQueue {
   private volatile TsFileInsertionEvent currentTsFileInsertionEvent;
   private volatile RetryableEvent<TabletInsertionEvent> currentTabletInsertionEvent;
   private volatile SubscriptionTsFileToTabletIterator currentToTabletIterator;
+  private volatile PipeTerminateEvent currentTerminateEvent;
 
   public SubscriptionPrefetchingQueue(
       final String brokerId,
@@ -174,6 +175,10 @@ public abstract class SubscriptionPrefetchingQueue {
       ((EnrichedEvent) currentTabletInsertionEvent.innerEvent)
           .clearReferenceCount(this.getClass().getName());
       currentTabletInsertionEvent = null;
+    }
+    if (Objects.nonNull(currentTerminateEvent)) {
+      currentTerminateEvent.clearReferenceCount(this.getClass().getName());
+      currentTerminateEvent = null;
     }
   }
 
@@ -467,6 +472,10 @@ public abstract class SubscriptionPrefetchingQueue {
    * {@link SubscriptionPrefetchingQueue#inputPendingQueue} is empty.
    */
   private synchronized void tryPrefetch() {
+    if (Objects.nonNull(currentTerminateEvent) && !tryCommitCurrentTerminateEvent()) {
+      return;
+    }
+
     while (!inputPendingQueue.isEmpty() || Objects.nonNull(currentTabletInsertionEvent)) {
       if (Objects.nonNull(currentTabletInsertionEvent)) {
         final RetryableState state = onRetryableTabletInsertionEvent(currentTabletInsertionEvent);
@@ -497,16 +506,10 @@ public abstract class SubscriptionPrefetchingQueue {
       }
 
       if (event instanceof PipeTerminateEvent) {
-        final PipeTerminateEvent terminateEvent = (PipeTerminateEvent) event;
-        // add mark completed hook
-        terminateEvent.addOnCommittedHook(this::markCompleted);
-        // commit directly
-        ((PipeTerminateEvent) event)
-            .decreaseReferenceCount(SubscriptionPrefetchingQueue.class.getName(), true);
-        LOGGER.info(
-            "Subscription: SubscriptionPrefetchingQueue {} commit PipeTerminateEvent {}",
-            this,
-            terminateEvent);
+        currentTerminateEvent = (PipeTerminateEvent) event;
+        if (!tryCommitCurrentTerminateEvent()) {
+          return;
+        }
         continue;
       }
 
@@ -549,6 +552,11 @@ public abstract class SubscriptionPrefetchingQueue {
   }
 
   private synchronized void tryPrefetchV2() {
+    if (Objects.nonNull(currentTerminateEvent)) {
+      tryCommitCurrentTerminateEvent();
+      return;
+    }
+
     if (!prefetchingQueue.isEmpty()) {
       return;
     }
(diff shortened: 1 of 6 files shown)


Caideyipi (commented):
I found two issues that seem worth fixing before merge:

1. `tryCommitCurrentTerminateEvent()` checks `prefetchingQueue.isEmpty()` and `inFlightEvents.isEmpty()` non-atomically. `poll()`/`pollV2()` remove an event from `prefetchingQueue` before putting it into `inFlightEvents`, and both polling and prefetch execution only hold the read lock, so they can run concurrently. If terminate draining runs in that transition window, the terminate event can still be committed while the last batch has just been polled but not yet tracked as in-flight. That preserves the early-completion race this PR is trying to fix. Please make the poll transition and terminate-commit check mutually exclusive, or include an explicit "poll transition in progress" state in the commit condition.

2. For tablet queues, a pending `currentTerminateEvent` is retried only from inside `tryPrefetch()`, but `executePrefetch()` calls `tryPrefetch()` only when `states.shouldPrefetch()` is true. If normal prefetching is throttled or disabled while the terminate event is waiting for the queue to drain, completion can be delayed until another poll path happens. The terminate-drain path should bypass the normal prefetch heuristic once `currentTerminateEvent` is set.

VGalaxies:
Addressed Caideyipi's review comments in `cfdf8dacc1`:

1. Made the prefetch-to-in-flight poll transition mutually exclusive with the terminate commit check by moving the transition into a synchronized helper and making `tryCommitCurrentTerminateEvent()` synchronized as well. This removes the window where a polled event is no longer in `prefetchingQueue` but not yet visible in `inFlightEvents`.

2. Updated `executePrefetch()` so a pending `currentTerminateEvent` bypasses the normal `states.shouldPrefetch()` heuristic and keeps retrying the drain/commit path directly.

Validation:
- `git diff --check`
- `mvn compile -pl iotdb-core/datanode` still fails with the same unrelated dependency/generated-source mismatches (`PipePeriodicalLogReducer`, relational grammar `identifier()`, `CalcMessages`, `IndexedBlockingReserveQueue`, etc.); the modified file is not reported in the compilation errors.

VGalaxies:
Added an integration test for the termination-drain scenario requested in review.\n\nWhat it covers:\n- table-model snapshot topic with TsFile subscription format\n- forced TsFile parsing via start-time so data is buffered in SubscriptionPipeTsFileEventBatch\n- large TsFile batch delay/size thresholds so snapshot rows stay pending until the terminate event arrives\n- a background consumer continuously drains and commits all emitted TsFile messages\n- the test asserts the subscription auto-completes and receiver data contains the expected snapshot rows\n\nVerification:\n`mvn verify -DskipUTs -Dit.test=org.apache.iotdb.subscription.it.dual.tablemodel.IoTDBSubscriptionTopicIT#testTsFileSnapshotDrainsPendingBatchBeforeTermination -DfailIfNoTests=false -Dfailsafe.failIfNoSpecifiedTests=false -pl integration-test -am -P MultiClusterIT2SubscriptionTableArchVerification -P with-integration-tests`\n\nResult: BUILD SUCCESS, Tests run: 1, Failures: 0, Errors: 0, Skipped: 0.

VGalaxies:
CI follow-up:\n\n- Investigated the previous `dual-tree-auto-enhanced (17, HighPerformanceMode, HighPerformanceMode, ubuntu-latest, 0)` failure. It failed in `IoTDBPipeClusterIT.testRegisteringNewDataNodeAfterTransferringData` with inconsistent row counts across DataNodes, which is unrelated to the subscription TsFile termination-drain change.\n- Pushed `ddd09219b2` to address the Sonar/checkstyle findings on this PR: removed the non-primitive volatile field access pattern for `currentTerminateEvent`, fixed the long log line, and placed overloaded methods next to each other.\n- Local verification passed:\n  - `mvn spotless:apply -pl iotdb-core/datanode`\n  - `mvn checkstyle:check -pl iotdb-core/datanode`\n  - `mvn verify -DskipUTs -Dit.test=org.apache.iotdb.subscription.it.dual.tablemodel.IoTDBSubscriptionTopicIT#testTsFileSnapshotDrainsPendingBatchBeforeTermination -DfailIfNoTests=false -Dfailsafe.failIfNoSpecifiedTests=false -pl integration-test -am -P MultiClusterIT2SubscriptionTableArchVerification -P with-integration-tests`\n- The new CI run on `ddd09219b2` has already passed `SonarCloud Code Analysis` and `subscription-table-arch-verification`; the remaining large matrix jobs and Codecov are still pending.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/broker/SubscriptionPrefetchingQueue.java
```diff
@@ -731,6 +710,43 @@ protected boolean onEvent() {
...
 
+  private synchronized boolean tryCommitCurrentTerminateEventIfPresent() {
+    if (Objects.isNull(currentTerminateEvent)) {
+      return false;
+    }
+    tryCommitCurrentTerminateEvent();
+    return true;
+  }
+
+  private synchronized boolean tryCommitCurrentTerminateEvent() {
```
jt2594838: The return value is not used?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/broker/SubscriptionPrefetchingQueue.java
```diff
@@ -731,6 +710,43 @@ protected boolean onEvent() {
...
+    }
+
+    // Add mark completed hook only when all subscription events have been consumed.
+    currentTerminateEvent.addOnCommittedHook(this::markCompleted);
+    currentTerminateEvent.decreaseReferenceCount(
+        SubscriptionPrefetchingQueue.class.getName(), true);
+    LOGGER.info(
+        "Subscription: SubscriptionPrefetchingQueue {} commit PipeTerminateEvent {}",
+        this,
+        currentTerminateEvent);
```
jt2594838: i18n

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/broker/SubscriptionPrefetchingQueue.java
VGalaxies: Addressed by splitting the check from the action: `hasCurrentTerminateEvent()` now owns the existence check, and `tryCommitCurrentTerminateEventIfPresent()` performs the commit without returning an ignored boolean.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/broker/SubscriptionPrefetchingQueue.java
VGalaxies: Addressed by moving these terminate-event log messages into `DataNodeMiscMessages` for both en and zh locales.

jt2594838 (approved):

PR #17933, opened 2026-06-12
Uncomment commented set-configuration-able items in the config template (fix enable_topology_probing & topology_probing_* hot/restart reload)

CRZbulabula:
## Description

Several configuration items declared as dynamically settable were left **commented out** in `iotdb-system.properties.template`, so `set configuration` rejected them with:

```
301: ignored config items: [...] because they are immutable or undefined.
```

The most visible case (PingCode V2-995) was `enable_topology_probing` (added in #17595): it is declared `effectiveMode: hot_reload` and the ConfigNode already implements the hot-reload start/stop logic (`ConfigManager.handleTopologyProbingHotReload`), yet `set configuration "enable_topology_probing"="true"` was rejected — so operators could only toggle the TopologyService by editing the file and restarting the ConfigNode (which triggers a leader election and a brief availability hit).

### Root cause

`ConfigurationFileUtils.getConfigurationItemsFromTemplate` only parses **uncommented** `key=value` lines into `configuration2DefaultValue`. When an item's value line is commented out, the key never enters that map, so `filterInvalidConfigItems` treats it as `undefined` and drops it **before** it can reach the apply path. The only `effectiveMode` that `set configuration` is *supposed* to reject is `FIRST_START`; every other mode (`hot_reload`, `restart`, `first_start_or_set_configuration`) should be accepted (written to the file, applied now for hot_reload or on next restart otherwise).

### Fix

Uncomment the affected items so they become bare defaults like every other working settable item (e.g. `slow_query_threshold`, `wal_mode`). No Java logic changes are required — the entire downstream chain was already correct; the commented template line was the only blocker.

Items uncommented:

| Item | effectiveMode |
| --- | --- |
| `enable_topology_probing=false` | hot_reload |
| `topology_probing_base_interval_in_ms=5000` | restart |
| `topology_probing_timeout_ratio=0.5` | restart |
| `cn_max_idle_client_count_for_each_node_in_client_manager=1000` | restart |
| `dn_max_idle_client_count_for_each_node_in_client_manager=1000` | restart |
| `region_group_allocate_policy=GCR` | restart |

All have real descriptor bindings (`ConfigNodeDescriptor` / `CommonDescriptor` / `IoTDBDescriptor`) and validated defaults, so uncommenting them is safe and behavior-preserving for fresh deployments.

### Tests

- **`ConfigurationFileUtilsTest.checkSettableItemsAreUncommentedInTemplate`** — a regression guard that scans the raw template and fails if any item whose `effectiveMode` permits `set configuration` (i.e. anything except `FIRST_START`) is left commented out. It tolerates keys that also have an uncommented variant elsewhere (e.g. the Windows/Unix path pairs such as `dn_data_dirs`), so it only flags items that are genuinely unreachable. This is what surfaced the additional `restart`-mode items beyond `enable_topology_probing`.
- **`IoTDBSetConfigurationIT.testSetTopologyProbingConfiguration`** — end-to-end: asserts `set configuration` accepts `enable_topology_probing` (hot_reload) and the two `topology_probing_*` items (restart) and persists them to the ConfigNode config file.

<hr>

This PR has:
- [x] been self-reviewed.

<hr>

- `iotdb-system.properties.template` (node-commons config template)
- `ConfigurationFileUtilsTest` (datanode unit test)
- `IoTDBSetConfigurationIT` (integration test)

### DIFF
--- a/integration-test/src/test/java/org/apache/iotdb/db/it/IoTDBSetConfigurationIT.java
+++ b/integration-test/src/test/java/org/apache/iotdb/db/it/IoTDBSetConfigurationIT.java
@@ -89,16 +89,20 @@ public class IoTDBSetConfigurationIT {
   }
 
   /**
-   * Regression test for V2-995: {@code enable_topology_probing} is declared {@code effectiveMode:
-   * hot_reload}, so {@code set configuration} must accept it (instead of rejecting it as "immutable
-   * or undefined") and persist it to the config file. This used to fail because the item was left
-   * commented out in the template, hiding it from the default-value map.
+   * Regression test for V2-995: the topology-probing config items were left commented out in the
+   * template, so {@code getConfigurationItemsFromTemplate} never recorded them as known defaults
+   * and {@code set configuration} rejected them as "immutable or undefined". After uncommenting
+   * them, {@code set configuration} must accept and persist them: {@code enable_topology_probing}
+   * (hot_reload) and {@code topology_probing_base_interval_in_ms} / {@code
+   * topology_probing_timeout_ratio} (restart).
    */
   @Test
-  public void testSetHotReloadTopologyProbing() {
+  public void testSetTopologyProbingConfiguration() {
     try (Connection connection = EnvFactory.getEnv().getConnection();
         Statement statement = connection.createStatement()) {
       statement.execute("set configuration \"enable_topology_probing\"=\"true\"");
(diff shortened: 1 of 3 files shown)


Caideyipi (approved):
Reviewed the configuration template and related set-configuration paths. I did not find any functional blocker; only noted a minor non-blocking test robustness point.

PR #17936, opened 2026-06-12
Subscription: support table-model column-filter for topics

VGalaxies:
## Purpose
Add a `column-filter` topic attribute for table-model subscriptions so a topic can
restrict which columns consumers receive, based on column metadata, applied uniformly
across historical, real-time pipe, and IoTConsensus WAL paths.

## What
- New optional topic attribute `column-filter` (table-model only); unset == `true`.
  Restricted boolean predicate over column-metadata fields aligned with
  `information_schema.columns`: `database`, `table_name`, `column_name`, `datatype`,
  `category`. Supports `=`/`!=`, `IN`, `LIKE`, `REGEXP`, `IS [NOT] NULL`, `AND`/`OR`/`NOT`.
- Independent `ColumnFilter.g4` grammar (decoupled from the query grammar) producing the
  reused relational AST nodes; a whitelist validator rejects unsupported syntax/fields.
- Resolved to a per-(database, table) column set at CREATE/ALTER (and on re-bind) against the
  current schema, cached per topic on the DataNode and used for O(1) runtime pruning.
  TAG columns are retained; ATTRIBUTE columns are not transmittable.
- Applied in all paths: IoTConsensus WAL converter, pipe tablet batch, and TsFile reseal
  (prune at tablet granularity; raw TsFile passthrough kept only when the filter is trivial).
- `timeSelected` propagated as an optional, backward-compatible per-table flag on the poll
  response (defaults to exposing time for old consumers).
- ALTER TOPIC hot-refreshes the cached matcher (no converter rebuild; in-flight not reprocessed).
- Removed the legacy consensus-only `column` regex attribute.

## Tests
- Unit: parser/validator/evaluator, binder, tablet pruner, topic config, poll response, record handler.
- Integration (table model): live/snapshot, record & TsFile, multi-consumer-group, historical+
  realtime consistency, ALTER rebind, empty-result progress, snapshot semantics, custom time
  column, altered data type, tree-view source-column binding, attribute drop, 1C3D cluster
  ALTER-after-owner-transfer, DataNode restart re-snapshot, invalid-expression rejection, permission.

## Notes
- The column-filter bound set is a non-persisted snapshot: stable between binds, re-derived
  from the current schema on CREATE/ALTER/restart/owner-transfer; not frozen for the topic lifetime.
- Tree-model topics are unaffected.


Caideyipi (commented):
Functional issues found during review.

Caideyipi:
Findings                                                                                                              
                                                                                                                        
  1. Valid REGEXP/string literals are rejected before parsing                                                           
     ColumnFilterParser.java#L139-L153                                                                                  
     validateUnsupportedSyntax scans the raw expression and rejects <, >, and + before tokenization, but it does not    
     skip quoted string literals. As a result, valid filters such as column_name REGEXP "s[0-9]+", column_name = "a+b", 
     or column_name LIKE "a>%" fail at topic creation time. This should either ignore characters inside quoted strings  
     or let the grammar/parser reject unsupported operators.                                                            
                                                                                                                        
  2. A time-only filter drops all rows for tagless tables                                                               
     ColumnFilterBinder.java#L67-L75, TabletColumnPruner.java#L78-L94                                                   
     A filter like column_name = "time" only sets timeSelected; it does not add any selected physical tablet column. For
     a table without TAG columns, TabletColumnPruner then sees an empty selected column set and returns null, so        
     consumers receive no rows. IoTDB table model allows tagless tables such as CREATE TABLE t (s1 INT64 FIELD), so this
     is a valid case. The implementation should either support time-only tablets or reject this unsupported shape during
     bind/validation.

integration-test/src/test/java/org/apache/iotdb/subscription/it/dual/tablemodel/IoTDBSubscriptionColumnFilterIT.java
```diff
@@ -0,0 +1,1523 @@
...
+            .bind(
+                new TopicConfig(attributes),
+                Map.of(database, Map.of(viewName, createTreeViewSchema(viewName))));
+    final ColumnFilterMatcher matcher =
+        ColumnFilterMatcher.fromBoundColumnFilter(boundColumnFilter);
+
+    Assert.assertTrue(matcher.match(database, viewName, "tag1"));
+    Assert.assertTrue(matcher.match(database, viewName, "s_src"));
+    Assert.assertFalse(matcher.match(database, viewName, "s_view"));
+  }
```
jt2594838: This does not seem to be an IT?

integration-test/src/test/java/org/apache/iotdb/subscription/it/dual/tablemodel/IoTDBSubscriptionColumnFilterIT.java
```diff
@@ -0,0 +1,1523 @@
...
+          topicName,
+          database,
+          viewName,
+          TopicConstant.FORMAT_RECORD_HANDLER_VALUE,
+          "column_name = \"s_view\"");
+    } finally {
+      cleanup(topicName, database);
+      dropTreeDatabase(senderEnv, treeDatabase);
+    }
+  }
```
jt2594838, line 427: Where is the check?

iotdb-client/subscription/src/main/java/org/apache/iotdb/rpc/subscription/payload/poll/SubscriptionPollResponse.java
```diff
@@ -77,6 +110,8 @@ private void serialize(final DataOutputStream stream) throws IOException {
     ReadWriteIOUtils.write(responseType, stream);
     payload.serialize(stream);
     commitContext.serialize(stream);
+    ReadWriteIOUtils.write(timeSelected, stream);
+    serializeTimeSelectedByTable(stream);
```
jt2594838, line 115: They are serialized, then why do you call them transient?

iotdb-client/subscription/src/main/java/org/apache/iotdb/rpc/subscription/payload/poll/SubscriptionPollResponse.java
```diff
@@ -114,7 +149,71 @@ public static SubscriptionPollResponse deserialize(final ByteBuffer buffer) {
...
+    for (int i = 0; i < databaseSize; ++i) {
+      final String databaseName = ReadWriteIOUtils.readString(buffer);
+      final int tableSize = ReadWriteIOUtils.readInt(buffer);
+      final Map<String, Boolean> tableMap = new HashMap<>();
+      for (int j = 0; j < tableSize; ++j) {
+        tableMap.put(ReadWriteIOUtils.readString(buffer), ReadWriteIOUtils.readBool(buffer));
+      }
+      result.put(databaseName, tableMap);
+    }
+    return copyTimeSelectedByTable(result);
```
jt2594838, line 189: Why copy here and there?
Will the map be modified after analysis?

iotdb-client/subscription/src/main/java/org/apache/iotdb/session/subscription/payload/SubscriptionRecordHandler.java
```diff
@@ -89,12 +106,57 @@ public void removeUserData() {
...
+      final boolean defaultTimeSelected,
+      final String databaseName,
+      final Tablet tablet) {
+    if (Objects.isNull(timeSelectedByTable) || timeSelectedByTable.isEmpty()) {
+      return defaultTimeSelected;
+    }
+    final Map<String, Boolean> tableMap = timeSelectedByTable.get(databaseName);
+    if (Objects.isNull(tablet)) {
+      return defaultTimeSelected;
+    }
```
jt2594838: Check tablet before calling map.get()

iotdb-client/subscription/src/main/java/org/apache/iotdb/session/subscription/payload/SubscriptionRecordHandler.java
```diff
@@ -89,12 +106,57 @@ public void removeUserData() {
...
+    final Map<String, Boolean> resolvedTableMap =
+        Objects.nonNull(tableMap)
+            ? tableMap
+            : getIgnoreCase(timeSelectedByTable, databaseName, null);
```
jt2594838: It would be better to store dbNames and tableNames in lowercase in the maps.
Then you may avoid iterating the map.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/subscription/SubscriptionInfo.java
```diff
@@ -374,41 +372,41 @@ private void validateConsensusProtocolSupport(final TopicConfig topicConfig)
...
+    if (columnFilterKeyCount > 1) {
       final String exceptionMessage =
           String.format(
-              "Failed to create or alter topic, %s is only supported for table topics",
-              TopicConstant.COLUMN_KEY);
+              "Failed to create or alter topic, duplicate %s attributes are not allowed",
+              TopicConstant.COLUMN_FILTER_KEY);
       LOGGER.warn(exceptionMessage);
       throw new SubscriptionException(exceptionMessage);
     }
```
jt2594838: Do other keys support duplications?
Better to have consistent behavior.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/table/DataNodeTableCache.java
```diff
@@ -313,6 +313,28 @@ public long getInstanceVersion() {
...
+                      entry.getValue().entrySet().stream()
+                          .collect(
+                              Collectors.toMap(
+                                  Map.Entry::getKey,
+                                  tableEntry -> new TsTable(tableEntry.getValue()),
+                                  (left, right) -> right,
+                                  ConcurrentHashMap::new)),
+                  (left, right) -> right,
+                  ConcurrentHashMap::new));
+    } finally {
```
jt2594838, line 440: Is it necessary to use concurrentMap for snapshot?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/agent/SubscriptionTopicAgent.java
```diff
@@ -93,14 +94,62 @@ public TPushTopicMetaRespExceptionMessage handleSingleTopicMetaChanges(
...
-        topicMetaKeeper.getTopicMeta(topicName), metaFromCoordinator);
+    final TopicMeta oldMeta = topicMetaKeeper.getTopicMeta(topicName);
+    TopicMeta.validateOwnerProgression(oldMeta, metaFromCoordinator);
     topicMetaKeeper.removeTopicMeta(topicName);
     topicMetaKeeper.addTopicMeta(topicName, metaFromCoordinator);
+    if (shouldRefreshColumnFilter(oldMeta, metaFromCoordinator)) {
+      SubscriptionAgent.broker().refreshColumnFilter(topicName, metaFromCoordinator.getConfig());
+    } else if (!metaFromCoordinator.getConfig().isTableTopic()) {
+      SubscriptionAgent.broker().dropColumnFilter(topicName);
+    }
```
jt2594838, line 111: Why would a tree topic be added with ColumnFilter in the first place.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/agent/SubscriptionTopicAgent.java
```diff
@@ -93,14 +94,62 @@ public TPushTopicMetaRespExceptionMessage handleSingleTopicMetaChanges(
...
+
+  private static String getAttributeIgnoreCase(
+      final TopicConfig topicConfig, final String key, final String defaultValue) {
+    return topicConfig.getAttribute().entrySet().stream()
+        .filter(entry -> key.equalsIgnoreCase(entry.getKey()))
+        .map(Map.Entry::getValue)
+        .filter(Objects::nonNull)
+        .findFirst()
+        .orElse(defaultValue);
+  }
```
jt2594838, line 153: May translate cases before constructing attributes.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/broker/consensus/ConsensusLogToTabletConverter.java
```diff
@@ -458,43 +472,18 @@ private List<Tablet> convertRelationalInsertRowNode(final RelationalInsertRowNod
...
+        buildTableModelTabletFromRow(
+            tableName, time, measurements, dataTypes, values, node.getColumnCategories());
+    if (Objects.isNull(tablet)) {
+      return Collections.emptyList();
     }
-    tablet.setRowSize(1);
 
-    return Collections.singletonList(tablet);
+    final Tablet prunedTablet =
+        TabletColumnPruner.pruneTableModelTablet(tablet, databaseName, getColumnFilterMatcher());
```
jt2594838, line 489: Is it possible to prune during the first construction so that no unnecessary objects will be created and copied.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/broker/consensus/ConsensusPrefetchingQueue.java
```diff
@@ -1813,6 +1819,25 @@ private boolean createAndEnqueueEvent(
...
+      if (Objects.nonNull(tablet) && Objects.nonNull(tablet.getTableName())) {
+        tableMap.put(
+            tablet.getTableName(), matcher.isTimeSelected(databaseName, tablet.getTableName()));
+      }
+    }
+    return tableMap.isEmpty()
+        ? Collections.emptyMap()
+        : Collections.singletonMap(databaseName, tableMap);
+  }
+
```
jt2594838, line 1877: Why not pass down the matcher directly?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/BoundColumnFilter.java
```diff
@@ -0,0 +1,146 @@
...
+    return MATCH_ALL;
+  }
+
+  public static BoundColumnFilter of(
+      final Map<TableKey, Set<String>> selectedColumnsByTable,
+      final boolean timeSelected,
+      final Map<TableKey, Boolean> timeSelectedByTable) {
+    final Map<TableKey, Set<String>> copied = new HashMap<>();
+    selectedColumnsByTable.forEach(
+        (key, value) -> copied.put(key, Collections.unmodifiableSet(new HashSet<>(value))));
```
jt2594838, line 60: Check the necessity of copying

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/ColumnFilterBinder.java
```diff
@@ -0,0 +1,166 @@
...
+    final Set<String> selectedColumnNames = new HashSet<>();
+    final Set<String> tagColumnNames = new HashSet<>();
+
+    for (final TsTableColumnSchema columnSchema : table.getColumnList()) {
+      if (Objects.isNull(columnSchema)) {
+        continue;
+      }
+      if (columnSchema.getColumnCategory() == TsTableColumnCategory.TAG) {
+        tagColumnNames.add(BoundColumnFilter.normalize(sourceColumnName(table, columnSchema)));
+      }
```
jt2594838, line 97: Continue？

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/ColumnFilterEvaluator.java
```diff
@@ -0,0 +1,179 @@
...
+        node.getEscape().map(expression -> ((StringLiteral) expression).getValue()).orElse(null);
+    return compileLikePattern(pattern, escape).matcher(left).matches();
+  }
+
+  @Override
+  public Boolean visitFunctionCall(final FunctionCall node, final ColumnMetadata context) {
+    final String left = fieldValue((Identifier) node.getArguments().get(0), context);
+    final String pattern = ((StringLiteral) node.getArguments().get(1)).getValue();
+    return Pattern.compile(pattern, Pattern.CASE_INSENSITIVE).matcher(left).matches();
+  }
```
jt2594838, line 120: Is this for a regex?
Is the function name check?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/ColumnFilterEvaluator.java
```diff
@@ -0,0 +1,179 @@
...
+      } else {
+        regex.append(Pattern.quote(String.valueOf(ch)));
+      }
+      escaping = false;
+    }
+    if (escaping) {
+      throw new IllegalArgumentException("LIKE pattern ends with escape character");
+    }
+    return Pattern.compile(regex.toString(), Pattern.CASE_INSENSITIVE);
+  }
```
jt2594838, line 159: How is LIKE processed by the query engine? Is it possible to reuse it?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/ColumnFilterMatcher.java
```diff
@@ -0,0 +1,204 @@
...
+  public boolean isTimeSelected() {
+    if (isMatchAll()) {
+      return true;
+    }
+    if (Objects.nonNull(boundColumnFilter)) {
+      return boundColumnFilter.isTimeSelected();
+    }
+    if (Objects.nonNull(selectedColumnNames)) {
+      return selectedColumnNames.contains("time");
+    }
```
jt2594838, line 114: Is the renamed time column properly processed here?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/ColumnFilterMatcher.java
```diff
@@ -0,0 +1,204 @@
...
+            });
+    if (result.isEmpty()) {
+      return Collections.emptyMap();
+    }
+
+    final Map<String, Map<String, Boolean>> copied = new HashMap<>();
+    result.forEach(
+        (database, tableMap) -> copied.put(database, Collections.unmodifiableMap(tableMap)));
+    return Collections.unmodifiableMap(copied);
+  }
```
jt2594838, line 164: Result is already a local variable, why copy it?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/ColumnFilterParser.java
```diff
@@ -0,0 +1,274 @@
...
+  Expression parse(final String rawColumnFilter) {
+    if (rawColumnFilter == null || rawColumnFilter.trim().isEmpty()) {
+      throw new ParsingException("column-filter should not be empty", null, 1, 1);
+    }
+    validateUnsupportedSyntax(rawColumnFilter);
+
+    final ColumnFilterLexer lexer = new ColumnFilterLexer(CharStreams.fromString(rawColumnFilter));
+    final CommonTokenStream tokenStream = new CommonTokenStream(lexer);
+    final org.apache.iotdb.db.relational.grammar.sql.ColumnFilterParser parser =
+        new org.apache.iotdb.db.relational.grammar.sql.ColumnFilterParser(tokenStream);
```
jt2594838: ColumnFilterParser -> AntlrColumnFilterParser to avoid name conflict?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/ColumnFilterParser.java
```diff
@@ -0,0 +1,274 @@
...
+      };
+
+  public Expression parseAndValidate(final String rawColumnFilter) throws SubscriptionException {
+    try {
+      final Expression expression = parse(rawColumnFilter);
+      ColumnFilterValidator.validate(expression);
+      return expression;
+    } catch (final ParsingException | IllegalArgumentException e) {
+      throw new SubscriptionException(
+          String.format("Invalid column-filter: %s", e.getMessage()), e);
```
jt2594838: i18n, check other places

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/ColumnFilterParser.java
```diff
@@ -0,0 +1,274 @@
...
+        }
+        throw new ParsingException("unsupported comparison operator '<'", null, 1, i + 1);
+      }
+      if (ch == '>') {
+        throw new ParsingException("unsupported comparison operator '>'", null, 1, i + 1);
+      }
+      if (ch == '+') {
+        throw new ParsingException("unexpected character '+'", null, 1, i + 1);
+      }
+    }
```
jt2594838, line 171: What if these are part of the column name?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/ColumnFilterParser.java
```diff
@@ -0,0 +1,274 @@
...
+
+      final Identifier field = toIdentifier(context.field());
+      if (Objects.nonNull(context.comparisonOperator())) {
+        return new ComparisonExpression(
+            Objects.nonNull(context.comparisonOperator().EQ())
+                ? ComparisonExpression.Operator.EQUAL
+                : ComparisonExpression.Operator.NOT_EQUAL,
+            field,
+            toStringLiteral(context.string(0)));
+      }
```
jt2594838, line 235: What does `validateUnsupportedSyntax` exactly do?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/ColumnFilterValidator.java
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  @Override
+  public Void visitFunctionCall(final FunctionCall node, final Void context) {
+    if (!REGEXP_LIKE.equalsIgnoreCase(node.getName().toString())
+        || node.isDistinct()
+        || node.getProcessingMode().isPresent()
+        || node.getArguments().size() != 2) {
+      throw invalid("only REGEXP is supported as regexp_like(field, pattern)");
+    }
```
jt2594838, line 124: Is it necessary to perform multi-pass validation and parsing? Can we squash them into one pass?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/TabletColumnPruner.java
```diff
@@ -0,0 +1,173 @@
...
+
+  public static Tablet pruneTableModelTablet(
+      final Tablet tablet, final String databaseName, final ColumnFilterMatcher matcher) {
+    if (Objects.isNull(tablet) || Objects.isNull(databaseName)) {
+      return tablet;
+    }
+
+    final ColumnFilterMatcher effectiveMatcher =
+        Objects.nonNull(matcher) ? matcher : ColumnFilterMatcher.matchAll();
+    final List<IMeasurementSchema> schemas = tablet.getSchemas();
```
jt2594838, line 52: If matchAll, return the tablet immediately.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/TabletColumnPruner.java
```diff
@@ -0,0 +1,173 @@
...
+
+    if (effectiveMatcher.shouldAutoRetainTagsAtRuntime()) {
+      for (int i = 0; i < schemas.size(); i++) {
+        if (isValidColumn(schemas, values, i)
+            && categories.get(i) == ColumnCategory.TAG
+            && hasValueArray(values, i)) {
+          selectedColumns[i] = true;
+        }
+      }
+    }
```
jt2594838, line 95: Merge the last if into the first one?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/TabletColumnPruner.java
```diff
@@ -0,0 +1,173 @@
...
+
+    return new Tablet(
+        tablet.getTableName(),
+        prunedSchemas,
+        prunedCategories,
+        tablet.getTimestamps(),
+        prunedValues,
+        prunedBitMaps,
+        tablet.getRowSize());
+  }
```
jt2594838, line 132: May add Tablet.pruneByIndex() for future reusing.

iotdb-core/datanode/src/test/java/org/apache/iotdb/db/subscription/columnfilter/ColumnFilterBinderTest.java
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  @Test
+  public void testAttributeMatchBindsTagsButNotAttribute() {
+    final BoundColumnFilter boundFilter =
+        new ColumnFilterBinder()
+            .bind(
+                createTableTopicConfig("category = \"ATTRIBUTE\""),
+                Collections.singletonMap(
+                    "db", Collections.singletonMap("sensors", createTableSchema())));
```
jt2594838, line 72: Should we directly reject this during parsing?

iotdb-core/relational-grammar/src/main/antlr4/org/apache/iotdb/db/relational/grammar/sql/SubscriptionColumnFilter.g4
jt2594838, line 1: ColumnFilter -> SubscriptionColumnFilter

Or add "Subscription" as one package level.

iotdb-core/relational-grammar/src/main/antlr4/org/apache/iotdb/db/relational/grammar/sql/SubscriptionColumnFilter.g4
```diff
@@ -0,0 +1,88 @@
...
+predicate
+    : booleanValue
+    | field comparisonOperator string
+    | field NOT? IN '(' string (',' string)* ')'
+    | field NOT? LIKE string (ESCAPE string)?
+    | field NOT? REGEXP string
+    | field IS NOT? NULL
+    | '(' booleanExpression ')'
+    ;
+
```
jt2594838, line 44: Function call is not present in the list.
Check and remove useless logic.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/broker/consensus/ConsensusLogToTabletConverter.java
```diff
@@ -512,51 +501,32 @@ private List<Tablet> convertRelationalInsertTabletNode(final RelationalInsertTab
...
-    final Object[] columns = node.getColumns();
-    final BitMap[] bitMaps = node.getBitMaps();
-    final int rowCount = node.getRowCount();
-    final List<Integer> matchedColumnIndices =
-        getMatchedTableColumnIndices(
-            measurements, dataTypes, columns, node.getColumnCategories(), true);
-    if (matchedColumnIndices.isEmpty()) {
+    if (Objects.isNull(measurements)
+        || Objects.isNull(dataTypes)
+        || Objects.isNull(node.getColumns())) {
```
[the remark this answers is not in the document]
Caideyipi: Fixed. `convertRelationalInsertTabletNode` now validates null/misaligned measurement/type/column arrays, skips malformed columns with warning logs, keeps schemas/columns/bitmaps aligned, and returns empty when no valid column remains. Added a regression test for malformed tablet-node columns.

integration-test/src/test/java/org/apache/iotdb/subscription/it/dual/tablemodel/IoTDBSubscriptionColumnFilterIT.java
Caideyipi: Removed that binder-only case from the IT. The binding-only coverage stays in unit tests.

integration-test/src/test/java/org/apache/iotdb/subscription/it/dual/tablemodel/IoTDBSubscriptionColumnFilterIT.java
Caideyipi: Added the actual subscription check: the IT now creates a consumer, inserts tree rows, polls record messages, and asserts that only the expected source column/tag are delivered while the view alias and unrelated field are absent.

iotdb-client/subscription/src/main/java/org/apache/iotdb/rpc/subscription/payload/poll/SubscriptionPollResponse.java
Caideyipi: Fixed. The serialized fields are no longer declared `transient`.

iotdb-client/subscription/src/main/java/org/apache/iotdb/rpc/subscription/payload/poll/SubscriptionPollResponse.java
Caideyipi: Kept a defensive immutable copy at the response boundary and normalize db/table keys while copying. This avoids caller-side mutation after analysis and lets the consumer side use direct lookups.

iotdb-client/subscription/src/main/java/org/apache/iotdb/session/subscription/payload/SubscriptionRecordHandler.java
Caideyipi: Fixed. `SubscriptionRecordHandler` now validates `tablet`, `databaseName`, and `tablet.getTableName()` before reading from the map.

iotdb-client/subscription/src/main/java/org/apache/iotdb/session/subscription/payload/SubscriptionRecordHandler.java
Caideyipi: Fixed. `timeSelectedByTable` db/table keys are normalized to lowercase when copied, and the handler now performs direct lowercase lookups instead of scanning case-insensitively.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/subscription/SubscriptionInfo.java
Caideyipi: Fixed for consistency. Known topic config keys are now checked case-insensitively for duplicates, not only `column-filter`; added a `mode` duplicate-key validation test as coverage.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/table/DataNodeTableCache.java
Caideyipi: Fixed. The snapshot collector now uses `HashMap` instead of `ConcurrentHashMap`.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/agent/SubscriptionTopicAgent.java
Caideyipi: ConfigNode validation rejects create/alter of tree topics with `column-filter`. The DataNode-side drop is kept defensive for stale/replayed/hot-updated metadata so a tree topic cannot retain an old column-filter matcher.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/agent/SubscriptionTopicAgent.java
Caideyipi: The hot-refresh comparison normalizes the relevant values before comparing. I kept the original topic attributes unchanged here because `TopicConfig`/`PipeParameters` own attribute lookup semantics, and mutating pushed coordinator metadata in the handler would be broader than needed.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/broker/consensus/ConsensusLogToTabletConverter.java
Caideyipi: Partially folded into construction for tablet nodes: malformed columns are filtered while building the tablet arrays so schemas/columns/bitmaps stay aligned. The actual column-filter pruning remains in `TabletColumnPruner` so row/tablet paths share the same tag/attribute/time-only behavior.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/broker/consensus/ConsensusPrefetchingQueue.java
Caideyipi: I kept the broker lookup here intentionally so the queue observes the latest matcher after topic-meta hot refresh. Passing a matcher captured earlier could become stale while the queue keeps running.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/BoundColumnFilter.java
Caideyipi: The copy is intentional. `BoundColumnFilter` is cached by the broker and reused by matchers, so it should be an immutable snapshot of the binder result rather than retaining caller-owned mutable maps/sets.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/ColumnFilterBinder.java
Caideyipi: Yes, this is intentional. Null column schemas are skipped so one malformed/unknown column does not fail binding for the whole table; only valid matched columns contribute to the bound result.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/ColumnFilterEvaluator.java
Caideyipi: `REGEXP` is represented internally as a `FunctionCall(regexp_like, ...)`. The validator checks the function name, argument count, and pattern before evaluation, so the evaluator only sees validated `regexp_like` calls.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/ColumnFilterEvaluator.java
Caideyipi: The query engine LIKE path has execution/runtime dependencies, while this evaluator only matches a few in-memory metadata strings. I kept a small SQL-like pattern compiler here to avoid coupling subscription metadata filtering to query execution internals.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/ColumnFilterMatcher.java
Caideyipi: Yes. Binding evaluates the filter against each table schema, including the visible time column name, and records per-table time selection. The existing binder test covers `time` versus a renamed `event_time` table.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/ColumnFilterMatcher.java
Caideyipi: Fixed. The extra local copy was removed; the local result/table maps are wrapped as unmodifiable directly.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/ColumnFilterParser.java
Caideyipi: Kept the subscription facade name as `ColumnFilterParser`; the generated ANTLR parser is referenced with its fully qualified name where needed to avoid ambiguity. Renaming the grammar would create broader generated-class churn, so I left that as-is.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/ColumnFilterParser.java
Caideyipi: Checked the surrounding subscription/config validation path. These parser/validator messages follow the existing subscription exception style in this area; I kept them local rather than adding a new i18n surface for the column-filter parser in this PR.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/ColumnFilterParser.java
Caideyipi: Fixed. The unsupported-syntax scanner now skips quoted identifiers/literals, including escaped double quotes, so `<`, `>`, and `+` inside column names or string literals are not rejected. Added a regression test.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/ColumnFilterParser.java
Caideyipi: `validateUnsupportedSyntax` is a pre-parser guard for clearer errors around unsupported operators/common mistakes before the ANTLR parse. It now ignores quoted text so legal quoted names/literals are not rejected by this guard.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/ColumnFilterValidator.java
Caideyipi: I kept parsing and semantic validation separated. The parser builds the AST; the validator enforces the supported metadata fields, predicates, and pattern validity. Squashing them would push semantic checks into the grammar visitor and make the diagnostics harder to keep consistent.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/TabletColumnPruner.java
Caideyipi: Fixed. `matchAll` now returns the original tablet immediately, with test coverage.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/TabletColumnPruner.java
Caideyipi: I kept the tag-retention pass separate from the matched-column pass because they apply different conditions: matched non-attribute columns are selected first, then tags are auto-retained only for runtime expression matching. Keeping them separate avoids mixing those two rules.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/TabletColumnPruner.java
Caideyipi: Agree this could be useful as a reusable `Tablet` API. For this PR I kept pruning local to avoid changing the TsFile-facing API/dependency surface; it can be extracted once there is another caller.

iotdb-core/datanode/src/test/java/org/apache/iotdb/db/subscription/columnfilter/ColumnFilterBinderTest.java
Caideyipi: I kept this allowed because the filter language describes column metadata, not only output columns. Matching `ATTRIBUTE` can be useful as a predicate to select the table/retain tags, while the pruner still excludes attribute columns from emitted tablets.

iotdb-core/relational-grammar/src/main/antlr4/org/apache/iotdb/db/relational/grammar/sql/SubscriptionColumnFilter.g4
Caideyipi: Kept the grammar name for now to avoid broad generated-parser churn. The runtime facade already avoids collision by referencing the generated parser with a fully qualified name. I can do the grammar/package rename separately if you prefer.

iotdb-core/relational-grammar/src/main/antlr4/org/apache/iotdb/db/relational/grammar/sql/SubscriptionColumnFilter.g4
Caideyipi: Generic function calls are not exposed by the user grammar. The `FunctionCall` path is only the internal AST representation for `REGEXP` as `regexp_like(field, pattern)`; arbitrary function-call syntax is rejected before evaluation.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/agent/SubscriptionTopicAgent.java
jt2594838: Add comments for all defensive logic to avoid confusion.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/columnfilter/ColumnFilterParser.java
jt2594838: Fix it now, do not add new info string that is out-of-i18n

iotdb-core/relational-grammar/src/main/antlr4/org/apache/iotdb/db/relational/grammar/sql/SubscriptionColumnFilter.g4
jt2594838: Do the rename.
Otherwise, it sounds like something for the query engine.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/agent/SubscriptionTopicAgent.java
Caideyipi: Added comments around the defensive DataNode-side drop. The comment now explains that ConfigNode rejects column-filter on tree topics and this branch is for stale or replayed metadata after table-topic to tree-topic transitions.

iotdb-core/relational-grammar/src/main/antlr4/org/apache/iotdb/db/relational/grammar/sql/SubscriptionColumnFilter.g4
Caideyipi: Done. Renamed the grammar to SubscriptionColumnFilter, including the file name, grammar name, root rule, and generated-parser references used by the subscription facade.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/agent/SubscriptionBrokerAgent.java
```diff
@@ -660,17 +665,24 @@ public void refreshConsensusQueueOrderMode(final String topicName, final String
   public void refreshColumnFilter(final String topicName, final TopicConfig topicConfig) {
     final ColumnFilterMatcher matcher;
     try {
+      final Map<String, Map<String, TsTable>> bindingTables =
+          getColumnFilterBindingTables(topicConfig);
+      if (Objects.isNull(bindingTables)) {
+        topicNameToColumnFilterMatcher.remove(topicName);
+        LOGGER.info(
+            "SubscriptionBrokerAgent: postpone refreshing column-filter matcher for topic [{}] because its table schema is not available locally",
+            topicName);
```
jt2594838: i18n

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/agent/SubscriptionBrokerAgent.java
```diff
@@ -660,17 +665,24 @@ public void refreshConsensusQueueOrderMode(final String topicName, final String
...
       matcher =
           ColumnFilterMatcher.fromBoundColumnFilter(
-              columnFilterBinder.bind(
-                  topicConfig, DataNodeTableCache.getInstance().getTableSnapshot()));
+              columnFilterBinder.bind(topicConfig, bindingTables));
     } catch (final Exception e) {
       LOGGER.warn(
           "SubscriptionBrokerAgent: failed to refresh column-filter matcher for topic [{}], use empty matcher to fail closed",
           topicName,
           e);
```
jt2594838: i18n

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/agent/SubscriptionBrokerAgent.java
Caideyipi: Fixed in 886ec2c98d4. The schema-unavailable column-filter matcher log now uses DataNodeMiscMessages, with both en and zh entries added.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/subscription/agent/SubscriptionBrokerAgent.java
Caideyipi: Fixed in 886ec2c98d4. The refresh-failure column-filter matcher log now uses DataNodeMiscMessages, and I moved the related success/lazy-refresh/drop matcher logs to the same i18n constants as well.

jt2594838 (approved):

PR #17938, opened 2026-06-15
Fix aggregate write-back output database metadata

Caideyipi:
## Description
This patch fixes tree-model aggregate-processor output events when processor.output.database is configured and the sink is write-back-sink.

Previously, aggregate output tablets rewrote the device path to the configured output database, but the raw tablet event database metadata could still inherit the source data region database. During write-back, partition lookup could then use mismatched database metadata and hit an NPE in DataPartition.

Changes:
- Preserve and allow overriding raw table/tree database metadata in raw tablet collectors.
- Make AggregateProcessor set generated tree output events to the configured output database.
- Add defensive DataPartition null checks to avoid unclear null-map NPEs.
- Add a unit test covering overridden tree database metadata on collected row output.

## Verification
- git diff --check

Targeted test was attempted but this local checkout cannot complete it: direct datanode test resolution misses local 2.0.10.1 artifacts; -am reaches upstream modules but fails before datanode in unrelated calc-commons generated NextFill compilation.

### DIFF
--- a/iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/partition/DataPartition.java
+++ b/iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/partition/DataPartition.java
@@ -221,8 +221,20 @@ public class DataPartition extends Partition {
     final List<TRegionReplicaSet> dataRegionReplicaSets = new ArrayList<>();
     final Map<TSeriesPartitionSlot, Map<TTimePartitionSlot, List<TRegionReplicaSet>>>
         dataBasePartitionMap = dataPartitionMap.get(databaseName);
+    if (dataBasePartitionMap == null) {
+      throw new RuntimeException(
+          "Database "
+              + databaseName
+              + " not exists and failed to create automatically because enable_auto_create_schema is FALSE.");
+    }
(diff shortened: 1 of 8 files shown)


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/partition/DataPartition.java
```diff
@@ -221,8 +221,20 @@ public List<TRegionReplicaSet> getDataRegionReplicaSetForWriting(
     final List<TRegionReplicaSet> dataRegionReplicaSets = new ArrayList<>();
     final Map<TSeriesPartitionSlot, Map<TTimePartitionSlot, List<TRegionReplicaSet>>>
         dataBasePartitionMap = dataPartitionMap.get(databaseName);
+    if (dataBasePartitionMap == null) {
+      throw new RuntimeException(
+          "Database "
+              + databaseName
+              + " not exists and failed to create automatically because enable_auto_create_schema is FALSE.");
```
jt2594838: i18n

Caideyipi: Fixed

PR #17941, opened 2026-06-15
Fix compaction writer size checkpointing

shuwenwei:
## Description

This PR fixes fast compaction writer checkpointing for variable-length data.

The previous implementation checked chunk size primarily by point-count checkpoints. For large TEXT/STRING/BLOB/OBJECT values, the writer could accumulate a large PageWriter/ChunkWriter buffer before the next point checkpoint. This patch adds per-subtask written-point-size checkpoint tracking and lets size checkpoints trigger the existing chunk-size check.

It also keeps checkpoint state isolated per sub compaction task and updates size accounting for direct page-flush paths so large binary pages can trigger checks promptly.

## Changes

- Track written point total size per subtask in compaction writers.
- Trigger chunk-size checks when either the point checkpoint or size checkpoint is reached.
- Enable size checkpointing only for chunk writers containing binary types.
- Account for direct non-aligned and aligned page flushes using serialized page sizes.
- Add a focused unit test proving binary-size checkpoints trigger chunk-size checks before point checkpoints.

## Validation

- Ran `git diff --cached --check` before commit.
- Ran `git diff --check` during local verification.
- Did not run Maven, spotless, or formatting checks.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/utils/writer/AbstractCompactionWriter.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/dataregion/compaction/execute/utils/writer/AbstractCompactionWriter.java
@@ -77,7 +77,7 @@ public abstract class AbstractCompactionWriter implements AutoCloseable {
   @SuppressWarnings("squid:S1170")
   private final long checkPoint = (targetChunkPointNum >= 10 ? targetChunkPointNum : 10) / 10;
 
-  private long lastCheckIndex = 0;
+  private long[] lastCheckIndexArray = new long[subTaskNum];
 
   // if unsealed chunk size is lower then this, then deserialize next chunk no matter it is
   // overlapped or not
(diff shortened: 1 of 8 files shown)


shuwenwei:
I found one path where the new size checkpoint is updated but never consumed.

`AbstractCompactionWriter` now increments `writtenPointTotalSizeArray` from direct page flushes, and `checkChunkSizeAndMayOpenANewChunk` can use that size checkpoint to force a chunk-size check. However, `FastInnerCompactionWriter` returns immediately after these direct page flushes:

- `flushAlignedPage` after `flushAlignedPageToChunkWriter(...)`
- `flushBatchedValuePage` after `flushAlignedPageToChunkWriter(...)`
- `flushNonAlignedPage` after `flushNonAlignedPageToChunkWriter(...)`

Unlike `FastCrossCompactionWriter`, these methods do not call `checkChunkSizeAndMayOpenANewChunk(...)` after the page is appended. For inner fast compaction with large TEXT/STRING/BLOB/OBJECT pages, the new `writtenPointTotalSizeArray` can cross the checkpoint, but no chunk-size check runs, so the writer may still keep accumulating pages until a later chunk boundary or `endMeasurement`.

Could you call `checkChunkSizeAndMayOpenANewChunk(fileWriter, chunkWriters[subTaskId], subTaskId)` after each successful direct page flush in `FastInnerCompactionWriter`, and add a test that covers the inner-fast page flush path?

JackieTien97 (approved):

PR #17951, opened 2026-06-16
Support readTsFile table function for external TsFiles

shuwenwei:
## Description

This PR adds relational `readTsFile` table function support for querying external TsFiles.

Main changes:
- Add `readTsFile` TVF planning/analyze support and schema collection through `TsFileSchemaCollector`.
- Validate explicit file paths by opening them as TsFiles, while directory inputs recursively scan `.tsfile` files and skip invalid TsFile contents by magic-number validation.
- Add external TsFile scan and aggregation scan plan nodes/operators.
- Add `ExternalTsFileQueryResource` to manage external TsFile readers, task partitioning, run-file merge reads, temporary files, and execution-lifetime memory reservation cleanup.
- Route related user-facing messages through DataNode query i18n messages.
- Add UT/IT coverage for external TsFile query resources and the `readTsFile` table function path.

## Tests

Not run in this turn.


Caideyipi (commented):
I found two issues that should be fixed before merging:

1. External TsFile resources can leak if planning fails after the resource is created.

`RelationPlanner.java:1647` creates the `ExternalTsFileQueryResource`, and `ExternalTsFileQueryResource.java:106-108` increments external file-reader references. Those resources are only released through `QueryExecution.stopAndCleanup()` at `QueryExecution.java:395`. However, if `execution.start()` throws during logical planning, distribution planning, or scheduling, `Coordinator.java:340-343` only releases frontend memory and schema locks. A timeout or optimizer/distribution exception after `planExternalTsFileScan()` can therefore leave reader references and temporary directories alive. Please either make `QueryExecution.start()` transition to failed and clean up on thrown planning exceptions, or have `Coordinator.execution()` release `queryContext.releaseExternalTsFileQueryResources()` when `execution.start()` fails before normal cleanup can run.

2. `read_tsfile` silently drops ATTRIBUTE columns.

`TsFileSchemaCollector.java:218-229` only collects `TAG` and `FIELD` columns, and `build()` at `TsFileSchemaCollector.java:292-307` only emits time, tags, and fields. If an external TsFile table schema contains `ColumnCategory.ATTRIBUTE`, those columns disappear from the TVF output schema. The execution path also creates `AlignedDeviceEntry(deviceID, new Binary[0])` in `ExternalTsFileQueryResource.java:125`, so attribute values are not available later either. Since table model supports ATTRIBUTE columns, this should either be implemented end to end or rejected explicitly instead of returning an incomplete schema.

Caideyipi:
I found one new issue that should be fixed before merging:


ead_tsfile does not appear to have an authorization gate for reading external server-side files. It is registered as a built-in table function in TableMetadataImpl.java, and StatementAnalyzer.visitTableFunctionInvocation() directly analyzes it. ReadTsFileTableFunction.analyze() then reads the user-supplied PATHS on the DataNode filesystem to collect TsFile schemas. This is a different capability from selecting an IoTDB table: any SQL user who can invoke the table function can ask the DataNode process to inspect external local paths, except for the current data-dir exclusion.

Please add an explicit privilege boundary for this function, for example admin / SYSTEM / MAINTAIN or another clearly chosen system privilege, before any path access or schema collection happens.

Also, the two earlier issues still look not fully addressed: external TsFile resources can still leak if planning/start throws after the resource is created, and ATTRIBUTE columns are still silently dropped instead of being supported or rejected explicitly.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/fragment/FragmentInstanceContext.java
```diff
@@ -632,6 +642,48 @@ public void setDevicePathsToContext(Map<IDeviceID, DeviceContext> devicePathsToC
     this.devicePathsToContext = devicePathsToContext;
   }
 
+  public void addExternalTsFileQueryResource(
+      ExternalTsFileQueryResource externalTsFileQueryResource) {
+    this.externalTsFileQueryResource = externalTsFileQueryResource;
```
JackieTien97, line 651: check here, if externalTsFileQueryResource is null, in case of one FI contains two read_tsfile.(maybe won't happen now, but someone in future may chnage this, like we may put left and right child of Join in one FI)

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/source/relational/ExternalTsFileAggTableScanOperator.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+  }
+
+  @Override
+  public void close() throws Exception {
+    if (deviceTaskReader != null) {
+      deviceTaskReader.close();
+      deviceTaskReader = null;
+    }
+    super.close();
```
JackieTien97:
```suggestion
  Exception exception = null;
  try {
    if (deviceTaskReader != null) {
      deviceTaskReader.close();
    }
  } catch (Exception e) {
    exception = e;
  } finally {
    deviceTaskReader = null;
    try {
      super.close();
    } catch (Exception e) {
      if (exception == null) {
        exception = e;
      } else {
        exception.addSuppressed(e);
      }
    }
  }

  if (exception != null) {
    throw exception;
  }
```

JackieTien97 (changes requested):

JackieTien97 (approved):

PR #17952, opened 2026-06-16
feat: add integration test support for StreamNode

suchenglong:
## Description

This PR introduces the StreamClusterIT test category in integration-test module to support full cluster integration tests of StreamNode.
The new test suite independently verifies all core StreamNode functions without modifying existing cluster test logic or assertions.

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>


integration-test/src/main/java/org/apache/iotdb/it/env/cluster/env/AIEnv.java
```diff
@@ -19,20 +19,81 @@
...
+  protected void initExtraNodes(
+      final List<ConfigNodeWrapper> configNodeWrappers,
+      final List<DataNodeWrapper> dataNodeWrappers,
+      final String testClassName) {
+    String seedConfigNode = configNodeWrappers.get(0).getIpAndPortString();
+    int dataNodePort = dataNodeWrappers.get(0).getPort();
+    startAINode(seedConfigNode, dataNodePort, testClassName);
+  }
+
+  private void startAINode(
```
shuwenwei: I suggest not making `AIEnv` privately own the AINode startup logic. In future scenarios we may need an environment that starts both AINode and StreamNode, and keeping AINode initialization embedded in `AIEnv` would force us to duplicate that logic or introduce combined env classes such as `AIStreamEnv`.

suchenglong: My idea is that if we need to start both AINode and StreamNode simultaneously, we can create a new Env class to internally launch AINode and StreamNode together.How do you feel about this?

shuwenwei: Creating a combined Env can work, but I think we should avoid requiring that Env to reimplement the AINode startup logic. Could we at least make the AINode startup path reusable within the env framework? For example, we could extract it into a small helper/initializer, or at minimum make `startAINode` protected instead of private. This way a future Env that needs both AINode and StreamNode can reuse the existing AINode initialization logic instead of copying it.

suchenglong: fix it

shuwenwei (approved):

JackieTien97 (approved):

PR #17955, opened 2026-06-16
Fix PBTree flush for negative child address

Caideyipi:
Fix PBTree flush when an updated non-measurement child still has a negative segment address. The parent record now allocates a segment for that child before serialization, preventing later flush attempts from repeatedly failing with negative-address nodes. Also propagate failures from force flush so test helpers do not loop forever when a flush cannot make progress. Tests: .\mvnw.cmd -pl iotdb-core/datanode -DskipTests spotless:check passed. Attempted target UT with -am, but local build was blocked by generated thrift sources missing javax.annotation in iotdb-thrift.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/impl/pbtree/schemafile/pagemgr/PageManager.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/impl/pbtree/schemafile/pagemgr/PageManager.java
@@ -293,8 +293,6 @@ public abstract class PageManager implements IPageManager {
             .entrySet()) {
       child = entry.getValue();
       actualAddress = getTargetSegmentAddress(curSegAddr, entry.getKey(), cxt);
-      childBuffer = RecordUtils.node2Buffer(child);
-
       curPage = getPageInstance(SchemaFile.getPageIndex(actualAddress), cxt);
       if (curPage.getAsSegmentedPage().read(SchemaFile.getSegIndex(actualAddress), entry.getKey())
           == null) {
@@ -304,6 +302,19 @@ public abstract class PageManager implements IPageManager {
                 node.getName(),
                 entry.getKey()));
       }
+
+      if (!child.isMeasurement() && getNodeAddress(child) < 0) {
+        if (child.isDevice() && child.getAsDeviceMNode().isUseTemplate()) {
+          throw new MetadataException(
+              String.format(
+                  "Adding or updating children of device using template [%s] is NOT allowed.",
+                  child.getFullPath()));
+        }
+        short estSegSize = estimateSegmentSize(child);
+        long glbIndex = preAllocateSegment(estSegSize, cxt);
+        SchemaFile.setNodeAddress(child, glbIndex);
+      }
+      childBuffer = RecordUtils.node2Buffer(child);
 
       // prepare alias comparison
       if (child.isMeasurement() && child.getAsMeasurementMNode().getAlias() != null) {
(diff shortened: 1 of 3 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/impl/pbtree/schemafile/pagemgr/PageManager.java
```diff
@@ -305,6 +303,19 @@ private void writeUpdatedChildren(ICachedMNode node, SchemaPageContext cxt)
                 entry.getKey()));
       }
 
+      if (!child.isMeasurement() && getNodeAddress(child) < 0) {
+        if (child.isDevice() && child.getAsDeviceMNode().isUseTemplate()) {
```
HTHou: Could you clarify why the updated-child path rejects using-template devices here? `writeNewChildren()` allocates a segment for non-measurement children without this check, while this path will now fail if an updated template device still has a negative address. If the invariant is that such nodes can never reach this branch, a regression test or a short comment would make that clear; otherwise this should probably follow the new-child allocation behavior and use the existing i18n message for any real template-child write rejection.

HTHou (commented):
Thanks for the fix. The core change makes sense to me: `RecordUtils.node2Buffer(child)` serializes the child segment address, so the updated-child path needs to assign a valid segment address before serializing a non-measurement child.

I have two things I think should be addressed or clarified before merging:

1. Please add a regression test for the exact path fixed here: an updated non-measurement child whose segment address is still negative, followed by `forceFlushAndRelease()` / `scheduleFlushAll()`. This is a persistence/liveness bug and the PR currently has no test guarding against the negative address being serialized again.

2. There is still an adjacent liveness hole in `scheduleFlushAll()`: after a region is added to `flushingRegionSet`, the async task can return early when `store == null` or `!regionToStore.containsKey(regionId)`, before reaching `executeFlush()` and its `finally` block that removes the region from `flushingRegionSet`. Since this PR is also trying to prevent force-flush helpers from looping forever, it would be good to either clear `flushingRegionSet` in an outer task `finally`, or explicitly leave a follow-up if this is intentionally out of scope.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/impl/pbtree/schemafile/pagemgr/PageManager.java
```diff
@@ -305,6 +303,19 @@ private void writeUpdatedChildren(ICachedMNode node, SchemaPageContext cxt)
...
+          throw new MetadataException(
+              String.format(
+                  "Adding or updating children of device using template [%s] is NOT allowed.",
+                  child.getFullPath()));
+        }
```
jt2594838: i18n

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/impl/pbtree/schemafile/pagemgr/PageManager.java
Caideyipi: Addressed in e35299c138a. I added a regression test for the updated non-measurement child with a negative segment address, and changed the updated-child path to follow writeNewChildren(): allocate a segment before serialization without adding an extra template-device rejection in this branch. I also fixed the scheduleFlushAll()/scheduleFlush early-return path so flushingRegionSet is cleared from an outer finally even when the store is already closed or removed.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/mtree/impl/pbtree/schemafile/pagemgr/PageManager.java
Caideyipi: Addressed in e35299c138a. I removed the hard-coded template-device exception from this updated-child allocation branch and aligned it with writeNewChildren(), so there is no non-i18n message here now. The real template-child write rejection remains in SchemaFile.writeMNode() with the existing i18n message.

jt2594838 (approved):

PR #17956, opened 2026-06-16
feat(client-cpp): add SessionC DATE/BLOB support and RowRecord getters

hongzhi-gao:
## Summary

- Add `TSDate_C` for typed `TS_TYPE_DATE` inserts in the C API
- Add `ts_row_record_get_date_int32` and `ts_row_record_get_string_byte_length` for query result readback
- Fix `TS_TYPE_DATE` handling in typed insert (`toCharPtrVec` / `freeCharPtrVec`)
- Extend `c_rowDelete` IT to cover DATE/BLOB and the new RowRecord getters

## Test plan

- [x] `./session_c_tests "[c_rowDelete]"`
- [x] `mvn -P with-cpp -pl iotdb-client/client-cpp -am verify` (sessionCIT passed)

### DIFF
--- a/iotdb-client/client-cpp/src/include/SessionC.h
+++ b/iotdb-client/client-cpp/src/include/SessionC.h
@@ -105,7 +105,6 @@ typedef enum {
 
 typedef enum { TS_COL_TAG = 0, TS_COL_FIELD = 1, TS_COL_ATTRIBUTE = 2 } TSColumnCategory_C;
 
-/** Calendar date for TS_TYPE_DATE typed inserts; pass a pointer via values[]. */
 typedef struct {
   int year;
   int month;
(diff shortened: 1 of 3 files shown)


jt2594838 (approved):

Caideyipi (commented):
I found two issues before this is ready:

1. `TS_TYPE_BLOB` is still serialized through `strlen(src)` in `toCharPtrVec`, and the downstream insert buffer path also uses `strlen(values[i])`. That silently truncates real binary payloads containing `\0`, while the PR title/API suggests BLOB support. The new test only uses the ASCII value `blobXY`, so it does not catch this. Please add a length-aware BLOB write path/API, or narrow the claim and add coverage for embedded-NUL data.

2. `ts_row_record_get_date_int32` can throw through the C ABI for the valid boundary date `1000-01-01`. `parseDateExpressionToInt` allows year 1000, but `parseIntToDate(10000101)` treats that same value as the empty-date sentinel, so reading that date back produces an invalid `IoTDBDate`; calling the new getter then calls `parseDateExpressionToInt` and may throw out of a C-linkage function. The getter should not throw, or the sentinel/range conflict should be fixed, with a regression test for `1000-01-01`.

PR #17960, opened 2026-06-16
[IOTDB-17797] Support lateral column aliases in table SELECT list

### LINKED ISSUE #17797: [Feature] Allow referencing SELECT column aliases in GROUP BY / ORDER BY (and support lateral column alias in SELECT)

DaZuiZui:
### Support lateral column aliases in table SELECT lists

This PR implements Part 2 of #17797 for the table-model analyzer: later `SingleColumn` SELECT items can reference aliases explicitly defined by earlier `SingleColumn` items in the same SELECT list.

Examples now supported:

```sql
SELECT s1 AS x, x + 1 AS y FROM table1;
SELECT s1 AS x, x + 1 AS y FROM table1 ORDER BY y;
SELECT s1 AS x, row_number() OVER (ORDER BY x) AS rn FROM table1;
SELECT s1 AS x, avg(s2) OVER (PARTITION BY x) AS a FROM table1;
```

### Name-resolution behavior

The LCA rewrite is left-to-right and only applies to unqualified identifiers in later SELECT items. It does not rewrite qualified references such as `t.x`, dereference expressions such as `x.y`, or identifiers inside subqueries.

Resolution order is:

1. local source column
2. visible previous SELECT aliases
3. existing analyzer resolution

If multiple previous aliases with the same canonical name are visible and no local source column wins, the analyzer raises a clear ambiguity error. `AllColumns` and `COLUMNS(...)` do not register reusable LCA aliases.

`GROUP BY` alias reuse uses the LCA-rewritten SELECT expression, while `ORDER BY` keeps the existing output-alias precedence. `WHERE` and `HAVING` still do not see SELECT aliases.

For `CAST(value AS type)`, only `value` participates in LCA rewriting. Type names are not treated as alias references.

### Implementation notes

The original AST is kept unchanged. The analyzer records each `SingleColumn`'s semantic expression after LCA rewriting and uses it for type analysis, source-column tracking, output expression analysis, and GROUP BY alias reuse. Output field names without explicit aliases are still derived from the original SELECT item text rather than from the rewritten expression.

Inline window specifications in later SELECT items are rewritten and registered against the rewritten `FunctionCall` nodes before expression analysis. Referencing an alias whose expression contains a window function is rejected explicitly.

Fixes #17797

<hr>

This PR has:
- [x] been self-reviewed.

<hr>

- `StatementAnalyzer`: SELECT-list LCA rewrite, rewritten SELECT expression tracking, output-scope name handling, and window metadata registration for rewritten expressions.
- `ExpressionRewriter` / `ExpressionTreeRewriter`: support for leaf expression rewrite hooks used by LCA expression copying.
- `SelectAliasReuseTest`: coverage for LCA chaining, collisions, ambiguity, aggregates, GROUP BY/ORDER BY reuse, window specs, CAST type-name collisions, WHERE/HAVING isolation, and output field names.
- `relational/analyzer/README.md`: documented table-model SELECT alias resolution rules.

##### Tests

```bash
./mvnw spotless:apply -pl iotdb-core/datanode
./mvnw test -pl iotdb-core/datanode -Dtest=SelectAliasReuseTest
```

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/analyzer/StatementAnalyzer.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/analyzer/StatementAnalyzer.java
@@ -36,28 +36,19 @@ import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.AllRows;
 import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.ArithmeticBinaryExpression;
 import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.ArithmeticUnaryExpression;
 import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.BetweenPredicate;
-import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.BinaryLiteral;
-import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.BooleanLiteral;
 import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.Cast;
 import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.CoalesceExpression;
 import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.Columns;
 import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.ComparisonExpression;
-import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.CurrentDatabase;
-import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.CurrentTime;
-import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.CurrentUser;
-import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.DecimalLiteral;
 import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.DereferenceExpression;
-import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.DoubleLiteral;
 import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.Except;
 import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.ExistsPredicate;
 import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.Expression;
 import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.Extract;
 import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.FieldReference;
 import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.Fill;
-import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.FloatLiteral;
 import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.FrameBound;
 import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.FunctionCall;
-import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.GenericLiteral;
 import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.GroupBy;
 import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.GroupingElement;
 import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.GroupingSets;
@@ -82,7 +73,6 @@ import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.NaturalJoin;
 import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.Node;
 import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.NotExpression;
 import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.NullIfExpression;
-import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.NullLiteral;
 import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.Offset;
 import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.OrderBy;
 import org.apache.iotdb.commons.queryengine.plan.relational.sql.ast.Parameter;
@@ -257,6 +247,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Locale;
@@ -557,10 +548,6 @@ public class StatementAnalyzer {
         ExpressionTreeRewriter.rewriteWith(rewriter, expression), rewriter.getReferences());
   }
 
-  private static Expression copyExpression(Expression expression) {
-    return ExpressionTreeRewriter.rewriteWith(new DeepCopyExpressionRewriter(), expression);
-  }
-
   private static final class LateralColumnAliasRewrite {
     private final Expression expression;
     private final Map<NodeRef<Expression>, Expression> references;
@@ -583,8 +570,7 @@ public class StatementAnalyzer {
   private static final class LateralColumnAliasRewriter extends ExpressionRewriter<Void> {
     private final Scope scope;
     private final SelectAliasResolver visibleAliases;
-    private final ImmutableMap.Builder<NodeRef<Expression>, Expression> references =
-        ImmutableMap.builder();
+    private final Map<NodeRef<Expression>, Expression> references = new LinkedHashMap<>();
 
     private LateralColumnAliasRewriter(Scope scope, SelectAliasResolver visibleAliases) {
       this.scope = requireNonNull(scope, "scope is null");
@@ -592,7 +578,7 @@ public class StatementAnalyzer {
     }
 
     private Map<NodeRef<Expression>, Expression> getReferences() {
-      return references.buildOrThrow();
+      return ImmutableMap.copyOf(references);
     }
 
     @Override
@@ -613,9 +599,9 @@ public class StatementAnalyzer {
                 "Lateral column alias '%s' containing window function is not supported",
                 node.getValue()));
       }
-      Expression copiedExpression = copyExpression(selectAlias.get().getRewrittenExpression());
-      references.put(NodeRef.of(copiedExpression), selectAlias.get().getRewrittenExpression());
-      return copiedExpression;
+      Expression rewrittenExpression = selectAlias.get().getRewrittenExpression();
+      references.put(NodeRef.of(rewrittenExpression), rewrittenExpression);
+      return rewrittenExpression;
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


DaZuiZui:
I have completed the implementation for Issue #17797 Part 2: supporting lateral column aliases in the table-model SELECT list.

The PR is available here:

https://github.com/apache/iotdb/pull/17960

This PR adds left-to-right SELECT-list alias resolution, keeps local input columns at higher priority than SELECT aliases, avoids rewriting qualified expressions and subqueries, preserves WHERE/HAVING semantics, and keeps the existing ORDER BY alias behavior. It also includes analyzer tests and a plan-level test to ensure reused aliases do not cause duplicated projection computation.

I have verified the changes with:

./mvnw spotless:apply -pl iotdb-core/datanode
./mvnw -nsu test -pl iotdb-core/datanode -Dtest=SelectAliasReuseTest
The PR is ready for review. It will close #17797 after being merged.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/relational/analyzer/StatementAnalyzer.java
```diff
@@ -1670,11 +2185,25 @@ private SelectAnalysis analyzeSelect(QuerySpecification node, Scope scope) {
               outputPosition++;
             }
           } else {
+            SelectAliasLookup visibleAliases = visibleAliasBuilder.build();
```
JackieTien97: Nit / follow-up suggestion: this builds a fresh immutable `SelectAliasLookup` for every `SingleColumn`. Since the `SelectAliasLookup` constructor copies the whole current alias map and each alias list, analyzing a very wide SELECT list with many reusable aliases ends up doing repeated prefix copies, i.e. roughly O(N^2) alias-map copying.

This is probably fine for normal SELECT lists, but it may be worth avoiding the repeated snapshots by using an incremental/shared immutable lookup structure, or by exposing a read-only lookup view over the builder whose visible state grows left-to-right.

Could you also add IT/regression coverage for this area? In particular, it would be useful to cover a wide SELECT-list LCA chain/reuse case, plus the edge cases around delimited/quoted alias case sensitivity, named `WINDOW` definitions not seeing SELECT aliases, `DISTINCT` with LCA and no `ORDER BY`, and LCA references in window frame bounds.

DaZuiZui: Thanks for the suggestion. Addressed in 3bd3a8d445: SELECT-list LCA rewriting now reads from a builder-backed SelectAliasResolver while walking the SELECT list, so it avoids rebuilding immutable snapshots for each SingleColumn and only snapshots once for later clauses. I also added regression coverage for a wide SELECT-list LCA chain, delimited alias case sensitivity, named WINDOW definitions not seeing SELECT aliases, DISTINCT with LCA and no ORDER BY, and LCA references in window frame bounds.

Verified with:

```
./mvnw spotless:apply -pl iotdb-core/datanode
./mvnw -nsu test -pl iotdb-core/datanode -Dtest=SelectAliasReuseTest
```

DaZuiZui:
Hi @JackieTien97 ,
Thanks for the careful review and helpful suggestions.

I have addressed the latest review comments in PR #17960. The SELECT-list LCA alias lookup now avoids rebuilding immutable snapshots for each SingleColumn, and I added regression coverage for the wide SELECT-list LCA chain, delimited alias case sensitivity, named WINDOW definitions, DISTINCT without ORDER BY, and LCA references in window frame bounds.

I verified the changes with:

./mvnw spotless:apply -pl iotdb-core/datanode
./mvnw -nsu test -pl iotdb-core/datanode -Dtest=SelectAliasReuseTest
Thanks again for the review.

JackieTien97 (changes requested):
**Request changes — performance/robustness defect on deep lateral-column-alias chains**

Thanks for this feature — functional/semantic behavior checks out (I ran all normal/boundary/exception cases end-to-end on a build containing this PR; the error-code/message deltas vs. the spec were doc-side and are being corrected there). One blocking issue on the **performance** side:

**Chained aliases cause ≈O(N²) logical-plan blowup, and an oversized `EXPLAIN (FORMAT JSON)` on deep chains.**

**Root cause:** in the LCA rewrite, `copyExpression(selectAlias.get().getRewrittenExpression())` inlines a deep copy of the *already-expanded* expression of the previous alias. For a chain `c0, c0+1 AS c1, c1+1 AS c2, …`, `cN`'s rewritten expression contains all of `c(N-1)`'s, so it becomes a depth-N tree — N columns ⇒ **O(N²)** plan-expression nodes. (This is a *different* O(N²) from the alias-snapshot one already addressed via `SelectAliasResolver`.)

**Measured** (2.0.7-SNAPSHOT + this PR, table dialect, `EXPLAIN ANALYZE (FORMAT JSON)` → `planStatistics`):

| variant (same width) | N | logicalPlanCostMs | `EXPLAIN (FORMAT JSON)` |
|---|---:|---:|---|
| **LCA chained** | 5 / 100 / 200 | **0.49 / 126.7 / 984.7** | N=200 → **305**, plan **12.96 MB > 12.58 MB** Thrift frame |
| non-chained control (`s1 + i`) | 100 / 200 | 0.63 / 1.02 | all OK (8.7 / 17 KB) |

Doubling N (100→200) multiplies logical-plan time ~7.8× (super-linear); the same-width non-chained SELECT stays linear and never approaches the frame limit — confirming the blowup is LCA-chaining-specific. Execution stays 1–3 ms; the cost is entirely in planning. Server-side error on the N=200 EXPLAIN-JSON: `TTransportException: Frame size (12959549) larger than protect max size (12582912)`.

**Repro** (table dialect):

```python
def chain(n):
    cols = ["s1 AS c0"] + [f"c{i-1} + 1 AS c{i}" for i in range(1, n+1)] + [f"c0 + c{n} AS final_value"]
    return "SELECT " + ", ".join(cols) + " FROM table1 WHERE time <= 3000"
```

```sql
CREATE DATABASE test; USE test;
CREATE TABLE table1 (device_id STRING TAG, s1 INT32 FIELD, s2 INT32 FIELD);
INSERT INTO table1(time,device_id,s1,s2) VALUES (1000,'d1',10,100),(2000,'d1',20,200),(3000,'d1',30,300);
-- then, for N = 5 / 100 / 200:
EXPLAIN ANALYZE (FORMAT JSON) <chain(N)>;   -- read planStatistics.logicalPlanCostMs
EXPLAIN (FORMAT JSON) <chain(N)>;           -- fails at N=200 with 305 / frame-size
```

Results are correct at all N (`final_value = 2*s1 + N`); this is purely a planning-cost / plan-size concern, not a correctness one.

**Impact:** benign for realistic short chains (N≤~10, sub-ms). Deep chains (N≥~100, e.g. machine-generated SQL) degrade sharply and can make `EXPLAIN (FORMAT JSON)` fail outright.

**Suggested directions:**
1. Represent chained aliases via shared subtrees/references in the plan instead of inlining deep copies (avoid O(N²)).
2. Chunk/stream large `EXPLAIN (FORMAT JSON)` responses, or relax the Thrift frame cap for them — note `EXPLAIN ANALYZE (FORMAT JSON)` is unaffected since it serializes statistics, not the expression trees.

DaZuiZui:
Thanks for the detailed review and the measurements.

I’ve addressed the performance/robustness issue you pointed out in the latest commit. The LCA rewrite no longer deep-copies/inlines the already-expanded expression from the previous alias. Instead, the planner reuses previous LCA projections through symbol mappings and appends SELECT projections in batches, so deep chained aliases should no longer cause the O(N^2) plan-expression / EXPLAIN JSON blowup.

I also added regression tests for this case, including:
- reusing a previous `CAST(...) AS x` projection without duplicating the Cast expression
- a 200-column chained LCA case that keeps plan expressions shallow and the plan JSON bounded

Verified with:

./mvnw test -pl iotdb-core/datanode -Dtest=SelectAliasReuseTest

All 40 tests passed. Please take another look when you have time.

DaZuiZui:
Hi @JackieTien97,

I have addressed the alias canonical matching behavior for the LCA changes in the latest commit. SELECT alias reuse now derives the alias key from `Identifier.getCanonicalValue()` and folds it to uppercase, so ordinary and delimited aliases are matched consistently for lateral alias reuse, e.g. `x`, `X`, `"x"`, and `"X"` all refer to the same SELECT alias name.

I also added regression coverage for exact delimited matches, case-differing delimited matches, delimited/unquoted mixed references, and duplicate canonical alias ambiguity.

Verified locally with:

```bash
./mvnw -nsu test -pl iotdb-core/datanode -Dtest=SelectAliasReuseTest,CreateTableStatementTest,TableConfigTaskVisitorTest
```

All 42 targeted tests passed.

PR #17975, opened 2026-06-17
Support hot reload for cluster runtime configs

CRZbulabula:
## Description

This PR adds hot-reload support for several cluster runtime configuration items managed by ConfigNode:

- `heartbeat_interval_in_ms`
- `continuous_query_min_every_interval_in_ms`
- `schema_region_group_extension_policy` and `data_region_group_extension_policy`
- `default_schema_region_group_num_per_database` and `default_data_region_group_num_per_database`
- `schema_region_per_data_node` and `data_region_per_data_node`
- `read_consistency_level`

The implementation updates ConfigNode and DataNode hot-reload paths, removes stale cached configuration snapshots from region extension and read-consistency paths, and reschedules heartbeat-related ConfigNode services when the heartbeat interval changes. Region-per-DataNode changes trigger max region group recalculation only on the current ConfigNode leader.

Load-balancing related parameters were also reviewed. Region extension policies and region-per-DataNode limits now support hot reload. `region_group_allocate_policy` and Ratis auto leader-balance settings remain restart-only because the corresponding balancer/allocator instances are selected during service initialization.

## Testing

- `mvn spotless:apply -pl iotdb-core/confignode`
- `mvn compile -pl iotdb-core/confignode,iotdb-core/datanode -am -DskipTests -DskipUTs`
- `mvn verify -DskipUTs -Dit.test=IoTDBSetConfigurationIT -DfailIfNoTests=false -Dfailsafe.failIfNoSpecifiedTests=false -Drat.skip=true -pl integration-test -am -P with-integration-tests`
- `git diff --check`

### DIFF
--- a/integration-test/src/main/java/org/apache/iotdb/it/env/cluster/config/MppCommonConfig.java
+++ b/integration-test/src/main/java/org/apache/iotdb/it/env/cluster/config/MppCommonConfig.java
@@ -466,13 +466,13 @@ public class MppCommonConfig extends MppBaseConfig implements CommonConfig {
   }
 
   @Override
-  public CommonConfig setDataRegionPerDataNode(double dataRegionPerDataNode) {
+  public CommonConfig setDataRegionPerDataNode(int dataRegionPerDataNode) {
     setProperty("data_region_per_data_node", String.valueOf(dataRegionPerDataNode));
     return this;
   }
 
   @Override
-  public CommonConfig setSchemaRegionPerDataNode(double schemaRegionPerDataNode) {
+  public CommonConfig setSchemaRegionPerDataNode(int schemaRegionPerDataNode) {
     setProperty("schema_region_per_data_node", String.valueOf(schemaRegionPerDataNode));
     return this;
   }
--- a/integration-test/src/main/java/org/apache/iotdb/it/env/cluster/config/MppSharedCommonConfig.java
+++ b/integration-test/src/main/java/org/apache/iotdb/it/env/cluster/config/MppSharedCommonConfig.java
@@ -479,14 +479,14 @@ public class MppSharedCommonConfig implements CommonConfig {
   }
 
   @Override
-  public CommonConfig setDataRegionPerDataNode(double dataRegionPerDataNode) {
+  public CommonConfig setDataRegionPerDataNode(int dataRegionPerDataNode) {
     dnConfig.setDataRegionPerDataNode(dataRegionPerDataNode);
     cnConfig.setDataRegionPerDataNode(dataRegionPerDataNode);
     return this;
   }
 
   @Override
-  public CommonConfig setSchemaRegionPerDataNode(double schemaRegionPerDataNode) {
+  public CommonConfig setSchemaRegionPerDataNode(int schemaRegionPerDataNode) {
     dnConfig.setSchemaRegionPerDataNode(schemaRegionPerDataNode);
     cnConfig.setSchemaRegionPerDataNode(schemaRegionPerDataNode);
     return this;
--- a/integration-test/src/main/java/org/apache/iotdb/it/env/remote/config/RemoteCommonConfig.java
+++ b/integration-test/src/main/java/org/apache/iotdb/it/env/remote/config/RemoteCommonConfig.java
@@ -339,11 +339,12 @@ public class RemoteCommonConfig implements CommonConfig {
   }
 
   @Override
-  public CommonConfig setDataRegionPerDataNode(double dataRegionPerDataNode) {
+  public CommonConfig setDataRegionPerDataNode(int dataRegionPerDataNode) {
     return this;
   }
 
-  public CommonConfig setSchemaRegionPerDataNode(double schemaRegionPerDataNode) {
+  @Override
+  public CommonConfig setSchemaRegionPerDataNode(int schemaRegionPerDataNode) {
     return this;
   }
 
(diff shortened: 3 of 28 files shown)


shuwenwei:
I have one concern about the new hot-reload behavior for `heartbeat_interval_in_ms`.

The PR correctly reschedules `HeartbeatService`, `StatisticsService`, `EventService`, and `RetryFailedTasksThread` when the interval changes. However, the Phi failure detector in `AbstractLoadCache` is constructed with `CONF.getHeartbeatIntervalInMs() * 200_000L`, so existing load caches may still keep the detector parameters derived from the old heartbeat interval after hot reload.

Could you confirm whether the failure detector threshold is expected to follow `heartbeat_interval_in_ms` hot reload as well? If yes, we probably need to rebuild/update the existing load-cache failure detectors when the heartbeat interval changes, or add a test/clarification showing that only the scheduling interval is intended to be hot-reloaded.

Caideyipi:
I found a few issues that look worth fixing before merge:

1. `heartbeat_interval_in_ms` hot reload is incomplete. The reload path reschedules `HeartbeatService`, `StatisticsService`, `EventService`, and retry tasks, but existing `LoadCache` node/region caches and `TopologyService` keep `PhiAccrualDetector` instances that were constructed with the old heartbeat interval. `PhiAccrualDetector.minHeartbeatStdNs` is final, so after changing the heartbeat interval the node/region/topology failure detection thresholds still use the previous interval. Relevant locations: `ConfigManager.java`, `LoadManager.java`, `AbstractLoadCache.java`, and `TopologyService.java`.

2. Rejected hot-reload values can still appear in `show configuration` as applied. Both ConfigNode and DataNode call `ConfigurationFileUtils.updateAppliedProperties(...)` before validating and applying the merged hot-reload properties. If a reload later fails, for example `set configuration heartbeat_interval_in_ms=-1`, the config file is not written, but `lastAppliedProperties` has already been updated, so `show configuration` can report a value that was not actually applied. Relevant locations: `ConfigNodeDescriptor.loadHotModifiedProps`, `IoTDBDescriptor.loadHotModifiedProps`, and `ConfigurationFileUtils.updateAppliedProperties`.

3. This PR changes `schema_region_per_data_node` and `data_region_per_data_node` from accepting numeric strings through `Double.parseDouble(...).intValue()` to strict `Integer.parseInt(...)`. Existing configs like `data_region_per_data_node=5.0`, which were previously accepted and match the old IT API type, will now fail startup or any ConfigNode hot reload because the merged config file is revalidated. If this behavior change is intentional, it should probably be called out; otherwise consider preserving backward compatibility.

CRZbulabula:
Thanks @shuwenwei, confirmed that the failure detector threshold is not intended to follow `heartbeat_interval_in_ms` hot reload in this PR. The hot reload path only reschedules heartbeat-related services.

I pushed a follow-up commit (`2e673122cc`) to make this explicit and avoid partial behavior: ConfigNode now keeps a startup heartbeat-interval snapshot for failure detector initialization, and `AbstractLoadCache` / `TopologyService` use that snapshot instead of the hot-reloaded heartbeat interval. Existing detectors are not rebuilt, and newly-created detectors after hot reload still use the startup value.

Verified with:
- `mvn spotless:apply -pl iotdb-core/confignode`
- `mvn compile -pl iotdb-core/confignode -DskipTests -DskipUTs`
- `git diff --check`

shuwenwei (approved):

CRZbulabula:
Thanks for the detailed review. I pushed `fecde7f229` to address the remaining issues.

What changed:
- Failure detector behavior: no further code change in this commit. As clarified in `2e673122cc`, failure detector thresholds are intentionally restart-only for this PR and now use the startup heartbeat interval snapshot. Hot reload only reschedules the heartbeat-related services.
- Rejected hot reload values: fixed the applied-property update ordering in both ConfigNode and DataNode so `show configuration` is only updated after a successful hot reload. I also fixed `ConfigManager.setConfiguration` to stop immediately when the local ConfigNode rejects a cluster-wide hot reload, so invalid values are not propagated to DataNodes or other ConfigNodes.
- Region-per-node parsing: preserved compatibility for integer-compatible decimal strings such as `2.0`, while rejecting non-integer decimals such as `1.9` instead of truncating them.

Added/updated IT coverage in `IoTDBSetConfigurationIT` for:
- rejected `heartbeat_interval_in_ms=-1` not polluting ConfigNode/DataNode `show configuration`;
- `schema_region_per_data_node` / `data_region_per_data_node` accepting `2.0` and rejecting `1.9`.

Verified with:
- `mvn spotless:apply -pl iotdb-core/confignode,iotdb-core/datanode`
- `mvn compile -pl iotdb-core/confignode,iotdb-core/datanode -DskipTests -DskipUTs`
- `mvn verify -DskipUTs -Dit.test=IoTDBSetConfigurationIT#testRejectedHotReloadDoesNotUpdateAppliedConfiguration+testHotReloadRegionGroupExtensionConfiguration -DfailIfNoTests=false -Dfailsafe.failIfNoSpecifiedTests=false -Drat.skip=true -pl integration-test -am -P with-integration-tests`
- `git diff --check`

Note: the same focused IT without `-Drat.skip=true` was blocked locally by an unrelated `.claude/scheduled_tasks.lock` RAT violation before tests started.

Caideyipi (approved):
Approved. I think the remaining ConfigNode config-file-missing edge case is non-blocking for this PR: when iotdb-system.properties is absent, ConfigManager.setConfiguration can apply the value in memory, set EXECUTE_STATEMENT_ERROR for the missing file warning, and then return before running the heartbeat/region/topology hot-reload hooks. That can leave the in-memory value changed without the corresponding reschedule/recalculation hook in this degraded path. Normal deployments with the system config file present are unaffected, and the rejected-value propagation/applied-configuration issues are fixed. It would still be cleaner to separate the missing-file warning from actual reload failure in a follow-up.

PR #17981, opened 2026-06-18
Prevent pipe sink task id from logging secrets

luoluoyuyu:
## Description
As the title said

<hr>

This PR has:
- [ ] been self-reviewed.
    - [ ] concurrent read
    - [ ] concurrent write
  for an unfamiliar reader.
  for code coverage.

<hr>

### DIFF
--- a/iotdb-api/pipe-api/src/main/java/org/apache/iotdb/pipe/api/customizer/parameter/PipeParameters.java
+++ b/iotdb-api/pipe-api/src/main/java/org/apache/iotdb/pipe/api/customizer/parameter/PipeParameters.java
@@ -441,7 +441,7 @@ public class PipeParameters {
       return value;
     }
 
-    private static boolean isHiddenKey(final String key) {
+    public static boolean isHiddenKey(final String key) {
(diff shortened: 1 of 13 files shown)


Caideyipi:
I found two issues that still look worth fixing:

1. `PipeSinkSubtaskManager` now restores the raw `attributeSortedString` into the internal `taskID`, which is needed for compatibility, but the same raw string is also passed into `PipeSinkSubtask` and later returned by `getAttributeSortedString()`. That value is used as the metrics `Tag.NAME` in `PipeDataRegionSinkMetrics` and `PipeSchemaRegionSinkMetrics`, so secrets such as `sink.scp.password=...` or `sink.ssl.trust-store-pwd=...` can still be exposed through metric labels. Please keep the raw id only for internal map/executor keys, and use the display/sanitized attribute string for metric labels.

2. `PipeSubtaskExecutor#getSafeSubtaskStr` is still used on the `SUBTASK_NOT_REGISTERED` fallback paths and only masks `password=...`. With the raw sink `taskID`, this can still log sensitive keys covered by `PipeParameters.ValueHider`, for example `ssl.trust-store-pwd`. It would be better to reuse the same hiding rule as `ValueHider`, or avoid logging raw ids in those fallback paths.

jt2594838 (approved):

PR #17985, opened 2026-06-18
Fix old pipe root user compatibility

Caideyipi:
## Description
When old pipe metadata created before the user and permission system is loaded, automatically enrich missing local IoTDB source and write-back sink credentials with the current root user and encrypted password.

This keeps upgraded old user pipes runnable while avoiding overwriting existing credentials, system pipes, or normal remote IoTDB thrift sinks.

## Tests
- `mvn -nsu -pl iotdb-core/node-commons,iotdb-core/confignode -Dtest=PipeTaskInfoAutoRestartTest "-Dsurefire.failIfNoSpecifiedTests=false" test`

### DIFF
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/pipe/PipeTaskInfo.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/pipe/PipeTaskInfo.java
@@ -80,7 +80,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
-import java.util.function.BiFunction;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
@@ -97,16 +97,17 @@ public class PipeTaskInfo implements SnapshotProcessor {
 
   // Pure in-memory object, not involved in snapshot serialization and deserialization.
   private final PipeTaskInfoVersion pipeTaskInfoVersion;
-  private final BiFunction<String, String, String> pipeUserPasswordProvider;
+  // Accepts a username and returns its current stored password for pipe authentication.
+  private final Function<String, String> pipeUserCurrentPasswordProvider;
 
   public PipeTaskInfo() {
     this(null);
   }
 
-  public PipeTaskInfo(final BiFunction<String, String, String> pipeUserPasswordProvider) {
+  public PipeTaskInfo(final Function<String, String> pipeUserCurrentPasswordProvider) {
     this.pipeMetaKeeper = new PipeMetaKeeper();
     this.pipeTaskInfoVersion = new PipeTaskInfoVersion();
-    this.pipeUserPasswordProvider = pipeUserPasswordProvider;
+    this.pipeUserCurrentPasswordProvider = pipeUserCurrentPasswordProvider;
   }
 
   /////////////////////////////// Lock ///////////////////////////////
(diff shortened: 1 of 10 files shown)


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/pipe/PipeTaskInfo.java
```diff
@@ -95,10 +97,16 @@ public class PipeTaskInfo implements SnapshotProcessor {
 
   // Pure in-memory object, not involved in snapshot serialization and deserialization.
   private final PipeTaskInfoVersion pipeTaskInfoVersion;
+  private final BiFunction<String, String, String> pipeUserPasswordProvider;
```
jt2594838: Explain the parameters

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/pipe/PipeTaskInfo.java
```diff
@@ -998,6 +1009,45 @@ public void processLoadSnapshot(final File snapshotDir) throws IOException {
...
+      return;
+    }
+    final boolean shouldEnrichSource = pipeStaticMeta.mayNeedCompatibleRootUserForIoTDBSource();
+    final boolean shouldEnrichSink = pipeStaticMeta.mayNeedCompatibleRootUserForWriteBackSink();
+    if (!shouldEnrichSource && !shouldEnrichSink) {
+      return;
+    }
+
+    final String rootUserName = CommonDescriptor.getInstance().getConfig().getDefaultAdminName();
+    final String password = pipeUserPasswordProvider.apply(rootUserName, null);
```
jt2594838: What is the point of the second parameter?

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/pipe/PipeTaskInfo.java
```diff
@@ -998,6 +1009,45 @@ public void processLoadSnapshot(final File snapshotDir) throws IOException {
...
+    if (Objects.isNull(password)) {
+      throw new PipeException(
+          String.format(
+              "Failed to enrich pipe %s with root user for compatibility because root user %s does not exist.",
+              pipeStaticMeta.getPipeName(), rootUserName));
```
jt2594838: i18n

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/pipe/PipeTaskInfo.java
Caideyipi: Addressed in 3801d49e817: changed this from BiFunction<String, String, String> to Function<String, String> and renamed it to pipeUserCurrentPasswordProvider. The single parameter now clearly means username, and the return value is the current stored password for pipe authentication.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/pipe/PipeTaskInfo.java
Caideyipi: Addressed in 3801d49e817: moved this error message into ConfigNodeMessages and added the constant to both en and zh i18n source sets.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/pipe/PipeTaskInfo.java
Caideyipi: Addressed in 3801d49e817: removed the second parameter from the pipe metadata layer. ConfigManager now adapts PermissionManager.login4Pipe with userName -> login4Pipe(userName, null), because this compatibility path only needs to fetch the current stored root password.

jt2594838 (approved):

PR #17987, opened 2026-06-18
C++ client: build with OpenSSL 3.x by default and bundle the runtime …

hongzhi-gao:
Enables SSL for the C++ session client out of the box and ships a self-contained package.

What
- Default WITH_SSL / with.ssl to ON (CMake + Maven), so the client and the CI packaging/verify jobs build with SSL by default.
- Bundle the OpenSSL shared libraries into the package lib/ (new InstallOpenSSLRuntime.cmake) so the SDK runs on machines without a system OpenSSL. iotdb_session gets an $ORIGIN / @loader_path runtime path, and the examples copy the bundled libs next to their binaries, so the package is self-contained (verified the example smoke test resolves the bundled libssl.so.3, not a system one).
- Bump the bundled Apache Thrift 0.21.0 -> 0.23.0.

OpenSSL version
- Use OpenSSL 3.x (Apache-2.0   [License | OpenSSL Library](https://openssl-library.org/source/license/)  ). OpenSSL 1.1.1 is EOL and not Apache-2.0, so it is never linked or redistributed.
- Note: OpenSSL 4.0 removed the legacy TLS-method APIs (TLSv1_method, SSLv3_method, ASN1_STRING_data) that Thrift's TSSLSocket still calls, so the CI pins OpenSSL 3.x rather than installing the latest.
- find_package(OpenSSL) uses the system OpenSSL when present; otherwise it builds OpenSSL 3.5.0 from source as shared libraries. The Linux packaging build forces the from-source build (-DIOTDB_OPENSSL_FROM_SOURCE=ON) because the manylinux_2_28 / AlmaLinux 8 baseline ships OpenSSL 1.1.1, preserving the glibc 2.28 floor while shipping 3.x.

Build tooling
- Build bison >= 3.7 from source when the system bison is too old for Thrift 0.23's grammar build (manylinux ships 3.0.4).
- Encode WITH_SSL in the bundled-Thrift build stamp and forward OPENSSL_ROOT_DIR so Thrift's TSSLSocket links the same OpenSSL that is bundled.

CI
- Install/point at OpenSSL 3.x (not 4.0): Windows uses a pinned prebuilt OpenSSL 3.5.x; macOS uses brew openssl@3; the manylinux job builds 3.x from source (installs the perl modules OpenSSL's Configure needs).

Tested
- https://github.com/hongzhi-gao/iotdb/actions/runs/28008675171
- Windows (MSVC / VS2022): iotdb_session.dll links and imports libssl-3-x64.dll / libcrypto-3-x64.dll, both bundled into lib/.
- Linux (Ubuntu via WSL): from-source OpenSSL 3.5.0 shared build, libiotdb_session.so has RUNPATH=$ORIGIN and the bundled libssl.so.3 / libcrypto.so.3 are resolved in preference to the system OpenSSL.


iotdb-client/client-cpp/examples/CMakeLists.txt
```diff
@@ -145,6 +167,12 @@ foreach(_t IN LISTS _example_targets)
           COMMAND ${CMAKE_COMMAND} -E copy_if_different
                   "${_iotdb_runtime}" $<TARGET_FILE_DIR:${_t}>
           COMMENT "Copy IoTDB runtime library next to ${_t}")
+      foreach(_ssl_lib IN LISTS _iotdb_sdk_ssl_runtime)
```
HTHou, line 170: This copies the bundled OpenSSL runtime libraries next to each example binary, but the `example-dist` target below still only stages the executables and `iotdb_session`. A user who builds `example-dist` and copies that `dist/` directory to a machine without OpenSSL will still fail at runtime. Could we also copy `_iotdb_sdk_ssl_runtime` into `_example_dist_dir` when it is populated?

HTHou (commented):
One follow-up from the C++ package review.

iotdb-client/client-cpp/examples/CMakeLists.txt
hongzhi-gao: example-dist now also stages _iotdb_sdk_ssl_runtime into dist/, so a copied dist runs without a system OpenSSL

jt2594838 (approved):

PR #17988, opened 2026-06-18
Supply the max_schema/data_region_group_num param to modify schema when create or alter database

zerolbsony:
Adjust the maximum quotas for database schemas based on the MAX_DATA_REGION_GROUP_NUM and MAX_SCHEMA_REGION_GROUP_NUM parameters:
`ALTER DATABASE root.sg WITH MAX_DATA_REGION_GROUP_NUM = 4, MAX_SCHEMA_REGION_GROUP_NUM = 3;`


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/schema/ClusterSchemaManager.java
```diff
@@ -908,14 +942,112 @@ public static TSStatus enrichDatabaseSchemaWithDefaultProperties(
...
+            consensusGroupType,
+            TConsensusGroupType.SchemaRegion.equals(consensusGroupType)
+                ? databaseSchema.getMaxSchemaRegionGroupNum()
+                : databaseSchema.getMaxDataRegionGroupNum(),
+            true);
+    return status.getCode() == TSStatusCode.SUCCESS_STATUS.getStatusCode() ? previousError : status;
+  }
+
+  private static TSStatus validateMaxRegionGroupNum(
+      final String database,
```
CRZbulabula: **Unused parameter.** The `database` parameter of `validateMaxRegionGroupNum` is never referenced in the method body (it reads policy/default/field-name purely from `consensusGroupType` and `CONF`). Likewise `validateMaxRegionGroupNumOnCreation` only forwards `databaseSchema.getName()` into it. Either drop the parameter, or use it in the error message for clearer diagnostics.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/schema/ClusterSchemaManager.java
```diff
@@ -908,14 +942,112 @@ public static TSStatus enrichDatabaseSchemaWithDefaultProperties(
...
+      final TSStatus previousError) {
+    final TSStatus status =
+        validateMaxRegionGroupNum(
+            databaseSchema.getName(),
+            consensusGroupType,
+            TConsensusGroupType.SchemaRegion.equals(consensusGroupType)
+                ? databaseSchema.getMaxSchemaRegionGroupNum()
+                : databaseSchema.getMaxDataRegionGroupNum(),
+            true);
+    return status.getCode() == TSStatusCode.SUCCESS_STATUS.getStatusCode() ? previousError : status;
```
CRZbulabula: **Error-aggregation pattern is replace, not accumulate.** `validateMaxRegionGroupNumOnCreation` returns `previousError` on success and `status` on failure, and the two call sites chain it as `errorResp = validateMaxRegionGroupNumOnCreation(schema, SchemaRegion, errorResp)` then again for `DataRegion`. If the SchemaRegion validation already produced a non-null error and the DataRegion validation also fails, the SchemaRegion error is silently overwritten (DataRegion wins). With current inputs only one error can occur so it's harmless today, but the pattern reads as if it accumulates when it actually replaces. The alter path already does the clearer `if (status != OK) return status;` per dimension — applying the same shape here on create would avoid the subtlety.

integration-test/src/test/java/org/apache/iotdb/db/it/utils/TestUtils.java
```diff
@@ -1216,6 +1216,7 @@ public static boolean tryExecuteNonQueriesWithRetry(
           }
         }
         connectionToUse = null;
+        statement = null;
```
CRZbulabula: **Unrelated drive-by change.** This `statement = null;` (and the PR's head commit "The old statement shall be reclaimed in the same manner as the original connection") is a statement/connection-lifecycle fix that is unrelated to the `MAX_*_REGION_GROUP_NUM` feature this PR is about. It looks like a legitimate resource-cleanup fix, but please either split it into its own PR or call it out in the PR description so reviewers and `git blame` aren't confused about why it's here.

CRZbulabula (commented):
A few non-blocking cleanups. One of them (the orphaned min-based ALTER validation) is on pre-existing lines that aren't part of this diff, so I'm noting it here rather than inline:

**Orphaned min-based ALTER validation** — `ClusterSchemaManager.java:242` and `:256` (`alterDatabase`): the `isSetMinSchemaRegionGroupNum()` / `isSetMinDataRegionGroupNum()` validation blocks (the "could only be increased" checks) are now unreachable from the SQL path, because after this PR no SQL statement ever sets `minSchemaRegionGroupNum`/`minDataRegionGroupNum` — the keyword now sets `max*` instead. These branches are now only hit by internal/Thrift callers that still populate the min fields. This isn't wrong, but it's easy to mistake for the active validation. Consider adding a comment clarifying it only applies to non-SQL callers, or confirm nothing else still sends `min*` via ALTER.

The inline comments below cover the remaining items.

integration-test/src/test/java/org/apache/iotdb/relational/it/schema/IoTDBDatabaseMaxRegionGroupNumIT.java
```diff
@@ -0,0 +1,100 @@
...
+            EnvFactory.getEnv().getConnection(BaseEnv.TABLE_SQL_DIALECT);
+        final Statement statement = connection.createStatement()) {
+      statement.execute(
+          "create database test_create with(max_schema_region_group_num=3, max_data_region_group_num=4)");
+      TestUtils.assertResultSetEqual(
+          statement.executeQuery(
+              "select database, max_schema_region_group_num, max_data_region_group_num "
+                  + "from information_schema.databases where database = 'test_create'"),
+          "database,max_schema_region_group_num,max_data_region_group_num,",
+          Collections.singleton("test_create,3,4,"));
```
CRZbulabula, line 90: To verify the demand number is set correctly, u should not only test the metadata (the number obtained via information_schema or show databases), but also write some data and check if the region groups are created successfully.

Not sure if we have these CIs currently, you should check and append when necessary.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/schema/ClusterSchemaManager.java
```diff
@@ -267,6 +268,27 @@ public TSStatus alterDatabase(
       }
     }
```
CRZbulabula: Both min region group num params are deprecated, you can delete them.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/schema/ClusterSchemaManager.java
```diff
@@ -507,60 +537,64 @@ public synchronized void adjustMaxRegionGroupNum() {
...
-      // The allocated SchemaRegionGroups will not be shrunk.
-      final int allocatedSchemaRegionGroupCount;
-      try {
-        allocatedSchemaRegionGroupCount =
-            getPartitionManager()
-                .getRegionGroupCount(databaseSchema.getName(), TConsensusGroupType.SchemaRegion);
-      } catch (final DatabaseNotExistsException e) {
-        // ignore the pre deleted database
-        continue;
+      int maxSchemaRegionGroupNum = databaseSchema.getMaxSchemaRegionGroupNum();
```
CRZbulabula, line 509: Current implementation of this for-loop can be further optimized:
1. the adjustment logic for both schema and data region groups share lots of codes, can be extracted to some common functions.
2. the `calcMaxRegionGroupNum` can be invoked at most twice.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/persistence/schema/ClusterSchemaInfo.java
```diff
@@ -267,6 +267,20 @@ public TSStatus alterDatabase(final DatabaseSchemaPlan plan) {
             currentSchema.getName(),
             currentSchema.getMaxDataRegionGroupNum());
```
CRZbulabula: Similar as above, these are useless codes now. You can delete them.

CRZbulabula (changes requested):
PTAL.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/manager/schema/ClusterSchemaManager.java
```diff
@@ -908,14 +942,112 @@ public static TSStatus enrichDatabaseSchemaWithDefaultProperties(
...
+    }
+    if (maxRegionGroupNum < allocatedRegionGroupCount) {
+      return new TSStatus(TSStatusCode.DATABASE_CONFIG_ERROR.getStatusCode())
+          .setMessage(
+              String.format(
+                  "%s should be greater than or equal to allocated %sRegionGroupNum: %d.",
+                  fieldName, isSchemaRegion ? "Schema" : "Data", allocatedRegionGroupCount));
+    }
 
     return StatusUtils.OK;
```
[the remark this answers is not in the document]
zerolbsony: Resolved

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/statement/metadata/DatabaseSchemaStatement.java
```diff
@@ -152,10 +152,10 @@ public String toString() {
...
         + ", timePartitionInterval="
         + timePartitionInterval
-        + ", schemaRegionGroupNum="
-        + schemaRegionGroupNum
-        + ", dataRegionGroupNum="
-        + dataRegionGroupNum
+        + ", maxSchemaRegionGroupNum="
+        + maxSchemaRegionGroupNum
+        + ", maxDataRegionGroupNum="
+        + maxDataRegionGroupNum
```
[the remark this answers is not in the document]
zerolbsony: Changed
image

CRZbulabula (approved):
LGTM!

PR #18001, opened 2026-06-22
Reclaim the statement variable to avoid executing stale statements in the subsequent new test case

zerolbsony:


HTHou (commented):
One cleanup concern in `integration-test/src/test/java/org/apache/iotdb/db/it/utils/TestUtils.java` around line 1212:

`Statement.close()` can itself throw `SQLException`. Since it now runs before `localConnection.close()` in the same `try`, a close failure here would skip closing the local connection and can leak it on the retry path. Could you split the statement and connection closes into separate `try/catch` blocks, or otherwise ensure `localConnection.close()` is still attempted even when closing the statement fails?

HTHou (approved):

PR #18014, opened 2026-06-24
Limit concurrent load tsfile type conversions

Caideyipi:
### Problem

When multiple load-tsfile datatype conversion tasks run concurrently, each conversion can open TsFile tablet parsers and reserve Pipe tablet memory. In the investigated failure, the receiver failed in `LoadTreeStatementDataTypeConvertExecutionVisitor` with `forceAllocateForTablet: failed to allocate because there's too much memory for tablets`, so the load stayed temporarily unavailable and the sender retried.

### Fix

Limit concurrent load-tsfile datatype conversions with a shared semaphore. The permit count is the smaller of:

- `load_tsfile_tablet_conversion_thread_count`
- an estimate derived from the Pipe tablet memory reject threshold and the tablet parser memory reserved per conversion

If Pipe memory management is disabled, the existing configured conversion thread count is preserved.

### Validation

- `mvn spotless:check -pl iotdb-core/datanode`
- `mvn -Ddevelocity.off=true compile -pl iotdb-core/datanode` was attempted; it reached javac but failed on existing generated-source/ANTLR/dependency issues unrelated to this converter change, e.g. generated freemarker fill/aggregation classes and parser symbols.

<hr>

This PR has:
- [x] been self-reviewed.
    - [x] concurrent write

<hr>

- `org.apache.iotdb.db.storageengine.load.converter.LoadTsFileDataTypeConverter`

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/load/converter/LoadTsFileDataTypeConverter.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/load/converter/LoadTsFileDataTypeConverter.java
@@ -21,9 +21,12 @@ package org.apache.iotdb.db.storageengine.load.converter;
 
 import org.apache.iotdb.common.rpc.thrift.TSStatus;
 import org.apache.iotdb.commons.conf.IoTDBConstant;
+import org.apache.iotdb.commons.pipe.config.PipeConfig;
 import org.apache.iotdb.commons.queryengine.common.SqlDialect;
 import org.apache.iotdb.db.auth.AuthorityChecker;
 import org.apache.iotdb.db.conf.IoTDBDescriptor;
+import org.apache.iotdb.db.i18n.StorageEngineMessages;
+import org.apache.iotdb.db.pipe.resource.PipeDataNodeResourceManager;
 import org.apache.iotdb.db.protocol.session.IClientSession;
 import org.apache.iotdb.db.protocol.session.InternalClientSession;
 import org.apache.iotdb.db.protocol.session.SessionManager;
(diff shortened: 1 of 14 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/load/converter/LoadTsFileDataTypeConverter.java
```diff
@@ -46,13 +48,66 @@
...
+        Math.max(
+            1L, IoTDBDescriptor.getInstance().getConfig().getPipeDataStructureTabletSizeInBytes());
+    final long maxReaderChunkSize = Math.max(0L, pipeConfig.getPipeMaxReaderChunkSize());
+    final long tableSize =
+        Math.max(
+            1L,
+            Math.min(tabletSize, IoTDBDescriptor.getInstance().getConfig().getTargetChunkSize()));
+
+    final long treeParserMemorySize = 2L * tabletSize + maxReaderChunkSize;
+    final long tableParserMemorySize = 2L * tabletSize + 2L * tableSize + maxReaderChunkSize;
```
jt2594838, line 94: Add some comment

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/load/converter/LoadTsFileDataTypeConverter.java
```diff
@@ -46,13 +48,66 @@
...
+        ((pipeConfig.getPipeDataStructureTabletMemoryBlockAllocationRejectThreshold()
+                + pipeConfig.getPipeDataStructureTsFileMemoryBlockAllocationRejectThreshold() / 2)
+            * PipeDataNodeResourceManager.memory().getTotalNonFloatingMemorySizeInBytes());
+  }
+
+  private static Optional<TSStatus> getInterruptedConversionStatus(final InterruptedException e) {
+    Thread.currentThread().interrupt();
+    return Optional.of(
+        new TSStatus(TSStatusCode.LOAD_TEMPORARY_UNAVAILABLE_EXCEPTION.getStatusCode())
+            .setMessage("Interrupted while waiting for tablet conversion slot: " + e.getMessage()));
```
jt2594838: i18n

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/load/converter/LoadTsFileDataTypeConverter.java
Caideyipi: Added comments explaining the per-conversion memory estimate and why we use the larger tree/table parser working set.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/storageengine/load/converter/LoadTsFileDataTypeConverter.java
Caideyipi: Done. Moved the interrupted tablet conversion slot message to StorageEngineMessages and added both en/zh i18n entries.

jt2594838 (approved):

PR #18023, opened 2026-06-24
Fix schema path display quoting

Caideyipi:
## Description

This PR fixes schema/metadata display paths so paths containing reserved words, numeric-looking nodes, or otherwise invalid raw identifiers are returned with necessary back quotes.

The change centralizes schema display-path formatting in SchemaPathUtils and applies it to schema scan results, child path/node results, last-query outputs used by order-by-heat, template path display, active schema scan output, and related in-memory schema result objects.

## Tests

- git diff --cached --check

Attempted but blocked by local Windows page-file/native-memory exhaustion before Maven could start or complete:

- mvn -pl iotdb-core/datanode -Dtest=SchemaPathDisplayOperatorTest,SchemaQueryScanOperatorTest test

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/process/last/AlignedUpdateLastCacheOperator.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/execution/operator/process/last/AlignedUpdateLastCacheOperator.java
@@ -32,6 +32,8 @@ import org.apache.tsfile.read.common.block.TsBlock;
 import org.apache.tsfile.utils.RamUsageEstimator;
 import org.apache.tsfile.utils.TsPrimitiveType;
 
+import static org.apache.iotdb.db.utils.SchemaPathUtils.getFullPathWithNecessaryBackQuotes;
+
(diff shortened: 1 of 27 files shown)


Caideyipi:
Closing this PR after re-checking the intended behavior.

The current definition is: if the identifier was stored with backquotes and those backquotes are preserved in the stored path representation, SHOW should output them; otherwise SHOW should not add backquotes by itself.

Existing path handling already follows that rule: PartialPath/PathUtils removes backquotes for identifiers that can be represented directly, and preserves them for cases that require them (for example numeric or dotted identifiers). The current IT expectations, including root.sg1.d1.123d staying unquoted, match that behavior.

This PR mainly changes display code to add SQL-safe backquotes at output time (for example reserved words / parser-sensitive identifiers), which is a broader compatibility change and caused the observed IoTDBSyntaxConventionIdentifierIT failure. I also checked the active region scan path: the old IDeviceID.toString() joins the stored device-id segments, so it does not by itself justify keeping the new display-path map, since the map is also generated with display-time quoting.

So there is no narrow fix in this PR that should be kept under the current feature definition.

PR #18029, opened 2026-06-25
Fix tag index memory accounting

Caideyipi:
## Description
This PR makes TagManager update tag index memory accounting based on actual map/set mutations instead of non-atomic size snapshots.

Example scenario:
- Two overlapping tag-alter requests hit the same SchemaRegion at the same time, e.g. one thread runs `ALTER TIMESERIES root.sg.d1.s1 ADD TAGS(k=v)` while another runs `ALTER TIMESERIES root.sg.d2.s1 DROP TAGS(k)`.
- The old code compared `tagIndex.size()`, `tagValueMap.size()`, and `measurementsSet.size()` before and after separate mutation steps.
- With that interleaving, those snapshots can change in between, so the region memory counter can drift even though only one real insertion or removal happened.

It also makes index removal tolerant of missing entries and uses a concurrent measurement set for tag-index readers.

## Tests
- Added `TagManagerTest#concurrentAddAndRemoveIndexEventuallyReleasesAllMemory`, which starts 16 threads repeatedly adding and removing the same tag index entry, then verifies memory returns to zero and the index can still be accounted normally afterward.
- `mvn spotless:apply -pl iotdb-core/datanode` passed.
- `git diff --check` passed.
- `mvn -pl iotdb-core/datanode -Dtest=TagManagerTest#concurrentAddAndRemoveIndexEventuallyReleasesAllMemory test` could not reach the test phase locally because DataNode compilation failed on unrelated generated/query/pipe classes, e.g. missing `AGGREGATION_OPERATOR_FROM_RAW_DATA`, `Accumulator`/`IFill`, and pipe inclusion constants from dependent modules/classes.


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/schemaengine/schemaregion/tag/TagManager.java
```diff
@@ -166,34 +165,33 @@ public void addIndex(String tagKey, String tagValue, IMeasurementMNode<?> measur
...
-        tagValueMap.computeIfAbsent(tagValue, v -> Collections.synchronizedSet(new HashSet<>()));
-    int tagValueMapNewSize = tagValueMap.size();
+    tagIndex.compute(
+        tagKey,
+        (key, tagValueMap) -> {
+          long memorySize = 0;
+          if (tagValueMap == null) {
+            tagValueMap = new ConcurrentHashMap<>();
+            // the last 4 is the memory occupied by the size of tagvaluemap
+            memorySize += RamUsageEstimator.sizeOf(tagKey) + 4;
```
jt2594838: Why is it only 4 bytes? The same below

Caideyipi: Good point. I replaced the hard-coded 4 with COLLECTION_SIZE_FIELD_BYTES = Integer.BYTES for the estimated collection size field and removed the misleading comments in ee56b901e8c.

Caideyipi: Followed up in 28a97160245: renamed the constant to INDEX_ENTRY_OVERHEAD_ESTIMATE_BYTES and added a comment clarifying that it is an int-sized accounting estimate for each indexed key/value/measurement reference, not a specific ConcurrentHashMap or Set field.

PR #18033, opened 2026-06-25
Use procedures for region create/delete to fix the delete-region ghost task

CRZbulabula:
## Problem

The ConfigNode-leader periodically drained a persistent **RegionMaintainer queue** (`PartitionManager.maintainRegionReplicas`, every 10s). For region **deletion** it sent **synchronous** `deleteRegion` RPCs to DataNodes.

For a data region this deletion can be slow (removing TsFiles + the consensus peer) and may **outlive the RPC timeout**. When it does:

1. The ConfigNode times out and retries the same task 10s later.
2. On the DataNode the **second** call re-runs the deletion, which now returns `SUCCESS` because the region is already gone (`ConsensusGroupNotExistException` is swallowed and deleting an already-removed region is a no-op).
3. The ConfigNode treats that `SUCCESS` as completion, polls the task off the queue, and forgets it — **while the first call is still running** on the DataNode.

The result is a *ghost task*: the cluster keeps executing a deletion the coordinator has already discarded.

## Approach

Region **deletion** moves off the fire-and-forget queue and onto a **Procedure**, while region **creation** (replica re-creation) stays on the RegionMaintainer queue, which never had the ghost-task race (a create returns a final, unambiguous status from a single RPC).

### Deletion → `RemoveRegionGroupProcedure`
A new `RemoveRegionGroupProcedure` deletes a whole region group, replica by replica, reusing the existing migration primitives (`submitDeleteOldRegionPeerTask` → `deleteOldRegionPeer`/`deleteLocalPeer`, then `waitTaskFinish` polling for the real terminal state). It needs no quorum, is idempotent, tolerates an already-absent peer, and **fails loudly rather than skipping**, so the parent can never drop a partition table while region data is still on disk. It gains persistence and retry across leader changes like the other region procedures.

Callers now spawn it as a child instead of offering delete tasks to the queue:
- `DeleteDatabaseProcedure` — one `RemoveRegionGroupProcedure` per region group, completed before the `DatabasePartitionTable` is removed.
- `CreateRegionGroupsProcedure` (no-quorum branch) — one `RemoveRegionGroupProcedure` to clean up the replicas that did get created.

To make this reuse safe, a latent bug in `RegionMigrateService.DeleteOldRegionPeerTask` was fixed: on a deletePeer/deleteRegion failure it called `taskFail` but **fell through to `taskSucceed`**, reporting failure as success — it now returns, and `deletePeer` treats `ConsensusGroupNotExistException` as success. This also hardens the RemoveRegionPeer / region-migration path.

### Creation stays on the RegionMaintainer queue
`PartitionManager.maintainRegionReplicas` and the `RegionCleaner` scheduler are kept, but now **only recreate failed replicas**. The dead delete branch (the `DELETE_REGION` async RPC) is removed, and legacy `RegionDeleteTask`s that an upgraded ConfigNode might replay from a consensus log or snapshot are filtered out at the two `PartitionInfo` ingestion points (`offerRegionMaintainTasks`, `processLoadSnapshot`) so they can never get stuck at a region's queue head and block re-creation of that region's other replicas.

### Cleanup / readability
- Removed the now-dead async `CnToDnAsyncRequestType.DELETE_REGION` enum + dispatch and the two orphaned `ManagerMessages` log constants (en + zh).
- **Refactored `maintainRegionReplicas` for readability now that the queue is create-only:** dropped the `currentType`/`switch` machinery and the duplicated SchemaRegion/DataRegion blocks, replaced the nested `Optional.ifPresent` + leader check with an early return, and extracted `submitRegionCreateTasks(type, tasks)` / `collectSuccessfulRegions(...)`. Per-region FIFO ordering, the failed-head-blocks-region semantics, and the in-memory/persisted poll lockstep are preserved; the loop now batches both region types per round (throughput-only change). Net **−50 lines** in the method.

### Upgrade safety (keep deserialization, don't delete)
`RegionDeleteTask`, the `RegionMaintainType.DELETE` enum value, and the `RegionMaintainTask.Factory` deserialization arms are kept **byte-identical to master** (master persisted these to the Ratis log and snapshots), so a rolling upgrade can still replay an old log / load an old snapshot — the delete tasks are deserialized and then dropped at ingestion. The retained `OfferRegionMaintainTasks` / `PollRegionMaintainTask` / `PollSpecificRegionMaintainTask` plan types likewise keep consensus-log replay compatible.

## Tests
- `RemoveRegionGroupProcedureTest` — serialization round-trip through `ProcedureFactory` (exercises the new `ProcedureType` code + factory registration) and the replica-deletion state machine.
- `PartitionInfoTest.testLegacyRegionDeleteTasksAreFiltered` — a legacy delete task offered to / loaded from a snapshot is dropped, leaving only create tasks queued, and survives a snapshot round-trip.
- Existing `ConfigPhysicalPlanSerDeTest` for the retained plan types stays green (proves consensus-log replay compatibility).

All confignode + datanode unit tests above pass; `spotless:check` and `checkstyle` are clean.


CRZbulabula:
Follow-up (6306e96): simplified the DataNode side. Instead of adding a new `deleteRegionAsync` RPC + `DeleteRegionTask`, `DeleteRegionProcedure` now **reuses the existing `deleteOldRegionPeer` plumbing** (`submitDeleteOldRegionPeerTask` → `DeleteOldRegionPeerTask`), which already does `deleteLocalPeer` + delete-region-data on a background thread keyed by `taskId` and is polled via `getRegionMaintainResult` — exactly what region-replica deletion needs.

This removes the duplicated `deleteRegionAsync` RPC, `DeleteRegionTask`, `submitDeleteRegionTask` (service + handler), the `DELETE_REGION_ASYNC` request type and its i18n constant (net −254 lines on the DataNode side).

It also fixes a latent bug in `DeleteOldRegionPeerTask.run()` that the reuse exposes: on a `deletePeer`/`deleteRegion` failure it called `taskFail` but did **not** `return`, then fell through to `taskSucceed`, so a failed deletion was reported as success. It now returns after `taskFail`, and `deletePeer` treats `ConsensusGroupNotExistException` as success so a retry converges cleanly. This also hardens the existing RemoveRegionPeer/migration path that already uses this task.

The ConfigNode-side `CreateRegionProcedure`/`DeleteRegionProcedure` remain separate procedures: `Add`/`RemoveRegionPeerProcedure` model peer-membership changes (coordinator, consensus pipes, transfer-leader, resetPeerList rollback, partition-table edits), which is a different concern from region-group create/delete, so reusing them there would drag in irrelevant/harmful steps.

PR #18035, opened 2026-06-26
Support IoTDBLocal in UDF for table model

Wei-hao-Li:


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/session/SessionManager.java
```diff
@@ -369,6 +369,23 @@ public IClientSession getCurrSession() {
     return currSession.get();
   }
 
+  /**
+   * Swap the ThreadLocal session and return the previous one. Used by UDF internal queries to
+   * temporarily install an internal session without removing the previous session from the session
+   * map.
+   */
+  public IClientSession exchangeCurrSession(IClientSession session) {
```
JackieTien97: what's the return value used for? if not used, don't call it exchangeXXXX

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/protocol/session/SessionManager.java
```diff
@@ -369,6 +369,23 @@ public IClientSession getCurrSession() {
...
+    IClientSession previous = currSession.get();
+    if (session != null) {
+      currSession.set(session);
+      sessions.putIfAbsent(session, placeHolder);
```
JackieTien97: where did you remove it from `sessions`?

JackieTien97 (changes requested):
Extend the IT coverage for the missing IoTDBLocal scenarios from the design doc. In particular, we should have an IT proving that IoTDBLocal.query() rejects non-query statements such as INSERT, CREATE TABLE, or DROP TABLE before any side effect happens; an IT where UDAF and table-function lifecycle methods actually call local.query() rather than only logging through IoTDBLocal; and, if possible, a cluster/partial-aggregation case that exercises combineState(state, rhs, local). These would close the current gap between the documented self-test scenarios and the PR coverage.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/udf/UDFResultSetImpl.java
```diff
@@ -0,0 +1,121 @@
...
+      } catch (IoTDBException e) {
+        throw new UDFException(e.getMessage(), e);
+      }
+      if (!batch.isPresent()) {
+        return false;
+      }
+      TsBlock currentBlock = batch.get();
+      rowIterator =
+          new RecordIterator(
+              Arrays.asList(currentBlock.getValueColumns()),
```
JackieTien97: should rearrange the column sequence according to DatasetHeader.columnIndex2TsBlockColumnIndexList, add an IT about this.

iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/execution/operator/source/relational/aggregation/UserDefinedAggregateFunctionAccumulator.java
```diff
@@ -46,17 +49,51 @@ public class UserDefinedAggregateFunctionAccumulator implements TableAccumulator
...
+
+  private void initIfNeeded() {
+    if (init) {
+      return;
+    }
+    init = true;
+    try {
+      aggregateFunction.beforeStart(functionArguments, ioTDBLocal);
+    } catch (UDFException e) {
+      throw new RuntimeException(
```
JackieTien97: never throw RuntimeException, use IoTDBRunTimeExeception instead, and use `EXECUTE_UDF_ERROR(1206)` status code

iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/execution/operator/source/relational/aggregation/UserDefinedAggregateFunctionAccumulator.java
```diff
@@ -66,23 +103,26 @@ public long getEstimatedSize() {
 
   @Override
   public TableAccumulator copy() {
-    return new UserDefinedAggregateFunctionAccumulator(analysis, aggregateFunction, inputDataTypes);
+    return new UserDefinedAggregateFunctionAccumulator(
+        analysis, aggregateFunction, functionArguments, inputDataTypes, ioTDBLocal, true);
```
JackieTien97:
```suggestion
        analysis, aggregateFunction, functionArguments, inputDataTypes, ioTDBLocal, init);
```

iotdb-core/calc-commons/src/main/java/org/apache/iotdb/calc/execution/operator/source/relational/aggregation/UserDefinedAggregateFunctionAccumulator.java
```diff
@@ -150,7 +191,8 @@ public boolean removable() {
 
   @Override
   public void close() {
-    aggregateFunction.beforeDestroy();
+    aggregateFunction.beforeDestroy(ioTDBLocal);
```
JackieTien97, line 201:
```suggestion
    if (init) {
      aggregateFunction.beforeDestroy(ioTDBLocal);
    }
```

only call `beforeDestroy` if we call beforeStart

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/planner/plan/FragmentInstance.java
```diff
@@ -83,6 +83,9 @@ public class FragmentInstance implements IConsensusRequest {
   private final boolean debug;
   private final boolean verbose;
 
+  // Outer query deadline (startTime + timeout) for IoTDBLocal on remote DataNodes.
+  private long outerQueryDeadlineMs = -1L;
```
JackieTien97: no need to add this field? we can infer this just using timeOut + System.currentTime

JackieTien97: avoid changing FI serde method, we would try our best to promise rolling upgrading.

JackieTien97 (changes requested):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/Coordinator.java
```diff
@@ -308,6 +308,17 @@ private ExecutionResult execution(
...
+    return execution(queryId, session, sql, userQuery, debug, false, iQueryExecutionFactory);
+  }
+
+  private ExecutionResult execution(
+      long queryId,
+      SessionInfo session,
+      String sql,
+      boolean userQuery,
+      boolean debug,
+      boolean readOnlyInternalQuery,
```
JackieTien97, line 320: It will always be false

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/udf/InternalQueryExecutor.java
```diff
@@ -0,0 +1,125 @@
...
+
+      IQueryExecution queryExecution = COORDINATOR.getQueryExecution(queryId);
+      if (queryExecution == null) {
+        throw new IoTDBException(
+            "Internal query execution not found",
+            TSStatusCode.INTERNAL_SERVER_ERROR.getStatusCode());
+      }
+
+      return new InternalQueryResult(queryExecution, internalSession, statementId, queryId, sql);
+    } catch (Exception e) {
```
JackieTien97, line 116: catch IoTDBException and IoTDBRuntimeException seprately, and rethrow them instead of using another  TSStatusCode.INTERNAL_SERVER_ERROR.getStatusCode() IoTDBException

JackieTien97 (changes requested):

JackieTien97 (approved):

PR #18048, opened 2026-06-29
Avoid login lock for pipe auth retry

Caideyipi:
## Summary
- slow down pipe retry intervals for authentication failures to avoid triggering the default login lock window
- preserve receiver auth status codes in sync and legacy handshake error messages
- add unit coverage for auth failure detection and retry interval behavior

## Tests
- `mvn spotless:apply -pl iotdb-core/node-commons`
- `mvn test -pl iotdb-core/node-commons -Dtest=PipeSleepIntervalTest`
- `git diff --check`

## Notes
- `mvn -pl iotdb-core/datanode -DskipTests compile` reached javac but failed in this environment with `Java heap space` / pagefile memory errors.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/sink/protocol/legacy/IoTDBLegacyPipeSink.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/sink/protocol/legacy/IoTDBLegacyPipeSink.java
@@ -278,8 +278,11 @@ public class IoTDBLegacyPipeSink implements PipeConnector {
     if (openSessionResp.getStatus().getCode() != TSStatusCode.SUCCESS_STATUS.getStatusCode()) {
       final String errorMsg =
           String.format(
-              "Failed to login to receiver %s:%s for legacy pipe transfer because %s",
-              ipAddress, port, openSessionResp.getStatus().getMessage());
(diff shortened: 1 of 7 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/sink/protocol/legacy/IoTDBLegacyPipeSink.java
```diff
@@ -278,8 +278,11 @@ private void openClientSession() throws TException {
     if (openSessionResp.getStatus().getCode() != TSStatusCode.SUCCESS_STATUS.getStatusCode()) {
       final String errorMsg =
           String.format(
-              "Failed to login to receiver %s:%s for legacy pipe transfer because %s",
-              ipAddress, port, openSessionResp.getStatus().getMessage());
+              "Failed to login to receiver %s:%s for legacy pipe transfer because code: %d, message: %s",
+              ipAddress,
+              port,
+              openSessionResp.getStatus().getCode(),
+              openSessionResp.getStatus().getMessage());
```
jt2594838, line 285: i18n

Caideyipi: Fixed. The legacy pipe login failure message has been moved into DataNodePipeMessages with both en and zh i18n entries.

PR #18049, opened 2026-06-29
Pipe: improve incremental memory estimation

Caideyipi:
﻿## Description
This PR improves the pipe creation memory admission check from a coarse per-pipe estimate to an incremental estimate based on the tasks and sink subtasks that this DataNode will actually create.

Changes include:
- Estimate memory by local region tasks, and also check newly added single-region tasks.
- Account for sink subtask reuse so shared sink pending queues/subtasks are not charged repeatedly.
- Reuse the actual sink subtask key and parallelism logic from PipeSinkSubtaskManager.
- Match task builder parameter blending/preprocessing during admission estimation.
- Estimate batch memory from connector.batch.size-bytes/sink.batch.size-bytes, including leader-cache endpoint sharding.
- Estimate TsFile send read buffer with pipeSinkReadFileBufferSize.
- Skip admission checks for idempotent create no-ops, while checking before dropping old tasks during recreate.

## Verification
- git -c core.preloadIndex=false diff --check -- iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/pipe/agent/task/PipeTaskAgent.java iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/PipeDataNodeTaskAgent.java iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/builder/PipeDataNodeTaskBuilder.java iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/subtask/sink/PipeSinkSubtaskManager.java
- mvn "-Ddevelocity.off=true" "-Dcheckstyle.skip=true" "-Dspotless.check.skip=true" "-DskipTests" compile -pl iotdb-core/node-commons

Also ran datanode compile:
- mvn "-Ddevelocity.off=true" "-Dcheckstyle.skip=true" "-Dspotless.check.skip=true" "-DskipTests" compile -pl iotdb-core/datanode
- It reached javac but failed on existing generated-source/dependency issues such as missing IFill, Accumulator, and org.apache.commons.collections4.comparators, not in the modified pipe files.


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/PipeDataNodeTaskAgent.java
```diff
@@ -850,17 +990,17 @@ private void calculateInsertNodeQueueMemory(final PipeParameters sourceParameter
...
             - allocatedMemorySizeInBytes;
-    if (remainingMemory < PipeConfig.getInstance().getPipeInsertNodeQueueMemory()) {
+    if (remainingMemory < needFloatingMemory) {
       final String message =
           String.format(
               "%s Need Floating memory: %d  bytes, free Floating memory: %d bytes",
-              PipeMessages.NOT_ENOUGH_MEMORY_FOR_PIPE,
-              PipeConfig.getInstance().getPipeInsertNodeQueueMemory(),
-              remainingMemory);
+              PipeMessages.NOT_ENOUGH_MEMORY_FOR_PIPE, needFloatingMemory, remainingMemory);
```
jt2594838: i18n

Caideyipi: Done. I moved the memory check details into `PipeMessages` with both English and Chinese i18n entries, and updated the code to use those message constants.

PR #18051, opened 2026-06-29
Improve pipe runtime memory allocation

Caideyipi:
## Summary
- Keep PR 18051's dynamic pipe memory allocation changes for TsFile parser, sink batch, and TsFile read buffers.
- Fix the TsFile -> Tablet parser memory self-lock seen in multi-pipe large TsFile sync: when tablet memory allocation fails, the same parser no longer keeps retrying while holding parser chunk/page/tablet memory.
- Bound the OOM recovery path so the TsFile event is retried by the outer pipe subtask after parser memory is released, instead of spinning forever inside one parsed tablet.

## Root cause fixed
In the failing path, `PipeTsFileInsertionEvent.consumeTabletInsertionEventsWithRetry` caught `PipeRuntimeOutOfMemoryCriticalException` and retried the same parsed tablet indefinitely. The parser still held temporary tablet/chunk/page memory while the consumer was trying to allocate memory for the downstream `PipeRawTabletInsertionEvent`. With multiple pipes parsing large TsFiles concurrently, this could saturate the tablet memory pool and form a live-lock: pipe stays RUNNING, but `sub_remain_event` does not decrease and no tablet is transferred.

## Implementation details
- Release parser temporary tablet memory after scan/query/table parsers hand a `Tablet` to a raw tablet event.
- On pipe OOM during TsFile tablet consumption:
  - release the parsed raw tablet event if it has not been queued/referenced;
  - close the current TsFile parser to release parser-side tablet/chunk/page/modification memory;
  - rethrow OOM so the outer pipe subtask keeps the original TsFile event as `lastEvent` and retries later.
- Ensure OOM is not hidden by processor lambdas that temporarily store exceptions in `AtomicReference` (`PipeProcessorSubtask`, `DownSamplingProcessor`, `AggregateProcessor`).
- Wake memory waiters when `PipeMemoryManager.resize(..., 0)` releases memory.
- Fix zero-size memory block registration so a failed expansion does not leave an unallocated block in `allocatedBlocks`; also register zero-size blocks expanded through `tryAllocate(block, ...)`.
- Update the OOM log message to describe the new release-and-retry behavior instead of infinite in-place retry.

## Tests
- Added UT coverage in `TsFileInsertionEventParserTest`:
  - `testScanParserReleasesTabletMemoryAfterRawTabletGenerated`
  - `testConsumeTabletInsertionEventsWithRetryReleasesParserOnOutOfMemory`
- `git diff --check`
- Attempted targeted UT:
  - `mvn -pl iotdb-core/datanode '-Dtest=TsFileInsertionEventParserTest#testScanParserReleasesTabletMemoryAfterRawTabletGenerated+testConsumeTabletInsertionEventsWithRetryReleasesParserOnOutOfMemory' '-Dspotless.check.skip=true' '-Dcheckstyle.skip=true' test`
  - blocked before test execution by existing unrelated main compile errors in `ConsensusSubscriptionSetupHandler`: missing `TopicConstant.COLUMN_KEY` and `TopicConstant.COLUMN_DEFAULT_VALUE`.
- Attempted module compile with reduced JVM memory pressure:
  - `mvn -pl iotdb-core/datanode -DskipTests '-Dspotless.check.skip=true' '-Dcheckstyle.skip=true' compile`
  - blocked by the same unrelated `TopicConstant.COLUMN_*` compile errors.

No new IT is added because this is a low-level parser/memory-manager recovery path and the deterministic UTs cover the parser-memory release and OOM cleanup behavior directly. The original 10-pipe/2kw scenario should be used as the regression workload for end-to-end validation.


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/event/common/tsfile/parser/scan/TsFileInsertionEventScanParser.java
```diff
@@ -235,6 +231,7 @@ public TabletInsertionEvent next() {
                   final Tablet tablet = getNextTablet();
                   // Record tablet metrics
                   recordTabletMetrics(tablet);
+                  releaseTabletMemoryBlock();
```
jt2594838: Where is the memory allocated?
May move it to the first line; otherwise, it looks like you are freeing the newly retrieved Tablet.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/sink/payload/evolvable/batch/PipeTabletEventBatch.java
```diff
@@ -126,7 +129,28 @@ public synchronized boolean onEvent(final TabletInsertionEvent event)
...
 
+  protected void increaseTotalBufferSizeAndUpdateMemoryBlock(final long bufferSize) {
+    if (bufferSize <= 0) {
+      return;
+    }
+
+    final long newTotalBufferSize = totalBufferSize + bufferSize;
+    PipeDataNodeResourceManager.memory()
+        .forceResize(allocatedMemoryBlock, Math.min(newTotalBufferSize, maxBatchSizeInBytes));
+    totalBufferSize = newTotalBufferSize;
```
jt2594838: newTotalBufferSize =  Math.min(totalBufferSize + bufferSize, maxBatchSizeInBytes)?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/sink/protocol/iotconsensusv2/payload/builder/IoTConsensusV2TransferBatchReqBuilder.java
```diff
@@ -137,27 +113,80 @@ public synchronized boolean onEvent(TabletInsertionEvent event)
...
+    while (events.size() > previousEventsSize) {
+      events.remove(events.size() - 1);
+    }
+    while (requestCommitIds.size() > previousRequestCommitIdsSize) {
+      requestCommitIds.remove(requestCommitIds.size() - 1);
+    }
+    while (batchReqs.size() > previousBatchReqsSize) {
+      batchReqs.remove(batchReqs.size() - 1);
+    }
+  }
```
jt2594838, line 178: subList.clear()?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/event/common/tsfile/parser/scan/TsFileInsertionEventScanParser.java
Caideyipi: Applied in c940475af1d: the previous parser-owned tablet memory is released before fetching the next tablet, so the ownership/order is clearer.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/sink/payload/evolvable/batch/PipeTabletEventBatch.java
Caideyipi: Applied in c940475af1d: totalBufferSize is clamped with Math.min(totalBufferSize + bufferSize, maxBatchSizeInBytes) before resizing.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/sink/protocol/iotconsensusv2/payload/builder/IoTConsensusV2TransferBatchReqBuilder.java
Caideyipi: Applied in c940475af1d: rollback now uses subList(...).clear() for events, commit ids, and batch requests.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/PipeDataNodeTaskAgent.java
jt2594838: i18n

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/sink/protocol/thrift/sync/IoTDBDataRegionSyncSink.java
```diff
@@ -612,73 +612,114 @@ protected void transferFilePieces(
...
+                  clientAndStatus.getLeft().getEndPoint(),
+                  (long) (req.getBody().length * weight)));
+      return PipeTransferFilePieceResp.fromTPipeTransferResp(
+          clientAndStatus.getLeft().pipeTransfer(req));
+    } catch (final Exception e) {
+      clientAndStatus.setRight(false);
+      throw new PipeConnectionException(
+          String.format("Network error when transfer file %s, because %s.", file, e.getMessage()),
+          e);
+    }
```
jt2594838, line 709: i18n

Caideyipi: Applied in 1370353caf5: reused the existing transfer-file network error i18n constant and added en/zh i18n for the transfer file status message.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/agent/task/PipeDataNodeTaskAgent.java
Caideyipi: Applied in 1370353caf5: moved the pipe meta report logs to DataNodePipeMessages with en/zh i18n constants.

PR #18052, opened 2026-06-29
Enable show queries can be executed when the available memory for Operators is insufficient

Wei-hao-Li:


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/queryengine/utils/cte/CteDataStore.java
```diff
@@ -48,6 +48,8 @@ public class CteDataStore implements Accountable {
   // reference count by CteScanReader
   private final AtomicInteger count;
 
+  private long operatorsMemoryReservedBytes;
```
JackieTien97: add some comments about this field

JackieTien97 (approved):

JackieTien97 (approved):

PR #18058, opened 2026-06-29
Support target database in write-back sink

Caideyipi:
### Changes
- Add `sink.database` / `connector.database` support for write-back sink data writes.
- Validate write-back sink database names on the sink side instead of relying on generic path validation.
- Use table-model create-database validation for table-model database names.
- Use tree-model create-database validation semantics for tree-model database names, including `root.` requirement, legal tree path parsing, database-name pattern check, and length limit.
- Support multi-level tree-model target databases such as `root.target.db`.
- Reject table-model writes when the event or target mapping produces an invalid table-model database, for example tree database `root.a.b` mapping to table database `a.b`.
- Rewrite target database for both table-model and tree-model insertion statements.
- Add UT coverage for target database validation boundaries, target database customization/mapping, raw tablet event rejection, and statement event rejection.
- Update tree-model write-back sink IT coverage to use a multi-level target database.

### Semantics
- `sink.database=target` is treated as a table-model database and mapped to tree database `root.target` for tree writes.
- `sink.database=root.target` and `sink.database=root.target.db` are treated as tree-model databases.
- A multi-level tree-model database such as `root.target.db` is valid for tree-model writes.
- The corresponding table-model database name for `root.target.db` is `target.db`, which is invalid in table model and must reject table-model events.
- Table-model events always validate their event database name with table-model create-database rules before applying a configured target database, so source values like `root.a.b` / `a.b` are rejected.
- Tree-model events always validate their event database name with tree-model create-database rules before applying a configured target database.

### Validation
- `git diff --check`
- `mvn -pl iotdb-core/datanode -Dtest=PipeSinkTest test` was attempted, but datanode compilation fails before test execution in existing generated Freemarker sources unrelated to this change, including missing `IFill`, `IFillFilter`, `LinearFill`, `Accumulator`, `org.apache.commons.collections4.comparators`, and `IoTDBConfig#getModeMapSizeThreshold()`.


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/sink/protocol/writeback/WriteBackSink.java
```diff
@@ -103,35 +115,126 @@
...
+
+  private static void validateTableModelDatabaseName(final String databaseName) {
+    try {
+      TableConfigTaskVisitor.validateDatabaseName(databaseName);
+    } catch (final Exception e) {
+      throw new PipeException(
+          String.format(
+              "The table-model database %s is invalid. It should not contain '%s', should match "
+                  + "the pattern %s, and the length should not exceed %d",
+              databaseName, PATH_SEPARATOR, IoTDBConfig.DATABASE_PATTERN, MAX_DATABASE_NAME_LENGTH),
```
jt2594838: i18n

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/sink/protocol/writeback/WriteBackSink.java
```diff
@@ -215,6 +325,30 @@ public void customize(
...
+    targetTreeModelDatabaseName = validateAndNormalizeTreeModelDatabaseName(targetDatabase);
+    final String tableModelDatabaseName =
+        PathUtils.unQualifyDatabaseName(targetTreeModelDatabaseName).toLowerCase(Locale.ENGLISH);
+    try {
+      TableConfigTaskVisitor.validateDatabaseName(tableModelDatabaseName);
+      targetTableModelDatabaseName = tableModelDatabaseName;
+    } catch (final Exception e) {
+      invalidTargetTableModelDatabaseName = tableModelDatabaseName;
+    }
+  }
```
jt2594838, line 358: Why are tree db and table db set in both branches?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/sink/protocol/writeback/WriteBackSink.java
```diff
@@ -384,6 +539,161 @@ private void doTransfer(final PipeStatementInsertionEvent pipeStatementInsertion
...
+      return devicePath;
+    }
+
+    try {
+      final String[] sourceDatabaseNodes = new PartialPath(sourceTreeModelDatabaseName).getNodes();
+      final String[] targetDatabaseNodes = new PartialPath(targetTreeModelDatabaseName).getNodes();
+      final String[] deviceNodes = devicePath.getNodes();
+      if (!startsWith(deviceNodes, sourceDatabaseNodes)) {
+        return devicePath;
+      }
```
jt2594838, line 678: When will this happen?

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/sink/protocol/writeback/WriteBackSink.java
Caideyipi: Addressed. I moved the newly added write-back sink database validation/rewrite exception messages into DataNodePipeMessages for both en and zh locales.

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/sink/protocol/writeback/WriteBackSink.java
Caideyipi: Good point. The previous wording was inaccurate: with tree/table isolation, a write-back sink task should only see one event model. I kept both normalized target names because WriteBackSink only receives sink parameters during validate/customize and does not know the isolated runtime model at that point; the incoming event model still selects the target name later. This also preserves support for table-style targets like target being rewritten to root.target for tree events, while keeping the invalid table-model conversion marker for tree targets such as root.target.db. I updated the code comment accordingly.

jt2594838 (approved):

PR #18069, opened 2026-06-30
Pipe: add ConfigNode memory control for log reducer

Caideyipi:
﻿## Summary
- Add ConfigNodeMemoryConfig and confignode_memory_proportion for ConfigNode Pipe/Free memory split.
- Wire PipePeriodicalLogReducer to DataNode and ConfigNode memory resize hooks.
- Keep ConfigNode Pipe memory exposed only as logReducerMemory, so no other CN module is attached to the new memory framework.

## Tests
- PASS: mvn spotless:apply -pl iotdb-core/node-commons,iotdb-core/confignode,iotdb-core/datanode
- PASS: mvn test -pl iotdb-core/node-commons,iotdb-core/confignode "-Dtest=PipePeriodicalLogReducerTest,ConfigNodeMemoryConfigTest" -DfailIfNoTests=false
- PASS: mvn compile -pl iotdb-core/node-commons,iotdb-core/confignode
- NOTE: mvn compile -pl iotdb-core/node-commons,iotdb-core/datanode reaches javac but fails on existing calc MemoryReservationManager#setHighestPriority mismatch unrelated to this Pipe change.

### DIFF
--- a/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeMemoryConfig.java
+++ b/iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeMemoryConfig.java
@@ -22,6 +22,7 @@ package org.apache.iotdb.confignode.conf;
 import org.apache.iotdb.commons.conf.TrimProperties;
 import org.apache.iotdb.commons.memory.MemoryConfig;
 import org.apache.iotdb.commons.memory.MemoryManager;
+import org.apache.iotdb.confignode.i18n.ConfigNodeMessages;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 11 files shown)


iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeMemoryConfig.java
```diff
@@ -0,0 +1,96 @@
...
+  /** The memory manager of on heap. */
+  private MemoryManager onHeapMemoryManager;
+
+  /** Memory manager for the pipe. */
+  private MemoryManager pipeMemoryManager;
+
+  public void init(final TrimProperties properties) {
+    String memoryAllocateProportion = properties.getProperty("confignode_memory_proportion", null);
+    if (memoryAllocateProportion == null) {
+      memoryAllocateProportion = properties.getProperty("config_node_memory_proportion", null);
```
jt2594838: just keep one

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeMemoryConfig.java
```diff
@@ -0,0 +1,96 @@
...
+      if (memoryAllocateProportion != null) {
+        LOGGER.warn(
+            "The parameter config_node_memory_proportion is deprecated, "
+                + "please use confignode_memory_proportion instead.");
+      }
```
jt2594838: i18n

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeMemoryConfig.java
Caideyipi: Done. Removed the compatibility fallback for `config_node_memory_proportion` and now only reads `confignode_memory_proportion`.

iotdb-core/confignode/src/main/java/org/apache/iotdb/confignode/conf/ConfigNodeMemoryConfig.java
Caideyipi: Done. The deprecated-parameter warning was removed together with the fallback, and the remaining ConfigNode memory log messages now use `ConfigNodeMessages` i18n constants.

PR #18079, opened 2026-07-01
Support attribute filters in table deletion

jt2594838:
## Summary
- Support table deletion with attribute filters by converting matched devices into deletion predicates
- Support IS NOT NULL for attribute filters and tag filters
- Add a SegmentNotNull TagPredicate for tag IS NOT NULL without device schema fetch
- Add integration tests for attribute/tag/time condition combinations and operator coverage

## Tests
- mvn spotless:apply -pl iotdb-core/datanode,integration-test -P with-integration-tests
- mvn verify -DskipUTs "-Dit.test=IoTDBDeletionTableIT#testDeleteDataByTagIsNotNullFilter" "-DfailIfNoTests=false" "-Dfailsafe.failIfNoSpecifiedTests=false" "-Drat.skip=true" -pl integration-test -am -PTableSimpleIT -P with-integration-tests
- mvn verify -DskipUTs "-Dit.test=IoTDBDeletionTableIT#testDeleteDataByAttributeFilterWithOtherOperators" "-DfailIfNoTests=false" "-Dfailsafe.failIfNoSpecifiedTests=false" "-Drat.skip=true" -pl integration-test -am -PTableSimpleIT -P with-integration-tests


Caideyipi:
Thanks for the work on supporting attribute filters in table deletion. I found a few issues that should be addressed before merge:

1. `IS NOT NULL` tag predicates are dropped when they are combined with attribute filters.

   In `AnalyzeUtils.parsePredicate`, the `IsNotNullPredicate` branch only handles `shouldQueryDevice()`. Unlike the `ComparisonExpression` and `IsNullPredicate` branches, it does not add tag-only predicates to `deviceFilterExpressions` when `shouldFilterDevice()` is true. Later, when any attribute predicate exists, `tagPredicate` is replaced by `new DeviceIn(deviceIDs)`, so the original `SegmentNotNull` predicate is lost.

   For example:
   ```sql
   DELETE FROM t WHERE site IS NOT NULL AND attr1 = 'red'
   ```
   may delete data from devices whose `attr1 = 'red'` but whose `site` tag is `NULL`. Please either include tag `IS NOT NULL` in the schema fetch predicate list, or preserve/intersect the original tag predicate when converting matched devices to `DeviceIn`.

2. The existing integration test expectation for `deviceId IS NOT NULL` looks stale.

   `testDeleteData` still expects:
   ```sql
   DELETE FROM vehicle1 WHERE time < 10 AND deviceId IS NOT NULL
   ```
   to fail with "Unsupported expression", but this PR adds tag `IS NOT NULL` support via `parseIsNotNull`. This test should now fail unless the expectation and subsequent data validation are updated.

3. Attribute comparison predicates need explicit validation before being sent to the schema fetcher.

   `parseComparison` returns `PredicateParseResult.attribute(...)` without checking the operator or right-hand value type. Inputs such as `attr1 = 1`, `attr1 = null`, or unsupported comparison operators can flow into later schema-filter conversion code and trigger casts/internal exceptions instead of a clear semantic error. Please validate attribute predicates here, or make the schema predicate conversion layer return a proper `SemanticException` for invalid attribute predicates.

PR #18081, opened 2026-07-01
Move IoTConsensus multi-data-dir region migration ITs to commit suite

hongzhi-gao:
### Content1
Move IoTConsensus multi-data-dir region migration ITs from `pass/daily/iotv1` to `pass/commit`, updating package names and static imports so they run in the commit CI suite instead of daily.

### Content2
Fix `IoTDBMigrateMultiRegionForIoTV1IT.multiRegionMigrateTest`, which failed in ClusterIT (1C3D) because `INSERTION1` only creates one schema region and one data region; with replication factor 1 they are often placed on different DataNodes, so no DataNode hosts two regions. Before multi-region migrate, colocate regions via `extend region` when needed.

<hr>

This PR has:
- [x] been self-reviewed.
- [ ] added comments explaining the "why" and the intent of the code wherever would not be obvious for an unfamiliar reader.
- [ ] added unit tests or modified existing tests to cover new code paths, ensuring the threshold for code coverage.

<hr>

- `org.apache.iotdb.confignode.it.regionmigration.pass.commit` — 3 IT classes moved from `pass.daily.iotv1`:
  - `IoTDBRegionMigrateWithDeletionMultiDataDirIT`
  - `IoTDBRegionMigrateWithDeletionMultiDataDirTableIT`
  - `RegionMigrateFileAssertions`
- `IoTDBMigrateMultiRegionForIoTV1IT` — fix region colocation before multi-region migrate

**Test plan**
- [x] `IoTDBRegionMigrateWithDeletionMultiDataDirIT` (ClusterIT)
- [x] `IoTDBRegionMigrateWithDeletionMultiDataDirTableIT` (TableClusterIT)
- [x] `IoTDBMigrateMultiRegionForIoTV1IT` (ClusterIT)

### DIFF
--- a/integration-test/src/test/java/org/apache/iotdb/confignode/it/regionmigration/pass/commit/IoTDBMigrateMultiRegionForIoTV1IT.java
+++ b/integration-test/src/test/java/org/apache/iotdb/confignode/it/regionmigration/pass/commit/IoTDBMigrateMultiRegionForIoTV1IT.java
@@ -37,6 +37,7 @@ import org.slf4j.LoggerFactory;
 
 import java.sql.Connection;
 import java.sql.Statement;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -51,6 +52,7 @@ import static org.apache.iotdb.util.MagicUtils.makeItCloseQuietly;
 @RunWith(IoTDBTestRunner.class)
 public class IoTDBMigrateMultiRegionForIoTV1IT extends IoTDBRegionOperationReliabilityITFramework {
   private static final String MULTI_REGION_MIGRATE_FORMAT = "migrate region %s from %d to %d";
+  private static final String EXPAND_FORMAT = "extend region %d to %d";
 
   private static final Logger LOGGER =
       LoggerFactory.getLogger(IoTDBMigrateMultiRegionForIoTV1IT.class);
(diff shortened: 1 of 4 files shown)


integration-test/src/test/java/org/apache/iotdb/confignode/it/regionmigration/pass/commit/IoTDBMigrateMultiRegionForIoTV1IT.java
```diff
@@ -170,6 +176,71 @@ public void multiRegionMigrateTest() throws Exception {
...
+    }
+    List<Integer> regionIds = new ArrayList<>(regionMap.keySet());
+    Assert.assertTrue("Need at least two regions to colocate", regionIds.size() >= 2);
+
+    int firstRegion = regionIds.get(0);
+    int secondRegion = regionIds.get(1);
+    int targetDataNode = regionMap.get(secondRegion).iterator().next();
+    if (regionMap.get(firstRegion).contains(targetDataNode)) {
+      targetDataNode = regionMap.get(firstRegion).iterator().next();
+      regionGroupExpand(statement, client, secondRegion, targetDataNode);
```
jt2594838: What does this case indicate?

Some datanode of Region2 already contains Region1?

hongzhi-gao: We pick Region2's DataNode, then check if Region1 is already there. If not, extend Region1 to that node (the usual case); if yes, extend Region2 instead. With RF=1, the latter is defensive — colocated regions return early.

PR #18106, opened 2026-07-03
Reduce pipe load success log noise

Caideyipi:
## Description
- Lower high-frequency successful pipe-generated load, receiver, and load-tsfile analysis logs from INFO to DEBUG.
- Preserve INFO visibility for user-triggered load-tsfile paths where possible, and keep generic compaction/file-lifecycle logs unchanged.
- Lazily create and close the load tsfile dispatcher executor to avoid unnecessary thread pool churn on local load paths.

## Validation
- git diff --check


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/queryengine/plan/scheduler/load/LoadTsFileDispatcherImpl.java
```diff
@@ -97,36 +97,44 @@ public LoadTsFileDispatcherImpl(
...
+                  dispatchOneInstance(instance);
+                } catch (FragmentInstanceDispatchException e) {
+                  return new FragInstanceDispatchResult(e.getFailureStatus());
+                } catch (Exception t) {
+                  LOGGER.warn(DataNodeQueryMessages.CANNOT_DISPATCH_FI_FOR_LOAD_OPERATION, t);
+                  return new FragInstanceDispatchResult(
+                      RpcUtils.getStatus(
+                          TSStatusCode.INTERNAL_SERVER_ERROR,
+                          "Unexpected errors: " + t.getMessage()));
+                }
```
jt2594838, line 136: i18n

Caideyipi: Done. Added en/zh i18n constants for the unexpected-error status message and switched both dispatch paths to use them.

PR #18111, opened 2026-07-06
Pipe: use session context in legacy receiver loaders

Caideyipi:
### Legacy receiver session context

Legacy pipe receiver loading now uses the session context supplied by the RPC request path. The loader API accepts this context explicitly so deletion and tsfile loading follow the same request context.

### Loader behavior

TsFile loading now follows the configured auto-create setting. The legacy receiver no longer performs database registration during handshake.

### Tests

- `git diff --check`

<hr>

This PR has:
- [x] been self-reviewed.

<hr>

- Legacy pipe receiver agent and loaders.
- Client RPC pipe data transport path.
- Legacy pipe receiver integration test coverage.

### DIFF
--- a/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/receiver/protocol/legacy/IoTDBLegacyPipeReceiverAgent.java
+++ b/iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/receiver/protocol/legacy/IoTDBLegacyPipeReceiverAgent.java
@@ -21,26 +21,18 @@
 package org.apache.iotdb.db.pipe.receiver.protocol.legacy;
 
 import org.apache.iotdb.common.rpc.thrift.TSStatus;
-import org.apache.iotdb.commons.audit.UserEntity;
 import org.apache.iotdb.commons.conf.CommonDescriptor;
-import org.apache.iotdb.commons.exception.IllegalPathException;
 import org.apache.iotdb.commons.i18n.PipeMessages;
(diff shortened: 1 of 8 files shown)


iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/receiver/protocol/legacy/IoTDBLegacyPipeReceiverAgent.java
```diff
@@ -262,6 +203,15 @@ private SyncIdentityInfo getCurrentSyncIdentityInfo() {
     }
   }
 
+  private SessionInfo getCurrentSessionInfo() {
+    final SessionManager sessionManager = SessionManager.getInstance();
+    if (!sessionManager.checkLogin(sessionManager.getCurrSession())) {
+      throw new PipeException("Legacy pipe receiver requires a logged-in session.");
```
jt2594838: i18n

jt2594838 (approved):

iotdb-core/datanode/src/main/java/org/apache/iotdb/db/pipe/receiver/protocol/legacy/IoTDBLegacyPipeReceiverAgent.java
Caideyipi: Fixed, thanks. Moved the message into DataNodePipeMessages for both en and zh locales.

PR #18116, opened 2026-07-06
Fix pipe history LoadTsFile after deleting aligned measurements

Caideyipi:
## Summary
- Skip inactive aligned measurements when flushing `AlignedWritableMemChunk` to TsFile.
- Keep active schema columns mapped to their underlying TVList column indexes during both aligned flush paths.
- Add a minimal Pipe history IT for receiver `enable_auto_create_schema=false` with unflushed aligned `delete timeseries` before after-batch history sync.

## Root cause
`delete timeseries` on an aligned measurement removes it from the active measurement map and marks its TVList column deleted, but the flush writer still used the full historical `schemaList`. A history Pipe created before the memtable was flushed could therefore produce a TsFile whose metadata still contained the deleted measurement, without a corresponding mod file. The receiver then tried to verify/create a schema that should no longer exist, and `LoadTsFile` failed or retried forever when auto-create schema was disabled.

## Tests
- `mvn spotless:check -pl iotdb-core/datanode,integration-test -P with-integration-tests`
- `git diff --check`
- `mvn -pl integration-test -am -P with-integration-tests -DskipUTs -Dit.test=IoTDBPipeReceiverAutoCreateDisabledIT#testAutoSplitHistoryTsFileWithUnflushedAlignedDeletionWhenReceiverAutoCreateSchemaDisabled -DfailIfNoTests=false -Dfailsafe.failIfNoSpecifiedTests=false verify` *(blocked before the IT by existing upstream local compile failure in `iotdb-thrift`: generated thrift sources import missing `javax.annotation` on JDK 17)*

### DIFF
--- a/integration-test/src/test/java/org/apache/iotdb/pipe/it/dual/treemodel/auto/basic/IoTDBPipeReceiverAutoCreateDisabledIT.java
+++ b/integration-test/src/test/java/org/apache/iotdb/pipe/it/dual/treemodel/auto/basic/IoTDBPipeReceiverAutoCreateDisabledIT.java
@@ -182,6 +182,43 @@ public class IoTDBPipeReceiverAutoCreateDisabledIT extends AbstractPipeDualTreeM
         new HashSet<>(Arrays.asList("root.sg.aligned,true,null,INF,")));
   }
 
+  @Test
+  public void
+      testAutoSplitHistoryTsFileWithUnflushedAlignedDeletionWhenReceiverAutoCreateSchemaDisabled()
+          throws Exception {
+    TestUtils.executeNonQueries(
+        senderEnv,
+        Arrays.asList(
+            "create database root.sg",
+            "create aligned timeseries root.sg.aligned(s0 INT32, s1 INT64, s2 DOUBLE)",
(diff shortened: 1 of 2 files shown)


Caideyipi:
Follow-up for the table-model CI failures:

The failed `IoTDBWindowFunction2IT` / `IoTDBWindowTVF2IT` cases expose a table-model edge case: a table device can have rows whose value columns are all null, but the device/id row still needs to be persisted in TsFile metadata. My previous early return skipped encoding when the active aligned value schema list was empty, which is valid for tree model (`ignoreAllNullRows=true`) but invalid for table model (`ignoreAllNullRows=false`).

I updated the guard to return only when `activeSchemaList.isEmpty() && ignoreAllNullRows`, so tree model still drops all-null/deleted aligned devices while table model still writes the time/device metadata for all-null value rows.

Local validation:
- `mvn spotless:check -pl iotdb-core/datanode,integration-test -P with-integration-tests`
- `git diff --check`

I also tried to run the affected ITs locally, but the local build is still blocked before tests by the existing `iotdb-thrift` generated-source compile issue: missing `javax.annotation` under JDK 17.

jt2594838 (approved):

PR #18118, opened 2026-07-06
Extend region migration speed limit to transfers and deletion

CRZbulabula:
## Summary
- Extend `region_migration_speed_limit_bytes_per_second` to throttle region migration transfer across consensus protocols that have real transfer paths.
- Keep IoTConsensus AddPeer snapshot transfer covered by the shared limiter, and add coverage for IoTConsensusV2 consensus pipe transfer and Ratis install snapshot transfer.
- Share the same limiter with region file deletion and wire DataRegion, SchemaRegion, WAL, snapshot, and consensus peer cleanup paths through rate-limited deletion helpers.
- Update the config template description and add unit coverage for rate-limited deletion, IoTConsensusV2 transfer size calculation, and Ratis snapshot chunk size calculation.

## Protocol coverage
- IoTConsensus: AddPeer snapshot transfer is rate-limited.
- IoTConsensusV2: tablet, tablet batch, delete event, and TsFile piece transfer requests are rate-limited.
- Ratis: install snapshot file chunks are rate-limited before sending.
- SimpleConsensus: no remote region migration transfer path; local region deletion cleanup remains rate-limited.

## Tests
- `mvn spotless:apply -pl iotdb-core/node-commons,iotdb-core/consensus,iotdb-core/datanode`
- `mvn test -pl iotdb-core/node-commons -am -Dtest=FileUtilsTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false`
- `mvn test -pl iotdb-core/consensus -am -Dtest=RateLimitedGrpcLogAppenderTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false`
- `mvn test -pl iotdb-core/datanode -am -Dtest=IoTConsensusV2RateLimiterTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false`
- `mvn compile -pl iotdb-core/consensus,iotdb-core/datanode -am -DskipTests`
- `mvn test-compile -DskipTests -P with-zh-locale -pl iotdb-core/consensus,iotdb-core/datanode -am`

Note: full reactor `mvn test-compile -DskipTests` reaches the distribution module and fails to resolve existing 2.0.11-SNAPSHOT zip artifacts for iotdb-server, iotdb-cli, and iotdb-confignode. The changed modules compile successfully.


iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/utils/FileUtils.java
```diff
@@ -111,13 +112,30 @@ public static void deleteFileOrDirectory(File file) {
...
       File[] files = file.listFiles();
       if (files != null) {
         for (File subfile : files) {
-          deleteFileOrDirectory(subfile, quietForNoSuchFile);
+          deleteFileOrDirectory(subfile, quietForNoSuchFile, deleteRateLimiter);
         }
       }
+    } else if (deleteRateLimiter != null && file.isFile()) {
+      deleteRateLimiter.accept(file.length());
     }
```
jt2594838, line 139: Deleting a file only involves some file system metadata modification and inode reclaim, which will not introduce such io. 
The rate-limiting step may slow down the migration significantly.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/utils/FileUtils.java
```diff
@@ -161,7 +179,16 @@ public static void deleteFileOrDirectoryWithRetry(File file) {
...
+  public static void deleteDirectoryAndEmptyParentWithRateLimiter(
+      File folder, LongConsumer deleteRateLimiter) {
+    deleteDirectoryAndEmptyParent(folder, deleteRateLimiter);
+  }
+
+  private static void deleteDirectoryAndEmptyParent(File folder, LongConsumer deleteRateLimiter) {
+    deleteFileOrDirectory(folder, false, deleteRateLimiter);
     final File parentFolder = folder.getParentFile();
     File[] files = parentFolder.listFiles();
     if (parentFolder.isDirectory() && (files == null || files.length == 0)) {
```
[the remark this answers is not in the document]
CRZbulabula: Fixed in e84fe31364. `deleteDirectoryAndEmptyParent(...)` now treats `parentFolder == null` as no parent to clean up, and I added `testDeleteDirectoryAndEmptyParentWithRateLimiterAndNoParent` to cover the relative-path/no-parent case.

iotdb-core/node-commons/src/main/java/org/apache/iotdb/commons/utils/FileUtils.java
CRZbulabula: Agreed, using the real file size for delete throttling is too conservative because removal is mostly metadata work. Fixed in e84fe31364 by moving cleanup to an independent `region_migration_file_remove_speed_limit_bytes_per_second` limiter and charging fixed estimated remove costs instead: 64 KiB per regular file and 4 KiB per directory. I also removed the IoTConsensusV2 transfer limiter so regular consensus pipe replication is not throttled by the migration transfer limit.

jt2594838 (approved):

